PR #6892, opened 2025-05-19
feat(CLI): Improve `iota name` commands output

### LINKED ISSUE #6869: IOTA-Names commands that execute transactions should output more concise results
## Description

Currently, the commands which execute transactions simply return the transaction result, which is not very nice for users to read, and it can be unclear whether the operation succeeded. It would be better to pull the relevant information from this result and display a simplified response based on the input command.

DaughterOfMars:
# Description of change

Improves the console output of `iota name` commands by adding custom output variants instead of printing the transaction outputs.

## Example

```
./target/debug/iota name subdomain register-node node.chloe2.iota -e "2025-12-20 00:00:00" -c -t 
Registered Record:
╭────────────────┬────────────────────────────────────────────────────────────────────╮
│ field          │ value                                                              │
├────────────────┼────────────────────────────────────────────────────────────────────┤
│ NFT ID         │ 0x74b40e4553e781c028f86a6d54b03d39cc61ae54ea9fe4a34aff9b13ed1eb3cd │
│ Target Address │ <n/a>                                                              │
│ Expiration     │ 1766188800000 (2025-12-20 00:00:00.000000000 UTC)                  │
│ S_AC           │ 1                                                                  │
│ S_ATE          │ 1                                                                  │
╰────────────────┴────────────────────────────────────────────────────────────────────╯
Created NFT:
╭────────────┬────────────────────────────────────────────────────────────────────╮
│ field      │ value                                                              │
├────────────┼────────────────────────────────────────────────────────────────────┤
│ ID         │ 0x0ac7c36b902e60b8d4d342652413607f69d55dae9b8e0eea704564420c98b786 │
│ Domain     │ node.chloe2.iota                                                   │
│ Expiration │ 1766188800000 (2025-12-20 00:00:00.000000000 UTC)                  │
╰────────────┴────────────────────────────────────────────────────────────────────╯
```

Closes #6869

## Type of change

I have run the commands and viewed the output.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Improved output for `iota name` commands.
- [ ] Rust SDK:
- [ ] REST API:


Thoralf-M:
For the commands where we do a transaction we should at least display the transaction id, so one has an easy way to find all the information that might be interesting and is not covered by the custom output

DaughterOfMars:
> For the commands where we do a transaction we should at least display the transaction id, so one has an easy way to find all the information that might be interesting and is not covered by the custom output

Done!

crates/iota/src/name_commands.rs
```diff
@@ -1311,6 +1790,29 @@ async fn fetch_renewal_config(context: &mut WalletContext) -> anyhow::Result<Ren
...
+    Fun: FnOnce(IotaTransactionBlockResponse) -> F,
+    F: futures::Future<Output = anyhow::Result<NameCommandResult>>,
+{
+    Ok(
+        if let IotaClientCommandResult::TransactionBlock(res) = res {
+            if !res.errors.is_empty() {
+                for err in res.errors {
+                    println!("{err}");
+                }
+                bail!("transaction failed");
```
Thoralf-M: Shouldn't this rather be before the errors with `:`?

DaughterOfMars: Yeah, I was thinking the same, but wasn't sure the best way to do it. I'll think about it a bit

Thoralf-M (approved):

samuel-rufi (approved):
Looks good to me

thibault-martinez:
I think we should at least have a flag to revert to the usual output, is that possible?

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -57,35 +57,70 @@ pub struct Summary {
...
+impl PTBCommandResult {
+    pub fn to_styled_str(self) -> StyledStr {
+        StyledStr::from(self.to_string())
     }
+}
 
+impl std::fmt::Display for PTBCommandResult {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Preview(ptbpreview) => ptbpreview.to_string(),
```
thibault-martinez:
```suggestion
            Self::Preview(ptb_preview) => ptb_preview.to_string(),
```

crates/iota-names/src/registry.rs
```diff
@@ -72,6 +75,12 @@ pub struct NameRecord {
     pub data: VecMap<String, String>,
 }
 
+impl NameRecord {
```
thibault-martinez: There is already an impl block, can you group them please?

DaughterOfMars:
> I think we should at least have a flag to revert to the usual output, is that possible?

So technically it's already possible by calling the `--display` flag like so:
```
./target/debug/iota name unset-target-address chloe2.iota --display input
```

Which will output the requested item. So to get everything you would need to manually specify
```
--display input,effects,events,object_changes,balance_changes
```

crates/iota-names/src/registry.rs
DaughterOfMars: done

crates/iota/src/client_ptb/ptb.rs
DaughterOfMars: done

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2113,18 +2113,18 @@ impl From<InputObjectKind> for IotaInputObjectKind {
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema, Clone)]
 #[serde(rename = "TypeTag", rename_all = "camelCase")]
-pub struct IotaTypeTag(String);
+pub struct IotaTypeTag(pub String);
```
thibault-martinez: I think I'd rather have a getter/AsRef/etc than a public field, to avoid having it inadvertently changed

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -45,9 +45,9 @@ pub struct PTB {
     pub display: HashSet<DisplayOption>,
 }
 
-pub struct PTBPreview<'a> {
```
thibault-martinez, line 48: Just curious, can you elaborate on the ownership change please?

DaughterOfMars: It was just an unnecessary lifetime when this data can be owned. Otherwise I would have to propagate it upward in the containing enums

crates/iota/src/client_commands.rs
```diff
@@ -1705,10 +1705,17 @@ impl IotaClientCommands {
...
-                IotaClientCommandResult::NoOutput
-            }
+            IotaClientCommands::PTB(ptb) => match ptb.execute(context).await? {
+                PTBCommandResult::CommandResult(iota_client_command_result) => {
+                    iota_client_command_result
+                }
+                res => {
+                    let s = res.to_styled_str();
+                    println!("{}", s.ansi());
+                    println!("{s}");
```
thibault-martinez: Why both prints?

crates/iota/tests/cli_tests.rs
```diff
@@ -4706,13 +4706,20 @@ async fn test_ptb_dev_inspect() -> Result<(), anyhow::Error> {
...
     .execute(context)
-    .await?;
-    assert!(res.contains("Execution Result\n  Return values\n    IOTA TypeTag: IotaTypeTag(\"0x1::string::String\")\n    Bytes: [5, 72, 101, 108, 108, 111]"));
+    .await?
+    else {
+        panic!("unexpected PTB result");
+    };
+    assert!(res.results.expect("missing results").iter().any(|res| {
+        res.return_values.iter().any(|(bytes, tag)| {
+            tag.as_ref() == "0x1::string::String" && bytes == &[5, 72, 101, 108, 108, 111]
```
thibault-martinez, line 4720: A pretty nice improvement

crates/iota/tests/cli_tests.rs
```diff
@@ -4728,29 +4735,37 @@ async fn test_ptb_display_args() -> Result<(), anyhow::Error> {
...
+        .execute(context)
+        .await?
+    else {
+        panic!("unexpected PTB result");
+    };
 
-    assert!(res.contains("Transaction Data"));
-    assert!(res.contains("Transaction Effects"));
+    assert!(res.transaction.is_some());
+    assert!(res.effects.is_some());
```
thibault-martinez, line 4750: Overall better to have to have the result than a string tbh

thibault-martinez, line 4750: Btw, why is this some even though we didn't pass `DisplayOption::Effects`?

crates/iota/src/client_commands.rs
DaughterOfMars: Well it's two lines 🤷‍♀️

crates/iota/tests/cli_tests.rs
DaughterOfMars: They are on the result because it defaults to all

crates/iota/src/client_commands.rs
thibault-martinez: But why the second one?

thibault-martinez: Aren't they kind of the same thing? What am I missing?

DaughterOfMars: Two newlines

Alex6323 (approved):

crates/iota/tests/cli_tests.rs
thibault-martinez: Sorry I'm not following, wouldn't it default to all if nothing was provided? We are providing `DisplayOption::Input` here

DaughterOfMars: Oh, I see. Hmm. Well it was like that before too, but I don't know. It is a bit weird.

thibault-martinez: Yeah I know it's not you, I just noticed when reviewing, not sure why this is passing the tests

thibault-martinez: Opened https://github.com/iotaledger/iota/issues/6935, let's have a look later

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -57,35 +57,70 @@ pub struct Summary {
...
+                Pretty(dev_inspect_results.as_ref()).to_string()
+            }
+            Self::Json(value) => {
+                serde_json::to_string_pretty(&value).map_err(|_| std::fmt::Error)?
+            }
+            Self::Help { long } => {
+                let mut buf = Vec::new();
+                if *long {
+                    ptb_description()
+                        .write_long_help(&mut buf)
```
thibault-martinez: Why not keep using render_help/render_long_help?

DaughterOfMars: Because the print directly to the console and this is a display impl

thibault-martinez: I don't think they do, no. `write_help` and `write_long_help` are actually deprecated and they say to use `render_help` and `render_long_help` instead.

DaughterOfMars: ah, you are correct, sorry. I misremembered. I did it just because of the return type. Fixed it anyway.

thibault-martinez:
> > I think we should at least have a flag to revert to the usual output, is that possible?
> 
> So technically it's already possible by calling the `--display` flag like so:
> 
> ```
> ./target/debug/iota name unset-target-address chloe2.iota --display input
> ```
> 
> Which will output the requested item. So to get everything you would need to manually specify
> 
> ```
> --display input,effects,events,object_changes,balance_changes
> ```

Have you tried it? I don't think it works in combination with your new stuff.

DaughterOfMars:
> > > I think we should at least have a flag to revert to the usual output, is that possible?
> > 
> > 
> > So technically it's already possible by calling the `--display` flag like so:
> > ```
> > ./target/debug/iota name unset-target-address chloe2.iota --display input
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which will output the requested item. So to get everything you would need to manually specify
> > ```
> > --display input,effects,events,object_changes,balance_changes
> > ```
> 
> Have you tried it? I don't think it works in combination with your new stuff.

Yes, I did

thibault-martinez:
> > Have you tried it? I don't think it works in combination with your new stuff.
> 
> Yes, I did

```
thibault@Mac ~/i/iota (dev-tools/better-name-command-output)> ./target/debug/iota name subdomain register-node test3.phone.iota --display input
[warn] Client/Server api version mismatch, client api version: 1.2.0-alpha, server api version: 1.1.0
Registered Record:
╭────────────────┬────────────────────────────────────────────────────────────────────╮
│ field          │ value                                                              │
├────────────────┼────────────────────────────────────────────────────────────────────┤
│ NFT ID         │ 0x557756520a4e23c85cf287bf50d1d76d115416468684d21431e365777ec2ab7f │
│ Target Address │ <n/a>                                                              │
│ Expiration     │ 1777993652616 (2026-05-05 15:07:32.616000000 UTC)                  │
╰────────────────┴────────────────────────────────────────────────────────────────────╯
Created NFT:
╭────────────┬────────────────────────────────────────────────────────────────────╮
│ field      │ value                                                              │
├────────────┼────────────────────────────────────────────────────────────────────┤
│ ID         │ 0xaa474179baef1e1750a09f370d92ec6e88cba9e99ef38bd02d8508109db40302 │
│ Domain     │ test3.phone.iota                                                   │
│ Expiration │ 1777993652616 (2026-05-05 15:07:32.616000000 UTC)                  │
╰────────────┴────────────────────────────────────────────────────────────────────╯
Transaction digest: BHiAwinKVmJc4GP7xvq9yQKi4M8akXR659YG9ErtKmzJ
```

What am I missing?

crates/iota/src/name_commands.rs
```diff
@@ -1311,6 +1790,26 @@ async fn fetch_renewal_config(context: &mut WalletContext) -> anyhow::Result<Ren
...
+    res: IotaClientCommandResult,
+    fun: Fun,
+) -> anyhow::Result<NameCommandResult>
+where
+    Fun: FnOnce(IotaTransactionBlockResponse) -> F,
+    F: futures::Future<Output = anyhow::Result<NameCommandResult>>,
+{
+    Ok(
+        if let IotaClientCommandResult::TransactionBlock(res) = res {
+            if !res.errors.is_empty() {
```
thibault-martinez, line 1881: I'm having a hard time wrapping my head around this check, can you elaborate please? Where was it done before?

crates/iota/src/name_commands.rs
```diff
@@ -507,65 +558,86 @@ impl NameCommand {
...
+                        IotaJsonValue::new(serde_json::to_value(address)?)?,
+                    ],
+                    gas_price: None,
+                    opts,
+                }
+                .execute(context)
+                .await?;
+
+                handle_transaction_result(res, async |res| {
+                    Ok(NameCommandResult::Transfer {
```
thibault-martinez, line 616: We are not showing more than the parameters here, I would revert to showing the usual output

crates/iota/src/name_commands.rs
```diff
@@ -972,23 +1090,30 @@ impl SubdomainCommand {
...
+                    ],
+                    gas_price: None,
+                    opts: opts.clone(),
+                }
+                .execute(context)
+                .await?;
+
+                handle_transaction_result(res, async |res| {
+                    Ok(NameCommandResult::ExtendExpiration {
+                        nft: get_owned_nft_by_name::<SubdomainRegistration>(&domain, context)
```
thibault-martinez, line 1176: I wonder if we also want to print the record entry since the expiration is also there (assuming it's changed in both)

crates/iota/src/name_commands.rs
```diff
@@ -1062,13 +1304,88 @@ impl std::fmt::Display for NameCommandResult {
                     write!(f, "no target address found for '{domain}'")
                 }
             }
+            Self::Register {
+                record,
+                nft,
+                digest: transaction,
+            } => {
+                writeln!(f, "Registered Record:")?;
```
thibault-martinez: Here and following, I would not capitalise Record

crates/iota/src/name_commands.rs
```diff
@@ -1062,13 +1304,88 @@ impl std::fmt::Display for NameCommandResult {
...
+                format_registry_entry(f, entry)?;
+                write!(f, "\nTransaction digest: {transaction}")
+            }
+            Self::SetUserData {
+                key,
+                value,
+                record,
+                digest: transaction,
+            } => {
+                writeln!(f, "Successfully set user data {key} to {value}")?;
```
thibault-martinez:
```suggestion
                writeln!(f, "Successfully set user data \"{key}\" to \"{value}\"")?;
```

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+            } => {
+                writeln!(f, "Successfully unset reverse lookup for {address}")?;
+                write!(f, "Transaction digest: {transaction}")
+            }
+            Self::UnsetUserData {
+                key,
+                record,
+                digest: transaction,
+            } => {
+                writeln!(f, "Successfully unset key {key}")?;
```
thibault-martinez:
```suggestion
                writeln!(f, "Successfully unset key \"{key}\"")?;
```

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+            .to_string();
+
+    let data = [
+        ("Start", start_datetime.to_string()),
+        ("End", end_datetime.to_string()),
+        (
+            "Current Bid",
+            auction.current_bid.balance.value().to_string(),
+        ),
+        ("Latest Bidder", auction.current_bid.id.id.bytes.to_string()),
```
thibault-martinez: What is this supposed to be? Print `winner` instead?

DaughterOfMars:
> > > Have you tried it? I don't think it works in combination with your new stuff.
> > 
> > 
> > Yes, I did
> 
> ```
> thibault@Mac ~/i/iota (dev-tools/better-name-command-output)> ./target/debug/iota name subdomain register-node test3.phone.iota --display input
> [warn] Client/Server api version mismatch, client api version: 1.2.0-alpha, server api version: 1.1.0
> Registered Record:
> ╭────────────────┬────────────────────────────────────────────────────────────────────╮
> │ field          │ value                                                              │
> ├────────────────┼────────────────────────────────────────────────────────────────────┤
> │ NFT ID         │ 0x557756520a4e23c85cf287bf50d1d76d115416468684d21431e365777ec2ab7f │
> │ Target Address │ <n/a>                                                              │
> │ Expiration     │ 1777993652616 (2026-05-05 15:07:32.616000000 UTC)                  │
> ╰────────────────┴────────────────────────────────────────────────────────────────────╯
> Created NFT:
> ╭────────────┬────────────────────────────────────────────────────────────────────╮
> │ field      │ value                                                              │
> ├────────────┼────────────────────────────────────────────────────────────────────┤
> │ ID         │ 0xaa474179baef1e1750a09f370d92ec6e88cba9e99ef38bd02d8508109db40302 │
> │ Domain     │ test3.phone.iota                                                   │
> │ Expiration │ 1777993652616 (2026-05-05 15:07:32.616000000 UTC)                  │
> ╰────────────┴────────────────────────────────────────────────────────────────────╯
> Transaction digest: BHiAwinKVmJc4GP7xvq9yQKi4M8akXR659YG9ErtKmzJ
> ```
> 
> What am I missing?

Ah, I think this worked before I fixed the ordering but doesn't anymore. I'll check it out

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+    ];
+    let mut table_builder = Table::builder(data);
+    table_builder.set_header(["field", "value"]);
```
thibault-martinez, line 1528: IMO these field/value wrappers are not necessary, bit verbose for not much

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+}
+
+fn build_name_record_table(table_builder: &mut TableBuilder, record: &NameRecord) {
+    table_builder.push_record(["NFT ID", record.nft_id.bytes.to_string().as_str()]);
+    table_builder.push_record([
+        "Target Address",
+        record
+            .target_address
+            .map(|address| address.to_string())
+            .unwrap_or_else(|| "<n/a>".to_owned())
```
thibault-martinez:
```suggestion
            .unwrap_or_else(|| "none".to_owned())
```
I don't think it's really Non Applicable 😬

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+        tabled::settings::Style::rounded().horizontals([HorizontalLine::new(
+            1,
+            TableStyle::modern().get_horizontal(),
+        )]),
+    );
+    write!(f, "{table}")
+}
+
+fn build_name_record_table(table_builder: &mut TableBuilder, record: &NameRecord) {
+    table_builder.push_record(["NFT ID", record.nft_id.bytes.to_string().as_str()]);
```
thibault-martinez, line 1593: Just curious, ID itself doesn't impl Display?

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1403,198 @@ impl std::fmt::Display for NameCommandResult {
...
+    ]);
+
+    for entry in &record.data.contents {
+        table_builder.push_record([&entry.key, &entry.value]);
     }
 }
 
+fn format_nft(f: &mut std::fmt::Formatter, nft: &IotaNamesRegistration) -> std::fmt::Result {
+    let expiration_datetime = DateTime::<Utc>::from(nft.expiration_time())
+        .format("%Y-%m-%d %H:%M:%S.%f UTC")
```
thibault-martinez, line 1619: We are doing these so much, maybe we should factorize (another PR maybe)

crates/iota/src/name_commands.rs
DaughterOfMars: It's very verbose for no real reason IMO

crates/iota/src/name_commands.rs
DaughterOfMars: 🤷‍♀️ up to you I guess

crates/iota/src/name_commands.rs
DaughterOfMars: But it's not the winner necessarily

crates/iota/src/name_commands.rs
DaughterOfMars: it's not possible to have a table without a header afaict so this is what I compromised with

crates/iota/src/name_commands.rs
DaughterOfMars: nope

crates/iota/src/name_commands.rs
DaughterOfMars: It wasn't, this is part of the improvement

crates/iota/src/name_commands.rs
thibault-martinez: It's definitely the coin of the winner, I would at least show the winner and rename this field because it's not the "bidder", it's the bidder's coin ID

DaughterOfMars: I mean the auction may not be over so this is just the latest bidder

thibault-martinez: Sure, doesn't change the fact that there is a winner field that we need to show + this is a bid, not a bidder

crates/iota/src/name_commands.rs
```diff
@@ -1311,6 +1854,30 @@ async fn fetch_renewal_config(context: &mut WalletContext) -> anyhow::Result<Ren
...
+    res: IotaClientCommandResult,
+    verbose: bool,
+    fun: Fun,
+) -> anyhow::Result<NameCommandResult>
+where
+    Fun: FnOnce(IotaTransactionBlockResponse) -> F,
+    F: futures::Future<Output = anyhow::Result<NameCommandResult>>,
+{
+    if verbose {
+        println!("{res}");
```
thibault-martinez:
```suggestion
        println!("{res}\n");
```
Easier to differentiate both outputs. Not sure if that's the best place to put it though, I'll leave it to you.

crates/iota/src/name_commands.rs
thibault-martinez: moot point now that we have the verbose flag

crates/iota/src/name_commands.rs
thibault-martinez: Yeah let's do it

crates/iota/src/name_commands.rs
thibault-martinez: Hmm alright, that sucks

crates/iota/src/name_commands.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/6953

crates/iota/src/name_commands.rs
DaughterOfMars: haha oh, I see. not sure how that happened.

crates/iota/src/name_commands.rs
```diff
@@ -1086,36 +1474,198 @@ impl std::fmt::Display for NameCommandResult {
...
+            .to_string();
+
+    let data = [
+        ("Start", start_datetime.to_string()),
+        ("End", end_datetime.to_string()),
+        (
+            "Current Bid",
+            auction.current_bid.balance.value().to_string(),
+        ),
+        ("Latest Bidder", auction.winner.to_string()),
```
thibault-martinez:
```suggestion
        ("Winner", auction.winner.to_string()),
```

crates/iota/src/name_commands.rs
thibault-martinez: Surely at some point we were printing if the transaction failed?

crates/iota/src/name_commands.rs
DaughterOfMars: How about this, I will print winner if the auction is over, otherwise I will print Bidder. Good?

crates/iota/src/name_commands.rs
DaughterOfMars: It may be part of the regular output but it probably doesn't look nice

crates/iota/src/name_commands.rs
thibault-martinez: Not good. This is currently the name of the field, let's not introduce a second name and be confusing. Feel free to open an issue on the SC repo if you want to change it globally.

DaughterOfMars: But it can be either thing depending on whether the auction is over

thibault-martinez: Well don't ask me if you are not waiting for my answer?

thibault-martinez: I disagree, winner is enough, it does not assume the auction is over. I would revert your last commit.

DaughterOfMars: I like it better this way. It's more clear to the user when the auction is still going vs over

thibault-martinez: It does not really matter if you like it or not, creating inconsistency is never good for the end user. We shouldn't deviate from the field's name. Add a little something next to the expiration if you want.

crates/iota/src/name_commands.rs
thibault-martinez: Alright

crates/iota/src/name_commands.rs
DaughterOfMars: @Thoralf-M @Alex6323 what do you guys think?

crates/iota/src/name_commands.rs
```diff
@@ -1286,6 +1286,7 @@ impl std::fmt::Display for NameCommandResult {
                 digest: transaction,
             } => {
                 writeln!(f, "Successfully placed bid for {}", auction.domain)?;
+                writeln!(f, "Auction Status: {}", auction.status())?;
```
thibault-martinez: Think I would have put it in format_auction but that's fine

crates/iota/src/name_commands.rs
```diff
@@ -997,51 +1187,134 @@ impl SubdomainCommand {
...
 impl std::fmt::Display for NameCommandResult {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::Auction(client_result) => client_result.fmt(f),
+            Self::AuctionBid {
+                auction,
+                digest: transaction,
+            } => {
+                writeln!(f, "Successfully placed bid for {}", auction.domain)?;
+                writeln!(f, "Auction Status: {}", auction.status())?;
```
thibault-martinez:
```suggestion
                writeln!(f, "Auction status: {}", auction.status())?;
```

crates/iota/src/name_commands.rs
```diff
@@ -997,51 +1187,134 @@ impl SubdomainCommand {
...
-                    f,
-                    "start:\t{start_datetime}\n\
-                    end:\t{end_datetime}\n\
-                    winner:\t{}\n\
-                    bid:\t{} ({})",
-                    auction.winner,
-                    auction.current_bid.balance.value(),
-                    auction.current_bid.id.id.bytes
-                )
+                writeln!(f, "Auction Status: {}", auction.status())?;
```
thibault-martinez:
```suggestion
                writeln!(f, "Auction status: {}", auction.status())?;
```

crates/iota/src/name_commands.rs
```diff
@@ -997,51 +1187,134 @@ impl SubdomainCommand {
...
+                format_auction(f, auction)
+            }
+            Self::AuctionStart {
+                auction,
+                digest: transaction,
+            } => {
+                writeln!(f, "Successfully started auction for {}", auction.domain)?;
+                writeln!(f, "Auction Status: {}", auction.status())?;
```
thibault-martinez:
```suggestion
                writeln!(f, "Auction status: {}", auction.status())?;
```

thibault-martinez (approved):

PR #6900, opened 2025-05-19
fix(wallet): Ledger Nano S address verification in popup

### LINKED ISSUE #6610: Web wallet ledger nano address verification
## Bug description

Verifying an address for a ledger nano account fails when opening just the wallet popup
Image

If I open the wallet page at chrome-extension://iidjkmdceolghepehaaddojmnjnkkija it works fine

Image

## Steps To reproduce the bug

1. Create a ledger nano profile
2. Click on the iota extension at the top right
3. Click on Receive
4. Click on "Verify Address"

## Expected behaviour

Address verification should work or is this a limitation of the browser and the extension can't communicate with the ledger directly? Then there should be another error to let the user know that this only works in the page chrome-extension://iidjkmdceolghepehaaddojmnjnkkija

## Actual behaviour

`Address verification failed!`

msarcev:

Closes https://github.com/iotaledger/iota/issues/6610


marc2332 (commented):
mind sharing a brief explanation on why fixes it? Like, why does "asking permission first" fail?

msarcev:
> mind sharing a brief explanation on why fixes it? Like, why does "asking permission first" fail?

I found that the current state with `requestPermissionsFirst = true` triggers the browser popup where you select a "Ledger" HID device.

Indeed like @Thoralf-M suspected in the task description, this doesn't work (not supported) in the extension "popup mode" (or at least i wasn't able to make it work in a couple hours of trying). 
The part that fails in popup mode is in `TransportWebHID.request()` call
```
  static async request() {
    const [device] = await requestLedgerDevices();
    return TransportWebHID.open(device);
  }
```
The behaviour is that it skips the "device selection" and returns an empty array.

Then i realised that we do use the ledger in "popup mode", you can obviously sign transaction and it works. All those calls are done with `requestPermissionsFirst = false` (default value) which basically calls directly `TransportWebHID.openConnected()`. This one just gets the connected ledger devices to WebHID and uses the first one.

In the end the reasoning for this is, if we don't `requestPermissionsFirst` when you have a ledger account imported in the wallet in a previous session and you go to sign a transaction, getting a client to call `ledgerClient.getPublicKey` (which triggers the address validation) also can get away without the HID popup.

evavirseda (approved):
works 
image

marc2332 (approved):

PR #6904, opened 2025-05-19
feat(core): [TrafficControl] Support allowlisting

nmrshll:
## Port [MystenLabs/sui@c76e196](https://github.com/MystenLabs/sui/commit/c76e196213c550b6cfec2b7d3d635cd70cc8a454) : [TrafficControl] Support allowlisting 
Upstream range: [1.34.2, 1.35.4) 
### Upstream PR(s): 
- https://github.com/MystenLabs/sui/pull/19242  

## Description 

In some cases, we may want to enable a more restrictive policy wherein the node must explicitly specify all IP's from which it will accept requests. Because `TrafficController` policies do not easily support allowlisting, this is instead supported by introducing a separate mode of operation for traffic controller, enabled by providing `allow_list` field in the `PolicyConfig`, which should map to a list of strings all parseable to `IpAddr`. 

When this config is present, we skip spawning a tally thread and ignore all calls to `TrafficController::tally`, and instead initialize an in memory allowlist of IPs. On subsequent calls to `TrafficController::check`, we check this list against the requestor IP. 

Dry run mode in this mode still works as expected, as do block metrics (any request that is not in the allowlist is tallied against the block metric).


crates/iota-core/src/authority_server.rs.orig
muXxer, line 1: Where did that came from? Remove the file.

crates/iota-json-rpc/src/axum_router.rs.orig
muXxer, line 1: Where did that came from? Remove the file.

crates/iota-e2e-tests/tests/traffic_control_tests.rs.orig
muXxer, line 1: Where did that came from? Remove the file.

crates/iota-core/src/traffic_controller/mod.rs.orig
muXxer, line 1: Where did that came from? Remove the file.

muXxer (changes requested):

muXxer (approved):

Issue #3990, opened 2024-11-11
Port upstream node team related changes [mainnet-v1.32.2..mainnet-v1.36.2)


muXxer:
- Ask other teams how they created the report (format, where etc).
- Ask other teams how they track upstream changes (how do they mark critical ones etc)
- We should remember the latest hash of the commits we checked, so we don't redo the same work
- Does it make sense to use the slipstream script?

## mainnet-v1.32.2 to mainnet-v1.36.2
Commits from 2024-09-05 to 2024-10-26.
Commit hash from [bb77882](https://github.com/MystenLabs/sui/commit/bb778828e36d53a7d91a27e55109f2f45621badc) to [3ada97c](https://github.com/MystenLabs/sui/commit/3ada97c109cc7ae1b451cb384a1f2cfae49c8d3e)
Upstream change details: [here](https://iotafoundation.atlassian.net/wiki/spaces/OS/pages/2311618707/mainnet+v1.32.2+~+v1.36.2)

### Commits to track from mainnet-v1.32.2 to mainnet-v1.36.2
- [x] (🚧 Bing) Check changes in upstream crates/sui-protocol-config (18)
- [x] (🚧 Bing) Check changes in upstream crates/sui-rest-api (8)
- [x] (🚧 Vivian) Check changes in upstream crates/sui-snapshot (1)
- [x] (🚧 Vivian) Check changes in upstream crates/sui-config (22)
- [x] (🚧 Vivian) Check changes in upstream crates/sui-archival (1)
- [x] (🚧 Vivian) Check changes in upstream crates/sui-authority-aggregation (2)
- [x] (🚧 Vivian) Check changes in upstream docker (5)
- [x] (🚧 Gokhan) Check changes in upstream crates/sui-node (7)
- [x] (🚧 Gokhan) Check changes in upstream crates/sui-network (2)
- [x] (🚧 Bing) Check changes in upstream crates/sui-core (66)
- [x] (🚧 Vivian) Check changes in upstream crates/sui-types (19)
- [x] (🚧 Vivian) cross crates commit (52)

bingyanglin:
### The following is the flow provided by @thibault-martinez 🙏 

- Get the list of commits for the crates between two dates. An example to find all the commits of November in 5 different crates: 
[...]

PR #3366, opened 2024-10-16
remove(node): remove `simplified_unwrap_then_delete` feature flag

muXxer:
# Description of change

This PR removes the deprecated protocol feature flag `simplified_unwrap_then_delete`. 
We don't need the backwards compatibility to former protocol versions.

## AuthorityStore
- removed `get_object_ref_prior_to_key`, not needed anymore
- removed `get_object_ref_prior_to_key_deprecated`, as it was deprecated and is not needed anymore
- removed `maybe_reaccumulate_state_hash`, because we don't need to recalculate the state hash for old versions
- `iter_live_object_set`, `range_iter_live_object_set` do not include `wrapped_objects` anymore

## AuthorityStorePruner
- killswitch for tombstone pruning was removed, tombstones are now always pruned

## AccumulatorStore
- `accumulate_effects_v1` was removed, long live `accumulate_effects_v2`
- `accumulate_cached_live_object_set_for_testing`, `accumulate_live_object_set`, `digest_live_object_set` do not include `wrapped_tombstones` anymore

## Links to any relevant issues

#3253

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Change checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #6907, opened 2025-05-20
upstream(node): Enable writeback cache by default

bingyanglin:
# Description of change

- Upstream range: [v1.34.2, v1.35.4)
- Port Sui's commit [Enable writeback cache by default (#19508)](https://github.com/MystenLabs/sui/commit/8761bf8b70de387bf9ec35025493688482793c60) and [Remove debug output (https://github.com/MystenLabs/sui/pull/19730)](https://github.com/MystenLabs/sui/commit/8ecb26f2e7115930b9931f382b87367bca1514b5) 
- Writeback cache has been stable on all mysten fullnodes plus our testnet validator for several months. Time to enable it by default everywhere
- Note that `iter_live_object_set` and `accumulate_cached_live_object_set_for_testing` do not include `wrapped_objects` anymore, which was addressed in https://github.com/iotaledger/iota/pull/3366

Part of #3990 

## Type of change

Ran the local network with `RUST_LOG=info cargo run --release --bin iota start --force-regenesis --with-faucet`

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -260,14 +260,23 @@ pub struct NodeConfig {
     pub iota_names_config: Option<IotaNamesConfig>,
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, Default)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub enum ExecutionCacheConfig {
-    #[default]
     PassthroughCache,
     WritebackCache {
+        /// Maximum number of entries in each cache. (There are several
+        /// different caches). If None, the default of 10000 is used.
         max_cache_size: Option<usize>,
     },
+}
+
+impl Default for ExecutionCacheConfig {
+    fn default() -> Self {
+        ExecutionCacheConfig::WritebackCache {
+            max_cache_size: None,
+        }
+    }
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -4829,7 +4829,8 @@ impl AuthorityState {
     pub(crate) fn iter_live_object_set_for_testing(
         &self,
     ) -> impl Iterator<Item = authority_store_tables::LiveObject> + '_ {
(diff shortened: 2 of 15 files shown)


muXxer (approved):

piotrm50 (approved):

jkrvivian (approved):

muXxer:
We need to update the feature branch PR description after merging this.
Does it affect the protocol parameters?

bingyanglin:
> Does it affect the protocol parameters?

I think **yes**.
It affects the default value of protocol parameter `max_cache_size`, and we need to update [snapshot_tests__network_config_snapshot_matches.snap](https://github.com/iotaledger/iota/pull/6907/files#diff-f6e7960311e49daa15c45695a187d049750e17705baad5392898090fa56e7557) to make the CI pass.

Thoralf-M (approved):

muXxer:
> > Does it affect the protocol parameters?
> 
> I think **yes**. It affects the default value of protocol parameter `max_cache_size`, and we need to update [snapshot_tests__network_config_snapshot_matches.snap](https://github.com/iotaledger/iota/pull/6907/files#diff-f6e7960311e49daa15c45695a187d049750e17705baad5392898090fa56e7557) to make the CI pass.

But I think this is not a protocol parameter, but a node config parameter. Thats why the snapshot file in the `iota-swarm-config/tests/snapshots/` changes, but not in the `iota-protocol-config/src/snapshots/`.
So I guess this is a non-breaking change. As far as I understand, these `EpochFlags` are used locally to remember what kind of configuration was used in that epoch by that node, so that for example you do not suddenly enable the writeback cache for old epochs, just because you changed that value now. It remembers the flags per epoch and stores them in the database.

bingyanglin:
> But I think this is not a protocol parameter, but a node config parameter. Thats why the snapshot file in the iota-swarm-config/tests/snapshots/ changes, but not in the iota-protocol-config/src/snapshots/.
> So I guess this is a non-breaking change. As far as I understand, these EpochFlags are used locally to remember what kind of configuration was used in that epoch by that node, so that for example you do not suddenly enable the writeback cache for old epochs, just because you changed that value now. It remembers the flags per epoch and stores them in the database.

Make sense to me. Thank for the explanation.

PR #6918, opened 2025-05-20
fix(iota-data-ingestion-core): make file updates atomic

### LINKED ISSUE #6889: [`iota-data-ingestion-core`]: FileProgressStore updates are not atomic
## Bug description

Currently, updating the file by calling `set_len(0)` followed by `write_all` is not atomic. If the process crashes between truncating and writing, the file may be left empty, resulting in loss of watermark data.

Using `OpenOptions::truncate(true)` is also insufficient, as a crash after truncation but before writing will similarly leave the file empty.

**Proposed Solution:**  
Adopt the `tempfile + rename` pattern. This approach writes new data to a temporary file first, and only replaces the original file with an atomic rename if the write succeeds. This ensures that the original file remains intact unless the entire update completes successfully, greatly reducing the risk of data loss due to partial writes or crashes.

sergiupopescu199:
# Description of change

This PR refactors the `FileProgressStore` to ensure atomic updates when storing watermark data.  
It leverages a temporary file to write the new data, and then uses an atomic rename operation to replace the original file with the updated version.
  
This approach guarantees that the watermark file is never left in a partially written or corrupted state, even in the event of a crash or unexpected shutdown, ensuring data integrity and reliability.

fixes #6889 

## Type of change

- Added 2 test cases to verify the results.
 ```shell
```

    - Existing tests passed 
    - Added 2 tests in the `iota-data-ingestion-core` crate that verifies the actual patch changes
    -  `file_progress_store_save_timeout_simulates_crash` & `file_progress_store`

> [!NOTE]
> The other checkboxes were removed since this patch does not directly affect the indexer performance, nevertheless running locally the ingestion framework for checkpoint processing risulted in the same performance as before, not visible regression were spotted.

## Change checklist

- [x] I have followed the contribution guidelines for this project

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### CODE REVIEW

File: crates/iota-data-ingestion-core/src/progress_store/file.rs [hunk 1]
```diff
@@ -95,11 +98,30 @@ impl FileProgressStore {
...
+            if cfg!(test) {
+                tokio::time::sleep(std::time::Duration::from_nanos(10)).await;
+            }
+        }
+
+        // Atomically replace the original file
+        tokio::fs::rename(&tmp_path, &self.path).await?;
+
+        // Re-open the file handle for further reads
+        self.file = Self::open_or_create_file(&self.path).await?;
```
kodemartin: At this point we know that the file exists, so why not open it directly and not use `open_or_create_file`?

It would avoid breaking the signature of `open_or_create_file` above.

kodemartin (approved):
looks good with a remark

crates/iota-data-ingestion-core/src/progress_store/file.rs
sergiupopescu199: Mostly because I wanted to reuse the already present logic, but thinking now it can be misleading. Will follow with a patch shortly.

The `open_or_create_file` is a private function so I suppose is not critical if we change its signature, internally it needs a ref of `PathBuf` avoiding cloning it in the `FileProgressStore::new` method.

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/6918/commits/36aa17e99dcef30c86c6b2869a56a395a3e7d380

crates/iota-data-ingestion-core/src/progress_store/file.rs
```diff
@@ -95,11 +98,30 @@ impl FileProgressStore {
...
+
+        {
+            let mut tmp_file = OpenOptions::new()
+                .write(true)
+                .create(true)
+                .truncate(true)
+                .open(&tmp_path)
+                .await?;
+            tmp_file.write_all(data.as_ref()).await?;
+
```
samuel-rufi, line 112: Should there be a `tmp_file.sync_all().await?;` after the `write_all` to ensure data is on disk?
https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_all

Without it, isn't there a risk that `rename()`could  point to a file that was never flushed to disk if the system crashes?

samuel-rufi (commented):
Wondering about the need of `sync_all`, please see below:

crates/iota-data-ingestion-core/src/progress_store/file.rs
sergiupopescu199: `sync_all()` was quite slow during testing (from nano to millis), but as a tradeoff we could use the [sync_data()](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_data) since we only need the file content and not its metadata, is a little bit faster. 

I guess is a good tradeoff for having robust data storage during crashes, what do you think?

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/6918/commits/9fca597f1dc3ac4fa2069be048d4deaf4593eaf3

samuel-rufi (approved):

PR #6928, opened 2025-05-20
upstream(node): Add mode to support signed variant of Discovery protocol

nmrshll:
## Port [MystenLabs/sui@96b5a14](https://github.com/MystenLabs/sui/commit/96b5a149dff31731453baf59fe67e33ccb9fdd27): Add mode to support signed variant of Discovery protocol. 
Upstream range: [1.34.2, 1.35.4) 
### Upstream PR(s): 
- https://github.com/MystenLabs/sui/pull/19587  

## Description 

Adds a new version of discovery where all NodeInfo messages are signed by the originator.


crates/iota/validator.info
Thoralf-M, line 1: Did you commit this by accident? Looks like it's from testing

bingyanglin (changes requested):
The *rej files should be removed.

crates/iota/validator.info
nmrshll: yes, I'll remove that

crates/iota-core/src/authority.rs
```diff
@@ -4005,7 +4005,7 @@ impl AuthorityState {
                 };
                 return Ok(Some((
                     (*transaction).clone().into_message(),
-                    TransactionStatus::Executed(cert_sig, effects.into_inner(), events),
+                    TransactionStatus::Executed(cert_sig, Box::new(effects.into_inner()), events),
```
bingyanglin: Does this exist in the upstream change PR?

crates/iota-core/src/safe_client.rs
```diff
@@ -253,7 +253,7 @@ impl<C: Clone> SafeClient<C> {
                 ))
             }
             TransactionStatus::Executed(cert_opt, effects, events) => {
-                let signed_effects = self.check_signed_effects_plain(digest, effects, None)?;
+                let signed_effects = self.check_signed_effects_plain(digest, *effects, None)?;
```
bingyanglin: Does this exist in the upstream change PR?

crates/iota-core/src/unit_tests/authority_aggregator_tests.rs
```diff
@@ -964,7 +964,7 @@ async fn test_handle_transaction_response() {
     let resp = HandleTransactionResponse {
         status: TransactionStatus::Executed(
             Some(cert_epoch_0.auth_sig().clone()),
-            sign_tx_effects(effects, 0, *name_0, key_0),
+            Box::new(sign_tx_effects(effects, 0, *name_0, key_0)),
```
bingyanglin: Does this exist in the upstream change PR? There seems also other places that are not changed in the upstream PR?

crates/iota-core/src/authority.rs
nmrshll: No, but without it, it introduced a cargo check / CI warning on our side (with our CI checks settings)
Moving the enum variant into a Box removes that warning.

crates/iota-core/src/safe_client.rs
nmrshll: Same here:
No, but without it, it introduced a cargo check / CI warning on our side (with our CI checks settings)
Moving the enum variant into a Box removes that warning.

crates/iota-core/src/unit_tests/authority_aggregator_tests.rs
nmrshll: Same here:
No, but without it, it introduced a cargo check / CI warning on our side (with our CI checks settings)
Moving the enum variant into a Box removes that warning.

nmrshll: and same thing in a couple other places where that change was ported. 
Is this not the right solution ? and if no, why ?

crates/iota-core/src/authority.rs
nmrshll: The warning was about enum variants being very different in byte size

muXxer:
Please remove the boxing, because the signed variant will become the default, and the unsigned variant will be removed anyways.

piotrm50 (approved):
Apart from the additional `Box` wrapping that is not present in the upstream - lgtm.

crates/iota-core/src/safe_client.rs
nmrshll: -> removed the box in the end since the problem it solves is going away in an ucoming commit port

crates/iota-core/src/authority.rs
nmrshll: -> removed the box in the end since the problem it solves is going away in an ucoming commit port

crates/iota-core/src/unit_tests/authority_aggregator_tests.rs
nmrshll: -> removed the box in the end since the problem it solves is going away in an ucoming commit port

bingyanglin (approved):

Thoralf-M (approved):

muXxer (approved):

Issue #5181, opened 2025-02-04
Fixes for overload protection (#20248) · MystenLabs/sui@6ea6bc9


bingyanglin:
[Fixes for overload protection (#20248) · MystenLabs/sui@6ea6bc9](https://github.com/MystenLabs/sui/commit/6ea6bc9b112bf34a3dcfb136430726924b03e0d0)
- For us default_max_txn_age_in_queue =1sec, this PR changes it from 200 ms to 500 ms

PR #6934, opened 2025-05-21
upstream(node): Lower per object queue age and length thresholds

bingyanglin:
# Description of change

- Upstream range: [1.35.4-1.36.2)
- Port Sui's commit [Lower per object queue age and length thresholds (#19851)](https://github.com/MystenLabs/sui/commit/e920c3e0cfc8673e0858c69a94d8bbc261b0fa27)
- Description from Sui's commit: The original limits on the object queue are set with 1s checkpoint interval. Having the oldest transaction on an object queued for 1s is almost tolerable. In hindsight it should be lower than the checkpoint
interval to ensure healthy system. Now checkpoint interval is targeting 0.2s - 0.25s, so lowering the limits further to 0.2s, to ensure smooth checkpoint constructions.
- Notes
  - The `default_max_txn_age_in_queue` was already updated to be 500ms by a newer upstream change and merged in https://github.com/iotaledger/iota/issues/5181, where the `default_max_transaction_manager_per_object_queue_length` was 20 (see [here](https://github.com/MystenLabs/sui/blob/cedac9a012356f6fd9cd6b862521b3963f86bf3b/crates/sui-config/src/node.rs#L922)), hence in this porting we update `default_max_transaction_manager_per_object_queue_length` to be 20.
  - The snapshot file was updated to pass tests

Part of #3990 

## Type of change

Ran the local network with `RUST_LOG=info cargo run --release --bin iota start --force-regenesis --with-faucet`

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -941,7 +941,7 @@ fn default_max_transaction_manager_queue_length() -> usize {
 }
 
 fn default_max_transaction_manager_per_object_queue_length() -> usize {
-    100
+    20
 }
 
 impl Default for AuthorityOverloadConfig {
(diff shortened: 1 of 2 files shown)


gokhan-simsek-iota:
@bingyanglin 
We did not port the following on purpose, right? I see that we changed this value to `500` in our code 3 months ago.
```diff
fn default_max_txn_age_in_queue() -> Duration {
-     Duration::from_secs(1)
+    Duration::from_millis(200)
}
```

gokhan-simsek-iota (approved):

muXxer (approved):

PR #6936, opened 2025-05-21
refactor(core): Factor out parallel live object indexing

gokhan-simsek-iota:
## Description of change

Upstream range: [v1.35.4, v1.36.2)

Port [MystenLabs/sui@6159973](https://github.com/MystenLabs/sui/commit/6159973ae01f7b74505271d5ae76681f6594fe59)
Factor out the parallel live object indexing, used for initializing the rest indexes, into more general purpose and reusable logic.

Part of https://github.com/iotaledger/iota/issues/3990

## Type of change

Make sure Rest API is enabled in config:

```
...
// disabled by default, needs to be enabled in config:
// fullnode.yaml 
enable-rest-api: true
```

```
cargo r --bin iota-node -- --config-path /tmp/fullnode.yaml
```

Run any query to see Rest API works:
```
$ curl -s localhost:9000/api/v1/system | jq | grep -w safe_mode
```
`  "safe_mode": false,`

### DIFF
--- a/crates/iota-core/src/lib.rs
+++ b/crates/iota-core/src/lib.rs
@@ -26,6 +26,7 @@ pub mod mock_consensus;
 pub mod module_cache_metrics;
 pub mod mysticeti_adapter;
 pub mod overload_monitor;
+mod par_index_live_object_set;
 pub(crate) mod post_consensus_tx_reorder;
 pub mod quorum_driver;
 pub mod rest_index;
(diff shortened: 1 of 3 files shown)


muXxer (approved):

muXxer:
We should port all the changes of the upstream PR in one PR.
The introduced feature of the upstream PR should be added by the infrastructure team tho.

So this should be closed.
Sorry, my bad.

PR #6941, opened 2025-05-21
chore(apps-core)!: Remove `CLOCK_PACKAGE_ID`

marc2332:
Small typo fix


Thoralf-M:
That's a breaking change :trollface: Should be `chore(apps-core)!:`

marc2332:
> That's a breaking change :trollface: Should be `chore(apps-core)!:`

is it really a breaking change if nothing can break ? 💡  Because the core package is only used by the wallet/dashhboard/explorer and they all get the latest version of it when they build 😄 

So its more like an internal change

Thoralf-M:
> > That's a breaking change :trollface: Should be `chore(apps-core)!:`
> 
> is it really a breaking change if nothing can break ? 💡 Because the core package is only used by the wallet/dashhboard/explorer and they all get the latest version of it when they build 😄
> 
> So its more like an internal change

Ah sorry, you're right, I just knew I used some clock import already, but it was `import { IOTA_CLOCK_OBJECT_ID } from "@iota/iota-sdk/utils";` 
But this makes me wonder, why are there multiple consts for the same thing? Shouldn't there be only a single place which then gets used in the other places?

marc2332:
> > > That's a breaking change :trollface: Should be `chore(apps-core)!:`
> > 
> > 
> > is it really a breaking change if nothing can break ? 💡 Because the core package is only used by the wallet/dashhboard/explorer and they all get the latest version of it when they build 😄
> > So its more like an internal change
> 
> Ah sorry, you're right, I just knew I used some clock import already, but it was `import { IOTA_CLOCK_OBJECT_ID } from "@iota/iota-sdk/utils";` But this makes me wonder, why are there multiple consts for the same thing? Shouldn't there be only a single place which then gets used in the other places?

Ohh yes, indeed.. seems like it was duplicated at some point, lemme clean this up

Thoralf-M (approved):

msarcev (approved):

PR #6961, opened 2025-05-22
chore(dependencies): Unpin chrono

DaughterOfMars:
# Description of change

Unpins the `chrono` dependency now that arrow has made new releases.

## Type of change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -528,58 +528,23 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "54.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b5ec52ba94edeed950e4a41f75d35376df196e8cb04437f7280a5aa49f20f796"
-dependencies = [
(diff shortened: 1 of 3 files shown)


Thoralf-M (approved):

thibault-martinez (approved):

Alex6323 (approved):

thibault-martinez (changes requested):
Unused dependencies

piotrm50 (approved):

thibault-martinez (approved):

kodemartin (approved):

DaughterOfMars:
Converted to draft as there are still (a lot of) downstream dependencies which need to update before this is possible.

eike-hass:
Is there another issue or PR where this effort is tracked?

DaughterOfMars:
> Is there another issue or PR where this effort is tracked?

There's almost nothing we can about this, unfortunately. There are six (**at minimum**, but this is when I stopped looking) 3rd-party crates which would need to be updated. Some of them are either unmaintained or very infrequently maintained. Some require small changes to update and some would need almost complete rewrites and the requisite knowledge to do this properly, which is no small challenge given they are cryptography crates which are heavily utilized. I believe that there is no path forward where we are able to do this update "quickly".

Issue #6429, opened 2025-04-15
Implement Data Manager (Data Reconstructor)


piotrm50:
- [ ] Manage commit data availability
- [ ] Fetch missing data from the author or peers who acknowledged the data
- [ ] Store data to DagState once available

PR #6978, opened 2025-05-23
feat(starfish): Update `Commit` creation to support transaction acknowledments

### LINKED ISSUE #6437: Update Commit Struct with Data Acks
- [ ] Track data acks across commits
- [ ] Include only data with quorum acks in new commit
- [ ] Add block refs for headers traversed in the commit (where we collected votes)
- [ ] Add block refs for certified data for which this commit reached quorum
- [ ] Optionally include bitmap of acknowledgers so that we know from which nodes the data can be requested.

piotrm50:
# Description of change

This PR implements modifying Commit struct and commit creation flow to support separate block header and transactions acknowledgments. It also extends the `Linearizer` component to support keeping track of transaction acknowledments and including them in newly created commits when they reach quorum of acks. It also extends test DAG builder to support transaction acks in a basic form. 

The next step is implementation of `DataManager` (#6429) and data synchronizer (#6881) and covering the new logic with tests.

Resolves #6437

## Type of change

Fixed tests of the components. The tests need to be expanded in a future PR. 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -636,7 +636,6 @@ pub struct VerifiedTransactions {
     transactions: Vec<Transaction>,
 
     /// The block reference of the block that contains the transactions.
-    #[expect(dead_code)]
     block_ref: BlockRef,
 
     /// Commitment of transactions in the block
@@ -670,6 +669,10 @@ impl VerifiedTransactions {
 
     pub fn transactions_commitment(&self) -> TransactionsCommitment {
         self.transactions_commitment
+    }
+
+    pub fn block_ref(&self) -> BlockRef {
+        self.block_ref
     }
 }
 
@@ -736,6 +739,11 @@ impl TestBlockHeader {
 
     pub fn set_commit_votes(mut self, commit_votes: Vec<CommitVote>) -> Self {
         self.block_header.commit_votes = commit_votes;
+        self
+    }
+
+    pub fn set_acknowledgments(mut self, acknowledgments: Vec<BlockRef>) -> Self {
+        self.block_header.acknowledgments = acknowledgments;
         self
     }
 
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -630,6 +630,7 @@ impl fmt::Debug for VerifiedBlockHeader {
 }
 
 /// VerifiedTransactions are transactions that correspond to an existing block
+#[derive(Clone)]
 pub struct VerifiedTransactions {
     #[expect(dead_code)]
     transactions: Vec<Transaction>,
@@ -638,22 +639,37 @@ pub struct VerifiedTransactions {
     #[expect(dead_code)]
     block_ref: BlockRef,
 
+    /// Commitment of transactions in the block
+    transactions_commitment: TransactionsCommitment,
+
     /// The serialized bytes of the transactions.
     #[expect(dead_code)]
     serialized: Bytes,
+}
+
+impl PartialEq for VerifiedTransactions {
+    fn eq(&self, other: &Self) -> bool {
+        self.transactions_commitment() == other.transactions_commitment()
+    }
 }
 
 impl VerifiedTransactions {
     pub(crate) fn new(
         transactions: Vec<Transaction>,
         block_ref: BlockRef,
+        transactions_commitment: TransactionsCommitment,
         serialized: Bytes,
     ) -> Self {
         Self {
             transactions,
             block_ref,
+            transactions_commitment,
             serialized,
         }
+    }
+
+    pub fn transactions_commitment(&self) -> TransactionsCommitment {
+        self.transactions_commitment
     }
 }
 
--- a/crates/starfish/core/src/commit_observer.rs
+++ b/crates/starfish/core/src/commit_observer.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{ops::Deref, sync::Arc, time::Duration};
+use std::{sync::Arc, time::Duration};
 
 use iota_metrics::monitored_mpsc::UnboundedSender;
 use parking_lot::RwLock;
@@ -10,9 +10,9 @@ use tokio::time::Instant;
 use tracing::{debug, info};
 
 use crate::{
-    CommitConsumer, CommittedSubDag,
+    BlockRef, CommitConsumer, CommittedSubDag,
     block_header::{BlockHeaderAPI, VerifiedBlockHeader},
-    commit::{CommitAPI, CommitIndex, load_committed_subdag_from_store},
+    commit::{CommitAPI, CommitIndex, PendingSubDag, load_pending_subdag_from_store},
     context::Context,
     dag_state::DagState,
     data_manager::DataManager,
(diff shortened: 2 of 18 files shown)


crates/starfish/core/src/block_header.rs
```diff
@@ -630,31 +630,50 @@ impl fmt::Debug for VerifiedBlockHeader {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.transactions_commitment() == other.transactions_commitment()
+    }
+}
+
 impl VerifiedTransactions {
     pub(crate) fn new(
         transactions: Vec<Transaction>,
         block_ref: BlockRef,
+        transactions_commitment: TransactionsCommitment,
```
VorobyevIlya, line 659: Do we need it here? We have transactions_commitment in the header. When we want to check if transactions correspond to the header, for example, when we receive header and data, we recalculate the commitment from serialized transactions.

crates/starfish/core/src/commit_observer.rs
```diff
@@ -75,7 +84,7 @@ impl CommitObserver {
     pub(crate) fn handle_commit(
         &mut self,
         committed_leaders: Vec<VerifiedBlockHeader>,
-    ) -> ConsensusResult<Vec<CommittedSubDag>> {
+    ) -> ConsensusResult<(Vec<PendingSubDag>, Vec<BlockRef>)> {
```
VorobyevIlya, line 96: Describe the meaning of the return values, especially `Vec<BlockRef>`

crates/starfish/core/src/core_thread.rs
```diff
@@ -67,6 +67,18 @@ pub trait CoreThreadDispatcher: Sync + Send + 'static {
...
 
+    // The following method will be called from data synchronizer to trigger output
+    // of commits that became solid in the DataManager
+    #[expect(unused)]
+    async fn add_data(
+        &self,
+        data: Vec<VerifiedTransactions>,
+    ) -> Result<BTreeSet<BlockRef>, CoreError>;
+
+    // The following method will be used by the data synchronizer to get
```
VorobyevIlya: unfinished sentence

crates/starfish/core/src/block_header.rs
piotrm50: This is used in `PartialEq` trait to avoid calculating the commitment multiple times. This should not be sent over the wire. Once you receive and Verify that the transactions are correct, then we should create `VerifiedTransactions` - just like with BlockHeader and VerifiedBlockHeader.

VorobyevIlya (approved):

crates/starfish/core/src/core_thread.rs
piotrm50: fixed

crates/starfish/core/src/commit_observer.rs
piotrm50: added a docstring

crates/starfish/core/src/data_manager.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use crate::{
+    BlockRef, CommittedSubDag, commit::PendingSubDag, context::Context, dag_state::DagState,
+};
+
+/// Block manager suspends incoming blocks until they are connected to the
+/// existing graph, returning newly connected blocks.
+/// TODO: As it is possible to have Byzantine validators who produce Blocks
+/// without valid causal history we need to make sure that BlockManager takes
+/// care of that and avoid OOM (Out Of Memory) situations.
```
polinikita: Rewrite this part

crates/starfish/core/src/linearizer.rs
```diff
@@ -91,16 +102,27 @@ impl Linearizer {
...
+            to_commit
+                .iter()
+                .flat_map(|block| {
+                    self.add_committed_transaction_acks(
+                        block.author(),
+                        block.acknowledgments().to_vec(),
+                    )
+                })
+                .unique()
+                .collect::<Vec<_>>(),
```
polinikita: Add a comment above about what is going on here. This is a part of the essential logic in Starfish, but it is easy to overlook it here. Also, I would compute the part with flat_map separetely and give a description to that; Otherwise, it is easy to overlook

crates/starfish/core/src/linearizer.rs
```diff
@@ -91,16 +102,27 @@ impl Linearizer {
                 .iter()
                 .map(|block| block.reference())
                 .collect::<Vec<_>>(),
+            to_commit
+                .iter()
+                .flat_map(|block| {
```
polinikita: Why flat_map here?

crates/starfish/core/src/linearizer.rs
```diff
@@ -205,17 +225,47 @@ impl Linearizer {
         // storage.
         self.dag_state.write().flush();
 
-        committed_sub_dags
+        pending_sub_dags
+    }
+
+    pub(crate) fn add_committed_transaction_acks(
```
polinikita, line 242: add description here

crates/starfish/core/src/linearizer.rs
```diff
@@ -205,17 +225,47 @@ impl Linearizer {
...
+            if !votes_collector.reached_threshold(&self.context.committee)
+                && votes_collector.add(authority, &self.context.committee)
+            {
+                acknowledged_data.push(block_ref);
+            }
+        }
+        acknowledged_data
+    }
+
+    pub(crate) fn recover_transaction_ack_tracker(
```
polinikita: Add a description. Maybe explain why no need to use Vec<BlockRef> while recovering.

crates/starfish/core/src/test_dag_parser.rs
```diff
@@ -88,7 +98,25 @@ fn parse_round<'a>(
         |input| parse_specified_connections(input, dag_builder),
     ))(input)?;
 
-    Ok((input, (round.parse().unwrap(), connections)))
+    // TODO: extend DAG parser with transaction acknowledgments. For now it's
+    //  assumed that transactions are available together with the block headers and
+    //  we acknowledge transactions of all ancestors.
```
polinikita, line 103: just a note that all ancestors method for acknowledging usually does not allow to commit all transaction data.

polinikita (approved):
Overall, it looks good to me. Added some comments to elaborate descriptions of functions.

crates/starfish/core/src/linearizer.rs
piotrm50: Removed this whole function and inlined the logic in commit_observer.rs. Also added some more comments explaining the logic there.

crates/starfish/core/src/linearizer.rs
piotrm50: added

crates/starfish/core/src/linearizer.rs
piotrm50: `add_committed_transaction_acks` returns a vector, so if I used a regular map I'd get a vector of vectors. I need a vector of unique block refs to transactions that reached quorum with acknowledmgents from this commit, so I use flat_map to flatten this vector of vectors into a vector.

crates/starfish/core/src/linearizer.rs
piotrm50: done

crates/starfish/core/src/data_manager.rs
piotrm50: Replaced with a TODO that will be resolved in the subsequent PR introducing the logic of data manager.

PR #7011, opened 2025-05-26
fix(wallet): keep updating countdown message and disable 'unlock' button

### LINKED ISSUE #6684: Improve locked account UX
if the account is temp locked, we could just disable the input and render the error always

Image
Image
Image

KeitoTadashi:

fixes #6684


apps/wallet/src/shared/accounts.ts
```diff
@@ -32,12 +32,15 @@ export interface AccountFromFinder {
 }
 
 export class AccountTooManyAttemptsError extends Error {
-    constructor(public remainingTime: number) {
+    public remainingTime: number;
+
+    constructor(remainingTime: number) {
```
marc2332: The previous code was also correct btw

marc2332: I see now it was eslint that complained about it :(

KeitoTadashi: :laughing:

evavirseda (dismissed):

panteleymonchuk (dismissed):

apps/wallet/src/ui/app/components/accounts/PasswordInputDialog.tsx
```diff
@@ -56,6 +57,28 @@ export function PasswordModalDialog({
...
+    // Clear the interval and error if the dialog closed but not unmounted
+    useEffect(() => {
+        if (!open && countdownError?.interval) {
+            clearInterval(countdownError.interval);
+            setCountdownError(null);
+        }
+    }, [open]);
+
+    // Clear the interval if the dialog unmounted
+    useEffect(() => {
```
msarcev, line 70: These two can't be merge into one `useEffect` ?
Like if you add the return to the first `useEffect` won't it just run when the component unmounts ?

apps/wallet/src/background/accounts/index.ts
```diff
@@ -300,14 +300,21 @@ export async function accountsHandleUIMessage(msg: Message, uiConnection: UiConn
         return true;
     }
     if (isMethodPayload(payload, 'getLockedState')) {
-        const { failedAttempts, lastFailedAttemptTime, isLockedOut, lockTimeMs } =
-            await getLockedState();
+        let remainingTime = 0;
+        const WALLET_LOCK_DURATION_IN_MS = 60 * MILLISECONDS_PER_SECOND;
```
marc2332: we should move this constant (is also in other functions) to another place so we can reuse it instead of duplicating it :D

marc2332 (changes requested):

msarcev (approved):

marc2332 (approved):

PR #7039, opened 2025-05-26
fix(iota): use gas coin as payment, if only a single gas coin is available

### LINKED ISSUE #6951: CLI: Make sure that all name commands work with a single gas coin
Some commands didn't work in the past, if one had only a single gas coin, as coin was used to split the payment and then there was no coin available for the gas payment anymore. In such a case, the amount should be split from the gas coin instead, so the transaction still works.
Go though each command that does a transaction and try it while having only a single gas coin.

Thoralf-M:
# Description of change

Use gas coin as payment, if only a single gas coin is available.
Also improved the error message when fetching the price config fails (usually happens if the package id was not set correct)

Fixes https://github.com/iotaledger/iota/issues/6951

## Type of change

Running `cargo run --features=iota-names name register test.iota` with a single gas coin and also with multiple gas coins (another name in this case), both work now.

If the package id was not set correct for the network, previously the error was
`Some(NotExists { object_id: 0xc0671ed8571ed80dae4ca18124c13d5cec36463837e5fdcc11591e2dbb6ba1ff })`, now it's `couldn't fetch pricing config: Some(NotExists { object_id: 0xc0671ed8571ed80dae4ca18124c13d5cec36463837e5fdcc11591e2dbb6ba1ff })`
We could even add something like `make sure to have the env variables set for IOTA-Names` or similar in this case, but not sure about this here

### DIFF
--- a/crates/iota/src/name_commands.rs
+++ b/crates/iota/src/name_commands.rs
@@ -335,11 +335,11 @@ impl NameCommand {
                 let price = fetch_pricing_config(&iota_client).await?.get_price(label)?;
                 let domain_name = domain.to_string();
                 let coin =
-                    select_coin_for_payment(domain_name.as_str(), coin, price, context).await?;
+                    select_coin_arg_for_payment(domain_name.as_str(), coin, price, context).await?;
                 let mut args = vec![
                     "--move-call iota::tx_context::sender".to_string(),
                     "--assign sender".to_string(),
-                    format!("--split-coins @{coin} [{price}]"),
+                    format!("--split-coins {coin} [{price}]"),
                     "--assign coins".to_string(),
                     format!(
                         "--move-call {}::payment::init_registration @{} '{domain_name}'",
@@ -413,14 +413,14 @@ impl NameCommand {
                     * years as u64;
                 let domain_name = domain.to_string();
                 let coin =
-                    select_coin_for_payment(domain_name.as_str(), coin, price, context).await?;
+                    select_coin_arg_for_payment(domain_name.as_str(), coin, price, context).await?;
                 let nft_id = get_owned_nft_by_name::<IotaNamesRegistration>(&domain, context)
                     .await?
                     .id();
                 let mut args = vec![
                     "--move-call iota::tx_context::sender".to_string(),
                     "--assign sender".to_string(),
-                    format!("--split-coins @{coin} [{price}]"),
+                    format!("--split-coins {coin} [{price}]"),
                     "--assign coins".to_string(),
                     format!(
                         "--move-call {}::payment::init_renewal @{} @{nft_id} {years}",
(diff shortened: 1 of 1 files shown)


crates/iota/src/name_commands.rs
```diff
@@ -1992,29 +1994,30 @@ impl PricingConfig {
     }
 }
 
-async fn select_coin_for_payment(
+// Returns "@<coin id>" or "gas" to be used as coin payment argument in a PTB
```
thibault-martinez, line 1997: Does "gas" assume the variable has been set already?

Thoralf-M: Not sure what you mean, which variable?

thibault-martinez: Why are we returning "gas", what is this?

Thoralf-M: That's how one references the gas coin in a PTB, providing the actual gas coin would result in an error, as coins in arguments are separated from the actual gas coins
```
iota client ptb \
--split-coins gas "[1000000000]" \ # the gas here
--assign coin
```

thibault-martinez: Ah yeah I see, got confused, thanks

thibault-martinez (approved):

crates/iota/src/name_commands.rs
```diff
@@ -1992,29 +1994,30 @@ impl PricingConfig {
...
+async fn select_coin_arg_for_payment(
     domain_name: &str,
     coin: Option<ObjectID>,
     price: u64,
     context: &mut WalletContext,
-) -> anyhow::Result<ObjectID> {
+) -> anyhow::Result<String> {
     Ok(match coin {
-        Some(coin) => coin,
+        Some(coin) => format!("@{coin}"),
```
DaughterOfMars, line 2005: Not sure I'm a fan of the double format :c

Thoralf-M: You mean here and then for the PTB string? If we don't do the `@` addition here, we have to match on the result of this function and add it or not depending on the gas coin or an object id

DaughterOfMars: Oh, I see. It's using the `gas` var. Is that always available?

thibault-martinez: I had the exact same brainfart, yes is is

DaughterOfMars: okie dokie then

DaughterOfMars (approved):

thibault-martinez:
@Thoralf-M Maybe this deserved a release note btw?

Thoralf-M:
@thibault-martinez I didn't add it, because it's not actually in the released binary atm

thibault-martinez:
> @thibault-martinez I didn't add it, because it's not actually in the released binary atm

Fair enough but we've been adding IOTA-Names CLI notes anyway

PR #7054, opened 2025-05-27
chore(iota-sdk): use re-exports in examples and doc comments

### LINKED ISSUE #7049: Crate imports in the Rust SDK examples and doc snippets depend on other crates
## Bug description

When using the example code as a basis for a new project I came across wrong crate imports.
Some examples are importing other crates directly like
 `use iota_json_rpc_types::IotaObjectDataOptions`
but this means my project would now depend on more crates that I need to import.

These symbols are already re-exported inside the sdk, so the correct import would be
 `use iota_sdk::rpc_types::{IotaObjectDataOptions};`

This occurs not only in the `examples/read_api` file but also in the in-line code doc snippets for the SDK.
Example:
```
    /// Get a parsed past object and version for the provided object ID.
    ///
    /// An object's version increases when the object is mutated, though it is
    /// not guaranteed that it increases always by 1. A past object can be used
    /// to understand how the object changed over time, i.e. what was the total
    /// balance at a specific version.
    ///
    /// # Examples
    ///
    /// ```rust,no_run
    /// use std::str::FromStr;
    ///
    /// use iota_json_rpc_types::IotaObjectDataOptions;
    /// use iota_sdk::IotaClientBuilder;
    /// use iota_types::base_types::{IotaAddress, ObjectID};
    ///
    /// #[tokio::main]
    /// async fn main() -> Result<(), anyhow::Error> {
    ///     let iota = IotaClientBuilder::default().build_localnet().await?;
    ///     let address = IotaAddress::from_str("0x0000....0000")?;
    ///     let owned_objects = iota
    ///         .read_api()
    ///         .get_owned_objects(address, None, None, None)
    ///         .await?;
    ///     // this code example assumes that there are previous owned objects
    ///     let object = owned_objects
    ///         .data
    ///         .get(0)
    ///         .expect(&format!("No owned objects for this address {}", address));
    ///     let object_data = object.data.as_ref().expect(&format!(
    ///         "No object data for this IotaObjectResponse {:?}",
    ///         object
    ///     ));
    ///     let object_id = object_data.object_id;
    ///     let version = object_data.version;
    ///     let past_object = iota
    ///         .read_api()
    ///         .try_get_parsed_past_object(
    ///             object_id,
    ///             version,
    ///             IotaObjectDataOptions {
    ///                 show_type: true,
    ///                 show_owner: true,
    ///                 show_previous_transaction: true,
    ///                 show_display: true,
    ///                 show_content: true,
    ///                 show_bcs: true,
    ///                 show_storage_rebate: true,
    ///             },
    ///         )
    ///         .await?;
    ///     Ok(())
    /// }
    /// ```
    pub async fn try_get_parsed_past_object(
        &self,
        object_id: ObjectID,
        version: SequenceNumber,
        options: IotaObjectDataOptions,
    ) -> IotaRpcResult<IotaPastObjectResponse> {
        Ok(self
            .api
            .http
            .try_get_past_object(object_id, version, Some(options))
            .await?)
    }
```

Here 2 other crates are imported, although all required symbols are re-exported in the `iota_sdk` crate

thibault-martinez:
Fixes https://github.com/iotaledger/iota/issues/7049
So that it doesn't force people to import more crates than necessary.

`iota_config` and `iota_keys` could maybe be made re-exports as well?

### DIFF
--- a/crates/iota-sdk/src/apis/read.rs
+++ b/crates/iota-sdk/src/apis/read.rs
@@ -58,8 +58,7 @@ impl ReadApi {
     /// ```rust,no_run
     /// use std::str::FromStr;
     ///
-    /// use iota_sdk::IotaClientBuilder;
-    /// use iota_types::base_types::IotaAddress;
+    /// use iota_sdk::{IotaClientBuilder, types::base_types::IotaAddress};
     ///
     /// #[tokio::main]
     /// async fn main() -> Result<(), anyhow::Error> {
@@ -100,8 +99,10 @@ impl ReadApi {
     /// ```rust,no_run
     /// use std::str::FromStr;
     ///
-    /// use iota_sdk::IotaClientBuilder;
-    /// use iota_types::base_types::{IotaAddress, ObjectID};
+    /// use iota_sdk::{
+    ///     IotaClientBuilder,
+    ///     types::base_types::{IotaAddress, ObjectID},
+    /// };
     ///
     /// #[tokio::main]
     /// async fn main() -> Result<(), anyhow::Error> {
@@ -182,9 +183,11 @@ impl ReadApi {
     /// ```rust,no_run
     /// use std::str::FromStr;
     ///
-    /// use iota_json_rpc_types::IotaObjectDataOptions;
-    /// use iota_sdk::IotaClientBuilder;
-    /// use iota_types::base_types::{IotaAddress, ObjectID};
+    /// use iota_sdk::{
+    ///     IotaClientBuilder,
+    ///     rpc_types::IotaObjectDataOptions,
+    ///     types::base_types::{IotaAddress, ObjectID},
+    /// };
     ///
     /// #[tokio::main]
     /// async fn main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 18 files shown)


crates/iota-sdk/src/apis/read.rs
```diff
@@ -247,9 +250,7 @@ impl ReadApi {
     /// ```rust,no_run
     /// use std::str::FromStr;
     ///
-    /// use iota_json_rpc_types::{IotaGetPastObjectRequest, IotaObjectDataOptions};
-    /// use iota_sdk::IotaClientBuilder;
-    /// use iota_types::base_types::{IotaAddress, ObjectID};
+    /// use iota_sdk::{types::base_types::{IotaAddress, ObjectID}IotaClientBuilder, rpc_types::{IotaGetPastObjectRequest, IotaObjectDataOptions}};
```
thibault-martinez: I know there is an issue here, just checking if the CI picks it up

thibault-martinez: It does

Thoralf-M (approved):

alexsporn (approved):
lgtm!
Regarding `iota_keys` and `iota_config` I am not even sure this should be part of the SDK tbh. This is functionality from the CLI. I don't think every project using the SDK should be touching the CLI config and keystore.

PR #7086, opened 2025-05-27
feat(iota-names-indexer): Add CLI

### LINKED ISSUE #7018: CLI configuration of iota-names-indexer
This should be provided via some config

_Originally posted by @Thoralf-M in https://github.com/iotaledger/iota/pull/6975#discussion_r2107264511_

DaughterOfMars:
# Description of change

Adds a simple CLI which can be added to in the future, and reorganizes the startup logic.

Closes #7018

### DIFF
--- a/crates/iota-names-indexer/src/main.rs
+++ b/crates/iota-names-indexer/src/main.rs
@@ -4,57 +4,162 @@
 mod metrics;
 mod worker;
 
-use std::{path::PathBuf, sync::Arc};
+use std::sync::Arc;
 
 use anyhow::Result;
-use iota_data_ingestion_core::{
-    DataIngestionMetrics, FileProgressStore, IndexerExecutor, ReaderOptions, WorkerPool,
-};
+use clap::Parser;
 use iota_names::config::IotaNamesConfig;
+use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
-use tracing::info;
-
-use self::{
-    metrics::{IotaNamesMetrics, METRICS, start_prometheus_server},
-    worker::IotaNamesWorker,
+use tracing::{error, info};
+use tracing_subscriber::{
+    EnvFilter, fmt::format::FmtSpan, layer::SubscriberExt, util::SubscriberInitExt,
 };
+
+use self::{metrics::IotaNamesMetrics, worker::IotaNamesWorker};
+use crate::metrics::PrometheusServer;
+
+// Define the `GIT_REVISION` and `VERSION` consts
+bin_version::bin_version!();
+
+#[derive(Parser)]
+#[command(
+    name = env!("CARGO_BIN_NAME"),
+    about = env!("CARGO_PKG_DESCRIPTION"),
+    author,
+    version = VERSION,
+    propagate_version = true,
+)]
+enum CLI {
+    Start {
+        /// The URL of an IOTA node to get data from.
+        #[arg(long, default_value = "http://localhost:9000")]
+        node_url: String,
+    },
+}
+
+impl CLI {
+    async fn execute(self) -> Result<()> {
+        match self {
+            CLI::Start { node_url } => {
+                info!("Starting IOTA Names Indexer");
+
+                let prometheus = PrometheusServer::new();
+                let registry = prometheus.registry();
+
+                let cancel_token = CancellationToken::new();
+
+                let mut tasks: JoinSet<Result<()>> = JoinSet::new();
+
+                // Spawn the prometheus API
+                let handle = cancel_token.clone();
+                tasks.spawn(async move {
+                    prometheus.start(handle).await?;
+                    Ok(())
+                });
+
+                // Spawn the metrics worker
+                let handle = cancel_token.clone();
+                tasks.spawn(async move {
+                    let worker = IotaNamesWorker::new(
+                        IotaNamesConfig::from_env().unwrap_or_default(),
+                        Arc::new(IotaNamesMetrics::new(&registry)),
+                    );
+
+                    tokio::select! {
+                        res = worker.run(&node_url, &registry, handle.clone()) => {res?;},
+                        _ = handle.cancelled() => {},
+                    }
+                    Ok(())
+                });
+
+                let mut exit_code = Ok(());
+
+                tokio::select! {
+                    res = interrupt_or_terminate() => {
+                        if let Err(err) = res {
+                            tracing::error!("subscribing to OS interrupt signals failed with error: {err}; shutting down");
+                            exit_code = Err(err);
+                        } else {
+                            tracing::info!("received ctrl-c or terminate; shutting down");
+                        }
+                    },
+                    res = tasks.join_next() => {
+                        if let Some(Ok(Err(err))) = res {
+                            tracing::error!("a worker failed with error: {err}");
+                            exit_code = Err(err);
+                        }
+                    },
+                }
+
+                cancel_token.cancel();
+
+                // Allow the user to abort if the tasks aren't shutting down quickly.
+                tokio::select! {
+                    res = interrupt_or_terminate() => {
+                        if let Err(err) = res {
+                            tracing::error!("subscribing to OS interrupt signals failed with error: {err}; aborting");
+                            exit_code = Err(err);
+                        } else {
+                            tracing::info!("received second ctrl-c or terminate; aborting");
+                        }
+                        tasks.shutdown().await;
+                        tracing::info!("runtime aborted");
+                    },
+                    _ = async { while tasks.join_next().await.is_some() {} } => {
+                        tracing::info!("runtime stopped");
+                    },
+                }
+
+                exit_code
+            }
+        }
+    }
+}
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    let _guard = telemetry_subscribers::TelemetryConfig::new()
-        .with_env()
-        .init();
-    info!("Starting IOTA Names Indexer...");
+    dotenvy::dotenv().ok();
 
-    let registry = start_prometheus_server();
+    set_up_logging()?;
 
-    METRICS.get_or_init(|| Arc::new(IotaNamesMetrics::new(&registry)));
-    let metrics = DataIngestionMetrics::new(&registry);
+    CLI::parse().execute().await?;
+    Ok(())
+}
 
-    let progress_store = FileProgressStore::new("./progress_store").await?;
+fn set_up_logging() -> Result<()> {
+    std::panic::set_hook(Box::new(|p| {
+        error!("{}", p);
+    }));
 
-    let cancel_token = CancellationToken::new();
-    let mut executor = IndexerExecutor::new(progress_store, 1, metrics, cancel_token);
+    let registry = tracing_subscriber::registry();
 
-    let worker = IotaNamesWorker::new(IotaNamesConfig::from_env().unwrap_or_default());
-    let worker_pool = WorkerPool::new(
-        worker,
-        "iota_names_reader".to_string(),
-        1,
-        Default::default(),
-    );
+    let registry = {
+        registry
+            .with(EnvFilter::from_default_env())
+            .with(tracing_subscriber::fmt::layer().with_span_events(FmtSpan::CLOSE))
+    };
 
-    executor.register(worker_pool).await.unwrap();
-    executor
-        .run(
-            PathBuf::from("./chk".to_string()), /* path to a local directory where checkpoints
-                                                 * are stored. */
-            Some("http://host.docker.internal:9000/api/v1".to_string()),
-            vec![],                   // optional remote store access options.
-            ReaderOptions::default(), // remote_read_batch_size.
-        )
-        .await
-        .unwrap();
+    registry.init();
+    Ok(())
+}
+
+pub async fn interrupt_or_terminate() -> Result<()> {
+    #[cfg(unix)]
+    {
+        use anyhow::anyhow;
+        use tokio::signal::unix::{SignalKind, signal};
+        let mut terminate = signal(SignalKind::terminate())
+            .map_err(|e| anyhow!("cannot listen to `SIGTERM`: {e}"))?;
+        let mut interrupt = signal(SignalKind::interrupt())
+            .map_err(|e| anyhow!("cannot listen to `SIGINT`: {e}"))?;
+        tokio::select! {
+            _ = terminate.recv() => {}
+            _ = interrupt.recv() => {}
+        }
+    }
+    #[cfg(not(unix))]
+    tokio::signal::ctrl_c().await?;
 
     Ok(())
 }
--- a/crates/iota-names-indexer/src/main.rs
+++ b/crates/iota-names-indexer/src/main.rs
@@ -30,7 +30,7 @@ bin_version::bin_version!();
     version = VERSION,
     propagate_version = true,
 )]
-enum CLI {
+enum Command {
     Start {
         /// The URL of an IOTA node to get data from.
         #[arg(long, default_value = "http://localhost:9000")]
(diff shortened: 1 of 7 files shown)


crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,62 @@ use iota_types::{
...
+            DataIngestionMetrics::new(registry),
+            token,
+        );
+        let worker_pool =
+            WorkerPool::new(self, "iota_names_reader".to_string(), 1, Default::default());
+        executor.register(worker_pool).await.unwrap();
+
+        info!("Connecting to node at {}", node_url);
+        executor
+            .run(
```
Thoralf-M: Why is the WorkerPool inside of the IotaNamesWorker::run?

Thoralf-M: I know you provide self, but this looks messed up

crates/iota-names-indexer/src/metrics.rs
```diff
@@ -26,11 +26,37 @@ impl IotaNamesMetrics {
...
+    pub(crate) async fn start(&self, token: CancellationToken) -> anyhow::Result<()> {
+        let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 9184);
+        info!("Starting prometheus metrics at {}", addr);
 
-pub(crate) fn start_prometheus_server() -> Registry {
-    let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 9184);
-    info!("Starting prometheus metrics at http://0.0.0.0:9184/metrics");
+        let app = Router::new()
+            .route(METRICS_ROUTE, get(iota_metrics::metrics))
+            .layer(Extension(self.registry_service.clone()));
```
thibault-martinez, line 46: Why do we need to go through axum now?

crates/iota-names-indexer/src/worker.rs
thibault-martinez: Kind of agree that conceptually it's a bit weird

crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,62 @@ use iota_types::{
...
+use prometheus::Registry;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, info};
 
-use crate::metrics::METRICS;
+use crate::IotaNamesMetrics;
 
 pub(crate) struct IotaNamesWorker {
     config: IotaNamesConfig,
+    metrics: Arc<IotaNamesMetrics>,
```
thibault-martinez, line 65: Nice, wanted to do that as well

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: Think I should pass it in instead? Are we actually going to support multiple workers?

crates/iota-names-indexer/src/metrics.rs
DaughterOfMars: It already was but this way we can control the graceful shutdown of the server

crates/iota-names-indexer/src/worker.rs
Thoralf-M: I think we need that for performance reason, otherwise syncing will be too slow. In the indexer they use 200 currently

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,162 @@
...
+
+                let mut exit_code = Ok(());
+
+                tokio::select! {
+                    res = interrupt_or_terminate() => {
+                        if let Err(err) = res {
+                            tracing::error!("subscribing to OS interrupt signals failed with error: {err}; shutting down");
+                            exit_code = Err(err);
+                        } else {
+                            tracing::info!("received ctrl-c or terminate; shutting down");
```
muXxer: Why not SIGINT instead of ctrl-c.
Maybe that signal wasn't send via "ctrl-c" in the first place :)

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,162 @@
...
+                cancel_token.cancel();
+
+                // Allow the user to abort if the tasks aren't shutting down quickly.
+                tokio::select! {
+                    res = interrupt_or_terminate() => {
+                        if let Err(err) = res {
+                            tracing::error!("subscribing to OS interrupt signals failed with error: {err}; aborting");
+                            exit_code = Err(err);
+                        } else {
+                            tracing::info!("received second ctrl-c or terminate; aborting");
```
muXxer: Why not SIGINT instead of ctrl-c.
Maybe that signal wasn't send via "ctrl-c" in the first place :)

muXxer (approved):

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: Okay, I'll refactor a bit

crates/iota-names-indexer/src/main.rs
DaughterOfMars: Yes but in docker you may want to trigger this other ways

crates/iota-names-indexer/src/main.rs
DaughterOfMars: I'm seeing double

crates/iota-names-indexer/src/main.rs
DaughterOfMars: Just realized I misread your comment

crates/iota-names-indexer/src/metrics.rs
```diff
@@ -26,11 +26,37 @@ impl IotaNamesMetrics {
...
+impl PrometheusServer {
+    pub(crate) fn new() -> Self {
+        Self {
+            registry_service: RegistryService::new(Registry::new()),
+        }
+    }
+
+    pub(crate) async fn start(&self, token: CancellationToken) -> anyhow::Result<()> {
+        let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 9184);
+        info!("Starting prometheus metrics at {}", addr);
```
thibault-martinez:
```suggestion
        info!("Starting prometheus metrics at {addr}");
```

crates/iota-names-indexer/src/metrics.rs
```diff
@@ -26,11 +26,37 @@ impl IotaNamesMetrics {
...
+            token.cancelled().await;
+        }
+        axum::serve(listener, app.into_make_service())
+            .with_graceful_shutdown(shutdown_signal(token))
+            .await?;
+
+        Ok(())
+    }
+
+    pub(crate) fn registry(&self) -> Registry {
```
thibault-martinez, line 59: I find the name a tiny bit misleading as if it was a getter but it returns a new one

DaughterOfMars: Well technically it is a reference, as the Registry wraps an Arc :)

crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,67 @@ use iota_types::{
...
-use tracing::debug;
-
-use crate::metrics::METRICS;
+use prometheus::Registry;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, info};
+
+use crate::IotaNamesMetrics;
+
+pub(crate) async fn run_iota_names_reader(
```
thibault-martinez, line 26: I would personally still have that in the start command, or its own executor module, but not too bothered about it

crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,67 @@ use iota_types::{
...
+    );
+    let worker_pool = WorkerPool::new(
+        worker,
+        "iota_names_reader".to_string(),
+        concurrency,
+        Default::default(),
+    );
+    executor.register(worker_pool).await?;
+
+    info!("Connecting to node at {}", node_url);
```
thibault-martinez:
```suggestion
    info!("Connecting to node at {node_url}");
```

crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,67 @@ use iota_types::{
...
+        progress_store,
+        1,
+        DataIngestionMetrics::new(registry),
+        token,
+    );
+    let worker_pool = WorkerPool::new(
+        worker,
+        "iota_names_reader".to_string(),
+        concurrency,
+        Default::default(),
```
thibault-martinez, line 45: I know it was already like that before, but I don't like `Default::default()` in this context, makes it hard to review/understand without an IDE

DaughterOfMars: If we want to import the type I need to add `backoff` to the manifest just for that

crates/iota-names-indexer/src/metrics.rs
thibault-martinez: So this function always return the same Registry?

crates/iota-names-indexer/src/worker.rs
thibault-martinez: And it's not re-exported by something we already import?

crates/iota-names-indexer/src/metrics.rs
DaughterOfMars: Yes

thibault-martinez: Fair enough then, thanks

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,167 @@
...
-use self::{
-    metrics::{IotaNamesMetrics, METRICS, start_prometheus_server},
-    worker::IotaNamesWorker,
+use tracing::{error, info};
+use tracing_subscriber::{
+    EnvFilter, fmt::format::FmtSpan, layer::SubscriberExt, util::SubscriberInitExt,
 };
 
+use self::{metrics::IotaNamesMetrics, worker::IotaNamesWorker};
+use crate::{metrics::PrometheusServer, worker::run_iota_names_reader};
```
thibault-martinez: merge them

thibault-martinez: (prefer self)

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,167 @@
...
+    version = VERSION,
+    propagate_version = true,
+)]
+enum Command {
+    Start {
+        /// The URL of an IOTA node to get data from.
+        #[arg(long, default_value = "http://localhost:9000")]
+        node_url: String,
+        #[arg(long, default_value_t = 1)]
+        num_workers: usize,
```
thibault-martinez, line 42: Add doc please

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,167 @@
...
+                // Spawn the metrics worker
+                let handle = cancel_token.clone();
+                tasks.spawn(async move {
+                    let worker = IotaNamesWorker::new(
+                        IotaNamesConfig::from_env().unwrap_or_default(),
+                        Arc::new(IotaNamesMetrics::new(&registry)),
+                    );
+
+                    tokio::select! {
+                        res = run_iota_names_reader(worker, &node_url, &registry, handle.clone(), num_workers) => {res?;},
```
thibault-martinez: Are the brackets really necessary around `{res?;}`?

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,167 @@
...
+                        if let Some(Ok(Err(err))) = res {
+                            tracing::error!("a worker failed with error: {err}");
+                            exit_code = Err(err);
+                        }
+                    },
+                }
+
+                cancel_token.cancel();
+
+                // Allow the user to abort if the tasks aren't shutting down quickly.
```
thibault-martinez, line 105: Are we doing this in the other indexers?

crates/iota-names-indexer/src/main.rs
```diff
@@ -4,57 +4,167 @@
...
+}
+
+fn set_up_logging() -> Result<()> {
+    std::panic::set_hook(Box::new(|p| {
+        error!("{}", p);
+    }));
+
+    let registry = tracing_subscriber::registry();
+
+    let registry = {
```
thibault-martinez: Why not chain the previous call instead of doing this?

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: nope

crates/iota-names-indexer/src/main.rs
DaughterOfMars: No, but it's good practice

crates/iota-names-indexer/src/main.rs
DaughterOfMars: not the same type

crates/iota-names-indexer/src/main.rs
DaughterOfMars: done

crates/iota-names-indexer/src/main.rs
DaughterOfMars: done

crates/iota-names-indexer/src/main.rs
DaughterOfMars: this was leftover from previous attempts, thanks.

crates/iota-names-indexer/src/main.rs
thibault-martinez: I tried locally, it's fine

crates/iota-names-indexer/src/worker.rs
```diff
@@ -15,30 +17,67 @@ use iota_types::{
...
+    );
+    executor.register(worker_pool).await?;
+
+    info!("Connecting to node at {node_url}");
+    executor
+        .run(
+            PathBuf::from("./chk".to_string()), /* path to a local directory
+                                                 * where
+                                                 * checkpoints
+                                                 * are stored. */
```
Alex6323: the formatter did that, right?

crates/iota-names-indexer/src/main.rs
DaughterOfMars: ah, fixed it

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: Probably. want me to change it?

crates/iota-names-indexer/src/main.rs
thibault-martinez: Feels like there has to be a better way to do this but ok

thibault-martinez (approved):

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: time to ✅  then :)

crates/iota-names-indexer/src/worker.rs
thibault-martinez: Yes

Thoralf-M (approved):

crates/iota-names-indexer/src/worker.rs
thibault-martinez: Can be later so we don't wait another CI cycle tho

DaughterOfMars: too late

thibault-martinez: alles gut

Alex6323: you don't have to if you're fine with it. But I wanted to make you aware of it in case you missed the formatter doing that. It certainly doesn't look so nice :trollface: 

also ... .to_string() should not be necessary I think

Alex6323 (approved):

PR #7123, opened 2025-05-28
feat(iota-names-indexer): Add event types

DaughterOfMars:
# Description of change

Adds event types to the indexer.


crates/iota-names-indexer/src/events.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use iota_names::{config::IotaNamesConfig, domain::Domain, registry::NameRecord};
+use iota_types::{base_types::IotaAddress, event::Event};
+use serde::{Deserialize, Serialize};
+
+pub(crate) enum IotaNamesEvent {
+    IotaNamesRegistry(IotaNamesRegistryEvent),
+    IotaNamesReverseRegistry(IotaNamesReverseRegistryEvent),
+    AuctionStarted(AuctionStartedEvent),
+    Bid(BidEvent),
```
thibault-martinez: If it's just `BidEvent` in the SC, can you please do a PR to rename it there to `AuctionBidEvent`?

DaughterOfMars: sure

DaughterOfMars: https://github.com/iotaledger/iota-names/pull/217

crates/iota-names-indexer/src/events.rs
```diff
@@ -0,0 +1,79 @@
...
+    AuctionExtended(AuctionExtendedEvent),
+    AuctionFinalized(AuctionFinalizedEvent),
+}
+
+impl IotaNamesEvent {
+    pub(crate) fn try_from_event(event: &Event, config: &IotaNamesConfig) -> anyhow::Result<Self> {
+        anyhow::ensure!(
+            event.package_id == config.package_address.into(),
```
thibault-martinez: Not too sure about this, why would it always be from the main package?

crates/iota-names-indexer/src/events.rs
Thoralf-M: Then also update it here now to AuctionBidEvent

crates/iota-names-indexer/src/events.rs
Thoralf-M: The package_id is the package which was called in a PTB which ended up with this event, right?
So if one writes a wrapper around iota-names and calls a function there, the package_id would be from the wrapper package
We should check if `event.type_.address` is equal to any of our package ids, so iota-names, subdomains, auction, wherever we defined an event

thibault-martinez: Is it not the package where the event was emitted?

DaughterOfMars: Is there actually a difference between `event.package_id` and `event.type_.address`?

Thoralf-M: > Is it not the package where the event was emitted?

Probably yes, I think I just mixed it up with the event form another package

> Is there actually a difference between event.package_id and event.type_.address?

Maybe not, try it
But as Thibault said, the evens we want can definitely be from different packages

Thoralf-M: Since we don't have all the package ids available in the config at the moment, maybe just add a TODO/issue and ignore the filtering here

crates/iota-names-indexer/src/events.rs
```diff
@@ -0,0 +1,79 @@
...
+    end_timestamp_ms: u64,
+}
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct AuctionFinalizedEvent {
+    domain: Domain,
+    start_timestamp_ms: u64,
+    end_timestamp_ms: u64,
+    winning_bid: u64,
+    winner: IotaAddress,
```
Thoralf-M: All the fields should be publicly accessible

crates/iota-names-indexer/src/events.rs
DaughterOfMars: It's fine for now to filter because the one package we do have is the one we need

Thoralf-M: No, auctions for example have a different package ID already

thibault-martinez: I think she was saying we don't yet have metrics for them anyway

crates/iota-names-indexer/src/events.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl IotaNamesEvent {
+    pub(crate) fn try_from_event(event: &Event, config: &IotaNamesConfig) -> anyhow::Result<Self> {
+        anyhow::ensure!(
+            event.package_id == config.package_address.into(),
+            "Invalid event package: {}",
+            event.package_id
+        );
+        Ok(match event.type_.name.as_str() {
+            "IotaNamesRegistryEvent" => Self::IotaNamesRegistry(bcs::from_bytes(&event.contents)?),
```
Thoralf-M: The bcs decoding doesn't work, it fails with `unexpected end of input`

crates/iota-names-indexer/src/worker.rs
```diff
@@ -70,19 +69,16 @@ impl IotaNamesWorker {
...
                 self.metrics.total_name_records.add(1);
-                // TODO: deserialize to get the name lengths
-                // let register_event =
-                //     bcs::from_bytes::<IotaNamesRegistryEvent>(&
-                // event_bcs_bytes)?;
-                // println!("Register event: {register_event:#?}");
             }
+            IotaNamesEvent::IotaNamesReverseRegistry(_event) => (),
+            IotaNamesEvent::AuctionStarted(_event) => (),
+            IotaNamesEvent::Bid(_event) => (),
```
thibault-martinez:
```suggestion
            IotaNamesEvent::AuctionBid(_event) => (),
```
?

crates/iota-names-indexer/src/events.rs
thibault-martinez: It's a String in the SC but an actual Domain here

crates/iota-names-indexer/src/events.rs
Thoralf-M: Ok not in this PR, but just moves the problem to the next person that wants to add metrics for them
Also it doesn't work like it's not, you would need to set the env variables in the indexer docker container

thibault-martinez: Regardless of all this, what do you think of making the return type `anyhow::Result<Option<Self>>` instead so the errors are only for actual IOTA-Names events? We can properly log errors then

Thoralf-M: Sounds like a good idea

Thoralf-M (approved):

thibault-martinez (approved):

PR #7128, opened 2025-05-28
feat(iota-names-indexer): Add balance metric

### LINKED ISSUE #7029: IOTA-Names metrics: balance of IOTA-Names over time

DaughterOfMars:
# Description of change

Adds a metric for the `IotaNames` object balance.


crates/iota-names-indexer/src/worker.rs
```diff
@@ -120,6 +126,32 @@ impl Worker for IotaNamesWorker {
...
+                Identifier::from_str("dummy_field")?,
+                MoveTypeLayout::Bool,
+            )],
+        }));
+        let balance_object_id = iota_types::dynamic_field::derive_dynamic_field_id(
+            self.config.object_id,
+            &TypeTag::Struct(Box::new(config_type)),
+            &IotaJsonValue::new(serde_json::json!({ "dummy_field": false }))?
+                .to_bcs_bytes(&layout)?,
+        )?;
```
thibault-martinez, line 145: Any way to not construct these 3 seemingly constant things for every single checkpoint?

thibault-martinez (changes requested):
Need to add it to the dashboard JSON

crates/iota-names-indexer/src/metrics.rs
```diff
@@ -11,6 +11,7 @@ use tracing::info;
 
 pub(crate) struct IotaNamesMetrics {
     pub total_name_records: IntGauge,
+    pub iota_names_balance: IntGauge,
```
thibault-martinez, line 14: Do we have any interest naming it to signify that it's a balance in IOTA? Since there is a balance dynfield per coin, even though there's only one for now

crates/iota-names-indexer/src/worker.rs
DaughterOfMars: Yes, but it might get messy. TBH I hate this code. I will be looking for a better way tomorrow

thibault-martinez: In the end we just need the `balance_object_id`, we could store it in the worker. Syncing will take a VERY long time already, anything we can spare here would be nice.

crates/iota-names-indexer/src/metrics.rs
DaughterOfMars: I don't right now

crates/iota-names-indexer/src/worker.rs
thibault-martinez: If you don't want to do it now, please open an issue 🙏🏻

thibault-martinez: I will open an issue then

thibault-martinez: https://github.com/iotaledger/iota/issues/7163

Thoralf-M (approved):

thibault-martinez (approved):

PR #7133, opened 2025-05-28
feat(iota-indexer, iota-types, iota-json-rpc-api, iota-graphql-rpc): Introduce `TransactionFilter::WrappedOrDeletedObject(object_id)`

### LINKED ISSUE #6756: `TransactionFilter::ChangedObject(object_id)` does not return deleted objects
## Description

Currently, when using `TransactionFilter::ChangedObject(object_id)` with the Indexer's `queryTransactionBlocks` RPC endpoint, transactions that **_changed_** the provided object are returned. This currently includes `created`, `mutated` and `unwrapped` objects. However, it doesn't include `deleted` objects.

Following issue should investigate if

1) if the filter should also return transactions for deleted objects
2) if it actually would be possible to also return transactions as part of the same query
3) if a new filter should be introduced (e.g. `TransactionFilter::DeletedObject(object_id)`
4) or if an existing transaction filter can/should be used to handle the query

---

Test which confirms the issue:

```rust
#[test]
fn get_trasaction_for_deleted_object() {
    let ApiTestSetup {
        runtime,
        store,
        client,
        cluster,
    } = ApiTestSetup::get_or_init();

    runtime.block_on(async move {
        indexer_wait_for_checkpoint(store, 1).await;

        // Publish NFT package and create an NFT
        let context = &cluster.wallet;
        let (package_id, _, _) = publish_nfts_package(context).await;

        let (sender, nft_object_id, tx_digest) = create_nft(context, package_id).await;

        indexer_wait_for_transaction(tx_digest, store, client).await;

        // Retrieve the latest object reference (which includes version) for deletion.
        let nft_object_ref = cluster.get_latest_object_ref(&nft_object_id).await;

        // Delete the NFT
        let delete_tx = delete_nft(context, sender, package_id, nft_object_ref).await;
        indexer_wait_for_transaction(delete_tx.digest, store, client).await;

        let res = client.query_transaction_blocks(IotaTransactionBlockResponseQuery {
            filter: Some(TransactionFilter::ChangedObject(nft_object_id)),
            options: Some(IotaTransactionBlockResponseOptions::full_content()),
        },
        None,
        None,
        None,
        ).await.expect("RPC call should succeed");

        let digests = res.data.iter().map(|tx| tx.digest).collect::<Vec<_>>();

        // Check if delete_tx.digest is in the result
        assert!(digests.contains(&delete_tx.digest), "Expected delete transaction to be found");
    });
}
```

Result: **`Expected delete transaction to be found`**

samuel-rufi:
# Description of change

Introduces a new `TransactionFilter::WrappedOrDeletedObject(object_id)` variant to filter transactions that either `wrapped`, `created and then wrapped`, `deleted`, or `unwrapped and then deleted` the specified object.

**Key additions:**
- adds a migration for the new Indexer `tx_wrapped_or_deleted_objects` table
- updates the checkpoint handler to index all tombstone objects i.e., wrapped, deleted, or unwrapped then deleted - as well as objects that were created and immediately wrapped. The latter are identified using the new `created_and_immediately_wrapped_objects` function, which derives such cases from the effects' object changes
- adds support for filtering in the Indexer JSON-RPC API and GraphQL. As with Sui, support was not added to `iota-json-rpc` since the required indices are not present and intentionally not added (https://github.com/MystenLabs/sui/pull/19614).
- adds relevant Indexer RPC and GraphQL test cases

Fixes https://github.com/iotaledger/iota/issues/6756

## Type of change

- Non-breaking change which adds functionality

DB migrations were tested through:
`diesel database reset --database-url postgres://postgres:postgrespw@localhost:5432/iota_indexer --migration-dir migrations/pg/`

Synchronization on a production-like database yet to do.
All other test classes are not necessary for this patch - we don't touch migration objects and don't touch Docker deployment.

- [X] Restart of indexer synchronization locally without resetting the database.

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [X] Indexer: Introduces a new `tx_wrapped_or_deleted_objects` table as part of a new database migration. Adds a `TransactionFilterV2::WrappedOrDeletedObject(object_id)` variant to filter transactions that either `wrapped`, `created and then wrapped`, `deleted`, or `unwrapped and then deleted` the specified object. Introduces the new `IndexerApi::query_transaction_blocks_v2` JSON-RPC API.
- [X] JSON-RPC: Introduces a new `TransactionFilterV2::WrappedOrDeletedObject(object_id)` variant to filter transactions that either `wrapped`, `created and then wrapped`, `deleted`, or `unwrapped and then deleted` the specified object. Introduces the new `IndexerApi::query_transaction_blocks_v2` JSON-RPC API.
- [X] GraphQL: Introduces a new `TransactionFilter::WrappedOrDeletedObject(object_id)` variant to filter transactions that either `wrapped`, `created and then wrapped`, `deleted`, or `unwrapped and then deleted` the specified object.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


piotrm50 (approved):
Only looked at `iota-types`changes (only crate I own) and it lgtm

crates/iota-indexer/src/types.rs
```diff
@@ -535,7 +535,13 @@ pub struct TxIndex {
     pub transaction_digest: TransactionDigest,
     pub checkpoint_sequence_number: u64,
     pub input_objects: Vec<ObjectID>,
+    /// Objects that were mutated, created and unwrapped objects. In other
+    /// words, all objects that still exist in the object state after this
+    /// transaction.
     pub changed_objects: Vec<ObjectID>,
+    /// Objects that were either wrapped immediately after being created,
+    /// deleted, or deleted immediately after being unwrapped.
+    pub wrapped_or_deleted_objects: Vec<ObjectID>,
```
kodemartin: This breaks backward compatibility of the Rust API. 

Could you think of an alternative, backward compatible way to add the new feature?

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -911,53 +911,67 @@ impl PgIndexerStore {
...
+                    tx_changed_objects,
+                    tx_wrapped_or_deleted_objects,
+                    tx_pkgs,
+                    tx_mods,
+                    tx_funs,
+                    tx_digests,
+                    tx_kinds,
+                )
+            },
+        );
```
kodemartin: Please consider simplifying.

See also [this comment](https://github.com/iotaledger/iota/pull/7235/files#r2132072644)

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE tx_wrapped_or_deleted_objects (
+                                  tx_sequence_number          BIGINT       NOT NULL,
+                                  object_id                   BYTEA        NOT NULL,
+                                  sender                      BYTEA        NOT NULL,
+                                  PRIMARY KEY(object_id, tx_sequence_number)
+);
+CREATE INDEX tx_wrapped_or_deleted_objects_tx_sequence_number_index ON tx_wrapped_or_deleted_objects (tx_sequence_number);
+CREATE INDEX tx_wrapped_or_deleted_objects_sender ON tx_wrapped_or_deleted_objects (sender, object_id, tx_sequence_number);
```
kodemartin, line 8: Why do we need this index?

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2366,6 +2366,10 @@ pub enum TransactionFilter {
     /// Query by changed object, including created, mutated and unwrapped
     /// objects.
     ChangedObject(ObjectID),
+    /// Query transactions that wrapped or deleted the specified object.
+    /// Includes transactions that either created and immediately wrapped
+    /// the object or unwrapped and immediately deleted it.
+    WrappedOrDeletedObject(ObjectID),
```
kodemartin: This isn't backward compatible either for the Rust API.

kodemartin (commented):
Hey @samuel-rufi, good job! I especially liked the use of `graphql-e2e-tests`. That being said, I left a few specific comments to consider. Plus I have the following general comments:

* This affects the tooling components, and albeit a new feature it is best to ship together. So please create a feature branch.
* You can consider consolidating the graphql-e2e-test into a single one: If I am not mistaken they all use the same package.
* Backward compatibility breaks for the Rust API. Please consider alternatives.

P.S. This could have also broken down into smaller PR. `json-rpc`, `indexer`, `graphql`. Please take it into consideration for future PRs.

crates/iota-test-transaction-builder/src/lib.rs
```diff
@@ -654,24 +654,42 @@ pub async fn emit_new_random_u128(
...
 
+/// Executes a transaction to publish the `nfts` package and returns the package
+/// id, id of the gas object used, and the digest of the transaction.
+pub async fn publish_nfts_package(
+    context: &WalletContext,
+) -> (ObjectID, ObjectID, TransactionDigest) {
+    publish_example_package(context, "nft").await
+}
+
+/// Executes a transaction to publish the `nfts` package and returns the package
```
Thoralf-M: Should be `simple_warrior`

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
samuel-rufi: Added to be in line with the index definitions of `tx_input_objects` and `tx_changed_objects`

crates/iota-test-transaction-builder/src/lib.rs
samuel-rufi: https://github.com/iotaledger/iota/pull/7133/commits/4c5f28a90f0f029b156cf6ed316ff82e31362d2a

crates/iota-indexer/src/types.rs
samuel-rufi: Good point. It breaks the `IndexerStore` trait which might be used by services outside of the crate. Introducing a `TxIndexV2` and an additional `persist_tx_indices_v2` solves this.

The checkpoint worker handler pipeline was updated to work with the new `TxIndexV2`. In addition, I changed the visibility of `CheckpointHandler::index_transaction` to private as it seems that the visibility of this function is not set correctly set, based on following:
- all other `CheckpointHandler` functions are private too, including the `index_transactions`
- Sui only has `index_transactions` which contains the extended logic of our `index_transaction`

Currently, there is a public function `pub async fn start_tx_checkpoint_commit_task<S>(...)` which also makes use of the `TxIndexV2` as part of the checkpoint handler pipeline. I checked our repo and the full Sui repo, it's not used outside of the iota-indexer. I changed the visibility to `pub(crate)` as I think it shouldn't be public as of a visibility bug. In general, to me it seems the visibility of types and functions in general is not properly set in many cases which makes versioning of core Rust APIs more difficult.

Please see the changes in: https://github.com/iotaledger/iota/pull/7133/commits/e4da2a5032f591310c671d809401d34e648a6bf5

Do you think of anything else that would need to be changed?

crates/iota-json-rpc-types/src/iota_transaction.rs
samuel-rufi: Will handle this shortly 👍

crates/iota-test-transaction-builder/src/lib.rs
```diff
@@ -654,7 +655,29 @@ pub async fn emit_new_random_u128(
...
+    );
+
+    let resp = context.execute_transaction_must_succeed(tx).await;
+    let package_id = get_new_package_obj_from_response(&resp).unwrap().0;
+    (package_id, resp.digest)
+}
+
+/// Executes a transaction to publish the `nft` package and returns the package
 /// id, id of the gas object used, and the digest of the transaction.
 pub async fn publish_nfts_package(
```
samuel-rufi, line 684: Should be adjusted in a future PR as it might produce flaky tests, please see https://github.com/iotaledger/iota/issues/7337

crates/iota-indexer/src/types.rs
kodemartin: I am afraid this is also breaking the API:

* Adding a new method in a public trait is a breaking change if no default implementation is provided. Implementors will have to implement the new method.
* Changing the visibility of any component from public to private is also a breaking change.

But the idea of migrating to a private API somehow is legit. Thus, you might want to consider something in the line of following:

* Use a new private type `TxIndexExtension` that includes only the new indices. This can be then safely extended in the future if we need to add new indices.
* Define a new private supertrait of `IndexerStore`  with a new method that persists the index extension. 
* Implement that for the `PgIndexerStore` or any relevant type that handles writing, and use the private methods internally instead of the `IndexerStore`.

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
kodemartin: The question remains though: Why do we need to add it? If we can leave without it we can save the implicit performance costs while writing.

crates/iota-indexer/src/handlers/checkpoint_handler.rs
```diff
@@ -511,6 +511,14 @@ impl CheckpointHandler {
             .map(|(object_ref, _owner, _write_kind)| object_ref.0)
             .collect::<Vec<_>>();
 
+        // Wrapped or deleted objects
+        let wrapped_or_deleted_objects = fx
```
miker83z: I wanted to bring a schema just to be sure about the changes. What you get here with these functions is:

With all_tombstones():
- `deleted` -> input_state: `exist`, output_state: `not_exist`
- `wrapped` -> input_state: `exist`, output_state: `not_exist`
- `unwrapped_then_deleted` -> input_state: `not_exist`, output_state: `not_exist`

With created_and_immediately_wrapped_objects():
- `created_then_wrapped` -> input_state: `not_exist`, output_state: `not_exist`

With all_changed_objects() (above this line I am commenting):
- `mutated` -> input_state: `exist`, output_state: `exist`
- `created` -> input_state: `not_exist`, output_state: `exist`
- `unwrapped` -> input_state: `not_exist`, output_state: `exist`


The 3 functions complement themselves, making it possible to reach all kinds of objects. 


Upstream, we find an additional field in `TxIndex` that allows to reach all kinds of objects (Including  `created_then_wrapped` and `unwrapped_then_deleted`): [affected_objects](https://github.com/MystenLabs/sui/pull/19618/files#diff-a90190dfcf57eae4500c4da22ce0cab4397028c6a4be2576e39f2b5755101e26R487) introduced in https://github.com/MystenLabs/sui/pull/19618 which uses the `iota_types::TransactionEffects` function object_changes() (which is already present in our sdk).
Keep in mind that `iota_types::TransactionEffects::object_changes()` here:
- provides a more complete set of objects in respect to `iota_json_rpc_types::IotaTransactionBlockResponse`'s field `object_changes`, which only [returns mutated and created objects](https://github.com/iotaledger/iota/blob/7a7adddee08b3ed7ea1cfe98c7320da0675a46a1/crates/iota-json-rpc/src/object_changes.rs#L17); moreover, `iota_json_rpc_types::IotaTransactionBlockEffects` doesn't have an equivalent object_changes() function or similar to obtain the same set of objects.
- provides the same set of objects as in `iota_graphql_rpc::TransactionBlockEffects` with the [function `object_changes`](https://github.com/iotaledger/iota/blob/7a7adddee08b3ed7ea1cfe98c7320da0675a46a1/crates/iota-graphql-rpc/src/types/transaction_block_effects.rs#L319).



I am not making any suggestion with this comment, I just wanted to point out what we have and what we don't!

crates/iota-types/src/effects/mod.rs
```diff
@@ -185,6 +185,26 @@ impl TransactionEffects {
             .collect()
     }
 
+    /// Returns all objects that were created + wrapped in the same transaction.
+    pub fn created_and_immediately_wrapped_objects(&self) -> Vec<(ObjectID, SequenceNumber)> {
```
miker83z: I'd call it `created_then_wrapped_objects` to keep the same style as `unwrapped_then_deleted`

crates/iota-indexer/src/types.rs
samuel-rufi: Hi @kodemartin , thanks for the suggestions. I applied the changes as discussed which should avoid any breaking changes, please see: https://github.com/iotaledger/iota/pull/7133/commits/fb81ac7f229b55ecf4e316414eaf776e7e61ea69

crates/iota-indexer/src/store/pg_indexer_store.rs
samuel-rufi: Simplified it as part of https://github.com/iotaledger/iota/pull/7133/commits/fb81ac7f229b55ecf4e316414eaf776e7e61ea69

crates/iota-types/src/effects/mod.rs
samuel-rufi: Adjusted in https://github.com/iotaledger/iota/pull/7133/commits/fb81ac7f229b55ecf4e316414eaf776e7e61ea69

crates/iota-indexer/src/handlers/checkpoint_handler.rs
samuel-rufi: Thanks @miker83z. What I maybe just want to add is that `all_tombstones` is about objects that lived before the transaction (that performs the operations on the objects), including for the `unwrapped_then_deleted`.

Also note, `unwrapped_then_deleted` and `created_then_wrapped` have the same input/output state but differ in the `IDOperation`.

Regarding `affected_objects`, we discussed it in the issue link https://github.com/iotaledger/iota/issues/6756 and decided to go with the minimal subset to provide the transaction filter feature. And as you mentioned, using the 3 functions above we reach all kind of objects, which are possible to fetch using the transaction filter.

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
samuel-rufi: Let me do some investigation on this

samuel-rufi: Yes, we need both indexes. The `tx_sequence_number` index for fast cursor-driven pagination.
And the composite index (sender, object_id, tx_sequence_number)  for any "by sender + object" filters, which we need in GraphQL if the additional transaction filter `sign_address` is specified:
https://github.com/iotaledger/iota/blob/1beda71adbf34bf07337a021ad93ffb7727da162/crates/iota-graphql-rpc/src/types/transaction_block/tx_lookups.rs#L270
https://github.com/iotaledger/iota/blob/1beda71adbf34bf07337a021ad93ffb7727da162/crates/iota-graphql-rpc/src/types/transaction_block/tx_lookups.rs#L442-L448

samuel-rufi:
> Hey @samuel-rufi, good job! I especially liked the use of `graphql-e2e-tests`. That being said, I left a few specific comments to consider. Plus I have the following general comments:
> 
> * This affects the tooling components, and albeit a new feature it is best to ship together. So please create a feature branch.
> * You can consider consolidating the graphql-e2e-test into a single one: If I am not mistaken they all use the same package.
> * Backward compatibility breaks for the Rust API. Please consider alternatives.
> 
> P.S. This could have also broken down into smaller PR. `json-rpc`, `indexer`, `graphql`. Please take it into consideration for future PRs.

@kodemartin regarding the open points left:

Regarding the consolidation of the GraphQL tests: I added them separately to improve maintainability and readability. I think the lack of proper commenting in the test files (at least I couldn't find a way to do it in the docs) makes it a bit challenging. Seems like we only can leave a comment as part of the GraphQL query itself. I will try to use that to restructure the tests.

The PR is targeting now a feature branch, I will coordinate this after https://github.com/iotaledger/iota/pull/7133#discussion_r2137330207 is resolved.

crates/iota-json-rpc-types/src/iota_transaction.rs
samuel-rufi: Fixed by https://github.com/iotaledger/iota/pull/7133/commits/73e8cc935ecfd0c0781d55eb6ef51271430e7aee

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2429,6 +2459,100 @@ impl Filter<EffectsWithInput> for TransactionFilter {
     }
 }
 
+#[serde_as]
+#[derive(Clone, Debug, JsonSchema, Serialize, Deserialize)]
```
kodemartin, line 2488: Add the [non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute to allow additions in the future without breaking the API.

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2429,6 +2459,100 @@ impl Filter<EffectsWithInput> for TransactionFilter {
...
+    FromAndToAddress { from: IotaAddress, to: IotaAddress },
+    /// Query txs that have a given address as sender or recipient.
+    FromOrToAddress { addr: IotaAddress },
+    /// Query by transaction kind
+    TransactionKind(IotaTransactionKind),
+    /// Query transactions of any given kind in the input.
+    TransactionKindIn(Vec<IotaTransactionKind>),
+}
+
+impl Filter<EffectsWithInput> for TransactionFilterV2 {
```
kodemartin: You can avoid keeping the same logic in two places if you implement a simple conversion method to `TransactionFilter`. E.g.

```rust
impl TransactionFilterV2 {
    fn as_v1(&self) -> Option<TransactionFilter> {
        match self {
            TransactionFilterV2::InputObject(o) => Some(TransactionFilter::InputObject(*o)),
            // ...repeat for all variants that exist in both...
            _ => None, // for new or incompatible variants
        }
    }
}

impl Filter<EffectsWithInput> for TransactionFilterV2 {
    fn matches(&self, item: &EffectsWithInput) -> bool {
        if let Some(v1) = self.as_v1() {
            return v1.matches(item);
        }
        // Fallback for new V2-only variants:
        match self {
            TransactionFilterV2::WrappedOrDeletedObject(o) => {
                item.effects.wrapped().iter()
                    .chain(item.effects.deleted())
                    .chain(item.effects.unwrapped_then_deleted())
                    .any(|oref| &oref.object_id == o)
            }
            // ...other V2-only logic...
            _ => false,
        }
    }
}
```

crates/iota-json-rpc-api/src/extended.rs
```diff
@@ -79,4 +79,19 @@ pub trait ExtendedApi {
     /// Exclusively served by the indexer.
     #[method(name = "getParticipationMetrics")]
     async fn get_participation_metrics(&self) -> RpcResult<ParticipationMetrics>;
+
+    /// Return list of transactions for a specified query criteria.
+    #[rustfmt::skip]
+    #[method(name = "queryTransactionBlocksV2")]
```
kodemartin: We can use the `version` attribute in the v1 `query_transaction_blocks` and keep the same method name here for web clients to allow seamless usage.

```rust
    #[method(name = "queryTransactionBlocks")]
```

crates/iota-indexer/src/types.rs
kodemartin: I have two remarks on the changes:

* We want to add indices in the future, not define alternative, so the enum does not seems plausible for `TxIndexExt`
* A `non_exhaustive` `TxIndexExt` could just wrap the `wrapped_or_deleted_objects` vector and be passed to the new `persist` method in `PgIndexerStoreExt` together with the inflexible `TxIndex` that we currently use.

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
kodemartin: This is a subtle issue. One has to consider if the write overhead with the additional index worth the while for supporting the `object_id + sender` combo.

In this class of queries `object_id` is the main query param followed by `tx_sequence_number` range. Then `sender` may or may not exists as an additional constraint. So ultimately if we don't add the `(sender, object_id, tx_sequence_number)` index we end up with having to scan all the senders that map to `object_id` for the given `tx_sequence_number` range. If this is in the order of 1000 it might not be an issue. Especially if we still restrain it in the index, eg. as part of the primary key `(object_id, tx_sequence_number, sender)`.

Anyway we can do some proper benchmarking and revisit this in the future, as it affects all `tx_indices`.

kodemartin (changes requested):

richardmands (approved):

crates/iota-json-rpc-types/src/iota_transaction.rs
samuel-rufi: Added in https://github.com/iotaledger/iota/pull/7133/commits/1bebfda263b2ab2a078a3f921ffa9c605aab8d76

crates/iota-json-rpc-types/src/iota_transaction.rs
samuel-rufi: Added in https://github.com/iotaledger/iota/pull/7133/commits/1bebfda263b2ab2a078a3f921ffa9c605aab8d76

crates/iota-indexer/src/types.rs
samuel-rufi: Hi @kodemartin, thanks, makes sense! Please see https://github.com/iotaledger/iota/pull/7133/commits/d1eb04787177670579f636e068f05d934f8edab2 where I followed your suggestions.

In addition, to more easily work with and provide the both types required (`TxIndex`, `TxIndexExt`) I kept `TxIndexV2`, which bundles them into a single wrapper.

crates/iota-json-rpc-api/src/extended.rs
samuel-rufi: Since we can’t guarantee that every client will send a version field (I saw it's optional), tagging only V1 with version <= "1.0.0" will push unversioned calls into V2 and break old clients. Shouldn't we instead leave V1 untagged?

And use the `_v2` suffix as we did for `get_latest_iota_system_state_v2` and `get_dynamic_field_object_v2`?

kodemartin: But we only add an additional variant parameter in the filter. The problem would occur if a client uses the filter AND requests a previous version of the server. 

Other than that the new client is fully backward-compatible (meaning one can issue JSON-RPC request with all other filter variants) to new and old versions of the JSON-RPC server.

samuel-rufi: I see and seems reasonable, added it here: https://github.com/iotaledger/iota/pull/7133/commits/2b1d957e18217348dc5d5b40f42aa5acd98ee012

crates/iota-indexer/migrations/pg/2025-05-28-174842_index_wrapped_or_deleted_objects/up.sql
samuel-rufi: I created a tracking issue here: https://github.com/iotaledger/iota/issues/7509

samuel-rufi:
Seems like the CI is failing after the latest change. Investigating

crates/iota-json-rpc-api/src/extended.rs
```diff
@@ -79,4 +79,19 @@ pub trait ExtendedApi {
...
+        &self,
+        /// the transaction query criteria.
+        query: IotaTransactionBlockResponseQueryV2,
+        /// An optional paging cursor. If provided, the query will start from the next item after the specified cursor. Default to start from the first item if not specified.
+        cursor: Option<TransactionDigest>,
+        /// Maximum item returned per page, default to QUERY_MAX_RESULT_LIMIT if not specified.
+        limit: Option<usize>,
+        /// query result ordering, default to false (ascending order), oldest record first.
+        descending_order: Option<bool>,
+    ) -> RpcResult<TransactionBlocksPage>;
```
kodemartin: @samuel-rufi just realized that this is part of the `ExtendedApi`. There is no reason to separate it. We can't version it properly like this and it is the same method. You can still add it in the `IndexerApi` and implement that in the node as well, throwing `UnsupportedFilter` error in case the new filter is used.

crates/iota-json-rpc-api/src/indexer.rs
```diff
@@ -45,7 +45,7 @@ pub trait IndexerApi {
 
     /// Return list of transactions for a specified query criteria.
     #[rustfmt::skip]
-    #[method(name = "queryTransactionBlocks")]
+    #[method(name = "queryTransactionBlocks", version <= "1.0.0")]
```
kodemartin: We have to account for the largest release version prior to the target develop. E.g.
```suggestion
    #[method(name = "queryTransactionBlocks", version <= "1.2.10")]
```

crates/iota-json-rpc-api/src/extended.rs
samuel-rufi: Moved it as part of: https://github.com/iotaledger/iota/pull/7133/commits/d9d2a68e78f483632f8579ab1af26231e2ed42b0

crates/iota-json-rpc-api/src/indexer.rs
samuel-rufi: So considering the latest stable release it should be 1.2.3.
I tested it locally with curl and if no header is provided, it will target v2, so also the new filter is available.
If a header is provided, as expected it will target V1 if the provided version is <= 1.2.3.

I also looked into the failing rpc-tests as they can't match the endpoint. I tried to inject a header, however this didn't help. I checked the raw request body that is received on the server and they use the correct internal method name, but it still somehow fail to match. When replacing the `client.query_transaction_blocks()` usages with `client.query_transaction_blocks_v2` the test passes.

Do you maybe know what I'm missing here? I assume they should at least work when a header is provided?

crates/iota-json-rpc-api/src/extended.rs
samuel-rufi: Resolved in favor of https://github.com/iotaledger/iota/pull/7133#discussion_r2164082274

crates/iota-json-rpc-api/src/indexer.rs
kodemartin: > So considering the latest stable release it should be 1.2.3. I tested it locally with curl and if no header is provided, it will target v2, so also the new filter is available. If a header is provided, as expected it will target V1 if the provided version is <= 1.2.3.

The latest is `1.2.3` but the current develop is `1.3.0`. So we would like to catch any patches `1.2.x` until this new feature lands. And while we don't have a `< 1.3.0` which would be what we wanted here, we can leave some room by increasing the patch number. Hence the `1.2.10` that leaves room for additional patches to `1.2.3`.

> 
> I also looked into the failing rpc-tests as they can't match the endpoint. I tried to inject a header, however this didn't help. I checked the raw request body that is received on the server and they use the correct internal method name, but it still somehow fail to match. When replacing the `client.query_transaction_blocks()` usages with `client.query_transaction_blocks_v2` the test passes.
> 
>
> Do you maybe know what I'm missing here? I assume they should at least work when a header is provided?


Aren't the tests using the Rust API? It makes total sense that you should pass the `_v2` method to make use of the new filter. The version used above and the headers are relevant to http clients like `curl`.

samuel-rufi: Thanks for the further clarification on the selected version.

> Aren't the tests using the Rust API? It makes total sense that you should pass the _v2 method to make use of the new filter. The version used above and the headers are relevant to http clients like curl.

Yes, the tests are using the Rust API. All clients that make use of the Rust API `client.query_transaction_blocks()` are sending the internal method name and will fail/reject, with or without header in this function:

```rust
    pub fn route<'c, 'a: 'c, 'b: 'c>(&'a self, method: &'b str, version: Option<&str>) -> &'c str {
        // Reject direct access to the old methods
        if self.route_to_methods.contains(method) {
            "INVALID_ROUTING"
        } else if self.disable_routing {
            method
        } else {
            // Modify the method name if routing is enabled
            match (version, self.routes.get(method)) {
                (Some(v), Some(route)) if route.matches(v) => route.route_to.as_str(),
                _ => method,
            }
        }
    }
```

Methods that are old will be pushed to "route_to_methods" and as the RPC stub did send the exact name using `client.query_transaction_blocks()` it fails with "INVALID_ROUTING".
Removing that branch makes the tests work, however we are no longer protected about unsupported direct function calls. A task for the future might be to revisit this routing logic and maybe reject on specific methods only.

Changed internal `client.query_transaction_blocks()` to `client.query_transaction_blocks_v2() `: https://github.com/iotaledger/iota/pull/7133/commits/ef54d05541e95d1077f0cbd3e3470eb6a6f3c2ee

kodemartin (commented):
@samuel-rufi it looks good after the latest changes.

We don't have visibility on the CI though. Please rebase the source and target branches on top of the latest `develop` to resolve this issue as well.

crates/iota-json-rpc-api/src/indexer.rs
kodemartin: > Removing that branch makes the tests work, however we are no longer protected about unsupported direct function calls. A task for the future might be to revisit this routing logic and maybe reject on specific methods only.

There is indeed something off with that branch, but I would expect us to support direct calls actually. This seems to be a side effect of tweaking with the RPC macro and associating the Rust client method to the versioned variant, which is hidden by the spec and hence by other clients.

If direct access is not supported then the Rust API is no longer backward compatible.

So I suggest we remove the branch altogether.

kodemartin (approved):
Looks good, subject to fixing `clippy` and passing the CI.

PR #7135, opened 2025-05-28
fix(indexer, json-rpc): Fix get total supply endpoint of coin api

### LINKED ISSUE #7134: [indexer, json-rpc]: Get total supply endpoint doesn't work for coin manager coins
## Bug description

Similar to https://github.com/iotaledger/iota/issues/5772 but for get_total_supply endpoint.

The root cause of the issue is the same, the code is looking for treasurycap/coinmetadata in package that initially, but those objects may be hidden inside of CoinManager object, potentially in a different package.

tomxey:
# Description of change

Fix get total supply endpoint of coin api

One coin manager case couldn't be supported by the node easily, a separate issue with more context was created here as a follow up: https://github.com/iotaledger/iota/issues/7507

Fixes https://github.com/iotaledger/iota/issues/7134

## Type of change

Unchecked test-classes were not needed for this patch as they don't touch ingestion or database migrations.   

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


samuel-rufi:
One Rust test fails currently as of a new error message. Needs to be revisited.
**EDIT**: test expectations had to be adjusted as of the additional coin manager branch.

The error message in the node was aligned to cover both cases. The error code `−32602` stays the same and therefore https://github.com/iotaledger/iota/pull/7135/commits/42d6b4706593213dde4da976cf4aa0bafa9a4e94 shouldn't be considered a breaking change. The JSON-RPC code is the only part of the error payload that downstream consumers are meant to rely on and the error messages should be treated as "human-readable hints" (not part of the formal contract).

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -2026,28 +2026,58 @@ impl IndexerReader {
...
-                    treasury_cap_obj_id
-                )))?;
-        Ok(TreasuryCap::try_from(treasury_cap_obj_object)?.total_supply)
+            .inspect_err(|e| tracing::error!("package_obj_type_cache poisoned: {:?}", e))
+            .map_err(|_| IndexerError::Generic("failed to acquire cache lock".into()))?;
+
+        let maybe_id = if let Some(&id) = cache.cache_get(&cache_key) {
+            id
+        } else {
+            let fetched = get_single_obj_id_from_package_publish(
```
kodemartin: It would be more efficient to use `get_singleton_object` that takes advantage of the `objects_package_module_name_full_type` index, similarly to how the `CoinManager` is retrieved currently.

The method used here first queries for the transaction that created the treasure cap, to lookup its id in the transaction effects and then do an additional query to get the object based on the id.

crates/iota-json-rpc/src/coin_api.rs
```diff
@@ -383,6 +376,70 @@ async fn find_package_object_id(
...
+    if let Ok(obj) = internal
+        .find_package_object(&tag.address.into(), TreasuryCap::type_(tag.clone()))
+        .await
+    {
+        let data = obj
+            .data
+            .try_as_move()
+            .ok_or_else(|| Error::Unexpected("Cannot get move contents".into()))?
+            .contents();
+        let tc = TreasuryCap::from_bcs_bytes(data).map_err(Error::from)?;
```
kodemartin, line 418: Similarly to `CoinManager` we could implement `TryFrom<Object>` and consolidate `coin_manager_total_supply` and `treasury_cap_total_supply` to use a common method for both types generic on a type `T: TryFrom<Object>`

kodemartin (changes requested):

crates/iota-indexer/src/indexer_reader.rs
samuel-rufi: Please see https://github.com/iotaledger/iota/pull/7135/commits/8ef90fde277aefd99843282038ee8eeaa0dfeff1

crates/iota-json-rpc/src/coin_api.rs
samuel-rufi: Please see https://github.com/iotaledger/iota/pull/7135/commits/83d3c827036441c437bd192de9df23f4dd522d5d

crates/iota-indexer/tests/rpc-tests/coin_api.rs
```diff
@@ -533,10 +533,141 @@ fn get_total_supply() {
...
+            cluster,
+            client,
+            immutable_metadata_coin_name.to_string(),
+        )
+        .await;
+
+        assert!(result_fullnode.is_some());
+        assert_eq!(result_fullnode, result_indexer);
+    });
+}
```
kodemartin: I think we can remove this. Testing the fullnode should not be part of the indexer testing suite.

kodemartin (approved):
Looks good, with one suggestion about a test case. Non-critical.

crates/iota-indexer/tests/rpc-tests/coin_api.rs
samuel-rufi: Removed it here: https://github.com/iotaledger/iota/pull/7135/commits/2c11898b82dc6abdaa516c9e8fb30d250e80141f

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -2026,28 +1963,65 @@ impl IndexerReader {
...
+    }
+
+    fn get_package_object_as<T>(&self, tag: StructTag) -> Result<Option<T>, IndexerError>
+    where
+        T: TryFrom<Object, Error = IotaError>,
+    {
+        let package_id: ObjectID = tag.address.into();
+        let cache_key = format!(
+            "{}{}",
+            package_id,
```
tomxey: nit: not necessarily an issue, but isn't package_id part of the tag.to_canonical_string(..) already?

No reason to have it twice in the cache key. (unless I'm wrong and its not present in the canonical str)

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -2026,28 +1963,65 @@ impl IndexerReader {
...
+        &self,
+        coin_struct: &StructTag,
+    ) -> Result<Option<Supply>, IndexerError> {
+        let tag = CoinManager::type_(coin_struct.clone());
+        Ok(self
+            .get_package_object_as::<CoinManager>(tag)?
+            .map(|mgr| mgr.treasury_cap.total_supply))
+    }
+
+    fn get_package_object_as<T>(&self, tag: StructTag) -> Result<Option<T>, IndexerError>
```
tomxey: I feel the name 'get_package_object' is a bit confusing. It's getting object by type, not limited to any particular package.
What would you say about just: 'get_object_as' ?

samuel-rufi: It gets a singleton object similarly as the previous `find_package_object` that was used to get the `TreasuryCap`.
Also it keeps populating the package type cache.
Therefore I would better keep the `get_package_object_as`.

crates/iota-indexer/src/indexer_reader.rs
samuel-rufi: Good catch, you're absolutely right. It's not needed as it's part of `to_canonical_string()` already. Appears it was like this already in the previous codebase https://github.com/iotaledger/iota/pull/7135/files#diff-59c7fa0841ea58f4030d1040643d35471b08f0d570bd91d8a35607cdeb86ccc3L2033-L2039)

Also, appears to be safe to change that as it's only used within this module.

Adjusted as part of  https://github.com/iotaledger/iota/pull/7135/commits/022c251f89bffc9af47b8827e49701c2ffee746e

crates/iota-indexer/src/indexer_reader.rs
tomxey: Actually, now that you mentioned it, I think we could also refactor the `package_obj_type_cache` name to just  `obj_type_cache`.

Originally package id was one of the parameters of the function and it was limiting the scope where we searched for the object. But now we are looking for object of particular type, not limited by any package id.

So my suggestion would be to remove the 'package' part both from `get_package_object_as` and `package_obj_type_cache`

samuel-rufi: Renamed it here: https://github.com/iotaledger/iota/pull/7135/commits/27bc7fe2dc5f41664cb478e1a0760c1f9beaed3d

samuel-rufi (approved):
Approving it with approval from @tomxey (as GH doesn't let him approve as he started the PR).

PR #7142, opened 2025-05-29
chore: fix clippy `large_enum_variant`

thibault-martinez:
Address the clippy lint `large_enum_variant` which has been made stable with 1.87.
The suggestion from clippy is to box the large variant, which is not always possible without breaking public API. For each instance you own, please confirm the right approach has been used.


thibault-martinez:
superseded by https://github.com/iotaledger/iota/pull/7545

PR #7146, opened 2025-05-29
upstream(iota-indexer): compatibility check using migration records

### LINKED ISSUE #7097: [indexer]: Compatibility check using migration records (#19156) (upstream)
upstream PR: https://github.com/MystenLabs/sui/pull/19156

sergiupopescu199:
# Description of change

This patch follows upstream changes in changing how compatibility checks are done by fully leveraging the migration records.

fixes #7097 

## Type of change

```shell
$ cargo test --features pg_integration --profile simulator -- --test-threads 1
$ cargo nextest run -p iota-graphql-rpc --features pg_integration --no-fail-fast --test-threads 1
```
- Started the indexer in **writer** & **reader** mode making sure it was syncing data as expected.

This patch affects the database migrations, especially during the reset database operations.

- [x] Restart of indexer synchronization locally resetting the database, testing migrations.

## Change checklist

- [x] I have followed the contribution guidelines for this project

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-indexer/src/schema.rs
```diff
@@ -592,67 +592,60 @@ diesel::joinable!(optimistic_tx_kinds -> optimistic_transactions (tx_insertion_o
...
-            tx_insertion_order,
-            tx_kinds,
-            tx_recipients,
-            tx_senders
-        );
-    };
-}
-pub use for_all_tables;
-for_all_tables!(diesel::allow_tables_to_appear_in_same_query);
+diesel::allow_tables_to_appear_in_same_query!(
```
tomxey: Did you use the `./scripts/indexer-schema/generate.sh` ?

crates/iota-indexer/src/db.rs
```diff
@@ -229,61 +241,101 @@ pub mod setup_postgres {
...
     }
 
+    /// Execute all unapplied migrations.
+    pub fn run_migrations(conn: &mut PoolConnection) -> Result<(), anyhow::Error> {
+        conn.run_pending_migrations(MIGRATIONS)
+            .map_err(|e| anyhow!("Failed to run migrations {e}"))?;
+        Ok(())
+    }
+
+    /// Checks that the local migration scripts is a prefix of the records in
```
tomxey: Actually, why do we check that it's a prefix, and not check that they are an exact match?

sergiupopescu199: We check that the local migrations are a prefix of the applied migrations in the database (rather than requiring an exact match) to allow for forward compatibility. This means the database can have additional migrations that were applied after the current client was built or deployed. As long as all local migrations have been applied in order, the client can work with a database that has newer migrations.

If we required an exact match, any new migration applied to the database—even if unrelated to the current client—would cause the check to fail. The prefix check ensures that the client’s expected migrations are present and in order, without being too strict about extra migrations that may have been added later.

This logic is implemented in our code, and the uniqueness of the `version` field (as a primary key) ensures there are no duplicates.

crates/iota-indexer/src/schema.rs
sergiupopescu199: will re-generate the schema using the script as I forgot about it, I suppose we should update the `README.md`, what do you think?

sergiupopescu199: re-generated schema using the script: https://github.com/iotaledger/iota/pull/7146/commits/88377e7e95778cf08ffbce762a529ed0c2ca04f0

tomxey: Yes, I think we could mention it in the README.
Not necessarily in this PR though.

crates/iota-indexer/src/db.rs
tomxey: In upstream the code runs migrations when starting writers, and runs prefix check when starting readers.

In our code we are running migrations in both cases now, so it seems to me the prefix case should not occur at all.
Consider modifying the check to verify that all migrations match exactly, not just the prefix.

crates/iota-indexer/src/db.rs
```diff
@@ -311,11 +363,86 @@ pub mod setup_postgres {
...
+        use crate::{
+            db::setup_postgres::{self, MIGRATIONS},
+            test_utils::TestDatabase,
+        };
+
+        fn database_url(db_name: &str) -> Secret<String> {
+            secrecy::Secret::new(format!(
+                "postgres://postgres:postgrespw@localhost:5432/{db_name}"
+            ))
+        }
```
kodemartin, line 410: This shares functionality with `common::get_indexer_db_url`, and is best to consolidate the two in `test_utils`.

Let's create a new issue to handle this.

crates/iota-indexer/src/db.rs
```diff
@@ -311,11 +363,86 @@ pub mod setup_postgres {
...
+        // Check that the migration records in the database created from the local
+        // schema pass the consistency check.
+        #[test]
+        fn db_migration_consistency_smoke_test() {
+            let mut database =
+                TestDatabase::new(database_url("db_migration_consistency_smoke_test"));
+            database.recreate();
+            let pool = database.to_connection_pool();
+            let mut conn = pool.get().unwrap();
+            setup_postgres::reset_database(&mut conn).unwrap();
```
kodemartin: This should be equivalent to
```rust
            database.reset_db();
```
no?

crates/iota-indexer/src/db.rs
```diff
@@ -311,11 +363,86 @@ pub mod setup_postgres {
...
+            setup_postgres::check_db_migration_consistency(&mut conn).unwrap();
```
kodemartin:
```suggestion
            setup_postgres::check_db_migration_consistency(&mut conn).unwrap();
            database.drop_if_exists();
```

crates/iota-indexer/src/db.rs
kodemartin: Please refactor accordingly the other tests after verifying.

crates/iota-indexer/src/db.rs
kodemartin: Please drop accordingly the test databases in the other tests as well.

crates/iota-indexer/src/db.rs
```diff
@@ -229,61 +241,101 @@ pub mod setup_postgres {
...
 
-        diesel::sql_query(
-            "
-        CREATE TABLE IF NOT EXISTS __diesel_schema_migrations (
-            version VARCHAR(50) PRIMARY KEY,
-            run_on TIMESTAMP NOT NULL DEFAULT NOW()
-        )",
-        )
-        .execute(conn)?;
+        conn.setup()?;
```
kodemartin, line 244: good one!

kodemartin (changes requested):
Looks good with some minor revision requests.

crates/iota-indexer/src/db.rs
```diff
@@ -229,61 +241,101 @@ pub mod setup_postgres {
...
+        Ok(())
+    }
+
+    /// Checks that the local migration scripts is a prefix of the records in
+    /// the database. This allows us run migration scripts against a DB at
+    /// anytime, without worrying about existing readers fail over.
+    /// We do however need to make sure that whenever we are deploying a new
+    /// version of either reader or writer, we must first run migration
+    /// scripts to ensure that there is not more local scripts than in the DB
+    /// record.
```
kodemartin: I would also add that this requires that new migrations are backward compatible with respect to the model.

crates/iota-indexer/src/db.rs
sergiupopescu199: Running migrations for both reader & writer is not correct, the reader could potentially reset the db which is not expected behavior, I've created an issue to fix this: https://github.com/iotaledger/iota/issues/7215

Based on this premise we'll leave the compatibility check for checking for prefixes.

crates/iota-indexer/src/db.rs
sergiupopescu199: https://github.com/iotaledger/iota/pull/7146/commits/0e8a4ff76cf1ae5bc72361b1c8a19cdedb5743fa

crates/iota-indexer/src/db.rs
sergiupopescu199: issue created: https://github.com/iotaledger/iota/issues/7224

crates/iota-indexer/src/db.rs
sergiupopescu199: https://github.com/iotaledger/iota/pull/7146/commits/fbf2ac857869957e304a80399ca6ff14640ee970

crates/iota-indexer/src/db.rs
sergiupopescu199: https://github.com/iotaledger/iota/pull/7146/commits/fbf2ac857869957e304a80399ca6ff14640ee970

crates/iota-indexer/src/db.rs
sergiupopescu199: https://github.com/iotaledger/iota/pull/7146/commits/fbf2ac857869957e304a80399ca6ff14640ee970

kodemartin (approved):
lgtm

tomxey (approved):

PR #7151, opened 2025-05-29
feat(iota): add `client remove-address` command

### LINKED ISSUE #7150: CLI: new command to remove an address in iota client
https://github.com/MystenLabs/sui/pull/21912

thibault-martinez:
Fixes https://github.com/iotaledger/iota/issues/7150

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Add `iota client remove-address` command
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-keys/src/keystore.rs
+++ b/crates/iota-keys/src/keystore.rs
@@ -37,10 +37,12 @@ pub enum Keystore {
     File(FileBasedKeystore),
     InMem(InMemKeystore),
 }
+
 #[enum_dispatch]
 pub trait AccountKeystore: Send + Sync {
     fn add_key(&mut self, alias: Option<String>, keypair: IotaKeyPair)
     -> Result<(), anyhow::Error>;
+    fn remove_key(&mut self, address: &IotaAddress) -> Result<(), anyhow::Error>;
     fn keys(&self) -> Vec<PublicKey>;
     fn get_key(&self, address: &IotaAddress) -> Result<&IotaKeyPair, anyhow::Error>;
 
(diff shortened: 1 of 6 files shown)


crates/iota/src/client_commands.rs
```diff
@@ -496,6 +496,8 @@ pub enum IotaClientCommands {
         #[arg(long, short)]
         profile_output: Option<PathBuf>,
     },
+    /// Remove an existing address by its alias or hexadecimal string.
```
Thoralf-M, line 499: I think this should mention that it removes the private key from the keystore and that there is no way to recover, if no backup was made

crates/iota-keys/src/keystore.rs
```diff
@@ -249,6 +251,13 @@ impl AccountKeystore for FileBasedKeystore {
         Ok(())
     }
 
+    fn remove_key(&mut self, address: &IotaAddress) -> Result<(), anyhow::Error> {
+        self.aliases.remove(address);
+        self.keys.remove(address);
+        self.save()?;
```
Thoralf-M, line 257: Do we also need to unset the active address here?

thibault-martinez: Good point. Just unset, or reset it to something else?

crates/iota/src/client_commands.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/7151/commits/94137616d5e36ee9a52362d3ae496ec70214c50a

crates/iota-keys/src/keystore.rs
Thoralf-M: Not sure how the logic is built around it, if it would automatically set a new one the next time we try to get it, then just unsetting it should be fine

thibault-martinez: I did it somewhere else because keystore has no concept of active address. Also confirmed that it already sets it to something else automatically.
https://github.com/iotaledger/iota/pull/7151/commits/73dbeda5780290a6d969c798ff710f893de07969

Thoralf-M (approved):

richardmands (approved):

PR #7152, opened 2025-05-29
refactor(indexer): use tx_global_order table

### LINKED ISSUE #6668: [indexer]: `persist_tx_insertion_order` may not preserve tx order from the checkpoint
During checkpoint indexing the `persist_tx_insertion_order` is called with transactions ordered the same way as they are ordered in a checkpoint with an intention that assigned insertion order will also follow the same ordering.

This however may not be the case, since implementation of `persist_tx_insertion_order` is splitting data in chunks, and is committing each chunk concurrently, which may lead to transaction order being randomly mixed.

This issue is to investigate this further, confirm that the issue exists, and fix it. (a naive solution would be to process the chunks sequentially, not concurrently)

Relevant piece of code: https://github.com/iotaledger/iota/blob/ee7dfe940ff05cc28fb23e2f927478ea73da0913/crates/iota-indexer/src/store/pg_indexer_store.rs#L2067

kodemartin:
# Description of change

This patch revisits the global ordering of optimistic and checkpointed transactions, by allowing ties in the global order number between transactions in a new table `tx_global_order`.

The table is similar to `tx_digests` with the addition of a monotonically increasing `optimistic_sequence_number` that will help us resolve deterministically the ties.

The logic requires that while executing optimistic transactions we assign `MAX(tx_sequence_number) FROM tx_digests` to as the global sequence number.

See also [comment in #6668](https://github.com/iotaledger/iota/issues/6668#issuecomment-2919267538) and the follow-up discussion.

This PR introduces the feature branch to target from now for the optimistic-indexing milestone.

Fixes #6668 

## Type of change

Manual synchronization benchmarks to verify successful insertions in the global order table

Further QA to be conducted for the feature branch at certain checkpoints.

### Release Notes

- [x] Indexer: Introduce `tx_global_order` table

[run-ci]


crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
```diff
@@ -0,0 +1,42 @@
+DROP TABLE IF EXISTS tx_insertion_order;
+
+-- It provides common ordering for optimistic and checkpointed transactions, whereas
+-- `tx_digests.tx_sequence_number` provides ordering only for checkpointed transactions.
+--
+-- The `sequence_number` in this table defaults to the checkpoint sequence number for
```
tomxey: Did you mean something like:
...defaults to the tx sequence number from the checkpoint...

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
```diff
@@ -0,0 +1,42 @@
...
+-- checkpointed transactions, and to the `SELECT MAX(tx_sequence_number) FROM
+-- tx_digests` at the time of insertion for optimistic transactions.
+--
+-- Deterministic global order is guaranteed by the composite index on
+-- `(insertion_order, optimistic_sequence_number)`, where
```
tomxey: Should be `(global_sequence_number, optimistic_sequence_number)` ?

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -667,12 +667,12 @@ impl IndexerReader {
         let optimistic_txs = run_query!(&self.pool, |conn| {
             optimistic_transactions::table
                 .inner_join(
-                    tx_insertion_order::table.on(optimistic_transactions::insertion_order
-                        .eq(tx_insertion_order::insertion_order)),
+                    tx_global_order::table.on(optimistic_transactions::sequence_number
+                        .eq(tx_global_order::optimistic_sequence_number)),
                 )
                 // we filter the tx_insertion_order table because it is indexed by digest,
```
tomxey: we filter the `tx_global_order` table

crates/iota-indexer/tests/ingestion_tests.rs
```diff
@@ -318,15 +331,20 @@ mod ingestion_tests {
...
-                .filter(tx_insertion_order::tx_digest.eq(digest.inner().to_vec()))
-                .select(tx_insertion_order::insertion_order)
-                .first::<i64>(conn)
+        let stored = read_only_blocking!(&pg_store.blocking_cp(), |conn| {
+            tx_global_order::table
+                .filter(tx_global_order::tx_digest.eq(digest.inner().to_vec()))
+                .select(TxGlobalOrder::as_select())
+                .first::<TxGlobalOrder>(conn)
         })
         .context("Failed reading tx insertion order from PostgresDB")?;
```
tomxey: Failed reading tx global order

crates/iota-indexer/src/models/transactions.rs
```diff
@@ -25,22 +25,39 @@ use serde::Deserialize;
...
+    /// guarantees deterministic ordering.
+    pub global_sequence_number: i64,
     pub tx_digest: Vec<u8>,
+    /// Monotonically increasing number that represents the order
+    /// of execution of optimistic transactions.
+    ///
+    /// To maintain these semantics the value should be `0` for
+    /// checkpointed transactions.
+    #[diesel(deserialize_as = i64)]
+    pub optimistic_sequence_number: Option<i64>,
```
tomxey, line 53: When read from DB it will always be Some(), it will only be set to None when writing optimistic tx to DB, to have the sequence number autoassigned, right?
Could we shortly mention it here in the doc?

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
```diff
@@ -0,0 +1,42 @@
...
+--
+-- In case of missing digests, the `tx_digests` table is used a fallback
+-- to resolve the transaction order. This is ok because optimistic transactions
+-- will be inserted only after creation of this table.
+CREATE TABLE tx_global_order (
+    tx_digest               BYTEA        PRIMARY KEY,
+    global_sequence_number  BIGINT       NOT NULL,
+    optimistic_sequence_number     BIGSERIAL
+);
+CREATE UNIQUE INDEX tx_global_order_seq_digest ON tx_global_order (global_sequence_number, optimistic_sequence_number);
```
tomxey, line 24: I feel that the most common use case would be to filter by tx_digest first, and then sort by this index.

I don't see this index being used in such scenario:
```
EXPLAIN ANALYZE
SELECT *
FROM tx_global_order
WHERE tx_digest IN ('\x1234', '\x5678')
ORDER BY global_sequence_number, optimistic_sequence_number;
```

It's only used if I remove the WHERE clause.

I'm not sure how much of an issue it is though.

What do you think?

tomxey (commented):
Looks good except for some details. Left a few comments.

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
kodemartin: b6499733ee

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
kodemartin: b6499733ee

crates/iota-indexer/src/indexer_reader.rs
kodemartin: b6499733ee

crates/iota-indexer/tests/ingestion_tests.rs
kodemartin: b6499733ee

crates/iota-indexer/src/models/transactions.rs
kodemartin: b6499733ee

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
kodemartin: This index is indeed not used in this case. But the objective here is different.

The primary design goal is to resolve transaction pages based using a `tx_digest` as a cursor, e.g. in the `IndexerApi`. That maps to a `(global_sequence_number, optimistic_sequence_number)` pair and thus the query that we want to optimize is

```sql
explain 
analyze select from tx_global_order 
where global_sequence_number >= 5 and optimistic_sequence_number > 3
order by global_sequence_number, optimistic_sequence_number limit 50;
```

tomxey (approved):

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
```diff
@@ -0,0 +1,43 @@
...
+-- number assigned in the checkpoint for checkpointed transactions, and to the
+-- `SELECT MAX(tx_sequence_number) FROM tx_digests` at the time of insertion
+-- for optimistic transactions.
+--
+-- Deterministic global order is guaranteed by the composite index on
+-- `(global_sequence_number, optimistic_sequence_number)`, where
+-- `optimistic_sequence_number` is the monotically increasing number
+-- that represents the order of execution for optimistic transactions.
+--
+-- In case of missing digests, the `tx_digests` table is used a fallback
```
samuel-rufi: small typo, should be "is used as a"

crates/iota-indexer/tests/ingestion_tests.rs
```diff
@@ -318,15 +331,20 @@ mod ingestion_tests {
...
+                .filter(tx_global_order::tx_digest.eq(digest.inner().to_vec()))
+                .select(TxGlobalOrder::as_select())
+                .first::<TxGlobalOrder>(conn)
         })
-        .context("Failed reading tx insertion order from PostgresDB")?;
+        .context("Failed reading `tx_global_order` from PostgresDB")?;
 
-        assert_eq!(actual_insertion_order, pre_existing_insertion_order);
+        assert_eq!(stored.global_sequence_number, global_sequence_number);
+        let expected_optimistic_sequence_number = 1;
```
samuel-rufi, line 343: The sequence number for optimistic transactions will be `SELECT MAX(tx_sequence_number) FROM tx_digests` at the time of insertion. So in that case it is 1. What I can't figure from the code changes, where does the optimistic sequence number field get set? Between
https://github.com/iotaledger/iota/pull/7152/files#diff-24c4275b8eee2f1692509dbc74e4f8aab2a378b8e83093f1b2237d47e5984f29R308-R313 we only populate the `tx_global_order` table with `optimistic_sequence_number: None`?

kodemartin: This patch handles the checkpoint transactions only. See the `committer.rs` changes.

For optimistic transactions we will just set `optimistic_sequence_number` to `None` and rely on the postgres sequence to generate it on the db server side.

samuel-rufi (approved):

crates/iota-indexer/tests/ingestion_tests.rs
samuel-rufi: Perfect thanks, then it is clear to me! 👌

crates/iota-indexer/migrations/pg/2025-05-29-072412_insertion-order-revisited/up.sql
kodemartin: 4086311632

PR #7162, opened 2025-05-30
fix(core): deserialize malformed types in tx (#19446)

### LINKED ISSUE #7081: [core,types] Deserialize malformed types in tx · MystenLabs/sui@6d8ceed
fix(core): deserialize malformed types in tx (#19446) · MystenLabs/sui@6d8ceed 

fix (protocol change)

- some changes are already merged (e.g., external crates), the rest should be implemented

TheMrAI:
# Description of change

Based on [changes](https://github.com/MystenLabs/sui/commit/6d8ceed4d727a6c9ce9f5879b3cd1b2e8605affa).
(For original description see the ticket.)

Most APIs, command line tools simply do not allow the user to construct syntactically incorrect transaction, but there are APIs which do. Namely ones in which users can send in an already prepared transaction as a byte sequence.

```
Usage: iota client execute-signed-tx [OPTIONS] --tx-bytes <TX_BYTES>
```

or the GraphQL API:

```
[executeTransactionBlock](https://graphql.testnet.iota.cafe/#)(
txBytes: [String](https://graphql.testnet.iota.cafe/#)!
signatures: [[String](https://graphql.testnet.iota.cafe/#)!]!
): [ExecutionResult](https://graphql.testnet.iota.cafe/#)!
```

or the Json RPC API:

```
iota_executeTransactionBlock
```

In terms of this change the cases in which an invalid **module** or **function** identifier manages to sneak through the checks the system would have deemed it a valid transaction, but it would have failed during execution. Because of this
it would have cost gas, forcing it the invalid transaction to be stored on-chain. This would cause downstream interfaces
to fail while trying to access the invalid transaction.

For example calling this is a correct call for the clock timestamp:
`cargo run -p iota -- client call --package 0x2 --module clock --function timestamp_ms --args 0x6`
While this would be incorrect:
`cargo run -p iota -- client call --package 0x2 --module _ --function timestamp_ms --args 0x6`

Now if you tried to execute the second function you would notice that the call fails as the tool properly protects the user. But if you manage to construct a byte sequence containing the same function call with an underscore as a module name, you could sign it and send it in (through the raw byte interfaces), and get the above described issue.

This change aims to mitigate these issues. Now the **module** and **function** identifiers are checked for being syntactically valid identifiers, so after the new protocol version, no more transactions can be logged on-chain.
Furthermore the GraphQL API has been modified so that it won't crash while trying to parse such an invalid entry.

We have added an e2e test for checking the modified behavior. 

fixes #7081 

## Type of change

Regenerate crates/iota-core/tests/staged/iota.yaml:
`cargo run -p iota-core --example generate-format -- print > crates/iota-core/tests/staged/iota.yaml`

Regenerate crates/iota-graphql-rpc/tests/snapshots/snapshot_tests__schema_sdl_export.snap
`cargo nextest run -p iota-graphql-rpc --features pg_integration --no-fail-fast --test-threads 1`

Regenerate/update iota-protocol-config snapshots:
`cargo nextest run -p iota-protocol-config`

Run iota-graphql-e2e-tests:
`cargo nextest run -p iota-graphql-e2e-tests --features pg_integration`

The new e2e tests:
`cargo nextest run -p iota-e2e-tests --test tx_bytes_validity_check`

### Release Notes

- [x] Protocol: Transactions that include invalid types will not be
signed by validators from protocol 10 onwards.
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL: `MoveType` can fail to provide a layout or abilities in
cases where it represents an invalid type.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-core/src/generate_format.rs
```diff
@@ -114,79 +140,105 @@ fn get_registry() -> Result<Registry> {
...
+    let si = StructInput {
+        address: AccountAddress::ZERO,
+        module: "foo".to_owned(),
+        name: "bar".to_owned(),
+        type_params: vec![TypeInput::Bool],
+    };
+    tracer.trace_value(&mut samples, &si).unwrap();
+
+    let event = iota_types::event::Event::random_for_testing();
+    tracer.trace_value(&mut samples, &event).unwrap();
```
TheMrAI, line 212: Event sample added as it contains an Identifier.

crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -560,6 +559,15 @@ impl Query {
             }))
     }
 
+    // TODO <- patch this after [aebab3c](https://github.com/MystenLabs/sui/commit/aebab3cf3218a8893417209fecc834c23619e780#diff-dc915efeed8b11d1f192223812017db449f824f2df0043dd95e207e75b0ba70e)
```
TheMrAI: Do we want such a comment to be included?
I have put in as a marker, but I don't like it.

crates/iota-package-resolver/src/lib.rs
```diff
@@ -1727,6 +1732,38 @@ fn ident(s: &str) -> Result<Identifier> {
     Identifier::new(s).map_err(|_| Error::NotAnIdentifier(s.to_string()))
 }
 
+pub fn as_type_tag(type_input: &TypeInput) -> Result<TypeTag> {
```
TheMrAI, line 1795: !!! This function is the same function as as_type_tag and into_type_tag from crates/iota-types/src/type_input.rs.
This triplication is very unnecessary and error prone. What would you prefer?
We keep one `as_type_tag(&self)` implemented in type_inputs.rs or the more Rustecean TryFrom (or TryInto if we don't have visibility).

I think it is prudent to stay with as_type_tag(&self) to keep future pulls a bit simpler. WDYT?

TheMrAI:
Please take care while reviewing this PR. I can't possible guarantee that no invariants got broken.

crates/iota-core/src/generate_format.rs
miker83z: Is this needed? I think the result would be the same if the Event was not explicitly put here.

crates/iota-package-resolver/src/lib.rs
miker83z: I vote for moving the logic into 1 new function and keeping the 3 function signatures as these are (pls, first check if maybe ths was done in a more recent upstream commit)

crates/iota-graphql-rpc/src/types/query.rs
miker83z: I guess it is fine for now!

crates/iota-core/src/generate_format.rs
TheMrAI: Unfortunately it is, otherwise the tracer will fail as it will not find a valid sample for the Identifier inside the Event.
It will try to serialize an empty string, upon which it will fail as that is forbidden within the Identifier.

crates/iota-types/src/type_input.rs
```diff
@@ -0,0 +1,378 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2024 IOTA Stiftung
```
jkrvivian:
```suggestion
// Modifications Copyright (c) 2025 IOTA Stiftung
```

jkrvivian (changes requested):

crates/iota-package-resolver/src/lib.rs
TheMrAI: Only managed to eliminate one of the triplets. This one would be too costly to fix.
For details see the Commits.

crates/iota-core/src/generate_format.rs
miker83z: Tested locally after the rebase, I still don't get why on Sui a sample for the Identifier is not needed here.

crates/iota-types/src/iota_sdk2_conversions.rs
```diff
@@ -1766,21 +1778,19 @@ impl TryFrom<Command> for crate::transaction::Command {
...
-                        .collect::<Result<_, _>>()?,
-                    arguments: move_call.arguments.into_iter().map(Into::into).collect(),
-                }))
-            }
+            Command::MoveCall(move_call) => Self::move_call(
+                move_call.package.into(),
+                crate::Identifier::new(move_call.module.as_str())
+                    .expect("invalid move call module identifier"),
+                crate::Identifier::new(move_call.function.as_str())
+                    .expect("invalid move call function identifier"),
```
Thoralf-M: Why expect here and not return an error?

jkrvivian (approved):

sdk/typescript/src/graphql/generated/2025.2/schema.graphql
miker83z, line 1: Please, remove these changes and create an issue for the tooling team to update this schema.

crates/iota-graphql-rpc/schema.graphql
```diff
@@ -343,7 +343,6 @@ String representation of an arbitrary width, possibly signed integer.
 """
 scalar BigInt
 
-
```
miker83z, line 346: Could you check if, when producing the schema locally, it contains these blank spaces or not?

crates/iota-core/src/generate_format.rs
miker83z: Could you point me to the part of the Event that is different in respect to upstream?

TheMrAI: You are right. I confused it with something else. Investigating.

TheMrAI: So for the time being I have checked the serde-reflection versions.
If I downgrade to serde-reflection 0.3.6 (same if I exactly copy the version pattern from sui and use 0.3.6 globally but for external crates 0.3.2) then we get error:

```
thread 'main' panicked at crates/iota-core/src/generate_format.rs:208:46:
called `Result::unwrap()` on an `Err` value: Custom("Failed to deserialize value: \"invalid value: integer `11`, expected variant index 0 <= i < 11\"") 
```
And it would point to the `tracer.trace_type::<TypeInput>(&samples).unwrap();` which is rather odd. Has 11 variants yes, why is it trying to serialize with the 11th?

If I upgrade to serde-reflection 0.5 then we get different error: 
```
thread 'main' panicked at crates/iota-core/src/generate_format.rs:318:35:
called `Result::unwrap()` on an `Err` value: MissingVariants(["CheckpointCommitment", "ConsensusDeterminedVersionAssignments", "GenesisObject", "TransactionData"])
```

But Sui is on version 0.3.6 and they don’t need Event sample and even if I update it to
0.4 it still passes. 
If I update their version to 0.5 they will error with:
```
thread 'main' panicked at crates/sui-core/src/generate_format.rs:248:35:
called `Result::unwrap()` on an `Err` value: MissingVariants(["CheckpointCommitment", "ConsensusDeterminedVersionAssignments", "GenesisObject", "TransactionData"])
```

As of now it is really unclear why this discrepancy occurs, nor is it trivially visible which type in the chain may cause an issue. My best guess is that perhaps serde-reflection isn't entirely stable and some subtle type differences may cause the problems.

crates/iota-graphql-rpc/schema.graphql
TheMrAI: It does not, I had to regenerate the schema and those spaces got deleted.

crates/iota-types/src/iota_sdk2_conversions.rs
TheMrAI: That behavior wasn't changed as per the feature pull. If at all possible we shouldn't change code when pulling changes or did you mean something else?

Thoralf-M: Ah it's the Command of the other SDK, then expect should be fine

sdk/typescript/src/graphql/generated/2025.2/schema.graphql
TheMrAI: Of course, but could you provide more context why?
I am not even sure why this change is here.

crates/iota-core/src/generate_format.rs
```diff
@@ -132,79 +150,105 @@ fn get_registry() -> Result<Registry> {
...
+
+    let si = StructInput {
+        address: AccountAddress::ZERO,
+        module: "foo".to_owned(),
+        name: "bar".to_owned(),
+        type_params: vec![TypeInput::Bool],
+    };
+    tracer.trace_value(&mut samples, &si).unwrap();
+
+    let event = iota_types::event::Event::random_for_testing();
```
TheMrAI: This Event sample shouldn't be necessary, but we must put it in otherwise serde-reflection will fail for us.
There is already an Event at the bottom of this file, we could move that up here or just leave a comment with this that it is here to workaround serde-reflection inconsistencies.

Now supplying another sample will not break anything, so the presence of this line shouldn't affect the outputs.

TheMrAI:
Please let me know what other tests should be executed.

crates/iota-graphql-rpc/src/types/move_type.rs
```diff
@@ -143,87 +143,98 @@ impl MoveType {
...
+        };
+
+        Ok(Some(MoveTypeLayout::try_from(layout).extend()?))
     }
 
-    /// The abilities this concrete type has.
-    async fn abilities(&self, ctx: &Context<'_>) -> Result<Vec<MoveAbility>> {
+    /// The abilities this concrete type has. Returns no abilities if the type
+    /// is invalid.
+    async fn abilities(&self, ctx: &Context<'_>) -> Result<Option<Vec<MoveAbility>>> {
```
kodemartin: You can return an empty `Vec` if the type is invalid. This way we preserve the existing guarantees of the API.

crates/iota-graphql-rpc/src/types/move_type.rs
```diff
@@ -143,87 +143,98 @@ impl MoveType {
     }
 
     /// Structured representation of the "shape" of values that match this type.
-    async fn layout(&self, ctx: &Context<'_>) -> Result<MoveTypeLayout> {
+    /// May return no layout if the type is invalid.
+    async fn layout(&self, ctx: &Context<'_>) -> Result<Option<MoveTypeLayout>> {
```
kodemartin: Instead of making this a nullable value for the API, I would suggest that you add a `MoveTypeLayout::InvalidType` variant and return that instead of `None`.

crates/iota-graphql-rpc/src/types/query.rs
kodemartin: I would rather suggest to create a tracking issue and remove the comment.

kodemartin (changes requested):
Left a couple of suggestion requiring minor revisions in `iota-graphql-rpc`. Other than that it looks good for the infra crates.

sdk/typescript/src/graphql/generated/2025.2/schema.graphql
TheMrAI: Never mind, got the context.

crates/iota-graphql-rpc/src/types/move_type.rs
```diff
@@ -143,87 +144,98 @@ impl MoveType {
...
+            ))
+        })?))
     }
 
     pub(crate) async fn abilities_impl(
         &self,
         resolver: &PackageResolver,
-    ) -> Result<AbilitySet, Error> {
-        resolver.abilities(self.native.clone()).await.map_err(|e| {
+    ) -> Result<Option<AbilitySet>, Error> {
```
kodemartin, line 201: Nit: If you maintain the signature here as well you might not need to refactor the calling function above at all.

crates/iota-graphql-rpc/src/types/move_type.rs
```diff
@@ -143,87 +144,98 @@ impl MoveType {
...
-        resolver
-            .type_layout(self.native.clone())
-            .await
-            .map_err(|e| {
-                Error::Internal(format!(
-                    "Error calculating layout for {}: {e}",
-                    self.native.to_canonical_display(/* with_prefix */ true),
-                ))
-            })
+    ) -> Result<Option<A::MoveTypeLayout>, Error> {
```
kodemartin, line 185: Nit: If you maintain the signature here as well you might not need to refactor the calling function above at all.

kodemartin (approved):
looks good @TheMrAI ! Left a couple of nitpicks that you may or may not consider further. Thank you!

piotrm50 (approved):

miker83z (approved):
LGTM!

crates/iota-core/src/generate_format.rs
```diff
@@ -56,6 +57,23 @@ use serde_reflection::{Registry, Result, Samples, Tracer, TracerConfig};
...
+/// The process is not automatic though, so all types that should
+/// be tracked must be presented to the [Tracer]. Whenever possible the
+/// [Tracer::trace_type] function should be used, but in cases when
+/// custom [serde::Deserialize] is implemented for a type with additional
+/// restrictions a [Tracer::trace_value] is likely necessary, so that [Tracer]
+/// may verify the type formats. This later requirement seems to be transitive.
+///
+/// For example **TypeA** implements a custom serializer, hence necessitating
+/// the use of [Tracer::trace_value], then every type that contains **TypeA**
+/// will require a sample to be provided.
```
valeriyr, line 76: Should we change the line:
```
// tracer.trace_value(&mut samples, ...)?;
```
to
```
// tracer.trace_value(&mut samples, ...).unwrap();
```
in the comment below?

crates/iota-graphql-rpc/src/types/move_type.rs
TheMrAI: I see your point, but am rather torn.

For one I try to stay as close to the feature pull as possible, even if I don't like it.
If I keep the original interface, then I can't really signal this new special case, because here it could only be an Error (in which case I would have to introduce a new type or match on its context above so that the empty vector could be returned).
Since, these aren't really that `pub` public functions, I could refactor abilities_impl, abilities function and the rest of the call sites to make it cleaner, but I feel that isn't in the scope of this PR.

The similar goes to the above layout you mentioned.

What do you think? I am not against fixing it, but probably would be best done in a follow up PR. This one is already a bit too complicated.

crates/iota-core/src/generate_format.rs
valeriyr: @TheMrAI @miker83z Looks like the problem is in the `GenesisTransaction` type, it doesn't contain events in Sui.
So, I think, the correct way to fix this issue should be to move the event tracing above this line:
```
tracer.trace_type::<TransactionKind>(&samples).unwrap();
```

valeriyr (approved):
The only point is that the extra event tracing can be removed. Otherwise, everything looks good to me.

crates/iota-core/src/generate_format.rs
TheMrAI: Agreed @valeriyr, moved the line.

crates/iota-core/src/generate_format.rs
TheMrAI: We can, but it will still be misleading, because the with the changes introduced in: https://github.com/iotaledger/iota/commit/ef3cd55a6427cdbf0c111336f4882cb94727f6d6
we already broke the separation that there is a trace values section followed by a trace types section.
This is how the code is in sui, and I have attempted to resolve the order for us, but it would introduce other tracing failures which didn't seem worth fixing.

Rather difficult to fix any problem with the current constraints.

TheMrAI: Not to mention that unwrapping is a downgrade in all honesty. When I was tracking down the issue it turned out that you can ask serde to provide some explanation for the failure which was admittedly a bit generic, but helpful non the less.
However, with this panicing unwrap, such niceties are lost.

crates/iota-core/src/generate_format.rs
miker83z: @TheMrAI we can close #7384, no?

PR #4128, opened 2024-11-19
Upgrade async-graphql deps

samuel-rufi:
# Description of change

Upgrade async-graphql deps

## Links to any relevant issues

Fixes https://github.com/iotaledger/iota/issues/4121

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How the change has been tested

```
cd crates/iota-graphql-e2e-tests
cargo nextest run --features pg_integration
```

```
cd crates/iota-graphql-rpc
cargo nextest run -p iota-graphql-rpc --features pg_integration --no-fail-fast --test-threads 1
```

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes

PR #1969, opened 2024-08-21
fix(iota-graphql-e2e-tests): restore graphql e2e tests

kodemartin:
# Description of change

This patch does the following:

* Fixes a bug introduced by the upgrade of `async_graphql` in #1106 (see cde7061556)
* Updates the baseline for the e2e tests, which among other things removes the query on the subsidies.
* Removes obsolete name service compatibility tests from `@iota/graphql-transport` 

@valeriyr, @miker83z, @pavlo these tests use the `iota-transactional-test-runner` and so the baseline updates mostly relate to the updated genesis, and different ordering of the results. I would appreciate if one of you could have a quick look to verify that the responses are plausible. To me it seems so.

## Links to any relevant issues

Fixes #1580

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

## How the change has been tested

Follow the instructions in the updated README file to run the tests locally.

## Change checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have checked that new and existing unit tests pass locally with my changes

PR #7180, opened 2025-05-30
refactor(test_adapter): refactors the query interpolation part of transactional test adapter

### LINKED ISSUE #7059: [test-adapter] Refactor query interpolation and change format for bcs · MystenLabs/sui@ae367c4
[test-adapter] refactor query interpolation and change format for bcs… · MystenLabs/sui@ae367c4 

refactor

- to check if feasible first

### LINKED ISSUE #7080: [transactional-test] Update tests · MystenLabs/sui@94a0944
[exp-to-insta][3/n] Move compiler transactional tests (#21044) · MystenLabs/sui@94a0944 

feat

- the external crates changes are already merged, the rest should be implemented

Dkwcs:
[run-ci]
# Description of change

According to [changes](https://github.com/MystenLabs/sui/commit/ae367c4#diff-7a132b89ef988a881b2c5d33b4cee85acb20e6be9c7fc645d26806d7ae307290).
I as see we don’t have iota-indexer-alt-e2e tests and iota-mvr-graphql-e2e, so changes only touch iota-graphql-e2e- tests.

Only these changes differ from Sui’s commit.
I assume it’s because of the version differences in fastcrypto or async_graphql.
To comply with how iota-graphql-rpc decodes the JSON cursor … this traces back to async_graphql = 7.0.7, which uses no padding for encoding/decoding.

`Ok(Base64::encode(cursor));
`

This likely defaults to:

 - STANDARD base64

 - with padding (=)

 - and possibly using the general character set (+ / /).

But the decoder expects:

- URL_SAFE_NO_PAD

- no padding, using - and _

- matching what async_graphql and the RPC layer parse.

```
return Ok(base64::Engine::encode(
        &base64::engine::general_purpose::URL_SAFE_NO_PAD,
        cursor,
));
```

Fixes #7059  .

## Type of change

For iota-graphql-e2e- tests run:

`cargo nextest run --features pg_integration`


crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1348,10 +1298,65 @@ impl IotaTestAdapter {
...
+    }
+
+    fn encode_cursor(&self, cursor: &str) -> anyhow::Result<String> {
+        // Cursor format is either bcs(object_id,n1,n2,...) or a json value,
+        // in which case we just return its base64 encoding.
+        let Some(args) = cursor
+            .strip_prefix("bcs(")
+            .and_then(|c| c.strip_suffix(")"))
+        else {
+            return Ok(base64::Engine::encode(
```
TheMrAI, line 1319: You meant this is the different section?
I checked the version of `base64ct` that we use here and mysten uses in the linked commit and if I didn't get confused:
(we) https://github.com/MystenLabs/fastcrypto/blob/2f502fd8570fe4e9cff36eea5bbd6fef22002898/fastcrypto/Cargo.toml#L13
(them)
https://github.com/MystenLabs/fastcrypto/blob/69d496c71fb37e3d22fe85e5bbfd4256d61422b9/fastcrypto/Cargo.toml#L13
both of which result in a `base64ct = { version = "1.5.3", features = ["alloc"] }`.

Quickly ran the tests after reverting this to `return Ok(Base64::encode(cursor));` and noticed from all the tests that fail none were expected to change according to the linked changelist.
So then I took a faulty test and ran it like so:
```
RUST_BACKTRACE=1 cargo nextest run -p iota-graphql-e2e-tests --nocapture --features pg_integration scan_limit/le_page/last
```
The backtrace helped me none, but the nocapture prints out all the stdout used in the tests. Made a diff on them:
image
(left using `Ok(Base64::encode(cursor))`)

So yeah it does mess up the padding, but why? Maybe we miss a graphql change perhaps or I mixed up the versions?

Dkwcs: That’s what I was talking about — the code is identical, but something (probably the version of crate) is different

TheMrAI: Okay, the description wasn't clear to me. 
But isn't the relevant crate version (base64) the same?

Didn't something change on the graphql parser side for us?

Did some test on the SUI side for their commit ae367c4b88fc7f3b1d18591e3d77f3395d33f6d8 and printed a Debug GraphQL query from the test.
We make:
`GraphQL query: "# cursors are null, and page flags are both false\n{\n  transactionBlocks(last: 3 before: \"eyJjIjo0LCJ0IjoyLCJpIjp0cnVlfQ==\" scanLimit: 2 filter: {recvAddress: \"0x8cca4e1ce0ba5904cea61df9242da2f7d29e3ef328fb7ec07c086b3bf47ca61a\" afterCheckpoint: 1 beforeCheckpoint: 4}) {\n    pageInfo {\n      hasPreviousPage\n      hasNextPage\n      startCursor\n      endCursor\n    }\n    edges {\n      cursor\n      node {\n        digest\n        effects {\n          checkpoint {\n            sequenceNumber\n          }\n        }\n      }\n    }\n  }\n}"`
They make:
`GraphQL query: "# cursors are null, and page flags are both false\n{\n  transactionBlocks(last: 3 before: \"eyJjIjo0LCJ0IjoyLCJpIjp0cnVlfQ==\" scanLimit: 2 filter: {affectedAddress: \"0xfccc9a421bbb13c1a66a1aa98f0ad75029ede94857779c6915b44f94068b921e\" afterCheckpoint: 1 beforeCheckpoint: 4}) {\n    pageInfo {\n      hasPreviousPage\n      hasNextPage\n      startCursor\n      endCursor\n    }\n    edges {\n      cursor\n      node {\n        digest\n        effects {\n          checkpoint {\n            sequenceNumber\n          }\n        }\n      }\n    }\n  }\n}"`

They also produce the == as we would, so I don't think `Ok(Base64::encode(cursor))` is the issue.

Dkwcs: About obvious things I only have noticed a difference in async-graphql version. I didn’t go further with an investigation.

But anyway we have other behavior competing to Sui's

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1277,69 +1271,25 @@ impl IotaTestAdapter {
...
-                use base64::Engine;
-
-                // To comply with how `iota-graphql-rpc` decodes the json cursor
-                // (see `iota_graphql_rpc::types::cursor::JsonCursor`).
-                //
-                // This traces back to `async_graphql = 7.0.7` that uses no padding for
-                // encoding/decoding.
-                let base64d = base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(s);
-
-                variables.insert(format!("cursor_{idx}"), base64d);
```
miker83z, line 1314: @Dkwcs continuing from the [thread above](https://github.com/iotaledger/iota/pull/7180#discussion_r2122104309). This part was already different from upstream. @samuel-rufi introduced it in #4128, so maybe he has more information to share on why this is necessary.

crates/iota-transactional-test-runner/src/test_adapter.rs
miker83z: Continues in https://github.com/iotaledger/iota/pull/7180#discussion_r2126228904

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Great point, thanks

samuel-rufi: Hello @miker83z @Dkwcs , `async-graphql’s` PR https://github.com/async-graphql/async-graphql/pull/1466/files upgraded to base64 0.21, which switched cursor encoding to URL-safe Base64 without padding (URL_SAFE_NO_PAD). Since JsonCursor in async-graphql 7.x only accepts that no-pad format, we must encode our cursors the same way; otherwise, GraphQL cursor decoding will fail.
The same patch was applied in https://github.com/iotaledger/iota/pull/1969/files#diff-b6d7fb8244703e36e6c090583b0323b30b90ee3b94744dd6c5aa7236a70f29c7R1206 by @kodemartin, maybe also this reference is helpful.

miker83z: Thanks! @Dkwcs could you please then copy the comment you removed here into the encoding part in `encode_cursor`?

Dkwcs: @miker83z 
Do you mean this one? Okay
```
                // To comply with how `iota-graphql-rpc` decodes the json cursor
                // (see `iota_graphql_rpc::types::cursor::JsonCursor`).
                //
                // This traces back to `async_graphql = 7.0.7` that uses no padding for
                // encoding/decoding.
```

Dkwcs: Done

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1348,10 +1298,70 @@ impl IotaTestAdapter {
...
+            .trim()
+            .parse()?;
+
+        let mut bytes = bcs::to_bytes(&id.to_vec())?;
+        for part in parts {
+            let n: u64 = part.trim().parse()?;
+            bytes.extend(bcs::to_bytes(&n)?);
+        }
+
+        Ok(Base64::encode(bytes))
```
TheMrAI, line 1339: Please confirm if we can use this call or the
`Ok(base64::Engine::encode(&base64::engine::general_purpose::URL_SAFE_NO_PAD,
                cursor,
            ));
        };`
     one.
     Apart from that it looks good to me.

TheMrAI (changes requested):

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Yeah, it fails with the error - `"Failed to parse \"String\": Invalid value was given to the function"`

TheMrAI (approved):

PR #7181, opened 2025-05-30
feat(starfish) Track who knows which block headers

### LINKED ISSUE #6879: Track who knows what for recent DAG (with traversing this recent DAG)

polinikita:
# Description of change

We add two new block structures to the DAG state. One stores the DAG structure of the block headers (only BlockRefs) and who knows the corresponding block headers. Another stores explicitly which authority does not know which block headers. The latter will be used in streaming for the pushing strategy. The former allows to efficiently update the latter.

We implement the update with traversing the DAG when a new block header is added and the eviction when flush() is called.

Fixes #6879 

## Type of change

CI

## Change checklist

- [ ] I have followed the contribution guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/dag_state.rs
+++ b/crates/starfish/core/src/dag_state.rs
@@ -4,7 +4,7 @@
 
 use std::{
     cmp::max,
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashSet, VecDeque},
     ops::Bound::{Excluded, Included, Unbounded},
     panic,
     sync::Arc,
@@ -103,6 +103,20 @@ pub(crate) struct DagState {
     // availability of transaction data from the corresponding blocks
     pending_acknowledgments: BTreeSet<BlockRef>,
 
+    // Keeps track of the most recent BlockHeaderDAG cordial knowledge (who knows which blocks)
+    // for each authority. This is a helper structure that is used primarily for traversing the
+    // recent DAG. This struct is evicted after flushing the dag state to storage and is not
+    // persisted.
+    // To access the cordial knowledge of a given block_ref, one shall retrieve it from
+    // `recent_dag_cordial_knowledge[block_ref.author][(block_ref.round, block_ref.digest)]`.
+    // The value is a tuple of (parents, who knows the block header).
+    recent_dag_cordial_knowledge:
+        Vec<BTreeMap<(Round, BlockHeaderDigest), (Vec<BlockRef>, HashSet<AuthorityIndex>)>>,
+    // Keeps tracks of block headers that are not known by the authority.
+    // Is used to ensure that we send block headers that are really needed
+    // to the authority, and not the ones that they already know.
+    block_headers_not_known_by_authority: Vec<BTreeSet<BlockRef>>,
+
     // Data to be flushed to storage.
     blocks_to_write: Vec<VerifiedBlock>,
     block_headers_to_write: Vec<VerifiedBlockHeader>,
@@ -195,6 +209,8 @@ impl DagState {
             commits_to_write: vec![],
             commit_info_to_write: vec![],
             pending_acknowledgments: BTreeSet::new(),
+            recent_dag_cordial_knowledge: vec![BTreeMap::new(); num_authorities],
+            block_headers_not_known_by_authority: vec![BTreeSet::new(); num_authorities],
             scoring_subdag,
             store: store.clone(),
             cached_rounds,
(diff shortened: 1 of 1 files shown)


crates/starfish/core/src/dag_state.rs
```diff
@@ -324,6 +337,85 @@ impl DagState {
...
+    fn update_cordial_knowledge(&mut self, block_header: &VerifiedBlockHeader) {
+        let block_reference = block_header.reference();
+        let block_author = block_reference.author;
+        let round_digest = (block_reference.round, block_reference.digest);
+
+        // Collect parents of the block header, which are not genesis
+        let parents = block_header
+            .ancestors()
+            .iter()
+            .filter(|parent| parent.round > 0)
```
VorobyevIlya: We usually use GENESIS_ROUND constant

crates/starfish/core/src/dag_state.rs
```diff
@@ -101,6 +101,17 @@ pub(crate) struct DagState {
     // availability of transaction data from the corresponding blocks
     pending_acknowledgments: BTreeSet<BlockRef>,
 
+    // Keeps track of the most recent BlockHeaderDAG cordial knowledge (who knows which blocks)
+    // for each authority. This is a helper structure that is used primarily for traversing the
+    // recent DAG. This struct is evicted after flushing the dag state to storage and is not
+    // persisted.
+    recent_dag_cordial_knowledge:
```
VorobyevIlya, line 113: It is not clear from the description how the information is stored in this structure

crates/starfish/core/src/dag_state.rs
```diff
@@ -938,6 +1030,56 @@ impl DagState {
...
+        let set = &self.block_headers_not_known_by_authority[authority_index.value()];
+
+        // Construct an upper bound (exclusive)
+        let upper_bound = BlockRef::new(round, AuthorityIndex::MAX, BlockHeaderDigest::MAX);
+
+        // Take all BlockRefs strictly less than the upper bound
+        set.range(..upper_bound).cloned().collect()
+    }
+
+    /// Updates the set of known block headers for a given authority. Make use
```
VorobyevIlya: From the code it looks like pruning everything from <= target_round. Description and the name doesn't correspond to the code

crates/starfish/core/src/dag_state.rs
```diff
@@ -938,6 +1030,56 @@ impl DagState {
...
+    #[expect(unused)]
+    pub(crate) fn update_known_block_headers_for_authority_by_round(
+        &mut self,
+        authority_index: AuthorityIndex,
+        target_round: Round,
+    ) {
+        // Construct an exclusive upper bound
+        let upper_bound = BlockRef::new(target_round, AuthorityIndex::MAX, BlockHeaderDigest::MAX);
+
+        // Split off entries strictly greater than upper_bound
```
VorobyevIlya: it is >=, not strictly >

crates/starfish/core/src/dag_state.rs
```diff
@@ -960,6 +1102,39 @@ impl DagState {
...
+    /// Evicts old cordial knowledge and pending acknowledgments. It is aligned
+    /// with the eviction method, thereby should be called every time the
+    /// eviction happens.
+    pub(crate) fn evict_cordial_knowledge(&mut self) {
+        // === 1. Evict from recent_dag_cordial_knowledge ===
+        for (authority_index, map) in self.recent_dag_cordial_knowledge.iter_mut().enumerate() {
+            let evict_round = self.evicted_rounds[authority_index];
+
+            // Only keep entries with round > evict_round
+            let keep_from = (evict_round + 1, BlockHeaderDigest::MIN);
```
VorobyevIlya, line 1123: Should we just use update_known_block_headers_for_authority_by_round function here? Looks like code duplication. Or is there a difference?

crates/starfish/core/src/dag_state.rs
polinikita: Good catch

VorobyevIlya (approved):

crates/starfish/core/src/dag_state.rs
```diff
@@ -324,6 +340,85 @@ impl DagState {
             .highest_accepted_authority_round
             .with_label_values(&[hostname])
             .set(highest_accepted_round_for_author as i64);
+        self.update_cordial_knowledge(block_header);
+    }
+
+    // Function updates who knows which BlockHeaders what after receiving a new
+    // block header, which is accepted. In particular, it traverses the DAG from
```
piotrm50, line 349: > after receiving a new block header, which is accepted

Wouldn't it make more sense to update cordial knowledge upon receiving the block in BlockManager? After the problems we recently saw with suspended blocks only updating it upon acceptance might make us constantly send the same blocks to Authorities creating those suspended blocks. Could it cause any other problems though? Not sure how potential gaps would be handled.

crates/starfish/core/src/dag_state.rs
```diff
@@ -324,6 +340,85 @@ impl DagState {
...
+                let traversed_parent_author = parent.author;
+                let traversed_parent_round_digest = (parent.round, parent.digest);
+
+                if self.recent_dag_cordial_knowledge[traversed_parent_author.value()]
+                    .contains_key(&traversed_parent_round_digest)
+                {
+                    let (_, who_knows_given_parent) = self.recent_dag_cordial_knowledge
+                        [traversed_parent_author.value()]
+                    .get_mut(&traversed_parent_round_digest)
+                    .expect("We should have this value as it is checked");
```
piotrm50, line 415: Using `match` would avoid `expect` here.

piotrm50 (approved):
Are we going to mark blocks that we disseminate to an authority as "known" by that authority to avoid sending the same blocks over and over in case the authority does not create any new blocks, or if those blocks are suspended?

We also need tests for this logic, but that probably needs to be implemented after we get rid of the current compilation errors.

crates/starfish/core/src/dag_state.rs
polinikita: You are right, but if "expect" panics, then something is wrong with the underlying logic in this function. I would fix the logic here in the first place, rather than making it recoverable with unexpected behavior.

crates/starfish/core/src/dag_state.rs
polinikita: No. Upon sending (in stream) a block header, we update our cordial knowledge immediately and will not resend the same block header to the same peer in our streaming.

Updating the knowledge in blockmanager is dangerous in my opinion, while it is possible to some extent. If a block_header is still suspended, it is not a part of cordial_knowledge_dag, and we will not be able to fully traverse the dag starting from such a vertex - also, this might require traversing too many times for one block header and might lead to unexpected results.

PR #7188, opened 2025-06-02
feat(ci): allow defining the feature flags per binary

alexsporn:
# Description of change

This change allows us to define the feature flags to be used when building each release binary.
This also removes the bridge binaries from the released binaries

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,14 +161,8 @@ jobs:
         run: |
           [ -f ~/.cargo/env ] && source ~/.cargo/env
 
-          binaries=$(jq -r 'to_entries[] | select(.key=="release_binaries") | .value | to_entries[] | @base64' ${{ env.BINARY_LIST_FILE }})
-          for entry in $binaries; do
-            _jq() {
-              echo "${entry}" | base64 --decode | jq -r "$1"
-            }
-            binary=$(_jq '.key')
-            features=$(_jq '.value | join(",")')
-
+          jq -r '.release_binaries | to_entries[] | [.key, (.value | join(","))] | @tsv' ${{ env.BINARY_LIST_FILE }} | while IFS=$'\t' read -r binary features; do
+            echo ">>> Building $binary with features: $features"
             if [[ -z "$features" ]]; then
               cargo build --release --bin "$binary"
             else
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -158,9 +158,23 @@ jobs:
 
       - name: Cargo build for ${{ matrix.os_type }} platform
         shell: bash
-        # Currently building in release mode, but we could also have debug builds for testing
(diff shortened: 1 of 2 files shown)


lzpap (approved):

.github/workflows/release.yml
```diff
@@ -158,9 +158,23 @@ jobs:
 
       - name: Cargo build for ${{ matrix.os_type }} platform
         shell: bash
-        # Currently building in release mode, but we could also have debug builds for testing
         run: |
-          [ -f ~/.cargo/env ] && source ~/.cargo/env ; cargo build --release --features indexer,gen-completions,tracing
+          [ -f ~/.cargo/env ] && source ~/.cargo/env
+
+          binaries=$(jq -r 'to_entries[] | select(.key=="release_binaries") | .value | to_entries[] | @base64' ${{ env.BINARY_LIST_FILE }})
```
TheMrAI: Is `jq` installed in the release workflow?

TheMrAI (approved):
LGTM, but can't really comment on the release flag section.

.github/workflows/release.yml
```diff
@@ -158,9 +158,23 @@ jobs:
...
+          for entry in $binaries; do
+            _jq() {
+              echo "${entry}" | base64 --decode | jq -r "$1"
```
Thoralf-M: Is the base64 encoding and decoding needed?

alexsporn: That would allow us to specify more complex objects per entry, but I refactored it out for now and instead use a tab-based separation.

Thoralf-M (approved):

.github/workflows/release.yml
```diff
@@ -158,9 +158,17 @@ jobs:
 
       - name: Cargo build for ${{ matrix.os_type }} platform
         shell: bash
-        # Currently building in release mode, but we could also have debug builds for testing
         run: |
-          [ -f ~/.cargo/env ] && source ~/.cargo/env ; cargo build --release --features indexer,gen-completions,tracing
+          [ -f ~/.cargo/env ] && source ~/.cargo/env
+
+          jq -r '.release_binaries | to_entries[] | [.key, (.value | join(","))] | @tsv' ${{ env.BINARY_LIST_FILE }} | while IFS=$'\t' read -r binary features; do
```
thibault-martinez, line 164: What's this `@tsv` thing?

thibault-martinez (approved):
Assuming you have tested it locally

.github/workflows/release.yml
alexsporn: tab-separated values as explained in the resolved comment above :-P

alexsporn: it does a:
`binaryname\tfeature1,feature2,feature3` per entry so that bash can read each entry and separate the components by the `\t`

PR #7225, opened 2025-06-04
feat(starfish): separate headers transaction storage

### LINKED ISSUE #6436: Update Storage for Header/Data Separation
This logic should be implemented in the DagState component as well as the store.
- [x] Store block headers separately (`(Round, Author, Digest) -> BlockHeader`) (Done as part of #6707)
- [ ] Store block data separately (`(Round, Author, Digest) -> / Data`)
- [ ] Trigger storing when data is available

piotrm50:
# Description of change

This PR updates the interfaces of Store and DAGState to support separate BlockHeader and Transaction handling needed for Starfish implementation. 

Resolves #6436 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/block_manager.rs
+++ b/crates/starfish/core/src/block_manager.rs
@@ -115,7 +115,7 @@ impl BlockManager {
         for block_header in accepted_block_headers.iter() {
             if let Some(block) = self.suspended_blocks.remove(&block_header.reference()) {
                 // for this accepted header we already have a block, so we add it to dag_state
-                self.dag_state.write().add_block(block, live);
+                self.dag_state.write().add_transactions(block, live);
             } else {
                 accepted_block_header_refs.insert(block_header.reference());
             }
@@ -123,7 +123,7 @@ impl BlockManager {
         for block in blocks {
             let block_ref = &block.reference();
             if accepted_block_header_refs.remove(block_ref) {
-                self.dag_state.write().add_block(block, live);
+                self.dag_state.write().add_transactions(block, live);
             } else {
                 self.suspended_blocks.insert(block.reference(), block);
             }
--- a/crates/starfish/core/src/commit.rs
+++ b/crates/starfish/core/src/commit.rs
@@ -877,7 +877,7 @@ mod tests {
             .map(|block| (block.reference(), block))
             .unzip();
         store
-            .write(WriteBatch::default().blocks(first_round_headers))
+            .write(WriteBatch::default().transactions(first_round_headers))
             .unwrap();
         blocks.append(&mut first_round_references.clone());
         // TODO: create some data for blocks in the first round
@@ -896,7 +896,7 @@ mod tests {
                         .build(),
                 );
                 store
-                    .write(WriteBatch::default().blocks(vec![block.clone()]))
+                    .write(WriteBatch::default().transactions(vec![block.clone()]))
                     .unwrap();
                 new_ancestors.push(block.reference());
                 blocks.push(block.reference());
--- a/crates/starfish/core/src/dag_state.rs
+++ b/crates/starfish/core/src/dag_state.rs
@@ -55,8 +55,8 @@ pub(crate) struct DagState {
     // authority. Note: all uncommitted block headers are kept in memory.
     recent_block_headers: BTreeMap<BlockRef, VerifiedBlockHeader>,
 
-    // Contains recent blocks (together with transactions) within CACHED_ROUNDS from the last
-    // committed round per authority. Note: all uncommitted blocks are kept in memory.
+    // Contains recent transactions within CACHED_ROUNDS from the last
+    // committed round per authority. Note: all uncommitted transactions are kept in memory.
     recent_transactions: BTreeMap<BlockRef, VerifiedTransactions>,
 
     // Indexes recent block headers refs by their authorities.
--- a/crates/starfish/core/src/dag_state.rs
+++ b/crates/starfish/core/src/dag_state.rs
@@ -19,7 +19,7 @@ use tracing::{debug, error, info};
 use crate::{
     block_header::{
         BlockHeaderAPI, BlockHeaderDigest, BlockRef, BlockTimestampMs, GENESIS_ROUND, Round, Slot,
-        VerifiedBlock, VerifiedBlockHeader, genesis_blocks,
+        VerifiedBlock, VerifiedBlockHeader, VerifiedTransactions, genesis_blocks,
     },
     commit::{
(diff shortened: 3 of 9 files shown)


crates/starfish/core/src/commit.rs
```diff
@@ -877,7 +877,7 @@ mod tests {
             .map(|block| (block.reference(), block))
             .unzip();
         store
-            .write(WriteBatch::default().blocks(first_round_headers))
+            .write(WriteBatch::default().transactions(first_round_headers))
```
polinikita, line 880: rename first_round_headers to first_round_blocks

crates/starfish/core/src/dag_state.rs
```diff
@@ -54,19 +54,12 @@ pub(crate) struct DagState {
 
     // Contains recent blocks (together with transactions) within CACHED_ROUNDS from the last
```
polinikita: Change the description to align with transactions

crates/starfish/core/src/block_manager.rs
```diff
@@ -114,15 +114,15 @@ impl BlockManager {
         for block_header in accepted_block_headers.iter() {
             if let Some(block) = self.suspended_blocks.remove(&block_header.reference()) {
                 // for this accepted header we already have a block, so we add it to dag_state
-                self.dag_state.write().add_block(block);
+                self.dag_state.write().add_transactions(block);
```
polinikita: Should verified_transactions from this block be as input here?

crates/starfish/core/src/block_manager.rs
```diff
@@ -114,15 +114,15 @@ impl BlockManager {
...
             } else {
                 accepted_block_header_refs.insert(block_header.reference());
             }
         }
         for block in blocks {
             let block_ref = &block.reference();
             if accepted_block_header_refs.remove(block_ref) {
-                self.dag_state.write().add_block(block);
+                self.dag_state.write().add_transactions(block);
```
polinikita: Should verified_transactions from this block be as input here?

crates/starfish/core/src/dag_state.rs
```diff
@@ -537,12 +560,16 @@ impl DagState {
     /// Gets the last proposed block from this authority.
     /// If no block is proposed yet, returns Genesis block.
     pub(crate) fn get_last_proposed_block(&self) -> VerifiedBlock {
```
polinikita, line 666: Is this function used somewhere in the context of block not block_header?

crates/starfish/core/src/dag_state.rs
```diff
@@ -537,12 +560,16 @@ impl DagState {
...
+        if let Some(last) = self.recent_headers_refs_by_authority[self.context.own_index].last() {
+            let header = self
+                .recent_block_headers
                 .get(last)
-                .expect("Block should be found in recent blocks")
-                .clone();
+                .expect("Block header should exist for the most recent blocks");
+            let transactions = self
+                .recent_transactions
+                .get(last)
```
polinikita, line 674: I am not sure that this is correct. For most recent block header, transactions might be not present in the dag_state.
I think we use this function only for metadata or for requesting last_block_header request.

crates/starfish/core/src/storage/mem_store.rs
```diff
@@ -111,48 +99,54 @@ impl Store for MemStore {
...
                     .cloned()
             })
             .collect();
-        Ok(block_headers)
+        Ok(transactions)
     }
 
-    fn contains_blocks(&self, refs: &[BlockRef]) -> ConsensusResult<Vec<bool>> {
+    // TODO: Do we need this method or will DAGState always try to read both headers
+    // and transactions seaprately?
```
polinikita: typo in separetely

crates/starfish/core/src/storage/rocksdb_store.rs
```diff
@@ -209,6 +202,9 @@ impl Store for RocksDBStore {
             .iter()
```
polinikita, line 202: When do we need to read_blocks()? Is it guruanteed that both block_headers and associated transaction data are present in the DB? If it is used, e.g., for commit_syncer, then the logic might not be correct, as some data may be unavailable for some headers.

polinikita (changes requested):
Left a few comments to resolve some inconsistencies and typos.

VorobyevIlya (approved):

crates/starfish/core/src/dag_state.rs
VorobyevIlya: But this is about our own block? Can transactions be not present for our block?

crates/starfish/core/src/storage/rocksdb_store.rs
VorobyevIlya: It is not guaranteed but if transaction data is missing then for such block function returns None. I think it is fine

crates/starfish/core/src/dag_state.rs
piotrm50: this is inside an `if let Some(last) = self.recent_headers_refs_by_authority[self.context.own_index].last()`

crates/starfish/core/src/storage/rocksdb_store.rs
piotrm50: We will remove `read_blocks` from Store as DAG State is reconstructing blocks from headers and transactions.

crates/starfish/core/src/block_manager.rs
piotrm50: This will be resolved in a subsequent PR

crates/starfish/core/src/block_manager.rs
piotrm50: This will be resolved in a subsequent PR

crates/starfish/core/src/dag_state.rs
piotrm50: Yes, in `recover` function of `core`.

crates/starfish/core/src/commit.rs
piotrm50: This will be resolved in a subsequent PR

crates/starfish/core/src/storage/rocksdb_store.rs
piotrm50: Yes, it can happen - it's used in the `scan_blocks_by_author` method which is currently not used anywhere. This might be cleaned up. However about your point - I think it's ok that if we don't have transactions for a header that we return None and the caller should be aware of it. Most likely we will remove those methods in the end as they seem to not be necessary, but let's wait and see.

PR #7235, opened 2025-06-05
feat(indexer): Introduce basic optimistic indexing

### LINKED ISSUE #7212: [indexer]: Introduce basic optimistic indexing logic
This issue is to introduce basic optimistic indexing logic to the feature branch, so that follow up PRs can improve upon this logic (potentially allowing the work to be parallelized).

The basic implementation doesn't have to solve all known issues (race conditions, dependency checks, pruning, etc.), those will be solved later on the feature branch.

tomxey:
# Description of change

Introduce basic optimistic indexing logic.

This basic implementation doesn't solve all known issues (race conditions, dependency checks, pruning, etc.), those will be solved later on the feature branch.

fixes https://github.com/iotaledger/iota/issues/7212

- [x] Patch-specific tests (correctness, functionality coverage) - **Adapted tests to see if basic optimistic indexing logic is working**

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/apis/write_api.rs
+++ b/crates/iota-indexer/src/apis/write_api.rs
@@ -133,19 +133,24 @@ impl OptimisticWriteApi {
         } = response;
         let tx_digest = *effects.transaction_digest();
 
-        if let (Some(input_objects), Some(output_objects)) = (input_objects, output_objects) {
-            let full_tx_data = CheckpointTransaction {
-                transaction,
-                effects,
-                events,
-                input_objects,
-                output_objects,
-            };
-            self.index_transaction(&full_tx_data).await?;
-        } else {
-            tracing::warn!(
-                "Cannot optimistically index because of missing in/out objs for tx: {tx_digest}"
-            );
+        match (input_objects, output_objects) {
+            (Some(input_objects), Some(output_objects))
+                if !input_objects.is_empty() && !output_objects.is_empty() =>
+            {
+                let full_tx_data = CheckpointTransaction {
+                    transaction,
+                    effects,
+                    events,
+                    input_objects,
+                    output_objects,
+                };
+                self.index_transaction(&full_tx_data).await?;
+            }
+            _ => {
+                tracing::warn!(
+                    "Cannot optimistically index because of missing in/out objs for tx: {tx_digest}"
+                );
+            }
         }
 
         let tx_block_response = self
--- a/crates/iota-indexer/src/apis/write_api.rs
+++ b/crates/iota-indexer/src/apis/write_api.rs
@@ -53,6 +53,10 @@ use crate::{
 
 pub(crate) struct WriteApi {
     fullnode: HttpClient,
+}
+
+pub(crate) struct OptimisticWriteApi {
+    write_api: WriteApi,
     fullnode_rest_client: iota_rest_api::Client,
     inner: IndexerReader,
     store: PgIndexerStore,
@@ -69,15 +73,23 @@ type TransactionDataToCommit = (
 );
 
 impl WriteApi {
+    pub fn new(fullnode_client: HttpClient) -> Self {
+        Self {
+            fullnode: fullnode_client,
+        }
+    }
+}
+
+impl OptimisticWriteApi {
     pub fn new(
-        fullnode_client: HttpClient,
+        write_api: WriteApi,
         fullnode_rest_client: iota_rest_api::Client,
         inner: IndexerReader,
         store: PgIndexerStore,
         metrics: IndexerMetrics,
     ) -> Self {
         Self {
-            fullnode: fullnode_client,
+            write_api,
             fullnode_rest_client,
             inner,
             store,
@@ -380,23 +392,11 @@ impl WriteApiServer for WriteApi {
         options: Option<IotaTransactionBlockResponseOptions>,
         request_type: Option<ExecuteTransactionRequestType>,
     ) -> RpcResult<IotaTransactionBlockResponse> {
-        let iota_transaction_response = match request_type {
-            None | Some(ExecuteTransactionRequestType::WaitForEffectsCert) => {
-                let mut node_response = self
-                    .fullnode
-                    .execute_transaction_block(tx_bytes, signatures, options.clone(), request_type)
-                    .await
-                    .map_err(error_object_from_rpc)?;
-                // it's not locally executed in indexer, no matter what is the status in the
-                // node
-                node_response.confirmed_local_execution = Some(false);
-                node_response
-            }
-            Some(ExecuteTransactionRequestType::WaitForLocalExecution) => {
-                self.execute_and_index_tx_effects(tx_bytes, signatures, options.clone())
-                    .await?
-            }
-        };
+        let iota_transaction_response = self
+            .fullnode
+            .execute_transaction_block(tx_bytes, signatures, options.clone(), request_type)
+            .await
+            .map_err(error_object_from_rpc)?;
         Ok(IotaTransactionBlockResponseWithOptions {
             response: iota_transaction_response,
             options: options.unwrap_or_default(),
@@ -435,7 +435,76 @@ impl WriteApiServer for WriteApi {
     }
 }
 
+#[async_trait]
+impl WriteApiServer for OptimisticWriteApi {
+    async fn execute_transaction_block(
+        &self,
+        tx_bytes: Base64,
+        signatures: Vec<Base64>,
+        options: Option<IotaTransactionBlockResponseOptions>,
+        request_type: Option<ExecuteTransactionRequestType>,
+    ) -> RpcResult<IotaTransactionBlockResponse> {
+        let iota_transaction_response = match request_type {
+            None | Some(ExecuteTransactionRequestType::WaitForEffectsCert) => {
+                let mut node_response = self
+                    .write_api
+                    .execute_transaction_block(tx_bytes, signatures, options.clone(), request_type)
+                    .await?;
+                // it's not locally executed in indexer, no matter what is the status in the
+                // node
+                node_response.confirmed_local_execution = Some(false);
+                node_response
+            }
+            Some(ExecuteTransactionRequestType::WaitForLocalExecution) => {
+                self.execute_and_index_tx_effects(tx_bytes, signatures, options.clone())
+                    .await?
+            }
+        };
+        Ok(IotaTransactionBlockResponseWithOptions {
+            response: iota_transaction_response,
+            options: options.unwrap_or_default(),
+        }
+        .into())
+    }
+
+    async fn dev_inspect_transaction_block(
+        &self,
+        sender_address: IotaAddress,
+        tx_bytes: Base64,
+        gas_price: Option<BigInt<u64>>,
+        epoch: Option<BigInt<u64>>,
+        additional_args: Option<DevInspectArgs>,
+    ) -> RpcResult<DevInspectResults> {
+        self.write_api
+            .dev_inspect_transaction_block(
+                sender_address,
+                tx_bytes,
+                gas_price,
+                epoch,
+                additional_args,
+            )
+            .await
+    }
+
+    async fn dry_run_transaction_block(
+        &self,
+        tx_bytes: Base64,
+    ) -> RpcResult<DryRunTransactionBlockResponse> {
+        self.write_api.dry_run_transaction_block(tx_bytes).await
+    }
+}
+
 impl IotaRpcModule for WriteApi {
+    fn rpc(self) -> RpcModule<Self> {
+        self.into_rpc()
+    }
+
+    fn rpc_doc_module() -> Module {
+        iota_json_rpc_api::WriteApiOpenRpc::module_doc()
+    }
+}
+
+impl IotaRpcModule for OptimisticWriteApi {
     fn rpc(self) -> RpcModule<Self> {
         self.into_rpc()
     }
--- a/crates/iota-indexer/src/apis/write_api.rs
+++ b/crates/iota-indexer/src/apis/write_api.rs
@@ -110,9 +110,6 @@ impl OptimisticWriteApi {
             .collect::<Result<Vec<_>, FastCryptoError>>()?;
 
         let transaction = Transaction::from_generic_sig_data(tx_data, sigs);
-
-        // TODO: shouldn't return type below be from rust-sdk types?
-        // Is this type correct?
         let response = self
             .fullnode_rest_client
             .execute_transaction(
--- a/crates/iota-indexer/src/apis/write_api.rs
+++ b/crates/iota-indexer/src/apis/write_api.rs
@@ -9,7 +9,7 @@ use diesel::{
     ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper, pg::sql_types, sql_query,
 };
 use downcast::Any;
-use fastcrypto::{encoding::Base64, traits::ToFromBytes};
+use fastcrypto::{encoding::Base64, error::FastCryptoError, traits::ToFromBytes};
 use iota_json_rpc::IotaRpcModule;
 use iota_json_rpc_api::{WriteApiClient, WriteApiServer, error_object_from_rpc};
 use iota_json_rpc_types::{
@@ -64,6 +64,15 @@ pub(crate) struct WriteApi {
     metrics: IndexerMetrics,
 }
 
+type TransactionDataToCommit = (
+    IndexedTransaction,
+    TxIndex,
+    Vec<IndexedEvent>,
+    Vec<EventIndex>,
+    BTreeMap<String, StoredDisplay>,
+    TransactionObjectChangesToCommit,
+);
+
 impl WriteApi {
     pub fn new(
         fullnode_client: HttpClient,
@@ -81,27 +90,23 @@ impl WriteApi {
         }
     }
 
-    async fn execute_and_optimistically_index_tx_effects(
+    async fn execute_and_index_tx_effects(
         &self,
         tx_bytes: Base64,
         signatures: Vec<Base64>,
         options: Option<IotaTransactionBlockResponseOptions>,
-    ) -> RpcResult<IotaTransactionBlockResponse> {
-        let tx_data: TransactionData =
-            bcs::from_bytes(&tx_bytes.to_vec().map_err(IndexerError::FastCrypto)?)
-                .map_err(IndexerError::Bcs)?;
-        let mut sigs = Vec::new();
-        for sig in signatures {
-            sigs.push(
-                GenericSignature::from_bytes(&sig.to_vec().map_err(IndexerError::FastCrypto)?)
-                    .map_err(IndexerError::FastCrypto)?,
-            );
-        }
+    ) -> Result<IotaTransactionBlockResponse, IndexerError> {
+        let tx_data: TransactionData = bcs::from_bytes(&tx_bytes.to_vec()?)?;
+        let sigs = signatures
+            .into_iter()
+            .map(|sig| GenericSignature::from_bytes(&sig.to_vec()?))
+            .collect::<Result<Vec<_>, FastCryptoError>>()?;
+
         let transaction = Transaction::from_generic_sig_data(tx_data, sigs);
 
         // TODO: shouldn't return type below be from rust-sdk types?
         // Is this type correct?
-        let result = self
+        let response = self
             .fullnode_rest_client
             .execute_transaction(
                 &ExecuteTransactionQueryParameters {
@@ -117,12 +122,11 @@ impl WriteApi {
 
         let TransactionExecutionResponse {
             effects,
-            finality: _,
             events,
-            balance_changes: _,
             input_objects,
             output_objects,
-        } = result;
+            ..
+        } = response;
         let tx_digest = *effects.transaction_digest();
 
         if let (Some(input_objects), Some(output_objects)) = (input_objects, output_objects) {
@@ -133,13 +137,15 @@ impl WriteApi {
                 input_objects,
                 output_objects,
             };
-            self.optimistically_index_transaction(&full_tx_data).await?;
+            self.index_transaction(&full_tx_data).await?;
         } else {
-            tracing::warn!("Cannot optimistically index because of missing in/out objs");
+            tracing::warn!(
+                "Cannot optimistically index because of missing in/out objs for tx: {tx_digest}"
+            );
         }
 
         let tx_block_response = self
-            .wait_for_and_return_tx_block_response(tx_digest, options.clone())
+            .wait_for_local_indexing(tx_digest, options.clone())
             .await?;
 
         Ok(IotaTransactionBlockResponseWithOptions {
@@ -149,7 +155,7 @@ impl WriteApi {
         .into())
     }
 
-    async fn wait_for_and_return_tx_block_response(
+    async fn wait_for_local_indexing(
         &self,
         tx_digest: TransactionDigest,
         options: Option<IotaTransactionBlockResponseOptions>,
@@ -184,7 +190,7 @@ impl WriteApi {
         .await
     }
 
-    async fn optimistically_index_transaction(
+    async fn index_transaction(
         &self,
         full_tx_data: &CheckpointTransaction,
     ) -> Result<(), IndexerError> {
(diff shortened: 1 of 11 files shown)


crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -39,11 +449,27 @@ impl WriteApiServer for WriteApi {
...
+            }
+            Some(ExecuteTransactionRequestType::WaitForLocalExecution) => {
+                self.execute_and_optimistically_index_tx_effects(
+                    tx_bytes,
+                    signatures,
+                    options.clone(),
+                )
+                .await?
+            }
+        };
```
kodemartin: We don't need to bind the logic to the `request_type` that is defined in the context of the node.

Instead it is ok to execute always optimistically here.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
         Self {
             fullnode: fullnode_client,
+            fullnode_rest_client,
+            inner,
+            store,
+            metrics,
+        }
+    }
+
+    async fn execute_and_optimistically_index_tx_effects(
```
kodemartin: I think we can do with less verbosity by removing the `optimistically` part:

```rust
    async fn execute_and_index_transaction(
```

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        &self,
+        tx_bytes: Base64,
+        signatures: Vec<Base64>,
+        options: Option<IotaTransactionBlockResponseOptions>,
+    ) -> RpcResult<IotaTransactionBlockResponse> {
```
kodemartin: You can use `Result<_, IndexerError>` here to avoid all these `map_err` below, and be consistent with this implementation.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        let tx_data: TransactionData =
+            bcs::from_bytes(&tx_bytes.to_vec().map_err(IndexerError::FastCrypto)?)
+                .map_err(IndexerError::Bcs)?;
+        let mut sigs = Vec::new();
+        for sig in signatures {
+            sigs.push(
+                GenericSignature::from_bytes(&sig.to_vec().map_err(IndexerError::FastCrypto)?)
+                    .map_err(IndexerError::FastCrypto)?,
+            );
+        }
```
kodemartin: Better to use `signatures.into_iter().map().collect::<Result<Vec<_>>()`.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        let transaction = Transaction::from_generic_sig_data(tx_data, sigs);
+
+        // TODO: shouldn't return type below be from rust-sdk types?
+        // Is this type correct?
```
kodemartin: This is ok. Why not?

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        let result = self
```
kodemartin: Nit: Add some semantics
```rust
        let response = self
```

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+            .map_err(|e| IndexerError::Generic(e.to_string()))?;
+
+        let TransactionExecutionResponse {
+            effects,
+            finality: _,
+            events,
+            balance_changes: _,
+            input_objects,
+            output_objects,
+        } = result;
```
kodemartin: You can [ignore remaining parts](https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html#remaining-parts-of-a-value-with-)
```suggestion
        let TransactionExecutionResponse {
            effects,
            events,
            input_objects,
            output_objects,
            ..
        } = result;
```

But I think you can even spare the reader of the additional lines of code if you just access the `result` fields directly later (e.g. `result.input_objects`).

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        let tx_digest = *effects.transaction_digest();
+
+        if let (Some(input_objects), Some(output_objects)) = (input_objects, output_objects) {
```
kodemartin: We need to ensure also `!iinput_objects.is_empty() and !output_objects.is_empty()`. A `match would then be more concise.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+            let full_tx_data = CheckpointTransaction {
+                transaction,
+                effects,
+                events,
+                input_objects,
+                output_objects,
+            };
+            self.optimistically_index_transaction(&full_tx_data).await?;
```
kodemartin: I would favor less verbosity
```rust
            self.index_transaction(&full_tx_data).await?;
```

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        } else {
+            tracing::warn!("Cannot optimistically index because of missing in/out objs");
```
kodemartin: Adding the `digest` would be useful.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        }
+
+        let tx_block_response = self
+            .wait_for_and_return_tx_block_response(tx_digest, options.clone())
```
kodemartin:
```rust
            .wait_for_local_indexing(tx_digest, options.clone())
```

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        })
+        .await
+    }
+
+    async fn optimistically_index_transaction(
+        &self,
+        full_tx_data: &CheckpointTransaction,
+    ) -> Result<(), IndexerError> {
+        let assigned_global_order = self
+            .get_or_assign_optimistic_tx_global_order(full_tx_data.transaction.digest())
```
kodemartin: If the order is already assigned then we shouldn't continue with indexing.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+            tx_indices,
+            indexed_events,
+            events_indices,
+            indexed_displays,
+            object_changes,
+        )
+        .await
+    }
+
+    async fn get_or_assign_optimistic_tx_global_order(
```
kodemartin: I think we can limit the scope of the method to `assign_tx_global_order` and return `None` if it is already assigned, so that we skip indexing.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+            &pool,
+            |conn| {
+                sql_query(
+                    r#"
+                        INSERT INTO tx_global_order (tx_digest, global_sequence_number)
+                        SELECT $1, MAX(tx_sequence_number) FROM tx_digests
+                        ON CONFLICT (tx_digest) DO NOTHING
+                    "#,
+                )
+                .bind::<sql_types::Bytea, _>(&tx_digest_bytes)
```
kodemartin: You can insert a `TxGlobalOrder` value directly, no? Why using the raw query?

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+            },
+            Duration::from_secs(30)
+        )?;
+
+        run_query_async!(&pool, |conn| {
+            tx_global_order::table
+                .select(TxGlobalOrder::as_select())
+                .filter(tx_global_order::tx_digest.eq(tx_digest_bytes))
+                .first::<TxGlobalOrder>(conn)
+        })
```
kodemartin: You can use [returning](https://docs.diesel.rs/master/diesel/query_builder/struct.InsertStatement.html#method.returning) to get the assigned value in one go.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        &self,
+        full_tx_data: &CheckpointTransaction,
+        assigned_global_order: &TxGlobalOrder,
+    ) -> IndexerResult<(
+        IndexedTransaction,
+        TxIndex,
+        Vec<IndexedEvent>,
+        Vec<EventIndex>,
+        BTreeMap<String, StoredDisplay>,
+        TransactionObjectChangesToCommit,
```
kodemartin: Consider wrapping into a type `TransactionDataToCommit` for conciseness.

kodemartin: You can then also wrap `CheckpointTransaction` into a `TransactionExtractor`, implement methods to extract each individual component to index and implement `TryFrom` or `From` for `CheckpointTransaction`.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -2,30 +2,440 @@
...
+        self.store
+            .persist_optimistic_tx_indices(Self::optimistic_tx_indices_from_indexed_tx_indices(
+                tx_indices,
+            ))
+            .await?;
+
+        Ok(())
+    }
+
+    fn optimistic_event_indices_from_indexed_event_indices(
```
kodemartin: Please consider simplifying: Copilot suggests

```rust
fn optimistic_event_indices_from_indexed_event_indices(
    event_indices: Vec<EventIndex>,
) -> OptimisticEventIndices {
    let splits: Vec<_> = event_indices.into_iter().map(|i| i.split()).collect();

    OptimisticEventIndices {
        optimistic_event_emit_packages: splits.iter().map(|t| t.0.clone().into()).collect(),
        optimistic_event_emit_modules: splits.iter().map(|t| t.1.clone().into()).collect(),
        optimistic_event_senders: splits.iter().map(|t| t.2.clone().into()).collect(),
        optimistic_event_struct_packages: splits.iter().map(|t| t.3.clone().into()).collect(),
        optimistic_event_struct_modules: splits.iter().map(|t| t.4.clone().into()).collect(),
        optimistic_event_struct_names: splits.iter().map(|t| t.5.clone().into()).collect(),
        optimistic_event_struct_instantiations: splits.iter().map(|t| t.6.clone().into()).collect(),
    }
}
```

crates/iota-indexer/src/lib.rs
```diff
@@ -230,16 +231,25 @@ impl Default for IotaNamesOptions {
 }
 
 pub async fn build_json_rpc_server(
+    store: PgIndexerStore,
     prometheus_registry: &Registry,
     reader: IndexerReader,
     config: &IndexerConfig,
     custom_runtime: Option<Handle>,
+    metrics: IndexerMetrics,
```
kodemartin: This breaks the public API of the library.

Alternatively we can add a new `OptimisticWriteApi` that wraps `WriteApi` and holds the new optimistic methods.

Then we can add a new function here with the new signature to build the optimistic rpc-server.

kodemartin (changes requested):

crates/iota-indexer/src/apis/write_api.rs
tomxey: Users could gain a bit of performance that way by not requesting local execution if they don't need it.
Or do we assume that anyone who wants performance will just use the node directly?

crates/iota-indexer/src/apis/write_api.rs
tomxey: The iota-rest-api is producing return value using type defined in https://github.com/iotaledger/iota/blob/8d452173d6ad65e8e6089cfe88d9c070686cc812/crates/iota-rest-api/src/transactions/execution.rs#L197 which contains TransactionEffects that comes from iota-rust-sdk.
The return value is serialized to bcs here: https://github.com/iotaledger/iota/blob/8d452173d6ad65e8e6089cfe88d9c070686cc812/crates/iota-rest-api/src/transactions/execution.rs#L169

-------------

Then the value is deserialized from bcs in https://github.com/iotaledger/iota/blob/43412a93f4ef752104460ef630daea0631621f52/crates/iota-rest-api/src/client/mod.rs#L123 

But the type used here is https://github.com/iotaledger/iota/blob/43412a93f4ef752104460ef630daea0631621f52/crates/iota-rest-api/src/client/mod.rs#L139 which is a different type (despite the same name) and the TransactionEffects there come from iota-types.

---------------

That's how the client was implemented before, and it works so far, but I wonder what guarantees we have about this.

crates/iota-indexer/src/apis/write_api.rs
tomxey: In general it is not 100% correct, since the order is always written at the beginning of the indexing, so it is not a guarantee that indexing is finished, or do we want to assume it now and fix it later?

crates/iota-indexer/src/apis/write_api.rs
tomxey: As above.

crates/iota-indexer/src/apis/write_api.rs
tomxey: Raw query makes it possible to do the read of tx_digests and insert in one query, otherwise it would be split into two queries.

crates/iota-indexer/src/apis/write_api.rs
tomxey: I think creting `TransactionDataToCommit`is a good idea.
Not sure about the `TransactionExtractor` though. Not all components that are being indexed are extracted independently, so it seems to me that implementation of the extractor would not be that simple.

crates/iota-indexer/src/apis/write_api.rs
tomxey: There is some cloning in the simplified version, but I suppose it should be ok. Applying.

crates/iota-indexer/src/apis/write_api.rs
tomxey: I think it looks clearer if this struct is unpacked. I will go with `..` for now.

crates/iota-indexer/src/apis/write_api.rs
tomxey: Why should we check for empty vectors? If the data is filled I think we can assume that it is correct.

crates/iota-indexer/src/apis/write_api.rs
tomxey: It would be possible, but with a bit more complex raw query, since with a basic one no value would be returned in case row is already present (`ON CONFLICT (tx_digest) DO NOTHING`).
I'll keep this comment on hold until it is decided what to do with remaining comments in this function.

crates/iota-indexer/src/lib.rs
tomxey: Is this library intended to be used by users outside of IF?

crates/iota-indexer/src/apis/write_api.rs
kodemartin: Does the rest-api that we use here wait for local execution? It doesn't seam like it: https://iotaledger.github.io/iota/src/iota_rest_api/transactions/execution.rs.html#67

crates/iota-indexer/src/apis/write_api.rs
kodemartin: We should rely on what the client returns in my opinion.

crates/iota-indexer/src/apis/write_api.rs
kodemartin: because we can actually exhaust the logic here without making assumptions about the length of the vectors.

crates/iota-indexer/src/apis/write_api.rs
kodemartin: If checkpoint indexing has been started we should make it so that even if it fails it will reindex the checkpoint. This should be a requirement for https://github.com/iotaledger/iota/issues/7290

crates/iota-indexer/src/apis/write_api.rs
kodemartin: https://github.com/iotaledger/iota/pull/7235#discussion_r2144689071

crates/iota-indexer/src/lib.rs
kodemartin: This is a public interface, and can be used by external users as well.

crates/iota-indexer/src/apis/write_api.rs
kodemartin: > Not all components that are being indexed are extracted independently

What components? The idea is to do what you now do inline with the `CheckpointTransaction` inside a wrapper struct with dedicated methods.

Please also note that in the future iteration we would like to make this logic part of the library to be used by the GraphQL.

crates/iota-indexer/src/apis/write_api.rs
kodemartin: Good point, thank you!

crates/iota-indexer/src/apis/write_api.rs
kodemartin: Still `RETURNING` would spare you one query.

crates/iota-indexer/src/apis/write_api.rs
tomxey: Seems like it's not waiting, which is to our favor, since indexer doesn't need it, and it's a bit faster that way.

kodemartin: Exactly, so the flag `WaitForLocalExecution` is irrelevant.

kodemartin: Plus we shouldn't rely on the node json-rpc at all.

crates/iota-indexer/src/apis/write_api.rs
tomxey: Okay, will use `RETURNING`

crates/iota-indexer/src/apis/write_api.rs
tomxey: Rewritten to return None in case order was already assigned earlier. This will effectively stop optimistic indexing of given tx.

crates/iota-indexer/src/apis/write_api.rs
tomxey: As in linked comment.

crates/iota-indexer/src/lib.rs
tomxey: Okay, I'll create `build_optimistic_json_rpc_server` and `OptimisticWriteApi`

crates/iota-indexer/src/apis/write_api.rs
tomxey: Ok, will add the check

crates/iota-indexer/src/apis/write_api.rs
tomxey: Ok, let's index optimistically in every case

crates/iota-indexer/src/apis/write_api.rs
tomxey: Extracted the logic to separate struct, and to separate file altogether. 83cada9b5c558700b949ed75bb2849afd45bc18a

crates/iota-indexer/src/apis/write_api.rs
tomxey: 9d4cd308af07e734c4e0fdf73b9a0b9d350bffdc

crates/iota-indexer/src/apis/write_api.rs
tomxey: bcd99dca719364eaba6fdc8dccafa0e3f1cd8047

crates/iota-indexer/src/apis/write_api.rs
tomxey: bcd99dca719364eaba6fdc8dccafa0e3f1cd8047

crates/iota-indexer/src/apis/write_api.rs
tomxey: bcd99dca719364eaba6fdc8dccafa0e3f1cd8047

crates/iota-indexer/src/apis/write_api.rs
tomxey: 1ff368ec08ae6273912c5b25284f26061ade19b0

crates/iota-indexer/src/lib.rs
tomxey: 6fed08b9353c8d77d66c3a7c3f476e99528c130f

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+pub struct OptimisticTransactionExecutor {
+    fullnode_rest_client: iota_rest_api::Client,
+    indexer_reader: IndexerReader,
+    store: PgIndexerStore,
+    metrics: IndexerMetrics,
+}
+
+impl OptimisticTransactionExecutor {
+    pub fn new(
+        fullnode_rest_client: iota_rest_api::Client,
```
kodemartin: I suggest we use the raw `url` instead of the client and construct the client as a side effect. This way we can abstract away the client and api implementation from the public API.

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+                sql_query(
+                    r#"
+                        INSERT INTO tx_global_order (tx_digest, global_sequence_number)
+                        SELECT $1, MAX(tx_sequence_number) FROM tx_digests
+                        ON CONFLICT (tx_digest) DO NOTHING
+                        RETURNING *;
+                    "#,
+                )
+                .bind::<sql_types::Bytea, _>(&tx_digest_bytes)
+                .load::<TxGlobalOrder>(conn)
```
kodemartin: You can take `Option<TxGlobalOrder>` directly

```rust
                .get_result::<TxGlobalOrder>(conn)
                .optional()
```

Requires `use diesel::OptionalExtension`.

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+            return Ok(());
+        };
+
+        let extractor = TransactionExtractor::new(
+            full_tx_data,
+            assigned_global_order
+                .optimistic_sequence_number
+                .expect("Optimistic sequence number is always set for data read from DB")
+                .try_into()
+                .unwrap(),
```
kodemartin: We should return an internal-server error instead.

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+            optimistic_tx,
+            optimistic_tx_indices,
+            optimistic_events,
+            optimistic_event_indices,
+            indexed_displays,
+            object_changes,
+        ))
+    }
+
+    fn optimistic_event_indices_from_indexed_event_indices(
```
kodemartin: Nit: Simplify to 
```rust
    fn optimistic_event_indices(
```
The function signature has semantics as well, so as we don't have to be so verbose with the function name.

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+            optimistic_event_struct_modules: splits.iter().map(|t| t.4.clone().into()).collect(),
+            optimistic_event_struct_names: splits.iter().map(|t| t.5.clone().into()).collect(),
+            optimistic_event_struct_instantiations: splits
+                .iter()
+                .map(|t| t.6.clone().into())
+                .collect(),
+        }
+    }
+
+    fn optimistic_tx_indices_from_indexed_tx_indices(tx_index: TxIndex) -> OptimisticTxIndices {
```
kodemartin: Ni: Simplity fo 
```rust
    fn optimistic_tx_indices(tx_index: TxIndex) -> OptimisticTxIndices {
```

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -0,0 +1,393 @@
...
+    ) -> Self {
+        Self {
+            fullnode_rest_client,
+            indexer_reader,
+            store,
+            metrics,
+        }
+    }
+
+    pub(crate) async fn execute_and_index_tx_effects(
```
kodemartin: Nit: I think it is safe to name it
```suggestion
    pub(crate) async fn execute_and_index_transaction(
```
Indexing `tx_effects` is abstracted under `index_transaction`

kodemartin (changes requested):
Suggested a few minor changes, and nitpicks, but otherwise looks good!

sergiupopescu199 (approved):

crates/iota-indexer/src/optimistic_indexing.rs
tomxey: Actually, why `optimistic_sequence_number` in TxGlobalOrder is i64? We could turn it into u64 and avoid a conversion here.

kodemartin: Is that possible? I think postgres only supports signed integers.

tomxey: Indeed, they are not supported by postgres. I'll handle the conversion error better here then

kodemartin (approved):

PR #7237, opened 2025-06-05
fix(indexer): ensure partition pruning in analytics queries

### LINKED ISSUE #7196: [indexer]: analytical queries on transactions cause lock contention
## Description

The `transactions` table is partitioned by `tx_sequence_number`. On each new epoch a new partition is created by the internal logic of the indexer writer. When the epochs advance in large number (e.g. > 700) so do the partitions and we noticed a performance degradation in local test networks. E.g.:

```
Advanced epoch partition 100 for table transactions elapsed=3.349437031
Advanced epoch partition 200 for table transactions elapsed=0.246397175
Advanced epoch partition 300 for table transactions elapsed=1.709999232
Advanced epoch partition 400 for table transactions elapsed=5.489943558
Advanced epoch partition 500 for table transactions elapsed=6.957459044
Advanced epoch partition 600 for table transactions elapsed=14.850626357
Advanced epoch partition 700 for table transactions elapsed=97.062561708
Advanced epoch partition 800 for table transactions elapsed=178.169522202
Advanced epoch partition 900 for table transactions elapsed=598.575338398
Advanced epoch partition 1000 for table transactions elapsed=671.470147444
Advanced epoch partition 1100 for table transactions elapsed=665.144655395
Advanced epoch partition 1200 for table transactions elapsed=794.534614893
Advanced epoch partition 1300 for table transactions elapsed=1278.656600706
```

This is found to be caused by the analytical queries (e.g. `persist_tx_count_metrics`) that try to join on the `transactions` table without specifying specific ranges of `tx_sequence_number` thus causing execution to scan all partitions. 

The queries should be optimized to enable [partition pruning](https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITION-PRUNING)

The problem has been verified by investigating lock contention in the server with:

```sql
SELECT pid, relation::regclass, locktype, mode, granted, query
FROM pg_locks
JOIN pg_stat_activity USING (pid)
WHERE relation::regclass::text LIKE 'transactions_partition%';
```

This showed specific queries getting `SharedAccessLock` on the transactions table.

This has been further verified by switching off the analytical worker, in which case performance returned to the expected levels.

kodemartin:
# Description of change

This patch refactors the `tx_count`, `address`, and `move_call` metrics queries to ensure partition pruning on the `transactions` table during query planing. 

The main change is that when joining the `transactions` table we now use bounds in the `WHERE` clause that are known at the initialization of the execution.

To this end we introduce two new procedures:

* `calculate_tx_count_metrics`: Performs a query to get the `tx_sequence_number` range first and then execute the metric calculation query. The use of a procedure allows a single query to the db.
* `persist_address_analytics`: Consolidates the two queries for `addresses` and `active_addresses` that were previously distinct to enforce partition pruning and a single join with the `transactions` table.

The `move_call` metrics query is also refactored accordingly.

Fixes #7196 

The patch has been tested against a local and a production-scale network to assess its efficiency. The key measure was the elapsed time for advancing partition for the table `transactions`. This measure has been compared between a setup without the analytical worker and one with the analytical worker. The two measures were found to be of the same order after the optimizations.

### Release Notes

- [x] Indexer: Optimize analytics queries by ensuring partition pruning


sergiupopescu199 (approved):
lgtm ✨

samuel-rufi (approved):
Fantastic PR, looks good to me.

PR #7293, opened 2025-06-10
fix(ts-sdk): Export `getGraphQLUrl`

marc2332:


.changeset/itchy-avocados-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@iota/iota-sdk': minor
+---
+
+Rexport `getGraphQLUrl` correctly
```
msarcev: Re-export (or Export) ? Rexport is not a thing

msarcev (changes requested):

.changeset/itchy-avocados-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@iota/iota-sdk': minor
```
begonaalvarezd: I would say its more like a patch

begonaalvarezd (changes requested):

begonaalvarezd (approved):

msarcev (approved):

PR #7306, opened 2025-06-10
upstream(node): core: factor out parallel live object indexing

muXxer:
# Description of change

- Upstream range: [v1.35.4, v1.36.2)
- Port commit: https://github.com/MystenLabs/sui/commit/6159973ae01f7b74505271d5ae76681f6594fe59
- Descriptions from commits
Factor out the parallel live object indexing, used for initializing the
rest indexes, into more general purpose and reusable logic.

Part of #3990 

### DIFF
--- a/crates/iota-core/src/rest_index.rs
+++ b/crates/iota-core/src/rest_index.rs
@@ -32,11 +32,9 @@ use typed_store::{
 };
 
 use crate::{
-    authority::{
-        AuthorityStore, authority_per_epoch_store::AuthorityPerEpochStore,
-        authority_store_tables::LiveObject,
-    },
+    authority::{AuthorityStore, authority_per_epoch_store::AuthorityPerEpochStore},
     checkpoints::CheckpointStore,
+    par_index_live_object_set::{LiveObjectIndexer, ParMakeLiveObjectIndexer},
 };
 
 const CURRENT_DB_VERSION: u64 = 0;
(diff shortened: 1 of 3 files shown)


crates/iota-core/src/rest_index.rs
```diff
@@ -732,3 +623,90 @@ fn try_create_coin_index_info(object: &Object) -> Option<(CoinIndexKey, CoinInde
...
+            }
+
+            Owner::Shared { .. } | Owner::Immutable => {}
+        }
+
+        // Look for CoinMetadata<T> and TreasuryCap<T> objects
+        if let Some((key, value)) = try_create_coin_index_info(&object) {
+            use std::collections::hash_map::Entry;
+
+            match self.coin_index.lock().unwrap().entry(key) {
```
semenov-vladyslav, line 687: What is going on here? Why is entry removed when the key is found? Why CoinIndexInfo::merge is non-commutative? Why the db table is not updated when the key is not found? What are the allowed transitions for value (eg. can it go from coin_metadata_object_id = Some() to coin_metadata_object_id = None or to treasury_object_id = Some())? Are there unit tests for various coin object transitions?

This logic is very non-trivial and should at least be commented.

semenov-vladyslav (changes requested):
Looks good except for one very weird place. IMHO, should be addressed.

semenov-vladyslav (approved):
The current version of the code doesn't make any breaking changes, just make the existing code more generic. The questions not answered will be investigated in another issue.

bingyanglin (approved):

crates/iota-core/src/rest_index.rs
muXxer: I created an issue #7317

PR #7321, opened 2025-06-11
chore(shared-crypto): reserve the former bridge event intent scope

miker83z:
# Description of change

Renamed IntentScope variant `BridgeEventUnused` to `BridgeEventDeprecated`.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/shared-crypto/src/intent.rs
+++ b/crates/shared-crypto/src/intent.rs
@@ -66,7 +66,8 @@ pub enum IntentScope {
     SenderSignedTransaction = 4, // Used for an authority signature on a user signed transaction.
     ProofOfPossession = 5,       /* Used as a signature representing an authority's proof of
                                   * possession of its authority key. */
-    Reserved = 6, // it was thought for bridge purposes but it was never included in messages.
+    BridgeEventDeprecated = 6, /* Deprecated. Should not be reused. Introduced for bridge
+                                * purposes but was never included in messages. */
     ConsensusBlock = 7, // Used for consensus authority signature on block's digest.
 }
 
--- a/crates/shared-crypto/src/intent.rs
+++ b/crates/shared-crypto/src/intent.rs
@@ -66,8 +66,8 @@ pub enum IntentScope {
     SenderSignedTransaction = 4, // Used for an authority signature on a user signed transaction.
     ProofOfPossession = 5,       /* Used as a signature representing an authority's proof of
                                   * possession of its authority key. */
-    BridgeEventUnused = 6, // for bridge purposes but it's currently not included in messages.
-    ConsensusBlock = 7,    // Used for consensus authority signature on block's digest
+    Reserved = 6, // it was thought for bridge purposes but it was never included in messages.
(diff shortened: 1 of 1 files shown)


crates/shared-crypto/src/intent.rs
```diff
@@ -66,8 +66,8 @@ pub enum IntentScope {
     SenderSignedTransaction = 4, // Used for an authority signature on a user signed transaction.
     ProofOfPossession = 5,       /* Used as a signature representing an authority's proof of
                                   * possession of its authority key. */
-    BridgeEventUnused = 6, // for bridge purposes but it's currently not included in messages.
-    ConsensusBlock = 7,    // Used for consensus authority signature on block's digest
+    Reserved = 6, // it was thought for bridge purposes but it was never included in messages.
```
Alex6323:
```suggestion
    Reserved = 6, // Reserved for later use. Introduced for bridge purposes but was never included in messages.
```

Alex6323: Feel free to ignore the suggestion if you prefer yours.

Alex6323 (approved):

crates/shared-crypto/src/intent.rs
Thoralf-M: I think even with that it's not really clear, like reserved for what?
Shouldn't there be some message like `Instead of adding a new variant, rename and use this one`?

muXxer: What if we ever reserve another one? Maybe `ReservedBridgeEvent` is better?

Thoralf-M: But is it actually reserved for an eventual bridge or can't we just use it for the next one we would add?

muXxer: No it's not reserved for an eventual bridge. And no such message was ever sent I guess, so it might be fine if we reuse it 🤷

miker83z: It's a dead number, shouldn't be reused.

Alex6323: okay, I guess there's no good solution then and we can just keep the old variant name around. I was thinking about calling it `Unused`, but if we ever have another one, then we need to add a suffix to those variants { Unused1, Unused2, ... }. Not sure if that's really better.

muXxer (approved):

crates/shared-crypto/src/intent.rs
miker83z: Discussed in a meeting in which the consensus was `BridgeEventDeprecated`, thank you all! -> [409d3cc](https://github.com/iotaledger/iota/pull/7321/commits/409d3ccbbcaaaa379c57cd73e23ff0eba7b7e9cf)

Thoralf-M (approved):

PR #7360, opened 2025-06-13
upstream(indexer): fix system-state values at epoch table

### LINKED ISSUE #7313: upstream(indexer): Add json serialized SuiSystemStateSummary to epochs table
## Description

https://www.github.com/MystenLabs/sui/pull/19428
https://www.github.com/MystenLabs/sui/pull/19532

kodemartin:
# Description of change

This patch fixes the problem of overriding the system-state value in the `epochs` table with the system state of the next epoch when we reach the epoch end. 

This implies that currently for each epoch `i` in the indexer db we have stored the system-state for `i+1` which leads to wrong results for the `iotax_getCommitteeInfo` request.

The issue if fixed by excluding the system-state from the `ON CONFLICT UPDATE` statement at epochs end, and by introducing a new migration that restores the correct system-state values for all epochs except for the first (`epoch_id = 0`). For the latter we have to backfill the correct value once we setup the necessary infrastructure.

Fixes #7313 

### Release Notes

- [x] Indexer: Fixes system-state values in epoch tables.


sergiupopescu199 (approved):
lgtm ✨

crates/iota-indexer/src/types.rs
```diff
@@ -242,7 +242,7 @@ impl IndexedEpochInfo {
     /// X. `network_total_tx_num_at_last_epoch_end` is needed to determine
     /// the number of transactions that occurred in the epoch X-1.
     pub fn from_end_of_epoch_data(
-        system_state_summary: &IotaSystemStateSummary,
+        _system_state_summary: &IotaSystemStateSummary,
```
tomxey, line 245: Could we just modify the usage and not pass this argument?

kodemartin: It would be a breaking change I am afraid.

tomxey (approved):

PR #7371, opened 2025-06-13
feat(consensus): Improve synchronizer

### LINKED ISSUE #7370: (mysticeti): prioritize the author of the block and nodes that reference this block (this can be easily retrieved from the BlockManager as we track suspended child blocks of a missing block).

NaitsabesMue:
# Description of change

This PR improves block sync performance when a single/few missing block blocks may suspend many others.

Problem:
Random selection in the periodic synchronizer often takes too long to locate blocks that are held by only a few authorities, leaving large portions of the DAG stalled.

Fix:
Add a knowledge-based fetch path to the periodic synchronizer:

- Leverage extended missing_blocks map from BlockManager, which now tracks for each missing block a set of authorities that are known to have it.
- Prioritize “knowledge-based” peers in the scheduler: when requesting missing blocks, first query exactly those authorities that should already hold them, rather than picking peers at random.
- Issue random‐peer requests alongside the knowledge‐based queries

This change avoids long delays in recovering blocks that live on only a few nodes by directing periodic sync requests to the right peers up front.

closes #7370

- X Basic tests (linting, compilation, formatting, unit/integration tests)
- X Patch-specific tests (correctness, functionality coverage)
- X I have added tests that prove my fix is effective or that my feature works
- X I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


polinikita:
We need to figure out how to make a consistent change. You are taking a block_manager as a parameter. However, for retrieving missing blocks in periodic synchronizer, we access these blocks through a comman in a core thread because we wanted to avoid concurrent access for this structure. So, if we have a block_manager in synchronizer, then I would remove a command from the core thread for retrieving missing blocks and read them directly.

consensus/core/src/block_manager.rs
```diff
@@ -721,6 +722,30 @@ impl BlockManager {
...
+
+    /// Returns the list of authority indices that authored suspended blocks
+    /// which list `missing` as one of their missing ancestors.
+    pub fn dependents_of(&self, missing: &BlockRef) -> Vec<AuthorityIndex> {
+        self.missing_ancestors
+            .get(missing)
+            .map(|children| {
+                children
+                    .iter()
+                    .filter_map(|child_ref| self.suspended_blocks.get(child_ref))
```
polinikita: Is this filter_map needed? I was thinking that we only put suspended block refs insides the value for a missing block

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1026,58 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+            let bm = block_manager.read();
+
+            missing_blocks
+                .iter()
+                .filter(|&missing| bm.dependent_count(missing) > HOT_DEPENDENT_THRESHOLD)
+                .for_each(|missing| {
+                    // Always bind authors as mutable to allow shuffling in non-test environments.
+                    // In test builds, the `mut` is unused, so we silence Clippy.
+                    #[allow(unused_mut)]
+                    let mut authors = bm.dependents_of(missing);
```
polinikita: The returned set is a vector where same authors can be considered multiple times. i would suggest to change that function to output each authors at most once or represent multiplicities.

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1026,58 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+
+                    #[cfg(not(test))]
+                    authors.shuffle(&mut ThreadRng::default());
```
polinikita: Shuffling might be an expensive operation given that we take only one peer from the set

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1026,58 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+            // Acquire the lock on the BlockManager
+            let bm = block_manager.read();
+
+            missing_blocks
+                .iter()
+                .filter(|&missing| bm.dependent_count(missing) > HOT_DEPENDENT_THRESHOLD)
+                .for_each(|missing| {
+                    // Always bind authors as mutable to allow shuffling in non-test environments.
+                    // In test builds, the `mut` is unused, so we silence Clippy.
+                    #[allow(unused_mut)]
```
polinikita: I think that too many flags inside this function. This makes it harder to understand. Also some unncessary discussion below on reasoning why the flag is needed

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1026,58 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+                    authors.shuffle(&mut ThreadRng::default());
+
+                    if let Some(peer) = authors.first().copied() {
+                        // BlockRef is Copy, so Clippy suggests using `*missing` instead of
+                        // `.clone()`. However, using `*missing` here breaks
+                        // type inference in this context, while `.clone()`
+                        // is clear, safe, and equivalent. We explicitly
+                        // allow Clippy's warning for clarity and correctness.
+                        #[allow(clippy::clone_on_copy)]
+                        let block_set = std::iter::once(missing.clone()).collect::<BTreeSet<_>>();
```
polinikita: For every such block we make a separete request. This should work differently and we need to aggregate all requests for the same peer

polinikita (changes requested):

consensus/core/src/synchronizer.rs
NaitsabesMue: Thanks, changed to a random index.

consensus/core/src/synchronizer.rs
NaitsabesMue: Changed the logic, choosing a random authority and then collecting all missing blocks it should know.

consensus/core/src/block_manager.rs
NaitsabesMue: Removed the filter map. Coded the logic directly.

consensus/core/src/synchronizer.rs
NaitsabesMue: Changed to a unique vector. Might add multiples (and sample weight) later if needed.

NaitsabesMue:
> We need to figure out how to make a consistent change. You are taking a block_manager as a parameter. However, for retrieving missing blocks in periodic synchronizer, we access these blocks through a comman in a core thread because we wanted to avoid concurrent access for this structure. So, if we have a block_manager in synchronizer, then I would remove a command from the core thread for retrieving missing blocks and read them directly.

 - Removed direct `BlockManager` RwLock reads in hot-fetch path  
- Use `get_suspended_blocks()` and `get_missing_ancestors()` on `core_dispatcher` to snapshot suspension state

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1029,67 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+        let highest_rounds = Self::get_highest_accepted_rounds(dag_state, &context);
+
+        // Hot-path: when the ratio of suspended to missing blocks exceeds
+        // HOT_FETCH_RATIO_THRESHOLD, proactively fetch a batch of missing
+        // blocks from the authors of blocks referencing them.
+        const HOT_FETCH_RATIO_THRESHOLD: usize = 50;
+        {
+            if suspended_blocks.len() > HOT_FETCH_RATIO_THRESHOLD {
+                // Step 1: Map authors to their missing blocks
+                let mut author_to_blocks: HashMap<AuthorityIndex, Vec<BlockRef>> = HashMap::new();
```
piotrm50: Variable name `author_to_blocks` is not entirely correct, as it also contains authorities that reference the missing block. Maybe rename to `authorities_to_blocks`.

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1029,67 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+        let mut request_futures = FuturesUnordered::new();
+
+        let highest_rounds = Self::get_highest_accepted_rounds(dag_state, &context);
+
+        // Hot-path: when the ratio of suspended to missing blocks exceeds
+        // HOT_FETCH_RATIO_THRESHOLD, proactively fetch a batch of missing
+        // blocks from the authors of blocks referencing them.
+        const HOT_FETCH_RATIO_THRESHOLD: usize = 50;
+        {
+            if suspended_blocks.len() > HOT_FETCH_RATIO_THRESHOLD {
```
piotrm50: Ratio is not calculated, but a length of suspended blocks is used.

consensus/core/src/synchronizer.rs
```diff
@@ -1156,10 +1228,34 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+    missing_ancestors: &BTreeMap<BlockRef, BTreeSet<BlockRef>>,
+) -> Vec<AuthorityIndex> {
+    if let Some(dependents) = missing_ancestors.get(missing) {
+        let mut seen = HashSet::with_capacity(dependents.len());
+        let mut result = Vec::new();
+        for dependent in dependents {
+            let sb = suspended_blocks
+                .get(dependent)
+                .expect("Suspended block for missing ancestor should exist.");
+            let author = sb.block.author();
```
piotrm50: Instead of exposing `block` we can create a getter to retrieve `author` directly.

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1029,67 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
     ) -> Vec<(BlocksGuard, Vec<Bytes>, AuthorityIndex)> {
+        let mut request_futures = FuturesUnordered::new();
+
+        let highest_rounds = Self::get_highest_accepted_rounds(dag_state, &context);
+
+        // Hot-path: when the ratio of suspended to missing blocks exceeds
+        // HOT_FETCH_RATIO_THRESHOLD, proactively fetch a batch of missing
+        // blocks from the authors of blocks referencing them.
+        const HOT_FETCH_RATIO_THRESHOLD: usize = 50;
+        {
```
piotrm50: Why the additional scope?

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1029,67 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+                        author_to_blocks.entry(author).or_default().push(*missing);
+                    }
+                }
+                // Step 2: Choose one random authority from the map
+
+                #[cfg(not(test))]
+                let chosen = author_to_blocks.iter().choose(&mut ThreadRng::default());
+
+                #[cfg(test)]
+                let chosen = author_to_blocks.iter().next();
```
piotrm50: Do I understand correctly that we're randomly selecting one authority to fetch a subset of all missing blocks? I think we should send request to the corresponding authorities for all the missing blocks rather than just one.

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1029,67 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+                            highest_rounds.clone(),
+                            FETCH_REQUEST_TIMEOUT,
+                            1,
+                        ));
+                    }
+                }
+            }
+        }
+
+        // Proceed with the usual fetching
```
piotrm50: Why do we continue with regular fetching if we've already sent requests for all (some?) missing blocks to their authors or authorities that have referenced them?

piotrm50 (dismissed):

consensus/core/src/synchronizer.rs
NaitsabesMue: We randomly select an authority and then request all missing blocks that are referenced by a block authored by this authority.

consensus/core/src/synchronizer.rs
NaitsabesMue: Not all missing blocks, see above. We could, however, remove these blocks from the subsequent fetches, or increase `MAX_AUTHORITIES_TO_FETCH_PER_BLOCK` by 1.

consensus/core/src/synchronizer.rs
NaitsabesMue: Thanks. The ratio calculation got lost.

consensus/core/src/synchronizer.rs
NaitsabesMue: done

consensus/core/src/synchronizer.rs
NaitsabesMue: Some artefact. Removed.

consensus/core/src/synchronizer.rs
NaitsabesMue: changed

consensus/core/src/block_manager.rs
```diff
@@ -683,6 +684,16 @@ impl BlockManager {
         self.missing_blocks.clone()
     }
 
+    /// Returns all suspended blocks.
+    pub(crate) fn suspended_blocks(&self) -> BTreeMap<BlockRef, SuspendedBlock> {
+        self.suspended_blocks.clone()
```
piotrm50: Another thing we might change is to not send the full SuspendedBlock but rather only its author as a value in the map.

consensus/core/src/core_thread.rs
```diff
@@ -159,6 +176,20 @@ impl CoreThread {
...
+                        CoreThreadCommand::GetSuspendedCount(sender) => {
+                            let _scope = monitored_scope("CoreThread::loop::get_suspended_count");
+                            let count=self.core.get_suspended_blocks().len();
+                            sender.send(count).ok();
+                        }
+                        CoreThreadCommand::GetMissingAncestorsFor(keys, sender) => {
+                            let _scope = monitored_scope("CoreThread::loop::get_missing_ancestor_for");
+                            // Fetch full map from Core, then filter to requested keys.
+                            let full_map = self.core.get_missing_ancestors();
+                            let filtered = keys.into_iter()
```
piotrm50: Maybe we can put this filter inside if that's the only usage of this method.

consensus/core/src/synchronizer.rs
```diff
@@ -1014,7 +1035,81 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+                    let metrics = &context.metrics.node_metrics;
+                    metrics
+                        .synchronizer_requested_blocks_by_peer
+                        .with_label_values(&[peer_hostname.as_str(), "hot"])
+                        .inc_by(block_refs.len() as u64);
+                    for block_ref in &block_refs {
+                        let block_hostname =
+                            &context.committee.authority(block_ref.author).hostname;
+                        metrics
+                            .synchronizer_requested_blocks_by_authority
```
piotrm50: We should count normal fetches as well, not only the hot ones with the new metric.

consensus/core/src/synchronizer.rs
```diff
@@ -1911,4 +2025,189 @@ mod tests {
...
+        let missing_anc = BTreeMap::from([(missing_ref, suspended_map.keys().cloned().collect())]);
+
+        // 4) Prepare mocks
+        let core_disp = Arc::new(SyncMockDispatcher::default());
+        core_disp.stub_missing_blocks(missing_blocks.clone()).await;
+        core_disp.stub_suspended_blocks_count(0).await;
+        core_disp.stub_missing_ancestors(missing_anc.clone()).await;
+
+        let network = Arc::new(MockNetworkClient::default());
+        // Stub *all* three authorities so none panic:
```
piotrm50:
```suggestion
        // Stub *all* ten authorities so none panic:
```

consensus/core/src/synchronizer.rs
```diff
@@ -1911,4 +2025,189 @@ mod tests {
...
+        // 6) Assert we got exactly two fetches - one hot path and one cold periodic
+        //    path,
+        assert_eq!(results.len(), 2);
+
+        // 7) The hot‐fetch went to peer 2
+        let (_hot_guard, hot_bytes, hot_peer) = &results[0];
+        assert_eq!(*hot_peer, AuthorityIndex::new_for_test(2));
+
+        // 8) The periodic fetch went to peer 1
+        let (_periodic_guard, _periodic_bytes, periodic_peer) = &results[1];
```
piotrm50: Why are two blocks returned if there is only one block in `missing_block` set? If it's a parent of the missing block then why is it returned from a different peer?

piotrm50 (commented):
Let's extend the tests with different cases where we have hot fetches and regular fetched blocks and where we don't have hot fetches at all (where we have both  alot of missing blocks and little missing blocks).

consensus/core/src/synchronizer.rs
NaitsabesMue: There are two requests. One via the hot path and another via the periodic path. In the hot path, we choose from those authorities that have created a suspended block referencing the missing blocks. In the periodic, we choose among all authorities.

consensus/core/src/block_manager.rs
NaitsabesMue: done

consensus/core/src/synchronizer.rs
```diff
@@ -1018,19 +1052,115 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+        #[cfg(not(test))]
+        let random_peers: Vec<AuthorityIndex> = random_candidates
+            .into_iter()
+            .choose_multiple(&mut rng, MAX_RANDOM_PEERS);
 
+        #[cfg_attr(test, allow(unused_mut))]
+        let mut all_missing_blocks: Vec<BlockRef> = missing_blocks.keys().cloned().collect();
+        #[cfg(not(test))]
+        all_missing_blocks.shuffle(&mut rng);
+
```
VorobyevIlya, line 1148: Do I understand correctly that from additional random peers we will ask random blocks from missing_block?
From non-random peers, we ask lexicographically first, and it makes sense for me: we want to fetch blocks from round 1 before round 10.
But these non-random guys may also not know blocks from round 1. So for me it makes more sense to ask for a first chunk here too.

polinikita (approved):
The logic of choosing peers from whom one should fetch the block can be further improved. For instance, one could also take into account from whom we received a block and propogate that knowledge to the block manager. Or the way how blocks are assigned to peers can always follow the rule from smallest-round block to highest round, but now one can leave such improvements for future work.

consensus/core/src/block_manager.rs
```diff
@@ -1108,6 +1151,80 @@ mod tests {
         }
     }
 
+    /// Tests that authorities that know about missing blocks are correctly
```
VorobyevIlya: sentence is not finished

VorobyevIlya (approved):

consensus/core/src/block_manager.rs
NaitsabesMue: finished it

NaitsabesMue:
> Let's extend the tests with different cases where we have hot fetches and regular fetched blocks and where we don't have hot fetches at all (where we have both alot of missing blocks and little missing blocks).

Tests are extended

PR #7385, opened 2025-06-13
refactor(apps-ui-kit): Make react a peer dependency

marc2332:
Currently you cannot use the ui kit/-icons with react v19, this PR makes it possible to do so
https://github.com/iotaledger/iota/pull/7385

In case you ask yourself if this is worth doing:

We do it in the [dapp kit](https://github.com/iotaledger/iota/blob/8cd8501d861e533e5413a2d9eba30554905223f9/sdk/dapp-kit/package.json#L91) and it is very [common](https://github.com/TanStack/query/blob/33d008bbb39f749588ab591d41fefa53f4e18c99/packages/react-query/package.json#L84) in the ecosystem

### DIFF
--- a/apps/ui-kit/package.json
+++ b/apps/ui-kit/package.json
@@ -48,7 +48,6 @@
         "react": "*"
     },
     "devDependencies": {
-        "react": "^18.3.1",
         "@chromatic-com/storybook": "^1.6.0",
         "@storybook/addon-essentials": "^7.1.0",
         "@storybook/addon-interactions": "^7.1.0",
@@ -65,6 +64,8 @@
         "cross-env": "^7.0.3",
         "memfs": "^4.11.1",
         "postcss": "^8.4.31",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
         "storybook": "^7.1.0",
         "storybook-dark-mode": "^4.0.2",
         "tailwindcss": "^3.3.3",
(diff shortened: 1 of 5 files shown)


apps/ui-kit/vite.config.ts
```diff
@@ -12,7 +12,7 @@ const packageConfig = {
         fileName: (format) => `index.${format}.js`,
     },
     rollupOptions: {
-        external: ['react', 'react-dom', 'tailwindcss'],
+        external: ['react', 'react-dom', 'tailwindcss', 'react/jsx-runtime'],
```
marc2332, line 15: This fixes a bug with react and ui kit in another project

"Cannot read properties of undefined (reading 'ReactCurrentDispatcher')"

Most probably it was trying to use a react 18 API, so the fix was to simply make it "dynamic" so that it uses the API from whatever react the end consumer of the library (18 or 19)  is using

apps/ui-kit/package.json
```diff
@@ -42,10 +42,11 @@
         "@radix-ui/react-visually-hidden": "^1.1.0",
         "classnames": "^2.5.1",
         "lodash.merge": "^4.6.2",
-        "react": "^18.3.1",
-        "react-dom": "^18.3.1",
         "react-number-format": "^5.2.2"
     },
+    "peerDependencies": {
+        "react": "^18 || ^19"
```
evavirseda: maybe we should add `react-dom` to `peerDependencies` too? we call `createPortal` (e.g in the Tooltip component), so already depend on react-dom, right?

evavirseda (changes requested):

evavirseda (approved):

msarcev (approved):

PR #7394, opened 2025-06-16
fix(node): prevent discovery private info leak

### LINKED ISSUE #6994: discovery: Add the possibility to minimize leakage of private peer information
In the struct `AllowlistedPeer`, why is the multiaddress optional? If we specify a peer that is allowlisted, then we should already know the peer upfront. Only adding by peer ID isn't a real usecase. In the same change, we can remove the peer ID field, and move the peer ID into the multiaddress as it was back then in hornet.

It also seems that nodes leak information about their own private peers if they are asked to. This is a result of the former behaviour, that a private node can be added where only the peer ID is known, but not the address. Is there really a usecase for this private info leak? There are even tests that test this weird behaviour. We should change that, and never leak any information about our private peers, no matter what.

semenov-vladyslav:
# Description of change

Nodes can advertise themselves as private during peer discovery. Previously, peers would leak info about their private known peers. This PR changes that logic, now, only public known peers are exposed during peer discovery.

Fixes #6994.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Prevent private peer info leaks during peer discovery.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-network/src/discovery/server.rs
+++ b/crates/iota-network/src/discovery/server.rs
@@ -2,12 +2,10 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{
-    collections::HashMap,
-    sync::{Arc, RwLock},
-};
+use std::sync::{Arc, RwLock};
 
 use anemo::{Request, Response};
+use iota_config::p2p::AccessType;
 use rand::seq::IteratorRandom;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


crates/iota-network/src/discovery/server.rs
```diff
@@ -56,54 +54,30 @@ impl Discovery for Server {
...
+        let mut known_not_connected_peers = state
+            .known_peers
+            .iter()
+            .filter_map(|(peer_id, peer_info)| {
+                (peer_info.access_type != AccessType::Private
+                    && !state.connected_peers.contains_key(peer_id))
+                .then_some(peer_info.inner().clone())
+            })
+            .choose_multiple(&mut rng, MAX_PEERS_TO_SEND - known_connected_peers.len());
+        known_connected_peers.append(&mut known_not_connected_peers);
```
muXxer: Should we maybe keep the old name "known_peers" instead? The variable contains connected and unconnected known peers maybe.

muXxer: Should we shuffle the final list, so that our peers can't directly determine to which peers we are connected?

semenov-vladyslav: Right, that wouldn't hurt.

jkrvivian (approved):

muXxer (approved):

PR #7404, opened 2025-06-16
feat(account_abstaction): basic smart account creation flow (PROTOTYPE)

### LINKED ISSUE #7272: [AA-prototype][move] Create a simple dynamic multisig account in Move
Loosely based on gnosis safe, the prototype is called "Dubai chocolate on-chain". 

Image

Implement a Move package with the following features:
- Make it possible to create a `AA Account` shared object type
  - This type should support the receiving of IOTA coins and storing them in a `Balance<IOTA>` field.
  - Through authorized operations should be possible to withdraw from such balance.
  -  The withdraw function can be called only passing in as input an `OwnerCap` capability and thus uses the capability pattern; the `OwnerCap` cap must be owned by a [multisig address](https://docs.iota.org/developer/cryptography/transaction-auth/multisig).
  - This means that before executing all possible operations listed below, a static multisig account must be created (e.g. using the CLI) with at least 2 actors and a threshold of 2 (e.g., Alice and Bob). 
- Make it possible to have an entry_point function where some transaction bytes are proposed. 
  - Such TX bytes can be generated through a CLI and then passed as input; we call this transaction as ProposedTX.
  - The ProposedTX should use a gas coin owned by the user who is generating it, e.g. Alice. This process should involve the use of [sponsorship](https://docs.iota.org/developer/iota-101/transactions/sponsored-transactions/about-sponsored-transactions).
  - The ProposedTX should contain one or more commands in which the `AA Account` shared object balance is withdrawn and transferred to a random address, using the `OwnerCap`.
  -  The entry_point function is invoked with a new transaction called EntryPointTX. 
  -  The entry_point function stores the ProposedTX bytes.
- Make it possible to Alice to reference the stored ProposedTX bytes and create a new transaction called SignatureTX
  - This TX allows Alice to digitally sign the ProposedTX bytes and store the signature
  - The signature is done off-chain and then passed as a parameter, the CLI can be used for this
  - On-chain the signature must be verified, e.g., invoking this https://github.com/iotaledger/iota/blob/c5d81c88c526f6d28d9a111c191ac693d94f7547/crates/iota-framework/packages/iota-framework/sources/crypto/ed25519.move#L13 
- Make it possible to Bob create a SignatureTX as well
- Make it possible to have an on-chain process where:
  1. Alice calls EntryPointTX
  2. Alice calls SignatureTX
  3. Bob calls SignatureTX
  4. The outcome is composed of ProposedTX bytes and the 2 signatures; we call this SignedTransaction; the 2 signature must be done such that [these can be combined off-chain](https://docs.iota.org/developer/cryptography/transaction-auth/multisig#step-7-combine-individual-signatures-into-a-multisig)
 - Make it possible to combine the signatures off-chain and have a SignedTransaction that can be executed like a normal transaction (i.e., submit it to a node for execution)

Dkwcs:
# Description of change
This feature introduces a multisig-controlled smart account on localnet, implementing the complete token withdrawal flow - from initialization to secure execution and balance verification.

It includes - Rust SDK usage + Move package.

Execution Flow
1.  Setup(in current implementation you have to have your local keystone and relevant addresses beforehand, then pass them in code - Alice, Bob, Recipient, Multisig(Alice + Bob)):
 - Load Alice, Bob, and Multisig addresses from local keystore.
 - Define a separate Recipient address to receive withdrawal.
 - Build a MultiSigPublicKey with threshold 2-of-2.

2. Funding
 - Request faucet tokens for: Alice, Bob, Multisig

3. AA Package Deployment
 - Compile the Move package account_abstraction from `../aa_move`.
 - Deploy via `publish_account_abstraction_package`.
 - Extract the resulting package_id.

4. Smart Account Initialization
 - Call `init_smart_account` to deploy an AA-compliant SmartAccount object.
 - Store resulting object IDs: SmartAccount, OwnerCap

5.  Deposit
 - Alice deposits coins to the SmartAccount using make_deposit_to_smart_account.
 
6. Prepare Withdrawal
 - Prepare programmable withdrawal transaction to send 1_000_000 tokens to Recipient.
 - Compute digest and construct TransactionData.
 
7.  ProposeTx Submission
 - Submit proposal to on-chain via `propose_tx_to_smart_account`.
 - Generates ProposedTx object.

8. Signatures Collection
 - Alice signs the proposal using her key; signature stored on-chain.
 - Bob follows, reaching the quorum (2 signatures).

9. Signature Reconstruction & Validation
 - Retrieve on-chain SignedTx object.
 - Restore individual signatures from stored BCS bytes.
 - Combine into valid GenericSignature using MultiSig::combine.
 
10. Execute Final Transaction
 - Call execute_transaction_block with:
 - Combined multisig signature + additional signature (e.g., Alice)

Final check(validates that the recipient account has exactly received the withdrawn amount):

`assert!(check_recipient_balance(&iota_client, coin_recipient_addr, withdraw_amount).await?);
`

Fixes #7272 .

After initial setup, run the command below in `aa_rust_sdk_flow` directory:
`cargo run`


account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+    let uid = object::new(ctx);
+    let id = *uid.as_inner();
+    let smart_account = SmartAccount { id: uid, balance: balance::zero<IOTA>() };
+
+    let owner_cap = OwnerCap { id: object::new(ctx), smart_acc_id: id, owner_addr: multisig_owner };
+    transfer::public_transfer(owner_cap, multisig_owner);
+    transfer::share_object(smart_account);
+}
+
+public fun deposit(smart_account: &mut SmartAccount, coin: Coin<IOTA>) {
```
lzpap: Sending assets to the `SmartAccount` shall happen the same way as it happens to any other address: by `address`, which in this case is the `ObjectID` of the `SmartAccount` object. Therefore, you need to implement a `receive` function on the type which then can deposit into the the `SmartAccount` object.

account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+
+const EInvalidAccountOwner: u64 = 0;
+const EInvalidOwnerAddr: u64 = 1;
+const ENotEnoughBalance: u64 = 2;
+
+public struct OwnerCap has key, store { id: UID, smart_acc_id: ID, owner_addr: address }
+
+public struct SmartAccount has key {
+    id: UID,
+    balance: Balance<IOTA>,
```
lzpap: For this example it is out of scope, but for a general application layer smart account you'd want a heterogenous data structure to hold your assets, for example a `Bag`.

account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+
+// Initializes a shared account and issues OwnerCap to multisig address
+public fun init_multisig_aa(multisig_owner: address, ctx: &mut TxContext) {
+    let uid = object::new(ctx);
+    let id = *uid.as_inner();
+    let smart_account = SmartAccount { id: uid, balance: balance::zero<IOTA>() };
+
+    let owner_cap = OwnerCap { id: object::new(ctx), smart_acc_id: id, owner_addr: multisig_owner };
+    transfer::public_transfer(owner_cap, multisig_owner);
+    transfer::share_object(smart_account);
```
lzpap: How does one delete a smart account from the ledger?

account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+}
+
+public fun deposit(smart_account: &mut SmartAccount, coin: Coin<IOTA>) {
+    coin::put(&mut smart_account.balance, coin);
+}
+
+public fun withdraw(
```
lzpap: It would be a more flexible pattern if `withdraw` returns a coin/balance which then the PTB assembler can use in a transfer/whatever other function as well.

account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+//     transfer::share_object(smart_account);
+// }
+
+// Initializes a shared account and issues OwnerCap to multisig address
+public fun init_multisig_aa(multisig_owner: address, ctx: &mut TxContext) {
+    let uid = object::new(ctx);
+    let id = *uid.as_inner();
+    let smart_account = SmartAccount { id: uid, balance: balance::zero<IOTA>() };
+
+    let owner_cap = OwnerCap { id: object::new(ctx), smart_acc_id: id, owner_addr: multisig_owner };
```
lzpap: If we ever want to add signer rotation as a feature to such smart accounts, it is needed to have some form of information saved inside the account about the current `OwnerCap`, so you can check if the privileged functions are being called with the correct ownercap.

account_abstraction/aa_move/sources/account_abstraction.move
```diff
@@ -0,0 +1,67 @@
...
+//     while (signer_count != owners.length()) {
+//         let owner_cap = OwnerCap { id: object::new(ctx), smart_acc_id: id, owner_addr: owners[signer_count]  };
+//         transfer::public_transfer(owner_cap, owners[signer_count]);
+//         signer_count = signer_count + 1;
+//     };
+//     transfer::share_object(smart_account);
+// }
+
+// Initializes a shared account and issues OwnerCap to multisig address
+public fun init_multisig_aa(multisig_owner: address, ctx: &mut TxContext) {
```
lzpap: Theoretically, we could ask for the pubkeys of the signers here and derive their and the multisig address on-chain too, no?

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,86 @@
...
+    tx_digest: vector<u8>,
+    tx_bytes: vector<u8>,
+    signatures: vector<vector<u8>>,
+    threshold: u64,
+}
+
+// Store raw transaction proposal
+public fun entry_point(
+    smart_account: &mut SmartAccount,
+    proposed_tx_digest: vector<u8>,
```
lzpap, line 40: tx_digest can also be derived on-chain from the tx_bytes, no?

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,86 @@
...
+public struct SignedTx has key, store {
+    id: UID,
+    tx_digest: vector<u8>,
+    tx_bytes: vector<u8>,
+    verified_signatures: vector<vector<u8>>,
+}
+
+/// Holds the raw bytes of the proposed transaction with signatures and their threshold
+public struct ProposedTx has key, store {
+    id: UID,
```
lzpap, line 30: Why do SignedTx and ProposedTx have to be objects? Can't they have simple `store`?

account_abstraction/aa_move/sources/tx_flow.move
Dkwcs: On chain yes, we discussed it with Mirko and  decided to simplify this part in prototype.

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: This function was simplified in order to have fast funding(good point anyway)


```
public fun deposit(smart_account: &mut SmartAccount, coin: Coin<IOTA>) {
    coin::put(&mut smart_account.balance, coin);
}
```

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: There’s no dedicated delete function, but we can achieve it through the usual flow for deleting shared objects - likely with multisignature approval

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: Good point, it can be useful in the real  further implementation

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: If I get you correct , we have a commented field  `//owner_cap: OwnerCap,` for appropriate possible feature.

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: An additional check required, but what is the point?

account_abstraction/aa_move/sources/tx_flow.move
Dkwcs: It’s just an implementation case, it was easier to verify those objects during the flow.
Theoretically we can have these information as a dynamic field, not dynamic object field.

account_abstraction/aa_move/sources/account_abstraction.move
lzpap: I'm not sure I understand what do you mean with "fast funding". The point I'm trying to make is the `SmartAccount` address must behave from the external point of view like a normal address. A normal address receives coins with `public_transfer(address)` and therefore the `SmartAccount` shall do it the same way, so it is not distinguishable and doesn't brake compatibility.

account_abstraction/aa_move/sources/account_abstraction.move
lzpap: It would be nice to see it implemented here ;)

account_abstraction/aa_rust_sdk_flow/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+        "***",
+    )?;
+    let coin_recipient_addr = IotaAddress::from_str(
+        "***",
+    )?;
+
+    // Generate multisig public key
+    let alice_pub_key = keystore.get_key(&alice_addr)?.public();
+    let bob_pub_key = keystore.get_key(&bob_addr)?.public();
+    let multisig_pub_key = MultiSigPublicKey::new(vec![alice_pub_key, bob_pub_key], vec![1, 2], 2)?;
```
lzpap: Imagine Bob and Alice are not using the same computer to store their secrets, but they manage a protocol treasury and don't even know each other. How do they derive the multisig address via SDK? They need some form of medium to mediate it, right?

My previous comment about deriving the multisig address from pubkeys on-chain was about this: why don't we use the chain to mediate this process?

account_abstraction/aa_rust_sdk_flow/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "aa_rust_sdk_flow"
+edition = "2024"
+version.workspace = true
+
+[dependencies]
+iota-sdk = { path = "../../crates/iota-sdk"}
+shared-crypto = { path ="../../crates/shared-crypto"}
+iota-move-build = { path ="../../crates/iota-move-build"}
+tokio = { version = "1.2", features = ["full"] }
+anyhow = "1.0"
```
TheMrAI: Can we set anyhow and tokio to the workspace versions?

account_abstraction/aa_rust_sdk_flow/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+    let transaction_response = iota_client
+        .quorum_driver_api()
+        .execute_transaction_block(
+            Transaction::from_generic_sig_data(tx_data, vec![multisig, alice_signature]),
+            IotaTransactionBlockResponseOptions::full_content(),
+            ExecuteTransactionRequestType::WaitForLocalExecution,
+        )
+        .await?;
+
+    print!("\n Withdraw signature exectuion info: ");
```
TheMrAI: NIT: Small typo.
Perhaps the two prints could be one, but not really important.

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[derive(Deserialize)]
+struct FaucetResponse {
+    task: String,
+    error: Option<String>,
+}
+
+pub async fn request_tokens(client: &IotaClient, address: IotaAddress) -> Result<()> {
+    let address_str = address.to_string();
+    let reqwest_client = Client::new();
```
TheMrAI, line 24: NIT: Wasn't this meant to be `request` instead of `reqwest`?

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
```diff
@@ -0,0 +1,89 @@
...
+                .and_then(|v| v.as_str())
+                .ok_or_else(|| anyhow!("Failed to parse coin ID from faucet response"))?;
+            break id.to_string();
+        }
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    };
+
+    let object_id = IotaObjectDataOptions::new().with_owner();
+    loop {
```
TheMrAI, line 73: Is the sleep loop necessary here?

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,86 @@
...
+    smart_account: &mut SmartAccount,
+    tx_id: ID,
+    pk: vector<u8>,
+    pure_signature: vector<u8>,
+    ctx: &mut TxContext,
+) {
+    let proposed_tx: &mut ProposedTx = dof::borrow_mut(smart_account.id_mut(), tx_id);
+    assert!(ed25519::ed25519_verify(&decode(pure_signature), &pk, &proposed_tx.tx_digest), EInvalidSignature);    
+    
+    proposed_tx.signatures.push_back(pure_signature);
```
TheMrAI: Could I just call sign_proposed_tx with Alice twice and succeed?

account_abstraction/aa_move/sources/tx_flow.move
TheMrAI: Could you write down these prototype shortcuts as comments?
I will definitely forget about the reasons otherwise.

account_abstraction/aa_move/README.md
```diff
@@ -0,0 +1,136 @@
...
+4. Deposit Coins into the Smart Account
+Deposit a coin object into the smart account’s balance:
+    ```bash
+    /Users/pk/iota client call --package $PACKAGE_ID --module account_abstraction --function deposit --args $AA $SOME_COIN_OBJ
+    ```
+
+5. Construct a Withdraw Transaction
+Create a serialized unsigned transaction to withdraw tokens from the smart account (served with Alice’s gas object):
+    ```bash
+    /Users/pk/iota client call --package $PACKAGE_ID --module account_abstraction --function withdraw --args $AA $OWNER_CAP 9999999 $RECIPIENT_ADDR --gas $ALICE_GAS_OBJ --gas-budget 10000000 --serialize-unsigned-transaction
```
lzpap: This command outputs the serialized tx bytes, but not the tx-digest that you need to pass in the following steps. Can you please clarify what is the difference between tx digest (from TransactionData API) and the digest that is being signed? Is the latter only used for sig verification? The one returned in TxEffects if the former, right?

account_abstraction/aa_move/README.md
```diff
@@ -0,0 +1,136 @@
...
+    Export the transaction bytes:
+    ```bash
+    export TX
+    export TX_DIGEST
+    ```
+
+6. Submit the Transaction Proposal
+Create a proposed transaction entry point using the serialized unsigned transaction:
+    ```bash
+    /Users/pk/iota client call --package $PACKAGE_ID --module tx_flow --function entry_point --args $AA $TX_DIGEST $TX 2
```
lzpap: Warning! This should not be executed by Alice, since then she will consume the gas object that she already committed into the proposed TX!

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,317 @@
...
+
+/// Calls the `init_multisig_aa` function to create a SmartAccount object.
+///
+/// # Returns
+/// A response that includes the created objects, including SmartAccount and
+/// OwnerCap.
+pub async fn init_smart_account(
+    iota_client: &IotaClient,
+    package_id: ObjectID,
+    alice_addr: IotaAddress,
```
TheMrAI: Is there a significance for it being `alice`'s address? It seems to be it could be anybody, it is more like an entity that pays.

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,317 @@
...
+    keystore: &FileBasedKeystore,
+) -> Result<IotaTransactionBlockResponse> {
+    let mut ptb_builder = ProgrammableTransactionBuilder::new();
+    let module = Identifier::new(AA_MODULE_NAME)?;
+    let function = Identifier::new("init_multisig_aa")?;
+
+    let arguments = vec![ptb_builder.pure(multisig_addr)?];
+    ptb_builder.command(Command::move_call(
+        package_id,
+        module.clone(),
```
TheMrAI: NIT: No need to clone.

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,317 @@
...
+/// Builds a withdrawal transaction from a SmartAccount,
+/// returning the transaction digest and prepared TransactionData.
+///
+/// This will later be used to propose the transaction and collect signatures.
+pub async fn prepare_withdraw_tx_data(
+    iota_client: &IotaClient,
+    alice_addr: IotaAddress,
+    multisig_addr: IotaAddress,
+    package_id: ObjectID,
+    smart_accoint_obj: ObjectRef,
```
TheMrAI: NIT: smart_account_obj perhaps?

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,317 @@
...
+
+/// Submits a deposit transaction into a SmartAccount using Alice's gas coin.
+pub async fn make_deposit_to_smart_account(
+    iota_client: &IotaClient,
+    keystore: &FileBasedKeystore,
+    alice_addr: IotaAddress,
+    package_id: ObjectID,
+    smart_account_obj: ObjectRef,
+) -> Result<()> {
+    let alice_coin = iota_client
```
TheMrAI: Does your `get_coin` work here as well?

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,86 @@
...
+    );
+    assert!(dof::exists_(smart_account.id_mut(), proposed_tx_id), 10);
+}
+
+// Signatures verification on-chain
+public fun sign_proposed_tx(
+    smart_account: &mut SmartAccount,
+    tx_id: ID,
+    pk: vector<u8>,
+    pure_signature: vector<u8>,
```
lzpap, line 65: It's really not clear to me what all these mean (what is pure sig, what is serialized sig from the offline sig step, etc.). I'd vote for pushing the whole serialized `GenericSignature` on-chain, which is `flag || sig || pk`, and then parse it only for Ed25519, where the first two bytes are `0`. Then you know you have 64 bytes sig and the remaining 32 bytes is the pk.

We could go even as far as doing the multisig building part on-chain if the threshold is met, so at the end of the day you really just need to dump the bytes prepared on-chain into the RPC's submit endpoint.

account_abstraction/aa_rust_sdk_flow/src/utils.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Default gas budget used for all programmable transactions.
+///
+/// This is passed to `TransactionData::new_programmable` to define
+/// how much gas the transaction is allowed to consume.
+pub const GAS_BUDGET: u64 = 10_000_000;
+
+/// Selects the first available coin object for the given address.
+///
+/// This is used to pay for gas or as an input to a transaction.
+pub async fn get_coin(iota_client: &IotaClient, addr: IotaAddress) -> Result<Coin> {
```
TheMrAI, line 35: I wonder if there are internal functions already doing this job.
If no, then you could perhaps make this more general by returning the iterator of the coin page in the result.
Then you could decide on the call-site if you want just one, or more.

TheMrAI (approved):
Looks good to me, nicely written code. If just 30% of our codebase had this quality I would be a happy gumba!

TheMrAI (commented):
There is just one thing I suppose I don't understand, and this is not directed at you necessarily Pavlo.
Why do all this hupla? 

In this setup, for such a multisig based AA account, there are at least 4 Tx that need to be executed to arrive at the desired result.
Send the proposal in, sign it by enough parties and then execute it.

Does this logic need to be implemented on chain?
Can't we just require Wallets or other tools to sign the propossed Tx appropriately and then send it in, AA checks if all good and executes it?

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
Dkwcs: It’s a reqwest::async_impl::client(Furthermore I copied that code from sdk utils). I guess it was written on purpose.

account_abstraction/aa_rust_sdk_flow/src/utils.rs
Dkwcs: It’s just a minor helper function in order to avoid chain of api calls duplication everywhere, it can be optimized further.

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
Dkwcs: I have to check it, not mine code.

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: Done

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: It’s mistype, yes

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: Yes

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: Yes, you’re right, it was like strict example of sender, but yeah, it can be anybody.

account_abstraction/aa_move/sources/tx_flow.move
Dkwcs: Yes, but it will fail later. But good point, I’ve fixed it.

account_abstraction/aa_rust_sdk_flow/src/main.rs
Dkwcs: Fixed

account_abstraction/aa_rust_sdk_flow/Cargo.toml
Dkwcs: I made it all based on workspace.

account_abstraction/aa_move/README.md
lzpap, line 1: This README doesn't really work. I tried to follow the steps, but at around step 5 you fail miserably due to not having the correct outputs from previous commands (TX_DIGEST) which are needed for the next ones. Also, in step 8, the expected format of the pubkeys and signatures can only be constructed in code, and even if you do that it is not really possible to pass them in via the CLI.

Therefore please revise this README and drop the CLI support altogether if the primary way is via the SDK. I managed to run the demo succesfully on the devnet after some trial and errors.

account_abstraction/aa_rust_sdk_flow/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    // Hardcoded user and contract addresses (you have to pass your own addresses)
+    let multisig_addr = IotaAddress::from_str("***")?;
+    let alice_addr = IotaAddress::from_str("***")?;
+    let bob_addr = IotaAddress::from_str("***")?;
+    let coin_recipient_addr = IotaAddress::from_str("***")?;
+
+    // Generate multisig public key
+    let alice_pub_key = keystore.get_key(&alice_addr)?.public();
+    let bob_pub_key = keystore.get_key(&bob_addr)?.public();
+    let multisig_pub_key = MultiSigPublicKey::new(vec![alice_pub_key, bob_pub_key], vec![1, 2], 2)?;
```
lzpap: `vec![1, 2], 2` means that bob has weight 2, and threshold is 2, so what's the point of Alice even being there? I'd rather use weights [1 , 1] and keep the threshold at 2.

lzpap (changes requested):

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
Thoralf-M: Good old https://crates.io/crates/reqwest

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: I see what you mean, but could you clarify do we need "receive" functionality here or we can just remove `deposit` function and use `public_transfer` to instead?

lzpap: If you send to an ObjectID via `public_transfer()` then it needs to be received via the object's `&mut id`, which only the "owner" should have access to. Take a look at [this example from the docs](https://docs.iota.org/developer/iota-101/objects/transfers/transfer-to-object#receiving-objects-and-adding-them-as-dynamic-fields). Instead of checking for an address, we should require the ownercap for the `accept_payment()` function.

account_abstraction/aa_move/sources/account_abstraction.move
lzpap: what we need to save inside the field is the ID of the current owner cap, so we can check in function invocations whether the correct cap is being passed or not.

account_abstraction/aa_rust_sdk_flow/src/main.rs
Dkwcs: Do you mean initializing    
` let multisig_pub_key = MultiSigPublicKey::new(vec![alice_pub_key, bob_pub_key], vec![1, 2], 2)?;
`
on-chain? 
As a prototype it doesn’t have a full-fledged implementation of all details.
We can spend some time on it if needs

account_abstraction/aa_move/README.md
Dkwcs: That’s why a left the warning. I’ll think how can I change the CLI README
```
The CLI doesn’t support transaction sponsorship.
  Additionally, it would require significant effort to handle encoding and decoding of signatures, public keys, and transaction bytes. 
  So, the following steps outline the general flow to provide an overview of the process
```

account_abstraction/aa_rust_sdk_flow/src/main.rs
Dkwcs: With 1, 1 the error is occurred 
`Invalid user signature`
Trying to figure it out

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: Done

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: Implemented

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: Done

account_abstraction/aa_move/sources/account_abstraction.move
Dkwcs: Implemented

account_abstraction/aa_rust_sdk_flow/src/faucet.rs
TheMrAI: Oh okay. 👍

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,94 @@
...
+) {
+    let proposed_tx: &mut ProposedTx = dof::borrow_mut(smart_account.id_mut(), tx_id);
+    assert!(
+        ed25519::ed25519_verify(&decode(pure_signature), &pk, &proposed_tx.tx_digest),
+        EInvalidSignature,
+    );
+
+    proposed_tx.signatures.insert(pure_signature);
+
+    if (proposed_tx.threshold == proposed_tx.signatures.size()) {
```
lzpap: this check is not valid: each address can have different weights, therefore the amount of signatures does not equal the sum of the weights.

In the example that you did (aa_rust_sdk_flow), a multisig address is generated out of 2 addresses with weights being 1 and 2, while the threshold is 2. The second address then could authorize a withdraw, however this check makes it so that it can't without the first address.

This is another reason why I think it would make sense to derive the multisig address on-chain: you know the public keys, addresses and their weights directly on-chain, so you can do the proper threshold check here.

account_abstraction/aa_rust_sdk_flow/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+mod tx_flow;
+mod utils;
+
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let iota_client = IotaClientBuilder::default().build_localnet().await?;
+
+    // Path to local keystore(use your own)
+    // For instance - /Users/user/.iota/iota_config/iota.keystore
+    let keystore = FileBasedKeystore::new(&PathBuf::from("***"))?;
```
lzpap: To make it easier to run this example, it would be better to define `keystore` as `InMemoryKeystore`  and initialize addresses (Alice, Bob) from secrets we hardcode here, they are for experimenting anyway.

lzpap (approved):

lzpap (approved):
There is a small bug with the multisig weights when checking signatures on-chain, but we can leave it as it is fo r this proof-of-concept.

One thing however that needs changing is the base branch: please create a feature branch for AA as these kinds of prototypes don't belong to `develop`.

account_abstraction/aa_rust_sdk_flow/src/main.rs
miker83z: [dd41fdf](https://github.com/iotaledger/iota/pull/7404/commits/dd41fdf13deffbbc9206158a0da55d2a396199dc)

account_abstraction/aa_move/sources/smart_account.move
```diff
@@ -0,0 +1,87 @@
...
+    let uid = object::new(ctx);
+    let id = *uid.as_inner();
+    let owner_cap = OwnerCap { id: uid, owner_addr: multisig_owner };
+
+    let smart_account = SmartAccount {
+        id: object::new(ctx),
+        balance: balance::zero<IOTA>(),
+        owner_cap_id: id,
+    };
+
```
miker83z, line 52: Not to be implemented for this prototype, but for the next time: every object created should have an event emit, e.g., `SmartAccountCreated` event or `OwnerCapCreated`; same for object destroy.

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,451 @@
...
+        .ok_or_else(|| anyhow!("No object changes found"))?;
+    let find_ref = |name: &str| -> Result<ObjectRef> {
+        let id = Identifier::new(name)?;
+        created
+            .iter()
+            .find_map(|obj_change| match obj_change {
+                ObjectChange::Created {
+                    object_type: StructTag { name: n, .. },
+                    ..
+                } if n == &id => Some(obj_change.object_ref()),
```
miker83z: nit: a better approach would be to check address, module and name, so that you are sure you are getting the unique id; e.g., with this code you would find all objects like 0xabc::def::SmartAccount, 0x123::ghi::SmartAccount, ...

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
```diff
@@ -0,0 +1,451 @@
...
+    )?;
+
+    let transaction_response = iota_client
+        .quorum_driver_api()
+        .execute_transaction_block(
+            Transaction::from_generic_sig_data(receive_tx_data, vec![multisig, depositor_sig]),
+            IotaTransactionBlockResponseOptions::full_content(),
+            ExecuteTransactionRequestType::WaitForLocalExecution,
+        )
+        .await?;
```
miker83z: It's not needed to use the multi sig here, anyone could call `receive_deposit`, so let's reuse the simple flow of the `depositor_addr`
```suggestion
    let receive_tx_data = TransactionData::new_programmable(
        depositor_addr,
        vec![depositor_gas_coin.object_ref()],
        ptb_builder.finish(),
        GAS_BUDGET,
        gas_price,
    );

    let depositor_sig: GenericSignature = keystore
        .sign_secure(
            &depositor_addr,
            &receive_tx_data,
            Intent::iota_transaction(),
        )?
        .into();

    let transaction_response = iota_client
        .quorum_driver_api()
        .execute_transaction_block(
            Transaction::from_generic_sig_data(receive_tx_data, vec![depositor_sig]),
            IotaTransactionBlockResponseOptions::full_content(),
            ExecuteTransactionRequestType::WaitForLocalExecution,
        )
        .await?;
```

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,94 @@
...
+public fun entry_point(
+    smart_account: &mut SmartAccount,
+    proposed_tx_digest: vector<u8>,
+    proposed_tx_bytes: vector<u8>,
+    threshold: u64,
+    ctx: &mut TxContext,
+) {
+    let proposed_tx = ProposedTx {
+        id: object::new(ctx),
+        tx_digest: proposed_tx_digest,
```
miker83z: @Dkwcs please, write down a comment saying that this digest could be also derived on-chain as suggested by @TheMrAI

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -0,0 +1,94 @@
...
+        signatures: vec_set::empty(),
+        threshold,
+    };
+    let proposed_tx_id = *proposed_tx.id.as_inner();
+    dof::add(
+        smart_account.id_mut(),
+        proposed_tx_id,
+        proposed_tx,
+    );
+    assert!(dof::exists_(smart_account.id_mut(), proposed_tx_id), 10);
```
miker83z: I don't think this assert is needed

account_abstraction/aa_rust_sdk_flow/src/sig_utils.rs
```diff
@@ -0,0 +1,102 @@
...
+/// # Returns
+/// A `String` containing the 128-character hex representation of the pure
+/// signature bytes.
+///
+/// # Notes
+/// This function assumes the Ed25519 scheme (flag byte 0x00) and extracts only
+/// the signature.
+pub fn extract_pure_signature(signature: &GenericSignature) -> String {
+    let flag_prefix = 2;
+    let pure_signature_length = 128;
```
miker83z: This would make it more explicit (Ed25519Signature is found in the fastcrypto lib): 
```suggestion
pub fn extract_pure_signature_ed25519(signature: &GenericSignature) -> String {
    let flag_prefix_length = 1 /* FLAG_LENGTH */ * 2;
    let pure_signature_length = Ed25519Signature::LENGTH * 2;
```

account_abstraction/aa_rust_sdk_flow/src/tx_flow.rs
```diff
@@ -0,0 +1,178 @@
...
+    withdraw_tx_data: &TransactionData,
+    keystore: &K,
+    addr: IotaAddress,
+    package_id: ObjectID,
+    smart_account_object: ObjectRef,
+) -> Result<IotaTransactionBlockResponse> {
+    let signature: GenericSignature = keystore
+        .sign_secure(&addr, &withdraw_tx_data, Intent::iota_transaction())?
+        .into();
+    let pub_key = keystore.get_key(&addr)?.public();
```
miker83z, line 147: You could get the public key also through `extract_pure_signature` (the last 32 bytes/64 chars). This would make clearer that the signature and pub keys are related in this example.

account_abstraction/aa_move/sources/tx_flow.move
miker83z: Reply for this comment and the one just above: 
We know it's feasible; however, I argue implementing the multisig combine/building on-chain would require too much time for the sake of this prototype (both for the implementation and learning the related cryptography). 
I propose to represent just the `MultiSigPublicKey` type in Move, that allows us to check the weights correctly.

miker83z (changes requested):
Very good prototype, left a couple of small change requests!

account_abstraction/aa_move/sources/tx_flow.move
miker83z: #7616

account_abstraction/aa_rust_sdk_flow/src/main.rs
miker83z: Fixed in #7616

account_abstraction/aa_move/README.md
miker83z: @Dkwcs could you remove the readme for the CLI and create a new one within the `aa_rust_sdk_flow` with a text similar to the description of this PR?

account_abstraction/aa_move/sources/tx_flow.move
miker83z: Continues in https://github.com/iotaledger/iota/pull/7404#discussion_r2174559821

account_abstraction/aa_rust_sdk_flow/src/main.rs
miker83z: Continues in https://github.com/iotaledger/iota/pull/7404#discussion_r2174559821

account_abstraction/aa_move/README.md
miker83z: This command returns some TX BCS bytes. If you use this bytes in the sign command, e.g.: 
```
 iota keytool sign --address [your-address] --data AAABACAxKyJs2YOcGjXTWZqhTUmuKgXYZRQ7ag8ClsYsaEzVBwEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEFZGVidWcFcHJpbnQBBAEBAACP9QeeFX/PDHAYKTCf66BmtHK4EWrBoXeMy+9sy/bkkgEBb3Ym6GYNhrF4eTRFwareu2O3BD39PcHV6SCRfx+YMgMAAAAAAAAAILNSBSd5/GQKB84zuVJkcf/2A/vA85uW3Wpq5CoOC+HXj/UHnhV/zwxwGCkwn+ugZrRyuBFqwaF3jMvvbMv25JLoAwAAAAAAADB6LQAAAAAAAA==
```
you will obtain something like this as output
```
╭───────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ iotaAddress   │ 0x8ff5079e157fcf0c701829309feba066b472b8116ac1a1778ccbef6ccbf6e492                                                                                               │
│ rawTxData     │ AAABACAxKyJs2YOcGjXTWZqhTUmuKgXYZRQ7ag8ClsYsaEzVBwEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEFZGVidWcFcHJpbnQBBAEBAACP9QeeFX/PDHAYKTCf66BmtHK4EWrBoXeMy+9sy/bk │
│               │ kgEBb3Ym6GYNhrF4eTRFwareu2O3BD39PcHV6SCRfx+YMgMAAAAAAAAAILNSBSd5/GQKB84zuVJkcf/2A/vA85uW3Wpq5CoOC+HXj/UHnhV/zwxwGCkwn+ugZrRyuBFqwaF3jMvvbMv25JLoAwAAAAAAADB6LQAA │
│               │ AAAAAA==                                                                                                                                                         │
│ intent        │ ╭─────────┬─────╮                                                                                                                                                │
│               │ │ scope   │  0  │                                                                                                                                                │
│               │ │ version │  0  │                                                                                                                                                │
│               │ │ app_id  │  0  │                                                                                                                                                │
│               │ ╰─────────┴─────╯                                                                                                                                                │
│ rawIntentMsg  │ AAAAAAABACAxKyJs2YOcGjXTWZqhTUmuKgXYZRQ7ag8ClsYsaEzVBwEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEFZGVidWcFcHJpbnQBBAEBAACP9QeeFX/PDHAYKTCf66BmtHK4EWrBoXeMy+9s │
│               │ y/bkkgEBb3Ym6GYNhrF4eTRFwareu2O3BD39PcHV6SCRfx+YMgMAAAAAAAAAILNSBSd5/GQKB84zuVJkcf/2A/vA85uW3Wpq5CoOC+HXj/UHnhV/zwxwGCkwn+ugZrRyuBFqwaF3jMvvbMv25JLoAwAAAAAAADB6 │
│               │ LQAAAAAAAA==                                                                                                                                                     │
│ digest        │ q03rpf0jGIWF/5d83M0KtfMsR2oericDz567IcfDOwA=                                                                                                                     │
│ iotaSignature │ AJaQV9aRpXgiF0+qLkrg2uEjc9SO2pMpDrensePU7wOqj+4WXTYR3p7JsRROVGFXX/BaAfSa+4ymLH3z56LsfAbcOknsLT4uB4TwgYPrxTAj/5z6Ipl+OEbh3CmJc5QAMQ==                             │
╰───────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

So the signature is not done on the TxDigest we are used to, but on the `rawIntentMsg`'s digest, obtained appending the `rawTxData` to the intent bytes. Finally, we will need this digest in order to verify the signature in Move.

account_abstraction/aa_move/sources/account_abstraction.move
miker83z: Continues in https://github.com/iotaledger/iota/pull/7404#discussion_r2174559821

account_abstraction/aa_rust_sdk_flow/src/tx_flow.rs
Dkwcs: You mean by returning a tuple?

account_abstraction/aa_move/README.md
Dkwcs: Done

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: Great point

account_abstraction/aa_rust_sdk_flow/src/smart_account.rs
Dkwcs: Done

account_abstraction/aa_move/sources/tx_flow.move
Dkwcs: Done

account_abstraction/aa_move/sources/tx_flow.move
Dkwcs: Done

account_abstraction/aa_rust_sdk_flow/src/sig_utils.rs
Dkwcs: Done

account_abstraction/aa_rust_sdk_flow/src/tx_flow.rs
miker83z: Yes, but not needed to be implemented rn.

miker83z (approved):

PR #7405, opened 2025-06-16
fix(ci): `external-crates` tests don't run on the CI

valeriyr:
# Description of change

The `external-crates` tests should run on the CI.


external-crates/move/crates/move-cli/tests/build_tests/migration_names/args.exp
```diff
@@ -92,9 +92,9 @@ Wrote patchfile out to: ./migration.patch
 
 Recorded edition in 'Move.toml'
 External Command `diff -r -s sources migration_sources`:
+Files sources/mod0.move and migration_sources/mod0.move are identical
 Files sources/mod_intermodule_errors.move and migration_sources/mod_intermodule_errors.move are identical
-Files sources/mod_let_errors.move and migration_sources/mod_let_errors.move are identical
 Files sources/mod_let.move and migration_sources/mod_let.move are identical
-Files sources/mod0.move and migration_sources/mod0.move are identical
+Files sources/mod_let_errors.move and migration_sources/mod_let_errors.move are identical
```
valeriyr, line 98: This file was changed because it was updated by mistake.
The problem was in a custom version of the `diff` tool installed using Homebrew.
We expect that `diff` compares files in lexicographical order, but my tool did it differently.

scripts/ci_tests/rust_tests.sh
```diff
@@ -465,9 +474,7 @@ function filter_and_run_tests() {
 
         # first run tests for external crates (they are not part of the workspace)
         if [ "$test_type" == $TEST_TYPE_NEXTEST ]; then
-            run_cargo_nextest "$combined_set_external" ".config/nextest_external.toml" "external-crates/move/Cargo.toml"
-        elif [ "$test_type" == $TEST_TYPE_SIMTEST ]; then
-            run_cargo_simtest "$combined_set_external" "external-crates/move/Cargo.toml"
```
valeriyr, line 470: The `external-crates` workspace can't be tested using the `simtest` tool because it doesn't contain the `simulator` profile. I don't think that it makes sense to add such a profile, so I removed the `run_cargo_simtest` command.

scripts/ci_tests/rust_tests.sh
```diff
@@ -465,9 +474,7 @@ function filter_and_run_tests() {
...
+            run_cargo_nextest "$combined_set_external" ".config/nextest_external.toml" "external-crates/move/Cargo.toml" "tracing"
```
valeriyr, line 477: All the changes related to `feature_set` were done for investigation.
There is a problem if we run the `external-crates` tests with the `--all-features` flag because in this case, we enable the `tiered-gas` flag that changes the costing schedule for bytecode instructions used by the tests. This leads to different results in some tests, such as `move-cli::tracing_testsuite`, for example.

A question here, if it is possible to enable all the features except `tiered-gas` without explicitly listing them in the script?

.github/workflows/_rust_tests.yml
```diff
@@ -153,7 +153,7 @@ jobs:
         # "RESTART_POSTGRES" are set in the env variables above
         run: |
           export CI_IS_RUST=${{ inputs.isRust }}
-          export CI_IS_EXTERNAL_CRATES=${{ inputs.inputs.isRust }}
+          export CI_IS_EXTERNAL_CRATES=${{ inputs.isRust }}
```
valeriyr: If we don't run simulation tests for external crates, should we remove `CI_IS_EXTERNAL_CRATES` and `CI_CHANGED_EXTERNAL_CRATES` from here?

external-crates/move/crates/move-cli/tests/build_tests/migration_names/args.exp
TheMrAI: This is a bit annoying.
Perhaps we can update such comparisons to use insta?

valeriyr: Probably in a separate issue. Maybe it is possible to change the command to make it work with different versions:
```
diff -r -s sources migration_sources
```

TheMrAI (approved):
LGTM

.github/workflows/_rust_tests.yml
muXxer: yes, you can remove them in this case.

muXxer (approved):

TheMrAI (approved):

scripts/ci_tests/rust_tests.sh
DaughterOfMars: It is unfortunately not possible

muXxer (approved):

thibault-martinez (approved):

PR #7421, opened 2025-06-17
feat(starfish): read headers to send with a block

### LINKED ISSUE #7368: reading additional block headers from the DAGState and marking them as known by the authority.
Send required headers based on our perception of what the peers know.

VorobyevIlya:
# Description of change

The PR modifies block streaming to include headers unknown to the peer.

#7368 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -313,8 +313,8 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
 
     async fn handle_subscribe_block_bundles_request(
         &self,
-        _peer: AuthorityIndex,
-        _last_received: Round,
+        peer: AuthorityIndex,
+        last_received: Round,
     ) -> ConsensusResult<BlockBundleStream> {
         fail_point_async!("consensus-rpc-response");
 
@@ -343,11 +343,11 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         // TODO::deal with possible error in try_from
         Ok(Box::pin(missed_blocks.chain({
             let dag_state = Arc::clone(&self.dag_state);
-            let peer = peer;
 
             broadcasted_blocks.map(move |block| {
                 let mut dag_state_guard = dag_state.write();
-                let block_headers = dag_state_guard.take_unknown_headers_for_authority(peer);
+                let block_headers =
+                    dag_state_guard.take_unknown_headers_for_authority(peer, block.round());
                 let block_bundle = BlockBundle {
                     verified_block: block,
                     verified_headers: block_headers,
@@ -702,7 +702,6 @@ impl SubscriptionCounter {
 /// It yields blocks that are broadcasted after the stream is created.
 type BroadcastedBlockStream = BroadcastStream<VerifiedBlock>;
 
-type BroadcastedBlockBundleStream = BroadcastStream<BlockBundle>;
 /// Adapted from `tokio_stream::wrappers::BroadcastStream`. The main difference
 /// is that this tolerates lags with only logging, without yielding errors.
 struct BroadcastStream<T> {
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -28,8 +28,8 @@ use crate::{
     dag_state::DagState,
     error::{ConsensusError, ConsensusResult},
     network::{
-        BlockBundleStream, BlockStream, NetworkService, SerializedBlock, SerializedBlockBundle,
-        SerializedHeaderAndTransactions,
+        BlockBundle, BlockBundleStream, BlockStream, NetworkService, SerializedBlock,
+        SerializedBlockBundle, SerializedHeaderAndTransactions,
     },
     stake_aggregator::{QuorumThreshold, StakeAggregator},
     storage::Store,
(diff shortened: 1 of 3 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -316,7 +316,45 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        // Return a stream of blocks that first yields missed blocks as requested, then
+        // new blocks.
+        // TODO::deal with possible error in try_from
+        Ok(Box::pin(missed_blocks.chain({
+            let dag_state = Arc::clone(&self.dag_state);
+            let peer = peer;
+
+            broadcasted_blocks.map(move |block| {
+                let mut dag_state_guard = dag_state.write();
+                let block_headers = dag_state_guard.take_unknown_headers_for_authority(peer);
```
polinikita: I am not sure that this is the right way to do. In the testbed, I took the causal past of a given block or at least all rounds below certain round. Here, we might also send blocks from the same round or even future rounds. Should we also feed also `block.round` to the function take_unknown_headers_for_authority()?

The current approach might be not a problem in general and might be even more pragmatic, but it is a bit unclear to me how it will work

polinikita (commented):
Overall, it looks good for me, but I would also use a given round as a parameter when taking unknown blocks for a given peer.

crates/starfish/core/src/authority_service.rs
VorobyevIlya: I added the block round as a parameter, and now the function takes only headers from previous rounds.

polinikita (approved):

crates/starfish/core/src/authority_service.rs
```diff
@@ -313,10 +313,48 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
     ) -> ConsensusResult<BlockBundleStream> {
-        unimplemented!("Unimplemented")
+        fail_point_async!("consensus-rpc-response");
+
+        let dag_state = self.dag_state.read();
+        // Find recent own blocks that have not been received by the peer.
+        // If last_received is a valid and more blocks have been proposed since then,
+        // this call is guaranteed to return at least some recent blocks, which
+        // will help with liveness.
+        // TODO:: do we need to add some headers here?
```
piotrm50, line 326: about TODO: I think it's better to stream those blocks separately to avoid creating a huge batch with many blocks in the beginning. Is it a problem to send multiple types through the stream?

crates/starfish/core/src/dag_state.rs
```diff
@@ -1120,6 +1122,47 @@ impl DagState {
...
+            );
+            let nth_element = set
+                .iter()
+                .nth(MAX_HEADERS_PER_BUNDLE)
+                .map_or(round_bound, |x| *x);
+            min(nth_element, round_bound)
+        };
+
+        let remaining = set.split_off(&split_point);
+        let block_refs: Vec<BlockRef> = set.into_iter().collect();
```
piotrm50: I think we can skip this iteration and create a vector in the for loop below.

crates/starfish/core/src/network/mod.rs
```diff
@@ -286,7 +286,84 @@ impl TryFrom<SerializedBlock> for SerializedHeaderAndTransactions {
...
+    fn try_from(serialized_block_and_headers: SerializedBlockAndHeaders) -> ConsensusResult<Self> {
+        let bytes = bcs::to_bytes(&serialized_block_and_headers)
+            .map_err(ConsensusError::SerializationFailure)?;
+        Ok(Self {
+            serialized_block_bundle: Bytes::from(bytes),
+        })
+    }
+}
+
+impl TryFrom<VerifiedBlock> for SerializedBlockBundle {
```
piotrm50, line 359: I don't think we should be creating VerifiedBlock without actually verifying the contents of the header and transactions. In `handle_subscribed_block` method we receive `SerializedBlock` instance and only create `VerifiedBlock` after successful verification. On the serialization part it's fine, but we cannot optimistically assume that what we receive is valid and correct.

piotrm50 (changes requested):

crates/starfish/core/src/dag_state.rs
VorobyevIlya: fixed

crates/starfish/core/src/authority_service.rs
VorobyevIlya: now we stream block bundles with empty vector of additional headers. So I guess there is some small overhead because of this bundle wrapper. Do you think it is better to stream serialization of some enum that can be block bundle or simple block?

crates/starfish/core/src/network/mod.rs
VorobyevIlya: Yes, but here we create serialization from verified block. We create a block from serialization in impl TryFrom<SerializedHeaderAndTransactions> for VerifiedBlock { in block_header.rs. It is used in fetching blocks in commit syncer and reading from the database. Check is not implemented, but there is a TODO there. We definitely should do verification during fetching, but I'm not sure about the database. I think it will be better if I do it in a branch about commit syncer

crates/starfish/core/src/authority_service.rs
piotrm50: ah, then it's fine I gues. I thought we're streaming only old headers - which I guess would make some sense as well as to avoid committing old data that is beyond acknowledge threshold?

crates/starfish/core/src/network/mod.rs
piotrm50: ah, ok. My bad, understood it the wrong way

piotrm50 (approved):

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Maybe, let's discuss it later. The change we need to make to support this is not small

PR #7424, opened 2025-06-17
refactor(iota-data-ingestion-core): separate `reader` & `executor` in `v1` modules & add `CheckppintReader` V2

### LINKED ISSUE #5811: [data-ingestion-core]: Read checkpoints from historical and live object store
## Description

Following #5808 and #5809 we need to provide a new `CheckpointReader` that fetches live data for the current epoch, and historical data for older epochs.

sergiupopescu199:
# Description of change

This PR introduces an alternative `CheckpointReader` that supports reading from hybrid historical sources. It lays the groundwork for separating the current implementations of `CheckpointReader` into versioned modules (`reader::v1`). The new `CheckpointReader` will reside in `reader::v2`, and the updated `IndexerExecutor` will introduce a new `run_with_config()` method which internally will use the new `CheckpointReader` implementation.

The new implementation adds the ability to read checkpoints from hybrid historical storage and, as an optimization, downloads checkpoints in batches. This minimizes redundant downloads, since a single `.chk` file can contain thousands of checkpoints. The live storage is now an optional component that can be declared alongside the historical store, reflecting its optional role by design.

While most of the logic in the new `CheckpointReader` remains similar to the previous implementation, the main difference lies in how checkpoint downloads are handled. More information can be found in [this comment](https://github.com/iotaledger/iota/pull/7425#issuecomment-2996622634).

fixes #5811 

- This includes tests and doctests
```shell
```
- refactored the ingestion tests to use the new IndexerExecutor, all tests passed.
- refactored the `examples/custom-indexer/rust/local_reader.rs` to refactor the `IndexerExecutor` to run the alternative sync process using the new `run_with_config` method   which internally uses the new `reader::v2::CheckpointReader`, checked the correctness and the behavior.

```rust
use std::sync::Arc;

use anyhow::Result;
use async_trait::async_trait;
use iota_data_ingestion_core::{
    DataIngestionMetrics, IndexerExecutor, ReaderOptions, ShimProgressStore, Worker, WorkerPool,
    reader::v2::{CheckpointReaderConfig, RemoteUrl},
};
use iota_types::full_checkpoint_content::CheckpointData;
use prometheus::Registry;

struct CustomWorker;

#[async_trait]
impl Worker for CustomWorker {
    type Message = ();
    type Error = anyhow::Error;

        // custom processing logic
        // print out the checkpoint number
        println!(
            "Processing checkpoint: {}",
            checkpoint.checkpoint_summary.to_string()
        );
        Ok(())
    }
}

#[tokio::main]
async fn main() -> Result<()> {
    tracing_subscriber::fmt::init();
    let config = CheckpointReaderConfig {
        remote_store_url: Some(RemoteUrl::HybridHistoricalStore {
            historical_url: "https://checkpoints.testnet.iota.cafe".into(),
            live_url: Some("https://checkpoints.testnet.iota.cafe/ingestion/live".into()),
        }),
        reader_options: ReaderOptions::default(),
        ..Default::default()
    };
    let metrics = DataIngestionMetrics::new(&Registry::new());
    let progress_store = ShimProgressStore(0);
    let token = CancellationToken::new();
    let mut executor = IndexerExecutor::new(progress_store, 1, metrics, token.child_token());
    let worker_pool = WorkerPool::new(CustomWorker, "workflow".to_string(), 5, Default::default());
    executor.register(worker_pool).await?;
    executor.run_with_config(config).await?;

    Ok(())
}
```

-  Verified that the move to `v1` didn't break any existing code used by other crates.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-data-ingestion-core/src/reader/fetch.rs
```diff
@@ -0,0 +1,177 @@
...
+/// This function scans the specified path for checkpoint files whose sequence
+/// number is greater than or equal to `current_checkpoint_number`. It attempts
+/// to deserialize each file into a [`CheckpointData`], collecting them
+/// into a vector. The process stops early if the provided `exceeds_capacity`
+/// function returns `true` for a checkpoint's sequence number,
+/// or when `MAX_CHECKPOINTS_IN_PROGRESS` files have been processed.
+pub async fn read_local_files_with_capacity_check<F>(
+    path: &Path,
+    current_checkpoint_number: CheckpointSequenceNumber,
+    exceeds_capacity: F,
```
kodemartin: Instead of passing a closure as an argument, you could define a trait with a method `exceeds_capacity` that implementors will have to implement and default implementations for the rest top-level functions here.

crates/iota-data-ingestion-core/src/reader/fetch.rs
kodemartin, line 1: Consider defining a trait for the common interface, with default implementaitons where applicable.

crates/iota-data-ingestion-core/src/executor/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) mod v1;
```
kodemartin: All `pub` components of the curernt `v1` that was also `pub` before, should be exposed to maintain backward compatibility.

crates/iota-data-ingestion-core/src/reader/mod.rs
kodemartin, line 1: All `pub` components of current `v1`, `fetch` that were public before the rename, should be exposed here with `pub use` for backward compatibility.

crates/iota-data-ingestion-core/src/tests.rs
kodemartin, line 1: Why are these changes necessary?

kodemartin (changes requested):

crates/iota-data-ingestion-core/src/tests.rs
sergiupopescu199: With the [latest refactor of `FileProgresStore`](https://github.com/iotaledger/iota/pull/6918) to make the updates atomic to have a more robust implementation for every update we run `sync_data()`, this adds a bit of latency ([link to PR conversation](https://github.com/iotaledger/iota/pull/6918#discussion_r2100247263)). 

At that time I ran the tests several times and they never failed, this time around 1 out of 10 times some tests cases failed (the ones which tested the checkpoint processing flow). I've increased the time to issues a shutdown signal to allow the FileProgressStore to save the progress state of the processed checkpoints.

crates/iota-data-ingestion-core/src/executor/mod.rs
sergiupopescu199: In `lib.rs`, the necessary components are re-exported to maintain backward compatibility. Although the `v1` module is marked as `pub(crate)`, its public items (such as `IndexerExecutor` and `setup_single_workflow`) are re-exported at the crate root (`lib.rs`) via `pub use`. This ensures that users can continue to access these components as before (e.g., `iota_data_ingestion_core::IndexerExecutor`), regardless of the internal module structure.

crates/iota-data-ingestion-core/src/reader/mod.rs
sergiupopescu199: Very similar to previous [comment](https://github.com/iotaledger/iota/pull/7424#discussion_r2154247105)

crates/iota-data-ingestion-core/src/reader/fetch.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7424/commits/8054b582dad7d2c4955a2efdbac81eba3411abf7

crates/iota-data-ingestion-core/src/reader/fetch.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7424/commits/8054b582dad7d2c4955a2efdbac81eba3411abf7

crates/iota-data-ingestion-core/src/reader/fetch.rs
```diff
@@ -0,0 +1,195 @@
...
+use object_store::{ObjectStore, path::Path as ObjectStorePath};
+use tokio::sync::mpsc;
+use tracing::{debug, info};
+
+use crate::{IngestionError, IngestionResult, MAX_CHECKPOINTS_IN_PROGRESS};
+
+pub type CheckpointResult = IngestionResult<(Arc<CheckpointData>, usize)>;
+
+/// Managing and processing checkpoint files in a directory.
+pub(crate) trait LocalCheckpointFileOps {
```
kodemartin: `Ops` is ambiguous. What about renaming `LocalRead`?

kodemartin (approved):
looks good!

crates/iota-data-ingestion-core/src/reader/fetch.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7424/commits/cbcfae532205b69d597fa4779babdd1d4ac0dc99

tomxey (approved):

PR #7425, opened 2025-06-17
refactor(iota-data-ingestion-core): Introduce `reader::v2::CheckpointReader`

### LINKED ISSUE #5811: [data-ingestion-core]: Read checkpoints from historical and live object store

sergiupopescu199:
# Description of change

This PR is the second part of introducing an alternative `CheckpointReader` that supports reading from hybrid historical sources. It adds the ability to read checkpoints from historical storage, and, as an optimization, downloads checkpoints in batches. This minimizes redundant downloads of the same checkpoint file, since a single `.chk` file can contain thousands of checkpoints. The live storage is now an optional component that can be declared alongside the historical store, reflecting its optional role by design.

While most of the logic in the new `CheckpointReader` remains similar to the previous implementation, the main difference lies in how checkpoint downloads are handled. More info can be found in this [comment](https://github.com/iotaledger/iota/pull/7425#issuecomment-2996622634).

fixes #5811 

- This includes tests and doctests
```shell
```
- refactored the ingestion tests to use the new IndexerExecutor, all tests passed.
- refactored the `examples/custom-indexer/rust/local_reader.rs` to refactor the `IndexerExecutor` to run the alternative sync process using the new `run_with_config` method   which internally uses the new `reader::v2::CheckpointReader`, checked the correctness and the behavior.

```rust
use std::sync::Arc;

use anyhow::Result;
use async_trait::async_trait;
use iota_data_ingestion_core::{
    DataIngestionMetrics, IndexerExecutor, ReaderOptions, ShimProgressStore, Worker, WorkerPool,
    reader::v2::{CheckpointReaderConfig, RemoteUrl},
};
use iota_types::full_checkpoint_content::CheckpointData;
use prometheus::Registry;

struct CustomWorker;

#[async_trait]
impl Worker for CustomWorker {
    type Message = ();
    type Error = anyhow::Error;

        // custom processing logic
        // print out the checkpoint number
        println!(
            "Processing checkpoint: {}",
            checkpoint.checkpoint_summary.to_string()
        );
        Ok(())
    }
}

#[tokio::main]
async fn main() -> Result<()> {
    tracing_subscriber::fmt::init();
    let config = CheckpointReaderConfig {
        remote_store_url: Some(RemoteUrl::HybridHistoricalStore {
            historical_url: "https://checkpoints.testnet.iota.cafe".into(),
            live_url: Some("https://checkpoints.testnet.iota.cafe/ingestion/live".into()),
        }),
        reader_options: ReaderOptions::default(),
        ..Default::default()
    };
    let metrics = DataIngestionMetrics::new(&Registry::new());
    let progress_store = ShimProgressStore(0);
    let token = CancellationToken::new();
    let mut executor = IndexerExecutor::new(progress_store, 1, metrics, token.child_token());
    let worker_pool = WorkerPool::new(CustomWorker, "workflow".to_string(), 5, Default::default());
    executor.register(worker_pool).await?;
    executor.run_with_config(config).await?;

    Ok(())
}
```

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+    /// Number of maximum concurrent requests to the remote store. Increase it
+    /// for backfills, higher values increase throughput but use more resources.
+    ///
+    /// Default: 10.
+    pub batch_size: usize,
+    /// Maximum memory (bytes) for batch checkpoint processing to prevent OOM
+    /// errors. Zero indicates no limit.
+    ///
+    /// Default: 0.
+    pub data_limit: usize,
```
kodemartin: You can wrap a value of `ReaderOptions` here.

It gives proper scope and semantics for this set of fields, and would make it more concise by reusing an existing type. It would also allow easier conversions between the two.

crates/iota-data-ingestion-core/src/executor/v2.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Main executor loop.
+    ///
+    /// # Error
+    ///
+    /// Returns an [`IngestionError::EmptyWorkerPool`] if no worker pool was
+    /// registered.
+    pub async fn run(
+        mut self,
+        config: CheckpointReaderConfig,
+    ) -> IngestionResult<ExecutorProgress> {
```
kodemartin: It is not clear to me why we need an `executor::v2`. All other methods here just use `inner` and it seems to me that we could just a new method `run_with_config` that takes `CheckpointReaderConfig` that encapsulates the new reader.

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+    Rest(String),
+    /// An object storage backend for checkpoint data.
+    ObjectStore {
+        /// The URL of the object store (e.g., S3 bucket, GCS bucket,
+        /// HTTP/WebDAV Storage).
+        object_store_url: String,
+        /// Additional options for configuring the remote store as key-value
+        /// pairs.
+        remote_store_options: Vec<(String, String)>,
+    },
```
kodemartin: This is the one that we want to deprecate, no? So why to include it in the new reader?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+}
+
+/// Represents a remote backend for checkpoint data retrieval.
+///
+/// This enum encapsulates the supported remote storage mechanisms that can be
+/// used by the ingestion framework to fetch checkpoint data. Each variant
+/// corresponds to a different type of remote source.
+enum RemoteStore {
+    ObjectStore(Box<dyn ObjectStore>),
```
kodemartin: Same as above: we want to deprecate this in the new reader in favor of the hybrid solutions.

kodemartin (changes requested):
@sergiupopescu199 I left a first round of comments that have to do with high-level refactorings.

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+    },
+};
+
+/// Available sources for checkpoint streams supported by the ingestion
+/// framework.
+///
+/// This enum represents the different types of remote sources from which
+/// checkpoint data can be fetched. Each variant corresponds to a supported
+/// backend or combination of backends for checkpoint retrieval.
+pub enum RemoteUrl {
```
kodemartin, line 49: implement common traits here: [c-common-traits](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+        live_url: Option<String>,
+    },
+}
+
+/// Represents a remote backend for checkpoint data retrieval.
+///
+/// This enum encapsulates the supported remote storage mechanisms that can be
+/// used by the ingestion framework to fetch checkpoint data. Each variant
+/// corresponds to a different type of remote source.
+enum RemoteStore {
```
kodemartin, line 82:
```
impl RemoteStore {
    async fn new(remote_url: RemoteUrl) -> Result<Self>
}
```

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+        let (checkpoint_tx, checkpoint_rx) = mpsc::channel(MAX_CHECKPOINTS_IN_PROGRESS);
+        let (gc_signal_tx, gc_signal_rx) = mpsc::channel(MAX_CHECKPOINTS_IN_PROGRESS);
+        let (shutdown_tx, shutdown_rx) = oneshot::channel();
+
+        let remote_fetcher: OptionFuture<_> = config
+            .remote_store_url
+            .map(|url| RemoteCheckpointFetcher::new(url, config.timeout_secs, config.batch_size))
+            .into();
+
+        let remote_fetcher = remote_fetcher.await.transpose()?;
```
kodemartin: Nit: Explicit is better than implicit
```suggestion
        let remote_fetcher = if let Some(url) = config.remote_store_url {
            Some(RemoteCheckpointFetcher::new(url, config.timeout_secs, config.batch_size).await?)
        } else {
            None
        };
```

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,717 @@
...
+                ..Default::default()
+            },
+            use_for_pruning_watermark: false,
+        };
+        HistoricalReader::new(config)
+            .inspect_err(|e| tracing::error!("Unable to instantiate historical reader: {e}"))
+    }
+
+    /// Fetch a single checkpoint from the live object store.
+    async fn fetch_from_live_object_store(
```
kodemartin: Nit:

```rust
    async fn fetch_live(
```

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/00b114b58bbb4aac2d837476f3c95cda6261d61b#diff-9f714a0efbe52313776b7c231fc3b7360c25ea9383f54a95b8e28cb63d0c89d4R49

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/00b114b58bbb4aac2d837476f3c95cda6261d61b#diff-9f714a0efbe52313776b7c231fc3b7360c25ea9383f54a95b8e28cb63d0c89d4L54-L62

crates/iota-data-ingestion-core/src/executor/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/06f3ccbc855a2580c0d3f9c16db58cb35d3ee062

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/00b114b58bbb4aac2d837476f3c95cda6261d61b#diff-9f714a0efbe52313776b7c231fc3b7360c25ea9383f54a95b8e28cb63d0c89d4R120

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/00b114b58bbb4aac2d837476f3c95cda6261d61b#diff-9f714a0efbe52313776b7c231fc3b7360c25ea9383f54a95b8e28cb63d0c89d4L79

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/00b114b58bbb4aac2d837476f3c95cda6261d61b#diff-9f714a0efbe52313776b7c231fc3b7360c25ea9383f54a95b8e28cb63d0c89d4R76-R113

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -16,13 +16,69 @@ use tokio::{
...
+    reader::{
+        v1::CheckpointReader,
+        v2::{CheckpointReader as CheckpointReaderV2, CheckpointReaderConfig},
+    },
     worker_pool::{WorkerPool, WorkerPoolStatus},
 };
 
+pub const MAX_CHECKPOINTS_IN_PROGRESS: usize = 10000;
+
+enum CheckpointReaderVariant {
```
kodemartin: Nit: `Variant` is implicit in the variants of the enum. So I suggest you simplify to
```rust
enum CheckpointReader {
```

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,565 @@
...
+        let latest_available_checkpoint = manifest
+            .next_checkpoint_seq_num()
+            .checked_sub(1)
+            .ok_or_else(|| {
+                IngestionError::HistoryRead("no checkpoint data in the remote store".into())
+            })?;
+
+        if self.current_checkpoint_number > latest_available_checkpoint {
+            return Err(IngestionError::CheckpointNotAvailableYet);
+        }
```
kodemartin: Why do you need to check the condition `self.current_checkpoint_number > latest_available_checkpoint` two times?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,565 @@
...
+
+        let manifest = historical_reader.get_manifest().await;
+
+        let latest_available_checkpoint = manifest
+            .next_checkpoint_seq_num()
+            .checked_sub(1)
+            .ok_or_else(|| {
+                IngestionError::HistoryRead("no checkpoint data in the remote store".into())
+            })?;
```
kodemartin: Isn't this what the `historical_reader.latest_available_checkpoint()` method does?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,565 @@
...
+                        .await?;
+                }
+            }
+            RemoteStore::HybridHistoricalStore { historical, live } => {
+                if let Err(err) = self.fetch_from_historical_object_store(historical).await {
+                    if matches!(err, IngestionError::CheckpointNotAvailableYet) {
+                        let live = match live {
+                            Some(live) => live,
+                            None => return Err(err),
+                        };
```
kodemartin: Nit:
```suggestion
                        let live = live.ok_or(err)?;
```

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,565 @@
...
+                || checkpoints[0].checkpoint_summary.sequence_number
+                    > self.current_checkpoint_number)
+    }
+
+    /// Fetch a single checkpoint from the live object store.
+    async fn fetch_from_live_object_store(
+        live: &dyn ObjectStore,
+        checkpoint_number: CheckpointSequenceNumber,
+    ) -> CheckpointResult {
+        let path = ObjectStorePath::from(format!("ingestion/live/{checkpoint_number}.chk"));
```
kodemartin: Here we should assume that the `ObjectStore` url is the path to the store of `*.chk` files directly. 

Making assumptions on the path being the parent of a `ingestion/live` is rather inflexible for users and error prone.

```suggestion
        let path = ObjectStorePath::from(format!("{checkpoint_number}.chk"));
```

kodemartin (commented):
Great job @sergiupopescu199!

Made a few more comments that would require minor changes.

I would like nevertheless to suggest that we do proper QA to test that syncing from the hybrid store in a downstream application works as expected.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: I've opted to remove this code entirely in favor of using the existing `reader::fetch::fetch_from_object_store` function. After refactoring, this would have been functionally equivalent, and reusing the existing function eliminates code duplication while also addressing the path flexibility concern you raised.

fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/executor.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: As describe in this [comment](https://github.com/iotaledger/iota/pull/7425#discussion_r2160969248), I've remove the function in favor of `reader::fetch::fetch_from_object_store`.

fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/5241e16fc87e6e1997c7325af2a426a5a988324d

kodemartin (approved):

sergiupopescu199:
@iotaledger/infrastructure 

The `iota-indexer` was refactored to use the new `CheckpointReader` implementation, and works as expected.  We used `devent` network to download checkpoints: 
- historical checkpoint stream endpoint: `https://checkpoints.devnet.iota.cafe`.
- live checkpoint stream endpoint: `https://checkpoints.devnet.iota.cafe/ingestion/live`.

As test case we simulated the synchronization of the indexer from genesis for a network including migration objects:
|                                              | Old CheckpointReader | New CheckpointReader |
| -------------------------------------------- | -------------------: | -------------------: |
| Ram usage                                    |        around 195 MB |        around 215 MB |
| Primary genesis checkpoint  sync time        |        24 min 28 sec |         18 min 4 sec |
| Object snapshot genesis checkpoint sync time |        34 min 36 sec |        32 min 32 sec |

> [!NOTE]
> The increased RAM usage is primarily because the new (v2) `CheckpointReader` downloads and buffers more checkpoints at a time, making them immediately available for processing. Unlike the old (v1) `CheckpointReader`, which only fetches up to `batch_size` checkpoints every `tick_interval_ms` (default 100ms) and then waits for the next interval, the v2 implementation continuously downloads checkpoints and sends them for processing until it reaches the configured max capacity of allowed checkpoints to be held in memory or a non transient error. This approach reduces latency for workers, as they can process checkpoints without having to wait for the next polling cycle. By keeping more checkpoints readily available, v2 ensures that workers are rarely idle, leading to better throughput and more efficient resource utilization.
>
> By default the indexer allows a maximum amount of checkpoints held in memory of around 20MB, (215MB - 195MB = 20 MB)

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,555 @@
...
+        backoff.multiplier = 1.0;
+
+        loop {
+            match self.fetch_and_send_to_channel().await {
+                Ok(_) => break,
+                Err(IngestionError::MaxCheckpointsCapacityReached) => break,
+                Err(err) => {
+                    match backoff.next_backoff() {
+                        Some(duration) => {
+                            if !err.to_string().contains("404") {
```
tomxey: Does the error type not allow to have a more elegant check for status code 404?
This check could catch many more cases, e.g. "Checkpoint number 24047 does not exist"

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,555 @@
...
+        let should_fetch_from_remote = self.should_fetch_from_remote(&checkpoints);
+        self.send_local_checkpoints_to_channel(checkpoints).await?;
+
+        if should_fetch_from_remote {
+            remote_source = ReadSource::Remote;
+            self.fetch_and_send_to_channel_with_retry().await;
+        }
+
+        info!(
+            "Read from {remote_source}. Current checkpoint number: {}, pruning watermark: {}",
```
tomxey, line 428: Am I correct that this function may read and send to channel checkpoints from both local and remote sources?
Is this log line correct in such case?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,555 @@
...
+            .expect("Failed to clean the directory");
+
+        loop {
+            tokio::select! {
+                _ = &mut self.shutdown_rx => break,
+                Some(watermark) = self.gc_signal_rx.recv() => {
+                    self.data_limiter.gc(watermark);
+                    self.gc_processed_files(watermark).expect("Failed to clean the directory");
+                }
+                Ok(Some(_)) | Err(_) = timeout(Duration::from_millis(self.reader_options.tick_interval_ms), inotify_rx.recv())  => {
```
tomxey, line 449: Do you know in what cases Err will be returned here? Are those transient errors that should be ignored?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -0,0 +1,555 @@
...
+                                debug!(
+                                    "remote reader retry in {} ms. Error is {err:?}",
+                                    duration.as_millis(),
+                                );
+                            }
+                            tokio::time::sleep(duration).await
+                        }
+                        // once reached the tip of the network, the historical reader can take some
+                        // time to issue a new checkpoint file, we reset the backoff in this case.
+                        None if matches!(err, IngestionError::CheckpointNotAvailableYet) => {
```
tomxey: It's rather a nitpick, but would it make sense to move this as the first match case in a form:
```
_ if matches!(err, IngestionError::CheckpointNotAvailableYet) => {
 info!(...)
 backoff.reset();
 tokio::time::sleep(duration).await
}
```
?

Right now you could get unlucky and have several failures because of `CheckpointNotAvailableYet` that will not reset the backoff, and then at the end of `max_elapsed_time` have some other random error (e.g. connection error) that will not reset the backoff and break the loop.

I don't think think that's a super important or likely case, but just for your consideration.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: In this context, `Err(_)` is returned **only** when the timeout elapses—specifically, when `inotify_rx.recv()` does not complete within the specified duration (`tick_interval_ms`). This is not a transient error from the channel itself, but rather the expected behavior of the `tokio::time::timeout` function when the inner future does not resolve in time.

To clarify:
- `Ok(Some(_))` — A message was received on `inotify_rx` before the timeout.
- `Err(_)` — The timeout elapsed before any message was received.

So, this match arm ensures that `self.sync().await` is called either when a new file event is received or when the timeout occurs, whichever happens first. There are no transient channel errors being ignored here; only the timeout is matched as `Err(_)`.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: In the current implementation, the logic already ensures that the backoff is only reset when **both** conditions are met:

1. We have exhausted all backoff retries (`next_backoff()` returns `None`).
2. The error is `IngestionError::CheckpointNotAvailableYet`.

This is handled by the match arm:
```rust
None if matches!(err, IngestionError::CheckpointNotAvailableYet) => {
    info!(
        "Resetting backoff, historical reader does not have the requested checkpoint yet"
    );
    backoff.reset();
}
```
If `next_backoff()` returns `None` and the error is **not** `CheckpointNotAvailableYet`, we fall through to the final arm, which breaks the loop:
```rust
None => {
    error!("remote reader transient error {:?}", err);
    break;
}
```
So, the code already prioritizes resetting the backoff only in the specific case where we’ve exhausted retries and the checkpoint is not yet available. This avoids unnecessary resets for other errors and ensures the loop only breaks for non-checkpoint-related errors after all retries.

For other errors, it’s actually preferred **not** to reset the backoff. The reason is that `fetch_and_send_to_channel_with_retry()` should eventually finish its execution—either by succeeding or by breaking out of the loop after all retries are exhausted. This allows the next call to `self.sync()` to proceed, which can then attempt to fetch local checkpoints from the ingestion path if remote fetching continues to fail.

By not resetting the backoff for other errors, we avoid getting stuck in an endless retry loop for persistent or unexpected errors, and we give the system a chance to recover by trying alternative sources on the next sync cycle.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: At the moment, a more elegant check isn’t possible because the error type here can originate from either the full node’s REST API or from various object store providers, each of which formats errors differently. Since the error handling is not unified and providers may not expose HTTP status codes in a consistent way, we’re limited to string matching for now. If we standardize error handling across sources in the future, we could definitely improve this check.

I suppose this could be delegated to a stand alone issue if we need a more robust solution, what do you think?

As a temporary solution we could match not the `404` but the `not found` substring.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7425/commits/e29e89a9ecd8e8a405b39a1baf963af9f5ff46db

crates/iota-data-ingestion-core/src/reader/v2.rs
tomxey: If you are sure that all error sources will include "not found" in the message then I think it would be a better pattern to match.

And maybe let's leave it at that, it's not that important to create a new issue for this I think. (the logic here is not critical)

crates/iota-data-ingestion-core/src/reader/v2.rs
tomxey: I meant to reset backoff only for `CheckpointNotAvailableYet`, not for all errors, but just to do it always, not only when the backoff time elapses.

sergiupopescu199: Ok now I see your point about always resetting the backoff on every `CheckpointNotAvailableYet` error. 

I've refactored the code, this approach is better because we avoid resetting or exhausting the backoff when encountering `CheckpointNotAvailableYet`. Instead, we immediately break out of the retry loop for this specific error, as well as for `MaxCheckpointsCapacityReached`. This prevents unnecessary remote retries and rapid polling when the checkpoint is not yet available, and allows the next sync cycle to proceed. As a result, the system can promptly check the local ingestion path for new checkpoints that may have become available, ensuring efficient resource usage and timely processing without putting extra load on the remote source.

fixed: https://github.com/iotaledger/iota/pull/7425/commits/bda698fe9095c98be2ec5038e9b57f4187aadff6

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: I've checked locally and for now both REST & Cloudflare Bucket does contain the "not found" substring in their error messages.

fixed: https://github.com/iotaledger/iota/pull/7425/commits/100c2c9f038667407016d0b5f10e9bc237ab0488

tomxey (approved):

PR #7426, opened 2025-06-17
feat(starfish): handle additional headers on receiving end

### LINKED ISSUE #7420: handling additional block headers on the receiving end

VorobyevIlya:
# Description of change

PR adds logic that handles headers received with a block in a block bundle.

#7420 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -420,7 +420,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 self.context
                     .metrics
                     .node_metrics
-                    .invalid_header_in_a_bundle
+                    .invalid_headers_in_a_bundle
                     .with_label_values(&[
                         peer_hostname.as_str(),
                         "handle_subscribed_block_bundle",
@@ -439,7 +439,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 self.context
                     .metrics
                     .node_metrics
-                    .invalid_header_in_a_bundle
+                    .invalid_headers_in_a_bundle
                     .with_label_values(&[
                         peer_hostname.as_str(),
                         "handle_subscribed_block_bundle",
@@ -455,6 +455,12 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
             let verified_block_header =
                 VerifiedBlockHeader::new_verified(signed_block_header, serialized_header);
             additional_block_headers.push(verified_block_header);
+            self.context
+                .metrics
+                .node_metrics
+                .valid_headers_in_a_bundle
+                .with_label_values(&[peer_hostname.as_str(), "handle_subscribed_block_bundle"])
+                .inc();
         }
 
         // 5. Observe headers and the block for the commit votes. When local commit is
@@ -465,6 +471,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         self.commit_vote_monitor.observe_block(&verified_block);
 
         // TODO:: add filtering for already processed blocks/headers
+        // TODO:: add metric for filtered blocks/headers
 
         // 6. Reject blocks when local commit index is lagging too far from quorum
         //    commit
@@ -513,6 +520,19 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         //    them
         // Normally, there should be no missing ancestors, as the headers are sent in
         // order of increasing rounds.
+
+        // TODO::Uncomment when the filter for headers is implemented, and already
+        // processed headers are removed from additional_block_headers. Before that it
+        // is incorrect
+        // self.context
+        // .metrics
+        // .node_metrics
+        // .received_unique_headers_from_a_bundle
+        // .with_label_values(&[
+        // peer_hostname.as_str(),
+        // "handle_subscribed_block_bundle",
+        // ])
+        // .inc_by(additional_block_headers.len() as u64);
 
         let mut missing_ancestors = self
             .core_dispatcher
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -83,6 +83,7 @@ impl<C: CoreThreadDispatcher> AuthorityService<C> {
 
 #[async_trait]
 impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
+    #[cfg(test)]
     async fn handle_subscribed_block(
         &self,
         peer: AuthorityIndex,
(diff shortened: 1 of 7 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -269,12 +269,239 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        // handle headers
+
+        self.context
+            .metrics
+            .node_metrics
+            .verified_blocks
+            .with_label_values(&[peer_hostname])
+            .inc();
+
+        let mut missing_ancestors = self
```
polinikita, line 537: Add comments here and below on what is done in this step. You have steps 1-4, but this one is logically even more important

crates/starfish/core/src/authority_service.rs
```diff
@@ -269,12 +269,239 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        }
+        self.commit_vote_monitor.observe_block(&verified_block);
+
+        // Reject blocks when local commit index is lagging too far from quorum commit
+        // index.
+        //
+        // IMPORTANT: this must be done after observing votes from the block, otherwise
+        // observed quorum commit will no longer progress.
+        //
+        // Since the main issue with too many suspended blocks is memory usage not CPU,
```
polinikita: I am not sure that with the push strategy it would be the case, so I would remove this comment and add a todo regarding filtering already processed blocks

polinikita (approved):
Overall looks fine to me, but I would add a few todos for future work here

crates/starfish/core/src/authority_service.rs
```diff
@@ -269,12 +269,239 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+            .with_label_values(&[peer_hostname])
+            .inc();
+
+        let mut missing_ancestors = self
+            .core_dispatcher
+            .add_block_headers(additional_block_headers)
+            .await
+            .map_err(|_| ConsensusError::Shutdown)?;
+
+        missing_ancestors.extend(
```
polinikita, line 548: This might be ok to add the main block after all these, but I would add a todo about optimization in these steps.

VorobyevIlya: what optimization are you talking about?

crates/starfish/core/src/authority_service.rs
```diff
@@ -269,12 +270,244 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+                verified_block.timestamp_ms(),
+                now,
+                forward_time_drift.as_millis(),
+            );
+            sleep(forward_time_drift).await;
+        }
+
+        // 4. Create block headers from bytes from a bundle
+
+        let mut additional_block_headers = vec![];
```
polinikita, line 410: I would add an extra check of the maximum number of headers and that their round is below the one of the main block

crates/starfish/core/src/authority_service.rs
polinikita: Here, all block headers undergo the main core task. Instead, one can
1. First process the main block and if not successful, feed the necessary block headers (not all, but those that are missing)
2. Add byte filter if the same block header is processed by another thread (since almost all block headers are expected to arrive from every other node). To this end, one needs to have a global structure that store the filter and reads before attempting to process a new blockheader in the form of Bytes

polinikita (changes requested):
Left new comments about todos and an extra check for additional block headers arriving with the bundle

crates/starfish/core/src/authority_service.rs
VorobyevIlya: check is added

crates/starfish/core/src/authority_service.rs
VorobyevIlya: 1. added a TODO. 2. There is already TODO about filter above step 6.

crates/starfish/core/src/authority_service.rs
```diff
@@ -269,12 +270,261 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+            return Err(ConsensusError::BlockRejected {
+                block_ref,
+                reason: format!(
+                    "Last commit index is lagging quorum commit index too much ({} < {})",
+                    last_commit_index, quorum_commit_index,
+                ),
+            });
+        }
+
+        self.context
```
piotrm50, line 512: We should have a metric for valid block headers in the bundle as well as the invalid blocks above should be separately counted for streamed block headers and full blocks.

crates/starfish/core/src/authority_service.rs
piotrm50: Are the headers in the bundle the oldest or the newest? 
> First process the main block and if not successful, feed the necessary block headers (not all, but those that are missing)

This seems a bit wasteful if the gap is more than one round and in fact we are actually missing more blocks - in this case the only thing we save is one Synchronizer run, or do I misunderstand something?

piotrm50 (changes requested):
Please are more detailed metrics for blocks in the bundle.

crates/starfish/core/src/authority_service.rs
VorobyevIlya: I created a separate metric for an invalid header in a bundle, and use it instead of the metric for invalid blocks above. Do you think we need a metric for valid headers in a bundle as well?

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Headers in the bundle are the oldest.
I also now have doubts about this optimisation. If the block is 'new' and the headers are 'old', will we just ignore them? That doesn't look good

piotrm50: Then the oldest blocks should in be accepted immediately most of the time (we should have a metric for that). I think sending the oldest blocks is a good approach as the newest blocks will be fetched through synchronizer so we will not be fetching from the same direction using two separate mechanisms.

crates/starfish/core/src/authority_service.rs
piotrm50: Yes, and also maybe a metric for how many headers from the bundle were accepted, how many headers already existed/were filtered out by the bytes filter to asses if the bundle is sending too many unnecessary blocks and if we can send them with a delay or something.

VorobyevIlya: Ok, I created a metric for received valid headers from bundles and added TODOS for metrics for filtered headers to implement them with the filter

crates/starfish/core/src/authority_service.rs
polinikita: No, I meant that we feed from the previous round only blocks that are missing. Regarding blocks from older rounds, we can feed all of them. I want to reduce a huge overlap in processing headers from the previous round mainly. This is very important from the results I observed in the testbed. For a very low load and a large committee, the computational overhead for the core thread and the concurrent access to dag_state was high because of quadratic number of block headers in every round. 

By the way, the contention might happen also in the commit observer, so I would make the optimization including that step

polinikita (approved):

piotrm50 (approved):

PR #7443, opened 2025-06-18
feat: Allow all name commands to be called on domains and subdomains

### LINKED ISSUE #7006: Make sure all IOTA-Names CLI commands work with names AND subdomains
If it actually makes sense for both.

DaughterOfMars:
# Description of change

Allows commands which can be run on both domains and subdomains to do so.

Closes #7006

## Type of change

### DIFF
--- a/crates/iota/src/name_commands.rs
+++ b/crates/iota/src/name_commands.rs
@@ -479,7 +479,7 @@ impl NameCommand {
                 opts,
             } => {
                 // Check ownership of the name off-chain to avoid potentially wasting gas
-                get_owned_nft_by_name::<IotaNamesRegistration>(&domain, context).await?;
+                get_proxy_nft_by_name(&domain, context).await?;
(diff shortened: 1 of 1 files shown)


crates/iota/src/name_commands.rs
```diff
@@ -479,7 +479,7 @@ impl NameCommand {
                 opts,
             } => {
                 // Check ownership of the name off-chain to avoid potentially wasting gas
-                get_owned_nft_by_name::<IotaNamesRegistration>(&domain, context).await?;
+                get_proxy_nft_by_name(&domain, context).await?;
```
thibault-martinez, line 534: Should also change package/module in the call?

DaughterOfMars: There is none. This one just calls it to verify

Thoralf-M (approved):

thibault-martinez (approved):

PR #7458, opened 2025-06-20
fix(explorer): Correct formatting over 9 decimals

marc2332:
before

image

now
image


evavirseda (approved):

msarcev:
Can we add a test for this ?

marc2332:
> Can we add a test for this ?

Where, the apps core package is not even a real package, not sure if adding a test is necessary for this either

msarcev:
> > Can we add a test for this ?
> 
> Where, the apps core package is not even a real package, not sure if adding a test is necessary for this either

There are already tests for the "subscript" formatting in `apps/core/src/hooks/__tests__/useFormatCoin.test.ts`. We could just add another case.

msarcev (approved):

Issue #4390, opened 2024-12-09
Port upstream node team related changes [mainnet-v1.36.2..mainnet-v1.38.4)


jkrvivian:
First commit: 68ac44a2b180547635cb6733b0fb9b2ab6d7d0a9
Last commit: b164e321fc9c51745043a3dc148ef95d258a6b0c
SINCE: 2024-10-26
UNTIL: 2024-12-05 
 
- [x] docker (1)
- [x] crates/sui-config (1)
- [x] crates/sui-core (12)
- [x] crates/sui-types (6)
- [x] crates/sui-protocol-config (6)
- [x] crates/sui-rest-api (10)
- [x] Cross-crate commits (13)

The list of commits is in [doc](https://iotafoundation.atlassian.net/wiki/x/DwBHig)

daria305:
List is categorized based on priority and relevant topic for Questionable categories. 
At least two of the changes listed were based on the commits not included in any of the summaries (v1.36.2, v1.38.3), we might consider expanding the report with listings for `sui-indexer` crate.

jkrvivian:
All sub-issues are completed. 🎉

PR #7465, opened 2025-06-20
upstream(core): Panic on db failure instead of returning Result

jkrvivian:
# Description of change

- Port commit: https://github.com/MystenLabs/sui/commit/9289d3dc46eb9f130575f9faba29f76eea8000a9

Part of #4390  

### DIFF
--- a/crates/iota-config/src/genesis.rs
+++ b/crates/iota-config/src/genesis.rs
@@ -334,14 +334,14 @@ impl UnsignedGenesis {
 
     pub fn has_randomness_state_object(&self) -> bool {
         self.objects()
-            .get_object(&IOTA_RANDOMNESS_STATE_OBJECT_ID)
+            .try_get_object(&IOTA_RANDOMNESS_STATE_OBJECT_ID)
             .expect("read from genesis cannot fail")
             .is_some()
     }
 
     pub fn has_bridge_object(&self) -> bool {
         self.objects()
-            .get_object(&GENESIS_IOTA_BRIDGE_OBJECT_ID)
+            .try_get_object(&GENESIS_IOTA_BRIDGE_OBJECT_ID)
             .expect("read from genesis cannot fail")
             .is_some()
     }
--- a/crates/iota-config/src/genesis.rs
+++ b/crates/iota-config/src/genesis.rs
@@ -32,7 +32,7 @@ use iota_types::{
         CertifiedCheckpointSummary, CheckpointContents, CheckpointSummary, VerifiedCheckpoint,
     },
     object::Object,
-    storage::ObjectStore,
+    storage::ObjectStoreFallible,
     transaction::Transaction,
 };
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
--- a/crates/iota-config/src/genesis.rs
+++ b/crates/iota-config/src/genesis.rs
@@ -32,7 +32,7 @@ use iota_types::{
         CertifiedCheckpointSummary, CheckpointContents, CheckpointSummary, VerifiedCheckpoint,
     },
     object::Object,
-    storage::ObjectStoreFallible,
+    storage::ObjectStoreNonFallible,
     transaction::Transaction,
 };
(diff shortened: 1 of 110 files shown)


TheMrAI:
Could you give more context what this change does?
"Panic on db failure instead of returning Result" sounds weird to me. DB failures are normal no? Should we crash the thread/app completely if they occur?

jkrvivian:
The commit changes certain database operations to panic (using `.expect("db error")`) instead of propagating errors as a `Result`. If a database operation fails due to I/O error or corruption, the node will immediately panic and crash, rather than returning an error for the caller.

crates/iota-config/src/genesis.rs
```diff
@@ -335,19 +335,15 @@ impl UnsignedGenesis {
...
-        self.objects()
-            .get_object(&IOTA_BRIDGE_OBJECT_ID)
-            .expect("read from genesis cannot fail")
-            .is_some()
+        self.objects().get_object(&IOTA_BRIDGE_OBJECT_ID).is_some()
     }
 
     pub fn has_coin_deny_list_object(&self) -> bool {
-        get_deny_list_root_object(&self.objects()).is_ok()
+        get_deny_list_root_object(&self.objects()).iter().len() > 0
```
muXxer: Is this logic correct? Shouldn't it be enough to check for "is_some" ?

crates/iota-e2e-tests/tests/bridge_tests.rs
```diff
@@ -30,7 +30,6 @@ async fn test_create_bridge_state_object() {
                 node.state()
                     .get_object_cache_reader()
                     .get_latest_object_ref_or_tombstone(IOTA_BRIDGE_OBJECT_ID)
-                    .unwrap()
```
muXxer: mhh we should rebase our branch, the bridge should be gone.

semenov-vladyslav (approved):

muXxer (approved):

TheMrAI:
> The commit changes certain database operations to panic (using `.expect("db error")`) instead of propagating errors as a `Result`. If a database operation fails due to I/O error or corruption, the node will immediately panic and crash, rather than returning an error for the caller.

This is why I am asking. An I/O error is normal and should be handled appropriately, by retrying or abandoning the operation.
What am I missing still?
Now if you say that these db failures mean we are in deep deep trouble and there is nothing to do I get it, but I am missing this context, hence asking.

muXxer:
> > The commit changes certain database operations to panic (using `.expect("db error")`) instead of propagating errors as a `Result`. If a database operation fails due to I/O error or corruption, the node will immediately panic and crash, rather than returning an error for the caller.
> 
> This is why I am asking. An I/O error is normal and should be handled appropriately, by retrying or abandoning the operation. What am I missing still? Now if you say that these db failures mean we are in deep deep trouble and there is nothing to do I get it, but I am missing this context, hence asking.

I wouldn't say that an I/O error is normal in terms of database access. Right now that burden is simply on the caller side, everywhere in the node. What do you do as the caller if your write request to the database fails. You are simply missing the context why it failed, and what do you even do in such situations? For example disc full, or wrong permissions, or running out on max open files. I agree with the change mysten did here, it was just another layer of complication, and everywhere was an "expect" or "unwrap" anyways. So panicing here is fine, the node operator has to take care of that.

kodemartin (changes requested):
This patch introduces breaking changes that affect any user of the underlying public API.

I would therefore suggest that you consider introducing the changes in a smoother way. You could define new traits that implement the methods with the new signatures (e.g. MaybeObjectStore) and then impl MaybeObjectStore for T: ObjectStore so that to expose the new functionality to implementors.

This way users can opt in in transitioning to the new API instead of being forced to do so.

piotrm50 (approved):

jkrvivian:
We will implement both non-fallible and fallible version of storage traits in the 2 PRs in order to resolve the review comments:
* #7768 
* #7769

PR #7466, opened 2025-06-20
chore(docs): update developer sidebar to remove duplicated entries

29988122:
Remove duplicated entries of move-concepts.
image

diff'd, confirmed that the two entries have the same content.


Dr-Electron (approved):
Thanks for the cleanup ❤️

salaheldinsoliman (approved):

PR #7480, opened 2025-06-20
feat(apps-ui-kit): improve ui-kit styles after multitheme 

### LINKED ISSUE #7457: Improve ui-kit styles after multitheme
Feedback:
### apps-ui-kit storybook app:
- In the `design tokens` section, all the pages don't take the styles until I go to a component and then come back to any page. It's more noticeable on the typeset page. This is not that important but I think it would be good to fix it
### apps-ui-kit components:
- **Badge**:
  -  LIGHT:
    - `warning` has wrong text color. Figma and prod are both `#422C00`
    - `error` has also wrong text color.
  - DARK:
    - `warning` has wrong background color `#422C00` -> `#2D1E00`. Text color is ok
- **Button**
  - Secondary button loses background color when disabled. (left pr deploy, right prod) image
  - Outlined button has different border color on dark image
- **Button Segment**:
  - When type is underlined, and theme is dark, and is disabled, the underline in production is the same color as the text, while the current in this PR is a bit more blue image
  - Underlined, lightmode: the border that appears on hover is wrongly positioned now (it should be placed on top of the other border. Weird though, when selected it gets placed correctly image image
- **Checkbox**:
  - Only on dark theme, if the checkbox is disabled and checked should have a gray brackground. [pr vs prod vs figma] image
- **InfoBox**
  - Icons has different color on both light and dark for success variant image
   - Error has wrong icon color in lightmode (also in type elevated)  image
   - Warning in dark theme has incorrect bg on type elevated  image
   - Success variant on type elevated has different icon colors (also on dark) image
- **Snackbar**
  - Text color when type is error and theme is light are wrong image
  - Type warning has also different text colors  image
  - Type warning theme dark has different background image
- **Scrollbar**
  - The scrollbar in darkmode has different colors. Seems that when looking at the `Scrollbar` page in the storybook it has the correct colors, but the scrollbar that is on the apps have a different color image
- **Input**
  - Caret has lost it's color 
  image image
  - Placeholder in darkmode has different color image
  - Seems the numeric input changes the colors for the max button. I don't know if this is an issue with the classes in the story or has to do with some colors being inherited image
- **TextArea**:
  - Caret missing color
  - When value is hidden, the skeleton is much more visible now in both light and dark image

Juligs:
fixes #7457


apps/ui-kit/src/lib/components/atoms/button/button.classes.ts
```diff
@@ -22,7 +22,7 @@ export const BACKGROUND_COLORS: Record<ButtonType, string> = {
 };
 
 export const DISABLED_BACKGROUND_COLORS: Record<ButtonType, string> = {
-    [ButtonType.Primary]: 'bg-neutral-80 dark:bg-neutral-30',
+    [ButtonType.Primary]: 'button-bg-color-disabled-primary',
     [ButtonType.Secondary]: 'button-bg-color-secondary',
     [ButtonType.Ghost]: 'bg-transparent',
     [ButtonType.Outlined]: 'bg-transparent button-border-color-outline',
```
evavirseda, line 28: in figma outlined button border is `#8892A1` and it has `#6e7787` (dark) :pray:

evavirseda (changes requested):
just one detail :pray:

evavirseda (approved):

PR #7485, opened 2025-06-20
feat(msim): Remove random cache init in simtests

jkrvivian:
# Description of change

The cache in the simtests was randomly choosen, which might yield random failures/results. We want to avoid that.

- Partly ported commit: https://github.com/MystenLabs/sui/commit/b09b0a8d36e1e03352811cc180d4bb92a2cb677f

Part of #4390  

### DIFF
--- a/crates/iota-core/src/execution_cache.rs
+++ b/crates/iota-core/src/execution_cache.rs
@@ -30,7 +30,7 @@ use crate::{
         AuthorityStore,
         authority_per_epoch_store::AuthorityPerEpochStore,
         authority_store::{ExecutionLockWriteGuard, IotaLockResult},
-        epoch_start_configuration::{EpochFlag, EpochStartConfigTrait, EpochStartConfiguration},
+        epoch_start_configuration::{EpochFlag, EpochStartConfiguration},
     },
     state_accumulator::AccumulatorStore,
     transaction_outputs::TransactionOutputs,
@@ -128,14 +128,9 @@ pub fn build_execution_cache(
 ) -> ExecutionCacheTraitPointers {
     let execution_cache_metrics = Arc::new(ExecutionCacheMetrics::new(prometheus_registry));
 
-    match epoch_start_config.execution_cache_type() {
-        ExecutionCacheConfigType::WritebackCache => ExecutionCacheTraitPointers::new(
-            WritebackCache::new(store.clone(), execution_cache_metrics).into(),
-        ),
-        ExecutionCacheConfigType::PassthroughCache => ExecutionCacheTraitPointers::new(
-            ProxyCache::new(epoch_start_config, store.clone(), execution_cache_metrics).into(),
-        ),
-    }
+    ExecutionCacheTraitPointers::new(
+        ProxyCache::new(epoch_start_config, store.clone(), execution_cache_metrics).into(),
+    )
 }
 
 /// Should only be used for iota-tool or tests. Nodes must use
--- a/crates/iota-core/src/execution_cache.rs
+++ b/crates/iota-core/src/execution_cache.rs
@@ -23,14 +23,14 @@ use iota_types::{
     transaction::{VerifiedSignedTransaction, VerifiedTransaction},
 };
 use prometheus::Registry;
-use tracing::instrument;
+use tracing::{error, instrument};
 
 use crate::{
     authority::{
         AuthorityStore,
         authority_per_epoch_store::AuthorityPerEpochStore,
         authority_store::{ExecutionLockWriteGuard, IotaLockResult},
-        epoch_start_configuration::{EpochFlag, EpochStartConfiguration},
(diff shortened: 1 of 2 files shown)


polinikita (approved):

Thoralf-M (approved):

bingyanglin (approved):

muXxer (approved):

muXxer:
That affects the node operators, please add a release note.

muXxer:
Let's merge this one after the fix in #7543

crates/iota-core/src/execution_cache.rs
```diff
@@ -112,31 +112,12 @@ pub enum ExecutionCacheConfigType {
...
     }
 
-    if std::env::var(DISABLE_WRITEBACK_CACHE_ENV_VAR).is_ok()
-        || matches!(config, ExecutionCacheConfig::PassthroughCache)
-    {
-        ExecutionCacheConfigType::PassthroughCache
-    } else {
-        ExecutionCacheConfigType::WritebackCache
-    }
+    ExecutionCacheConfigType::WritebackCache
```
muXxer: I would revert this change here, So we can still specify PassthroughCache in the config if we want.
(the use random cache for simtests can be removed)

crates/iota-core/src/execution_cache.rs
```diff
@@ -913,15 +900,10 @@ macro_rules! implement_passthrough_traits {
...
-                // than risk forking because ProxyCache::reconfigure_cache was not
-                // called.
-                panic!(
-                    "reconfigure_cache should not be called on a {}",
-                    stringify!($implementor)
-                );
+                // Since we now use WritebackCache directly at startup (if the epoch flag is
+                // set), this can be called at reconfiguration time. It is a no-op.
+                // TODO: remove this once we completely remove ProxyCache.
+                std::future::ready(()).boxed()
```
muXxer, line 900: Can we somehow check if the implementor is the proxy cache and then panic with the old message, otherwise do the no-op?

muXxer (changes requested):

crates/iota-core/src/execution_cache.rs
jkrvivian: Proxy cache implements its own `reconfigure_cache`, so it'll not trigger the implementation here.
Writeback cache and Passthrough cache share this implementation, it's no-op for both. (bc no need to switch cache)

muXxer: Yeah but's the whole point of this panic here, if for whatever reason that slipped somehow through for proxy_cache, it needs to panic here, because then the reconfigure_cache was not called.

jkrvivian: This is the `reconfigure_cache` of proxy cache, it will not go through the one above
https://github.com/iotaledger/iota/blob/fc692a2807ea994e1e0a0badaa6786b5428026be/crates/iota-core/src/execution_cache/proxy_cache.rs#L384-L389

https://github.com/iotaledger/iota/blob/fc692a2807ea994e1e0a0badaa6786b5428026be/crates/iota-core/src/execution_cache/proxy_cache.rs#L77-L97

jkrvivian: I don't see how it'll slip through with proxy cache. It only panics if we call `reconfigure_cache` from the Writeback or Passthrough cache in proxy cache directly, but there's no such behaviour in the current codebase.

crates/iota-core/src/execution_cache.rs
```diff
@@ -145,9 +128,15 @@ pub fn build_execution_cache(
...
-        ProxyCache::new(epoch_start_config, store.clone(), execution_cache_metrics).into(),
-    )
+
+    match epoch_start_config.execution_cache_type() {
+        ExecutionCacheConfigType::WritebackCache => ExecutionCacheTraitPointers::new(
+            WritebackCache::new(store.clone(), execution_cache_metrics).into(),
+        ),
+        ExecutionCacheConfigType::PassthroughCache => ExecutionCacheTraitPointers::new(
+            ProxyCache::new(epoch_start_config, store.clone(), execution_cache_metrics).into(),
+        ),
```
jkrvivian: @muXxer If we use PassthroughCache instead of ProxyCache here, we can remove the entire ProxyCache and the `reconfigure_cache` function. However, if we want to retain the flexibility to switch between different cache implementations, it may be worth keeping them.

semenov-vladyslav: We should keep the old code here. With this match, if WritebackCache is enabled (at node startup), it'll be enabled forever, as ProxyCache is responsible for switching cache types.

crates/iota-core/src/execution_cache.rs
semenov-vladyslav: I don't see the problem here. It's pretty easy to track how `reconfigure_cache` is called without panics and tests.

muXxer (approved):

PR #7487, opened 2025-06-20
upstream(node): Add nodes with unpruned history in simtest

jkrvivian:
# Description of change

- Port commit: https://github.com/MystenLabs/sui/commit/c27594503a7b129a2ca1ae1b03cf78a2b5388fbc
- Descriptions from commits
  Simtest `test_simulated_load_shared_object_congestion_control` seems to have been occasionally failing the assertion `assertion failed: results.num_successful_transactions > 0` , ex:

https://github.com/MystenLabs/sui/actions/runs/11949884838/job/33310255818

https://github.com/MystenLabs/sui/actions/runs/11082671210/job/30796002007

Investigating the issue this looks to be related to the Full Node falling behind on state sync from the rest of the network (due to connectivity issues etc) and when coming back effectively rest of the network has advanced and also pruned checkpoint history that would be necessary for the FN in order to advance. Any transactions at this point can not successfully execute as local state is missing dependencies which can never be fetched.

This PR is making two of the validators having unpruned history so if that happens FN can still sync state when falls behind.

Part of #4390  

### DIFF
--- a/crates/iota-benchmark/tests/simtest.rs
+++ b/crates/iota-benchmark/tests/simtest.rs
@@ -196,7 +196,7 @@ mod test {
     #[sim_test(config = "test_config()")]
     async fn test_simulated_load_small_committee_reconfig() {
         iota_protocol_config::ProtocolConfig::poison_get_for_min_version();
-        let test_cluster = build_test_cluster(1, 5_000 /* , 0 */).await;
+        let test_cluster = build_test_cluster(1, 5_000, 0).await;
         test_simulated_load(test_cluster, 120).await;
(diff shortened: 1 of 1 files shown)


bingyanglin (approved):

muXxer (approved):
Any idea why the tests are failing? I will rerun them.

Oh actually the error is that `build_test_cluster` is missing an argument.

muXxer (changes requested):
Please fix `build_test_cluster` arguments.

muXxer (approved):

PR #7488, opened 2025-06-20
feat(iota-e2e-tests): add e2e test for invalid Identifier check

TheMrAI:
Adding an e2e test checking whether or not a transaction constructed containing invalid module or function Identifiers can be sent in.

In protocol version 9 these can still be accepted, but from version 10 they should be rejected outright.

### Tests exucted

```cargo nextest run -p iota-e2e-tests --test tx_bytes_validity_check```


valeriyr (approved):
LGTM!

PR #7494, opened 2025-06-21
feat(consensus): Always fetch own blocks from a quorum and expect a quorum of subscriptions

### LINKED ISSUE #7251: Change the condition to fetch last proposed block from the network in Mysticeti
Currently the condition whether a node should fetch the last own proposed block from the network is the following:
```rust
        let sync_last_known_own_block = boot_counter == 0
            && dag_state.read().highest_accepted_round() == 0
            && !context
                .parameters
                .sync_last_known_own_block_timeout
                .is_zero();
        info!("Sync last known own block: {sync_last_known_own_block}");
```
I think is wrong - if a validator has written to the DB before it was restarted, then the middle condition (`dag_state.read().highest_accepted_round() == 0`) will be `false`. because this value is updated as we fill the DAG State's fields upon startup. 

What this middle condition means is that you only sync your latest block when you start from scratch, but when you write to the DB, you assume that it's correctly persisted, but as we learned multiple times this is not always true because we flush data to disk with a delay, so if a node suddenly crashes then some data will be gone.

polinikita:
# Description of change

Two main changes that are logically connected: 
 1 - Fetching own block during startup: When starting the consensus node, we always try to fetch own latest block even if the consensus database is non-empty. 
 2 - Attempt to propose new blocks only with a quorum of subscriptions: Now we control the number of subscriptions to our block stream. Stop proposing blocks when there is no quorum of subscriptions.

Both changes are made to enhance the network's reliability following crashes and restarts. In particular, we propose and disseminate our own block only when we have a quorum of connections; thereby, if our block is disseminated to a quorum, we will fetch it from at least one connection during startup due to quorum intersection.

Why 1 is important: Some validators reported the creation of equivocating blocks after the restart. That could happen because if the node first erased the consensus db and then partially filled it with some blocks, it didn't ask the network about its latest block due to failed `dag_state.read().highest_accepted_round() == 0` condition. After creating a new block, it then attempted to add another block from the same slot that was already committed. **The change 1** ensures that we always fetch our own latest block from a quorum (by stake) of the network.

Why 2 is important: To propose a new block, we examine active connections - specifically, subscriptions for the produced block stream. Previously, one active subscription was enough to propose new blocks. However, we observed occasionally that when there are not enough active subscriptions, the node could spam the network with its long hidden chain of blocks, which eventually results in a big problem of block synchronization. The change 2 prevents this situation by requiring a quorum (by stake) of active subscriptions.

Fixes #7251 

CI

### Release Notes

- [x] Protocol: always fetch own block from a quorum of the network and stop proposing new blocks when a quorum of subscriptions is gone.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/consensus/core/src/authority_service.rs
+++ b/consensus/core/src/authority_service.rs
@@ -570,8 +570,21 @@ impl SubscriptionCounter {
     fn increment(&self, peer: AuthorityIndex) -> Result<(), ConsensusError> {
         let mut counter = self.counter.lock();
         counter.count += 1;
+        let original_subscription_by_peer = counter.subscriptions_by_authority[peer];
         counter.subscriptions_by_authority[peer] += 1;
-
+        let mut total_stake = 0;
+        for (authority_index, _) in self.context.committee.authorities() {
+            if counter.subscriptions_by_authority[authority_index] >= 1 || self.context.own_index == authority_index {
+                total_stake += self.context.committee.stake(authority_index);           
+            }
+        }
+        // Stake of subscriptions before a new peer was subscribed
+        let previous_stake = if original_subscription_by_peer == 0 {
+            total_stake - self.context.committee.stake(peer)
+        } else {
+            total_stake 
+        };
+        
         let peer_hostname = &self.context.committee.authority(peer).hostname;
         self.context
             .metrics
@@ -579,18 +592,34 @@ impl SubscriptionCounter {
             .subscribed_by
             .with_label_values(&[peer_hostname])
             .set(1);
-        if counter.count == 1 {
+        // If the subscription count reaches quorum, notify the dispatcher and get ready to propose blocks.
+        if !self.context.committee.reached_quorum(previous_stake) && self.context.committee.reached_quorum(total_stake) {
             self.dispatcher
-                .set_subscriber_exists(true)
+                .set_quorum_subscribers_exists(true)
                 .map_err(|_| ConsensusError::Shutdown)?;
         }
+        // Drop the counter after sending the command to the dispatcher
+        drop(counter);
         Ok(())
     }
 
     fn decrement(&self, peer: AuthorityIndex) -> Result<(), ConsensusError> {
         let mut counter = self.counter.lock();
         counter.count -= 1;
+        let original_subscription_by_peer = counter.subscriptions_by_authority[peer];
         counter.subscriptions_by_authority[peer] -= 1;
+        let mut total_stake = 0;
+        for (authority_index, _) in self.context.committee.authorities() {
+            if counter.subscriptions_by_authority[authority_index] >= 1 || self.context.own_index == authority_index {
+                total_stake += self.context.committee.stake(authority_index);
+            }
+        }
+        // Stake of subscriptions before a peer was dropped
+        let previous_stake = if original_subscription_by_peer == 1 {
+            total_stake + self.context.committee.stake(peer)
+        } else {
+          total_stake
+        };
 
         if counter.subscriptions_by_authority[peer] == 0 {
             let peer_hostname = &self.context.committee.authority(peer).hostname;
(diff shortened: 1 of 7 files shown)


consensus/core/src/authority_service.rs
```diff
@@ -579,19 +594,39 @@ impl SubscriptionCounter {
...
+        if !self.context.committee.reached_quorum(previous_stake)
+            && self.context.committee.reached_quorum(total_stake)
+        {
             self.dispatcher
-                .set_subscriber_exists(true)
+                .set_quorum_subscribers_exists(true)
                 .map_err(|_| ConsensusError::Shutdown)?;
         }
+        // Drop the counter after sending the command to the dispatcher
+        drop(counter);
```
VorobyevIlya, line 607: why do we need to drop here? it will be out of scope anyway after the next line

consensus/core/src/authority_service.rs
```diff
@@ -570,7 +570,22 @@ impl SubscriptionCounter {
     fn increment(&self, peer: AuthorityIndex) -> Result<(), ConsensusError> {
         let mut counter = self.counter.lock();
         counter.count += 1;
+        let original_subscription_by_peer = counter.subscriptions_by_authority[peer];
         counter.subscriptions_by_authority[peer] += 1;
+        let mut total_stake = 0;
+        for (authority_index, _) in self.context.committee.authorities() {
+            if counter.subscriptions_by_authority[authority_index] >= 1
```
VorobyevIlya, line 577: Should the values here be only 0 or 1? Why is it >=?

consensus/core/src/authority_service.rs
```diff
@@ -603,11 +638,17 @@ impl SubscriptionCounter {
...
+        if self.context.committee.reached_quorum(previous_stake)
+            && !self.context.committee.reached_quorum(total_stake)
+        {
             self.dispatcher
-                .set_subscriber_exists(false)
+                .set_quorum_subscribers_exists(false)
                 .map_err(|_| ConsensusError::Shutdown)?;
         }
+        // Drop the counter after sending the command to the dispatcher
+        drop(counter);
```
VorobyevIlya, line 651: again, why do we need to drop?

VorobyevIlya (approved):
left some minor comments, but overall looks good

consensus/core/src/authority_service.rs
polinikita: Not entirely sure how to extend the lifetime of this variable such that the lock is persisted until sending to dispatcher is done. I found it strange that the commands could happen out of the order leading to unexpected behaviour.

consensus/core/src/authority_service.rs
polinikita: Yes, in my opinion it should 0,1. Still not sure why these variables are tracked as integers...

piotrm50: It's most likely to account for validators opening more than one subscription connection.

alexsporn: I guess in theory an authority could subscribe multiple times. Interestingly the prometheus metric is always set to 0,1 instead of `counter.subscriptions_by_authority[peer]`

piotrm50 (commented):
lgtm

consensus/core/src/authority_service.rs
piotrm50: Drop is needed because otherwise lock is dropped after its last use. Before it wasn't necessary because `if` statement used the `counter` variable which extended its lifespan, but now that the condition changed we need to extend it using `drop`.

piotrm50 (approved):

alexsporn:
Please backport this to the 1.3.x release once merged

PR #7506, opened 2025-06-23
chore(CLI): Better error for leaf `set-target-address`

### LINKED ISSUE #7505: `iota name set-target-address` doesn't work for leaves
```bash
thibault@Mac ~/i/iota (develop)> ./target/debug/iota name set-target-address leaf.nounny.iota 0x6b427c797f0bc1d93941ba8ab8d395a5bfe4b73e7faac2011696ba50dc149161
[warn] Client/Server api version mismatch, client api version: 1.4.0-alpha, server api version: 1.2.2-rc
no matching owned SubdomainRegistration found for leaf.nounny.iota
```

DaughterOfMars:
## Description 

Leaf subdomains cannot have their target addresses set or unset, so this PR adds an error when the registry entry exists but there is no NFT.

Closes #7505

### DIFF
--- a/crates/iota/src/name_commands.rs
+++ b/crates/iota/src/name_commands.rs
@@ -575,6 +575,11 @@ impl NameCommand {
                 opts,
             } => {
                 let entry = get_registry_entry(&domain, &iota_client).await?;
+                if entry.name_record.is_leaf_record() {
+                    bail!(
+                        "cannot set target address for leaf subdomain; try removing and recreating the subdomain instead."
+                    );
+                }
                 let new_address =
                     get_identity_address(new_address.map(KeyIdentity::Address), context)?;
                 if entry
@@ -584,11 +589,7 @@ impl NameCommand {
                 {
                     bail!("target address is already set to the given value");
                 }
-                let Some(nft) = get_proxy_nft_by_id_opt(entry.id, &domain, context).await? else {
-                    bail!(
-                        "cannot set target address for leaf subdomain; try removing and recreating the subdomain instead."
-                    );
-                };
+                let nft = get_proxy_nft_by_id(entry.id, &domain, context).await?;
                 let iota_names_config = get_iota_names_config(&iota_client).await?;
 
                 let res = IotaClientCommands::Call {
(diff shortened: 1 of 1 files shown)


Thoralf-M (approved):

crates/iota/src/name_commands.rs
```diff
@@ -583,7 +584,11 @@ impl NameCommand {
                 {
                     bail!("target address is already set to the given value");
                 }
-                let nft = get_proxy_nft_by_name(&domain, context).await?;
+                let Some(nft) = get_proxy_nft_by_id_opt(entry.id, &domain, context).await? else {
```
thibault-martinez: You could have just checked if the entry you already have is a leaf record or not

DaughterOfMars: How?

thibault-martinez: `entry.name_record.is_leaf_record()`?

DaughterOfMars: Oh, I forgot there is something special about leaf expirations :) I'll change it then, thanks

thibault-martinez: All the new stuff is not really necessary anymore then

DaughterOfMars: Not necessary, but better to search by ID than iterate over all owned objects when we can

crates/iota/src/name_commands.rs
```diff
@@ -1888,6 +1930,18 @@ async fn get_proxy_nft_by_name(
...
 
+async fn get_proxy_nft_by_id(
+    nft_id: ObjectID,
+    domain: &Domain,
+    context: &mut WalletContext,
+) -> anyhow::Result<IotaNamesNftProxy> {
+    Ok(if domain.is_sld() {
+        IotaNamesNftProxy::Domain(get_owned_nft_by_id(nft_id, None, context).await?)
+    } else {
+        IotaNamesNftProxy::Subdomain(get_owned_nft_by_id(nft_id, None, context).await?)
```
Thoralf-M: This doesn't work for subdomain nodes?

DaughterOfMars: Aw, dang, you're right :c I hoped the NFT ID on the name record would be the wrapper but it's not

Thoralf-M: Yeah it's a bit annoying also for the tooling team, as there is no deterministic way to get from the NFT ID to the wrapper

thibault-martinez (approved):

PR #7514, opened 2025-06-24
chore(cli): better auction errors

### LINKED ISSUE #317: Better error when claiming an already claimed auction
```bash
thibault@Mac ~/i/iota (develop)> ./target/debug/iota name auction claim nounny.iota
[warn] Client/Server api version mismatch, client api version: 1.4.0-alpha, server api version: 1.2.2-rc
Error executing transaction '3c9XaRvnPnSbsZTzxC2YGLauKZnvqdESVGFcAtkUkV3d': 2nd command aborted within function '0x0000000000000000000000000000000000000000000000000000000000000002::dynamic_field::remove_child_object' at instruction 0 with code 1
thibault@Mac ~/i/iota (develop) [1]>
```

thibault-martinez:
# Description of change

Prints
```bash
auction for "tibo.iota" has already been claimed
```
instead of
```bash
Some(Deleted { object_id: 0x80bb58993010e80237313c895c0ad5a0148c270eae1d7c3eb0717608a375980f, version: SequenceNumber(5367), digest: o#7gyGAp71YXQRoxmFBaHxofQXAipvgHyBKPyxmdSJxyvz })
```

and
```bash
auction for "tibo.iota" does not exist
```

instead of
```bash
Some(NotExists { object_id: 0xfb1c0453a99750dd59a1bb208c06da0b8b1b00cd15cc9928e757a990a3009709 })
```

Fixes https://github.com/iotaledger/iota-names/issues/317


crates/iota/src/name_commands.rs
```diff
@@ -1888,7 +1882,10 @@ async fn get_proxy_nft_by_name(
     })
 }
 
-async fn get_registry_entry(domain: &Domain, client: &IotaClient) -> anyhow::Result<RegistryEntry> {
+async fn get_registry_entry(
+    domain: &Domain,
+    client: &IotaClient,
+) -> Result<RegistryEntry, RpcError> {
```
DaughterOfMars, line 1897: Why change the result?

thibault-martinez: Because the return of `get_object_from_bcs` changed but I guess I could map this one

DaughterOfMars: why did they change in the first place?

thibault-martinez: Because there is two layers of error coming from the API, an actual error variant and on success case an object that can contain an error. We can't properly match that wrapped error if we keep a generic anyhow error

DaughterOfMars: Then why don't we pull out the inner error before returning the object?

thibault-martinez: That's kind of what I'm doing? There are 2 sources or error so I'm combining them into one so the caller can properly handle it

DaughterOfMars: Yeah but why not just use anyhow?

thibault-martinez: Because then I can't match it anymore?

DaughterOfMars: I'm saying you would do the match here instead:
```
if let Some(error) = object_response.error {
    return Err(RpcError::IotaObjectResponse(error));
}
```

DaughterOfMars (approved):

Thoralf-M (approved):

PR #7403, opened 2025-06-16
feat(indexer): Do not reindex objects in checkpoint indexing, fix race conditions for display table

tomxey:
# Description of change

Do not reindex objects in checkpoint indexing, fix race conditions for display table

In general this PR makes sure that it's safe to index old transactions, in a sense that newer data will not be overwritten by older data. (with an exception of concurrent indexing, and indexing transactions for which dependencies are not yet indexed)

This PR is a first step to fix race conditions on `objects` table. To fix them completely https://github.com/iotaledger/iota/issues/7290 and https://github.com/iotaledger/iota/issues/6321 would need to be completed.
This PR fixes race conditions on `display` table completely.

## Links to any relevant issues

fixes https://github.com/iotaledger/iota/issues/7289

## How the change has been tested

`cargo test --profile simulator --package iota-indexer --test rpc-tests --all-features`

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage) - partially tested, race conditions on `display` table are fixed so they are tested, tests for `objects` table can be added later when remaining issues are closed

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

- [ ] Synchronization of the indexer from genesis for a network including migration objects.
- [ ] Restart of indexer synchronization locally without resetting the database.
- [ ] Restart of indexer synchronization on a production-like database.
- [ ] Deployment of services using Docker.
- [ ] Verification of API backward compatibility.

### Release Notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

Issue #7289, opened 2025-06-10
feat(indexer): disable reindexing of optimistic transactions


kodemartin:
## Description

Following #6320

> As discussed, there are three conditions to eliminate the race conditions altogether:
> 
> 1. Disable reindexing of the checkpoint transactions that have been indexed optimistically. Suffice it to check it there is an entry in `tx_global_order`. 
> 2. Ensure that if an optimistic transactions gets an `tx_global_order` entry, it will be (or will have been) indexed.
> 3. Ensure that all transaction dependencies of an optimistic transaction are indexed before proceeding with optimistic indexing. 

This task is for handling (1). That is we want to disable reindexing of checkpoint transactions that already have a row in `tx_global_order` from the optimistic execution path.

tomxey:
We are only concerned about the `objects` and `display` table, because changes to other tables are not causing race conditions.

Since checkpoint indexing commits transactions in batches and we are determining which transactions are already indexed at a time when batches are already formed, then we need a way to remove effects of a transactions from the batch.

The objects in a batch are a mixture of updated/deleted objects from all transactions in a batch.
To remove effects of given transaction from the batch we see what objects were output of given transaction, and then we remove the exact object versions from the set of objects to commit.

For display table it would be enough to always check version before updating the row, as part of the sql update query.

Prepared draft PR: https://github.com/iotaledger/iota/pull/7403

PR #7518, opened 2025-06-24
refactor(indexer): improve on disabling reindexing of optimistic transactions

kodemartin:
# Description of change

This patch follows #7403 to apply the following improvements:

* Retrieve the transaction digests that touched the objects while extracting data from `CheckpointData` to avoid lookups afterwards.
* Check if the associated transaction for an object is optimistically indexed at insertion time, avoiding an additional query on the application side to filter indexed transaction before the insert. This eliminates any race conditions that might stem from the two queries as well.

> [!NOTE]
> A transaction is considered as optimistically index if `tx_global_order.optimistic_sequence_number > 0`. 
>
> We add additional semantics to `optimistic_sequence_number` to avoid an extra column or different table altogether.

While implementing this, it was challenging to find an ergonomic and idiomatic way to introduce the changes without breaking the API, or adding incoherent types in the existing types that would further complicate proper versioning.

To this end, this patch also introduces the `rolling` module, as a means to introduce the **next** stable version of the indexer while maintaining backward compatibility to the existing one. This is experimental, and the versioning scheme is still to be specified, so please indulge me by not putting too much focus on its structure which is subject to change.

Part of #7289

Run the `v1.2.3` version of the indexer for a fixed set of checkpoint from a local network, and then ran the same set of checkpoints against the version of the indexer introduced with this patch. The resulting `objects` table were identical.

Other QA tests seemed redundant at this point.

### DIFF
--- a/crates/iota-indexer/src/models/objects.rs
+++ b/crates/iota-indexer/src/models/objects.rs
@@ -446,6 +446,72 @@ impl StoredObject {
     }
 }
 
+/// Holds a granular collection of [`StoredObject`] fields.
+///
+/// This can be used to consume an iterator of [`StoredObject`]
+/// to construct the collection of the individual fields.
+///
+/// ## Examples
+///
+/// ```ignore
+/// use iota_indexer::models::objects::{StoredObject, StoredObjects};
+/// fn construct_data() -> Vec<StoredObject> {
+///     Default::default()
+/// }
+///
+/// let mut stored_objects = StoredObjects::default();
+/// stored_objects.extend(construct_data());
+/// ```
+///
+/// Or using [`unzip`][std::iter::Iterator::unzip]:
+///
+/// ```ignore
+/// use iota_indexer::models::objects::{StoredObject, StoredObjects};
+/// use iota_types::digests::TransactionDigest;
+///
+/// fn construct_data() -> Vec<(StoredObject, TransactionDigest)> {
+///     Default::default()
+/// }
+///
+/// let (objects, digests): (StoredObjects, Vec<_>) = construct_data().into_iter().unzip();
+/// ```
+#[derive(Debug, Clone, Default)]
+pub(crate) struct StoredObjects {
+    pub(crate) object_ids: Vec<Vec<u8>>,
+    pub(crate) object_versions: Vec<i64>,
+    pub(crate) object_digests: Vec<Vec<u8>>,
+    pub(crate) owner_types: Vec<i16>,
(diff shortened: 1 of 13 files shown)


crates/iota-indexer/src/rolling/extract.rs
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+        latest_live_objects.into_values()
+    }
+
+    pub(crate) fn iter_removed_objects(
+        &'chk self,
+    ) -> impl Iterator<Item = (ObjectRef, TransactionDigest)> + 'chk {
+        let mut eventually_removed_object_refs = BTreeMap::new();
+        for tx in self.0.transactions.iter() {
```
tomxey: nit: could we extract self.0 to a variable, to give it some semantics?
here and above

crates/iota-indexer/src/models/objects.rs
```diff
@@ -446,6 +446,72 @@ impl StoredObject {
...
+/// use iota_types::digests::TransactionDigest;
+///
+/// fn construct_data() -> Vec<(StoredObject, TransactionDigest)> {
+///     Default::default()
+/// }
+///
+/// let (objects, digests): (StoredObjects, Vec<_>) = construct_data().into_iter().unzip();
+/// ```
+#[derive(Debug, Clone, Default)]
+pub(crate) struct StoredObjects {
```
tomxey, line 479: Seems the only purpose of this type is to simplify `persist_changed_objects`

Could we place it in the same file as `persist_changed_objects`? To treat it more like helper struct rather than general purpose model, since it's usage is quite specific.

kodemartin: It is a helper struct indeed, but it is also essential to storing mutliple objects using raw queries. In this sense it is used a "model" of multiple objects for the write path, and hence I believe that it fits the scope of this module.

It also relates closely to `StoreObject` as well, and maintaining them together seems rather intuitive.

crates/iota-indexer/src/rolling/extract.rs
kodemartin: https://github.com/iotaledger/iota/pull/7518/commits/a9bcd8a8b9095ab5934e168d3f7c25b3bd79152e

sergiupopescu199 (approved):

tomxey (approved):

kodemartin:
Localnet failure tracked by https://github.com/iotaledger/iota/issues/7560

PR #7520, opened 2025-06-24
feat(indexer): Make optimistic indexing commit all data in single transaction

### LINKED ISSUE #7431: feat(indexer): Modify optimistic indexing logic to insert all data in a single transaction
Created from: https://github.com/iotaledger/iota/issues/7290#issuecomment-2980252305

We need to modify optimistic indexing logic to insert all data (creation of tx_global_order, modifications of objects table, transactions, etc.) in a single transaction.
If such DB commit would try to write row to tx_global_order that already exists there, then the commit should fail and optimistic indexing of such tx should be skipped (because existence of row in `tx_global_order` means that checkpoint indexing started for such tx, so from now on checkpoint indexing is responsible for indexing this tx)

We could introduce a new metric to monitor how often optimistic indexing commit fails (we would expect it to be a very rare case)

tomxey:
# Description of change

This PR modifies optimistic indexing logic so that it commits all data related to execution of single TX inside of one DB transaction.
This means that the fact of optimistically indexing a TX is atomic and combined with unique columns check in tx_global_order table will ensure that given TX will either be indexed on optimistic path, or checkpoint path, but not both.

fixes https://github.com/iotaledger/iota/issues/7431

`cargo test --profile dev-nodebug --package iota-indexer  --features pg_integration -- --test-threads=1`
`cargo test --profile dev-nodebug --package iota-indexer --test rpc-tests --all-features`

The change was manually tested by adding artificial delays inside of optimistic indexing logic:
 - inside optimistic indexing DB transaction, but **before** tx global order is assigned - results in unique constraint violation, and optimistic indexing is aborted - this is as expected
 - inside optimistic indexing DB transaction, but **after** tx global order is assigned - results in checkpoint indexing waiting until optimistic indexing for given TX finishes - this is correct from data consistency perspective, although a bit unexpected, still this should be a rather rare case

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/optimistic_indexing.rs
+++ b/crates/iota-indexer/src/optimistic_indexing.rs
@@ -5,7 +5,7 @@ use std::{
     time::Duration,
 };
 
-use diesel::{OptionalExtension, RunQueryDsl, sql_query, sql_types};
+use diesel::{PgConnection, RunQueryDsl, result::DatabaseErrorKind, sql_query, sql_types};
 use downcast::Any;
 use fastcrypto::{encoding::Base64, error::FastCryptoError, traits::ToFromBytes};
 use iota_json_rpc_types::{IotaTransactionBlockResponse, IotaTransactionBlockResponseOptions};
@@ -22,7 +22,9 @@ use crate::{
     errors::IndexerError,
     handlers::{
         TransactionObjectChangesToCommit,
-        checkpoint_handler::{CheckpointHandler, try_extract_df_kind},
+        checkpoint_handler::{
+            CheckpointHandler, IndexedTransactionComponents, try_extract_df_kind,
+        },
     },
     indexer_reader::IndexerReader,
     metrics::IndexerMetrics,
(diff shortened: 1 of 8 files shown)


crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -172,68 +176,92 @@ impl OptimisticTransactionExecutor {
...
+        .or_else(|e| match e {
+            // This error means that checkpoint indexing was faster than the optimistic
+            // indexing. Let's just return and let checkpoint indexing handle
+            // the transaction.
+            IndexerError::PostgresUniqueTxGlobalOrderViolation(_) => Ok(()),
+            _ => Err(e),
+        })
+        .map_err(|e| {
+            IndexerError::PostgresWrite(format!("Failed to persist optimistic tx: {:?}", e))
+        })
```
kodemartin: Nit: You can simplify with `match`
```rust
        match tokio::task::spawn_blocking({
            let this: OptimisticTransactionExecutor = self.clone();
            let full_tx_data = full_tx_data.clone();
            move || this.index_transaction(&full_tx_data)
        })
        .await
        .map_err(|e| {
            tracing::error!("Failed to join optimistic index_transaction: {e}");
            IndexerError::from(e)
        })? {
            Ok(_) | Err(IndexerError::UniqueViolation) => Ok(()),
            Err(e) => Err(IndexerError:PostgresWrite)
        }
```

kodemartin (approved):
Looks good!

sergiupopescu199 (approved):

PR #7541, opened 2025-06-25
feat: Enable AA to dynamically access owned assets

### LINKED ISSUE #7440: [AA-prototype][account] Enable AA to dynamically access owned assets
## Problem
If an AA account is represented through a shared object, there should be a way to make its assets directly accessible or introspectable from the on-chain logic of the AA account itself. This is doable using dynamic fields and creating a logic for fields keys, e.g., using the child object id or type. This would allow to implement custom rules, such as budget caps (e.g., "don’t spend more than 1000 IOTA"), because the AA logic can dynamically access the assets it owns.

In this scenario, we have the UX hurdle of having the AA account shared object to receive each object transferred to this. In fact, when using the standard interface for transfers to objects, i.e., `iota_framework::transfer::public_transfer`, transfered objects can only be received and not automatically set as child of the receiving object.  
In the current design:
- objects owned by an IOTA address are explicitly passed as input to a transaction for being manipulated;
- objects owned by an Object ID are explicitly Received into a transaction for being manipulated;
- objects being child of another object, e.g., a dynamic object field, are implicitly passed as input to a transaction for being manipulated when the parent is explicitly passed as input.  

## Goal
Allow AA accounts implemented as shared objects to dynamically access their owned assets (e.g., balance, NFTs, tokens) for runtime introspection and enforcement of custom logic. Meanwhile, make so that AA account shared objects are treated differently from the rest of the objects and don't need to "receive" objects after these where transferred to them, but use them directly. 

### Others
- Do NOT focus on the object id/AA account id derivation. Right now, let's just assume that a random object id/AA account id is enough.
- Executing a PTB must involve the account’s shared object.
- Assure that locking the shared object effectively locks all assets owned by the account (should be already like this).

valeriyr:
# Description of change

> [!WARNING]  
> It is a prototype, not a final solution. It acts as a proof of concept.

Allow AA accounts implemented as shared objects to dynamically access their owned assets (e.g., balance, NFTs, tokens) for runtime introspection and enforcement of custom logic. Meanwhile, make so that AA account shared objects are treated differently from the rest of the objects and don't need to "receive" objects after these where transferred to them, but use them directly.

Let's imagine the following:
1. `Account Address` = magic(Sender, Salt, ...)
2. `Account Shared Object ID` = magic(Account Address)
3. We can execute a PTB `only` by using the related shared object because the verification and execution logic is inside its implementation.
4. It means that if we "lock" an account shared object, we "lock" all the assets that are owned by the related account address. On the logic level, it means that we can load them at runtime, like it is done with dynamic fields, so theoretically we can add native functions for this.
5. In this case, the on-chain account logic can have access to the owned assets.

fixes #7440

## Run the example

> [!NOTE]  
> All the following commands need to be executed on the proposal branch.

1. Run a local network:

```sh
cargo run -p iota -- start --force-regenesis --with-faucet
```

2. Request Gas from the faucet:

```sh
cargo run -p iota client faucet
```

3.  Create an `Account` shared object:

```sh
cargo run -p iota client call --package 0x2 --module account --function create
```

4. Run the `test` function:

```sh
iota client call --package 0x2 --module account --function test --args $ACCOUNT_ADDRESS $COIN_ADDRESS --gas-budget 9999999
```


iota-execution/latest/iota-move-natives/src/account.rs
```diff
@@ -0,0 +1,120 @@
...
+    // let address_asset_borrow_child_object_cost_params = context
+    //     .extensions_mut()
+    //     .get::<NativesCostTable>()
+    //     .address_asset_borrow_child_object_cost_params
+    //     .clone();
+    // native_charge_gas_early_exit!(
+    //     context,
+    //     address_asset_borrow_child_object_cost_params.
+    // address_asset_borrow_child_object_cost_base );
+
```
lzpap, line 81: I thought the first argument is at the bottom of the stack, that is here you'd pop `asset`, no?

From pop_arg comment:
```
/// Return the argument at the top of the stack.
///
/// Arguments are passed to a native as a stack with first arg at the bottom of
/// the stack. Calling this API can help in making the code more readable.
/// It's good practice to pop all arguments in locals of the native function on
/// function entry.
```

lzpap (commented):
I don't see the `test` function in `account.move` that I could execute to try it out. Can we grab the object owned address by value?

Additionally, we could only get the asset by id, right? So after all, when you construct the ptb you need to know the ids of the assets that you want to access, which is pretty similar to passing them as `Receiving` arguments. The only difference I guess is we could write rules inside the account as to when/what can be "received".

What do you think about "automatic receiving", meaning to put assets that are sent to accounts directly into a move structure inside the account, such that you can access `X amount of Y coins` owned by the account instead of having to know the ids of the objects that the account owns.

valeriyr:
> I don't see the `test` function in `account.move` that I could execute to try it out. Can we grab the object owned address by value?
> 
> Additionally, we could only get the asset by id, right? So after all, when you construct the ptb you need to know the ids of the assets that you want to access, which is pretty similar to passing them as `Receiving` arguments. The only difference I guess is we could write rules inside the account as to when/what can be "received".
> 
> What do you think about "automatic receiving", meaning to put assets that are sent to accounts directly into a move structure inside the account, such that you can access `X amount of Y coins` owned by the account instead of having to know the ids of the objects that the account owns.

@lzpap, thank you for the feedback.

Looks like you need to reload the PR page, the `test` function is added in the last push.

In this example, yes, you can get an asset only by ID, but we can consider other options, like by type.

Automatic receiving is a bit of a different concept, I think. In my example, we have a standard IOTA account and the related on-chain shared object. Assets are owned by the account address, not by an object; however, it is still possible.

PR #7545, opened 2025-06-25
chore: address clippy lints and bump rust toolchain to 1.87

DaughterOfMars:
## Description 

Ignores or in some cases fixes the new stable clippy `large_enum_variant` and `result_large_err` lints.

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -14,3 +14,4 @@ disallowed-methods = [
   # bincode::deserialize_from is easy to shoot your foot with
   { path = "bincode::deserialize_from", reason = "use bincode::deserialize instead" },
 ]
+large-error-threshold = 256
--- a/consensus/core/src/storage/store_tests.rs
+++ b/consensus/core/src/storage/store_tests.rs
@@ -13,6 +13,7 @@ use crate::{
 };
 
 /// Test fixture for store tests. Wraps around various store implementations.
+#[expect(clippy::large_enum_variant)]
 enum TestStore {
     RocksDB((RocksDBStore, TempDir)),
     Mem(MemStore),
--- a/crates/iota-core/src/authority_aggregator.rs
+++ b/crates/iota-core/src/authority_aggregator.rs
@@ -740,6 +740,7 @@ where
 
             type RequestResult<S> = Result<Result<S, IotaError>, tokio::time::error::Elapsed>;
 
+            #[expect(clippy::large_enum_variant)]
             enum Event<S> {
                 StartNext,
                 Request(AuthorityName, RequestResult<S>),
--- a/crates/iota-replay/src/fuzz.rs
+++ b/crates/iota-replay/src/fuzz.rs
@@ -183,7 +183,6 @@ impl ReplayFuzzer {
     }
 }
 
-#[expect(clippy::large_enum_variant)]
 #[derive(Debug, Error, Clone)]
 pub enum ReplayFuzzError {
     #[error(
--- a/crates/iota/src/client_ptb/ptb.rs
+++ b/crates/iota/src/client_ptb/ptb.rs
@@ -60,7 +60,7 @@ pub struct Summary {
 pub enum PTBCommandResult {
     Preview(PTBPreview),
     Summary(Summary),
-    CommandResult(IotaClientCommandResult),
+    CommandResult(Box<IotaClientCommandResult>),
     DevInspect(Box<DevInspectResults>),
     Json(serde_json::Value),
     Help { long: bool },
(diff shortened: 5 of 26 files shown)


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.86"
+channel = "1.87"
```
thibault-martinez, line 2: I remember that there is a second instance of the rust version to bump, a dockerfile or something

consensus/core/src/storage/store_tests.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
 };
 
 /// Test fixture for store tests. Wraps around various store implementations.
+#[expect(clippy::large_enum_variant)]
```
thibault-martinez, line 16: Why not box if private?

rust-toolchain.toml
DaughterOfMars: found it

consensus/core/src/storage/store_tests.rs
DaughterOfMars: It's a test thing so it just doesn't feel important

thibault-martinez: Fair enough

crates/iota-core/src/authority_aggregator.rs
```diff
@@ -740,6 +740,7 @@ where
 
             type RequestResult<S> = Result<Result<S, IotaError>, tokio::time::error::Elapsed>;
 
+            #[expect(clippy::large_enum_variant)]
             enum Event<S> {
```
thibault-martinez, line 744: private?

DaughterOfMars: yeah but only used in deserializing so no big deal

crates/iota-replay/src/fuzz.rs
```diff
@@ -183,7 +183,6 @@ impl ReplayFuzzer {
     }
 }
 
-#[expect(clippy::large_enum_variant)]
```
thibault-martinez, line 186: How come some are not required anymore? Because a nested type changed?

.clippy.toml
```diff
@@ -14,3 +14,4 @@ disallowed-methods = [
   # bincode::deserialize_from is easy to shoot your foot with
   { path = "bincode::deserialize_from", reason = "use bincode::deserialize instead" },
 ]
+large-error-threshold = 256
```
thibault-martinez, line 17: Did you just find a balance between increasing this and expecting some other?

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -60,7 +60,7 @@ pub struct Summary {
 pub enum PTBCommandResult {
     Preview(PTBPreview),
     Summary(Summary),
-    CommandResult(IotaClientCommandResult),
+    CommandResult(Box<IotaClientCommandResult>),
```
thibault-martinez, line 63: Breaking change?

crates/iota-replay/src/fuzz.rs
DaughterOfMars: Yeah it seems that way

.clippy.toml
DaughterOfMars: I thought the default was a little low, since only a few digests could surpass it. So I just doubled it and I think it's more reasonable now.

crates/iota/src/client_ptb/ptb.rs
DaughterOfMars: Yes but this is the CLI

thibault-martinez: So why did you expect `KeyToolCommand`?

DaughterOfMars: Because the commands themselves have to follow rules for the clap derive and I wasn't sure boxing the values would be a good idea.

thibault-martinez: Alright

thibault-martinez (approved):

crates/iota/src/name_commands.rs
```diff
@@ -1314,7 +1314,7 @@ pub enum NameCommandResult {
         burned: IotaNamesRegistration,
         digest: TransactionDigest,
     },
-    CommandResult(IotaClientCommandResult),
+    CommandResult(Box<IotaClientCommandResult>),
```
TheMrAI, line 1317: What is the reason for Boxing changes around the code? It is not apparent to me.

piotrm50 (approved):

crates/iota/src/name_commands.rs
DaughterOfMars: Enums are union types, which means that they must allocate the number of bytes of the largest variant. Boxing reduces the variant size to the size of a Box.

valeriyr (approved):

muXxer (approved):

kodemartin (approved):

Issue #6147, opened 2025-03-31
Port upstream node team related changes [mainnet-v1.38.4..mainnet-v1.39.4)


muXxer:

muXxer:
Merged in #7826

PR #7553, opened 2025-06-26
upstream(node): Add cache size params to ExecutionCacheConfig

jkrvivian:
# Description of change

- Port commit: https://github.com/MystenLabs/sui/commit/01275fb3199fbd15628c03288525d2ef6c353762

Part of #6147
  
### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): - Execution cache sizes can now be adjusted in the node config
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -290,22 +290,27 @@ pub enum ExecutionCacheConfig {
 
 impl Default for ExecutionCacheConfig {
     fn default() -> Self {
-        ExecutionCacheConfig::WritebackCache {
-            max_cache_size: None,
-            package_cache_size: None,
-            object_cache_size: None,
-            marker_cache_size: None,
-            object_by_id_cache_size: None,
-            transaction_cache_size: None,
-            executed_effect_cache_size: None,
-            effect_cache_size: None,
-            events_cache_size: None,
-            transaction_objects_cache_size: None,
-        }
+        ExecutionCacheConfig::default_writeback_cache()
     }
 }
 
 impl ExecutionCacheConfig {
+    /// Creates a default `WritebackCache` configuration with sensible defaults.
+    pub fn default_writeback_cache() -> Self {
+        ExecutionCacheConfig::WritebackCache {
+            max_cache_size: None,                 // defaults to 10000
+            package_cache_size: None,             // defaults to 1000
+            object_cache_size: None,              // defaults to max_cache_size
+            marker_cache_size: None,              // defaults to object_cache_size
+            object_by_id_cache_size: None,        // defaults to object_cache_size
+            transaction_cache_size: None,         // defaults to max_cache_size
+            executed_effect_cache_size: None,     // defaults to transaction_cache_size
+            effect_cache_size: None,              // defaults to executed_effect_cache_size
+            events_cache_size: None,              // defaults to transaction_cache_size
+            transaction_objects_cache_size: None, // defaults to 1000
+        }
+    }
+
     pub fn max_cache_size(&self) -> u64 {
         match self {
             ExecutionCacheConfig::PassthroughCache => fatal!("invalid cache config"),
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -307,127 +307,129 @@ impl Default for ExecutionCacheConfig {
 
 impl ExecutionCacheConfig {
     pub fn max_cache_size(&self) -> u64 {
-        std::env::var("IOTA_MAX_CACHE_SIZE")
-            .ok()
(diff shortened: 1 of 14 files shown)


crates/iota-core/src/execution_cache.rs
```diff
@@ -139,13 +139,14 @@ pub fn choose_execution_cache(config: &ExecutionCacheConfig) -> ExecutionCacheCo
 }
 
 pub fn build_execution_cache(
+    cache_config: &ExecutionCacheConfig,
     epoch_start_config: &EpochStartConfiguration,
     prometheus_registry: &Registry,
     store: &Arc<AuthorityStore>,
 ) -> ExecutionCacheTraitPointers {
     let execution_cache_metrics = Arc::new(ExecutionCacheMetrics::new(prometheus_registry));
```
muXxer, line 130: On upstream the code looks like this:
```
    match epoch_start_config.execution_cache_type() {
        ExecutionCacheConfigType::WritebackCache => ExecutionCacheTraitPointers::new(
            WritebackCache::new(cache_config, store.clone(), execution_cache_metrics).into(),
        ),
        ExecutionCacheConfigType::PassthroughCache => ExecutionCacheTraitPointers::new(
            ProxyCache::new(epoch_start_config, store.clone(), execution_cache_metrics).into(),
        ),
    }
```
otherwise we can't choose the passthrough cache anymore.

muXxer (changes requested):

crates/iota-core/src/execution_cache.rs
jkrvivian: This is the change from https://github.com/iotaledger/iota/pull/7485, we should hold this until 7485 get merged.

piotrm50 (approved):

crates/iota-benchmark/tests/simtest.rs
```diff
@@ -29,7 +29,10 @@ mod test {
             workload_configuration::{WorkloadConfig, WorkloadConfiguration, WorkloadWeights},
         },
     };
-    use iota_config::{AUTHORITIES_DB_NAME, IOTA_KEYSTORE_FILENAME, node::AuthorityOverloadConfig};
+    use iota_config::{
+        AUTHORITIES_DB_NAME, ExecutionCacheConfig, IOTA_KEYSTORE_FILENAME,
```
muXxer: I still don't see where this new import is used 😅

muXxer (changes requested):

crates/iota-benchmark/tests/simtest.rs
jkrvivian: I did a force push [7afc87](https://github.com/iotaledger/iota/commit/e7afc87212cb47de0535c9ce5126243687f8d6f9)

muXxer (approved):

crates/iota-config/src/node.rs
```diff
@@ -269,18 +270,167 @@ pub enum ExecutionCacheConfig {
...
     }
 }
 
+impl ExecutionCacheConfig {
+    pub fn max_cache_size(&self) -> u64 {
+        std::env::var("IOTA_MAX_CACHE_SIZE")
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .unwrap_or_else(|| match self {
+                ExecutionCacheConfig::PassthroughCache => fatal!("invalid cache config"),
```
semenov-vladyslav: Shouldn't matching on self and getting env var be swapped? Getting max cache size on passthrough cache should panic in any case independently of whether env var is set or not.

semenov-vladyslav (approved):
Looks good except for weird unwraps in `ExecutionCacheConfig`.

crates/iota-config/src/node.rs
muXxer: I kinda disagree.

The type of the `ExecutionCacheConfig` can only be set via config file (as far as I understood, see `build_execution_cache()`. So if a node operator changes cache settings by setting explicit environment variables, the node should notify him by a panic during the startup of the node, that something is wrong.

muXxer: Ah sorry I misunderstood. I thought you DON'T want to panic, but you mean it the other way around. I agree, we should always panic if someone calls "max_cache_size" for the Passthrough cache (and also in all other functions below).

jkrvivian: Updated in https://github.com/iotaledger/iota/pull/7553/commits/3d00c7aea4166742f2220589e4ae368d65db273d

semenov-vladyslav (approved):

PR #7557, opened 2025-06-26
feat(wallet): integrated new LedgerSigner

alexsporn:
I switched the wallet `LedgerSigner` to internally use the `LedgerSigner` from the `@iota/signers` pkg.

It probably makes sense to clean up the `WalletSigner` a bit to make it more compatible. The new `LedgerSigner` has separate signing functions for messages and transactions and the `WalletSigner` does the same thing and expects a generic `signData` function which already has a serialized `Intent`. The new `LedgerSigner` needs to differentiate because for a `Transaction` it will try to fetch all necessary objects to be able to clear-sign the transaction.

Alternatively we could keep the `signData` one and unpack the intent again, but a proper refactor makes more sense here.


apps/wallet/src/ui/app/background-client/backgroundServiceSigner.ts
```diff
@@ -26,8 +30,34 @@ export class BackgroundServiceSigner extends WalletSigner {
...
+        return {
+            bytes: toB64(input.message),
+            signature,
+        };
+    }
+
+    async signTransaction(input: {
+        transaction: Uint8Array | Transaction;
+    }): Promise<SignedTransaction> {
+        const bytes = await this.prepareTransaction(input.transaction);
```
alexsporn: maybe do this in the walletsigner so that any signer implementing the interface does not forget to do this?

msarcev (approved):

panteleymonchuk (approved):

PR #7563, opened 2025-06-27
fix(apps-ui-kit): remove unnecessary root light class and set default theme for MultiTheme

### LINKED ISSUE #7561: Bring back colorful loader in ui kit
dashboard loader: before multi theme pr  vs with multi theme pr

Image

Image

Juligs:
fixes #7561

### DIFF
--- a/apps/ui-kit/src/lib/styles/index.css
+++ b/apps/ui-kit/src/lib/styles/index.css
@@ -12,8 +12,7 @@
 @import 'tailwindcss/utilities';
 @import './utilities.css';
 
-:root,
-:root.dark {
+:root {
     /* FONTS */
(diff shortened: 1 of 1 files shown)


apps/ui-kit/src/lib/styles/index.css
```diff
@@ -12,7 +12,7 @@
 @import 'tailwindcss/utilities';
 @import './utilities.css';
 
-:root.light,
+:root,
 :root.dark {
```
marc2332: This could be simplified to just this I guess then right?
```suggestion
:root {
```

marc2332 (changes requested):

VmMad (approved):

marc2332 (approved):

PR #7567, opened 2025-06-27
feat(evm-bridge): add support for vUSD coin

### LINKED ISSUE #7479: Add VirtueUSD bridging support
As requested by the ecosystem team, we should add VirtueUSD bridging support

We should probably update our envs to be able to define allowed tokens on each network.

🙏🏼 Please update the confluence CI with any env updates

> These new tokens are "registered" by the EVM chain owner (us) so that a deposit via the bridge from L1 creates the respective ERC20 asset in EVM, and a withdrawal from EVM reverts that (burns the ERC20, releases the L1 asset)

Depends on https://github.com/iotaledger/iota-evm-bridge/issues/86

brancoder:
# Description of change

resolves #7479 

~~needs https://github.com/iotaledger/iota/pull/7640 to fix the tests~~
~~needs https://github.com/iotaledger/iota/pull/7749 to fix the tests~~

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


msarcev:
There is an additional scroll bar that appears when the coin selector is used (and shifts the screen slightly)
image

apps/core/src/components/VirtualList.tsx
```diff
@@ -56,7 +56,7 @@ export function VirtualList<T>({
         const el = containerRef.current;
         if (!el) return;
 
-        const resizeObserver = new ResizeObserver((entries) => {
+        const resizeObserver = new ResizeObserver((_entries) => {
```
msarcev, line 59: Is this change really needed ?

apps/evm-bridge/src/components/CoinSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { useSortedCoinsBalances } from '../hooks/useSortedCoinsBalances';
+
+export function CoinSelector() {
+    const { watch, setValue } = useFormContext<DepositFormData>();
+    const { coinType: selectedCoinType, isFromLayer1 } = watch();
+
+    const { sortedCoinsBalanceL1, sortedCoinsBalanceL2 } = useSortedCoinsBalances();
+
+    // Track previous list state
+    const prevListLengthRef = useRef(0);
```
msarcev: Nit: What list length ? I suggest `previousSelectedCoinBalanceLength`

apps/evm-bridge/src/components/CoinSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const prevIsFromLayer1Ref = useRef(isFromLayer1);
```
msarcev: Nit: `previousIsFromLayer1`

apps/evm-bridge/src/components/CoinSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    useEffect(() => {
+        const currentLength = sortedCoinsBalance.length;
+        const lengthChanged = currentLength !== prevListLengthRef.current;
+        const directionChanged = isFromLayer1 !== prevIsFromLayer1Ref.current;
+
+        // Only reset selection when list contents change or direction changes
+        if ((lengthChanged && currentLength > 0) || directionChanged) {
+            // Only update if there are coins to select and the desired coin exists
+            if (currentLength > 0 && sortedCoinsBalance[0]) {
```
msarcev: Isn't the `currentLength > 0` check in the outer condition enough ?

apps/evm-bridge/src/components/CoinSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+                const firstCoinType = sortedCoinsBalance[0].coinType;
+
+                // Only update if different from current selection
+                if (selectedCoinType !== firstCoinType) {
```
msarcev: What is this for exactly ?
We reset to first selected coin if the supported coin list from apps backend suddenly changes, is that it ?

apps/evm-bridge/src/hooks/useAllCoinsMetadata.ts
```diff
@@ -0,0 +1,135 @@
...
+    const queriesResults = useQueries({
+        queries: uniqueCoinTypes.map((coinType) => ({
+            queryKey: ['coin-metadata', coinType],
+            queryFn: async () => {
+                if (coinType === IOTA_TYPE_ARG) {
+                    return IOTA_COIN_METADATA;
+                }
+
+                try {
+                    const rpcData = await client.getCoinMetadata({ coinType });
```
msarcev: Does this ever work ?

apps/evm-bridge/src/lib/utils/transaction/createL1DepositTransaction.ts
```diff
@@ -0,0 +1,86 @@
...
+            : tx.splitCoins(primaryCoinInput, [amount]);
+
+        iscTx.placeCoinInBag({
+            bag,
+            coin,
+            coinType,
+        });
+    }
+
+    const transfers: [string, number | bigint][] = [[IOTA_TYPE_ARG, amount]];
```
msarcev: `number` type doesn't seem needed here.

apps/evm-bridge/src/hooks/useSortedCoinsBalances.ts
```diff
@@ -0,0 +1,71 @@
...
+    const { availableBalance: availableIotaBalanceL1, isLoading: isLoadingIotaL1 } =
+        useAvailableIotaBalanceL1();
+
+    const { data: coinsBalanceL1, isLoading: isLoadingAllBalancesL1 } =
+        useGetAllBalances(addressL1);
+    const { recognized: recognizedL1, pinned: pinnedL1 } = useSortedCoinsByCategories(
+        coinsBalanceL1 || [],
+        knownEvmCoins,
+    );
+    const sortedCoinsBalanceL1 = [...recognizedL1, ...pinnedL1];
```
msarcev: In what sense are these "sorted" ? Why do we need them sorted ?

apps/evm-bridge/src/lib/schema/bridgeForm.schema.ts
```diff
@@ -1,51 +1,92 @@
...
+            [BridgeFormInputName.IsDepositAddressManualInput]: z.boolean().default(false),
+            [BridgeFormInputName.CoinType]: z.string().default(IOTA_TYPE_ARG),
+            [BridgeFormInputName.ReceivingAddress]: z.string().trim(),
             [BridgeFormInputName.DepositAmount]: z
                 .string()
                 .trim()
                 .refine(
                     (value) => {
-                        return new BigNumber(value).isGreaterThanOrEqualTo(MINIMUM_SEND_AMOUNT);
+                        return new BigNumber(value).isGreaterThanOrEqualTo(0);
```
msarcev, line 27: We went from `0.1` to `0`. Are you sure this is ok ?

msarcev (changes requested):

apps/evm-bridge/src/hooks/useSortedCoinsBalances.ts
brancoder: Its the same approach we use int the wallet CoinSelector where we put "recognized" (iota) coins on top and other coins bellow it

apps/evm-bridge/src/lib/schema/bridgeForm.schema.ts
brancoder: this is a general condition now for all supported native coins and that is why we compare to 0. the iota specific condition is still there but its moved to the line https://github.com/iotaledger/iota/pull/7567/files#diff-5ecac873b227f408a26fd5a334669f8b93d76e32d4064f526bfcddfbc7e5a39bR71

msarcev:
Test gas values seam to need some adjustment.

apps/evm-bridge/src/components/CoinSelector.tsx
brancoder: this is not related to the supported coin list from apps backend. 
coin selection switches to the first coin from the list in 2 cases:

1. change direction (isFromLayer1) - so basically to select the first coin in the "new" form (ex. switching from L1->L2 to L2-L1) 

2.  user bridges the entire coin balance of selected coin -  in this case the selected coin in coin selector switches to the first coin in list because user no longer has any of the selected coins.

apps/core/src/components/VirtualList.tsx
brancoder: this was triggering build fail in the CI

msarcev (changes requested):
When testing with "fresh addresses" (IOTA address that wasn't used in EVM transactions and new EVM address with no balance):

- Sending any custom coins minted with `tooling-resources` goes through
- Sending VirtueUSD **does not**

apps/evm-bridge/src/components/CoinSelector.tsx
```diff
@@ -0,0 +1,69 @@
...
+            sortedCoins.find((coin) => coin.coinType === selectedCoinType)?.coinType ||
+            sortedCoins[0]?.coinType;
+
+        // Skip if first coin matches current selection
+        if (!coinTypeToSelect || coinTypeToSelect === selectedCoinType) return;
+
+        // Reset selection to first coin
+        setValue(BridgeFormInputName.CoinType, coinTypeToSelect, {
+            shouldValidate: true,
+            shouldTouch: true,
```
marc2332: Simplified
```suggestion
    const sortedCoins = isFromLayer1 ? sortedCoinsBalanceL1 : sortedCoinsBalanceL2;
    const sortedCoinsCoinTypes = sortedCoins.map(coin => coin.coinType);

    // Reset the selected coinType when the user coinTypes or the direction change
    useEffect(() => {
        // Get first coin as default selection
        const firstCoinType = sortedCoinsCoinTypes[0];

        // Skip if first coin matches current selection
        if (!firstCoinType || firstCoinType === selectedCoinType) return;

        // Reset selection to first coin
        setValue(BridgeFormInputName.CoinType, firstCoinType, {
            shouldValidate: true,
            shouldTouch: true,
        });
    }, [JSON.stringify(sortedCoinsCoinTypes), isFromLayer1, setValue]);
```

apps/evm-bridge/src/hooks/useAllCoinsMetadata.ts
```diff
@@ -0,0 +1,135 @@
...
+        });
+
+        return result;
+    }, [uniqueCoinTypes]);
+
+    return {
+        metadata: metadataMap,
+        isLoading: loadingMap,
+        isAnyLoading: queriesResults.some((query) => query.isLoading),
+        isAllLoaded: queriesResults.every((query) => query.isFetched),
```
marc2332: Simplified
```suggestion
    return useMemo(() => {
        const metadata: Record<string, CoinMetadata | null> = {};
        const isLoading: Record<string, boolean> = {};
        uniqueCoinTypes.forEach((coinType, index) => {
            const query = queriesResults[index];
            metadata[coinType] = query.data || null;
            isLoading[coinType] = query.isLoading;
        });
        return {
            metadata,
            isLoading,
        };
    }, [queriesResults]);
```

marc2332 (changes requested):

apps/core/src/components/VirtualList.tsx
msarcev: link ?

msarcev (commented):
- Why do we have so many `Balance` related hooks ? Here we should possibly simplify.
```
| | | useAllCoinsMetadata.ts
| | | useAvailableBalance.ts
| | | useAvailableIotaBalanceL1.ts
| | | useAvailableIotaBalanceL2.ts
| | | useBalance.ts
| | | useBuildL1DepositTransaction.ts
| | | useGasEstimateL2.ts
| | | useGetAllBalancesL2.ts
| | | useSortedCoinsBalances.ts

```
Also i suggest dropping the `All` from wording since everything is multi token now anyway.
- `useGetAllBalancesL2` takes an address but afaik we always use the current L2 address, so we can just do that inside the hook.

apps/evm-bridge/src/components/CoinSelector.tsx
msarcev: Don't commit through github, fix extra parenthesis at the bottom

marc2332 (approved):

msarcev (approved):

PR #7568, opened 2025-06-27
feat(starfish): evict transactions when last available committed dag is advanced

### LINKED ISSUE #7239: (starfish): consider eviction when the transactions are sequenced
Right now, eviction depends on which leader block is committed. Instead, it should (in addition) take into consideration the latest commited_sub_dag to avoid reading from memory when sending transactions out

polinikita:
# Description of change

We now track the round of latest committed leader block such that the corresponding sequenced transactions are locally available. Because of transaction data separation, block headers could be available and transactions not, we have two different methods to evict: for block headers, when a new pending committed dag is created; for transactions, when a new solid committed dag is created (all transaction data that is sequenced is locally available to us). 

Eviction of transactions is triggered when the required round is updated and the flush method is called

Fixes #7239 

### Release Notes

- [x] Protocol: Proper eviction of transactions
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/dag_state.rs
+++ b/crates/starfish/core/src/dag_state.rs
@@ -28,6 +28,7 @@ use crate::{
     },
     context::Context,
     leader_scoring::{ReputationScores, ScoringSubdag},
+    linearizer::MAX_LINEARIZER_DEPTH,
     storage::{Store, WriteBatch},
     threshold_clock::ThresholdClock,
 };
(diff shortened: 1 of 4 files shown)


crates/starfish/core/src/dag_state.rs
```diff
@@ -83,7 +85,12 @@ pub(crate) struct DagState {
...
     last_commit_round_advancement_time: Option<std::time::Instant>,
 
-    // Last committed rounds per authority.
+    // Round of the last committed leader which created a commit with available transactions. Does
+    // not persist across restarts and after recovery. All transactions below this round minus
+    // MAX_TRANSACTIONS_ACK_DEPTH minus MAX_LINEARIZER_DEPTH are evicted from memory.
+    last_available_commit_leader_round: Option<Round>,
+
+    // Rounds for latest blocks traversed by linearizer per authority.
     last_committed_rounds: Vec<Round>,
```
VorobyevIlya, line 95: I see that committed has been changed to traversed in several places. Is it because only leader blocks are committed, and non-leader blocks are traversed? Should we also change the variable name here too?

VorobyevIlya (approved):
looks good

crates/starfish/core/src/dag_state.rs
polinikita: I changed the description to make it clearer. Renaming of fields is also needed, but should be consistent, which is harder to achieve due to the extensive usage of the word commit everywhere

PR #7573, opened 2025-06-30
feat(indexer): mark checkpoint transactions as indexed

### LINKED ISSUE #7430: feat(indexer): Modify DB schema to store indexed property for each transaction
Created from https://github.com/iotaledger/iota/issues/7290#issuecomment-2980252305

We need to modify DB schema to have `indexed` property mapped to all rows in `tx_global_order`.

It needs to be decided in scope of this issue whether this column will be part of `tx_global_order` table, or it will be stored in a separate table.

### LINKED ISSUE #7432: feat(indexer): Modify checkpoint indexing logic to use the `indexed` property
Created from: https://github.com/iotaledger/iota/issues/7290#issuecomment-2980252305

The scope of this ticket is to use the property introduced in #7430 in checkpoint indexing pipeline.

Checkpoint indexing should set the property to `False` at the beginning of indexing at a time `tx_global_order` is assigned.
Checkpoint indexing should skip indexing of all txs that had the property set to `True` earlier.
When the transaction is indexed, the property should be set to `True`.

kodemartin:
# Description of change

This patch introduces an `IndexStatus` for checkpoint transactions that transitions from `Started -> Completed` after a transactions has been indexed. This is mediated through `tx_global_order.optimistic_sequence_number`.

Fixes #7430, #7432

### Infrastructure QA

This patch adds an additional db operation on top of existing indexing operations. As such it tested only through integration tests.

### DIFF
--- a/crates/iota-indexer/src/models/transactions.rs
+++ b/crates/iota-indexer/src/models/transactions.rs
@@ -52,8 +52,10 @@ pub struct TxGlobalOrder {
     /// Monotonically increasing number that represents the order
     /// of execution of optimistic transactions.
     ///
-    /// To maintain these semantics the value should be `0` for
-    /// checkpointed transactions.
+    /// To maintain these semantics the value should be non-positive for
+    /// checkpointed transactions. More specifically we allow values
+    /// in the set `[0, -1]` to represent the index status of checkpointed
+    /// transactions. See also [`CheckpointTxGlobalOrder`].
     ///
     /// Optimistic transactions should set this value to `None`,
     /// so that it is auto-generated on the database.
@@ -71,16 +73,14 @@ impl From<&IndexedTransaction> for CheckpointTxGlobalOrder {
     }
 }
 
+/// Stored value for checkpointed transactions.
+///
+/// Differs from [`TxGlobalOrder`] in that it allows values
+/// for `optimistic_sequence_number` in the set `[0, -1]`
+/// that represent the index status.
 #[derive(Clone, Debug, Queryable, Insertable, QueryableByName, Selectable)]
 #[diesel(table_name = tx_global_order)]
 pub(crate) struct CheckpointTxGlobalOrder {
-    /// Number that represents the global ordering between optimistic and
-    /// checkpointed transactions.
-    ///
-    /// Optimistic transactions will share the same number as checkpointed
-    /// transactions. In this case, ties are resolved by the
-    /// `(global_sequence_number, optimistic_sequence_number)` pair that
-    /// guarantees deterministic ordering.
     pub(crate) global_sequence_number: i64,
     pub(crate) tx_digest: Vec<u8>,
     /// The index status of checkpointed transactions.
--- a/crates/iota-indexer/src/models/transactions.rs
+++ b/crates/iota-indexer/src/models/transactions.rs
@@ -4,7 +4,14 @@
 
 use std::sync::Arc;
 
-use diesel::prelude::*;
+use diesel::{
+    backend::Backend,
+    deserialize::{FromSql, Result as DeserializeResult},
+    expression::AsExpression,
+    prelude::*,
+    serialize::{Output, Result as SerializeResult, ToSql},
+    sql_types::BigInt,
+};
 use iota_json_rpc_types::{
     BalanceChange, IotaEvent, IotaTransactionBlock, IotaTransactionBlockEffects,
     IotaTransactionBlockEvents, IotaTransactionBlockResponse, IotaTransactionBlockResponseOptions,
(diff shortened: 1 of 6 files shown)


crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -869,6 +872,49 @@ impl PgIndexerStore {
...
+                );
+                Ok::<(), IndexerError>(())
+            },
+            PG_DB_COMMIT_SLEEP_DURATION
+        )
+        .tap_ok(|_| {
+            let elapsed = guard.stop_and_record();
+            info!(
+                elapsed,
+                "Persisted {} chunked txs insertion order", num_transactions
```
tomxey: The log here and below could be updated as the function is not about insertion order.

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -2648,6 +2661,77 @@ impl IndexerStoreExt for PgIndexerStore {
...
+        let futures = chunks.into_iter().map(|c| {
+            self.spawn_blocking_task(move |this| {
+                this.update_status_for_checkpoint_transactions_chunk(c)
+            })
+        });
+
+        futures::future::try_join_all(futures)
+            .await
+            .map_err(|e| {
+                tracing::error!("Failed to join update_tx_insertion_order_chunk futures: {e}",);
```
tomxey: The log messages in this function should be updated as it's not about tx insertion order.

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -2648,6 +2661,77 @@ impl IndexerStoreExt for PgIndexerStore {
...
+
+        let chunks = chunk!(tx_order, self.config.parallel_chunk_size);
+        let futures = chunks
+            .into_iter()
+            .map(|c| self.spawn_blocking_task(move |this| this.persist_tx_global_order_chunk(c)));
+
+        futures::future::try_join_all(futures)
+            .await
+            .map_err(|e| {
+                tracing::error!("Failed to join persist_tx_insertion_order_chunk futures: {e}",);
```
tomxey: As above, it's not tx insertion order anymore.

crates/iota-indexer/src/models/transactions.rs
```diff
@@ -54,12 +61,69 @@ pub struct TxGlobalOrder {
...
-            optimistic_sequence_number: Some(0),
+            index_status: Some(IndexStatus::Started),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Queryable, Insertable, QueryableByName, Selectable)]
+#[diesel(table_name = tx_global_order)]
+pub(crate) struct CheckpointTxGlobalOrder {
+    /// Number that represents the global ordering between optimistic and
```
tomxey: Maybe let's not repeat all the docstrings from the previous model, but just mention what are the differences, and why do we have a second model?

Now it's purpose is a bit confusing on the first glance, and the docstrings are more prone to getting outdated.

crates/iota-indexer/src/models/transactions.rs
```diff
@@ -54,12 +61,69 @@ pub struct TxGlobalOrder {
...
+    ///
+    /// Optimistic transactions will share the same number as checkpointed
+    /// transactions. In this case, ties are resolved by the
+    /// `(global_sequence_number, optimistic_sequence_number)` pair that
+    /// guarantees deterministic ordering.
+    pub(crate) global_sequence_number: i64,
+    pub(crate) tx_digest: Vec<u8>,
+    /// The index status of checkpointed transactions.
+    ///
+    /// This essentially reserves the values `[0, -1]` of the
```
tomxey, line 88: I think we should mention in the docstring for TxGlobalOrder that -1 value also is allowed and what is it's meaning. Right now that docstring is a bit outdated.

kodemartin: df8d5d8443

crates/iota-indexer/src/models/transactions.rs
kodemartin: df8d5d8443

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: df8d5d8443

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: df8d5d8443

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: df8d5d8443

tomxey (approved):

samuel-rufi (approved):

Issue #6148, opened 2025-03-31
Port upstream node team related changes [mainnet-v1.39.4..mainnet-v1.40.3)


muXxer:

muXxer:
Merged with #7885

PR #7578, opened 2025-06-30
upstream(core): Add support for callbacks on body chunks

jkrvivian:
# Description of change

- Upstream range: [v1.39.4, v1.40.3)
- Port commit: https://github.com/MystenLabs/sui/commit/9a50f393298d2aa038a674b107081be39760eb5d
- Description:
Add support for callbacks on body chunks. This also has the consequence that the callback handlers are now held until the response stream has concluded, meaning we'll be able to have a more accurate measurement of inflight requests than we previously did.

Part of #6148   

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,7 +221,7 @@ async-recursion = "1.0.4"
 async-trait = "0.1.61"
 aws-config = "1.5.6"
 aws-sdk-dynamodb = "1.42"
-axum = { version = "0.8", default-features = false, features = ["tokio", "http1", "http2", "json", "matched-path", "original-uri", "form", "query", "ws", "macros"] }
+axum = { version = "0.8", default-features = false, features = ["tokio", "http2", "json", "matched-path", "original-uri", "form", "query", "ws", "macros"] }
 axum-extra = { version = "0.10", features = ["typed-header"] }
 axum-server = { git = "https://github.com/bmwill/axum-server.git", rev = "f44323e271afdd1365fd0c8b0a4c0bbdf4956cb7", version = "0.6", default-features = false, features = ["tls-rustls"] }
 backoff = { version = "0.4.0", features = ["futures", "futures-core", "pin-project-lite", "tokio", "tokio_1"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -274,7 +274,7 @@ http = "1"
 http-body = "1"
 humantime = "2.1.0"
 hyper = "1"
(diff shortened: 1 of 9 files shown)


muXxer (approved):

consensus/core/src/network/tonic_network.rs
```diff
@@ -1010,12 +1010,12 @@ impl MakeCallbackHandler for MetricsCallbackMaker {
 }
 
 impl ResponseHandler for MetricsResponseCallback {
-    fn on_response(self, response: &http::response::Parts) {
-        self.on_response(response)
+    fn on_response(&mut self, response: &http::response::Parts) {
+        MetricsResponseCallback::on_response(self, response)
```
semenov-vladyslav, line 1014: `self.on_response(response)` will resolve to `MetricsResponseCallback::on_response`, there'll be no infinite loop, so we can keep the old code.

consensus/core/src/network/tonic_network.rs
```diff
@@ -1010,12 +1010,12 @@ impl MakeCallbackHandler for MetricsCallbackMaker {
...
     }
 
-    fn on_error<E>(self, err: &E) {
-        self.on_error(err)
+    fn on_error<E>(&mut self, err: &E) {
+        MetricsResponseCallback::on_error(self, err)
```
semenov-vladyslav, line 1018: Same here.

Cargo.toml
```diff
@@ -274,7 +274,7 @@ http = "1"
 http-body = "1"
 humantime = "2.1.0"
 hyper = "1"
-hyper-rustls = { version = "0.27", default-features = false, features = ["webpki-roots", "http2", "ring", "tls12"] }
+hyper-rustls = { version = "0.27", default-features = false, features = ["webpki-roots", "http1", "http2", "ring", "tls12"] }
```
semenov-vladyslav, line 277: Why do we need `"http1"` here? To enable http1 `enable_http1()` [should be called](https://docs.rs/hyper-rustls/latest/hyper_rustls/struct.HttpsConnectorBuilder.html#method.enable_http1).

semenov-vladyslav (approved):
Looks good, but a few minor comments.

piotrm50 (approved):

TheMrAI (approved):

consensus/core/src/network/tonic_network.rs
jkrvivian: Thanks for checking this! But I'll prefer to keep the upstream changes.

PR #7579, opened 2025-06-30
feat: first release, move tests FAILS

### LINKED ISSUE #7441: [AA-prototype][account] Create a new native function for generating a precomputed object id
## Problem
In the current design, creating a [new](https://github.com/iotaledger/iota/blob/c237b7e9f6dd882027d194132911c36a2551c885/crates/iota-framework/packages/iota-framework/sources/object.move#L173) object means invoking the [derive_id](https://github.com/iotaledger/iota/blob/c237b7e9f6dd882027d194132911c36a2551c885/crates/iota-framework/packages/iota-framework/sources/tx_context.move#L76) native function. This function creates an id based on the transaction hash digest, thus very difficult (if not impossible) to precompute.
This is an issue for AA because, in the case in which an AA account is represented through a shared object, then the AA id cannot be precomputed (this feature would enable different use cases and enhance user onboarding and scalability). 
This is the same problem faced in Ethereum when proposing the ERC4337 that led to the use of the [CREATE2 opcode](https://docs.openzeppelin.com/cli/2.8/deploying-with-create2).

## Goal
Add a new [native function similar to derive_id](https://github.com/iotaledger/iota/blob/c237b7e9f6dd882027d194132911c36a2551c885/iota-execution/latest/iota-move-natives/src/tx_context.rs#L29) such that object ids can be precomputed, but won't have a collision with "normal" object ids and test it in Move.

This new object id kind could be derived as:
```
hash(flag, iota_address, salt)
```
where:
- `flag` is a fixed byte that allows to avoid object id collisions (check [`HashingIntentScope`](https://github.com/iotaledger/iota/blob/c237b7e9f6dd882027d194132911c36a2551c885/crates/shared-crypto/src/intent.rs#L196)) -> could be `0xf2`;
- `iota_address` can be any IOTA address, i.e., an address derived from a private key or another object id; it is not necessarily the sender address;
- `salt` is an arbitrary value provided by the sender

theiari:
It seems there're some issues regarding linking between rust and move. Not entirely sure on how should I properly test the move native, since `iota move test` fails.

I tried following mainly `tx_context` flow to add:

- move module `deterministic_object_id.move`
- move native `derive_id_with_salt` 
- rust function `derive_id_with_salt` someway similar to`[`derive_id`](https://github.com/iotaledger/iota/blob/develop/crates/iota-types/src/base_types.rs#L1301)

Ahead of this current commit, I also tried

- Adding a few missing lines to `iota-framework` on [lib.rs](https://github.com/iotaledger/iota/blob/develop/crates/iota-framework/src/lib.rs) related to the native function.
- Followed this README: https://github.com/iotaledger/iota/tree/d17b7f77ee7d3217a4284db24259301c4110f716/crates/iota-framework, even though `all_natives` does not exist.
- Just in case, I also tried running `update_all_snapshots.sh` in [`iota/scripts/`](https://github.com/iotaledger/iota/blob/develop/scripts/update_all_snapshots.sh).

However, when I try to run `iota move test`, around **300 out of 500 tests** fail with the error:

UNEXPECTED_VERIFIER_ERROR (code 2017) and some reference to the native function I am trying to add.

### DIFF
--- a/crates/iota-framework/packages/iota-framework/sources/object.move
+++ b/crates/iota-framework/packages/iota-framework/sources/object.move
@@ -25,6 +25,7 @@ public use fun uid_to_address as UID.to_address;
 
 /// Allows calling `.to_bytes` on a `UID` to get a `vector<u8>`.
 public use fun uid_to_bytes as UID.to_bytes;
+use iota::deterministic_object_id;
 
 /// The hardcoded ID for the singleton IOTA System State Object.
 const IOTA_SYSTEM_STATE_OBJECT_ID: address = @0x5;
@@ -174,6 +175,14 @@ public fun new(ctx: &mut TxContext): UID {
     UID {
         id: ID { bytes: ctx.fresh_object_address() },
     }
+}
+
+/// Create a new object with a precomputed objectId.
+/// Using a fixed 123 to avoid collisions
+public fun new_precomputed(iota_address: address, salt: vector<u8>): UID {
+    let id = deterministic_object_id::precomputed_object_id(iota_address, &salt);
+    record_new_uid(id);
+    UID { id: ID { bytes: id } }
 }
 
 /// Delete the object and it's `UID`. This is the only way to eliminate a `UID`.
--- a/crates/iota-framework/packages/iota-framework/sources/object.move
+++ b/crates/iota-framework/packages/iota-framework/sources/object.move
@@ -48,6 +48,9 @@ const IOTA_BRIDGE_ID: address = @0x9;
 /// Sender is not @0x0 the system address.
 const ENotSystemAddress: u64 = 0;
 
+/// u64 flag used to produce precomputed object IDs.
+const PRECOMPUTED_OBJECT_ID_FLAG: u64 = 123;
+
 /// An object ID. This is used to reference IOTA Objects.
 /// This is *not* guaranteed to be globally unique--anyone can create an `ID` from a `UID` or
 /// from an object, and ID's can be freely copied and dropped.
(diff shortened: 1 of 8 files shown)


Dkwcs:
We have to rebuild the `iota` crate and run move test based on it(just `cargo build` will be enough).
If we run `move test` with the new iota binary the tests will pass.

Then if we run `UPDATE=1 cargo insta test` we have some errors:

```
   │  
37 │ ╭ /// Create an `address` that has not been used. As it is an object address, it will never
38 │ │ /// occur as the address for a user.
   │ ╰────────────────────────────────────^ Documentation comment cannot be matched to a language item
```
   
```
38 │ ╭ /// Create an `address` that has not been used. As it is an object address, it will never
39 │ │ /// occur as the address for a user.
40 │ │ /// In other words, the generated address is a globally unique object ID.
   │ ─────────────────────────────────^ Documentation comment cannot be matched to a language item
```

I guess we can handle it somehow, just to know how to write Move comment correctly

theiari:
Not sure about `native fun derive_id_with_salt(...)`visibility. 

Right now it's `public` but maybe it's better to have it `private`?

crates/iota-framework/packages/iota-framework/sources/object.move
```diff
@@ -176,6 +180,14 @@ public fun new(ctx: &mut TxContext): UID {
     }
 }
 
+/// Create a new object with a precomputed objectId.
+/// Using a fixed 123 to avoid collisions
+public fun new_precomputed(iota_address: address, salt: vector<u8>): UID {
```
lzpap, line 185: this function creates an UID, therefore it needs to ensure that each ID is actually unique. For the `new` function above, uniqueness is ensured via the tx context, for accounts, we need to gate access to this function via the account creation flow.

Therefore, this function should be public(package), such that it canbe called from the (future) account framework module but not by any outsider package.

lzpap: +1: in the account creation flow we need to check a global table of existing accounts

PR #7582, opened 2025-06-30
feat(CLI): Support IOTA Names instead of addresses in CLI commands

### LINKED ISSUE #321: CLI: resolve IOTA-Names addresses
Should they be resolved globally or only within the name subcommand?
Local alias address should have the priority?

DaughterOfMars:
## Description 

Adds support for using IOTA Names instead of addresses throughout the CLI (anywhere that currently uses `KeyIdentity`).

```
./target/debug/iota client balance thoralf.iota
╭────────────────────────────────────────╮
│ Balance of coins owned by this address │
├────────────────────────────────────────┤
│ ╭────────────────────────────────────╮ │
│ │ coin  balance (raw)  balance       │ │
│ ├────────────────────────────────────┤ │
│ │ IOTA  143812600399   143.81 IOTA   │ │
│ ╰────────────────────────────────────╯ │
╰────────────────────────────────────────╯
./target/debug/iota client balance chloe.iota
no target address set for chloe.iota
```

## Related Issues

Closes https://github.com/iotaledger/iota-names/issues/321

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: Add support for IOTA-Names address resolution in command arguments.
- [ ] Rust SDK:


crates/iota/src/key_identity.rs
```diff
@@ -31,6 +33,10 @@ impl FromStr for KeyIdentity {
         if s.starts_with("0x") {
             Ok(KeyIdentity::Address(IotaAddress::from_str(s)?))
         } else {
+            #[cfg(feature = "iota-names")]
+            if let Ok(name) = s.parse() {
```
thibault-martinez, line 37: Isn't the `KeyIdentity::Address` parsing above going to fail for names that start with `0x` like `0xff.iota` ?

DaughterOfMars: true, I'll fix it

crates/iota/src/key_identity.rs
```diff
@@ -41,19 +47,32 @@ impl Display for KeyIdentity {
         let v = match self {
             KeyIdentity::Address(x) => x.to_string(),
             KeyIdentity::Alias(x) => x.to_string(),
+            KeyIdentity::Name(x) => x.to_string(),
```
thibault-martinez, line 51: No feature gate?

crates/iota/src/key_identity.rs
```diff
@@ -41,19 +47,32 @@ impl Display for KeyIdentity {
...
 ) -> Result<IotaAddress, Error> {
     if let Some(addr) = input {
-        get_identity_address_from_keystore(addr, ctx.config().keystore())
+        match addr {
+            KeyIdentity::Address(x) => Ok(x),
+            KeyIdentity::Alias(x) => Ok(*ctx.config().keystore().get_address_by_alias(x)?),
+            #[cfg(feature = "iota-names")]
+            KeyIdentity::Name(name) => {
+                let client = ctx.get_client().await?;
+                let entry = crate::name_commands::get_registry_entry(&name, &client).await?;
```
thibault-martinez: Why not use the lookup endpoint?

crates/iota/src/key_identity.rs
DaughterOfMars: oop

crates/iota/src/key_identity.rs
DaughterOfMars: well this is what the CLI does right now so I just copied that

Thoralf-M: I also prefer that until we have a solution for local networks

Thoralf-M (approved):

thibault-martinez (commented):
This PR adds IOTA-Names support for `KeyIdentity` but there are probably lots of places not using `KeyIdentity` in the first place, would you do that in another PR?

DaughterOfMars:
> This PR adds IOTA-Names support for `KeyIdentity` but there are probably lots of places not using `KeyIdentity` in the first place, would you do that in another PR?

I added a few more places that were missed in name_commands. There are three more places where this could be used but I'm not sure they are needed.

1. `iota validator display-metadata`
2. `iota validator report-validator`
3. `iota validator serialize-payload-pop`

Thoralf-M:
What about `iota client call` and `iota client ptb`? They don't use KeyIdentity

thibault-martinez:
`IotaCommand::{Start, Genesis}` as well, but I wouldn't necessarily do it, wdyt?

thibault-martinez:
> What about `iota client call` and `iota client ptb`? They don't use KeyIdentity

Wouldn't these require changing the parser? If yes, not sure I'd do it (now)

Thoralf-M:
> Wouldn't these require changing the parser? If yes, not sure I'd do it (now)

Yes I think so, also fine not to do them in this PR, just wanted to mention it
And yeah no point in supporting it for `IotaCommand::{Start, Genesis}`

thibault-martinez:
> > Wouldn't these require changing the parser? If yes, not sure I'd do it (now)
> 
> Yes I think so, also fine not to do them in this PR, just wanted to mention it And yeah no point in supporting it for `IotaCommand::{Start, Genesis}`

Same goes for these then? Not worth it?

iota validator display-metadata
iota validator report-validator
iota validator serialize-payload-pop

thibault-martinez (changes requested):
Clippy failing

thibault-martinez:
conflicts

thibault-martinez:
What do you think about `NameCommand::Register::set_target_address`?

DaughterOfMars:
> What do you think about `NameCommand::Register::set_target_address`?

I already supported it

thibault-martinez:
> > What do you think about `NameCommand::Register::set_target_address`?
> 
> I already supported it

Ah so the parameter is a `String` because of `default_missing_value` ?

crates/iota/src/key_identity.rs
```diff
@@ -28,9 +30,13 @@ impl From<IotaAddress> for KeyIdentity {
 impl FromStr for KeyIdentity {
     type Err = anyhow::Error;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s.starts_with("0x") {
-            Ok(KeyIdentity::Address(IotaAddress::from_str(s)?))
+        if let Ok(address) = s.parse() {
+            Ok(KeyIdentity::Address(address))
         } else {
+            #[cfg(feature = "iota-names")]
+            if let Ok(name) = s.parse() {
```
thibault-martinez, line 37: An alias could technically have the shape of an IOTA-Names but I guess it's asking for troubles anyway

crates/iota/src/key_identity.rs
```diff
@@ -41,19 +47,33 @@ impl Display for KeyIdentity {
...
 /// Get the IotaAddress corresponding to this key identity.
 /// If no string is provided, then the current active address is returned.
-pub fn get_identity_address(
+pub async fn get_identity_address(
     input: Option<KeyIdentity>,
-    ctx: &mut WalletContext,
+    ctx: &WalletContext,
 ) -> Result<IotaAddress, Error> {
     if let Some(addr) = input {
-        get_identity_address_from_keystore(addr, ctx.config().keystore())
```
thibault-martinez, line 56: Why are we not replacing the calls to `get_identity_address_from_keystore` with `get_identity_address`?

thibault-martinez: We didn't want to handle IOTA-Names in the Keytool?

crates/iota/src/key_identity.rs
DaughterOfMars: I mean I could possibly handle the fall through there

crates/iota/src/key_identity.rs
DaughterOfMars: it takes the context but this command only uses the keytool

thibault-martinez (approved):

crates/iota/src/name_commands.rs
```diff
@@ -98,8 +98,8 @@ pub enum NameCommand {
         coin: Option<ObjectID>,
         /// The address or alias to which the name will point. If the flag is
         /// specified without a value, the current active address will be used.
-        #[arg(long, require_equals = true, default_missing_value = "", num_args = 0..=1)]
```
thibault-martinez, line 101: 100% sure non of these are required anymore?

crates/iota/src/key_identity.rs
thibault-martinez: How would that look like?

crates/iota/src/name_commands.rs
DaughterOfMars: Yeah I tested it :)

DaughterOfMars: Even the = isn't needed anymore :)))

thibault-martinez (approved):

Thoralf-M (approved):

PR #7589, opened 2025-06-30
feat(ci): always use the latest test scripts from commit that triggered the workflow

muXxer:
# Description of change

This PR changes the nightly workflow in case an `IOTA_REF` was given in the input parameters.
The tests will still checkout the given `IOTA_REF`, but the simtest script is used from the commit that triggered the workflow.

This enables us to update the simtest script, but test older versions of the code.

It also adds some boolean flags, so we can skip jobs we don't need while investigating issues.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -111,5 +111,9 @@ jobs:
       - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           ref: ${{ env.IOTA_REF }}
+      - name: Get latest simtest script from commit that triggered this workflow
+        run: |
+          git fetch origin ${{ github.sha }}
+          git checkout ${{ github.sha }} -- scripts/simtest/simtest-run.sh
       - name: Run simtest
         run: scripts/simtest/simtest-run.sh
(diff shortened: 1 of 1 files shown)


alexsporn (approved):

.github/workflows/nightly.yml
```diff
@@ -69,29 +81,39 @@ jobs:
...
   split-cluster:
+    if: ${{ !inputs.rustTestsOnly && !inputs.simTestsOnly }}
     uses: ./.github/workflows/_split_cluster.yml
 
   rosetta:
+    if: ${{ !inputs.rustTestsOnly && !inputs.simTestsOnly }}
     uses: ./.github/workflows/_rosetta.yml
 
   simtest:
+    if: ${{ inputs.simTestsOnly || (!inputs.rustTestsOnly && !inputs.simTestsOnly) }}
```
DaughterOfMars: can be reduced

```suggestion
    if: ${{ inputs.simTestsOnly || !inputs.rustTestsOnly }}
```

DaughterOfMars: same above

DaughterOfMars:
I'm a little confused as the changesets don't seem to match the title/description

DaughterOfMars (approved):

PR #7595, opened 2025-07-01
feat(indexer): wait for transaction dependencies

### LINKED ISSUE #6321: [indexer]: Handle case when optimistic indexing encounters transaction that depends on not yet indexed transactions
We need to decide what to do in case optimistic indexing encounters transaction that depends on not yet indexed transactions.

Should we wait for the dependencies? Or return an error?

How is the node handling it? Is it e.g. guaranteed that transactions for the same shared object are ordered in the same way for every fullnode?

kodemartin:
# Description of change

This patch introduces logic to wait for the transaction dependencies while executing optimistically.

It also adds exponential backoff for checking the index status of the dependencies, to accommodate concurrent optimistic execution of the dependencies (e.g. in the case of shared objects).

Fixes #6321 

No additional tests are considered necessary, as this affects the WriteAPI of the json-rpc server.

### DIFF
--- a/crates/iota-indexer/src/indexer_reader.rs
+++ b/crates/iota-indexer/src/indexer_reader.rs
@@ -630,33 +630,79 @@ impl IndexerReader {
         stored_txn.try_into_iota_transaction_effects()
     }
 
-    /// Count how many of the given transactions have been indexed.
+    /// Expensive check to assert whether all transactions
+    /// are indexed.
+    ///
+    /// It uses both the `tx_global_order` table
+    /// and the `checkpoints` table to cover old transactions.
+    fn deep_check_all_transactions_are_indexed(
+        &self,
+        digests: &[TransactionDigest],
+    ) -> Result<bool, IndexerError> {
+        let (checkpointed, optimistic) = self
+            .multi_get_transactions(digests)?
+            .into_iter()
+            .partition::<Vec<_>, _>(|tx| tx.checkpoint_sequence_number >= 0);
+        if !optimistic.is_empty() {
+            let num_optimistic = optimistic.len();
+            let optimistic_digests = optimistic
+                .into_iter()
+                .map(|tx| tx.transaction_digest)
+                .collect::<HashSet<_>>();
+            let num_indexed =
+                self.count_indexed_tx_global_orders(optimistic_digests.into_iter())?;
+            if num_indexed as usize != num_optimistic {
+                return Ok(false);
+            }
+        }
+        let Some(max_transaction_checkpoint) = checkpointed
+            .iter()
+            .map(|tx| tx.checkpoint_sequence_number)
+            .max()
+        else {
+            return Ok(false);
+        };
+        Ok(self
+            .get_checkpoint(CheckpointId::SequenceNumber(
+                max_transaction_checkpoint as u64,
+            ))?
+            .is_some())
+    }
+
+    pub(crate) async fn deep_check_all_transactions_are_indexed_in_blocking_task(
+        &self,
+        digests: Vec<TransactionDigest>,
+    ) -> Result<bool, IndexerError> {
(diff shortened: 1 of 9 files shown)


crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -630,6 +630,35 @@ impl IndexerReader {
...
+    /// Any transaction with a non-zero optimistic sequence number
+    /// is considered as indexed.
+    fn count_indexed_transactions(
+        &self,
+        digests: &[TransactionDigest],
+    ) -> Result<i64, IndexerError> {
+        let digests = digests
+            .iter()
+            .map(|digest| digest.inner().to_vec())
+            .collect::<HashSet<_>>();
```
tomxey: nit: I feel Vec would be enough, since it will most likely only be iterated over to generate some sql.

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -630,6 +630,35 @@ impl IndexerReader {
...
+        run_query!(&self.pool, |conn| {
+            tx_global_order::table
```
tomxey, line 693: should we somehow handle old transactions, that are indexed but don't have entry in tx_global_order?

kodemartin: ab6e5041ea

crates/iota-indexer/src/indexer_reader.rs
kodemartin: We still want to deduplicate digests as we rely on matching the number of rows to the number of given transactions. Transferred the conversion on a higher level though

ab6e5041ea

crates/iota-indexer/src/indexer_reader.rs
tomxey: Related comment, for visibility: https://github.com/iotaledger/iota/commit/ab6e5041ea1c005b8922ac6e60e84ae066fda45b#r161268043

kodemartin: Yep as discussed I opted in performing the check concurrently
https://github.com/iotaledger/iota/pull/7595/commits/25d0dac047ba5fd9d8a5a680ded10f8b0d259691

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -630,6 +630,81 @@ impl IndexerReader {
...
+    /// are indexed.
+    ///
+    /// It uses both the `tx_global_order` table
+    /// and the `checkpoints` table to cover old transactions.
+    fn deep_check_all_transactions_are_indexed(
+        &self,
+        digests: &[TransactionDigest],
+    ) -> Result<bool, IndexerError> {
+        let (checkpointed, optimistic) = self
+            .multi_get_transactions(digests)?
```
tomxey: Isn't it possible that some transactions will not be returned from this call at all?
I think in such case we should also return false right away.

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -630,6 +630,81 @@ impl IndexerReader {
...
+            if num_indexed as usize != num_optimistic {
+                return Ok(false);
+            }
+        }
+        let Some(max_transaction_checkpoint) = checkpointed
+            .iter()
+            .map(|tx| tx.checkpoint_sequence_number)
+            .max()
+        else {
+            return Ok(false);
```
tomxey: Actually, if we have no checkpointed transactions to check, shouldn't we return true?

kodemartin: true, good catch: 6270a5c594

crates/iota-indexer/src/indexer_reader.rs
kodemartin: 6270a5c594

tomxey (approved):

sergiupopescu199 (approved):

PR #7603, opened 2025-07-01
feat(indexer): Introduce backfilling infrastructure

### LINKED ISSUE #7365: upstream(indexer): introduce backfilling infrastructure
## Description

Port `backfill` utilities from upstream filtering and consolidating the following patches

* [indexer: SQL Backfill command (#19359)](https://github.com/MystenLabs/sui/pull/19359) {"side-effects": "major", "type": "feature"}
* [Indexer: A few improvements to backfill tool (#19441)](https://github.com/MystenLabs/sui/pull/19441)
* [Indexer: Add a generic backfill template](https://github.com/MystenLabs/sui/pull/19510)
* [Indexer: Add ingestion based backfill](https://github.com/MystenLabs/sui/pull/19636)
* [indexer: backfill tx_affected_objects](https://github.com/MystenLabs/sui/pull/19675)
* [indexer: tx_affected_objects ingestion-based backfill](https://github.com/MystenLabs/sui/commit/b902e8213d5ff7f62d2ad6baf4091527f21d227d)
* [Indexer: Fix a bug in epochs system state json backfill](https://github.com/MystenLabs/sui/pull/19695)
* [fix(backfill): chunk up writes to DB](https://github.com/MystenLabs/sui/pull/19699)
* [chore(indexer): drop redundant curly braces](https://github.com/MystenLabs/sui/commit/a9b5784e6183d9f83ef91bc6c29d5506247c5e5a)
* [chore(indexer): reduce visibility of crates](https://github.com/MystenLabs/sui/commit/372f6b62e3746ac3445305b5211d7e39cb99bfa8)
* [fix(indexer): fix default db path in SQL ingestion scripts](https://github.com/MystenLabs/sui/commit/81fcff61ded72986fa4e8065736174c72494fdac)
* [indexer: backfill for tx_affected_addresses](https://github.com/MystenLabs/sui/commit/6a579d6be31245b360df5dd76afc7b8baafd1429)

samuel-rufi:
# Description of change

Introduces the backfill infrastructure for the indexer. Consolidates relevant upstream patches, filters out backfill workflows for unsupported database tables, and improves error handling, testing, API compatibility, progress tracking, refactoring, type & trait naming, and comments.

**Upstream patches:**

* [indexer: SQL Backfill command (#19359)](https://github.com/MystenLabs/sui/pull/19359) {"side-effects": "major", "type": "feature"}
* [Indexer: A few improvements to backfill tool (#19441)](https://github.com/MystenLabs/sui/pull/19441)
* [Indexer: Add a generic backfill template](https://github.com/MystenLabs/sui/pull/19510)
* [Indexer: Add ingestion based backfill](https://github.com/MystenLabs/sui/pull/19636)
* [indexer: backfill tx_affected_objects](https://github.com/MystenLabs/sui/pull/19675)
* [indexer: tx_affected_objects ingestion-based backfill](https://github.com/MystenLabs/sui/commit/b902e8213d5ff7f62d2ad6baf4091527f21d227d)
* [Indexer: Fix a bug in epochs system state json backfill](https://github.com/MystenLabs/sui/pull/19695)
* [fix(backfill): chunk up writes to DB](https://github.com/MystenLabs/sui/pull/19699)
* [chore(indexer): drop redundant curly braces](https://github.com/MystenLabs/sui/commit/a9b5784e6183d9f83ef91bc6c29d5506247c5e5a)
* [chore(indexer): reduce visibility of crates](https://github.com/MystenLabs/sui/commit/372f6b62e3746ac3445305b5211d7e39cb99bfa8)
* [fix(indexer): fix default db path in SQL ingestion scripts](https://github.com/MystenLabs/sui/commit/81fcff61ded72986fa4e8065736174c72494fdac)
* [indexer: backfill for tx_affected_addresses](https://github.com/MystenLabs/sui/commit/6a579d6be31245b360df5dd76afc7b8baafd1429)

Fixes https://github.com/iotaledger/iota/issues/7365

This PR doesn’t touch existing ingestion pipelines or synchronization. It doesn’t include any actual backfill implementation, instead, it delivers the infrastructure and tooling needed to create backfill jobs. However, relevant patch-specific tests have been added.

### Release Notes

- [ ] Protocol:
- [X] Indexer: Adds a new `RunBackfill` CLI command to backfill DB tables.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-indexer/src/backfill/sql/task.rs
```diff
@@ -0,0 +1,224 @@
...
+            self.key_column,
+            *range.start(),
+            *range.end()
+        );
+
+        let mut conn = get_pool_connection(&pool)?;
+
+        diesel::sql_query(&query)
+            .bind::<BigInt, _>(start as i64)
+            .bind::<BigInt, _>(end as i64)
```
tomxey: For my understanding, why do we need the bind, if we specify the range already in the query above?

crates/iota-indexer/src/backfill/ingestion/task.rs
```diff
@@ -0,0 +1,213 @@
...
+                } else {
+                    break;
+                }
+            }
+            if start <= end {
+                self.notify.notified().await;
+            } else {
+                break;
+            }
+        }
```
tomxey: Seems to me that this could be simplified into:

```
        while start <= end {
            while let Some((_, processed)) = self
                .ready_checkpoints
                .remove(&(start as CheckpointSequenceNumber))
            {
                processed_data.extend(processed);
                start += 1;
            }

            if start <= end {
                self.notify.notified().await;
            }
        }
```

Or am I missing something?

crates/iota-indexer/src/backfill/sql/task.rs
samuel-rufi: Good catch! The query should use placeholders instead of populating the params directly. The binds can kept as they are. Fixed: https://github.com/iotaledger/iota/pull/7603/commits/0ff0952adebb6c22102d47b374e6e3826f065294

crates/iota-indexer/src/backfill/ingestion/task.rs
samuel-rufi: You're absolutely right. This upstream code should be refactored as well, please see: https://github.com/iotaledger/iota/pull/7603/commits/0ff0952adebb6c22102d47b374e6e3826f065294

crates/iota-indexer/src/config.rs
```diff
@@ -183,6 +223,43 @@ pub enum Command {
     HelpDeprecated,
 }
 
+#[derive(Subcommand, Clone, Debug)]
+#[non_exhaustive]
+pub enum CommandV2 {
```
sergiupopescu199: From my understanding, a breaking change is something that causes existing code to fail, behave differently, or require modification in order to continue working as before. In our case, the `Command` enum was introduced to define CLI commands, and `CommandV2` appears to have been added to support new subcommands. However, since adding a new subcommand is a non-breaking, backward-compatible change, it would have been possible to add these directly to the original `Command` enum.

Maintaining both `Command` and `CommandV2` seems redundant, as `CommandV2` simply extends the original functionality without breaking compatibility. Therefore, I suggest removing the old `Command` enum and renaming `CommandV2` to `Command` for clarity and simplicity. This will help reduce confusion and keep the codebase clean, as there is no need to maintain two separate enums for CLI commands when a single, unified enum will suffice.

Let me know what you think 🙏

crates/iota-indexer/src/backfill/task.rs
```diff
@@ -0,0 +1,20 @@
...
+use std::ops::RangeInclusive;
+
+use async_trait::async_trait;
+
+use crate::{db::ConnectionPool, errors::IndexerError};
+
+/// Encapsulates the logic to process and persist data for a chunk of
+/// checkpoints.
+#[async_trait]
+pub(crate) trait BackfillTask: Send + Sync {
```
sergiupopescu199: Nit: In async Rust where "task" often implies a spawned asynchronous unit of work. The trait `BackfillTask` is more of a strategy or operation definition, not a running async task. what do you think if we rename it to `Backfill`?

I saw that on upstream they did the same but I suppose we could improve on that 🙏

we could remove the `task.rs` file and place the trait definition in the `mod.rs`

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+pub(crate) mod ingestion;
+pub mod runner;
+pub(crate) mod sql;
+pub(crate) mod task;
+
+// Subcommands for selecting a backfill task to run.
+// Each variant corresponds to a different backfill implementation.
+#[derive(Subcommand, Clone, Debug)]
+#[non_exhaustive]
+pub enum BackfillTaskKind {
```
sergiupopescu199: Nit: we could rename it to `BackfillKind`.

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/// Selects the concrete ingestion backfill task to run.
+/// Each variant of `IngestionBackfillKind` must correspond to a type that
+/// implements the `IngestionBackfill` trait.
+#[derive(ValueEnum, Clone, Debug)]
+#[non_exhaustive]
+pub enum IngestionBackfillKind {}
+
+pub(crate) async fn get_backfill_task(
```
sergiupopescu199: Nit: this could be renamed `get_backfill`. It should return the renamed Trait, `Arc<Backfill>`

crates/iota-indexer/src/backfill/sql/task.rs
```diff
@@ -0,0 +1,221 @@
...
+
+use crate::{
+    backfill::task::BackfillTask,
+    db::{ConnectionPool, get_pool_connection},
+    errors::IndexerError,
+};
+
+/// A backfiller that runs SQL queries in parallel to update a range of rows in
+/// a database table.
+pub(crate) struct SqlBackfillTask {
```
sergiupopescu199: Nit: `SqlBackfill`, we also could rename the file to `sql_backfill.rs` since task is a bit misleading.

crates/iota-indexer/src/backfill/runner.rs
```diff
@@ -0,0 +1,150 @@
...
+            rate = final_rate,
+            "Completed backfill"
+        );
+
+        Ok(())
+    }
+}
+
+/// Creates chunks based on the total range and chunk size.
+fn generate_chunks(
```
sergiupopescu199: Nit: we could rename it as ` chunk_range_stream`

really good approach you took compared to upstream!!!! 🌸

crates/iota-indexer/src/backfill/runner.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    async fn run_impl(
+        pool: ConnectionPool,
+        config: BackfillConfig,
+        total_range: RangeInclusive<usize>,
+        task: Arc<dyn BackfillTask>,
+    ) -> Result<(), IndexerError> {
+        let timer = Instant::now();
+        let processed_counter = Arc::new(AtomicUsize::new(0));
+        // Keeps track of the checkpoint ranges (using starting checkpoint number)
```
sergiupopescu199: Here we don't do only checkpoint ranges but others aswell

```suggestion
        // Keeps track of the ranges (using starting range number)
```

crates/iota-indexer/src/backfill/ingestion/adapter.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2024 IOTA Stiftung
```
sergiupopescu199: Nit: should include 2025, please update the other files aswell 🙏

crates/iota-indexer/src/backfill/ingestion/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use std::sync::Arc;
+
+use iota_types::full_checkpoint_content::CheckpointData;
+
+use crate::{db::ConnectionPool, errors::IndexerError};
+
+/// Processes checkpoints and commits processed data to the database.
+#[async_trait::async_trait]
+pub(crate) trait IngestionBackfillHandler: Send + Sync {
```
sergiupopescu199: Nit: we could rename it as `IngestionBackfill`.

crates/iota-indexer/src/backfill/runner.rs
```diff
@@ -0,0 +1,150 @@
...
+            secs = elapsed,
+            rate = final_rate,
+            "Completed backfill"
+        );
+
+        Ok(())
+    }
+}
+
+/// Creates chunks based on the total range and chunk size.
```
sergiupopescu199:
```suggestion
/// Returns an asynchronous stream that yields consecutive, non-overlapping subranges ("chunks")
/// from the given inclusive range, each with a maximum length of `chunk_size`.
///
/// This is useful for processing a large range in smaller, manageable pieces, such as
/// batching database queries or parallelizing work.
///
/// # Example
///
/// ```rust
/// use futures::StreamExt;
///
/// let range = 0..=10;
/// let chunk_size = 3;
/// let mut stream = chunk_range_stream(range, chunk_size);
///
/// // This will yield: 0..=2, 3..=5, 6..=8, 9..=10
/// while let Some(chunk) = stream.next().await {
///     println!("{:?}", chunk);
/// }
/// ```
```

sergiupopescu199 (commented):
Amazing work Samuel! ✨

- Left some comments and nits. Additionally, I think it would be very helpful to include some usage examples. For instance, [the upstream project provides scripts](https://github.com/MystenLabs/sui/tree/main/crates/sui-indexer/src/backfill/backfill_instances/sql_backfills) that demonstrate common workflows—we could adapt those and add them as examples in our README. This would make it easier for users to get started and understand typical use cases.

- License modifications should include 2025 🙏

crates/iota-indexer/src/config.rs
samuel-rufi: Hey @sergiupopescu199 thanks for your comments! 🙏 

Regarding the `Command` type, I was considering what you mentioned. Clap could handle the additional command in a non breaking way. From a Rust lib API point of view it's however a breaking change because of the `pub` visibility and it doesn't implement `#[non_exhaustive]` / allow for new variants. Therefore to avoid any possible breaking change here, I opt in to introduce `CommandV2`.

sergiupopescu199: @samuel-rufi You’re right that for a public Rust library, adding new variants to a non-`#[non_exhaustive]` enum is a breaking change. However, as far as I understand, the indexer is primarily an application, not a library intended for external consumption. In this context, the `Command` enum could be kept private (or at least crate-private), which would allow us to evolve it without worrying about breaking downstream code.

For our use case, the main compatibility concerns are on the JSON-RPC and CLI interfaces. Adding new subcommands to the CLI is generally considered a backward-compatible, non-breaking change, since it doesn’t affect existing functionality or users.

If we don’t intend for external crates to match on `Command` directly, I think we can safely unify the enums and simplify the codebase. But if you see a specific scenario where this would break existing consumers, please let me know 🙏

crates/iota-indexer/src/backfill/ingestion/adapter.rs
samuel-rufi: Good catch, updated! https://github.com/iotaledger/iota/pull/7603/commits/6791fba7af3b0cb8d6b77a25524ba15bf4d63724

crates/iota-indexer/src/backfill/task.rs
samuel-rufi: Renamed it!

crates/iota-indexer/src/config.rs
samuel-rufi: You have a good point here @sergiupopescu199. We defined the introduction of this new command as `{"side-effects": "major", "type": "feature"}`. If we want to avoid any breaking changes and therefore don't break the existing definition of this type, then we need to stick to it as proposed in the PR for now. Also what I have in mind is the introduction of the _new_ `pub struct IndexerConfig` (which contains `Command`) which further hardened the config type definition. Would it be okay for you if we eventually adjust this in a follow up PR once we revisit this internally?

sergiupopescu199: @samuel-rufi Thank you very much for the detailed explanation, I suppose it can be always revisited in subsequent PRs 🙏.

sergiupopescu199 (approved):

crates/iota-indexer/src/config.rs
kodemartin: I have to agree with @sergiupopescu199 here. The `config` module is logically bound to the application and is thus highly unlikely that some other application or library code uses it. It is for structural convenience that it is included in the `lib`, so as not to add a different module structure for the application.

If you still think it is worthwhile preserving the Rust API for `config` instead of adding a new `V2` type, you could nest the extended version of `IndexerConfig` and `Command` in a `config` module of the application itself, so as to better distinguish what is part of the lib and what part of the application.

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+pub(crate) trait Backfill: Send + Sync {
+    async fn backfill_range(
+        &self,
+        pool: ConnectionPool,
+        range: &RangeInclusive<usize>,
+    ) -> Result<(), IndexerError>;
+}
+
+// Subcommands for selecting a backfill task to run.
+// Each variant corresponds to a different backfill implementation.
```
kodemartin: Use `///` instead of `//`.

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Subcommand, Clone, Debug)]
+#[non_exhaustive]
+pub enum BackfillKind {
```
kodemartin, line 36: Following the discussion on distinguishing scopes between application and lib: Should this part of the application or the lib?

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        format!("postgres://postgres:postgrespw@localhost:5432/{db_name}")
+    }
+
+    /// Counts how many rows exist in the target table.
+    #[derive(QueryableByName, Debug)]
+    pub(crate) struct RowCount {
+        #[diesel(sql_type = BigInt)]
+        pub(crate) cnt: i64,
+    }
+}
```
kodemartin: Please consider merging this into `test_utils`, as we have overlapping logic.

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use crate::{backfill::sql::sql_backfill::SqlBackfill, db::ConnectionPool, errors::IndexerError};
+
+pub(crate) mod ingestion;
+pub mod runner;
+pub(crate) mod sql;
+
+/// Encapsulates the logic to process and persist data for a chunk of
+/// checkpoints.
+#[async_trait]
```
kodemartin, line 23: Do we need to use the `async_trait` now that `async fn` is stable in traits?

crates/iota-indexer/src/backfill/runner.rs
kodemartin, line 1: Could you outline the benefits of the version with respect to [upstream](https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer/src/backfill/backfill_runner.rs). Because the latter is simpler, and seems similar in performance?

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+use clap::{Subcommand, ValueEnum};
+
+use crate::{backfill::sql::sql_backfill::SqlBackfill, db::ConnectionPool, errors::IndexerError};
+
+pub(crate) mod ingestion;
+pub mod runner;
+pub(crate) mod sql;
+
+/// Encapsulates the logic to process and persist data for a chunk of
+/// checkpoints.
```
kodemartin: The range is a range of rows for `SqlBackfill`. Could you generalize the description accordingly?

kodemartin (commented):
Looks good @samuel-rufi, with a few comments to further discuss.

crates/iota-indexer/src/config.rs
samuel-rufi: I understand and agree with your and @sergiupopescu199's reasoning here. I wanted to make sure we are aligned on how to handle such changes. To clarify my intent behind introducing the V2 type: I primarily wanted to ensure to avoid any breaking changes, potential side effects and preserve the existing public, exhaustive type. I also favor integrating the new variant directly into the existing type, as it reduces complexity, improves maintainability, and is unlikely to cause any issues. I reverted the commit: [1e68a03](https://github.com/iotaledger/iota/pull/7603/commits/1e68a034a128a7dbb7e418f696dcbf1f5134d33a).

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: I think having this part of the library should be fine. I mean it's a parameter for `BackfillRunner` so having this also part of the library seems reasonable.

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Fixed in https://github.com/iotaledger/iota/pull/7603

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Fixed in https://github.com/iotaledger/iota/pull/7603/commits/00007a36df715f051d898b9417649e725372dbf0

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Fixed in https://github.com/iotaledger/iota/pull/7603

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: Our version mainly refactors the upstream code, improves on error handling, providing clearer logs, and explicitly indicating restart points in case something goes wrong, which should reduce debugging time. The upstream implementation instead relies on tasks panicking.

tomxey (approved):

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Moved as part of https://github.com/iotaledger/iota/pull/7603/commits/43170b9e6df1f70748b442d05520537848875364

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Only if dynamic dispatch (`dyn Trait`) is required. Native async trait methods, despite being stable, are not yet directly compatible with `dyn Trait` without explicit future boxing. We could remove `async_trait` by manually adding minimal boilerplate. However, since `async_trait` is still used in this crate and explicitly solves this issue ([docs.rs](https://docs.rs/async-trait/latest/async_trait/)), I suggest we can reconsider this dependency optimization at a later stage.

crates/iota-indexer/src/test_utils.rs
```diff
@@ -370,3 +370,17 @@ impl<'a> IotaTransactionBlockResponseBuilder<'a> {
         }
     }
 }
+
+/// Returns a database URL for testing purposes.
+/// It uses a default user and password, and connects to a local PostgreSQL
+/// instance.
+pub fn test_database_url(db_name: &str) -> String {
+    format!("postgres://postgres:postgrespw@localhost:5432/{db_name}")
+}
```
kodemartin: This still overlaps semantically with `TestDatabase`.  

And there is a private function that replicates the functionality in `db::tests`.

crates/iota-indexer/src/backfill/runner.rs
kodemartin: These are valid points.

I still don't see why you had to refactor `chunk_range_stream` and revise the concurrent processing of ranges. This implementation is more verbose.

crates/iota-indexer/src/backfill/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+use async_trait::async_trait;
+use clap::{Subcommand, ValueEnum};
+
+use crate::{backfill::sql::sql_backfill::SqlBackfill, db::ConnectionPool, errors::IndexerError};
+
+pub(crate) mod ingestion;
+pub mod runner;
+pub(crate) mod sql;
+
+/// Encapsulates the logic to process and persist data for a range.
```
kodemartin: Now the description does not specify at all what `range` is supposed to be. As a user of the trait I would like to know what are the semantics of this `range` to properly design the implementors. 

For example, it is fair to ask after reading this: For a `range` of what?

Could you provide a general description for the semantics of `range` in the context of the logic introduced with this patch?

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: You're right, the revision of `chunk_range_stream` wasn't really necessary; it only slightly helps with memory usage by lazily generating ranges. If you prefer, I'm completely fine with reverting this part.
The adjustments in concurrent processing (`map + buffer_unordered + try_for_each`) were necessary to support proper error propagation, as `for_each_concurrent` doesn't allow returning errors directly.

crates/iota-indexer/src/backfill/mod.rs
samuel-rufi: Good point, I have clarified the doc. Please see: https://github.com/iotaledger/iota/pull/7603/commits/00007a36df715f051d898b9417649e725372dbf0

crates/iota-indexer/src/test_utils.rs
samuel-rufi: Adjusted `test_database_url` to `db_url`. The idea of the "test" prefix it should be really just a tool for testing. As we are part of the `test_utils` context, `db_url` should be sufficient.
Please see: https://github.com/iotaledger/iota/pull/7603/commits/6a939c2a58a46011699eb7f8a23ed6d2f0cbb76c

crates/iota-indexer/src/backfill/runner.rs
kodemartin: `create_chunk_iter` was lazy as well, returning an `Iterator ` of ranges. 

`chunk_range_stream`  helps avoiding the `mpsc` channel and producing the stream on top of the iterator, but one could argue that generating the stream is counter-intuitive, as there is no underlying async operation to produce the ranges

And there are alternative (e.g. `try_for_each_concurrently`) that might have helped with handling errors.

No need to revert now.

crates/iota-indexer/src/backfill/runner.rs
```diff
@@ -0,0 +1,177 @@
...
+                    in_progress.lock().await.insert(start_cp);
+
+                    // Execute backfill for the range
+                    if let Err(e) = backfill.backfill_range(pool, &range).await {
+                        let min_range_restart_num = {
+                            let mut guard = in_progress.lock().await;
+                            guard.remove(&start_cp);
+                            guard.iter().next().cloned().unwrap_or(start_cp)
+                        };
+                        error!("Chunk {start_cp}-{end_cp} failed. Minimum range restart number: {min_range_restart_num}. Error: {e}", );
```
kodemartin: These futures are run concurrently, so `min_range_restart_num` might very well be the start of the last range of the concurrent batch.

What if no other ranges have been processed until this fails? Wouldn't the user be hinter toward restarting from a range that does not include the unprocessed ranges?

crates/iota-indexer/src/backfill/runner.rs
```diff
@@ -0,0 +1,177 @@
...
+                    let total = counter.fetch_add(count, Ordering::Relaxed) + count;
+                    let elapsed = timer.elapsed().as_secs_f64();
+                    let avg_rate = total as f64 / elapsed;
+                    info!(
+                        processed = total,
+                        secs = elapsed,
+                        rate = avg_rate,
+                        min_range_start,
+                        "Avg backfill speed"
+                    );
```
kodemartin: What is the `min_range_start` supposed to inform about here?

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: Good catch! It indeed might be not the accurate restart number as it will very probably have gaps on failure.
Returning the exact restart number would require at least the book keeping of all currently running chunks.
This might be an optimization for the future.
In the worst case the restart point would be at `failing_start - max_concurrency * chunk_size`. An option would be to provide that entry point as part of the error message (this can be however also calculated by the user manually).

At this point I would suggest we remove the `Minimum range restart number: {min_range_restart_num}.` log as it's not accurate. And add a restart point (accurate or "by at most") in a follow up PR.

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: It can be considered as an additional progress log, similarly as done in upstream. But after having a second look at this I feel like this doesn't really add value as we have "processed" which seems sufficient. Will remove

samuel-rufi: Fixed as part of https://github.com/iotaledger/iota/pull/7603/commits/bb29d2e432fc4e702abc308df2552b6d135b6fd4

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: Fixed as part of https://github.com/iotaledger/iota/pull/7603/commits/bb29d2e432fc4e702abc308df2552b6d135b6fd4

crates/iota-indexer/src/backfill/runner.rs
samuel-rufi: Thanks for clarifying @kodemartin, you are right in upstream they also generate the chunks lazily. My main intent was streamlining the code a bit, and I had also thought numeric range slicing would be cheap enough to avoid the extra buffering, in order to get rid of the mpsc. But as you pointed out, the actual benefit is marginal. Totally see your point and will keep it in mind for future changes.

kodemartin (approved):

PR #7604, opened 2025-07-01
feat(consensus): Introduce validator scoring metrics

oliviasaa:
# Description of change

This PR introduces a set of misbehavior-related metrics to be stored by all validators. These same metrics are exposed via Prometheus. 

Each validator will track, for each of their peers:
- Equivocations
- Missing block proposals
- Faulty blocks sent to them

Bases on these values, validators also calculate their peers' score. 

Notes:
- Faulty blocks are divided into provable and unprovable, since those misbehaviors could be treated differently.
- Equivocations and missing block counts are divided into cached and uncached; uncached metrics will never decrease, and rounds that are not cached anymore can be considered "more final" than cached rounds. The uncached metrics also persist in storage.

### Release Notes

- [ ] Protocol: Introduces new Prometheus metrics for validators, which are also locally stored in the node.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


consensus/core/src/dag_state.rs
```diff
@@ -246,6 +246,25 @@ impl DagState {
             );
         }
 
+        // Initialize the scoring metrics according to the stored blocks
+        let threshold_clock_round = state.threshold_clock_round();
+        for index in 0..num_authorities {
+            let authority_index = state.context.committee.to_authority_index(index).unwrap();
+            let eviction_round = state.evicted_rounds[index];
```
cyberphysic4l: Why do we use `index` here rather than `authority_index`? What's the difference?

consensus/core/src/metrics.rs
```diff
@@ -80,21 +87,161 @@ const SIZE_BUCKETS: &[f64] = &[
...
+        validator: AuthorityIndex,
+        hostname: &str,
+        blocks_in_cache: &BTreeSet<BlockRef>,
+        mut evicted_blocks_per_round: Vec<u32>,
+        threshold_clock_round: u32,
+        eviction_round: u32,
+    ) {
+        // Update metrics according to the blocks from evicted rounds.
+        let missing_blocks_from_evicted_rounds =
+            evicted_blocks_per_round.extract_if(.., |x| x == &0).count() as u64;
```
cyberphysic4l: I think you should use `filter` here is you want to count the number of rounds with missing blocks. Here, you are counting the number of rounds that have at least one block.

consensus/core/Cargo.toml
```diff
@@ -32,6 +32,7 @@ quinn-proto.workspace = true
 rand.workspace = true
 rustls.workspace = true
 serde.workspace = true
+strum.workspace = true
```
cyberphysic4l: What is this?

cyberphysic4l (changes requested):

consensus/core/Cargo.toml
oliviasaa: I was using this dependency at some point, but now it's not used anymore. Let me delete it. 
(I was using this to iterate over enums)

consensus/core/src/dag_state.rs
oliviasaa: I don't understand the question. Do you mean why I'm calling this `index` or why are we iterating over integers instead of over the authorities vector?
Iterating over the authorities vector indeed sounds cleaner...

consensus/core/src/metrics.rs
oliviasaa: I want to count the number of rounds with missing blocks and remove those entries from `evicted_blocks_per_round`, so later I can use 

`let equivocations_from_evicted_rounds = evicted_blocks_per_round.iter().sum::<u32>() as u64
            - evicted_blocks_per_round.len() as u64;` 
 
 to count the equivocations

oliviasaa: `extract_if` returns the extracted part, so when I call `count()` this gives me the size of the returned iterator (which is the part of the vector with 0s). This line is actually counting the zeroes, not the other part.

`evicted_blocks_per_round` is mutated by this line, and after this, it will only have the rounds with at least 1 block, that's why we can use it to count the equivocations

cyberphysic4l: Right, I understand now, thanks for explaining.

consensus/core/src/dag_state.rs
```diff
@@ -246,6 +246,25 @@ impl DagState {
...
+                .scan_block_rounds_by_author(authority_index)
+                .expect("Database error");
+            state.context.metrics.load_scoring_metrics_from_storage(
+                authority_index,
+                hostname,
+                stored_block_rounds,
+                threshold_clock_round,
+                eviction_round,
+            );
+        }
```
cyberphysic4l:
```suggestion
        for (authority_index, _) in state.context.committee.authorities() {
            let eviction_round = state.evicted_rounds[authority_index];
            let hostname = &state.context.committee.authority(authority_index).hostname;
            let stored_block_rounds = state
                .store
                .scan_block_rounds_by_author(authority_index)
                .expect("Database error");
            state.context.metrics.load_scoring_metrics_from_storage(
                authority_index,
                hostname,
                stored_block_rounds,
                threshold_clock_round,
                eviction_round,
            );
        }
```

consensus/core/src/dag_state.rs
cyberphysic4l: Well I was confused why you bother to get the `authority_index` at all when you already have `index` which seems to be the same value, and then you use them interchangeably. I made a suggestion to just iterate over `authority_index`, see if that makes sense or I have missed something.

consensus/core/src/dag_state.rs
cyberphysic4l: Actually, you can also use the authority itself in the iterator to get the hostname too.
```suggestion
        for (authority_index, authority) in state.context.committee.authorities() {
            let eviction_round = state.evicted_rounds[authority_index];
            let hostname = &authority.hostname;
            let stored_block_rounds = state
                .store
                .scan_block_rounds_by_author(authority_index)
                .expect("Database error");
            state.context.metrics.load_scoring_metrics_from_storage(
                authority_index,
                hostname,
                stored_block_rounds,
                threshold_clock_round,
                eviction_round,
            );
        }
```

consensus/core/src/dag_state.rs
oliviasaa: it's just that `authority_index` has a different type than `index` (`AuthorityIndex` vs an integer). But maybe I should iterate directly over the authority vector itself

oliviasaa: ah, I see that you suggested this already below 😅 
Thanks! 🙏

consensus/core/src/dag_state.rs
oliviasaa: done!

consensus/core/src/dag_state.rs
```diff
@@ -246,6 +246,40 @@ impl DagState {
             );
         }
 
+        // Load the stored metrics relative to rounds that were not loaded to cache.
+        let metrics = state.store.scan_metrics().expect("Database error");
```
piotrm50: `metrics` variable name could be more specific imo

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+                .faulty_blocks_unprovable_by_authority
+                .store(faulty_blocks_unprovable_by_authority, Ordering::Relaxed);
+            self.scoring_metrics.stored[authority.value()]
+                .equivocations_by_authority
+                .store(equivocations_by_authority, Ordering::Relaxed);
+            self.scoring_metrics.stored[authority.value()]
+                .missing_proposals_by_authority
+                .store(missing_proposals_by_authority, Ordering::Relaxed);
+
+            self.node_metrics
```
piotrm50, line 197: I don't think we need to reset the metrics in the beginning as the values will be set to zero anyway.

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+            equivocations_by_authority: self.scoring_metrics.stored[validator.value()]
+                .equivocations_by_authority
+                .load(Ordering::Relaxed),
+            missing_proposals_by_authority: self.scoring_metrics.stored[validator.value()]
+                .missing_proposals_by_authority
+                .load(Ordering::Relaxed),
+        })
+    }
+
+    pub(crate) fn load_scoring_metrics_from_storage(
```
piotrm50: Rename to something like `initialize_scoring_metrics`. We're not loading metrics from storage in this function.

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+                .uncached_missing_proposals_by_authority
+                .with_label_values(&[hostname])
+                .inc_by(missing_proposals_by_authority);
+
+            // Update score
+            self.update_score(None);
+        }
+    }
+
+    pub(crate) fn calculate_scoring_metrics_from_storage(
```
piotrm50: how is storage related to this logic?

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
     10_000_000.0,
 ]; // size in bytes
 
+const ERRORS_FROM_UNSIGNED_BLOCK_VERIFICATION: [&str; 6] = [
```
piotrm50: I'd rather avoid using hardcoded strings to later compare them against error names. Maybe you can replace those with helper functions like:
```
fn is_from_unsigned_block_verification(err: &BlockError) -> bool {
    matches!(
        err,
        BlockError::WrongEpoch
            | BlockError::UnexpectedGenesisBlock
            | BlockError::InvalidAuthorityIndex
            | BlockError::SerializationFailure
            | BlockError::MalformedSignature
            | BlockError::SignatureVerificationFailure
    )
}
```

consensus/core/src/metrics.rs
```diff
@@ -800,3 +1179,813 @@ impl NodeMetrics {
...
+            core_dispatcher.clone(),
+            rx_block_broadcast,
+            dag_state,
+            store,
+        ));
+        (keys, context, core_dispatcher, authority_service)
+    }
+
+    impl ValidatorScoreMetrics {
+        pub(crate) fn uncached_missing_proposals_by_authority(&self) -> Vec<u64> {
```
piotrm50, line 1451: Maybe rename to `stored_` to be in line with `self.stored`

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+        let cached_equivocations =
+            number_of_cached_blocks.saturating_sub(unique_cached_block_rounds) as u64;
+        let missing_blocks_in_cached_rounds = threshold_clock_round
+            .saturating_sub(eviction_round + unique_cached_block_rounds as u32 + 1)
+            as u64;
 
-#[cfg(test)]
-pub(crate) fn test_metrics() -> Arc<Metrics> {
-    initialise_metrics(Registry::new())
+        self.scoring_metrics.cached[validator.value()]
```
piotrm50: We have `Index<AuthorityIndex> for Vec<T>` so there is no need to call `.value()`.

piotrm50 (changes requested):
Looks good overall, only added some small comments. Also, please add PR description as well as a bit more context in the comments describing the overall flow as it wasn't initially clear why we separate cached vs stored metrics. Request another review when you respond to the comments.

consensus/core/src/metrics.rs
```diff
@@ -800,3 +1179,813 @@ impl NodeMetrics {
...
+pub(crate) struct Score(AtomicU64);
+impl Score {
+    pub(crate) fn new() -> Self {
+        Self(AtomicU64::new(0))
+    }
+
+    pub(crate) fn store(&self, value: u64) {
+        self.0.store(value, Ordering::Relaxed);
+    }
+}
```
cyberphysic4l: Move this above the ValidatorScoreMetrics where it is used.

consensus/core/src/metrics.rs
```diff
@@ -800,3 +1179,813 @@ impl NodeMetrics {
...
+            cached,
+            score,
+        }
+    }
+}
+
+pub(crate) struct Score(AtomicU64);
+impl Score {
+    pub(crate) fn new() -> Self {
+        Self(AtomicU64::new(0))
```
cyberphysic4l: I think this should be initialised with a perfect score assuming no prior misbehaviours
```suggestion
        Self(AtomicU64::new(u32::MAX))
```

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+                .uncached_missing_proposals_by_authority
+                .with_label_values(&[hostname])
+                .reset();
+            self.node_metrics
+                .uncached_missing_proposals_by_authority
+                .with_label_values(&[hostname])
+                .inc_by(missing_proposals_by_authority);
+
+            // Update score
+            self.update_score(None);
```
cyberphysic4l: Shouldn't this be Some(authority)? Otherwise you are updating all authorities score every time in this loop. I think you should argument non optional anyway, it's weird to read with the option.

consensus/core/src/metrics.rs
```diff
@@ -77,24 +84,354 @@ const SIZE_BUCKETS: &[f64] = &[
...
+            .uncached_equivocations_by_authority
+            .with_label_values(&[hostname])
+            .inc_by(evicted_equivocations);
+        self.node_metrics
+            .uncached_missing_proposals_by_authority
+            .with_label_values(&[hostname])
+            .inc_by(missing_blocks_in_evicted_rounds);
+
+        // Update score
+        self.update_score(Some(validator));
```
cyberphysic4l: Why not just call `update_authority_score` here?

consensus/core/src/metrics.rs
```diff
@@ -800,3 +1179,813 @@ impl NodeMetrics {
         }
     }
 }
+
+pub(crate) struct ValidatorScoreMetrics {
```
cyberphysic4l: I would rename this to `ScoringMetrics` to match the variable name, and everywhere we have a name that ends in `ScoreMetrics` I would rename to `ScoringMetrics`. `Score` will then only be used in names that relate to the actual validator score which combines the scoring metrics.

consensus/core/src/authority_service.rs
```diff
@@ -29,7 +29,6 @@ use crate::{
     storage::Store,
     synchronizer::SynchronizerHandle,
 };
-
```
cyberphysic4l: place this line back

consensus/core/src/storage/mod.rs
```diff
@@ -16,8 +16,8 @@ use crate::{
     block::{BlockRef, Round, VerifiedBlock},
     commit::{CommitInfo, CommitRange, CommitRef, TrustedCommit},
     error::ConsensusResult,
+    metrics::StoredScoreMetricsU64,
 };
-
```
cyberphysic4l, line 20: place this line back in

cyberphysic4l (changes requested):

consensus/core/src/metrics.rs
oliviasaa: ah, of course!

consensus/core/src/metrics.rs
oliviasaa: sorry, this should be outside of the loop.
But I removed the option anyway, it seems that I pushed partial changes.

consensus/core/src/metrics.rs
oliviasaa: yeah, you are right, let me change that.

consensus/core/src/metrics.rs
oliviasaa: I just pushed changes that address all the comments regarding functions names. I changed a little bit their structure as well

piotrm50 (changes requested):
Left a comment about one last detail that I would change.

consensus/core/src/metrics.rs
oliviasaa: done

consensus/core/src/metrics.rs
oliviasaa: done

consensus/core/src/metrics.rs
oliviasaa: done

consensus/core/src/metrics.rs
```diff
@@ -80,21 +87,271 @@ const SIZE_BUCKETS: &[f64] = &[
...
+
+    fn update_missing_blocks_and_equivocations(
+        &self,
+        missing_blocks: u64,
+        equivocations: u64,
+        hostname: &str,
+        authority: AuthorityIndex,
+        metric_type: &str,
+    ) {
+        assert!(
```
piotrm50: Using an Enum would be safer than a string, because if there is a typo somewhere then the code won't compile in the first place rather than panic during runtime.

oliviasaa: done

consensus/core/src/dag_state.rs
oliviasaa: done

consensus/core/src/metrics.rs
```diff
@@ -80,21 +87,274 @@ const SIZE_BUCKETS: &[f64] = &[
...
+                .inc_by(faulty_blocks_unprovable_by_authority);
+            self.scoring_metrics.uncached[authority]
+                .faulty_blocks_provable_by_authority
+                .store(faulty_blocks_provable_by_authority, Ordering::Relaxed);
+            self.scoring_metrics.uncached[authority]
+                .faulty_blocks_unprovable_by_authority
+                .store(faulty_blocks_unprovable_by_authority, Ordering::Relaxed);
+
+            // Update all scores
+            self.update_scores();
```
cyberphysic4l: Why do you need to loop over all authorities here inside the loop? Can't you just use `update_authority_score` here and delete this function entirely?

consensus/core/src/metrics.rs
```diff
@@ -814,3 +1116,907 @@ impl NodeMetrics {
...
+            equivocations_by_authority: AtomicU64::new(0),
+            missing_proposals_by_authority: AtomicU64::new(0),
+        }
+    }
+}
+
+// This struct is used in storage. It holds the same data as
+// `UncachedScoreMetrics`, but uses `u64` instead of `AtomicU64`.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub(crate) struct StoredScoreMetricsU64 {
```
cyberphysic4l:
```suggestion
pub(crate) struct StoredScoringMetricsU64 {
```

consensus/core/src/metrics.rs
```diff
@@ -814,3 +1116,907 @@ impl NodeMetrics {
...
+        Self {
+            uncached,
+            cached,
+            score,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct UncachedScoreMetrics {
```
cyberphysic4l:
```suggestion
pub(crate) struct UncachedScoringMetrics {
```

consensus/core/src/metrics.rs
```diff
@@ -814,3 +1116,907 @@ impl NodeMetrics {
...
+        Self {
+            faulty_blocks_provable_by_authority: AtomicU64::new(0),
+            faulty_blocks_unprovable_by_authority: AtomicU64::new(0),
+            equivocations_by_authority: AtomicU64::new(0),
+            missing_proposals_by_authority: AtomicU64::new(0),
+        }
+    }
+}
+
+pub(crate) struct CachedScoreMetrics {
```
cyberphysic4l:
```suggestion
pub(crate) struct CachedScoringMetrics {
```

consensus/core/src/metrics.rs
```diff
@@ -814,3 +1116,907 @@ impl NodeMetrics {
...
+            ],
+            [
+                vec![0; committee_size],
+                vec![0; committee_size],
+                vec![0; committee_size],
+                vec![0; committee_size]
+            ]
+        );
+
+        // TODO: documentation
```
cyberphysic4l: This TODO can probably be addressed now, right?

cyberphysic4l (approved):
Few small requests for final changes, otherwise lgtm

consensus/core/src/metrics.rs
oliviasaa: sorry, I did not push this change again, apparently

consensus/core/src/metrics.rs
oliviasaa: I also thought I had changed that already....

consensus/core/src/metrics.rs
oliviasaa: addressed in https://github.com/iotaledger/iota/commit/442ce0bbac4418002d7f1295a76229a3a16afe22

consensus/core/src/metrics.rs
oliviasaa: addressed in https://github.com/iotaledger/iota/commit/442ce0bbac4418002d7f1295a76229a3a16afe22

consensus/core/src/metrics.rs
oliviasaa: addressed in https://github.com/iotaledger/iota/commit/442ce0bbac4418002d7f1295a76229a3a16afe22

consensus/core/src/metrics.rs
oliviasaa: addressed in https://github.com/iotaledger/iota/commit/442ce0bbac4418002d7f1295a76229a3a16afe22

consensus/core/src/metrics.rs
oliviasaa: deleted the check in https://github.com/iotaledger/iota/commit/f924403bfd3ceb4ba1b972971bcc0442b6f70ae7, since it's actually unrelated to metrics

piotrm50 (approved):

PR #7605, opened 2025-07-01
feat(aa-example): Hardcode a keystore for the example

miker83z:
# Description of change

This makes the example work without manually creating addresses.

### DIFF
--- a/account_abstraction/aa_rust_sdk_flow/src/main.rs
+++ b/account_abstraction/aa_rust_sdk_flow/src/main.rs
@@ -1,16 +1,17 @@
 // Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{path::PathBuf, str::FromStr};
+use std::{fs, path::PathBuf, str::FromStr};
 
 use anyhow::{Ok, Result};
+use bip32::DerivationPath;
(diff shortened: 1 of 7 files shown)


Dkwcs (approved):
LGTM

account_abstraction/aa_rust_sdk_flow/src/main.rs
```diff
@@ -35,20 +36,61 @@ mod smart_account;
 mod tx_flow;
 mod utils;
 
+/// Got from iota-genesis-builder/src/stardust/test_outputs/alias_ownership.rs
+const MAIN_ADDRESS_MNEMONIC: &str = "few hood high omit camp keep burger give happy iron evolve draft few dawn pulp jazz box dash load snake gown bag draft car";
+
+/// Creates a temporary keystore.
+fn setup_keystore() -> Result<FileBasedKeystore, anyhow::Error> {
```
lzpap: Why not use InMemoryKeystore btw? Why do we need to set up a file for this?

miker83z: I copied it from the stardust examples, I guess it makes sense there and not here -> [12e7de0](https://github.com/iotaledger/iota/pull/7605/commits/12e7de0e00a3eb575fb11a91bfc43c72a99a5611)

Issue #7499, opened 2025-06-23
Investigate why `known_peers.shuffle(&mut rng);` causes simtests to fail


muXxer:
See #7484 for more info. We should investigate why this causes the simtest `test_simulated_load_restarts` to fail.

muXxer:
This is not the cause for the current nightlys to fail. Unfortunately we can't reproduce the current bug locally.
Maybe we can use "git bisect" to find the cause, by running it on CI?

muXxer:
It was introduced in the range of d149ab570dde65cf6fc1605dcad8db8b14573bfd...a2693041cf4ae634bd2996a346b1f5a9561995ca

muXxer:
We should add `MSIM_TEST_SEED` and a filter to filter the tets that should be run on the nightly workflow, and then do the binary search for the causing commit in the range ourselves.

muXxer:
The issue seems to be caused by the changes introduced in #7164

muXxer:
After #7602 gets merged, we should introduce the shuffling again.

muXxer:
We re-added shuffling in #7706 and #7711

PR #7609, opened 2025-07-01
fix: fix for crash recovery

VorobyevIlya:
# Description of change

This PR should fix the flaky/failing nightly simtest test_simulated_load_restarts.
It is based on upstream https://github.com/MystenLabs/sui/pull/21929
Also, the function pub fn enqueue_transactions_for_execution is added in authority.rs

#7499

### DIFF
--- a/crates/iota-node/src/lib.rs
+++ b/crates/iota-node/src/lib.rs
@@ -1530,47 +1530,68 @@ impl IotaNode {
         ))
     }
 
+    /// Re-executes pending consensus certificates, which may not have been
+    /// committed to disk before the node restarted. This is necessary for
+    /// the following reasons:
+    ///
+    /// 1. For any transaction for which we returned signed effects to a client,
+    ///    we must ensure that we have re-executed the transaction before we
+    ///    begin accepting grpc requests. Otherwise we would appear to have
+    ///    forgotten about the transaction.
+    /// 2. While this is running, we are concurrently waiting for all previously
+    ///    built checkpoints to be rebuilt. Since there may be dependencies in
+    ///    either direction (from checkpointed consensus transactions to pending
+    ///    consensus transactions, or vice versa), we must re-execute pending
+    ///    consensus transactions to ensure that both processes can complete.
+    /// 3. Also note that for any pending consensus transactions for which we
+    ///    wrote a signed effects digest to disk, we must re-execute using that
+    ///    digest as the expected effects digest, to ensure that we cannot
+    ///    arrive at different effects than what we previously signed.
     async fn reexecute_pending_consensus_certs(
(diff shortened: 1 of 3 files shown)


crates/iota-node/src/lib.rs
```diff
@@ -1530,47 +1530,68 @@ impl IotaNode {
...
-        let pending_consensus_certificates = epoch_store
-            .get_all_pending_consensus_transactions()
-            .into_iter()
-            .filter_map(|tx| {
-                match tx.kind {
-                    // shared object txns will be re-executed by consensus replay
-                    ConsensusTransactionKind::CertifiedTransaction(tx)
-                        if !tx.contains_shared_object() =>
+        let mut pending_consensus_certificates = Vec::new();
+        let mut additional_certs = Vec::new();
```
polinikita, line 1555: This seems to be unused (only pushes).

polinikita (approved):
Looks good, I would only remove one unused vector.

semenov-vladyslav (approved):
The fix was to increase the two timeouts. Alright then...

crates/iota-node/src/lib.rs
muXxer: this needs to stay!

crates/iota-node/src/lib.rs
```diff
@@ -1530,47 +1530,68 @@ impl IotaNode {
...
             .collect::<Vec<_>>();
 
-        info!("reexecuting pending consensus certificates: {:?}", digests);
+        info!(
+            "reexecuting {} pending consensus certificates: {:?}",
+            digests.len(),
+            digests
+        );
 
         state.enqueue_with_expected_effects_digest(pending_consensus_certificates, epoch_store);
```
muXxer, line 1596: We are missing something here!
```diff
state.enqueue_with_expected_effects_digest(pending_consensus_certificates, epoch_store);
+state.enqueue_transactions_for_execution(additional_certs, epoch_store);
```
Since we don't have that on our side, I think the equivalent is the following (please double check):
```diff
state.enqueue_with_expected_effects_digest(pending_consensus_certificates, epoch_store);
+state.transaction_manager().enqueue(additional_certs, epoch_store);
```

muXxer (changes requested):

crates/iota-node/src/lib.rs
VorobyevIlya: added a line and enqueue_transactions_for_execution function

crates/iota-node/src/lib.rs
VorobyevIlya: now it is used

muXxer (approved):

polinikita (approved):

PR #7614, opened 2025-07-01
refactor(iota-indexer): Move indexer RPC tests that should run in isolation

### LINKED ISSUE #6902: Indexer RPC tests with custom clusters may conflict with shared runtime globals
Currently, some Indexer RPC tests that run under the `shared_test_runtime` feature use the tokio runtime from `ApiTestSetup` to spawn a custom cluster due to inflexibilities of the provided `ApiTestSetup`.

This can lead to issues. For example:

The `get_chain_identifier` response from the node in the custom cluster may not match the response of the indexer.

The root cause is that the global `CHAIN_IDENTIFIER` is initialized once and reused. When a custom cluster is spawned, it can result in inconsistent behavior due to this global already being set. Also the other way, if a custom cluster with own runtime would be added to the RPC tests, it might impact existing tests relying on `ApiTestSetup`.

To resolve this, we should:

1. Identify all Indexer RPC tests that rely on this nested cluster setup.
2. Move these tests into a separate location.
3. Run them in isolation, outside of the `shared_test_runtime` feature.

samuel-rufi:
# Description of change

Most of our Indexer RPC tests run under the `shared_test_runtime` feature and share a single Tokio runtime via `ApiTestSetup::get_or_init()`. A handful of tests however spin up their own cluster with `start_test_cluster_with_read_write_indexer` (either alongside or instead of the shared runtime). Because of different globals in our crates (e.g. `OnceCell<ChainIdentifier>`, ...), these values are initialized exactly once in the first test and then reused:

- If a custom cluster test runs after the shared‐runtime tests, its globals are already pre-initialized, leading to wrong globals.
- If custom‐cluster tests run first, they set the global and might influence the other tests that rely on `ApiTestSetup`.

This PR extracts all tests that spawn a custom cluster and places them in `tests/rpc-tests-isolated/`. Cargo and nextest will compile each file into its own binary resulting into unset globals on start .

The isolated tests require the `pg_integration` feature and can be run in the indexer crate using `cargo nextest run --features pg_integration --test-threads 1`.

15 tests have been moved. The full Indexer nextest run ends within 2-3 minutes on my machine which seems acceptable.
As a follow up of this PR, we could investigate if we eventually could rewrite some of these tests in order to work with the shared test runtime and move those tests back again.

When introducing this change, care was taken not to break the CI test run or require any changes to the CI.

Fixes https://github.com/iotaledger/iota/issues/6902

**Tested it locally by setting up multiple tests that initialize the chain identifier global in the node.
Checked if the chain identifier RPC responses matches the chain identifier derived from the checkpoint 0 digest - which was the case for all the test instances.**

Doesn't touch ingestion or synchronization, so there is not need to check these test-classes.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


kodemartin (commented):
Hey @samuel-rufi, this is a big refactoring and brings about a significant increase in the test runtime as you reported.

While the problem with static `OnceCell` values initialized is certainly there, I think there might be alternative solutions that are simpler and more efficient.

For example, the `CHAIN_IDENTIFIER` static value is something that is specific to the node, and the node json-rpc uses it for relieving the node database from the respective queries of the checkpoint index.

If we could reset the value then it would solve the problem. And while we can't do that, we can workaround by querying the fullnode rpc for the genesis checkpoint and then calculating the chain-id for our tests. Something like:

```rust
     let fullnode_genesis = cluster.rpc_client().get_checkpoint(0.into()).await.unwrap();
     let fullnode_chain_identifier = ChainIdentifier::from(fullnode_genesis.digest).to_string();
     let indexer_chain_identifier = client.get_chain_identifier().await.unwrap();
 
     assert_eq!(fullnode_chain_identifier, indexer_chain_identifier);
```

That is ok, because we don't test the node JSON-RPC here, but the indexer JSON-RPC.

Simiilarly, if there are other static values that have a similar impact in the tests, it would seem more plausible to address the issues one-by-one, by trying to bypass the values (or reset them if possible) in the same way.

samuel-rufi:
Hey @kodemartin , thanks for the review! I share your view, the testing times are not ideal. As you said, we should rather simply handle each case one by one and reset the values if really needed / or use work arounds as you suggested (which I did for the pruning test if I remember correctly). As our tests currently seem to be fine (we don't notice flaky behavior) I think we should rather close this one. But I think it's good to know that flaky behavior might happen when working with globals directly or indirectly in our tests and we should be aware of that. So I think as a result we can agree on doing resets (if really needed) or do work arounds.
Therefore I will close this PR.

PR #7616, opened 2025-07-01
feat: Add an on-chain multisig check

miker83z:
# Description of change

Adds a new module for representing the MultisigPublicKey in Move and then check its weights and threshold correctly


account_abstraction/aa_move/sources/multisig_pubkey.move
```diff
@@ -0,0 +1,61 @@
...
+/// Create a new MultiSigPublicKey with the given public keys, weights, and threshold.
+public fun new(pks: vector<vector<u8>>, weights: vector<u8>, threshold: u16): MultiSigPublicKey {
+    assert!(
+        pks.length() > 0 
+        && weights.length() > 0 
+        && threshold > 0 
+        && pks.length() == weights.length() 
+        && pks.length() <= MAX_SIGNER_IN_MULTISIG
+        && weights.contains(&0) == false
+        && weights.fold!(0 as u16, |acc, e| acc + (e as u16)) == threshold,
```
lzpap: Why are we saying that the sum of the weights must equal threshold? Why not sum(weights) >= threshold?

account_abstraction/aa_move/sources/multisig_pubkey.move
```diff
@@ -0,0 +1,61 @@
...
+        EInvalidMultiSigPublicKey,
+    );
+    pks.zip_do_ref!(&vector::tabulate!(pks.length(), |i| i), |a, b| {
```
lzpap, line 31: If the intention here is to check for duplicate pks, then I'd rather just define `MultiSigPublikKey.pks` as a `VecSet<vector<u8>>`.

lzpap (approved):
Other than the two small comments, i could run it locally and works well!

account_abstraction/aa_move/sources/multisig_pubkey.move
miker83z: Yes, nice catch!

account_abstraction/aa_move/sources/multisig_pubkey.move
miker83z: Normally, with this trick you'd save some gas for largely sized vectors. However, here we limit the vector length to just 10 pks, so at the end it's was just a fancy way to refresh the my macros knowledge 😅

account_abstraction/aa_move/sources/multisig_pubkey.move
miker83z: [dac0004](https://github.com/iotaledger/iota/pull/7616/commits/dac000496c0772d5385214b47076a7b63f3c87cd)

account_abstraction/aa_move/sources/tx_flow.move
```diff
@@ -65,30 +67,31 @@ public fun sign_proposed_tx(
     ctx: &mut TxContext,
 ) {
     let proposed_tx: &mut ProposedTx = dof::borrow_mut(smart_account.id_mut(), tx_id);
+    assert!(proposed_tx.multisig.contains(&pk), EInvalidSignature);
```
Dkwcs, line 69: What is the point to check public key here?

Dkwcs (approved):
LGTM

account_abstraction/aa_move/sources/tx_flow.move
miker83z: Before this, anyone could execute this function with a valid ed25519 signature, not just Alice and Bob.

PR #7622, opened 2025-07-02
feat(dapp-kit): Add an opt-in flag to wait for the transaction in the `useSignAndExecuteTransaction` hook

marc2332:
This way we dont need to wrap the call to `signAndExecuteTransaction` inside a `useMutation` that also calls `client.waitForTransaction`, we can just use `useSignAndExecuteTransaction` directly from a normal event handler


msarcev (approved):

brancoder (approved):
LGTM

PR #7624, opened 2025-07-02
feat(iota-core): extend gas price feedback for multiple consensus commits

### LINKED ISSUE #6353: [Task (core-protocol-research)]: Extend gas price feedback mechanism to use data from multiple consensus commits
## Description

Extend the gas price feedback mechanism to use data from multiple consensus commits for gas price estimation.

## Motivation

If accepted, the current gas price feedback mechanism (https://github.com/iotaledger/iota/pull/6280 and https://github.com/iotaledger/iota/pull/6490) uses information only from a single consensus commit to calculate `suggested_gas_price`. This calculation can be improved by using data from multiple consensus commit rounds, leading to more accurate gas price suggestions.

## Requirements

1. For each consensus commit, collect congestion data: gas prices and estimated execution times of scheduled transactions.
2. Retain the data across multiple consensus commits.
3. Calculate suggested gas price for cancelled transactions.

## Open questions (optional)

One open question that I could not figure out how to solve in an optimal way is how to retain the data used for the calculations across multiple consensus commit. Specifically, In `AuthorityPerEpochStore`, we would like to store some data (e.g., a hash map keyed by commit round number, and the values correspond to congestion info per commit) that should be updated each commit, and it should be available across multiple consensus commits. We have been thinking about a few possible solutions:

- add a new field directly to `AuthorityPerEpochStore` as a `Mutex`,
- add a new table to `AuthorityEpochTables` to store the data,

but not sure which one (or something else) would be recommended.

Yes.

roman1e2f5p8s:
# Description of change

This PR extend the gas price feedback mechanism to use the data from multiple consensus commit rounds for the suggested gas price calculations.

Instead of calculating the suggested gas price using the data from a single commit round (the one in which the transaction is cancelled), in this PR, we calculate the suggested gas price at each commit a transaction is deferred, save the suggested gas price to the deferred transactions table, and update it every time a lower suggested gas price is found.

Closes #6353.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -154,6 +154,41 @@ impl CertLockGuard {
 
 type JwkAggregator = GenericMultiStakeAggregator<(JwkId, JWK), true>;
 
+/// An alias type for a collection used to hold previously deferred
+/// transactions, where `Option<u64>` is used to hold suggested gas
+/// price for transactions deferred due to shared object congestion
+/// (`None` for transactions deferred due to randomness not available).
```
muXxer:
```suggestion
/// (`None` for transactions deferred due to "randomness not available").
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -154,6 +154,41 @@ impl CertLockGuard {
...
+/// Holds a verified sequenced consensus transaction that is deferred
+/// and optionally a suggested gas price for that transaction.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DeferredTransaction {
+    /// Deferred verified sequenced consensus transaction.
+    transaction: VerifiedSequencedConsensusTransaction,
+
+    /// Suggested gas price is `Some(u64)` for transactions deferred due
+    /// to shared object congestion, and it is `None` for transactions
+    /// deferred due to randomness not available.
```
muXxer:
```suggestion
    /// deferred due to "randomness not available".
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -192,8 +227,14 @@ pub enum ConsensusCertificateResult {
         start_time: ExecutionTime,
     },
     /// The transaction should be re-processed at a future commit, specified by
-    /// the DeferralKey
-    Deferred(DeferralKey),
+    /// `deferral_key`. `suggested_gas_price_as_opt` indicates a gas price that
+    /// the certificate would need to pay to be scheduled in a consensus commit.
+    /// For certificates deferred due to randomness not available, the
```
muXxer:
```suggestion
    /// For certificates deferred due to "randomness not available", the
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -615,7 +656,7 @@ pub struct AuthorityEpochTables {
     active_jwks: DBMap<(u64, (JwkId, JWK)), ()>,
 
     /// Transactions that are being deferred until some future time
-    deferred_transactions: DBMap<DeferralKey, Vec<VerifiedSequencedConsensusTransaction>>,
+    deferred_transactions: DBMap<DeferralKey, Vec<DeferredTransaction>>,
```
muXxer: I guess that won't work.

Nodes won't be able to ever load old epochs this way, because you change the serialised representation of the data.
I would say you need to introduce `deferred_transactions_v2` instead, and use that based on some feature flag.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3482,33 +3538,47 @@ impl AuthorityPerEpochStore {
...
+                                    deferral_key,
+                                    suggested_gas_price_as_opt: None,
+                                }
                             }
                             DeferralReason::SharedObjectCongestion(congested_objects) => {
                                 authority_metrics
                                     .consensus_handler_congested_transactions
                                     .inc();
+
+                                // FIX: roman: call suggested gas price calculator here
```
muXxer: Is there something missing or can this comment be removed?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3482,33 +3538,47 @@ impl AuthorityPerEpochStore {
...
+                                let current_commit_suggested_gas_price = self.reference_gas_price();
+                                let suggested_gas_price = previously_deferred_tx_digests
+                                    .get(certificate.digest())
+                                    .map_or_else(
+                                        || current_commit_suggested_gas_price,
+                                        |deferral_key_suggested_gas_price_pair| {
+                                            deferral_key_suggested_gas_price_pair.1.expect(
```
muXxer: And we are 100% sure this will never happen? :)

muXxer (changes requested):

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: Can be removed. I will do that.

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: I would say so. If a certificate has already been deferred, it means its digest is present in `previously_deferred_tx_digests`. Second, for txs deferred due to congestion, we always set `Some(suggested_gas_price)`, so this unwrapping should never panic.

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: Hmm, do nodes store and load deferred transactions in/from perpetual tables? I thought deferred transactions live only within an epoch, as `AuthorityEpochTables`'s docstring says the epoch tables contain data that is only valid within an epoch. I would like to understand where in the code we save deferred transactions from one epoch to another, because I could not find that.

piotrm50: Yes, but you still need to restart the node at some point during an epoch to deploy a new version of the node. Ideally, you want to do that without cleaning the databases, as this would lead to more downtime (imagine that 2/3rd of the nodes do this). Without introducing _v2 as Max suggested, if you don't clean the database then the node won't be able to start correctly as it won't correctly read data from the current epoch's DB.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: So we only put transactions into this map if they are deferred due to congestion, and transactions deferred due to missing randomness are not present here, right?

piotrm50 (changes requested):
Please introduce the _v2 DB table as suggested by Max.

cyberphysic4l (commented):
The logic all looks good to me. Only comment would be agreeing with the comments about needing a v2 for that deffered transaction type.

roman1e2f5p8s:
Converted to draft as it needs more work, namely, to introduce `deferred_transactions_v2` table.

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: Got it. Thanks for the explanation.
I will add `deferred_transactions_v2` DB table as Max suggested.

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: Transactions deferred due to unavailable randomness **are also added** to the same `deferred_transactions` DB map and the same `previously_deferred_tx_digests` map (if this is what you mean by a 'map' here) as transactions deferred due to congestion.

We only put suggested gas price (always as `Some(suggested_gas_price)`) for transactions deferred/cancelled due to congestion, because that gas price is the one the sender would have to pay to get their tx scheduled, and we can infer some value from the sequencer outputs.

For transactions deferred/cancelled due to unavailable randomness, we cannot suggest any gas price, as the decision of whether to defer or schedule a transaction that uses randomness does not depend on tx's gas price in any way. So, for transactions deferred due to unavailable randomness, we always attach `None` as suggested gas price. We need to do that because both types of deferred transactions are put in the same basket/map.

Going back to the Max's question here. If a transaction is deferred due to congestion for the first time, we are using the suggested gas price from the current consensus commit. But if the transaction has already been deferred, that means we have `Some(suggested_gas_price)` from the previous commit round, and we just compare it with the suggested gas price from the current commit and take the minimum. So, the unwrapping in question should never fail as long as we always set `Some(suggested_gas_price)` for transactions deferred due to congestion, which is what we are doing there. Hope that is more clear now.

piotrm50 (approved):

cyberphysic4l (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3482,34 +3581,62 @@ impl AuthorityPerEpochStore {
...
+                                    }
                                 } else {
                                     // Cancel the transaction that has been deferred for too long.
 
-                                    let suggested_gas_price = shared_object_congestion_tracker
-                                        .compute_suggested_gas_price(&certificate)
-                                        .expect(
-                                            "cancelled transaction must have at least one shared \
-                                            object and calculated suggested gas price",
-                                        );
```
muXxer, line 3505: Wouldn't that change the behaviour, or at least remove that check, as long as `congested_objects_gas_price_feedback_mechanism` isn't enabled?

muXxer (changes requested):

crates/iota-core/src/authority/authority_per_epoch_store.rs
muXxer: Ahh I see, this change is from the feature branch, not from develop, sorry.

muXxer (approved):

PR #7632, opened 2025-07-02
feat(iota-framework): new type AccountRegistry has been added.

### LINKED ISSUE #7607: [AA-prototype][account] Create a system AccountRegistry shared object to manage precomputed object id collisions
Based on the prototype #7441, add a new feature:
Part 1:
- Add a new type to the framework, e.g., AccountRegistry, and then add it as parameter to the `new_precomputed` function
  - This is a unique shared object that is passed every time someone wants to create a new precomputed id
  - This shared object contains simply a `Table` of `ID`s;
  - During the execution of `new_precomputed`, if the newly created `ID` is present in this `Table`, then the `new_precomputed` must fail, otherwise add the new `ID` into the `Table`
- The AccountRegistry type is implemented as part of the iota-framework and has a known object id
  - Ideally the object id should be hardcoded with something like 0x7 but for this prototype can also be random
  - This should be something like the `public struct Clock has key` in `clock.move` in the framework; then the instance of the type, i.e., the unique system shared object, should be passed as parameter to the `new_precomputed` function in `object.move`.
- Create 2 PTBs to test this flow, one that successfully creates a new id and one that fails

Part 2:
- In a second iteration, after that the first part works, make it so that when an Account object is destroyed then the `ID` is removed from the `Table`

Dkwcs:
# Description of change

New type AccountRegistry has been added and passed as a parameter to the `new_precomputed` function.
AccountRegistry - is a unique shared object that is passed every time someone wants to create a new precomputed id.
AccountRegistry ID - is hardcoded with 0x10.
`deterministic_object_id_tests.move` tests have been updated.

Fixes #7607 .

Run Move tests - `deterministic_object_id_tests.move`


crates/iota-framework/packages/iota-framework/sources/account_registry.move
```diff
@@ -0,0 +1,66 @@
...
+
+/// Singleton shared object that exposes time to Move calls.
+public struct AccountRegistry has key {
+    id: UID,
+    // Ids of generated accounts
+    account_ids: Table<ID, ID>,
+}
+
+/// Create and share the singleton AccountRegistry -- this function is
+/// called exactly once, during genesis.
```
Thoralf-M, line 23: So how do we get this in the existing networks if it runs only during genesis?

crates/iota-framework/packages/iota-framework/sources/account_registry.move
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/// Add new account id to the table
+public fun add(self: &mut AccountRegistry, account_id: ID) {
+    assert!(!self.account_ids.contains(account_id), EIdAlreadyExists);
+    self.account_ids.add(account_id, account_id); // Probably we need another collection
+}
+
+/// Remove account id from the table
+public fun remove(self: &mut AccountRegistry, account_id: ID) {
```
Thoralf-M, line 42: This needs to be somehow protected, so only the account owner can remove it?

PR #7667, opened 2025-07-03
upstream(core): Return BTreeMap<ObjectID, &Object> in checkpoint_input_objects

jkrvivian:
# Description of change

- Port commit: https://github.com/MystenLabs/sui/commit/f66f5ea889eb046faf4cdd349643880fa19fe157

NOTE: we don't have iota-indexer-alt crate, so only changes in iota-types are applied.

Part of #4390  

### DIFF
--- a/crates/iota-types/src/full_checkpoint_content.rs
+++ b/crates/iota-types/src/full_checkpoint_content.rs
@@ -94,7 +94,7 @@ pub struct CheckpointTransaction {
     pub transaction: Transaction,
     /// The effects produced by executing this transaction
     pub effects: TransactionEffects,
-    /// The events, if any, emitted by this transactions during execution
+    /// The events, if any, emitted by this transaction during execution
     pub events: Option<TransactionEvents>,
     /// The state of all inputs to this transaction as they were prior to
     /// execution.
--- a/crates/iota-types/src/full_checkpoint_content.rs
+++ b/crates/iota-types/src/full_checkpoint_content.rs
@@ -2,13 +2,13 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 
 use serde::{Deserialize, Serialize};
 use tap::Pipe;
 
 use crate::{
-    base_types::ObjectRef,
+    base_types::{ObjectID, ObjectRef},
     effects::{
         IDOperation, ObjectIn, ObjectOut, TransactionEffects, TransactionEffectsAPI,
         TransactionEvents,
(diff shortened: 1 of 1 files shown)


muXxer (approved):

crates/iota-types/src/full_checkpoint_content.rs
```diff
@@ -57,11 +57,24 @@ impl CheckpointData {
...
     }
 
-    pub fn input_objects(&self) -> Vec<&Object> {
-        self.transactions
-            .iter()
-            .flat_map(|tx| &tx.input_objects)
-            .collect()
+    /// Returns all objects that are used as input to the transactions in the
+    /// checkpoint, and already exist prior to the checkpoint.
+    pub fn checkpoint_input_objects(&self) -> BTreeMap<ObjectID, &Object> {
```
semenov-vladyslav, line 62: What is the point of this function? Why is it so complicated? Why does it need additional hash set `output_objects_seen`?

crates/iota-types/src/full_checkpoint_content.rs
```diff
@@ -57,11 +57,24 @@ impl CheckpointData {
...
+        let mut output_objects_seen = HashSet::new();
+        let mut checkpoint_input_objects = BTreeMap::new();
+        for tx in self.transactions.iter() {
+            for obj in tx.input_objects.iter() {
+                let id = obj.id();
+                if output_objects_seen.contains(&id) || checkpoint_input_objects.contains_key(&id) {
```
semenov-vladyslav, line 68: Why not just `checkpoint_input_objects.insert(id, obj);`? If we need to keep the first object (in case there are multiple for the same id), we can use `entry` and insert into only vacant entries.

crates/iota-types/src/full_checkpoint_content.rs
```diff
@@ -79,7 +92,7 @@ pub struct CheckpointTransaction {
     pub transaction: Transaction,
     /// The effects produced by executing this transaction
     pub effects: TransactionEffects,
-    /// The events, if any, emitted by this transaction during execution
+    /// The events, if any, emitted by this transactions during execution
```
semenov-vladyslav:
```suggestion
    /// The events, if any, emitted by this transaction during execution
```

semenov-vladyslav (changes requested):

crates/iota-types/src/full_checkpoint_content.rs
jkrvivian: Fixed in [56d4e76](https://github.com/iotaledger/iota/pull/7667/commits/56d4e766d3cbe64c8554aad79b3697910f554761)

crates/iota-types/src/full_checkpoint_content.rs
jkrvivian: Fixed in [56d4e76](https://github.com/iotaledger/iota/pull/7667/commits/56d4e766d3cbe64c8554aad79b3697910f554761)

crates/iota-types/src/full_checkpoint_content.rs
jkrvivian: Fixed in [56d4e76](https://github.com/iotaledger/iota/pull/7667/commits/56d4e766d3cbe64c8554aad79b3697910f554761)

crates/iota-types/src/full_checkpoint_content.rs
```diff
@@ -57,11 +57,16 @@ impl CheckpointData {
...
+    /// Returns all objects that are used as input to the transactions in the
+    /// checkpoint, and already exist prior to the checkpoint.
+    pub fn checkpoint_input_objects(&self) -> BTreeMap<ObjectID, &Object> {
+        let mut checkpoint_input_objects = BTreeMap::new();
+        for tx in self.transactions.iter() {
+            for obj in tx.input_objects.iter() {
+                checkpoint_input_objects.entry(obj.id()).or_insert(obj);
+            }
+        }
+        checkpoint_input_objects
```
DaughterOfMars: This can now be further simplified

```suggestion
        self.transactions
            .iter()
            .flat_map(|tx| tx.input_objects.iter().map(|obj| (obj.id(), obj)))
            .collect()
```

muXxer: Unfortunately your suggestion is outdated because the former simplification which was done was actually wrong and we reverted it.

semenov-vladyslav (approved):

muXxer (approved):

PR #7697, opened 2025-07-03
feat(iota-types): introduce `CancelledTransactionsV2`

roman1e2f5p8s:
# Description of change

Currently existing (tuple) variant `CancelledTransactions(Vec<(TransactionDigest, Vec<(ObjectID, SequenceNumber)>)>)` looks complex and is not very readable and self-documented. This PR adds a new (struct) variant to the `ConsensusDeterminedVersionAssignments` enum. Additionally, since versions of cancelled objects are pretty big numbers (`larger than u64::MAX / 2`), this PR also adds serialization of `SequenceNumber` as `BigInt<u64>`. 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-types/src/messages_consensus.rs
+++ b/crates/iota-types/src/messages_consensus.rs
@@ -16,12 +16,14 @@ use fastcrypto_tbls::dkg_v1;
 use fastcrypto_zkp::bn254::zk_login::{JWK, JwkId};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 
 use crate::{
     base_types::{
         AuthorityName, ConciseableName, ObjectID, ObjectRef, SequenceNumber, TransactionDigest,
     },
     digests::ConsensusCommitDigest,
+    iota_serde::BigInt,
     messages_checkpoint::{
         CheckpointSequenceNumber, CheckpointSignatureMessage, CheckpointTimestamp,
     },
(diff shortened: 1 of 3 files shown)


crates/iota-types/src/messages_consensus.rs
```diff
@@ -31,12 +33,38 @@ use crate::{
...
 
+/// Holds object ID and assigned version for cancelled shared objects.
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+pub struct VersionAssignment {
+    /// ID of shared object that appear in cancelled transactions.
+    pub object_id: ObjectID,
+
+    /// Version assigned to cancelled shared object. It embeds the
+    /// cancellation reason.
```
Thoralf-M, line 44: > It embeds the cancellation reason

What does this mean? Is it encoded in the number?

roman1e2f5p8s: Yes. Looking at the sequence number of a cancelled object, one can infer the cancellation reason.

Thoralf-M (approved):

piotrm50 (commented):
Is this going to be used in some future PR?

roman1e2f5p8s:
> Is this going to be used in some future PR?

Not really. The main idea behind introducing these changes is two-fold: (i) replace the tuple variant with a struct variant (see how it is done in [iota-rust-sdk](https://github.com/iotaledger/iota-rust-sdk/blob/cd97687c9316045351d7db8b389ea381a5c335e1/crates/iota-rust-sdk/src/types/transaction/mod.rs#L208)), and (ii) serialize versions of cancelled objects as `BigInt` since they are `> u64::MAX / 2`.

roman1e2f5p8s:
Converted to draft as it needs additional changes:

- protocol config feature flag;
- identical changes on the `iota-rust-sdk` side.

alexsporn:
I don't see the value this PR brings. It is not adding any new functionality, just replacing a tuple with a struct. I am not a friend of changing the protocol types without an explicit reason. So if this PR is not introducing new features I would skip it.

roman1e2f5p8s:
> I don't see the value this PR brings. It is not adding any new functionality, just replacing a tuple with a struct. I am not a friend of changing the protocol types without an explicit reason. So if this PR is not introducing new features I would skip it.

I see your point. Another reason behind this PR is serialization of cancelled objects' versions as `BigInt`: if this also does not justify the PR, we can close it. But I would then propose to make these changes whenever we introduce `ConsensusCommitPrologueV2`.

roman1e2f5p8s:
It is better to make these changes during the introduction of `ConsensusCommitPrologueV2`.
Added reminder about that in https://github.com/iotaledger/iota/pull/8065.

PR #7704, opened 2025-07-04
feat(keys)!: introduce a new keystore v2 format and migrate from the current one while keeping a backup of the files

### LINKED ISSUE #5849: Investigate adding Ledger support to CLI
See how feasible it is to use a ledger nano to generate addresses/sign transactions, for example to use in `iota client ptb`
Potentially useful resources
https://github.com/ledger-community/rust-ledger
https://github.com/alamgu/alamgu-generic-cli/tree/main/src/commands
https://github.com/iotaledger/ledger.rs (what we used for Stardust, no stax/flex support)

alexsporn:
# Description of change

* Introduce a new keystore v2 format. If a keystore v1 file (`iota.keystore` + `iota.aliases`) is found then they are automatically migrated while keeping a copy of the old data named as (`iota.keystore.migrated` + `iota.aliases.migrated`).

```json
{
  "version": 2,
  "keys": [
    {
      "alias": "xenodochial-amethyst",
      "address": "0x6a0e268bfec6dc7da4ab9b8af60e345a55f5cc34793f1d9c21f7f8faee6f4b68",
      "key": {
        "type": "external",
        "value": {
          "source": "ledger",
          "derivation_path": "m/44'/4218'/0'/0'/0'",
          "public_key_base64": "AA2itppp6zFjp9f6onzcTSyDbp8KymT2gCpQAHj7KL+o"
        }
      }
    },
    {
      "alias": "zealous-agates",
      "address": "0xaa0ebe1d2a37d4726a47b18b80ef23dbbd773d5594429ac738d69861dc9d49ce",
      "key": {
        "type": "external",
        "value": {
          "source": "some_kms",
          "public_key_base64": "AMkjtEDVFIxuAIbTTUHxCU0zVxujbL6A5ovdB/9rRxLz"
        }
      }
    },
    {
      "alias": "condescending-jet",
      "address": "0xfeebd0a88fb283e698e142bbfa87f27b76ac675cda361c317a6f0419dee0a699",
      "key": {
        "type": "key_pair",
        "value": "iotaprivkey1qpkfvwk4re9hd04c900hy4d2h6qtyqks0j58gr877xs6668q6czg2vt6ysm"
      }
    }
  ]
}
```

* The new keystore v2 supports adding keys from external sources, where only the public key is added to the store and an optional bip32 derivation path. This allows the CLI to store references to keys managed from e.g. Ledger hardware wallets.

Fixes #5849

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Introduces a new keystore v2 format. If a keystore v1 file (`iota.keystore` + `iota.aliases`) is found then they are automatically migrated while keeping a copy of the old data named as (`iota.keystore.migrated` + `iota.aliases.migrated`).
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-keys/src/keystore.rs
```diff
@@ -425,62 +523,165 @@ impl FileBasedKeystore {
         })
     }
 
-    pub fn set_path(&mut self, path: &Path) {
-        self.path = path.to_path_buf();
+    fn needs_migration(path: &PathBuf) -> Result<bool, anyhow::Error> {
```
Thoralf-M, line 536: This file got quite large now, mind to split it into multiple files?
Maybe something like this
```
keystore/
	mod.rs
	file-based.rs
	in-memory.rs
	migration.rs
```

alexsporn: I can do it for sure, but it will make the diff not really readable.
Mind if I do this in a separate PR after these changes are done?

crates/iota-keys/src/keystore.rs
```diff
@@ -40,11 +42,10 @@ pub enum Keystore {
 
 #[enum_dispatch]
 pub trait AccountKeystore: Send + Sync {
-    fn add_key(&mut self, alias: Option<String>, keypair: IotaKeyPair)
-    -> Result<(), anyhow::Error>;
+    fn add_key(&mut self, alias: Option<String>, key: StoredKey) -> Result<(), anyhow::Error>;
```
thibault-martinez: If you take an `impl Into<StoredKey>` instead it would be backwards compatible, you wouldn't need to sprinkle `.into()` everywhere, and it would be generally easier to use

crates/iota-keys/src/keystore.rs
```diff
@@ -40,11 +42,10 @@ pub enum Keystore {
...
     fn remove_key(&mut self, address: &IotaAddress) -> Result<(), anyhow::Error>;
     fn keys(&self) -> Vec<PublicKey>;
-    fn get_key(&self, address: &IotaAddress) -> Result<&IotaKeyPair, anyhow::Error>;
+    fn get_key(&self, address: &IotaAddress) -> Result<&StoredKey, anyhow::Error>;
```
thibault-martinez, line 53: This is a breaking change, if we're fine with it

crates/iota-keys/src/keystore.rs
Thoralf-M: Sure

crates/iota-keys/src/keystore.rs
alexsporn: Yes, I would also like to change `fn keys()` to also return `StoredKey` instead

crates/iota-keys/src/keystore.rs
alexsporn: Nice trick, thanks! https://github.com/iotaledger/iota/pull/7704/commits/d9f7b83c135d5112ee3c185d9d84ed6463696ca4

crates/iota-keys/src/serde_derivation_path.rs
```diff
@@ -0,0 +1,54 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        let opt = Option::<String>::deserialize(deserializer)?;
+        match opt {
+            Some(s) => Ok(Some(
+                DerivationPath::from_str(&s).map_err(serde::de::Error::custom)?,
+            )),
+            None => Ok(None),
+        }
```
thibault-martinez: You could write the whole function body like this, shorter and you can get rid of the `#[expect(dead_code)]` L18
```suggestion
    Option::<String>::deserialize(deserializer)?
        .map(|s| super::deserialize(serde::de::IntoDeserializer::into_deserializer(s)))
        .transpose()
```

alexsporn: Done in https://github.com/iotaledger/iota/pull/7704/commits/8658001120baa7dab483360887ec7ef10dff779b
I also added some tests in the serde files

crates/iota/src/client_commands.rs
```diff
@@ -3057,11 +3057,20 @@ pub(crate) async fn dry_run_or_execute_or_serialize(
...
+                StoredKey::KeyPair(_) => context.config().keystore().sign_secure(
+                    &tx_data.sender(),
+                    &tx_data,
+                    Intent::iota_transaction(),
+                )?,
+                StoredKey::External { source, .. } => {
+                    bail!("External signing is not supported for source: {source}")
+                }
+            }
+        };
```
Thoralf-M, line 3073: Isn't the match in sign_secure() be sufficient?
```Rust
        match stored_key {
            StoredKey::KeyPair(keypair) => Ok(Signature::new_secure(intent_msg, keypair)),
            StoredKey::External { source, .. } => Err(signature::Error::from_source(format!(
                "sign_secure is not supported for external type: {source} [{address}]",
            ))),
        }
```

crates/iota/src/keytool.rs
```diff
@@ -742,8 +765,17 @@ impl KeyToolCommand {
...
+
+                let key = keystore.get_key(&address)?;
+                let iota_signature = match key {
+                    StoredKey::KeyPair(_) => {
+                        keystore.sign_secure(&address, &intent_msg.value, intent_msg.intent)?
+                    }
+                    StoredKey::External { source, .. } => {
+                        bail!("External signing is not supported for source: {source}")
+                    }
+                };
```
Thoralf-M, line 777: Same here, isn't the match in sign_secure() for the key type sufficient?

crates/iota/src/client_commands.rs
Thoralf-M: Just saw that you even replaced it in the ledger PR

alexsporn: The idea was to already prepare the CLI for matching specific external sources. And yes, this is already implemented in the other PR that builds on top of this one.

crates/iota/src/keytool.rs
alexsporn: See above, this is because I split the PR into 2 and kept the structure prepared to match over the external source

crates/iota/src/client_commands.rs
Thoralf-M: Ah ok a new sign_secure() is only implemented in the ledger PR, didn't compare side by side before and thought it's just calling the same again, reverting this change

crates/iota-keys/src/keystore.rs
```diff
@@ -175,9 +197,101 @@ pub struct Alias {
...
+    KeyPair(IotaKeyPair),
+    External {
+        source: String,
+        #[serde(
+            with = "serde_derivation_path::option",
+            default,
+            skip_serializing_if = "Option::is_none"
+        )]
+        derivation_path: Option<DerivationPath>,
+        public_key_base64: String,
```
Thoralf-M: How about adding `with_flag` here? At least in your example the values contain the prefix byte for the type
And just to make sure, for IotaKeyPair you removed it, because we can just always compute it on the fly and therefore don't need it in the keystore file, right?

alexsporn: This is the same representation and naming that was in the aliases file. It is made so that we can match which public key is behind the address since we can't derive it if it is not a keypair. It uses the flag because it needs to support any type of public key. We can rename it to `with_flag` here if you think it would be more consistent with the keytool output, but I think that is nothing that should concern any user.

Thoralf-M: Yes please rename, already created confusions if the type flag was there, but not mentioned

crates/iota-keys/src/keystore.rs
```diff
@@ -175,9 +197,101 @@ pub struct Alias {
...
+    },
+}
+
+impl Clone for StoredKey {
+    fn clone(&self) -> Self {
+        match self {
+            StoredKey::KeyPair(keypair) => StoredKey::KeyPair(keypair.copy()),
```
DaughterOfMars: Not really worth implementing manually just to copy here.

crates/iota-keys/src/keystore.rs
```diff
@@ -175,9 +197,101 @@ pub struct Alias {
...
+    }
+
+    pub fn external_source(&self) -> Option<String> {
+        match self {
+            StoredKey::KeyPair(_) => None,
+            StoredKey::External { source, .. } => Some(source.clone()),
+        }
+    }
+
+    pub fn try_keypair(&self) -> Result<&IotaKeyPair, anyhow::Error> {
```
DaughterOfMars: Can we rename this to `try_as_keypair` or just `as_keypair`?

crates/iota-keys/src/keystore.rs
```diff
@@ -329,13 +465,13 @@ impl AccountKeystore for FileBasedKeystore {
...
         let new_alias_name = self.update_alias_value(old_alias, new_alias)?;
-        self.save_aliases()?;
+        self.save()?;
         Ok(new_alias_name)
     }
 }
 
 impl FileBasedKeystore {
-    pub fn new(path: &PathBuf) -> Result<Self, anyhow::Error> {
+    pub fn new_from_v1(path: &PathBuf) -> Result<Self, anyhow::Error> {
```
DaughterOfMars, line 446: What is `v1` referring to here? Maybe better to call this something more specific

crates/iota-keys/src/serde_derivation_path.rs
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2025 IOTA Stiftung
```
DaughterOfMars: I don't think this file is necessary. There are existing impls for types that impl `ToString` and `FromStr`

crates/iota/src/validator_commands.rs
```diff
@@ -213,37 +213,42 @@ impl IotaValidatorCommand {
...
-                };
-                let account_key_file_name = dir.join("account.key");
+
+                let account_key = context.config().keystore().get_key(&iota_address)?;
+
+                if let StoredKey::KeyPair(keypair) = account_key {
+                    let account_keypair = match keypair {
+                        IotaKeyPair::Ed25519(pair) => IotaKeyPair::Ed25519(pair.copy()),
+                        _ => panic!(
+                            "Other account key types supported yet, please use Ed25519 keys for now."
```
DaughterOfMars: Missed a few words here

crates/iota-keys/src/keystore.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/7704/commits/428675e134184eb1568b6863847796c99390855e

Thoralf-M (approved):

crates/iota-keys/src/keystore.rs
alexsporn: v1 is the old keystore format, v2 is the new one

crates/iota-keys/src/serde_derivation_path.rs
alexsporn: We either implement this serde serialization for the foreign type or we have to use `String` everywhere and have getters that convert. I found this solution better.

crates/iota/src/validator_commands.rs
alexsporn: Lol, I just moved the existing panic and did not read the message. Changed it in https://github.com/iotaledger/iota/pull/7704/commits/2f3c5ab74deb630e424f71a82d80061047b28665

crates/iota-keys/src/keystore.rs
alexsporn: Implemented Clone trait on IotaKeypair https://github.com/iotaledger/iota/pull/7704/commits/32195ce0e8a4ca4f88d81d4b2b8ec4944f806913

crates/iota-keys/src/serde_derivation_path.rs
alexsporn: Replaced this file with:
```
#[serde_as(as = "Option<DisplayFromStr>")]
#[serde(skip_serializing_if = "Option::is_none")
```

https://github.com/iotaledger/iota/pull/7704/commits/d2c49af468ac8003e1a5a7dffedde2daf310502c

crates/iota-keys/src/keystore.rs
alexsporn: Renamed to `as_keypair` in https://github.com/iotaledger/iota/pull/7704/commits/62feef7bd7be78d9c04c241d31d1bc3ce062d8d9

crates/iota-rosetta/src/main.rs
```diff
@@ -208,37 +198,36 @@ async fn wait_for_iota_client(rpc_address: String) -> IotaClient {
...
-        .iter()
-        .map(|kpstr| {
-            let key = IotaKeyPair::decode(kpstr);
-            key.map(|k| (IotaAddress::from(&k.public()), k))
-        })
-        .collect::<Result<BTreeMap<_, _>, _>>()
-        .unwrap();
+fn read_prefunded_account(path: &PathBuf) -> Result<Vec<PrefundedAccount>, anyhow::Error> {
+    let key_store =
+        FileBasedKeystore::new(path).map_err(|e| anyhow!("Failed to open keystore file: {}", e))?;
```
Thoralf-M:
```suggestion
        FileBasedKeystore::new(path).map_err(|e| anyhow!("Failed to open keystore file: {e}"))?;
```

crates/iota-keys/src/keystore.rs
```diff
@@ -175,9 +198,73 @@ pub struct Alias {
...
+    #[serde(with = "serde_iota_keypair")]
+    KeyPair(IotaKeyPair),
+    External {
+        source: String,
+
+        #[serde_as(as = "Option<DisplayFromStr>")]
+        #[serde(skip_serializing_if = "Option::is_none")]
+        derivation_path: Option<DerivationPath>,
+        #[serde(rename = "public_key_base64_with_flag", with = "serde_publickey")]
+        public_key: PublicKey,
```
DaughterOfMars: spacing is weird here

crates/iota-rosetta/src/main.rs
alexsporn: Done in https://github.com/iotaledger/iota/pull/7704/commits/db926050524a96d631e0116ba1f39d72159cd22c

crates/iota-keys/src/keystore.rs
alexsporn: Changed in https://github.com/iotaledger/iota/pull/7704/commits/db926050524a96d631e0116ba1f39d72159cd22c

DaughterOfMars (approved):

muXxer (approved):
Files owned by us look good.

crates/iota-keys/src/serde_publickey.rs
thibault-martinez, line 1: IMO should rename this file to `serde_public_key.rs` as it's inconsistent with our usage of `public_key` and `PublicKey`

crates/iota-rosetta/src/main.rs
```diff
@@ -208,37 +198,36 @@ async fn wait_for_iota_client(rpc_address: String) -> IotaClient {
...
-                }
-                IotaKeyPair::Secp256k1(k) => {
-                    (Hex::encode(k.private().as_bytes()), CurveType::Secp256k1)
-                }
-                IotaKeyPair::Secp256r1(k) => {
-                    (Hex::encode(k.private().as_bytes()), CurveType::Secp256r1)
+        .filter_map(|(address, _)| {
+            let key = key_store.get_key(address).ok()?;
+            match key.as_keypair() {
+                Ok(keypair) => {
```
kodemartin: Nit: You can simplify to
```rust
let keypair = key_store.get_key(address).ok()?.as_keypair().ok()?;
let (privkey, curve_type) = ...
```

kodemartin (approved):
lgtm

crates/iota/src/client_commands.rs
```diff
@@ -3057,11 +3057,20 @@ pub(crate) async fn dry_run_or_execute_or_serialize(
             tx_data,
         ))
     } else {
-        let signature = context.config().keystore().sign_secure(
-            &tx_data.sender(),
-            &tx_data,
-            Intent::iota_transaction(),
-        )?;
+        let signature = {
+            let key = context.config().keystore().get_key(&tx_data.sender())?;
```
thibault-martinez, line 3061: Can't you just call `sign_secure` on the keystore?

crates/iota/src/validator_commands.rs
```diff
@@ -213,37 +213,42 @@ impl IotaValidatorCommand {
...
-                    IotaKeyPair::Ed25519(account_key) => IotaKeyPair::Ed25519(account_key.copy()),
-                    _ => panic!(
-                        "Other account key types supported yet, please use Ed25519 keys for now."
-                    ),
-                };
-                let account_key_file_name = dir.join("account.key");
+
+                let account_key = context.config().keystore().get_key(&iota_address)?;
+
+                if let StoredKey::KeyPair(keypair) = account_key {
```
miker83z, line 223: Shouldn't it check the pub key being ed25519 also for the `StoredKey::External` variant?

miker83z (changes requested):

crates/iota-rosetta/src/main.rs
alexsporn: Done in https://github.com/iotaledger/iota/pull/7704/commits/4e999cc7fa9f05dcced516692dc0a9c6fa0f31af

crates/iota-keys/src/serde_publickey.rs
alexsporn: what about `serde_iota_keypair.rs`? `KeyPair` is used but then also `keypair`

thibault-martinez: You are right, it should be `key_pair`, `publickey` caught my attention more. I think the file name should follow the type name but I'll leave it to you.

crates/iota/src/validator_commands.rs
alexsporn: Tbh I don't see a reason why non-ed25519 would not be allowed to work. The protocol only cares about an `IotaAdddress` and not about any public key for the account. The proof-of-possession is done over the address and the address just needs to be able to hold the `OperationCap`.

TheMrAI (approved):

crates/iota/src/validator_commands.rs
```diff
@@ -213,37 +213,42 @@ impl IotaValidatorCommand {
...
+                    let account_keypair = match keypair {
+                        IotaKeyPair::Ed25519(pair) => IotaKeyPair::Ed25519(pair.copy()),
+                        _ => panic!(
+                            "Only Ed25519 accounts are supported, please use Ed25519 keys for now."
+                        ),
+                    };
+                    let account_key_file_name = dir.join("account.key");
+                    make_key_files(account_key_file_name.clone(), false, Some(account_keypair))?;
```
thibault-martinez: Is it actually fine to call this command with an external stored key? We wouldn't have this file. I'm not too versed into the validator steps I don't know if it's relevant for further use.

alexsporn: Yes, this file is not necessary, it is just a "backup" copy of the key from the keystore that you used. The validators won't need it and it is actually not encouraged to have it on the validator since it is the key that controls the rewards the validator earns.

crates/iota-keys/src/serde_publickey.rs
alexsporn: Done in https://github.com/iotaledger/iota/pull/7704/commits/3ff2c89d97b63f6ebb6d02b68f8034829c6c35bf

Kept keypair as is, because we don't have a consistent naming in the repo for that right now

crates/iota/src/validator_commands.rs
miker83z: I don't remember if there are other implicit checks about the account key being ed25519 in the code, probably not. Nonetheless, this change currently accepts non-ed25519 StoredKey::External keys and rejects non-ed25519 StoredKey::KeyPair keys, so should be harmonized.

alexsporn: Done in https://github.com/iotaledger/iota/pull/7704/commits/55f4382b2230e25b0a36fa28a177fc6eeef962b1

crates/iota-keys/src/keystore.rs
```diff
@@ -425,62 +533,165 @@ impl FileBasedKeystore {
...
+        let file = FileBasedKeystoreFile {
+            version: 2,
+            keys: self
+                .keys
+                .iter()
+                .map(|(address, key)| AliasedKey {
+                    alias: self
+                        .aliases
+                        .get(address)
+                        .map_or_else(|| self.create_alias(None).unwrap(), |a| a.alias.clone()),
```
thibault-martinez, line 675: Is it not possible to not have an alias?

alexsporn: No, if you don't provide one, a random one is always generated. The only edge-case where no alias was supported previously was when you had a legacy .keystore file and no .aliases file, so aliases were created for those when you loaded up the keystore.

thibault-martinez (approved):

miker83z (approved):

PR #7746, opened 2025-07-08
feat(gRPC): Add subscription to events including filtering

### LINKED ISSUE #7726: gRPC: Add subscription to events including filtering
We want to be able to subscribe to different events as it was done in the websocket API.

bingyanglin:
# Description of change

Add event filters for client subscription

fixes #7726

### Release Notes

- [x] Nodes (Validators and Full nodes): Add subscription to events including filtering in gRPC


crates/iota/src/iota_commands.rs
```diff
@@ -772,6 +776,21 @@ async fn start(
                     .dir(config_path.clone())
                     .with_iota_names_config(iota_names_config);
             }
+
+            // Apply gRPC configuration if enabled
+            if enable_grpc_api {
+                if let Some(grpc_config) = grpc_api_config {
+                    info!(
+                        "Enabling gRPC API for fullnode with config: {:?}",
+                        grpc_config
```
Thoralf-M:
```suggestion
                        "Enabling gRPC API for fullnode with config: {grpc_config:?}"
```

crates/iota-core/src/subscription_handler.rs
```diff
@@ -104,10 +107,44 @@ impl SubscriptionHandler {
...
+                    }
+                } else {
+                    debug!(
+                        event_index = index,
+                        tx_digest =? effects.transaction_digest(),
+                        event_type = %event.type_.name,
+                        "No gRPC subscribers, skipping broadcast"
+                    );
+                }
+            }
```
muXxer: I wonder if we just simply can reuse the "event_streamer" for gRPC as well. Isn't exactly doing what we need?
We could also just subscribe to that if someone subscribes to the gRPC event stuff. Maybe we should get the latest fixes for the affected files in "indexer_api.rs" and then move the functionality to a more common place? Maybe to the "subscription_handler.rs"?

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+service EventService {
+  rpc StreamEvents (EventStreamRequest) returns (stream Event);
+}
+
+message EventStreamRequest {
+  EventFilter filter = 1;
+}
+
+message Event {
+  bytes event_data = 1;
```
muXxer: Why isn't that the former created `BcsData` type?

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+  EventID event_id = 2;
+  optional uint64 timestamp_ms = 3;
+}
+
+message EventID {
+  uint64 tx_seq = 1;
+  uint64 event_seq = 2;
+  string tx_digest = 3;
```
muXxer: arent tx_digests always 32 bytes? No need for base64 strings here.

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Filter by sender address
+message SenderFilter {
+  string sender = 1;           // Sender address (hex string)
+}
+
+// Filter by transaction digest
+message TransactionFilter {
+  string tx_digest = 1;        // Transaction digest (base64 string)
```
muXxer: arent tx_digests always 32 bytes? No need for base64 strings here.

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Match all events (catch-all filter)
+message AllFilter {
+  // Empty - matches all events
+}
+
+// Filter by sender address
+message SenderFilter {
+  string sender = 1;           // Sender address (hex string)
```
muXxer: Not 100% sure, but aren't sender always addresses, and 32 byte long?
Then there would be no need for a string.

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+    SenderFilter sender = 8;
+    TransactionFilter transaction = 9;
+    MoveModuleFilter move_module = 10;
+    TimeRangeFilter time_range = 11;
+  }
+}
+
+// Filter by Move event type (package::module::event_name)
+message MoveEventTypeFilter {
+  string address = 1;      // Package ID (hex string)
```
muXxer: we should create an Address type here, and this should also always be 32 bytes.

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+
+// Filter by Move event field value
+message MoveEventFieldFilter {
+  string path = 1;         // Field path (e.g., "anchor")
+  string value = 2;        // Value to match (as JSON string)
+}
+
+// Filter by package ID
+message PackageFilter {
+  string package_id = 1;   // Package ID (hex string)
```
muXxer: Address type, 32 bytes

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Filter by package and module
+message MoveEventModuleFilter {
+  string package_id = 1;   // Package ID (hex string)
```
muXxer: Address type, 32 bytes

crates/iota-grpc-api/proto/event.proto
muXxer: We can reuse the address type here.

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Filter by transaction digest
+message TransactionFilter {
+  string tx_digest = 1;        // Transaction digest (base64 string)
+}
+
+// Filter by transaction execution module (different from event definition module)
+message MoveModuleFilter {
+  string package_id = 1;       // Package ID (hex string)
```
muXxer: Address type, 32 bytes

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,104 @@
...
+}
+
+message EventStreamRequest {
+  EventFilter filter = 1;
+}
+
+message Event {
+  bytes event_data = 1;
+  EventID event_id = 2;
+  optional uint64 timestamp_ms = 3;
```
semenov-vladyslav:
```suggestion
  EventID event_id = 1;
  optional uint64 timestamp_ms = 2;
  bytes event_data = 3;
```

crates/iota-grpc-api/src/bcs_event.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Convert BCS bytes back to IotaEvent
+pub fn try_from_bcs_bytes(bytes: &[u8]) -> Result<IotaEvent, bcs::Error> {
+    let bcs_event: BcsIotaEvent = bcs::from_bytes(bytes)?;
+    Ok(IotaEvent::from(bcs_event))
+}
+
+/// Internal BCS-compatible version of IotaEvent
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BcsIotaEvent {
```
semenov-vladyslav:
```suggestion
struct BcsIotaEvent {
```

crates/iota-grpc-api/src/bcs_event.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl From<BcsIotaEvent> for IotaEvent {
+    fn from(bcs_event: BcsIotaEvent) -> Self {
+        Self {
+            id: bcs_event.id,
+            package_id: bcs_event.package_id,
+            transaction_module: bcs_event.transaction_module,
+            sender: bcs_event.sender,
+            type_: bcs_event.type_,
+            parsed_json: serde_json::from_str(&bcs_event.parsed_json).unwrap_or_default(),
```
semenov-vladyslav: BCS is supposed to be canonical. Not sure this invariant is preserved here.

crates/iota-grpc-api/src/bcs_event.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! BCS serialization utilities for IotaEvent.
+//!
+//! IotaEvent cannot be directly BCS serialized because it uses
+//! serde_json::Value, and annotations like #[serde(flatten)].
+
+use iota_json_rpc_types::IotaEvent;
```
semenov-vladyslav: Should this whole `mod` be moved to `iota_json_rpc_types`?

crates/iota-grpc-api/proto/event.proto
muXxer: maybe you can define another file with common types, and move the `BcsData` there?

crates/iota-grpc-api/src/bcs_event.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn try_to_bcs_bytes(event: &IotaEvent) -> Result<Vec<u8>, bcs::Error> {
+    let bcs_event = BcsIotaEvent::from(event.clone());
+    bcs::to_bytes(&bcs_event)
+}
+
+/// Convert BCS bytes back to IotaEvent
+pub fn try_from_bcs_bytes(bytes: &[u8]) -> Result<IotaEvent, bcs::Error> {
+    let bcs_event: BcsIotaEvent = bcs::from_bytes(bytes)?;
+    Ok(IotaEvent::from(bcs_event))
+}
```
muXxer: Do we actually need that if we reuse the `BcsData`?

crates/iota/src/iota_commands.rs
```diff
@@ -772,6 +776,18 @@ async fn start(
...
                     .with_iota_names_config(iota_names_config);
             }
+
+            // Apply gRPC configuration if enabled
+            if enable_grpc_api {
+                if let Some(grpc_config) = grpc_api_config {
+                    info!("Enabling gRPC API for fullnode with config: {grpc_config:?}");
+                    swarm_builder = swarm_builder.with_fullnode_grpc_api_config(grpc_config);
+                } else {
+                    info!("gRPC API enabled but no grpc-api-config provided, using default");
```
muXxer:
```suggestion
                    warn!("gRPC API enabled but no grpc-api-config provided, using default");
```

muXxer (changes requested):

crates/iota-grpc-api/src/types.rs
```diff
@@ -30,6 +31,11 @@ pub trait CheckpointDataBroadcaster {
     fn send(&self, data: &CheckpointData) -> anyhow::Result<()>;
 }
 
+/// Trait for broadcasting events
+pub trait EventBroadcaster {
```
semenov-vladyslav: Should these Broadcaster traits be unified into the following?

```rs
pub trait Broadcaster<T> {
    fn send(&self, data: &T) -> anyhow::Result<()>;
}
```

or

```rs
pub trait Broadcaster {
    type Data;
    fn send(&self, data: &Self::Data) -> anyhow::Result<()>;
}
```

crates/iota-grpc-api/src/event_service.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Build AND filter by chaining EventFilter::And
+fn build_and_filter(filters: Vec<EventFilter>) -> Result<EventFilter, Status> {
+    build_chained_filter(filters, "AND", EventFilter::And)
+}
+
+/// Build OR filter by chaining EventFilter::Or
+fn build_or_filter(filters: Vec<EventFilter>) -> Result<EventFilter, Status> {
+    build_chained_filter(filters, "OR", EventFilter::Or)
+}
```
semenov-vladyslav: What "other logic"?
```suggestion
```

crates/iota-grpc-api/src/event_service.rs
```diff
@@ -0,0 +1,231 @@
...
+            value: serde_json::Value::String(f.value.clone()),
+        }),
+        Some(Filter::Package(f)) => Ok(EventFilter::Package(parse_object_id(&f.package_id)?)),
+        Some(Filter::MoveEventModule(f)) => Ok(EventFilter::MoveEventModule {
+            package: parse_object_id(&f.package_id)?,
+            module: parse_identifier(&f.module, "module name")?,
+        }),
+        Some(Filter::And(f)) => {
+            let filters = parse_filter_list(&f.filters)?;
+            build_and_filter(filters)
```
semenov-vladyslav:
```suggestion
            Ok(EventFilter::All(filters))
```

crates/iota-grpc-api/src/event_service.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+        Some(Filter::Or(f)) => {
+            let filters = parse_filter_list(&f.filters)?;
+            build_or_filter(filters)
```
semenov-vladyslav:
```suggestion
            Ok(EventFilter::Any(filters))
```

crates/iota-grpc-api/src/types.rs
semenov-vladyslav: These traits do not seem to be used at all. What is their purpose? To have `impl Broadcaster for ()` for testing which is not used?

crates/iota-grpc-api/src/event_service.rs
```diff
@@ -0,0 +1,231 @@
...
+use tracing::debug;
+
+use crate::{
+    bcs_event::try_to_bcs_bytes,
+    events::{Event, EventId, EventStreamRequest, event_service_server::EventService},
+    types::GrpcEventBroadcaster,
+};
+
+pub struct EventGrpcService {
+    pub event_broadcaster: GrpcEventBroadcaster,
```
semenov-vladyslav: Should it be like this:

```suggestion
pub struct EventGrpcService<T: EventBroadcaster> {
    pub event_broadcaster: T,
```

so that `EventBroadcaster` trait is justified?

crates/iota-grpc-api/src/bcs_event.rs
```diff
@@ -0,0 +1,70 @@
...
+    base_types::{IotaAddress, ObjectID},
+    event::EventID,
+};
+use move_core_types::{identifier::Identifier, language_storage::StructTag};
+use serde::{Deserialize, Serialize};
+
+/// Convert IotaEvent to BCS bytes
+pub fn try_to_bcs_bytes(event: &IotaEvent) -> Result<Vec<u8>, bcs::Error> {
+    let bcs_event = BcsIotaEvent::from(event.clone());
+    bcs::to_bytes(&bcs_event)
```
semenov-vladyslav:
```suggestion
    bcs::to_bytes(&bcs_event).unwrap()
```
It seems that `bcs::to_bytes` fails in really exceptional cases.

crates/iota-grpc-api/src/event_service.rs
```diff
@@ -0,0 +1,231 @@
...
+    build_chained_filter(filters, "AND", EventFilter::And)
+}
+
+/// Build OR filter by chaining EventFilter::Or
+fn build_or_filter(filters: Vec<EventFilter>) -> Result<EventFilter, Status> {
+    build_chained_filter(filters, "OR", EventFilter::Or)
+}
+
+// Convert IotaEvent to protobuf Event
+impl TryFrom<&IotaEvent> for Event {
```
semenov-vladyslav: Should it be unfallible?
```suggestion
impl From<&IotaEvent> for Event {
```

crates/iota-grpc-api/src/server.rs
```diff
@@ -54,16 +58,24 @@ impl GrpcServerHandle {
...
-/// services. Currently includes the checkpoint streaming service, but can be
-/// extended to host additional services in the future.
+/// and event streaming services. Currently includes the checkpoint streaming
+/// and event streaming services, but can be extended to host additional
+/// services in the future.
 pub async fn start_grpc_server(
     grpc_reader: Arc<GrpcReader>,
+    grpc_event_broadcaster: GrpcEventBroadcaster,
     config: crate::Config,
+    shutdown_token: CancellationToken,
```
semenov-vladyslav, line 71: It looks like `CancellationToken` should not be a part of `GrpcReader`.

crates/iota-node/src/lib.rs
```diff
@@ -2306,8 +2360,28 @@ async fn build_grpc_server(
...
+    let shutdown_token = CancellationToken::new();
+
+    // Create GrpcReader with the server-level shutdown token
+    let grpc_reader = Arc::new(GrpcReader::from_rest_state_reader(
+        rest_read_store,
+        shutdown_token.clone(),
+    ));
+
+    let grpc_event_broadcaster =
+        grpc_event_broadcaster.expect("gRPC event broadcaster should exist when gRPC is enabled");
```
semenov-vladyslav:
```suggestion
    let Some(grpc_event_broadcaster) = grpc_event_broadcaster else {
        return Err(anyhow!("gRPC event broadcaster should exist when gRPC is enabled"));
    };
```

crates/iota-node/src/lib.rs
```diff
@@ -1752,8 +1773,24 @@ impl IotaNode {
                 guard.as_ref().map(|handle| {
                     let tx = handle.checkpoint_summary_broadcaster().clone();
                     Box::new(move |summary: &CertifiedCheckpointSummary| {
-                        if let Err(e) = tx.send(summary) {
-                            warn!("Failed to send checkpoint summary: {e}");
+                        if tx.receiver_count() > 0 {
+                            match tx.send(summary) {
```
semenov-vladyslav: A receiver might have unsubscribed in between these two lines. Why not let `tx.send` handle `receiver_count` itself?

crates/iota-core/src/subscription_handler.rs
```diff
@@ -104,10 +107,44 @@ impl SubscriptionHandler {
...
-        for event in events.data.clone() {
-            if let Err(e) = self.event_streamer.try_send(event) {
+        for (index, event) in events.data.clone().iter().enumerate() {
+            // Send to existing event streamer
+            if let Err(e) = self.event_streamer.try_send(event.clone()) {
                 error!(error =? e, "Failed to send event to dispatch");
             }
+
+            // Also send to gRPC broadcast channel if available
+            if let Some(ref grpc_event_broadcaster) = self.grpc_event_broadcaster {
```
semenov-vladyslav: Should `self.grpc_event_broadcaster` use `self.event_streamer`? `self.event_streamer` seems to be designed to handle exactly this case: it only sends events to those subscribers with matching filters. `EventGrpcService` would just duplicate filtering logic.

semenov-vladyslav (changes requested):

crates/iota-grpc-api/src/bcs_event.rs
bingyanglin: Previously I defined a simplified Event proto where we need those functions. Now I removed all the `bcs_event` module by just including all IotaEvent fileds in our protobuf fields in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out. I refactored the `package_id` to be the newly defined Address type which is 32 bytes in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out. I refactored the `package_id` to be the newly defined Address type which is 32 bytes in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out. I refactored the `package_id` to be the newly defined Address type which is 32 bytes in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out. I added the 32-byte Address type in the common.proto in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/proto/event.proto
bingyanglin: Yes as the `Event` defined in `crates/iota-types/src/event.rs`, the sender is 32-byte long and we can reuse the address type. Fixed it at 8dc4a7a61d433c863eaa023bdf32d1329fd25262

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out, I created a `TransactionDigest` type to be 32 bytes in the common.proto in 8dc4a7a61d433c863eaa023bdf32d1329fd25262

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out, I created a TransactionDigest type to be 32 bytes in the common.proto in https://github.com/iotaledger/iota/commit/8dc4a7a61d433c863eaa023bdf32d1329fd25262.

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for pointing this out, I created a `common.proto` to put all the shared types in https://github.com/iotaledger/iota/commit/8dc4a7a61d433c863eaa023bdf32d1329fd25262.

crates/iota-grpc-api/src/event_service.rs
bingyanglin: Like XOR, NOT, NAND, etc.? I intended to do it in this way so it might be easier to extend in the future.

crates/iota-grpc-api/src/bcs_event.rs
bingyanglin: No it's not canonical. I refactored the Event protobuf to make it contain all the fields of IotaEvent, so as to remove the previous `bcs_event` module in https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262. In this way we don't have this issue anymore.

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks for the refactory, I put the event_data in the last field of the refactored Event protobuf in https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262.

crates/iota-grpc-api/src/bcs_event.rs
bingyanglin: Thanks for pointing this out. I removed the all bcs_event module because now our Event protobuf contains all the fields of IotaEvent in 8dc4a7a61d433c863eaa023bdf32d1329fd25262.

crates/iota-grpc-api/src/bcs_event.rs
bingyanglin: Thanks for pointing this out. After refactory I removed the whole bcs_event module in 8dc4a7a61d433c863eaa023bdf32d1329fd25262

crates/iota-grpc-api/src/event_service.rs
bingyanglin: But Maybe we don't need that anyway. Let's remove them as you suggested, thanks!

crates/iota-grpc-api/src/event_service.rs
bingyanglin: Agreed. I also refactored all the Event protobuf and also make it unfallible in 8dc4a7a61d433c863eaa023bdf32d1329fd25262

crates/iota-core/src/subscription_handler.rs
bingyanglin: Agreed. I did the refactory accordingly in [f06a81d](https://github.com/iotaledger/iota/pull/7746/commits/f06a81d38dd7d978820da6fa3957d8f1c019f214). Thanks for this valuable comment!

crates/iota-grpc-api/src/bcs_event.rs
bingyanglin: Thanks. I removed all the bcs_event module after refactory in [8dc4a7a](https://github.com/iotaledger/iota/pull/7746/commits/8dc4a7a61d433c863eaa023bdf32d1329fd25262).

crates/iota-grpc-api/src/types.rs
bingyanglin: The EventBroadcaster is used because we used the send before. Now after refactory by reusing the event_streamer in [f06a81d](https://github.com/iotaledger/iota/pull/7746/commits/f06a81d38dd7d978820da6fa3957d8f1c019f214) we just need subscriber.
For the newly introduce `impl EventSubscriber for () {` it is needed because it's for unit testing, like in `checkpoint_stream.rs` we need a dummy subscriber as the followings, 
```rust
    // Use the no-op EventSubscriber implementation for unit type
    let dummy_event_subscriber = Arc::new(()) as Arc<dyn EventSubscriber>;
```    
otherwise we need to define a dummy subscriber structure and implement the trait which might not be necessary.

crates/iota-core/src/subscription_handler.rs
bingyanglin: Yes thanks for the valuable comment. I refactored it by reusing the event_streamer in [f06a81d](https://github.com/iotaledger/iota/pull/7746/commits/f06a81d38dd7d978820da6fa3957d8f1c019f214).

crates/iota-node/src/lib.rs
bingyanglin: Thanks for pointing this out. After refactory in [f06a81d](https://github.com/iotaledger/iota/pull/7746/commits/f06a81d38dd7d978820da6fa3957d8f1c019f214) we don't need the broadcaster anymore.

crates/iota-node/src/lib.rs
bingyanglin: Because in this way (our previous approach) we will get lots of warn-level logs in the Error branch when we have no subscribers yet, which we observed by using iota CLI testing. And in the Error branch I think it's better to have warn level log instead of debug level logs, so I refactored it to be the current way. Also the racing condition might be a rare case? I'm not 100% sure about this.

crates/iota-grpc-api/src/server.rs
bingyanglin: Previously we don't have this shutdown_token inside GrpcReader because we created it inside it. The `GrpcReader` needs this `shutdown_token` to break the loop inside `create_checkpoint_stream`. But I guess the best approach might be pass the `shutdown_token` from the gprc server level so we can reuse the same token for any other new services in the future. I refactored it a bit so now we hold the `shutdown_token` in `CheckpointGrpcService`, and then pass it into the `GrpcReader` in [ba81b1b](https://github.com/iotaledger/iota/pull/7746/commits/ba81b1baada955c51e9105a1d239e91078fb8662).

crates/iota-grpc-api/src/event_service.rs
bingyanglin: Thanks for the comment. We removed this one and only have subscriber now in [f06a81d](https://github.com/iotaledger/iota/pull/7746/commits/f06a81d38dd7d978820da6fa3957d8f1c019f214).

crates/iota-grpc-api/src/server.rs
semenov-vladyslav: Right, having a single cancellation token for the whole grpc server makes more sense to me. If we wanted a more granular control for checkpoint and event services we could get child tokens from it, but it's too complex for now.

crates/iota-node/src/lib.rs
semenov-vladyslav: I implied that the design is not great:

1. it exposes and makes use of internals (`receiver_count`),
2. `send` is a method of `CheckpointSummaryBroadcaster` trait while `receiver_count` is a method of `GrpcCheckpointSummaryBroadcaster` type.

Arguably, a better approach would be to move sending and tracing stuff into a separate method (eg. `send_traced` or `send` itself) so that the call site is not bloated with not-so-relevant boilerplate code.

bingyanglin: Agreed, makes sense, I created a separate method `send_tranced` for the broadcasters, and in the Error handling I also used two different log levels so it should be clean and no race-condition now in [8c01a95](https://github.com/iotaledger/iota/pull/7746/commits/8c01a9528b118fbbe2dccc4ad666380b6da48341). I preserve the `receiver_count()` method because I use it several times in the `send_traced`. Also in case we might need it in other places in the future.

crates/iota-core/src/subscription_handler.rs
```diff
@@ -104,8 +105,9 @@ impl SubscriptionHandler {
         }
 
         // serially dispatch event processing to honor events' orders.
-        for event in events.data.clone() {
-            if let Err(e) = self.event_streamer.try_send(event) {
+        for event in events.data.clone().iter() {
+            // Send to unified event streamer (serves both JSON-RPC and gRPC subscribers)
+            if let Err(e) = self.event_streamer.try_send(event.clone()) {
```
piotrm50: Why is the event cloned now if no new usage is added below?

piotrm50 (commented):
Only looked at `iota-core` changes.

crates/iota-core/src/subscription_handler.rs
bingyanglin: Good catch, revert the changes in [216d7ef](https://github.com/iotaledger/iota/pull/7746/commits/216d7ef47639963a50a82ca84ef9765c22fbe391). Thanks for pointing this out.

piotrm50 (approved):

jkrvivian (approved):

crates/iota-grpc-api/proto/event.proto
```diff
@@ -0,0 +1,85 @@
...
+
+// Filter by transaction execution module (different from event definition module)
+message MoveModuleFilter {
+  iota.grpc.common.Address package_id = 1;       // Package ID
+  string module = 2;                             // Module name
+}
+
+// Filter by Move event type (package::module::event_name)
+message MoveEventTypeFilter {
+  iota.grpc.common.Address address = 1;      // Package ID
```
semenov-vladyslav:
```suggestion
  iota.grpc.common.Address package_id = 1;      // Package ID
```

crates/iota-grpc-api/src/types.rs
```diff
@@ -46,6 +56,34 @@ impl GrpcCheckpointSummaryBroadcaster {
...
+        match self.send(summary) {
+            Ok(()) => {
+                debug!(
+                    "Sent checkpoint summary #{} to {} gRPC subscriber(s)",
+                    *summary.data().sequence_number(),
+                    self.receiver_count()
+                );
+            }
+            Err(e) => {
+                if self.receiver_count() > 0 {
```
semenov-vladyslav: [Docs](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/struct.SendError.html) say that error can only be in one case -- no receivers, so only "debug" branch is needed.

crates/iota-grpc-api/src/types.rs
```diff
@@ -71,6 +109,34 @@ impl GrpcCheckpointDataBroadcaster {
...
+        match self.send(data) {
+            Ok(()) => {
+                debug!(
+                    "Sent checkpoint data #{} to {} gRPC subscriber(s)",
+                    data.checkpoint_summary.data().sequence_number,
+                    self.receiver_count()
+                );
+            }
+            Err(e) => {
+                if self.receiver_count() > 0 {
```
semenov-vladyslav: Same here.

semenov-vladyslav (changes requested):
Looks good except for a couple minor nits.

crates/iota-grpc-api/proto/event.proto
bingyanglin: Thanks. Modified all the related code in [8121115](https://github.com/iotaledger/iota/pull/7746/commits/8121115825d2d318bfda5eb77ebc67986a63581b).

crates/iota-grpc-api/src/types.rs
bingyanglin: Thanks, modified in [f7dcd36](https://github.com/iotaledger/iota/pull/7746/commits/f7dcd362c3bcedd0c9e744ab588de2a03780f488).

crates/iota-grpc-api/src/types.rs
bingyanglin: Thanks, modified in [f7dcd36](https://github.com/iotaledger/iota/pull/7746/commits/f7dcd362c3bcedd0c9e744ab588de2a03780f488).

semenov-vladyslav (approved):

PR #7751, opened 2025-07-08
fix(apps-core): Fix fallback icon image size and its refreshing

marc2332:


evavirseda (changes requested):
we need to keep large size :pray: 
image

evavirseda (changes requested):
this need to be reviewed, because it's not right after the change
image

msarcev (approved):

evavirseda (approved):

PR #7768, opened 2025-07-09
feat(node): Prefix fallible storage methods with `try_`

muXxer:
# Description of change

We want to add non-fallible functions to the storage traits, but therefore we need to clean up the existing interfaces first.
This PR is prefixing all fallible function in the stores and caches with `try_`.
This PR is renaming functions only, there are no changes of the logic. The fallible functions will be added in a follow-up PR.

### DIFF
--- a/crates/iota-analytics-indexer/src/handlers/transaction_handler.rs
+++ b/crates/iota-analytics-indexer/src/handlers/transaction_handler.rs
@@ -222,9 +222,9 @@ mod tests {
 
         // Create a checkpoint which should include the transaction we executed.
         let checkpoint = sim.create_checkpoint();
-        let checkpoint_data = sim.get_checkpoint_data(
+        let checkpoint_data = sim.try_get_checkpoint_data(
             checkpoint.clone(),
-            sim.get_checkpoint_contents_by_digest(&checkpoint.content_digest)?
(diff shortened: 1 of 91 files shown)


jkrvivian (approved):

TheMrAI (approved):
If you have a plan for this, then I am fine with it.
But I hope this won't be the beginning after which we want to prefix every fallible function with `try_*`.

DaughterOfMars (commented):
I do not agree with this. For most cases it is simply adding verbosity without meaningfully differentiating between anything. It makes sense in cases where you have two similar methods and one may return a Result, but not for any case where a method may return a Result.

muXxer:
@DaughterOfMars there will be both versions of those functions, the fallible ones and the non-fallible ones. I just wanted to split this in two PRs, so that this PR here doesn't change any logic, but only does renaming. The second PR will introduce a macro that automatically implements the non-fallible version for every method that is prefixed with "try_" and returns a result.
We wanted to port upstream changes in the first place, and everything there was made non-fallible, but other teams didn't agree with this breaking trait definition change, so we needed to have both versions now, fallible and non-fallible.

piotrm50 (approved):

DaughterOfMars (approved):

kodemartin (approved):

PR #7786, opened 2025-07-11
feat(starfish): implement and benchmark filter for additional pushed headers

### LINKED ISSUE #7523: feat(starfish) Benchmark Incoming Bytes Filter in Cordial Dissemination
Implement the incoming bytes filter and perform a benchmark with and without the incoming bytes filter in cordial dissemination to estimate the potential benefit of using a filter. We can test different distributions of incoming messages - from all unique messages to receiving the same message and see what's the overhead of introducing the filter and if it can bring performance boost.

VorobyevIlya:
# Description of change

This PR introduces a filter for headers. 
In the Starfish protocol, when validator v1 sends its block to validator v2, it also sends additional headers that validator v2 doesn't know (up to the best knowledge of v1). It can lead to a situation where the same header is received multiple times. We don't want to spend resources on processing the header multiple times, so we introduce a filter to detect duplicates as early as possible.

PR also introduces tests for the function handling received bundles (blocks with additional headers). Also, some changes in test_dag_builder were introduced for convenient testing.

Performance with and without a filter was measured for the following scenarios. 
1. Each validator sends its block for round r with all block headers from the previous round. It means that all blocks are accepted, and all additional headers should be filtered if the filter is on and processed multiple times if it is off.
2. Each validator sends only its block without any additional headers.

| Rounds | Validators | Filter | Time 1 | Time 2 |
|--------|------------|--------|--------|--------|
| 100    | 50         | false  | 157.6    | 11.9    |
| 100    | 50         | true   | 40.3   | 12.1    |
| 100    | 100        | false  | 1022   | 77.4   |
| 100    | 100        | true   | 297.9  | 77.7   |

#7523 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/starfish/core/src/authority_service.rs
```diff
@@ -37,6 +41,39 @@ use crate::{
...
+    fn new() -> Self {
+        Self {
+            header_digests: DashSet::new(),
+            queue: Mutex::new(VecDeque::new()),
+        }
+    }
+
+    async fn add_batch(&self, digests: Vec<BlockHeaderDigest>) {
+        for digest in digests.iter() {
+            self.header_digests.insert(digest.clone());
```
polinikita: As a suggestion, if `insert` says that such a digest already in the dashset, then we can remove this block header from the batch that is going to be send to the core thread. This will make a second check whether a block header with a given digest is already going to be processed by another thread

crates/starfish/core/src/authority_service.rs
```diff
@@ -276,6 +321,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         &self,
         peer: AuthorityIndex,
         serialized_block_bundle: SerializedBlockBundle,
+        filter: bool,
```
polinikita: I would remove this bool from this argument. The benefit of having the filter is very clear and I don't think that it will be a specific protocol parameter.

crates/starfish/core/src/authority_service.rs
```diff
@@ -1300,4 +1377,306 @@ mod tests {
...
+                        context.committee.to_authority_index(peer).unwrap(),
+                        serialized_block_bundle,
+                        filter,
+                    )
+                    .await
+                    .expect("bundle is expected to be processed successfully");
+                total_duration += start.elapsed();
+            }
+        }
+        println!("Total time: {:?}", total_duration);
```
polinikita: I don't think we need to track timing in all these places, and I would keep this time tracking private. Also, I would try to check that the filter is working as intended

polinikita (changes requested):
Overall looks good. I give a few suggestions to change (bool variable, remove potential concurrent block headers from sending to the core)

crates/starfish/core/src/authority_service.rs

crates/starfish/core/src/authority_service.rs
VorobyevIlya: removed timing tracking from here, added some asserts

crates/starfish/core/src/authority_service.rs
VorobyevIlya: removed

crates/starfish/core/src/authority_service.rs
```diff
@@ -37,6 +41,48 @@ use crate::{
...
+
+    #[cfg(test)]
+    fn size(&self) -> usize {
+        self.header_digests.len()
+    }
+
+    async fn add_batch(&self, digests: Vec<BlockHeaderDigest>) -> Vec<BlockHeaderDigest> {
+        let mut already_inserted = vec![];
+        for digest in digests.iter() {
+            if !self.header_digests.insert(*digest) {
```
piotrm50, line 67: Is using `DashSet` safe here if we are calling `insert` and `remove` while it's unlocked inbetween?

crates/starfish/core/src/authority_service.rs
```diff
@@ -37,6 +41,48 @@ use crate::{
...
+                already_inserted.push(*digest);
+            }
+        }
+        let mut queue = self.queue.lock().await;
+        for digest in digests {
```
piotrm50, line 72: Why do need another loop here and another `while` loop after that? Can we compress all this logic in a single loop?

crates/starfish/core/src/authority_service.rs
```diff
@@ -145,7 +197,9 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
 
         let verified_block_header = VerifiedBlockHeader::new_verified(
             signed_block_header,
-            serialized_block_and_transactions.serialized_block_header,
+            serialized_block_and_transactions
+                .serialized_block_header
+                .clone(),
```
piotrm50: Maybe we can avoid `.clone()` here by destructuring `serialized_block_and_transactions`?

crates/starfish/core/src/authority_service.rs
```diff
@@ -522,31 +589,46 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        additional_block_headers.retain(|block_header| {
+            if index < digests_to_exclude.len()
+                && block_header.digest() == digests_to_exclude[index]
+            {
+                index += 1;
+                false
+            } else {
+                true
+            }
+        });
```
piotrm50: I think this logic should not be split into `filter out` above and `add values to a filter`, because if two concurrent tasks are handling blocks, then it's possible that duplicated blocks pass the filter in concurrent tasks and only later they are added to the filter by those two tasks. 

I think when we are filtering out blocks we should also add new blocks to the filter.

crates/starfish/core/src/metrics.rs
```diff
@@ -130,10 +130,10 @@ pub(crate) struct NodeMetrics {
...
     pub(crate) synchronizer_fetched_blocks_by_authority: IntCounterVec,
     pub(crate) invalid_blocks: IntCounterVec,
-    pub(crate) invalid_headers_in_a_bundle: IntCounterVec,
-    pub(crate) valid_headers_in_a_bundle: IntCounterVec,
-    #[allow(dead_code)]
-    pub(crate) received_unique_headers_from_a_bundle: IntCounterVec,
+    pub(crate) invalid_headers_in_bundles: IntCounterVec,
+    pub(crate) valid_headers_in_bundles: IntCounterVec,
+    pub(crate) filtered_headers_in_bundles: IntCounterVec,
+    pub(crate) received_unique_headers_from_bundles: IntCounterVec,
```
piotrm50, line 136: Please describe better if the `authority` label refers to block author or the peer from which we received the block, as is done in other metrics (e.g. 

> pub(crate) block_manager_missing_blocks: IntGauge,
> pub(crate) block_manager_missing_blocks_by_authority: IntCounterVec,

piotrm50 (changes requested):

crates/starfish/core/src/authority_service.rs
VorobyevIlya: yes, inside the data is stored in different buckets, and there are internal locks for safe access

crates/starfish/core/src/authority_service.rs
VorobyevIlya: we can do it in a single loop, but is it better? I did it in two loops to minimize the time when queue is locked

crates/starfish/core/src/authority_service.rs
VorobyevIlya: yes, good idea. I modified the code

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Do you mean that we should add headers to the filter before checks? I think that we can't do it before the check when we verify that the round of received header is smaller than the block's round. And we do it after deserialization. So to move insertion into filter to filtering we need to remove this check. We can do it, but I'm not sure that we should.
For now, I separated the insertion of digests of additional headers from insertion of digest of the main block and moved header's insertion as early as possible.

crates/starfish/core/src/metrics.rs
VorobyevIlya: done

muXxer:
I think this PR needs to be rebased. I see way too many changes from former PRs, which makes a review impossible.

polinikita (approved):

crates/starfish/core/src/authority_service.rs
```diff
@@ -522,31 +579,52 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        for block_header in additional_block_headers.iter() {
+            digests_to_add_to_filter.push(block_header.digest())
+        }
+        digests_to_add_to_filter.push(verified_block.digest());
+        let digests_to_exclude = self
+            .received_block_headers
+            .add_batch(digests_to_add_to_filter)
+            .await;
+        let mut index = 0;
+        additional_block_headers.retain(|block_header| {
```
piotrm50, line 597: Does this work though? I guess this only works if it's guaranteed that the relative ordering in both vectors is the same. I think it would be safer to implement it like:

```
additional_block_headers.retain(|block_header| {
    !digests_to_exclude.iter().any(|digest| block_header.digest() == *digest)
});
```

crates/starfish/core/src/authority_service.rs
piotrm50: I see that here we filter some things out again.

piotrm50 (changes requested):

piotrm50 (approved):

PR #7804, opened 2025-07-14
feat(iota-framework): Allow unlocking of expired `TimelockedStakedIota` without unstaking

### LINKED ISSUE #6823: Allow unlocking of expired `TimelockedStakedIota` without unstaking
## Description

`TimelockedStakedIota` can currently only be handled in the wallet-dashboard even if they already vested.
To allow the dashboard to convert them to normal staking objects without requiring an unstake/restake operation I propose to add these 2 functions to `TimelockedStakedIota`:

```move
/// The lock has not expired yet.
const ENotExpiredYet: u64 = 3;

/// Function to unlock the `StakedIota` from a `TimelockedStakedIota`  based on the epoch start time.
public fun unlock(self: TimelockedStakedIota, ctx: &TxContext): StakedIota {
    // Unpack the timelock. 
    let (staked, expiration_timestamp_ms, _) = unpack(self);

    // Check if the lock has expired.
    assert!(expiration_timestamp_ms <= ctx.epoch_timestamp_ms(), ENotExpiredYet);

    staked
}

/// Function to unlock the `StakedIota` from a `TimelockedStakedIota` based on the `Clock` object.
public fun unlock_with_clock(self: TimelockedStakedIota, clock: &Clock): StakedIota {
    // Unpack the timelock. 
    let (staked, expiration_timestamp_ms, _) = unpack(self);

    // Check if the lock has expired.
    assert!(expiration_timestamp_ms <= clock.timestamp_ms(), ENotExpiredYet);

    staked
}
```

Note: both these functions are adapted from the ones found in `TimeLock`

## Motivation

This allows to convert already vested and staked timelocks into normal stakes without unstaking.

No.

valeriyr:
# Description of change

`TimelockedStakedIota` can currently only be handled in the wallet-dashboard even if they already vested.
To allow the dashboard to convert them to normal staking objects without requiring an unstake/restake operation two related functions were added.

fixes #6823


crates/iota-framework/packages/iota-system/sources/timelocked_staking.move
```diff
@@ -13,8 +14,10 @@ use std::string::String;
 
 /// For when trying to stake an expired time-locked balance.
 const ETimeLockShouldNotBeExpired: u64 = 0;
-/// Incompatible objects when joining TimelockedStakedIota
+/// Incompatible objects when joining `TimelockedStakedIota`.
 const EIncompatibleTimelockedStakedIota: u64 = 1;
+/// For when trying to unlock a not-expired `TimelockedStakedIota`.
+const ETimelockedStakedIotaShouldBeExpired: u64 = 2;
```
Thoralf-M: Can we add new errors as vector or is there something that speaks against it?
```move
    #[error]
    const ETimelockedStakedIotaIsNotExpired: vector<u8> = b"TimelockedStakedIota is not expired.";
```

muXxer (approved):
Files owned by us look good.

crates/iota-framework/packages/iota-system/sources/timelocked_staking.move
valeriyr: Should be good, I don't know any reason why not to use it. Updated here: 401b5e6cbf

crates/iota-framework/packages/iota-system/tests/timelocked_delegation_tests.move
```diff
@@ -638,6 +639,270 @@ fun test_add_remove_stake_mul_bal_flow() {
...
+    // Stake 2 IOTA
+    stake_timelocked_with(STAKER_ADDR_1, VALIDATOR_ADDR_1, 2, 10, scenario);
+
+    // Increment epoch timestamp.
+    scenario.ctx().increment_epoch_timestamp(10);
+
+    scenario.next_tx(STAKER_ADDR_1);
+    {
+        let timelocked_staked_iota = scenario.take_from_sender<TimelockedStakedIota>();
+        // The `timelocked_staked_iota` is expired so it sfould be unlocked.
```
TheMrAI: Minor typo? should?

crates/iota-framework/packages/iota-system/sources/timelocked_staking.move
```diff
@@ -207,6 +211,30 @@ public fun request_withdraw_stake_non_entry(
...
+    staked
+}
+
+/// Unlock the `StakedIota` from a `TimelockedStakedIota` based on the `Clock` object.
+public fun unlock_with_clock(self: TimelockedStakedIota, clock: &Clock): StakedIota {
+    // Unpack the `TimelockedStakedIota`.
+    let (staked, expiration_timestamp_ms, _) = unpack(self);
+
+    // Check if the lock has expired.
+    assert!(expiration_timestamp_ms <= clock.timestamp_ms(), ETimelockedStakedIotaShouldBeExpired);
```
TheMrAI, line 233: Just wondering, don't we have to check here that the received Clock object is the system Clock?
Or it is impossible to have another Clock?

crates/iota-framework/packages/iota-system/tests/timelocked_delegation_tests.move
```diff
@@ -638,6 +639,270 @@ fun test_add_remove_stake_mul_bal_flow() {
...
+    let scenario = &mut scenario_val;
+    // Stake 2 IOTA
+    stake_timelocked_with(STAKER_ADDR_1, VALIDATOR_ADDR_1, 2, 10, scenario);
+
+    scenario.next_tx(STAKER_ADDR_1);
+    {
+        let timelocked_staked_iota = scenario.take_from_sender<TimelockedStakedIota>();
+        // The `timelocked_staked_iota` is still not-expired so this should fail.
+        let staked_iota = timelocked_staked_iota.unlock(scenario.ctx());
+        test_utils::destroy(staked_iota);
```
TheMrAI, line 710: Just for my learning experience.
`unlock` here asserts, thus exiting immediately right? We need to destroy the objects so that the compiler is happy not because we have to clean up?

TheMrAI (approved):
LGTM

crates/iota-framework/packages/iota-system/tests/timelocked_delegation_tests.move
valeriyr: fixed in 301dd66115

crates/iota-framework/packages/iota-system/sources/timelocked_staking.move
valeriyr: No, it is impossible. We use the `iota::clock::Clock` type. There is only one its instance and it can be created only by a system transaction.

crates/iota-framework/packages/iota-system/tests/timelocked_delegation_tests.move
valeriyr: Yes, you are right. The assert happens at runtime, the compiler can't predict this.

PR #7769, opened 2025-07-09
feat: add non-fallible storage methods via macros

muXxer:
# Description of change

This PR adds non-fallible storage methods via newly added proc-macros.

```rust
#[generate_non_fallible_fn("Error message")]
fn try_get_value(key: &str) -> Result<String, Error> {
    // implementation
}

// The macro generates:
fn get_value(key: &str) -> String {
    try_get_value(key).expect("Error message")
}

#[extend_trait_with_non_fallible("Error message")]
pub trait MyService {
    fn try_get_data(&self) -> Result<String, Error>;
    
    fn try_get_many(&self, ids: &[u32]) -> Result<Vec<String>, Error> {
        // default implementation
    }
    
    // The macro will add:
    fn get_data(&self) -> String;

    fn get_many(&self, ids: &[u32]) -> Vec<String> {
        self.try_get_many(ids).expect("Error message")
    }
}

#[extend_impl_with_non_fallible("Error message")]
impl MyService for MyServiceImpl {
    fn try_get_data(&self) -> Result<String, Error> {
        // implementation
    }
    
    // The macro will add:
    fn get_data(&self) -> String {
        self.try_get_data().expect("Error message")
    }
}
```

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #7817, opened 2025-07-15
feat(core): add non-fallible storage methods

muXxer:
# Description of change

This PR adds non-fallible storage methods. It's an alternative solution to #7769 without macros 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1158,7 +1158,7 @@ impl AuthorityState {
     pub async fn try_execute_immediately(
         &self,
         certificate: &VerifiedExecutableTransaction,
-        mut expected_effects_digest: Option<TransactionEffectsDigest>,
+        expected_effects_digest: Option<TransactionEffectsDigest>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<(TransactionEffects, Option<ExecutionError>)> {
         let _scope = monitored_scope("Execution::try_execute_immediately");
@@ -1175,10 +1175,10 @@ impl AuthorityState {
             .get_transaction_cache_reader()
             .try_get_executed_effects(tx_digest)?
         {
-            if let Some(expected_effects_digest) = expected_effects_digest {
+            if let Some(expected_effects_digest_inner) = expected_effects_digest {
                 assert_eq!(
                     effects.digest(),
-                    expected_effects_digest,
+                    expected_effects_digest_inner,
                     "Unexpected effects digest for transaction {:?}",
                     tx_digest
                 );
@@ -1189,13 +1189,13 @@ impl AuthorityState {
         let input_objects =
             self.read_objects_for_execution(tx_guard.as_lock_guard(), certificate, epoch_store)?;
 
-        if expected_effects_digest.is_none() {
-            // We could be re-executing a previously executed but uncommitted transaction,
-            // perhaps after restarting with a new binary. In this situation, if
-            // we have published an effects signature, we must be sure not to
-            // equivocate. TODO: read from cache instead of DB
-            expected_effects_digest = epoch_store.get_signed_effects_digest(tx_digest)?;
-        }
+        // If no expected_effects_digest was provided, try to get it from storage.
+        // We could be re-executing a previously executed but uncommitted transaction,
+        // perhaps after restarting with a new binary. In this situation, if
+        // we have published an effects signature, we must be sure not to
+        // equivocate. TODO: read from cache instead of DB
+        let expected_effects_digest =
+            expected_effects_digest.or(epoch_store.get_signed_effects_digest(tx_digest)?);
 
         self.process_certificate(
             tx_guard,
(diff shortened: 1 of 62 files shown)


kodemartin (approved):

crates/iota-core/src/authority.rs
```diff
@@ -2220,10 +2231,14 @@ impl AuthorityState {
         Ok(epoch_store.reference_gas_price())
     }
 
-    pub fn is_tx_already_executed(&self, digest: &TransactionDigest) -> IotaResult<bool> {
+    pub fn try_is_tx_already_executed(&self, digest: &TransactionDigest) -> IotaResult<bool> {
         self.get_transaction_cache_reader()
             .try_is_tx_already_executed(digest)
     }
+    pub fn is_tx_already_executed(&self, digest: &TransactionDigest) -> bool {
```
jkrvivian: Add a blank line here 😁 
```suggestion
    }
    
    pub fn is_tx_already_executed(&self, digest: &TransactionDigest) -> bool {
```

crates/iota-core/src/authority.rs
```diff
@@ -3346,15 +3361,20 @@ impl AuthorityState {
     }
 
     #[instrument(level = "trace", skip_all)]
-    pub async fn get_object(&self, object_id: &ObjectID) -> IotaResult<Option<Object>> {
+    pub async fn try_get_object(&self, object_id: &ObjectID) -> IotaResult<Option<Object>> {
         self.get_object_store()
             .try_get_object(object_id)
             .map_err(Into::into)
     }
+    pub async fn get_object(&self, object_id: &ObjectID) -> Option<Object> {
```
jkrvivian: Also blank line
```suggestion
    }
    
    pub async fn get_object(&self, object_id: &ObjectID) -> Option<Object> {
```

jkrvivian (changes requested):

crates/iota-core/src/authority.rs
```diff
@@ -4334,17 +4354,27 @@ impl AuthorityState {
...
 
-    pub async fn get_object_or_tombstone(
+    pub async fn try_get_object_or_tombstone(
         &self,
         object_id: ObjectID,
     ) -> IotaResult<Option<ObjectRef>> {
         self.get_object_cache_reader()
             .try_get_latest_object_ref_or_tombstone(object_id)
     }
+    pub async fn get_object_or_tombstone(&self, object_id: ObjectID) -> Option<ObjectRef> {
```
jkrvivian: Add a blank line
```suggestion
    }
    
    pub async fn get_object_or_tombstone(&self, object_id: ObjectID) -> Option<ObjectRef> {
```

jkrvivian (changes requested):

jkrvivian (approved):

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -1141,9 +1141,7 @@ impl WritebackCache {
         // pending_consensus_transactions until after the transaction has executed.
         let Some((_, outputs)) = self.dirty.pending_transaction_writes.remove(tx) else {
             assert!(
-                !self
-                    .try_is_tx_already_executed(tx)
-                    .expect("read cannot fail"),
+                !self.try_is_tx_already_executed(tx)?,
```
valeriyr, line 1144: Just in case, would we like to return an error here instead the expect call?

crates/iota-core/src/authority.rs
```diff
@@ -4787,8 +4824,7 @@ impl AuthorityState {
         // reconfiguration anyway.
         if self
             .get_transaction_cache_reader()
-            .try_is_tx_already_executed(tx_digest)
-            .expect("read cannot fail")
+            .try_is_tx_already_executed(tx_digest)?
```
valeriyr, line 4827: and here?

roman1e2f5p8s (approved):

crates/iota-single-node-benchmark/src/mock_storage.rs
```diff
@@ -145,7 +145,7 @@ impl ChildObjectResolver for InMemoryObjectStore {
         child: &ObjectID,
         child_version_upper_bound: SequenceNumber,
     ) -> IotaResult<Option<Object>> {
-        Ok(self.try_get_object(child).unwrap().and_then(|o| {
+        Ok(self.try_get_object(child)?.and_then(|o| {
```
valeriyr, line 148: and here?

crates/iota-core/src/authority.rs
```diff
@@ -3346,15 +3362,22 @@ impl AuthorityState {
...
         self.get_object_store()
             .try_get_object(object_id)
             .map_err(Into::into)
     }
 
+    /// Non-fallible version of `try_get_object`.
+    pub async fn get_object(&self, object_id: &ObjectID) -> Option<Object> {
+        self.try_get_object(object_id)
+            .await
+            .expect("storage access failed")
```
DaughterOfMars, line 3375: it might actually be better to just unwrap here since the error message may be more precise

valeriyr (approved):
LGTM!

crates/iota-core/src/execution_cache/writeback_cache.rs
muXxer: We should go through the whole codebase and stop propagating errors, if we panic anyways at the end. This function here is only used by the top level function "reconfigure" where we have an "expect" in place. So for this PR it's not a problem, but you are right, maybe we should change that logic everywhere.

crates/iota-core/src/authority.rs
muXxer: This is used in `make_checkpoint`, where there is a retry every second in case it fails, instead of panicking.

crates/iota-single-node-benchmark/src/mock_storage.rs
muXxer: This is only used by the `fetch_child_object_unbounded` macro in `iota-execution`, and handled as a storage error anyways, so I guess propagating is better than panicking.
```rust
            .map_err(|msg| {
                PartialVMError::new(StatusCode::STORAGE_ERROR).with_message(format!("{msg}"))
            })?;
```

PR #7825, opened 2025-07-16
feat(indexer): Fetch optimistic events in ReadApi::get_events

### LINKED ISSUE #7794: [indexer]: refactor ReadApi::get_events to fetch optimistic events

tomxey:
# Description of change

Fetch optimistic events in ReadApi::get_events

Note that optimistically indexed event doesn't have **timestamp_ms** assigned yet, so from the API perspective event will have **timestamp_ms=None** at the beginning, and then value will be set to **Some(..)** after event is checkpoint indexed.

fixes #7794

`cargo test --profile dev-nodebug --package iota-indexer --test rpc-tests --all-features`

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/indexer_reader.rs
+++ b/crates/iota-indexer/src/indexer_reader.rs
@@ -1265,7 +1265,9 @@ impl IndexerReader {
         let iota_tx_events =
             tx_events_to_iota_tx_events(tx_events, self.package_resolver(), digest, timestamp_ms)
                 .await?;
-        Ok(iota_tx_events.map_or(vec![], |ste| ste.data))
+        Ok(iota_tx_events.map_or(vec![], |transaction_block_events| {
+            transaction_block_events.data
+        }))
     }
 
     pub async fn try_get_checkpointed_transaction_events(
--- a/crates/iota-indexer/src/indexer_reader.rs
+++ b/crates/iota-indexer/src/indexer_reader.rs
@@ -1252,12 +1252,12 @@ impl IndexerReader {
     ) -> Result<Vec<iota_json_rpc_types::IotaEvent>, IndexerError> {
         let checkpointed_events = self.try_get_checkpointed_transaction_events(digest).await?;
 
-        let (timestamp_ms, serialized_events) = match checkpointed_events {
-            Some((timestamp_ms, serialized_events)) => {
-                (Some(timestamp_ms as u64), serialized_events)
-            }
-            None => (None, self.get_optimistic_transaction_events(digest).await?),
-        };
+        let (timestamp_ms, serialized_events) =
+            if let Some((timestamp, events)) = checkpointed_events {
+                (Some(timestamp as u64), events)
+            } else {
+                (None, self.get_optimistic_transaction_events(digest).await?)
+            };
 
         let events = stored_events_to_events(serialized_events)?;
         let tx_events = TransactionEvents { data: events };
--- a/crates/iota-indexer/src/indexer_reader.rs
+++ b/crates/iota-indexer/src/indexer_reader.rs
@@ -1250,8 +1250,30 @@ impl IndexerReader {
         &self,
         digest: TransactionDigest,
     ) -> Result<Vec<iota_json_rpc_types::IotaEvent>, IndexerError> {
+        let checkpointed_events = self.try_get_checkpointed_transaction_events(digest).await?;
+
+        let (timestamp_ms, serialized_events) = match checkpointed_events {
+            Some((timestamp_ms, serialized_events)) => {
+                (Some(timestamp_ms as u64), serialized_events)
+            }
+            None => (None, self.get_optimistic_transaction_events(digest).await?),
+        };
+
+        let events = stored_events_to_events(serialized_events)?;
+        let tx_events = TransactionEvents { data: events };
+
+        let iota_tx_events =
+            tx_events_to_iota_tx_events(tx_events, self.package_resolver(), digest, timestamp_ms)
+                .await?;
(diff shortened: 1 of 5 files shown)


crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -1339,8 +1339,30 @@ impl IndexerReader {
...
+            None => (None, self.get_optimistic_transaction_events(digest).await?),
+        };
+
+        let events = stored_events_to_events(serialized_events)?;
+        let tx_events = TransactionEvents { data: events };
+
+        let iota_tx_events =
+            tx_events_to_iota_tx_events(tx_events, self.package_resolver(), digest, timestamp_ms)
+                .await?;
+        Ok(iota_tx_events.map_or(vec![], |ste| ste.data))
```
sergiupopescu199: Nit: Does `ste` stand for `StoredTransactionEvents`? If so, maybe we should rename it for clarity, especially since `tx_events_to_iota_tx_events` returns an `IotaTransactionBlockEvents`.

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -1339,8 +1339,30 @@ impl IndexerReader {
...
+    }
+
+    pub async fn try_get_checkpointed_transaction_events(
```
sergiupopescu199, line 1273: Mostly curious: why do we emphasize `checkpointed` in `try_get_checkpointed_transaction_events`? Since the indexer gets data from checkpoints, isn’t that already implicit?

sergiupopescu199 (approved):
lgtm!

crates/iota-indexer/src/indexer_reader.rs
tomxey: I would say that the new default would be to get data from all sources, like`get_transaction_events_in_blocking_task` does.
If we are limiting ourselves to only one data source then I think it's good to add the `checkpointed` part, to at least stand in contrast with `get_optimistic_transaction_events`.

crates/iota-indexer/src/indexer_reader.rs
tomxey: Yeah, the old name doesn't seem to make sense, will refactor.

crates/iota-indexer/src/indexer_reader.rs
sergiupopescu199: What I was referring to is that `try_get_checkpointed_transaction_events` retrieves data from the `transactions` table, which, by default, is populated through the ingestion pipeline as new checkpoints arrive.

On the other hand, `get_optimistic_transaction_events` is clear in its intent—it emphasizes that it fetches optimistic transaction events from a separate table.

My main concern is why we specifically highlight "checkpointed" for the standard table. Since this is our primary, committed data source, it might be more intuitive to treat it as the default and only call out other sources (like optimistic) when necessary.

> I would say that the new default would be to get data from all sources  

Could you clarify what you mean by "all sources"? Which sources are you referring to?

tomxey: By all sources I meant optimistic and standard tables.

If I would remove `checkpointed` from the function name we would have:
 - try_get_transaction_events - that reads events from standard tables
 - get_optimistic_transaction_events - that reads events from optimistic tables
 - get_transaction_events_in_blocking_task - that reads events from both standard and optimistic table

Especially there is no semantic difference in `try_get_transaction_events` and `get_transaction_events_in_blocking_task`, and I wanted to have some distinction.

sergiupopescu199: Thank you for the explanation. Resolving the conversation!

crates/iota-indexer/tests/rpc-tests/read_api.rs
```diff
@@ -904,6 +908,81 @@ fn get_events() {
     });
 }
 
+#[test]
+fn get_fresh_event() -> Result<(), anyhow::Error> {
```
samuel-rufi: nit: maybe a more clear test name?

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -1250,8 +1250,32 @@ impl IndexerReader {
         &self,
         digest: TransactionDigest,
     ) -> Result<Vec<iota_json_rpc_types::IotaEvent>, IndexerError> {
+        let checkpointed_events = self.try_get_checkpointed_transaction_events(digest).await?;
+
+        let (timestamp_ms, serialized_events) = match checkpointed_events {
```
samuel-rufi: nit: maybe this match can be simplified to something like this:

```
    let (timestamp_ms, serialized_events) = if let Some((timestamp, events)) = checkpointed_events {
        (Some(timestamp as u64), events)
    } else {
        (None, self.get_optimistic_transaction_events(digest).await?)
    };
```

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -939,6 +940,40 @@ async fn wrap_basic_object(
     Ok((res, wrapped_obj_id))
 }
 
+pub(crate) async fn update_basic_object(
```
samuel-rufi: nit: I personally think we can avoid this helper function if not required to share with other tests. To me it seems they additionally blow up the file a bit, and it lacks some context. I think these two calls can be indeed part of the test.

samuel-rufi (approved):
Looks good! @tomxey 🌷 
Just some nits.

PR #7828, opened 2025-07-16
feat(evm-bridge): add native token evm bridge tests

### LINKED ISSUE #7708: Add native-token e2e tests to evm-bridge
After https://github.com/iotaledger/iota/pull/7567
evm-bridge supports sending different tokens.

We need to cover these cases with e2e tests

brancoder:
# Description of change

adds tests for native tokens 
resolves #7708 


msarcev:
Would prefer to have new files with the added tests, instead of making the current two specs longer with `serial`.
`serial` makes the tests take longer (no parallel execution), and if one "step" fails in the chain we don't get to see the results of the rest.

brancoder:
> Would prefer to have new files with the added tests, instead of making the current two specs longer with `serial`. `serial` makes the tests take longer (no parallel execution), and if one "step" fails in the chain we don't get to see the results of the rest.

The tests are set up this way to keep wallet creations and faucet requests to a minimum, so we don't hit the faucet limit.

panteleymonchuk (dismissed):

VmMad (approved):

panteleymonchuk (approved):

msarcev (approved):

PR #7830, opened 2025-07-16
feat(ci): upload split cluster logs as artifacts on failure

muXxer:
# Description of change

We need a way to check why the split cluster tests fail, so the logs files of the nodes will be uploaded as artifacts on failure.

### DIFF
--- a/.github/workflows/_split_cluster.yml
+++ b/.github/workflows/_split_cluster.yml
@@ -1,6 +1,7 @@
 name: Split Cluster Check
 
 on:
+  workflow_dispatch:
   workflow_call:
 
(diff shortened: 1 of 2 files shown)


alexsporn (approved):

TheMrAI (approved):

Dr-Electron (approved):
Tested with

```
act workflow_dispatch -P self-hosted-x64=catthehacker/ubuntu:full-latest -W .github/workflows/_split_cluster.yml --artifact-server-path $PWD/.artifacts
```

and commented out `if: failure()` line. Docs are "uploaded" correctly

PR #7845, opened 2025-07-17
chore(CLI): Add leaf subname limitation info

### LINKED ISSUE #446: Add warning about not being able to track leaf nodes
In:
- CLI logs
- CLI command docs
- SC function docs?

Alex6323:
# Description of change

This PR makes it transparent to the CLI user, that leaf subnames cannot be tracked in the CLI and must be done separatedly.

Closes https://github.com/iotaledger/iota-names/issues/446

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota/src/name_commands.rs
```diff
@@ -65,7 +65,7 @@ pub enum NameCommand {
     #[command(subcommand)]
     Auction(AuctionCommand),
     /// Check the availability of a name and return its price if available.
-    /// Subnames are always free to register by the parent name owner.
+    /// Subnames are always free to register by the parent name owner
```
thibault-martinez: Why are you removing this dot but not the previous one?

crates/iota/src/name_commands.rs
```diff
@@ -1473,7 +1474,7 @@ impl std::fmt::Display for NameCommandResult {
             Self::List(nfts) => {
                 let mut table_builder = TableBuilder::default();
 
-                table_builder.set_header(["id", "name", "expiration"]);
+                table_builder.set_header(["id", "name or node subname", "expiration"]);
```
thibault-martinez: I don't think this is necessary

crates/iota/src/name_commands.rs
Alex6323: Because I thought we were doing it like this ... only the last sentence is without punctuation :thinking:

crates/iota/src/name_commands.rs
Alex6323: okay :+1:

crates/iota/src/name_commands.rs
thibault-martinez: Is it what we are actually doing?

Alex6323: I checked the other commands. We are a bit inconsistent. Sometimes we always use proper punctuation, sometimes we don't, See `IotaClientCommands` for example.

crates/iota/src/name_commands.rs
Alex6323: https://github.com/iotaledger/iota/pull/7845/commits/85010fd7449c40ff2e0e3128b643023325a2e038

crates/iota/src/name_commands.rs
Alex6323: https://github.com/iotaledger/iota/pull/7845/commits/85010fd7449c40ff2e0e3128b643023325a2e038

Thoralf-M (approved):

crates/iota/src/name_commands.rs
```diff
@@ -1023,8 +1024,8 @@ impl AuctionCommand {
 #[derive(Parser)]
 #[command(rename_all = "kebab-case")]
 pub enum SubnameCommand {
-    /// Register a new leaf subname, which can only be managed by the parent's
-    /// NFT
+    /// Register a new leaf subname, which is managed by its parent, i.e. a name
+    /// or a node subname NFT. Leaf subnames are not tracked by the CLI
```
DaughterOfMars: This feels a little odd, both the location and the phrasing. I'd maybe put it on the `list` command since that's the place where the confusion will be?

Alex6323: I agree. Want me to keep it as before?

DaughterOfMars: I feel like it's not clear enough why they wont appear. Maybe a simple explanation about the lack of nft?

thibault-martinez: Also I think `Leaf subnames are not tracked by the CLI` is a bit confusing as it may give the impression than even resolving wouldn't work. The only thing that doesn't work is listing.

crates/iota/src/name_commands.rs
```diff
@@ -85,7 +85,8 @@ pub enum NameCommand {
         /// records will be returned.
         key: Option<String>,
     },
-    /// List the names owned by the given address, or the active address
+    /// List the names and node subnames owned by the given address, or the
+    /// active address. Note, that leaf subnames are not listed by the CLI.
```
thibault-martinez:
```suggestion
    /// active address. Note that leaf subnames are not listed by the CLI.
```

crates/iota/src/name_commands.rs
```diff
@@ -1023,8 +1024,9 @@ impl AuctionCommand {
 #[derive(Parser)]
 #[command(rename_all = "kebab-case")]
 pub enum SubnameCommand {
-    /// Register a new leaf subname, which can only be managed by the parent's
-    /// NFT
+    /// Register a new leaf subname, which will NOT create an NFT, but instead,
+    /// is managed by its parent NFT. Therefore, it won't be listed by the
+    /// CLI.
```
thibault-martinez: So many commas 😆

thibault-martinez (approved):

crates/iota/src/name_commands.rs
Alex6323: okay, asked chattie about it and will push its suggestion :trollface:

crates/iota/src/name_commands.rs
Alex6323: https://github.com/iotaledger/iota/pull/7845/commits/6eca638e311452aea6dd685fbebf7fabc01c760d

crates/iota/src/name_commands.rs
```diff
@@ -1028,8 +1029,8 @@ impl AuctionCommand {
 #[derive(Parser)]
 #[command(rename_all = "kebab-case")]
 pub enum SubnameCommand {
-    /// Register a new leaf subname, which can only be managed by the parent's
-    /// NFT
+    /// Register a new leaf subname, which will NOT create an NFT but instead
+    /// is managed by its parent NFT. Therefore it won't be listed by the CLI.
```
DaughterOfMars: I tried to explain a bit more in depth without being overly verbose. I think clap will only display the first line by default, and additional lines with a special flag, but I'm not sure.

```suggestion
    /// Register a new leaf subname.
    /// Leaf subnames have no associated NFT; their ownership is determined
    /// by the ownership of the parent's NFT. Unfortunately, this makes
    /// it difficult to find leafs owned by a particular address. For this
    /// reason, leaf subnames are not included in the `list` command.
```

DaughterOfMars: And if we like this then I propose we add it to the `list` command too, slightly reworded

Alex6323: I would prefer your text to be part of a separate CLI documention, that goes more into the "why" question. In order to be as concise as possible, I personally would prefer for the CLI to restrict itself to answer the "what" question. Of course only as a rule of thumb, here we're kinda forced to hint at why it can't be listed.

Alex6323: @thibault-martinez @Thoralf-M what do you think?

Alex6323: @thibault-martinez @Thoralf-M what do you think?

Thoralf-M: > to be part of a separate CLI documention

Where would this be? If it's not here, then it will probably become outdated quickly and also won't help people that only use --help

Alex6323: In `cli.mdx` there has to be a reference to a new `./cli/iota_names.mdx` document, where this can be explained in detail. I personally don't think that phrases like "Unfortunately..." belong in the CLI, this belongs to an actual document. But that's just my opinion!! 

Why would it get outdated quickly? Same could be said for the other CLI documents then. No?

Thoralf-M: Yes and the issue is that they are already outdated right now :cry: Because it's not part of the code and we don't update the docs for every little change we do here

Alex6323: At some point the CLI will stabilize, and docs won't be outdated that quickly.

thibault-martinez: I agree with Alex that here I would just explain the "what" and not the "why. Remember that we will have proper IOTA Names documentation and we can expand the "why" there.

DaughterOfMars: Okay, but the what is not useful here because it doesn't make sense without the why

Alex6323: I am still not convinced that the CLI is actually the right place for the user to understand WHY a command works in a specific way. Disadvantages I see are:
* overly verbose CLI docs leading to a messy terminal view
* leaking implementation details
* quicker desynchronization between code and docs

thibault-martinez: > Okay, but the what is not useful here because it doesn't make sense without the why

I completely disagree, we can definitely have the what without the why, and I think we should

Alex6323: after some discussion we decided to just state the facts: https://github.com/iotaledger/iota/pull/7845/commits/ae91c7642e0983a2c0143208e22cf4135e6c3143

crates/iota/src/name_commands.rs
```diff
@@ -90,7 +90,9 @@ pub enum NameCommand {
         /// records will be returned.
         key: Option<String>,
     },
-    /// List the names owned by the given address, or the active address
+    /// List the names and subnames owned by the given address, or the
+    /// active address. Note that leaf subnames are not listed by the `list`
+    /// command.
```
thibault-martinez:
```suggestion
    /// List the names and subnames owned by the given address, or the
    /// active address. Note that leaf subnames are not listed.
```
I think it's a bit weird that the command mentions itself

crates/iota/src/name_commands.rs
```diff
@@ -1130,8 +1131,9 @@ impl AuctionCommand {
 #[derive(Parser)]
 #[command(rename_all = "kebab-case")]
 pub enum SubnameCommand {
-    /// Register a new leaf subname, which can only be managed by the parent's
-    /// NFT
+    /// Register a new leaf subname, which will NOT create an NFT but instead
+    /// is managed by its parent NFT. Note that leave subnames are not listed by
```
thibault-martinez:
```suggestion
    /// is managed by its parent NFT. Note that leaf subnames are not listed by
```

crates/iota/src/name_commands.rs
```diff
@@ -90,8 +90,9 @@ pub enum NameCommand {
         /// records will be returned.
         key: Option<String>,
     },
-    /// List the names and node subnames owned by the given address, or the
-    /// active address. Note that leaf subnames are not listed by the CLI.
+    /// List the names and subnames owned by the given address, or the
+    /// active address. Note that leaf subnames are not listed by the `list`
+    /// command.
```
DaughterOfMars:
```suggestion
    /// active address. Note that leaf subnames are not listed by this command.
```

crates/iota/src/name_commands.rs
```diff
@@ -1030,7 +1031,8 @@ impl AuctionCommand {
 #[command(rename_all = "kebab-case")]
 pub enum SubnameCommand {
     /// Register a new leaf subname, which will NOT create an NFT but instead
-    /// is managed by its parent NFT. Therefore it won't be listed by the CLI.
+    /// is managed by its parent NFT. Note that leave subnames are not listed by
```
DaughterOfMars: I wouldn't even mention it here

crates/iota/src/name_commands.rs
Alex6323: oh no. what did I do here?

crates/iota/src/name_commands.rs
Alex6323: hmm, I actually would, because it might make the user reconsider their decision.

DaughterOfMars: I think it's completely unrelated to this command and is confusing

crates/iota/src/name_commands.rs
Alex6323: that already has been addressed by https://github.com/iotaledger/iota/pull/7845/commits/6539a5ffa1c14c0c05e779fadaaa4fa48e58d71c

crates/iota/src/name_commands.rs
Alex6323: sorry, but I disagree ... it's making transparent to the user that there's a catch with using leaf subnames. The whole point of this PR is warning the user about them UPFRONT.
@Thoralf-M @thibault-martinez opinions?

thibault-martinez: It is completely related and we also have a log about it in the output of this command. It's only consistent that we put it in the docs so user are aware beforehand.

DaughterOfMars: I think the point is not to warn the user but to inform them of a limitation in one particular command. The decision to use leafs or nodes should be based on reading the docs, not the register command help text

thibault-martinez: I don't disagree with this but I don't see the harm in having it. If they haven't yet read the docs at least they are aware 🤷🏻

thibault-martinez: I don't disagree with this but I don't see the harm in having it. If they haven't yet read the docs at least they are aware 🤷🏻

Alex6323: the flaw in your argument is the SHOULD :trollface: @DaughterOfMars

Alex6323: ~Alright, if we're all happy, then I remove it. I don't actually mind that hard.~ Keeping it.

thibault-martinez: I won't push too hard for it so feel free to merge, but it doesn't make any sense to me. We can avoid inevitable confusion with a simple mention, zero reason not to do it.

PR #7848, opened 2025-07-17
feat(iota-indexer): Add `tx_wrapped_or_deleted_objects` backfill job

samuel-rufi:
# Description of change

Adds an ingestion backfill job to backfill the new Indexer `tx_wrapped_or_deleted_objects` table.
Inspired by following upstream patch: https://github.com/MystenLabs/sui/commit/b902e8213d5ff7f62d2ad6baf4091527f21d227d

## Links to any relevant issues

Fixes https://github.com/iotaledger/iota/issues/7631

## How the change has been tested

- [X] Basic tests (linting, compilation, formatting, unit/integration tests)
- [X] Patch-specific tests (correctness, functionality coverage)

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

Tested backfilling of a local Indexer DB instance with an ingestion worker that gets data from the devnet.

- [ ] Synchronization of the indexer from genesis for a network including migration objects.
- [X] Restart of indexer synchronization locally without resetting the database.
- [ ] Restart of indexer synchronization on a production-like database.
- [ ] Deployment of services using Docker.
- [X] Verification of API backward compatibility.

### Release Notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

PR #7851, opened 2025-07-17
fix(iota-indexer): Fix consumption of `ready_checkpoints` for ingestion based backfill

### LINKED ISSUE #7849: Ingestion backfill task removes checkpoints beyond asssigned chunk range
## Bug description

When performing backfill tasks in chunks, the current implementation can inadvertently remove checkpoints beyond its assigned range from the `ready_checkpoints` buffer. This occurs if subsequent checkpoints are already processed and buffered, causing following chunks to stall because their data have already been consumed.

This issue appeared as part of https://github.com/iotaledger/iota/pull/7848

The problematic loop that possible can consumes data outside of the chunk backfill range:
https://github.com/iotaledger/iota/blob/develop/crates/iota-indexer/src/backfill/ingestion/task.rs#L89-L95

## Context

**Components** (select all that apply):

- [X] `iota-indexer`
- [ ] `iota-json-rpc`
- [ ] `iota-graphql-rpc`
- [ ] Other: __________

**Issue type** (select one):

- [ ] JSON-RPC API bug (failing request, unexpected response, etc.)
- [X] Crate bug (logic bug, runtime error, performance issue, etc.)

## Version & Environment

Specify exactly which version, commit, or branch you're using, and where you observed the issue:

- **Version (tag), commit hash, or branch**: https://github.com/iotaledger/iota/tree/feat/rpc-graphql-filter-wrapped-or-deleted-objects
- **Environment** (Mainnet, Testnet, Devnet, Local, other, etc.): Local

## Steps to reproduce the bug

Run the backfill task for example with:
```sh
cargo run --bin iota-indexer -- --db-url "postgres://postgres:postgrespw@localhost/iota_indexer_local" run-backfill 10 1000000 ingestion tx-wrapped-or-deleted-objects http://localhost:9000/api/v1
```

## Expected behaviour

The backfill job should terminate.

## Actual behaviour

The backfill job possibly doesn't terminate as the backfilling of chunks may stall.

samuel-rufi:
# Description of change

The current ingestion backfill implementation may remove checkpoints beyond its assigned range from the `ready_checkpoints`. This occurs if subsequent checkpoints are already processed and buffered by the data ingestion worker, causing following chunks to stall which wait on their data as the data has already been consumed.

Fixes https://github.com/iotaledger/iota/issues/7849

Manually tested with the backfill job, introduced in https://github.com/iotaledger/iota/pull/7848

This PR doesn’t touch existing ingestion pipelines or synchronization. It doesn’t include any actual backfill implementation, instead, it's only a backfill infrastructure component.

- [X] Restart of indexer synchronization locally without resetting the database.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-indexer/src/backfill/ingestion/task.rs
```diff
@@ -86,19 +86,28 @@ impl<T: IngestionBackfill> Backfill for IngestionBackfillTask<T> {
...
             {
                 processed_data.extend(processed);
                 start += 1;
-            }
-
-            if start <= end {
+            } else {
+                info!(
+                    "Waiting for processed data for checkpoint sequence number {}",
+                    start
```
sergiupopescu199:
```suggestion
                    "Waiting for processed data for checkpoint sequence number {start}"
```

sergiupopescu199 (approved):
lgtm!!

Really good catch, this issues was not obvious at all!

kodemartin (commented):
looks good @samuel-rufi! 

The only suggestions I have is to modify `ingestion::task::tests::ingestion_backfill_writes_to_db` so that it covers the issue solved with this patch.

kodemartin (approved):

PR #7863, opened 2025-07-17
PROTOTYPE: Vm lang/view function ability prototype

### LINKED ISSUE #7805: [move] Create a prototype for adding a view function ability
Based on Aptos implementation, but not limited to that (we could resort to a different kind of view function declaration or usage of the iota-adapter):
- [x] record the “view“ as function ability in Move
    - [x] must have a return type 
    - [x] additionally we could try to limit the fun parameters to all be read-only (as a second step in the prototype)
- [x] update the verifier
- [ ] (check why Aptos doesn’t allow to pass the signer as parameter)

Material
- [View functions | Aptos Move by Example](https://move-developers-dao.gitbook.io/aptos-move-by-example/advanced-concepts/view-functions)
- [aptos-core/aptos-move/framework/src/extended_checks.rs at 28b542fbfc3bd2a5a06a5d71ad66ccf5a4b837d5 · aptos-labs/aptos-core](https://github.com/aptos-labs/aptos-core/blob/28b542fbfc3bd2a5a06a5d71ad66ccf5a4b837d5/aptos-move/framework/src/extended_checks.rs#L680)
- [aptos-core/aptos-move/aptos-vm/src/verifier/view_function.rs at 28b542fbfc3bd2a5a06a5d71ad66ccf5a4b837d5 · aptos-labs/aptos-core](https://github.com/aptos-labs/aptos-core/blob/28b542fbfc3bd2a5a06a5d71ad66ccf5a4b837d5/aptos-move/aptos-vm/src/verifier/view_function.rs#L32)

TheMrAI:
PROTOTYPE! Not meant to be merged, just for observational/discussion purposes!

Try it out:
From the root of iota run:
```rust
cargo run -p iota -- move -p view_example build
```
Then in `view_example/sources/viewing.move`, uncomment failing code snippets.

Observe the comments in the commits themselves. Shortcuts have been taken, whenever possible.

TLDR:
- adding the "view" designator as a modifier/attribute is relative easy
- properly checking if the view function is a view function with an external Verifier looks dubious (but I may have missed something)
- have to check what aptos is doing, but at a cursory glance they had the view checks in a different stage


external-crates/move/crates/move-binary-format/src/file_format.rs
```diff
@@ -563,6 +563,7 @@ pub struct FunctionDefinition {
     pub visibility: Visibility,
     /// Marker if the function is intended as an entry function. That is
     pub is_entry: bool,
+    pub is_view: bool,
```
Dkwcs, line 566: I think minor comment would be good here, I mean in order to follow and maintain comment workflow.
And it'll help to explain what is view for others.

Dkwcs (commented):
LGTM, I run it and it works

miker83z:
Closed because it is a PoC. It will be used as reference software for some specifications.

PR #7864, opened 2025-07-18
fix(node): refactor ExecutionCacheConfig

### LINKED ISSUE #7847: Split `cache_type` from `ExecutionCacheConfig`
Users can't specify the settings for the writeback cache via config right now in case they have selected the passthrough cache. But they still can set them via env variables. This is an odd behaviour.
We should do something like this instead:

```rust
pub enum ExecutionCacheType {
    PassthroughCache,
    WritebackCache
}

pub struct ExecutionCacheConfig {
        cache_type: ExecutionCacheType,

        /// Maximum number of entries in each cache. (There are several
        /// different caches). If None, the default of 10000 is used.
        max_cache_size: Option<u64>,

        package_cache_size: Option<u64>, // defaults to 1000

        object_cache_size: Option<u64>, // defaults to max_cache_size
        marker_cache_size: Option<u64>, // defaults to object_cache_size
        object_by_id_cache_size: Option<u64>, // defaults to object_cache_size

        transaction_cache_size: Option<u64>, // defaults to max_cache_size
        executed_effect_cache_size: Option<u64>, // defaults to transaction_cache_size
        effect_cache_size: Option<u64>,      // defaults to executed_effect_cache_size

        events_cache_size: Option<u64>, // defaults to transaction_cache_size

        transaction_objects_cache_size: Option<u64>, // defaults to 1000
    },
}
```

semenov-vladyslav:
# Description of change

ProxyCache crates both Passthrough and Writeback caches and thus requires both configs, while previously ExecutionCacheConfig was an enum which could only provide config for one cache. This PR simplifies the definition of ExecutionCacheConfig allowing to keep all config values.

Fixes #7847.


crates/iota-config/src/node.rs
```diff
@@ -263,178 +262,124 @@ pub struct NodeConfig {
     pub iota_names_config: Option<IotaNamesConfig>,
 }
 
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum ExecutionCacheConfigType {
```
muXxer:
```suggestion
pub enum ExecutionCacheType {
```

muXxer (changes requested):

piotrm50 (approved):

muXxer (approved):

crates/iota-config/src/node.rs
semenov-vladyslav: Done

muXxer (approved):

PR #7868, opened 2025-07-18
feat(iota-synthetic-ingestion): Add `iota-synthetic-ingestion` application

### LINKED ISSUE #7793: feat: synthetic ingestion data
## Description

The goal of this task is to port [sui-synthetic-ingestion](https://github.com/MystenLabs/sui/blob/main/crates/sui-synthetic-ingestion/src/synthetic_ingestion.rs) application to the repository.

The application should be able to generate synthetic checkpoint data to run backfill and ingestion benchmarks with applications using `iota-data-ingestion-core`. One such example is `iota-indexer`.

samuel-rufi:
# Description of change

This PR introduces functionality to generate synthetic checkpoint data used for benchmarking and testing ingestion processes. It provides configurable parameters such as the number of checkpoints, checkpoint sizes, and the initial sequence number.

Inspired by https://github.com/MystenLabs/sui/blob/main/crates/sui-synthetic-ingestion/src/synthetic_ingestion.rs

Fixes https://github.com/iotaledger/iota/issues/7793

The change was tested by the provided unit tests.
Also generation of relevant `*.chk` files was tested using

`cargo run --bin iota-synthetic-ingestion --ingestion-dir synthetic-test --num-checkpoints 200 --checkpoint-size 10`

The patch doesn't touch existing ingestion logic or migration objects.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


sergiupopescu199 (changes requested):
Please also add the `iota-synthetic-ingestion` crate to `.github/crates-filters.yml` file. To ensure that CI will execute `crates-changes` & `external-changes` jobs on this crate aswell 🙏

sergiupopescu199 (approved):
lgtm 🌸

crates/iota-synthetic-ingestion/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+iota-types.workspace = true
+simulacrum.workspace = true
+telemetry-subscribers.workspace = true
+
+[dev-dependencies]
+# external dependencies
+tempfile.workspace = true
+
+# internal dependencies
+iota-storage.workspace = true
```
muXxer: does this even need to be a dev dependency if it's already a normal dependency?

crates/iota-types/src/mock_checkpoint_builder.rs
```diff
@@ -68,6 +68,24 @@ impl MockCheckpointBuilder {
...
+    /// Override the next checkpoint number to generate.
+    /// This can be useful to generate checkpoints with specific sequence
+    /// numbers.
+    pub fn override_next_checkpoint_number(
+        &mut self,
+        checkpoint_number: u64,
+        validator_keys: &impl ValidatorKeypairProvider,
+    ) {
+        if checkpoint_number > 0 {
+            let mut summary = self.previous_checkpoint.as_ref().unwrap().data().clone();
```
muXxer: What if the former `previous_checkpoint` was already None?
Maybe an expect would be better here if we can't avoid the panic (for tests only anyways).
Can we please still add a comment here.... it's a bit hard to understand why the function is called "overwrite next checkpoint number", but then you do some magic here with the previous checkpoint.

crates/iota-types/src/mock_checkpoint_builder.rs
```diff
@@ -143,13 +161,17 @@ impl MockCheckpointBuilder {
             epoch,
             sequence_number: self
                 .previous_checkpoint
-                .sequence_number
-                .checked_add(1)
-                .unwrap(),
-            network_total_transactions: self.previous_checkpoint.network_total_transactions
+                .as_ref()
+                .map(|c| c.sequence_number + 1)
+                .unwrap_or_default(),
```
muXxer: What is the default value for a `VerifiedCheckpoint`?

muXxer (changes requested):

crates/iota-synthetic-ingestion/Cargo.toml
samuel-rufi: Good catch, removed it: https://github.com/iotaledger/iota/pull/7868/commits/6d9ed150fcf700ef3df5b02074aac8180fefd95e

crates/iota-types/src/mock_checkpoint_builder.rs
samuel-rufi: If the previous checkpoint is absent, it should default the sequence number for the checkpoint to be built to `0`. To make this more clear I changed it to use `unwrap_or(0)`: https://github.com/iotaledger/iota/pull/7868/commits/acd6e93e738ae4fde633335055e049c87bfd4f68

crates/iota-types/src/mock_checkpoint_builder.rs
samuel-rufi: Good point, improved the docs for it and changed it to use `.expect()` (which should be fine as mentioned as it's for testing purposes): https://github.com/iotaledger/iota/pull/7868/commits/38720973d5d6b1db6d59aca4ad1b05b32c45fd56

muXxer (approved):

kodemartin (changes requested):
When testing with `iota-indexer`, the generated genesis checkpoint cannot be indexed 

```
transient worker execution error CheckpointProcessing("Failed to read or deserialize system state related data structures on-chain: IotaSystemStateWrapper object not found") for checkpoint 0
```
As a consequence no data is persisted to the database.

This needs to be fixed in order to use the application for its intended purposes.

samuel-rufi:
Hey @kodemartin, thanks for highlighting this issue - I fully agree that we need to address it.

Currently, the `sui-synthetic-ingestion` application doesn't support benchmarking scenarios involving genesis data, leading directly to the issues you've identified. Specifically, when `config.starting_checkpoint = 0` is specified, the generator removes the original genesis checkpoint (`0.chk`) produced by Simulacrum, generating a synthetic checkpoint 0 without genesis data (see: https://github.com/MystenLabs/sui/blob/main/crates/sui-synthetic-ingestion/src/synthetic_ingestion.rs#L49-L51).

The benchmark command in `sui-indexer-alt` runs benchmarks exclusively on synthetic workloads without processing/indexing genesis:
[https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt/src/args.rs#L78-L83](https://github.com/MystenLabs/sui/blob/ac1fa6c44480079e0a043431bbb0988c3ca93479/crates/sui-indexer-alt/src/args.rs#L78-L83)
https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt/src/benchmark.rs#L42-L47
https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt/src/benchmark.rs#L72.

Upon further review, I think it's preferable not to enforce exclusion of genesis data, at least not in the generator itself. Applications consuming the generated data should independently decide whether or not to include genesis in their benchmarks. This would allow not only targeted benchmarking but also general ingestion tests. It would also better support potential future scenarios involving custom genesis configurations.

Therefore, I propose preserving the original checkpoint 0 generated by Simulacrum. Initially, I aligned with the upstream design, but given these practical implications, it's now clear that retaining the original genesis checkpoint offers greater flexibility. I'll make this adjustment shortly.

samuel-rufi:
Turborepo Hierarchy / e2e / Localnet CI should be fixed by: https://github.com/iotaledger/iota/pull/7986

kodemartin (approved):

alexsporn (approved):
checked the Cargo and codeowners files. lgtm

crates/iota-types/src/mock_checkpoint_builder.rs
```diff
@@ -29,7 +29,7 @@ pub trait ValidatorKeypairProvider {
 /// checkpoint builder. It's mostly used by simulations, tests and benchmarks.
 #[derive(Debug)]
 pub struct MockCheckpointBuilder {
-    previous_checkpoint: VerifiedCheckpoint,
+    previous_checkpoint: Option<VerifiedCheckpoint>,
```
TheMrAI: This class now looks so odd.
We are sure no invariants are broken here?
For example you have to have a checkpoint to create a `MockCheckpointBuilder`, but then you can simply eliminate it using `override_next_checkpoint_number`, but then epoch value will be still set on it.

Similarly if you set a checkpoint back enough with `override_next_checkpoint_number` will mean now your checkpoint sequence is not strictly monotonic. Can that cause issue anywhere?

crates/iota-synthetic-ingestion/README.md
```diff
@@ -0,0 +1,34 @@
...
+Synthetic ingestion data generator for benchmarking and testing database ingestion performance.
+
+Provides functionality to generate synthetic checkpoint data consisting of transactions for benchmarking or testing purposes. It simulates transaction execution and produces serialized checkpoint files, which can later be loaded into memory or ingested by a database.
+
+## Usage
+
+Use the CLI to generate synthetic checkpoint data:
+
+```sh
+cargo run -- --ingestion-dir ./path/to/checkpoints \
```
TheMrAI: This is just a preference of mine, but if possible within a repository, should provide the appropriate build/testing commands from the root of the repository.
So that navigating within the folder tree isn't necessary.

crates/iota-synthetic-ingestion/src/synthetic_ingestion.rs
```diff
@@ -0,0 +1,210 @@
...
+        sim.create_checkpoint();
+        effects.created()[0].0
+    };
+
+    // When generating a workload that includes the genesis state, retain the
+    // initial checkpoints `0.chk` (genesis state) and `1.chk` (initial gas
+    // provisioning) to accurately represent the state history.
+    // For starting_checkpoint >= 1, remove existing checkpoints (0 and 1)
+    // and generate a consistent workload.
+    if starting_checkpoint > 0 {
```
TheMrAI, line 88: The check doesn't match with the comment above.
// For starting_checkpoint >= 1,

crates/iota-synthetic-ingestion/src/synthetic_ingestion.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct Config {
+    /// Directory to write the ingestion data to.
+    #[arg(long)]
+    pub ingestion_dir: PathBuf,
+    /// Starting checkpoint sequence number for workload generation.
+    /// Useful for benchmarking or testing against a non-empty database.
+    #[arg(long, default_value_t = Self::DEFAULT_STARTING_CHECKPOINT)]
+    pub starting_checkpoint: u64,
+    /// Number of checkpoints to generate.
+    /// If `starting_checkpoint` is 0 (default), two additional initial
```
TheMrAI, line 34: This edge case behavior looks rather confusing.
If I set starting checkpoint to 0 then I get two additional checkpoint files, so the next checkpoint will actually start from 2?
But if it is 1, it starts from one, 2 from two.

Is a cleaner abstraction perhaps possible? This looks like we are asking for trouble.

TheMrAI (approved):
The vm-lang maintained part contains the change provided by the linked Mysten change.
In that regard I approve, for the rest I made some comments, but won't hold you up with any of them.

crates/iota-synthetic-ingestion/src/synthetic_ingestion.rs
samuel-rufi: I considered what you mean here, unfortunately support for these two cases introduces some additional complexity in the generator. If the user starts generation from zero, we should preserve the history for consistency. 
If the user starts generation for a non-empty database (e.g. start from 1000) files from that point should be generated. The output of the generator appears plausible and shouldn't really require much user consideration. I also thought about the naming of the `num_checkpoints` variable, however it appears to fit both cases better than alternatives. So at this point, I would stick with this logic if you don't mind, as it appears to balance usability with the additional complexity introduced.

crates/iota-synthetic-ingestion/src/synthetic_ingestion.rs
samuel-rufi: Fixed https://github.com/iotaledger/iota/pull/7868/commits/130ea2e7c7c5622a4b4a95c53ae7fbb7eb51856b

crates/iota-synthetic-ingestion/README.md
samuel-rufi: Fixed by https://github.com/iotaledger/iota/pull/7868/commits/8962537f927d0c1ecd6ba9949f3641d937ae5a17

crates/iota-types/src/mock_checkpoint_builder.rs
samuel-rufi: Good catch! Although upstream provides this flexibility by not strictly enforcing monotonicity (which can sometimes help in certain testing scenarios), I’ve explicitly added an assertion to enforce strict checkpoint sequence ordering. Following that, overriding checkpoint 0 will no longer be possible, and I removed the `Option`.

crates/iota-synthetic-ingestion/README.md
TheMrAI: Legend 💪

PR #7906, opened 2025-07-21
feat(iota-core): add unit tests for the gas price feedback mechanism

roman1e2f5p8s:
# Description of change

- This PR adds unit tests to test the gas price feedback mechanism. The aim is to test all the components: the mechanism itself, calculations of suggested gas price and its propagation across multiple commits.
- Instead of panic, the calculator will now suggest reference gas price (or max clearing gas price) if transaction duration exceeds `max_execution_duration_per_commit`.
- Renamed `congested_objects_gas_price_feedback_mechanism` feature flag to `congestion_control_gas_price_feedback_mechanism`.

```console
cargo simtest -p iota-core unit_tests gas_price_feedback_tests
```

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
+++ b/crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
@@ -9,10 +9,11 @@ use iota_protocol_config::{
 use iota_types::{
     base_types::{IotaAddress, ObjectID, ObjectRef},
     crypto::{AccountKeyPair, get_key_pair},
-    effects::TransactionEffectsAPI,
+    effects::{TransactionEffects, TransactionEffectsAPI},
+    executable_transaction::VerifiedExecutableTransaction,
     object::Object,
     programmable_transaction_builder::ProgrammableTransactionBuilder,
-    transaction::{ObjectArg, Transaction},
+    transaction::{ObjectArg, Transaction, VerifiedCertificate},
 };
 use move_core_types::ident_str;
 use rand::seq::SliceRandom;
@@ -159,7 +160,7 @@ impl GasPriceFeedbackTester {
         effects.created()[0].0
     }
 
-    async fn build_increment_both_counters_tx(
+    async fn build_increment_both_counters_transaction(
         &self,
         gas_object_id: &ObjectID,
         // gas_price: u64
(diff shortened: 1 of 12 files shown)


thibault-martinez (approved):
Approving dev-tools

muXxer (approved):

TheMrAI (approved):
vm-lang parst lgtm

crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
```diff
@@ -0,0 +1,1825 @@
...
+        .send_certificates_to_consensus_for_scheduling(&certificates)
+        .await;
+}
+
+// Test that the suggested gas price calculator panics if
+// `max_execution_duration_per_commit` is set too low such that even
+// one transaction cannot fit in a commit.
+#[tokio::test]
+#[should_panic] // because `max_execution_duration_per_commit` is set too low.
+async fn max_execution_duration_per_commit_too_low_in_total_gas_budget_mode() {
```
cyberphysic4l: This is a fair test case, but do we actually have any safe guards against a certain transaction having gas budget greater than max_execution_duration_per_commit? It seems possible that it could happen.

cyberphysic4l (commented):
Tests look good, just have one concern about the fact that the suggestion calculation can panic in the case that gas budget of a transaction exceeds the max_execution_duration_per_commit.

crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
roman1e2f5p8s: Thanks for looking at the tests. No, it seems we do not have any safeguards against a certain transaction having gas budget greater than `max_execution_duration_per_commit`. Looking at the numbers in Sui for `TotalGasBudgetWithCap` mode, `max_accumulated_txn_cost_per_object_in_mysticeti_commit` was set `37_000_000`, while max gas budget was 50K SUI, so seems it would be possible to send a transaction that would never fit. In IOTA, max gas budget is currently set to 50 IOTA, but we never has `TotalGasBudget` mode enabled. I think it would be just safer to return the maximum gas price there instead of panic, I will change that part. Thanks for spotting this.

roman1e2f5p8s: Nice catch, @cyberphysic4l! I changed the logic for txs with durations exceeding `max_execution_duration_per_commit`.

piotrm50 (approved):

cyberphysic4l (approved):

PR #7910, opened 2025-07-21
refactor(iota-move-build): Refactor BuildConfig

### LINKED ISSUE #7859: [iota-build] Refactors to `BuildConfig` · MystenLabs/sui@94dc4f8
https://github.com/MystenLabs/sui/commit/94dc4f8ed7d8519a4a499612ac594dd348e19808

TheMrAI:
# Description of change

`new_for_testing` more explicitly specifies the set values and
`build_from_resolution_graph` has been separated to multiple functions.

Removed BuildConfig, Default trait implementation.
Ideally the Default trait should be kept, with properly documented expectations, but since Sui
is removing it and generally refuses to rely on standard behaviors
we can't trust, that it won't change in unexpected ways either.

To mitigate further issues iota-genesis-builder now explicitly sets
all required configurations. This ensures build failure if a new
field is added and that it won't depend on any undocumented behaviors.

Furthermore we introduced a BuildConfig::new() function, by which
users of iota-sdk and iota-move-build, may be able to construct
a BuildConfig. Otherwise they can't as they simply don't and shouldn't
have access to move_package::BuildConfig (external move crate struct).
In fact this highlights some design flaws in iota-move-build as
external users of BuildConfig may see there is a public config field
on the struct, yet have no actual access to it.

fixes #7859 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


valeriyr (changes requested):
Why didn't you remove the Default trait implementation?
```
impl Default for BuildConfig {
...
}
```
Looks like it was the main reason of this change.

muXxer (approved):
LGTM for the files we own.
image

crates/iota-sdk/examples/transaction_builder/move_package.rs
```diff
@@ -26,8 +26,13 @@ async fn main() -> Result<(), anyhow::Error> {
...
 
-    let package_path = Path::new("../../examples/move/first_package");
-    let module = BuildConfig::default().build(package_path)?;
+    let package_path = [
+        env!("CARGO_MANIFEST_DIR"),
+        "../../examples/move/first_package",
+    ]
+    .iter()
+    .collect::<PathBuf>();
+    let module = BuildConfig::new(true, false, None).build(&package_path)?;
```
miker83z:
```suggestion
    let module = BuildConfig::new(
        true,  // run_bytecode_verifier
        false, // print_diags_to_stderr
        None,  // chain_id
    )
    .build(&package_path)?;
```

crates/iota-sdk/examples/transaction_builder/move_package.rs
```diff
@@ -80,7 +85,7 @@ async fn main() -> Result<(), anyhow::Error> {
         .expect("missing upgrade cap");
 
     // In reality you would like to do some changes to the package before upgrading
-    let module = BuildConfig::default().build(package_path)?;
+    let module = BuildConfig::new(true, false, None).build(&package_path)?;
```
miker83z:
```suggestion
    let module = BuildConfig::new(
        true,  // run_bytecode_verifier
        false, // print_diags_to_stderr
        None,  // chain_id
    )
    .build(&package_path)?;
```

miker83z (approved):
LGTM with a nit

crates/iota-sdk/examples/transaction_builder/move_package.rs
TheMrAI: Isn't that a bit redundant?
rust-analyser (or other tooling) will show you the name of the fields, because I have kept them exactly as you show in the comments.

miker83z: I think for booleans is a good practice to have some comments like this and this is done several times in the codebase. Moreover, I don't assume everyone uses the rust-analyzer.  

I was in doubt about adding the comment for the chain_id Option as well, but I guess it also helps there when the option is None.

crates/iota-move-build/src/lib.rs
```diff
@@ -275,42 +338,22 @@ pub fn build_from_resolution_graph(
...
     }
-
-    Ok(CompiledPackage {
-        package,
-        published_at,
-        dependency_ids,
-        bytecode_deps,
-        dependency_graph: resolution_graph.graph,
-    })
+    // TODO(https://github.com/iotaledger/iota/issues/69): Run Move linker
```
Thoralf-M: This was wrongly changed before
```suggestion
    // TODO(https://github.com/MystenLabs/sui/issues/69): Run Move linker
```

crates/iota/tests/cli_tests.rs
```diff
@@ -5199,6 +5199,7 @@ async fn test_tree_shaking_package_with_unpublished_deps() -> Result<(), anyhow:
 }
 
 #[sim_test]
+#[ignore] // TODO: DVX-786
```
Thoralf-M, line 5210: What does this mean? Why is it ignored?

crates/iota-sdk/examples/transaction_builder/move_package.rs
TheMrAI: Not just rust-analyser, other tools also can detect these and can provide you the appropriate values.
Very few write code without appropriate IDEs and especially Rust, given its complexity, and who do, are UFOs enough that they don't care.

In general I would consider it bad practice to add such comments into arguments. For one if someone refactors the name of one variable, then they have to search and replace all these support comments, which would be a needles complexity in development cycles.

It would be better to have it as a structure and rely on the Rust language.

```
let config = BuildConfig {
    run_bytecode_verifier: true,
    print_diags_to_stderr: true,
    chain_id: None
}
```

Now you can't really use this, because of the `MoveBuildConfig`, that the user doesn't have access too.

So in summary, I agree, with your point that currently it isn't expressive (easy to read enough). But the proposed solution wouldn't fix the real issue either, just attempt to hide it.
The interface should be changed, to be properly user facing with appropriate expectations, but we can't really do that.

crates/iota-move-build/src/lib.rs
TheMrAI: Fixed in: 37fcd716cd429717bcdef2e4de913e8ea744d9a4

crates/iota-sdk/examples/transaction_builder/move_package.rs
valeriyr: As an option for consideration, you can do this as well:
```rust
let run_bytecode_verifier = false;
let print_diags_to_stderr = false;
let chain_id = None;

let module = BuildConfig::new(
        run_bytecode_verifier,
        print_diags_to_stderr,
        chain_id,
    )
...
```
For me, it is better than comments because it is checked by the compiler.

valeriyr: oh, it is a function, sorry, then it will not work. only for better reading :)

TheMrAI: Fixed it: 6df49a77490622e9ce94d786d06f16116c295180
As an extra removed the code duplication for double initialization as well.

valeriyr (approved):
LGTM!

jkrvivian (approved):

crates/iota-sdk/Cargo.toml
```diff
@@ -50,6 +50,8 @@ tempfile.workspace = true
 
 # internal dependencies
 iota-move-build.workspace = true
+move-package.workspace = true
+move-compiler.workspace = true
```
TheMrAI, line 53: Summoning: @Thoralf-M , @thibault-martinez , @DaughterOfMars 
Please observe the move_packages.rs example.
It is unclear to me what the example is supposed to demonstrate, because many types it uses aren't exposed to the users. The sdk doesn't seem to propagate them.

For example `BuildConfig` can only be used because iota-move-build has been added as a dev dependency. But it is a dev dependency, so its access will be restricted.
Similarly I added move-package and move-compiler so that `BuildConfig` can be properly configured. Please advise if it is good or what should happen.

DaughterOfMars: I'm not sure what you mean. If a type is used in an example, it must be publicly exposed by the crate.

TheMrAI: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples
```
Examples can use the public API of the package’s library. They are also linked with the [[dependencies]](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) and [[dev-dependencies]](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) defined in Cargo.toml.
```
Here in the move-package example only seems to compile because iota-move-build was added as a dev dependency. 
So if the examples are meant to show some external user how to use the iota-sdk, this example may be problematic as the sdk itself doesn't provide some types, to which the user may or may not have access to otherwise (I am unsure what iota-sdk meant to represent so I may be mixing assumptions here).

Thoralf-M: > For example BuildConfig can only be used because iota-move-build has been added as a dev dependency. But it is a dev dependency, so its access will be restricted.

Everyone can add it as dependency to his own project and use it like in the example

DaughterOfMars: Talked about it in a call and we concluded that there is no issue here 👍

TheMrAI: Good thanks. Just wanted to make sure. :)

PR #7918, opened 2025-07-22
feat(iota): Simplify the `pay-iota` command

### LINKED ISSUE #7791: CLI: add simpler way to send IOTA to another address
## Description

client pay-iota is too complicated with the params, we should have something like
`iota client send <address> <amount>` no other params required

## Motivation

Easier to use, current commands are complicated

DaughterOfMars:
# Description of change

Simplifies the `pay-iota` command by making the input coins optional. If they are not provided, then coins satisfying the requested amounts will be selected and merged into the gas coin input so that it can be split out and transferred.

Closes #7791

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Simplify the `pay-iota` command by making the input coins optional.
- [ ] Rust SDK:
- [ ] REST API:


crates/iota/src/client_commands.rs
```diff
@@ -3235,3 +3241,36 @@ async fn check_protocol_version_and_warn(client: &IotaClient) -> Result<(), anyh
...
+    context: &mut WalletContext,
+) -> anyhow::Result<Vec<ObjectID>> {
+    let mut coins = Vec::new();
+
+    let sender = get_identity_address(sender.map(Into::into), &context).await?;
+    let mut gas_coins = context
+        .gas_objects(sender)
+        .await?
+        .iter()
+        // Ok to unwrap() since `get_gas_objects` guarantees gas
```
Thoralf-M:
```suggestion
        // Ok to unwrap() since `gas_objects` guarantees gas
```
?

crates/iota/src/client_commands.rs
```diff
@@ -3235,3 +3241,36 @@ async fn check_protocol_version_and_warn(client: &IotaClient) -> Result<(), anyh
...
+        .map(|(_val, object)| GasCoin::try_from(object).unwrap())
+        .collect::<Vec<_>>();
+    // Sort in ascending order
+    gas_coins.sort_unstable_by(|c1, c2| c1.value().cmp(&c2.value()));
+    while amount > 0 {
+        if gas_coins.is_empty() {
+            anyhow::bail!("insufficient funds for requested amounts");
```
Thoralf-M: Can we print the available and required amount here?
Also since the gas needs to be payed as well from this amount, we should probably select required amount + some extra amount for gas fee

crates/iota/src/client_commands.rs
```diff
@@ -325,9 +325,10 @@ pub enum IotaClientCommands {
     /// no extra gas coin is required.
     PayIota {
         /// The input coins to be used for pay recipients, including the gas
-        /// coin.
+        /// coin. If not provided, coins will be select automatically which
```
Alex6323:
```suggestion
        /// coin. If not provided, coins will be selected automatically which
```

crates/iota/src/client_commands.rs
```diff
@@ -1395,7 +1396,7 @@ impl IotaClientCommands {
                 opts,
             } => {
                 ensure!(
-                    !input_coins.is_empty(),
+                    !input_coins.as_ref().is_some_and(|v| v.is_empty()),
```
Alex6323, line 1473:
```suggestion
                    !input_coins.as_ref().is_some_and(|v| !v.is_empty()),
```
? :thinking:

crates/iota/src/client_commands.rs
```diff
@@ -3235,3 +3241,36 @@ async fn check_protocol_version_and_warn(client: &IotaClient) -> Result<(), anyh
...
+        // Ok to unwrap() since `get_gas_objects` guarantees gas
+        .map(|(_val, object)| GasCoin::try_from(object).unwrap())
+        .collect::<Vec<_>>();
+    // Sort in ascending order
+    gas_coins.sort_unstable_by(|c1, c2| c1.value().cmp(&c2.value()));
+    while amount > 0 {
+        if gas_coins.is_empty() {
+            anyhow::bail!("insufficient funds for requested amounts");
+        }
+        let coin = if let Some(idx) = gas_coins.iter().position(|c| c.value() >= amount) {
```
Alex6323: could employ a `binary_search_by` now, I guess

crates/iota/src/client_commands.rs
DaughterOfMars: nah that's not right

crates/iota/src/client_commands.rs
DaughterOfMars: Well the idea is to find the first with high enough value without going over, so I'm not sure a binary search works since it tries to find an exact value

crates/iota/src/client_commands.rs
Alex6323: ah, now I get it :see_no_evil:

crates/iota/src/client_commands.rs
DaughterOfMars: I printed the needed amount but in order to print the available amount I would need to sum it up before doing the loop which seems a bit weird. I can do it though, if you think it's worth it

crates/iota/src/client_commands.rs
Alex6323: ah, yes, I see. we want the smallest that just has enough value. Makes sense. And without this conditition picking the last gas coin of the vec and checking its value would be even better than a binary search.

Alex6323 (approved):

muXxer (approved):

crates/iota/src/client_commands.rs
Thoralf-M: It would be `amount - amount_remaining`, no?

DaughterOfMars: Oh, true 😆  nice!

Thoralf-M (approved):

thibault-martinez:
Clippy

thibault-martinez:
conflicts

crates/iota/src/client_commands.rs
```diff
@@ -336,9 +336,10 @@ pub enum IotaClientCommands {
     /// no extra gas coin is required.
     PayIota {
         /// The input coins to be used for pay recipients, including the gas
-        /// coin.
+        /// coin. If not provided, coins will be selected automatically which
+        /// fulfill the requested amounts.
```
thibault-martinez, line 340: I think this sentence reads a bit weird

crates/iota/src/client_commands.rs
```diff
@@ -3422,3 +3453,39 @@ async fn check_protocol_version_and_warn(client: &IotaClient) -> Result<(), anyh
...
+    gas_coins.sort_unstable_by_key(|c| c.value());
+    let mut amount_remaining = amount;
+    while amount_remaining > 0 {
+        if gas_coins.is_empty() {
+            anyhow::bail!(
+                "insufficient funds for requested amount: {amount}, available: {}",
+                amount - amount_remaining
+            );
+        }
+        let coin = if let Some(idx) = gas_coins.iter().position(|c| c.value() >= amount_remaining) {
```
thibault-martinez: Why do we not just pop since they are sorted?

crates/iota/src/client_commands.rs
DaughterOfMars: I think it's fine 🤷‍♀️

crates/iota/src/client_commands.rs
DaughterOfMars: I did it because I wanted to choose the least amount while still fulfilling the requirement, but actually I think it may not be necessary after all since it's provided as the gas coin

thibault-martinez: But the way it is at the moment, you either select the first one or you go to the other branch and pop, same thing

thibault-martinez: Ahhh wait it's ascending? My bad

crates/iota/src/client_commands.rs
thibault-martinez: Alright

thibault-martinez (approved):

PR #7921, opened 2025-07-22
feat(consensus): Improvements on the validator scoring metrics

oliviasaa:
# Description of change

- General readability improvements to the validator scoring mechanism.
- Exposes the Score as a Prometheus metrics
- Adds more extensive tests for the equivocations and missing blocks scoring metrics

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


consensus/core/src/dag_state.rs
```diff
@@ -246,36 +246,21 @@ impl DagState {
...
-            .authorities()
-            .map(|(_, x)| x.hostname.as_str())
-            .collect();
-        state
-            .context
-            .metrics
-            .initialize_uncached_scoring_metrics(recovered_scoring_metrics, hostnames);
-
-        // Initialize the scoring metrics relative to rounds that were loaded to cache.
+        let blocks_in_cache = &state.recent_refs_by_authority;
```
piotrm50: Do we need those variables? Would it be possible to pass those references directly?

consensus/core/src/dag_state.rs
```diff
@@ -246,36 +246,21 @@ impl DagState {
...
-            state.context.metrics.initialize_cached_scoring_metrics(
-                authority_index,
-                hostname,
-                block_rounds_in_cache,
-                threshold_clock_round,
-                eviction_round,
-            );
-        }
+        let eviction_rounds = &state.evicted_rounds;
+        let context = &state.context;
```
piotrm50: context should be cloned, as it's wrapped in an Arc. You can check how it's done in other places.

consensus/core/src/metrics.rs
```diff
@@ -98,15 +99,25 @@ enum MetricType {
...
         hostname: &str,
+        recent_refs: &BTreeSet<BlockRef>,
         eviction_round: u32,
-        last_evicted_round: u32,
-        recent_refs_by_authority: &BTreeSet<BlockRef>,
+        last_eviction_round: u32,
         threshold_clock_round: u32,
     ) -> Option<StoredScoringMetricsU64> {
+        // threshold_clock_round will be always at least 1. However, this method will
+        // panic if threshold_clock_round = 0 (subtraction with overflow). We
```
piotrm50: To prevent accidental panics I think we could additionally use `saturating_sub` function.

consensus/core/src/dag_state.rs
oliviasaa: no, I just thought this way would be more readable. I'll remove them

consensus/core/src/dag_state.rs
oliviasaa: done

consensus/core/src/metrics.rs
oliviasaa: done

consensus/core/src/metrics.rs
```diff
@@ -1451,62 +1514,197 @@ mod tests {
...
+        matches!(
+            stored_metrics,
+            Some(StoredScoringMetricsU64 {
+                faulty_blocks_provable: 0,
+                faulty_blocks_unprovable: 0,
+                equivocations: 0,
+                missing_proposals: 1
+            })
+        );
+    }
```
cyberphysic4l: Are most of these scenarios in the test actually never supposed to happen based on the gc logic? Maybe you could comment on each scenario and say something about what it actually means or when that might happen, or is it just a random few combinations of numbers? No harm to have tests showing it won't panic I suppose anyway.

cyberphysic4l (approved):

consensus/core/src/metrics.rs
oliviasaa: yes, they are not supposed to happen. I'm just testing this because, IIUC, the functions should not panic "independently". I'll write some comments about the tests, at least explaining that

consensus/core/src/metrics.rs
piotrm50: Comment is correct as the function will not panic due to this check as well as saturating_sub

piotrm50 (approved):

consensus/core/src/metrics.rs
oliviasaa: done

consensus/core/src/metrics.rs
oliviasaa: done

PR #7945, opened 2025-07-23
fix(core): use gas_owner() instead of sender() for gas object creation

### LINKED ISSUE #7892: iota-core: dry run ignores potential gas sponsor
## Description

Instead of the sender `transaction.sender()`, the gas owner `transaction.gas_owner()` should be used in https://github.com/iotaledger/iota/blob/32894142e11554c81d3ebd9eb94865b103aa9f2c/crates/iota-core/src/authority.rs#L1765 and probably also
https://github.com/iotaledger/iota/blob/32894142e11554c81d3ebd9eb94865b103aa9f2c/crates/iota-core/src/authority.rs#L1950

## Motivation

Doing a dry run like this (--gas-sponsor gets added in this PR https://github.com/iotaledger/iota/pull/7881)
```
ACTIVE_ADDRESS=$(iota client active-address)
# Get the first non-active address from the list
GAS_SPONSOR=$(iota client addresses --json | jq -r --arg active $ACTIVE_ADDRESS '
  .addresses[] | select(.[1] != $active) | .[1]
' | head -n 1)
COIN_OBJECT_ID=$(iota client gas --json | jq -r '.[0].gasCoinId')

iota client pay \
    --input-coins $COIN_OBJECT_ID \
    --amounts 0 \
    --recipients $ACTIVE_ADDRESS \
    --gas-sponsor $GAS_SPONSOR \
    --dry-run
```
will have the sender also as owner of the gas coin, which is unexpected as it doesn't take a potential gas sponsor into account, so in the response the gas object will be owned by ACTIVE_ADDRESS instead of GAS_SPONSOR

## Requirements

1. Use the gas owner instead of the sender address for the created gas object

No.

jkrvivian:
# Description of change

1. Replace `transaction.sender()` with `transaction.gas_owner()` when creating mock gas objects in `dry_exec_transaction_impl()` and `simulate_transaction_impl()`.
This ensures that gas objects are properly owned by the gas sponsor in sponsored transactions, rather than the transaction sender.

2. Include the mock gas object in the transaction's gas payment for both `dry_exec_transaction_impl()` and `simulate_transaction_impl()`. Previously, it wasn't added, causing misalignment of `transaction.gas()` in `check_objects()`.

Fixes #7892 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1772,7 +1772,7 @@ impl AuthorityState {
         let mut gas_object_refs = transaction.gas().to_vec();
         let reference_gas_price = epoch_store.reference_gas_price();
         let ((gas_status, checked_input_objects), mock_gas) = if transaction.gas().is_empty() {
-            let sender = transaction.sender();
+            let sender = transaction.gas_owner();
             // use a 1B iota coin
             const NANOS_TO_IOTA: u64 = 1_000_000_000;
             const DRY_RUN_IOTA: u64 = 1_000_000_000;
@@ -1957,7 +1957,7 @@ impl AuthorityState {
         // make a gas object if one was not provided
         let mut gas_object_refs = transaction.gas().to_vec();
         let ((gas_status, checked_input_objects), mock_gas) = if transaction.gas().is_empty() {
-            let sender = transaction.sender();
+            let sender = transaction.gas_owner();
             // use a 1B iota coin
             const NANOS_TO_IOTA: u64 = 1_000_000_000;
             const DRY_RUN_IOTA: u64 = 1_000_000_000;
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1769,6 +1769,7 @@ impl AuthorityState {
         )?;
 
         // make a gas object if one was not provided
+        let mut transaction = transaction;
         let mut gas_object_refs = transaction.gas().to_vec();
         let reference_gas_price = epoch_store.reference_gas_price();
         let ((gas_status, checked_input_objects), mock_gas) = if transaction.gas().is_empty() {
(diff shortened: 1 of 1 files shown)


muXxer (approved):

Thoralf-M (approved):

Thoralf-M:
resolve_transaction_transfer_with_sponsor test fails

piotrm50 (approved):

crates/iota-e2e-tests/tests/rest.rs
```diff
@@ -163,6 +173,11 @@ async fn resolve_transaction_transfer_with_sponsor() {
         sender: sender.into(),
         gas_payment: Some(UnresolvedGasPayment {
             objects: vec![],
+            // objects: vec![UnresolvedObjectReference {
+            //     object_id: sponsor_gas_coins.0.into(),
+            //     version: None,
+            //     digest: None,
+            // }],
```
Thoralf-M: Why did you comment this? It should be possible for the user to provide the gas coin

TheMrAI: Yeah, but in the test it will be sponsored no?
I think the comments should just be removed.

TheMrAI (commented):
Just remove the commented sections if they are no longer necessary. Otherwise LGTM

crates/iota-e2e-tests/tests/rest.rs
jkrvivian: I forced-push a commit to add the mock gas object to transaction gas payment in `dry_exec_transaction_impl()` and `simulate_transaction_impl()`.
The commented codes are removed.

TheMrAI (approved):

crates/iota-e2e-tests/tests/rest.rs
Thoralf-M: But now you're always just creating a new gas object out of thin air, no? As a user I might want to do a dry run and provide an actual gas object myself

Thoralf-M: Ok I see you only create the gas coin if no payment was provided, but then why did the test need to be modified? Now you're only testing the case where no payment is provided

jkrvivian: Tests remain unchanged.

Thoralf-M: Ah sorry, the force pushes make it a bit hard to track, so all good now

Thoralf-M (approved):

muXxer (approved):

PR #7559, opened 2025-06-26
feat(iota-core): import congestion tracker

salaheldinsoliman:
# Description of change

This PR imports `CongestionTracker`, whose objective is to suggest a gas price for shared-object transactions via dry running.

- Porting commits: 
  - https://github.com/MystenLabs/sui/commit/d4a321caf9b5257f00ec4a044edfdb784aeef766
  - https://github.com/MystenLabs/sui/commit/ffd74dac4ccdbf9b6979b681a87a3c83bfd25e9a
  - https://github.com/MystenLabs/sui/commit/2d9a68beb38a2de09c56d1d665833bf71abcc251
- Description:
  - No logic changes, I only did some renaming and added doc strings in `congetion_tracker.rs`.

## Links to any relevant issues

Does not have a dedicated issue, but fixes #8021.

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

PR #7946, opened 2025-07-23
feat(iota-core): adjust weights for gas price prediction based on online learning algorithm

vekkiokonio:
# Description of change

This PR introduces a predictive mechanism for estimating the minimum gas price required for transactions involving shared objects.

The predictor builds upon historical congestion data collected by the `CongestionTracker` module (see #7559) and dynamically adjusts object-specific weights using an online gradient descent (OGD) algorithm.

The core function, `get_suggested_gas_price_with_ogd`, estimates a gas price based on the set of shared objects accessed by a transaction. It aims to balance execution success (avoiding underestimation) with user cost efficiency (avoiding overestimation).

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Introduced computation of object hotness based on online gradient descent improving the upstream's logic for recommended suggested gas price calculations.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


salaheldinsoliman:
This PR is the other approach of vectorising and labeling a Tx based on the touched objetcs: #7980

crates/iota-core/src/authority.rs
```diff
@@ -1823,7 +1823,12 @@ impl AuthorityState {
         let module_cache =
             TemporaryModuleResolver::new(&inner_temp_store, epoch_store.module_cache().clone());
 
-         let mut layout_resolver =
+        let tx_clone = transaction.clone();
+        let tracker_clone = self.congestion_tracker.clone();
+
+        let suggested_gas_price = tracker_clone.get_suggested_gas_price_with_ogd(tx_clone);
+
```
roman1e2f5p8s:
```suggestion
```
This is unnecessary cloning that can be avoided.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -140,40 +160,52 @@ impl CongestionTracker {
                 .map(|id| id.id),
         )
     }
+
+    /// For all the mutable shared inputs, sum the hotness of the objects.
+    /// More sophisticated prediction can be implemented.
+    pub fn get_suggested_gas_price_with_ogd(&self, transaction: TransactionData) -> Option<u64> {
```
roman1e2f5p8s:
```suggestion
    pub fn get_suggested_gas_price_with_ogd(&self, transaction: &TransactionData) -> Option<u64> {
```

crates/iota-core/src/authority.rs
```diff
@@ -1882,9 +1887,7 @@ impl AuthorityState {
                 events,
                 object_changes,
                 balance_changes,
-                suggested_gas_price: self
-                    .congestion_tracker
-                    .get_suggested_gas_prices(&transaction),
+                suggested_gas_price,
```
roman1e2f5p8s:
```suggestion
                suggested_gas_price: self
                    .congestion_tracker
                    .get_suggested_gas_price_with_ogd(&transaction),
```
Move it to the beginning before the `input` field, then you can avoid cloning transaction in the `input`.
If not clear, please check how this is done on the base branch.

roman1e2f5p8s (commented):
Cloning can be avoided.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -12,22 +12,19 @@ use iota_types::{
...
-/// congested) shared objects.
-type TransactionGasPriceMutSharedObjectsPair = (u64, Vec<ObjectID>);
-
 /// Alias for type holding congestion info per checkpoint.
 type CongestionInfoMap = HashMap<ObjectID, CongestionInfo>;
 
 /// Holds tracked per-object congestion info.
 #[derive(Clone, Copy, Debug)]
-struct CongestionInfo {
+pub struct CongestionInfo {
```
roman1e2f5p8s:
```suggestion
struct CongestionInfo {
```
Used only in this module, so does not need to be public.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -42,8 +39,15 @@ struct CongestionInfo {
 
     /// Lowest gas price of clearing transaction(s) accessing the object.
     lowest_clearing_gas_price: Option<u64>,
+
+    /// The hotness of an object corresponds to the expected tip to pay for a
+    /// successful execution.
+    pub hotness: f64,
```
roman1e2f5p8s:
```suggestion
    hotness: f64,
```
Used only in this module, so does not need to be public.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -83,21 +111,16 @@ impl CongestionInfo {
...
-    /// Key-value-based cache storing congestion info of objects.
-    object_congestion_info: Cache<ObjectID, CongestionInfo>,
-}
-
-impl Default for CongestionTracker {
-    fn default() -> Self {
-        Self::new()
-    }
+    pub reference_gas_price: u64,
+    pub object_congestion_info: Cache<ObjectID, CongestionInfo>,
```
roman1e2f5p8s:
```suggestion
    reference_gas_price: u64,
    object_congestion_info: Cache<ObjectID, CongestionInfo>,
```
Used only in this module, so does not need to be public.

roman1e2f5p8s (commented):
Please make fields used only in `congestion_tracker.rs` private.
Above, I suggested three appropriate changes, which can be committed in one batch from the `Files changed` tab.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -18,9 +18,14 @@ use crate::execution_cache::TransactionCacheRead;
 /// Capacity of the congestion tracker's cache.
 const CONGESTION_TRACKER_CACHE_CAPACITY: u64 = 10_000;
 
-/// Alias type for holding transaction's gas price and mutable (or
-/// congested) shared objects.
-type TransactionGasPriceMutSharedObjectsPair = (u64, Vec<ObjectID>);
+/// Threshold for hotness below which an object is considered cold.
+const HOTNESS_CUTOFF: f64 = 1.0;
```
roman1e2f5p8s: Would be nice to add (directly to the doc string) which values this const should accept.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -18,9 +18,14 @@ use crate::execution_cache::TransactionCacheRead;
...
+
+/// Controls how quickly congestion feedback updates object hotness.
+const HOTNESS_ADJUSTMENT_FACTOR: f64 = 0.2; // > 0.0
```
roman1e2f5p8s: Did you mean congestion tracker? Also, I would suggest putting accepted values directly in the docstring.

```suggestion
/// Controls how quickly congestion tracker updates object hotness.
/// Values should be > 0.0.
const HOTNESS_ADJUSTMENT_FACTOR: f64 = 0.2;
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -18,9 +18,14 @@ use crate::execution_cache::TransactionCacheRead;
...
+
+/// Controls how quickly hotness decays for objects not seen in congestion.
+const HOTNESS_DECAY_FACTOR: f64 = 1.5; // > 1.0 = decay, 1.0 = no decay
```
roman1e2f5p8s: Add accepted values to the doc string:

```suggestion
/// Controls how quickly hotness decays for objects not seen in congestion.
/// Values should be >= 1.0: set to > 1.0 for decay, or 1.0 for no decay.
const HOTNESS_DECAY_FACTOR: f64 = 1.5;
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -42,6 +47,10 @@ struct CongestionInfo {
 
     /// Lowest gas price of clearing transaction(s) accessing the object.
     lowest_clearing_gas_price: Option<u64>,
+
+    /// The hotness of an object corresponds to the expected tip to pay for a
+    /// successful execution.
+    hotness: f64,
```
roman1e2f5p8s: Hotness cannot be negative, right? In the doc string, I would add what values hotness can take.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
...
     }
 
-    /// Update the highest congestion gas price with the new `gas_price`.
-    fn update_highest_congestion_gas_price(&mut self, gas_price: u64) {
-        self.highest_congestion_gas_price = self.highest_congestion_gas_price.max(gas_price);
+    fn update_hotness(
+        &mut self,
+        new: &CongestionInfo,
+        number_congested_transactions: usize,
+        _number_cleared_transactions: usize,
```
roman1e2f5p8s: Remove it from functions if it is not used.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
...
+        } else {
+            self.hotness = old_hotness;
+        }
+    }
+
+    fn update_hotness_for_new_object(
+        &mut self,
+        new: &CongestionInfo,
+        number_congested_transactions: usize,
+        _number_cleared_transactions: usize,
```
roman1e2f5p8s: Remove it from functions if it is not used.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
...
+                -new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;
+        }
+        // Ensure hotness is non-negative
+        self.hotness = self.hotness.max(0.0);
+    }
+
+    fn update_for_cancellation(&mut self, now: CheckpointTimestamp, gas_price: u64) {
+        self.latest_congestion_time = now;
+        self.highest_congestion_gas_price =
+            std::cmp::max(self.highest_congestion_gas_price, gas_price);
```
roman1e2f5p8s:
```suggestion
        /// Update the highest gas price and the latest time with the data from a
    /// congested transaction.
    fn update_for_congested_tx(&mut self, time: CheckpointTimestamp, gas_price: u64) {
        self.latest_congestion_time = time;
        self.highest_congestion_gas_price = self.highest_congestion_gas_price.max(gas_price);
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
         }
     }
 
-    /// Update the highest congestion gas price with the new `gas_price`.
-    fn update_highest_congestion_gas_price(&mut self, gas_price: u64) {
-        self.highest_congestion_gas_price = self.highest_congestion_gas_price.max(gas_price);
+    fn update_hotness(
```
roman1e2f5p8s, line 87: If you want to keep both functions for updating hotness, I would suggest renaming this function to `update_hotness_for_existing_object`, but I think there should only be one function for updating hotness, see comment on `update_hotness_for_new_object` below.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
...
+            // Update object i's hotness based on the congestion events following the
+            // formula: hotness_i -= SUM(tx)[SUM(i in tx)(hotness_i- gas_price_feedback_tx)]
+            // * LEARNING_RATE / number_congested_transactions.
+            self.hotness -=
+                new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;
+            // Hotness decrease is capped by HOTNESS_DECAY_FACTOR.
+            self.hotness = self.hotness.max(old_hotness / HOTNESS_DECAY_FACTOR);
+        } else {
+            self.hotness = old_hotness;
+        }
```
roman1e2f5p8s: The update in the else branch is redundant. I would rewrite the function as follows:

```suggestion
        if number_congested_transactions > 0 {
            // Update object i's hotness based on the congestion events following the
            // formula: hotness_i -= SUM(tx)[SUM(j in tx)(hotness_j - gas_price_feedback)]
            // * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions.
            let new_hotness = self.hotness
                - new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;
            // Hotness decrease is capped by HOTNESS_DECAY_FACTOR.
            self.hotness = new_hotness.max(self.hotness / HOTNESS_DECAY_FACTOR);
        }
```

Also, note the changes in the comment. We don't have any sum here, so maybe the comment should be updated too.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,9 +73,44 @@ impl CongestionInfo {
...
+        new: &CongestionInfo,
+        number_congested_transactions: usize,
+        _number_cleared_transactions: usize,
+    ) {
+        if number_congested_transactions > 0 {
+            self.hotness =
+                -new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;
+        }
+        // Ensure hotness is non-negative
+        self.hotness = self.hotness.max(0.0);
```
roman1e2f5p8s: This logic is very similar to the function above. Why not having a single function to update the hotness?
Something like this:

```rust
     fn update_hotness(
        &mut self,
        new: &CongestionInfo,
        number_congested_transactions: usize,
        new_object: bool,
    ) {
        if number_congested_transactions > 0 {
            // Update object i's hotness based on the congestion events following the
            // formula: hotness_i -= SUM(tx)[SUM(j in tx)(hotness_j - gas_price_feedback)]
            // * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions.
            let new_hotness = self.hotness
                - new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;

            if !new_object {
                // Hotness decrease is capped by HOTNESS_DECAY_FACTOR.
                self.hotness = new_hotness.max(self.hotness / HOTNESS_DECAY_FACTOR);
            } else {
                // Ensure hotness is non-negative
                self.hotness = new_hotness.max(0.0);
            }
        }
    }
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -83,21 +127,16 @@ impl CongestionInfo {
 /// `CongestionTracker` tracks objects' congestion info.
 /// The info is then used to calculated a suggested gas price.
 pub struct CongestionTracker {
-    /// Key-value-based cache storing congestion info of objects.
+    reference_gas_price: u64,
     object_congestion_info: Cache<ObjectID, CongestionInfo>,
```
roman1e2f5p8s, line 133:
```suggestion
    /// Key-value cache for storing congestion info of objects.
    object_congestion_info: Cache<ObjectID, CongestionInfo>,
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -110,10 +149,9 @@ impl CongestionTracker {
         effects: &[TransactionEffects],
     ) {
         // Containers for checkpoint's congestion and clearing transactions data.
-        let mut congestion_txs_data: Vec<TransactionGasPriceMutSharedObjectsPair> =
-            Vec::with_capacity(effects.len());
-        let mut clearing_txs_data: Vec<TransactionGasPriceMutSharedObjectsPair> =
+        let mut congestion_txs_data: Vec<(u64, Vec<ObjectID>, u64)> =
             Vec::with_capacity(effects.len());
+        let mut clearing_txs_data: Vec<(u64, Vec<ObjectID>)> = Vec::with_capacity(effects.len());
```
roman1e2f5p8s: These types `Vec<(u64, Vec<ObjectID>, u64)>` and `Vec<(u64, Vec<ObjectID>)>` appear multiple times in the code, and it is not clear which `u64` is what. I would suggest adding at least comments for `u64`s here. But a better idea would be to create a dedicated struct to hold this data:

```rust
struct TxData {
    gas_price: u64,
    shared_objects: Vec<ObjectID>,
    feedback_gas_price: Option<u64>,
}
```

Then we could simply use `Vec<TxData>` and the code and tests would be more readable.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -169,19 +215,60 @@ impl CongestionTracker {
...
+
+    /// Returns a map of all objects and their hotness values.
+    pub fn get_all_hotness(&self) -> HashMap<ObjectID, f64> {
+        let mut hotness = HashMap::new();
+
+        for entry in self.object_congestion_info.iter() {
+            hotness.insert(*entry.0, entry.1.hotness);
+        }
+
+        hotness
```
roman1e2f5p8s:
```suggestion
        self.object_congestion_info
            .iter()
            .map(|entry| (*entry.0, entry.1.hotness))
            .collect()
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -169,19 +215,60 @@ impl CongestionTracker {
...
 
 impl CongestionTracker {
     /// Process checkpoint's congestion and clearing transactions info.
     fn process_congestion_and_clearing_txs_data(
         &self,
         time: CheckpointTimestamp,
-        congestion_txs_data: &[TransactionGasPriceMutSharedObjectsPair],
-        clearing_txs_data: &[TransactionGasPriceMutSharedObjectsPair],
+        congestion_txs_data: &[(u64, Vec<ObjectID>, u64)],
+        clearing_txs_data: &[(u64, Vec<ObjectID>)],
```
roman1e2f5p8s: I would highly recommend adding a new type for those:

```rust
struct TxData {
    gas_price: u64,
    shared_objects: Vec<ObjectID>,
    feedback_gas_price: Option<u64>,
}
```

and then use

```rust
        congestion_txs_data: &[TxData],
        clearing_txs_data: &[TxData],
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
+    ) -> Option<f64> {
+        let mut total_hotness = 0.0;
+
+        for object_id in objects {
+            if let Some(info) = self.get_congestion_info(object_id) {
+                total_hotness += info.hotness;
+            }
+        }
+        Some(total_hotness)
+    }
```
roman1e2f5p8s: The function does not need to return an `Option`, and can be rewritten as follows:

```suggestion
    fn get_total_hotness_for_objects(&self, objects: impl Iterator<Item = ObjectID>) -> f64 {
        objects
            .filter_map(|object_id| self.get_congestion_info(object_id))
            .map(|info| info.hotness)
            .sum()
    }
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -169,19 +215,60 @@ impl CongestionTracker {
...
+                transaction
+                    .shared_input_objects()
+                    .into_iter()
+                    .filter(|id| id.mutable)
+                    .map(|id| id.id),
+            )
+            .unwrap_or(0.0);
+
+        Some(self.reference_gas_price + hotness as u64)
+    }
```
roman1e2f5p8s: The function does not need to return an `Option`, and can be rewritten as follows:

```suggestion
    pub fn get_suggested_gas_price_with_ogd(&self, transaction: &TransactionData) -> u64 {
        let hotness = self
            .get_total_hotness_for_objects(
                transaction
                    .shared_input_objects()
                    .into_iter()
                    .filter(|obj| obj.mutable)
                    .map(|obj| obj.id),
            );

        self.reference_gas_price + hotness as u64
    }
```

BTW, the function is never used? Where is it planned to be called?

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
+                if let Some(info) = congestion_info_map.get_mut(object) {
+                    info.hotness += hotness_adjustment;
+                }
+            }
         }
 
         for (gas_price, objects) in clearing_txs_data {
-            objects.iter().for_each(|object_id| {
-                match congestion_info_map.entry(*object_id) {
+            for object in objects {
```
roman1e2f5p8s:
```suggestion
            objects.iter().for_each(|object_id| {
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
+                // We only record clearing prices if the object has observed cancellations
+                // recently
+                match congestion_info_map.entry(*object) {
```
roman1e2f5p8s:
```suggestion
                match congestion_info_map.entry(*object_id) {
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
                     Entry::Occupied(entry) => {
                         entry.into_mut().update_for_clearing_tx(time, *gas_price);
                     }
                     Entry::Vacant(entry) => {
-                        // We only record clearing prices if the object has experienced
-                        // congestion recently.
-                        if let Some(prev) = self.get_congestion_info(*object_id) {
+                        if let Some(prev) = self.get_congestion_info(*object) {
```
roman1e2f5p8s:
```suggestion
                        if let Some(prev) = self.get_congestion_info(*object_id) {
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
                                 highest_congestion_gas_price: prev.highest_congestion_gas_price,
                                 latest_clearing_time: Some(time),
                                 lowest_clearing_gas_price: Some(*gas_price),
+                                hotness: prev.hotness,
                             });
                         }
                     }
                 }
-            });
+            }
```
roman1e2f5p8s:
```suggestion
            });
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
-                        highest_congestion_gas_price: *gas_price,
-                        latest_clearing_time: None,
-                        lowest_clearing_gas_price: None,
-                    });
-            });
+        congestion_txs_data: &[(u64, Vec<ObjectID>, u64)],
+        clearing_txs_data: &[(u64, Vec<ObjectID>)],
+    ) -> HashMap<ObjectID, CongestionInfo> {
+        let mut congestion_info_map: HashMap<ObjectID, CongestionInfo> = HashMap::new();
+        let mut object_id_per_tx: Vec<ObjectID> = Vec::new();
```
roman1e2f5p8s:
```suggestion
    ) -> CongestionInfoMap {
        let mut congestion_info_map = CongestionInfoMap::new();
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -169,19 +215,60 @@ impl CongestionTracker {
...
+                    .filter(|id| id.mutable)
+                    .map(|id| id.id),
+            )
+            .unwrap_or(0.0);
+
+        Some(self.reference_gas_price + hotness as u64)
+    }
+
+    /// Returns a map of all objects and their hotness values.
+    pub fn get_all_hotness(&self) -> HashMap<ObjectID, f64> {
```
roman1e2f5p8s, line 236: The function is not called anywhere, I think it should be removed.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -222,72 +309,143 @@ impl CongestionTracker {
...
+            // Adjust hotness based on the sum of hotness of objects in the transaction
+            // minus the gas price feedback
+            for object in &object_id_per_tx {
+                let hotness_adjustment = hotness_per_tx - (*gas_price_feedback as f64)
+                    + (self.reference_gas_price as f64);
+
+                if let Some(info) = congestion_info_map.get_mut(object) {
+                    info.hotness += hotness_adjustment;
+                }
+            }
```
roman1e2f5p8s:
```suggestion
            let hotness_per_tx = self
                .get_total_hotness_for_objects(objects.iter().cloned());

            objects.iter().for_each(|object_id| {
                match congestion_info_map.entry(*object_id) {
                    Entry::Occupied(entry) => {
                        entry.into_mut().update_for_cancellation(time, *gas_price);
                    }
                    Entry::Vacant(entry) => {
                        entry.insert(CongestionInfo {
                            latest_congestion_time: time,
                            highest_congestion_gas_price: *gas_price,
                            latest_clearing_time: None,
                            lowest_clearing_gas_price: None,
                            hotness: 0.0,
                        });
                    }
                }

                // Adjust hotness based on the sum of hotness of objects in the transaction
                // plus the difference between reference gas price and feedback gas price.
                let hotness_adjustment = hotness_per_tx - (*gas_price_feedback as f64)
                    + (self.reference_gas_price as f64);
                congestion_info_map
                    .entry(*object_id)
                    .and_modify(|info| info.hotness += hotness_adjustment);
            });
```

I am wondering if the hotness adjustment is computed correctly. Shouldn't it be `hotness_per_tx + gas_price_feedback - reference_gas_price`?

roman1e2f5p8s (changes requested):
- Some for loops can be rewritten using iterators.
- A few functions are not currently used and probably should be removed.
- The new type for congestion/clearing txs data is highly recommended.

Left more details in the comments and suggestions.

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: The first idea was to use `number_cleared_transactions` to have a more sophisticated way of reducing hotness but the current naive approach seems to be good enough

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: This function is supposed to be replacing Sui's estimation (`get_suggested_gas_price`) in `dry_exec_transaction_impl`

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: It's mostly used for debugging at this stage. I'd not remove right now as it's useful when running experiments.

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: Nope. The loss function we use to adjust the hotness is given by the difference between our suggested price `p'` and the feedback `q`. The suggested price is trivially computed as the sum of the hotnesses of all objects in the tx plus the reference gas price, that gives: `p' - q = (sum(hotness) + reference_gas_price) - q`

crates/iota-core/src/congestion_tracker.rs
roman1e2f5p8s: Okay, then it should be invoked there? Because `dry_exec_transaction_impl` is still using Sui's way.

crates/iota-core/src/congestion_tracker.rs
roman1e2f5p8s: Why not creating another branch out of this one, and keep there functionality you need for running experiment?
If you don't remove this, it will be merged and stay unused on develop.

roman1e2f5p8s (approved):
LGTM. I would only add release notes for Nodes in the PR description, something like "Replaced the upstream's logic for recommended suggested gas price calculations with more advanced calculations based on online gradient descent." But if you don't want to add any release notes, we should then remove the whole release notes section, to make the CI happy.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,8 +85,35 @@ impl CongestionInfo {
...
+            if !is_new {
+                let new_hotness = self.hotness
+                    - new.hotness * HOTNESS_ADJUSTMENT_FACTOR
+                        / number_congested_transactions as f64;
+                // Hotness decrease is capped by HOTNESS_DECAY_FACTOR.
+                self.hotness = new_hotness.max(self.hotness / HOTNESS_DECAY_FACTOR);
+            } else {
+                self.hotness =
+                    -new.hotness * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions as f64;
+                // Ensure hotness is non-negative
```
muXxer: shouldn't that comment be in the next line?

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -64,8 +85,35 @@ impl CongestionInfo {
...
+        is_new: bool,
+    ) {
+        if number_congested_transactions > 0 {
+            // Update object i's hotness based on the congestion events following the
+            // formula: hotness_i -= SUM(tx)[SUM(j in tx)(hotness_j - gas_price_feedback)]
+            // * HOTNESS_ADJUSTMENT_FACTOR / number_congested_transactions.
+            if !is_new {
+                let new_hotness = self.hotness
+                    - new.hotness * HOTNESS_ADJUSTMENT_FACTOR
+                        / number_congested_transactions as f64;
```
muXxer: Can't we calculate that term before the "if", and reuse it in both cases?
Idk, maybe it makes it a bit more readable? You can ignore my comment, just a suggestion.

muXxer (approved):

muXxer (approved):

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -18,13 +18,30 @@ use crate::execution_cache::TransactionCacheRead;
...
-/// congested) shared objects.
-type TransactionGasPriceMutSharedObjectsPair = (u64, Vec<ObjectID>);
+/// Threshold for hotness below which an object is considered cold.
+/// Values should be > 0.0. If HOTNESS_CUTOFF = 0.0, then no pruning will
+/// happen.
+const HOTNESS_CUTOFF: f64 = 1.0;
+
+/// Controls how quickly congestion tracker updates object hotness.
+/// Values should be > 0.0. If HOTNESS_CUTOFF = 0.0, then no pruning will
+/// happen.
```
VorobyevIlya: Repetition of lines above. Is it supposed to be like that?

VorobyevIlya (approved):

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: Good catch! I'll fix it

Thoralf-M (approved):

PR #7951, opened 2025-07-23
chore(*): Unpin tokio

DaughterOfMars:
## Description 

Unpins the `tokio` version.

NOTE: This means that simtests will be broken if the lock file tokio version is ever updated because of a new patch version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -348,7 +348,7 @@ test-fuzz = "3.0.4"
 thiserror = "1.0.40"
 # tokio is defined at a specific version to support patching with iota-sim
 # Do not upgrade without updating https://github.com/iotaledger/iota-sim first
-tokio = "=1.46.1"
+tokio = "1.46.1"
 tokio-rustls = { version = "0.26", default-features = false, features = ["tls12", "ring"] }
 tokio-stream = { version = "0.1.14", features = ["sync", "net"] }
 tokio-util = "0.7.10"


TheMrAI:
Could you give some more context why the unpinning is necessary given the problem it may cause?

DaughterOfMars:
> Could you give some more context why the unpinning is necessary given the problem it may cause?

It's not necessary, but it does allow library users to be more flexible with their tokio version. Since they don't care about simtests, it doesn't matter if they resolve a higher patch version.

Cargo.toml
```diff
@@ -348,7 +348,7 @@ test-fuzz = "3.0.4"
 thiserror = "1.0.40"
 # tokio is defined at a specific version to support patching with iota-sim
 # Do not upgrade without updating https://github.com/iotaledger/iota-sim first
-tokio = "=1.46.1"
+tokio = "1.46.1"
```
TheMrAI, line 351: Hmm, hmm. Does this mean that when any crate using tokio is updated to a newer version using a newer tokio version which technically could be compatible (from cargo's perspective), then they can potentially force this version to be elevated as well and CI breaks?
In which case from our side the only way to resolve the issue is to pin the version again or downgrade the offending library?
Just thinking aloud mostly here.

DaughterOfMars: No. It just means that if we accidentally update the version in the lock file (this can only be the patch version because tokio is past 1.0.0) then our CI will break in that PR. It's not very serious but it's annoying.

muXxer (approved):

thibault-martinez (approved):

TheMrAI (approved):
Then it is ok.

PR #7969, opened 2025-07-24
upstream(iota-indexer): chunk to avoid PG parameter limit

### LINKED ISSUE #7889: upstream(indexer): chunk to avoid PG parameter limit
## Description

[sui#19754](https://www.github.com/MystenLabs/sui/pull/19754)
`{"side-effects": "medium", "type": "fix/critical"}`

sergiupopescu199:
# Description of change

This PR adds changes from upstream, the actual interested code is already present in latest `develop`, for a detailed  explanation on why we have already this change please see the issue comment:
- https://github.com/iotaledger/iota/issues/7889#issuecomment-3112528449

- Nevertheless int this PR we add the two missing test cases which verifies that events and tx indices does not hit the Postgres limit.
- The `persist_tx_indices_chunk` was refactored to use the already present `persist_chunk_into_table!` macro to store chunked data to db, as `persist_event_indices_chunk` already does, this refactor removes duplicated code making it more readable and maintainable.

fixes #7889

- run new added tests
```shell
cargo test -p iota-indexer --features pg_integration -- --test-threads 1
```
- synced iota-indexer with `devnet` data to check if genesis checkpoint was synced correctly and also checked that reliability requirements  were not exceeded. No regressions found during testing.
- synced iota-indexer with `testnet` data and also restarted it without resetting the database.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/types.rs
+++ b/crates/iota-indexer/src/types.rs
@@ -416,6 +416,27 @@ impl EventIndex {
     }
 }
 
+#[cfg(feature = "pg_integration")]
+impl EventIndex {
+    /// Generate a random event index for testing purposes.
+    pub fn random() -> Self {
+        use rand::Rng;
+
+        let mut rng = rand::thread_rng();
+        EventIndex {
+            tx_sequence_number: rng.gen(),
+            event_sequence_number: rng.gen(),
+            sender: IotaAddress::random_for_testing_only(),
+            emit_package: ObjectID::random(),
+            emit_module: rng.gen::<u64>().to_string(),
+            type_package: ObjectID::random(),
+            type_module: rng.gen::<u64>().to_string(),
+            type_name: rng.gen::<u64>().to_string(),
+            type_instantiation: rng.gen::<u64>().to_string(),
+        }
+    }
+}
+
 #[derive(Debug, Copy, Clone)]
 pub enum OwnerType {
     Immutable = 0,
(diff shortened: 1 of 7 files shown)


crates/iota-indexer/src/types.rs
```diff
@@ -543,6 +564,64 @@ pub struct TxIndex {
     pub move_calls: Vec<(ObjectID, String, String)>,
 }
 
+#[cfg(feature = "pg_integration")]
+impl TxIndex {
+    const MAX_OBJECTS: usize = 1000;
```
samuel-rufi: nit: Since these constants are only used within the `random()` function, should we maybe move them there? As they don't seem to have semantic meaning for `TxIndex` impl itself.

samuel-rufi (approved):
Hey @sergiupopescu199, looks good to me! 🌞

crates/iota-indexer/src/types.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7969/commits/97fe091bf8eb48280971c5c6ea5ce49ef2a5169c

crates/iota-indexer/src/types.rs
```diff
@@ -416,6 +416,27 @@ impl EventIndex {
     }
 }
 
+#[cfg(feature = "pg_integration")]
```
kodemartin: I would suggest exposing this for all tests:

```suggestion
#[cfg(test)]
```

crates/iota-indexer/src/types.rs
```diff
@@ -543,6 +564,60 @@ pub struct TxIndex {
     pub move_calls: Vec<(ObjectID, String, String)>,
 }
 
+#[cfg(feature = "pg_integration")]
```
kodemartin: Same as above
```suggestion
#[cfg(test)]
```

kodemartin (commented):
lgtm with a couple of suggestions.

kodemartin (approved):

crates/iota-indexer/src/types.rs
sergiupopescu199: `#[cfg(test)]` is only enabled on unit tests, tests marked as `#[test]` under the `src` directory.

Tests under the `tests` directory are treated as integration tests, and they’re treated as independent crates that have the library as a dependency. Items with `#[cfg(test)]` will not be compiled.

As an alternative, I could add this:
```rust
// compiles when either condition is true
#[cfg(any(test, feature = "pg_integration"))]
```

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/7969/commits/5f601e42b9603eab8de7de10f29c082f5f622b75

crates/iota-indexer/src/types.rs
sergiupopescu199: based on comment: https://github.com/iotaledger/iota/pull/7969#discussion_r2231355414
fixed: https://github.com/iotaledger/iota/pull/7969/commits/5f601e42b9603eab8de7de10f29c082f5f622b75

PR #7995, opened 2025-07-25
upstream(indexer): add epochs.first_tx_sequence_number

### LINKED ISSUE #7364: upstream(indexer): Add first_tx_sequence_number to epochs table to decouple from checkpoints table
## Description

Following #7313 

https://github.com/MystenLabs/sui/pull/19773

kodemartin:
# Description of change

This patch follows the respective [upstream change](https://www.github.com/MystenLabs/sui/pull/19773) in adding a new column `first_tx_sequence_number` in the `epochs` table to decouple the calculation of `epoch_total_transactions` from the `checkpoints` table.

But in doing so we add a couple of additional changes:

1. The `first_tx_sequence_number` is backfilled during the migration itself and thus it is made NOT NULL. This is feasible because the values can be derived from the `epoch_total_transactions` and the `checkpoints` table. Furthermore, since the size of the `epochs` table is small, we don't expect this to affect deployment.
2. We rename `epoch_total_transactions` to `network_total_transactions` and recalculate its values to hold the total network transactions by the end of the epoch. `epoch_total_transactions` can be then calculated through `network_total_transactions - first_tx_sequence_number`. This allows decoupling completely the extraction operations (what is also referred to as the reading side) from the loading/persisting operations.

    So far we were indexing the `epoch_total_transactions` which is not available in the checkpoint data at the start or the end of the epoch. So we were making a db read to fetch the number of total network transactions from the db. `network_total_transactions` is available at epoch end, and `first_tx_sequence_number` is available at epoch start, so it is more plausible to store these.

In the process a flaky test has been updated to ensure successful test run.

Fixes #7364 

### Release Notes

- [x] Indexer: Adds a new column `epochs.first_tx_sequence_number` and repurposes `epoch.epoch_total_transactions` to `epoch.network_total_transactions`. This decouples epoch info from the checkpoints table, and eliminates db queries while extracting the data to persist. On the web interfaces (JSON-RPC, GraphQL) the view on the epoch data remains intact.

### DIFF
--- a/crates/iota-graphql-e2e-tests/tests/epoch/epoch_start_to_end.move
+++ b/crates/iota-graphql-e2e-tests/tests/epoch/epoch_start_to_end.move
@@ -1,0 +1,159 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --protocol-version 9 --simulator --accounts C
+
+//# create-checkpoint
+
+//# advance-epoch
+
+//# programmable --sender C --inputs 10000000000 @C
+//> SplitCoins(Gas, [Input(0)]);
+//> TransferObjects([Result(0)], Input(1));
+
+//# programmable --sender C --inputs 5000000000 @C
+//> SplitCoins(Gas, [Input(0)]);
+//> TransferObjects([Result(0)], Input(1));
+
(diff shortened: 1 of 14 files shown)


sergiupopescu199 (approved):
lgtm! 🐳

samuel-rufi (approved):
Looks fantastic @kodemartin. 
Great that we can perform the backfill as part of the migration, fortunately the epochs table allows for that.

PR #7996, opened 2025-07-25
feat(sdk): Add `IotaClientConfig` consistency checks

### LINKED ISSUE #5523: feat(CLI): proper `iota.keystore` existence check and configured `active_address` validation?
## Description

Should we add an existence check for the keystore file that is referenced in the config.yaml. 

Also ... do we think it would be a good idea to check the address stored in `active_address` in the `config.yaml` against the keystore to always ensure the correctness and consistency of such configured values.

Currently, if the config points to an invalid keystore path depending on what the user calls, one is not made aware of that until a command is invoked, that actually requires access to the keystore.

E.g. calls like 
```
iota client --client.config client.yaml active-address
```

then return `None` despite `config.yaml` saying otherwise.

If the `keystore` is not available due to a temporarily broken symlink (user who does point it to a separate device like a usb stick or whatever), the user might unnecessarily create a new keystore instead of fixing the true underlying issue (properly mounting the usb stick).

IMHO we should always start with a valid, consistent and functioning configuration.

Alex6323:
# Description of change

This PR introduces some consistency checks for the `IotaClientConfig` to make sure:
* that the active address (if set) is contained in the keystore
* that the active env (if set) is part of the envs list

fixes #5523 

I tested each one of the new features by modifying `client.yaml` manually and observe its results/errors.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-sdk/src/wallet_context.rs
```diff
@@ -55,6 +55,24 @@ impl WalletContext {
...
+            ensure!(
+                addresses.contains(active_address),
+                "active address does not belong to configured keystore"
+            );
+        }
+
+        if let Some(active_env) = &config.active_env {
+            ensure!(
+                config.get_env(active_env).is_some(),
+                "unknown active environment"
```
Thoralf-M: Wondering if we should add the path to the config folder in the message or give some advice how this can be resolved

Alex6323: yeah, why not.. I'll try to come up with something reasonable.

Alex6323: https://github.com/iotaledger/iota/pull/7996/commits/4d32a3926a54e7661e94fd1ce38819f84c6547ef

Thoralf-M (approved):

DaughterOfMars (approved):

Issue #6149, opened 2025-03-31
Port upstream node team related changes [mainnet-v1.40.3..mainnet-v1.41.2)


muXxer:

muXxer:
Closed by #8008

PR #7998, opened 2025-07-28
upstream(core): Remove the direct dependency from `iota-core` to `iota-move-build`

jkrvivian:
# Description of change

- Upstream range: [v1.40.3, v1.41.2)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/e54c57db0ca21a7546ed7d14be57a35492f9fc4b
- Description:
Previously, `iota-move-build` is linked into `iota-core` as a normal
dependency. The only reason for this is that some of the `iota-core`
testing utilities depend on `iota-move-build` and some other crates
depend on the `iota-core` testing utilities. This PR cleans this up a bit
so that the testing infrastructure that depends on `iota-move-build` is
not re-exported, and therefore `iota-move-build` only needs to be a
dev-dependency of `iota-core`.

Specifically, the `iota::test_utils` code that is only used
inside of `iota-core` was split off into `iota::unit_test_utils`, and gated
this file behind `#[cfg(test)]`. The tiny bit of shared
testing code that uses `iota-move-build` was moved into
`iota_move_build::test_utils`.

Similarly,
`iota_core::authority::authority_test_utils` into
`iota_core::authority::auth_unit_test_utils`
were split.

Part of #6149   

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -195,6 +195,10 @@ mod batch_verification_tests;
 #[path = "unit_tests/coin_deny_list_tests.rs"]
 mod coin_deny_list_tests;
 
+#[cfg(test)]
+#[path = "unit_tests/auth_unit_test_utils.rs"]
+pub mod auth_unit_test_utils;
+
 pub mod authority_test_utils;
 
 pub mod authority_per_epoch_store;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5721,6 +5721,7 @@ dependencies = [
  "iota-json",
  "iota-json-rpc-types",
  "iota-keys",
+ "iota-move-build",
  "iota-sdk 1.4.0-alpha",
  "iota-swarm",
  "iota-swarm-config",
--- a/crates/iota-cluster-test/Cargo.toml
+++ b/crates/iota-cluster-test/Cargo.toml
@@ -34,6 +34,7 @@ iota-indexer.workspace = true
 iota-json.workspace = true
 iota-json-rpc-types.workspace = true
 iota-keys.workspace = true
+iota-move-build.workspace = true
 iota-sdk.workspace = true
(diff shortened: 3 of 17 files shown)


muXxer (approved):

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -2047,7 +2046,7 @@ async fn diagnose_split_brain(
         checkpoint_seq = local_summary.sequence_number,
         "Running split brain diagnostics..."
     );
-    let time = Utc::now();
+    let time = SystemTime::now();
```
TheMrAI: What is this sneaky change here eh?

valeriyr (approved):

crates/iota-core/src/authority.rs
```diff
@@ -2701,8 +2704,11 @@ impl AuthorityState {
     }
 
     pub fn unixtime_now_ms() -> u64 {
-        let ts_ms = Utc::now().timestamp_millis();
-        u64::try_from(ts_ms).expect("Travelling in time machine")
+        let now = SystemTime::now()
```
TheMrAI: We are aware that SystemTime is a non-monotonic clock correct?

TheMrAI (approved):
From vm-lang persective it looks good.

crates/iota-core/src/checkpoints/mod.rs
muXxer: Sorry, that was a missing rebase, so another PR was accidentally contained in the changes.

crates/iota-core/src/authority.rs
muXxer: Sorry, that was a missing rebase, so another PR was accidentally contained in the changes.

Regarding your question:
`SystemTime` with `duration_since(UNIX_EPOCH)` is time zone neutral and deterministic.

PR #8003, opened 2025-07-28
feat(iota): simplify iota client setup

### LINKED ISSUE #7968: CLI: remove Config file doesn't exist prompt
## Description

First time one uses the CLI it will print
`Config file ["/home/t/.iota/iota_config/client.yaml"] doesn't exist, do you want to connect to an IOTA Full node server [y/N]?`
`n` case here is useless as it will just error with `no environment exists for the client`
So the prompte should be removed/merged with the next one to select a default network or provide an own URL

## Motivation

Less steps in the setup

Thoralf-M:
# Description of change

Simplify iota client setup by removing an obsolete prompt that would just error with `no environment exists for the client` if one doesn't write `y`

Fixes #7968 

`cargo run client` now directly displays `Select a default network [mainnet|testnet|devnet|localnet], or enter a custom IOTA full node server URL (defaults to testnet if not specified):`

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Removed obsolete prompt from `iota client` if no `client.yaml` existed
- [ ] Rust SDK:
- [ ] REST API:


crates/iota/src/iota_commands.rs
```diff
@@ -1192,42 +1192,32 @@ fn prompt_for_environment(
             print!(
                 "Creating config file [{wallet_conf_path:?}] with default (Testnet) full node server and ed25519 key scheme."
             );
-        } else {
-            print!(
-                "Config file [{wallet_conf_path:?}] doesn't exist! Do you want to connect to an IOTA full node server? [y/N]: "
-            );
+            return Ok(IotaEnv::testnet());
```
DaughterOfMars: Why return? Shouldn't we still allow choosing a network? I mean I guess you're taking "acccept defaults" to mean all defaults but this feels not so useful. 🤷‍♀️

Thoralf-M: That's the same behavior as before and the point of the accept default, if you want to select a network you don't accept the default (you can still select a default if you want when you select manually)

Alex6323: An aspect that disturbs me a bit is that the print seems badly placed here, because it's not doing the things it says it does, i.e. it's not creating a config file at all here, neither pre-selecting a scheme. It seems this should be moved to the caller tbh, because here it's just returning the default testnet env.

And another thing is that we probably should return early in line 1189, so we get rid of nesting most of the code in this function.

crates/iota/src/iota_commands.rs
```diff
@@ -1192,42 +1192,32 @@ fn prompt_for_environment(
...
-                        } else {
-                            alias
-                        };
-                        Ok(IotaEnv::new(alias, url))
+
+        print!(
+            "Select a default network [mainnet|testnet|devnet|localnet], or enter a custom IOTA full node server URL (defaults to testnet if not specified): "
+        );
+        let url = read_line()?;
+        match url.trim() {
```
Alex6323: I didnt think of it in my PR but maybe also lowercase the input here.

crates/iota/src/iota_commands.rs
```diff
@@ -1192,42 +1192,32 @@ fn prompt_for_environment(
...
-                            "custom".to_string()
-                        } else {
-                            alias
-                        };
-                        Ok(IotaEnv::new(alias, url))
+
+        print!(
+            "Select a default network [mainnet|testnet|devnet|localnet], or enter a custom IOTA full node server URL (defaults to testnet if not specified): "
+        );
+        let url = read_line()?;
```
Alex6323: nit: can we rename `url` to `input`?

docs/content/developer/getting-started/connect.mdx
```diff
@@ -63,18 +63,10 @@ iota client
 
 If this is the first time you run the command, you will be prompted to configure the client.
 
-#### 1. Create `client.yaml`
+#### 1. Create `client.yaml` with selected network
```
Alex6323:
```suggestion
#### 1. Create a client config (`client.yaml`)
```

Thoralf-M: But the config is not only for a single network :trollface: Only a single network is active

crates/iota/src/iota_commands.rs
Thoralf-M: Sure

crates/iota/src/iota_commands.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/8003/commits/1cc1796a9db91b9db5a2cc44c28b3e3c3d943cb5

crates/iota/src/iota_commands.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/8003/commits/1cc1796a9db91b9db5a2cc44c28b3e3c3d943cb5

crates/iota/src/iota_commands.rs
Thoralf-M: Problem with the print is that you would then also need to have the same checks in another place, not helping with the readability
Did the early return here https://github.com/iotaledger/iota/pull/8003/commits/1cc1796a9db91b9db5a2cc44c28b3e3c3d943cb5

docs/content/developer/getting-started/connect.mdx
Thoralf-M: Maybe just 
```
#### 1. Create a client config (`client.yaml`)
```
?

Alex6323: fine with me!

Alex6323: I changed the suggestion!

Alex6323 (approved):

vivekjain23 (approved):

DaughterOfMars (approved):

PR #8018, opened 2025-07-28
upstream(node): Fix race condition for transaction cache

jkrvivian:
# Description of change

- Upstream range: [v1.40.3, v1.41.2)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/11e5ec9bbed9c9baed97511b5b87aa16f5702e6e
- Description:
Bug was (likely) as follows:

      Reader thread               | Writer thread (state sync)

                                    invalidate tickets
      get ticket
      miss cache
      read database
      insert to cache               insert to cache (racing)
      ticket is valid
      panic
                                    write to database

The writer thread must insert to db first so that the db read
cannot find an old value while holding a valid ticket

Part of #6149   

### DIFF
--- a/crates/iota-core/src/execution_cache/cache_types.rs
+++ b/crates/iota-core/src/execution_cache/cache_types.rs
@@ -295,6 +295,8 @@ where
             check_ticket()?;
 
             // Ticket expiry should make this assert impossible.
+            // Note: value and entry versions can be equal, which is allowed for genesis
+            // initialization
             if entry.is_newer_than(&value) {
                 debug_fatal!("entry is newer than value");
             } else {
--- a/crates/iota-core/src/execution_cache/cache_types.rs
+++ b/crates/iota-core/src/execution_cache/cache_types.rs
@@ -9,6 +9,7 @@ use std::{
     sync::{Arc, atomic::AtomicU64},
 };
 
+use iota_common::debug_fatal;
 use iota_types::base_types::SequenceNumber;
 use moka::sync::Cache as MokaCache;
 use parking_lot::Mutex;
@@ -293,10 +294,12 @@ where
             let mut entry = entry.value().lock();
             check_ticket()?;
 
-            // Ticket expiry makes this assert impossible.
-            // TODO: relax to debug_assert?
-            assert!(!entry.is_newer_than(&value), "entry is newer than value");
-            *entry = value;
+            // Ticket expiry should make this assert impossible.
+            if entry.is_newer_than(&value) {
+                debug_fatal!("entry is newer than value");
+            } else {
+                *entry = value;
+            }
         }
 
         Ok(())
--- a/crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
+++ b/crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
@@ -1329,6 +1329,11 @@ async fn latest_object_cache_race_test() {
 }
 
 #[tokio::test]
+// This test verifies that concurrent transaction insertions and reads work
+// correctly without race conditions. It specifically tests the fix that ensures
+// store operations happen before cache operations. It ensures atomicity by
+// writing to the persistent store first, then updating the cache, so readers
+// never see stale or inconsistent data.
 async fn test_transaction_cache_race() {
     telemetry_subscribers::init_for_testing();
     let mut s = Scenario::new(None, Arc::new(AtomicU32::new(0))).await;
@@ -1369,8 +1374,6 @@ async fn test_transaction_cache_race() {
     };
 
     let t2 = {
-        let txns = txns.clone();
-        let cache = cache.clone();
         let barrier = barrier.clone();
         std::thread::spawn(move || {
             for (tx, _) in txns {
--- a/crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
+++ b/crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
@@ -1329,6 +1329,62 @@ async fn latest_object_cache_race_test() {
 }
 
 #[tokio::test]
+async fn test_transaction_cache_race() {
+    telemetry_subscribers::init_for_testing();
+    let mut s = Scenario::new(None, Arc::new(AtomicU32::new(0))).await;
+    let cache = s.cache.clone();
+    let mut txns = Vec::new();
+
+    for i in 0..1000 {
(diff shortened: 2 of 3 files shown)


muXxer (approved):

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -2232,6 +2232,12 @@ impl StateSyncAPI for WritebackCache {
         transaction: &VerifiedTransaction,
         transaction_effects: &TransactionEffects,
     ) -> IotaResult {
+        self.store
+            .insert_transaction_and_effects(transaction, transaction_effects)
+            .map_err(IotaError::from)?;
```
semenov-vladyslav: Do we need `.map_err(IotaError::from)` there? IIRC, the `?` operator should do that already.

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -2248,15 +2254,17 @@ impl StateSyncAPI for WritebackCache {
...
+        Ok(())
     }
 
     fn try_multi_insert_transaction_and_effects(
         &self,
         transactions_and_effects: &[VerifiedExecutionData],
     ) -> IotaResult {
+        self.store
+            .multi_insert_transaction_and_effects(transactions_and_effects.iter())
+            .map_err(IotaError::from)?;
```
semenov-vladyslav: Same here.

crates/iota-core/src/execution_cache/cache_types.rs
```diff
@@ -293,10 +294,12 @@ where
...
 
-            // Ticket expiry makes this assert impossible.
-            // TODO: relax to debug_assert?
-            assert!(!entry.is_newer_than(&value), "entry is newer than value");
-            *entry = value;
+            // Ticket expiry should make this assert impossible.
+            if entry.is_newer_than(&value) {
+                debug_fatal!("entry is newer than value");
+            } else {
+                *entry = value;
```
semenov-vladyslav, line 303: Now this happens when "entry is not newer than value". Should it be when "value is newer than entry"?

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -1328,6 +1328,62 @@ async fn latest_object_cache_race_test() {
...
+        s.do_tx().await;
+
+        let outputs = s.take_outputs();
+        let tx = (*outputs.transaction).clone();
+        let effects = outputs.effects.clone();
+
+        txns.push((tx, effects));
+    }
+
+    let barrier = Arc::new(std::sync::Barrier::new(2));
```
semenov-vladyslav, line 1355: `Arc` is not needed here.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -1328,6 +1328,62 @@ async fn latest_object_cache_race_test() {
...
+                    )]);
+                }
+            }
+        })
+    };
+
+    let t2 = {
+        let txns = txns.clone();
+        let cache = cache.clone();
+        let barrier = barrier.clone();
```
semenov-vladyslav, line 1377: No need for these `clones`.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -1328,6 +1328,62 @@ async fn latest_object_cache_race_test() {
     invalidator.join().unwrap();
 }
 
+#[tokio::test]
+async fn test_transaction_cache_race() {
```
semenov-vladyslav, line 1337: Does this test trigger "entry is newer than value" assert? Did this test fail before this commit? I'd leave an explainer comment.

semenov-vladyslav (changes requested):
Looks good but a few minor things.

crates/iota-core/src/execution_cache/cache_types.rs
jkrvivian: Do you mean to change the logic into the other way around?

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
jkrvivian: Why is it not needed? It's used in different threads.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
jkrvivian: The fix is moving store insertion before cache insertions. "entry is newer than value" doesn't relate to the bug 😄

crates/iota-core/src/execution_cache/cache_types.rs
semenov-vladyslav: I am just asking what should be the right logic. If adding objects with the same version is allowed (why? would that be a duplicate?), then it's fine as it is, otherwise I think we should change it.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
semenov-vladyslav: Right, thought it would be sufficient to only use a reference since Barrier's `wait` method requires reference. But the thread may outlive the barrier object itself. In docs to Barrier they use thread::scope to overcome that.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
semenov-vladyslav: I thought that wrong cache insertions could trigger "entry is newer than value" error. Otherwise what issue fixing cache insertions solve?

jkrvivian: Yes, you're right!

crates/iota-core/src/execution_cache/cache_types.rs
jkrvivian: Explained in https://github.com/iotaledger/iota/pull/8018/commits/f159974ece0323d8ef72a33934b8d8dabee39b4e

semenov-vladyslav (approved):
There can in fact be two objects with the same version inserted into cache.

Issue #6151, opened 2025-03-31
Port upstream node team related changes [mainnet-v1.42.2..mainnet-v1.43.1)


muXxer:

muXxer:
Merged with #8073

PR #8055, opened 2025-07-31
upstream(core): Make transaction execution synchronous

jkrvivian:
# Description of change

- Upstream range: [v1.42.2, v1.43.1)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/0c7c26ed8907fea4f88f010eda0348adc212fbb1
- Description:

Transaction execution doesn't need to be async, so this PR removes that.
This won't have any significant performance impacts, because most of the
previous await points were probably "fake" in that they were unlikely to actually yield execution.

We made some more functions than upstream "sync", because we were touching them anyways.
Also, in the json-rpc, we made the code a bit more efficient and fixed some wrong metrics and different behavior if the cache was force disabled (filter was not applied).

Part of #6151    

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1178,7 +1178,7 @@ impl AuthorityState {
     ///
     /// Should only be called within iota-core.
     #[instrument(level = "trace", skip_all)]
-    pub async fn try_execute_immediately(
+    pub fn try_execute_immediately(
         &self,
         certificate: &VerifiedExecutableTransaction,
         expected_effects_digest: Option<TransactionEffectsDigest>,
@@ -1262,13 +1262,11 @@ impl AuthorityState {
         certificate: &VerifiedCertificate,
     ) -> IotaResult<(VerifiedSignedTransactionEffects, Option<ExecutionError>)> {
         let epoch_store = self.epoch_store_for_testing();
-        let (effects, execution_error_opt) = self
-            .try_execute_immediately(
-                &VerifiedExecutableTransaction::new_from_certificate(certificate.clone()),
-                None,
-                &epoch_store,
-            )
-            .await?;
+        let (effects, execution_error_opt) = self.try_execute_immediately(
+            &VerifiedExecutableTransaction::new_from_certificate(certificate.clone()),
+            None,
+            &epoch_store,
+        )?;
         let signed_effects = self.sign_effects(effects, &epoch_store)?;
         Ok((signed_effects, execution_error_opt))
     }
@@ -1693,7 +1691,7 @@ impl AuthorityState {
         self.prepare_certificate(&execution_guard, certificate, input_objects, epoch_store)
     }
 
-    pub async fn dry_exec_transaction(
+    pub fn dry_exec_transaction(
         &self,
         transaction: TransactionData,
         transaction_digest: TransactionDigest,
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1257,7 +1257,7 @@ impl AuthorityState {
     /// Test only wrapper for `try_execute_immediately()` above, useful for
     /// checking errors if the pre-conditions are not satisfied, and
     /// executing change epoch transactions.
-    pub async fn try_execute_for_test(
+    pub fn try_execute_for_test(
         &self,
         certificate: &VerifiedCertificate,
     ) -> IotaResult<(VerifiedSignedTransactionEffects, Option<ExecutionError>)> {
@@ -1272,12 +1272,11 @@ impl AuthorityState {
     }
 
     /// Non-fallible version of `try_execute_for_test()`.
-    pub async fn execute_for_test(
+    pub fn execute_for_test(
         &self,
         certificate: &VerifiedCertificate,
     ) -> (VerifiedSignedTransactionEffects, Option<ExecutionError>) {
         self.try_execute_for_test(certificate)
-            .await
             .expect("try_execute_for_test should not fail")
     }
 
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -861,7 +861,7 @@ impl AuthorityState {
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<VerifiedSignedTransaction> {
         // Ensure that validator cannot reconfigure while we are signing the tx
-        let _execution_lock = self.execution_lock_for_signing().await;
+        let _execution_lock = self.execution_lock_for_signing();
 
         let tx_digest = transaction.digest();
         let tx_data = transaction.data().transaction_data();
@@ -919,9 +919,11 @@ impl AuthorityState {
         // The call to self.set_transaction_lock checks the lock is not conflicting,
         // and returns ConflictingTransaction error in case there is a lock on a
         // different existing transaction.
-        self.get_cache_writer()
-            .try_acquire_transaction_locks(epoch_store, &owned_objects, signed_transaction.clone())
-            .await?;
+        self.get_cache_writer().try_acquire_transaction_locks(
+            epoch_store,
+            &owned_objects,
+            signed_transaction.clone(),
+        )?;
 
         Ok(signed_transaction)
     }
@@ -1188,7 +1190,7 @@ impl AuthorityState {
         let tx_digest = certificate.digest();
 
         // Acquire a lock to prevent concurrent executions of the same transaction.
-        let tx_guard = epoch_store.acquire_tx_guard(certificate).await?;
+        let tx_guard = epoch_store.acquire_tx_guard(certificate)?;
 
         // The cert could have been processed by a concurrent attempt of the same cert,
         // so check if the effects have already been written.
@@ -1225,7 +1227,6 @@ impl AuthorityState {
             expected_effects_digest,
             epoch_store,
         )
-        .await
         .tap_err(|e| info!(?tx_digest, "process_certificate failed: {e}"))
         .tap_ok(
             |(fx, _)| debug!(?tx_digest, fx_digest=?fx.digest(), "process_certificate succeeded"),
@@ -1331,7 +1332,7 @@ impl AuthorityState {
     }
 
     #[instrument(level = "trace", skip_all)]
-    pub(crate) async fn process_certificate(
+    pub(crate) fn process_certificate(
         &self,
         tx_guard: CertTxGuard,
         certificate: &VerifiedExecutableTransaction,
@@ -1348,9 +1349,7 @@ impl AuthorityState {
             }
         });
 
-        let execution_guard = self
-            .execution_lock_for_executable_transaction(certificate)
-            .await;
+        let execution_guard = self.execution_lock_for_executable_transaction(certificate);
         // Any caller that verifies the signatures on the certificate will have already
         // checked the epoch. But paths that don't verify sigs (e.g. execution
         // from checkpoint, reading from db) present the possibility of an epoch
@@ -1431,7 +1430,7 @@ impl AuthorityState {
             }
         }
 
-        fail_point_async!("crash");
+        fail_point!("crash");
 
         self.commit_certificate(
             certificate,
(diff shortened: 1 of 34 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -1225,7 +1227,6 @@ impl AuthorityState {
             expected_effects_digest,
             epoch_store,
         )
-        .await
```
semenov-vladyslav, line 1228: Why is the current function `try_execute_immediately` still `async fn`? There's no `await`s in it. Looks like we can make it `fn`.

crates/iota-core/src/authority.rs
```diff
@@ -1721,10 +1717,10 @@ impl AuthorityState {
         }
 
         self.dry_exec_transaction_impl(&epoch_store, transaction, transaction_digest)
-            .await
```
semenov-vladyslav, line 1724: The current functions `dry_exec_transaction` can be made non-`async` as well.

crates/iota-core/src/authority.rs
semenov-vladyslav: See `dry_exec_transaction_for_benchmark`.

crates/iota-core/src/authority.rs
```diff
@@ -3103,11 +3096,14 @@ impl AuthorityState {
...
-    pub async fn execution_lock_for_executable_transaction(
+    pub fn execution_lock_for_executable_transaction(
         &self,
         transaction: &VerifiedExecutableTransaction,
     ) -> IotaResult<ExecutionLockReadGuard> {
-        let lock = self.execution_lock.read().await;
+        let lock = self
+            .execution_lock
+            .try_read()
+            .map_err(|_| IotaError::ValidatorHaltedAtEpochEnd)?;
```
semenov-vladyslav, line 3104: Shouldn't it be known in advance whether "validator halted at epoch end" so that we don't hit this error?
Also, the previous asynchronous version would just await for the lock. Should it be `blocking_read` instead of `try_read`?

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -514,14 +514,13 @@ impl WritebackCache {
...
 
-    async fn write_object_entry(
+    fn write_object_entry(
         &self,
         object_id: &ObjectID,
         version: SequenceNumber,
         object: ObjectEntry,
     ) {
         trace!(?object_id, ?version, ?object, "inserting object entry");
-        fail_point_async!("write_object_entry");
```
semenov-vladyslav, line 524: Should it be `fail_point!`?

semenov-vladyslav (commented):
There's a few things worth double-checking.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1391,40 +1367,31 @@ impl IndexStore {
...
-                Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner)
-            })
-            .await
-            .unwrap()
-            .map_err(|e| {
-                IotaError::Execution(format!("Failed to read all balance from DB: {e:?}"))
-            });
+            Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner).map_err(
+                |e| IotaError::Execution(format!("Failed to read all balance from DB: {:?}", e)),
+            )?;
```
semenov-vladyslav: There should be return instead of this `?`: if it's `Ok` the inner value will just get dropped and the function will continue.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1391,40 +1367,31 @@ impl IndexStore {
...
         if force_disable_cache {
-            return spawn_blocking(move || {
-                Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner)
-            })
-            .await
-            .unwrap()
-            .map_err(|e| {
-                IotaError::Execution(format!("Failed to read all balance from DB: {e:?}"))
-            });
+            Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner).map_err(
```
semenov-vladyslav: The logic is incorrect/misleading:
- if `force_disable_cache` is `false` then the cache will be used and the filter to retain non-zero balance coins be applied;
- if it's `true` then there's no such filter.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1391,40 +1367,31 @@ impl IndexStore {
...
-                .unwrap()
-                .map_err(|e| {
-                    IotaError::Execution(format!("Failed to read all balance from DB: {e:?}"))
-                })
+            .get_with(owner, move || {
+                Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner).map_err(
+                    |e| {
+                        IotaError::Execution(format!("Failed to read all balance from DB: {:?}", e))
+                    },
+                )
```
semenov-vladyslav: Should the retain filter be applied here in the init closure (which will efficiently cache filtered coins) instead of afterwards?

semenov-vladyslav (changes requested):
Coin index stuff should be fixed.

crates/iota-core/src/execution_cache/writeback_cache.rs
jkrvivian: Upstream removed it. 👀

crates/iota-core/src/authority.rs
jkrvivian: But `execution_lock_for_executable_transaction` is called by an async function, using `blocking_read` will panic.

crates/iota-core/src/jsonrpc_index.rs
jkrvivian: Fixed in https://github.com/iotaledger/iota/pull/8055/commits/17c85a5d2a210120fe03b3d33614bf694b40b4ae

crates/iota-core/src/jsonrpc_index.rs
jkrvivian: Fixed in https://github.com/iotaledger/iota/pull/8055/commits/17c85a5d2a210120fe03b3d33614bf694b40b4ae

crates/iota-core/src/authority.rs
jkrvivian: Before during reconfiguration, the method would block indefinitely waiting for the lock, which could cause deadlocks. With non-blocking `try_read`, the deadlock scenario should be prevented.

semenov-vladyslav: I hope waiting on a read lock cannot cause deadlocks in this case, otherwise it would indicate internal logic error.
`try_read` is ok here because the only possible case when a lock is acquired for exclusive access is during reconfiguration after which epoch would increase and thus the check in the next line would fail.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1324,41 +1319,27 @@ impl IndexStore {
...
-                    metrics_cloned,
-                    coin_index_cloned,
-                    owner,
-                    cloned_coin_type,
-                )
-            })
-            .await
-            .unwrap()
-            .map_err(|e| IotaError::Execution(format!("Failed to read balance frm DB: {e:?}")));
+            Self::get_balance_from_db(metrics_cloned, coin_index_cloned, owner, cloned_coin_type)
```
semenov-vladyslav: I think there should be return instead of `?`.

semenov-vladyslav (changes requested):
Missed one spot.

crates/iota-core/src/authority.rs
```diff
@@ -1261,13 +1262,11 @@ impl AuthorityState {
...
     ) -> IotaResult<(VerifiedSignedTransactionEffects, Option<ExecutionError>)> {
         let epoch_store = self.epoch_store_for_testing();
-        let (effects, execution_error_opt) = self
-            .try_execute_immediately(
-                &VerifiedExecutableTransaction::new_from_certificate(certificate.clone()),
-                None,
-                &epoch_store,
-            )
-            .await?;
+        let (effects, execution_error_opt) = self.try_execute_immediately(
```
muXxer, line 1265: Does `try_execute_for_test` need to async then?

crates/iota-core/src/jsonrpc_index.rs
muXxer: good catch, I also think so

crates/iota-json-rpc/src/authority_state.rs
```diff
@@ -326,9 +326,7 @@ impl StateRead for AuthorityState {
         TransactionEffects,
         Option<ObjectID>,
     )> {
-        Ok(self
```
muXxer, line 329: I wonder if we also need to spawn a tokio task here?

muXxer (changes requested):

crates/iota-core/src/jsonrpc_index.rs
muXxer: I would move it into the `get_all_balances_from_db`

crates/iota-core/src/authority.rs
```diff
@@ -861,7 +861,7 @@ impl AuthorityState {
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<VerifiedSignedTransaction> {
         // Ensure that validator cannot reconfigure while we are signing the tx
-        let _execution_lock = self.execution_lock_for_signing().await;
+        let _execution_lock = self.execution_lock_for_signing();
```
semenov-vladyslav, line 864: Looks like the current fn `handle_transaction_impl` is also non-async. You can run `cargo clippy -- -D clippy::unused_async` to check for it.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1391,45 +1371,33 @@ impl IndexStore {
...
+            self.caches.all_balances.get_with(owner, move || {
+                Self::get_all_balances_from_db(metrics_cloned, coin_index_cloned, owner).map_err(
+                    |e| {
+                        IotaError::Execution(format!("Failed to read all balance from DB: {:?}", e))
+                    },
+                )
+            })?
+        };
+
+        let mut filtered_map = (*balances).clone();
```
semenov-vladyslav: Why not just declare `balances` as `mut`? No need to `clone` this way.

semenov-vladyslav (changes requested):
Please run clippy to check for `clippy::unused_async` lint.

crates/iota-core/src/jsonrpc_index.rs
jkrvivian: Move the filter to `get_all_balances_from_db` in https://github.com/iotaledger/iota/pull/8055/commits/e8234aa76a742226c0db33eaed296292dd218f9d

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1451,6 +1445,10 @@ impl IndexStore {
                 },
             );
         }
+
+        // We do not want to return coins with 0 balance
+        balances.retain(|_, TotalBalance { num_coins, .. }| *num_coins > 0);
```
muXxer: why do we need to filter at the end, if we can directly skip the insertion in the first place?
I would put a check inside the for loop, before the `let coin_type`.

crates/iota-core/src/authority.rs
muXxer: We might fix that in another dedicated PR.

crates/iota-core/src/jsonrpc_index.rs
muXxer: Fixed with https://github.com/iotaledger/iota/pull/8055/commits/219f281bcdc9eac33ed7605e0bfd863cd5179365

muXxer (approved):

semenov-vladyslav (approved):

bingyanglin (approved):

piotrm50 (approved):

sergiupopescu199 (approved):
lgtm!

Much cleaner and more intuitive to follow the without those "fake async" functions!

Reviewed only the files owned by the `infra` team:

- `crates/iota-json-rpc/src/authority_state.rs`
- `crates/iota-json-rpc/src/transaction_execution_api.rs`

Issue #8054, opened 2025-07-31
starfish: add metric for transaction latency


polinikita:
Now, only block latency is computed. We need to enhance the metrics with the transaction latency to see how it is changed compared to Mysticeti

PR #8064, opened 2025-07-31
feat(starfish): add transaction latency metrics

polinikita:
# Description of change

We introduce the transaction latency which is larger than the block latency. We report the transaction latency (only for block headers in memory) and the actual number of committed transactions

#8054 

### Release Notes

- [x] Protocol: new transaction latency metric
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/commit_observer.rs
+++ b/crates/starfish/core/src/commit_observer.rs
@@ -53,6 +53,9 @@ pub(crate) struct CommitObserver {
     sender: UnboundedSender<CommittedSubDag>,
     /// Persistent storage for blocks, commits and other consensus data.
     store: Arc<dyn Store>,
+    /// Dag state for direct access to block headers
+    dag_state: Arc<RwLock<DagState>>,
+
     leader_schedule: Arc<LeaderSchedule>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/starfish/core/src/commit_observer.rs
```diff
@@ -315,7 +324,60 @@ impl CommitObserver {
...
+                    .sum::<usize>() as f64,
+            );
+
+            let block_refs_for_committed_txs = commit
+                .transactions
+                .iter()
+                .map(|tx| tx.block_ref())
+                .collect::<Vec<_>>();
+
+            // TODO: decide whether we want to read from storage here or not as it could
```
piotrm50: I think we can ignore metrics for old transactions on a node that is experiencing some problems anyway  (networking or otherwise) and not add more IO requests.

piotrm50 (approved):

crates/starfish/core/src/commit_observer.rs
polinikita: added a new function that reads only cached headers. So, the transaction latency could be represented in a better way that it is, but it is still a good indicator of the actual e2e latency

PR #8074, opened 2025-08-01
fix(iota-data-ingestion-core): add timeout to historical reader

### LINKED ISSUE #8072: fix(iota-data-ingestion-core): no timeout on HisotricalReader read operations in CheckpointReader v2
## Description

Fix the `CheckpointReader` using the `HisotricalReader` to have a timeout on read operation for files gathered form S3 since the object store client `HisotricalReader` relies on does not have timeouts.

sergiupopescu199:
# Description of change

This PR fixes a corner case where the GET requests from the `HisotricalReader` to the `S3 ObjectStore` could hang for around 2 hours. This was because the `ObjectStore client` the `HisotricalReader` relies does disable all timeouts (this change was inherited from upstream). 

The fix applies a `tokio::time::timeout`, the duration can be configured as it reads the `ReaderOptions.timeout_secs` field.

fixes #8072 

- Ran a custom indexer and manitored it's logs for some hours to assert the fix works. in this amount of time it happened 2 times, once the timeout was reached, it exited the loop and once entered again it was able to successfully fetched the requested file.

> [!NOTE]
> The indexer does not use the `CheckpointReader V2` so this path does not affect the normal operation of the indexer. QA tests were skipped.

### Release Notes

- [ ] Protocol:
- [x] Indexer: the `iota-data-ingestion-core` `CheckpointReader` applies the `ReaderOptions.timeout_secs` also for historical checkpoint reads, previously only live checkpoint reads were respecting the provided timeout value.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-data-ingestion-core/src/reader/v2.rs
+++ b/crates/iota-data-ingestion-core/src/reader/v2.rs
@@ -117,7 +117,6 @@ impl RemoteStore {
                 };
                 let historical = HistoricalReader::new(config)
                     .inspect_err(|e| error!("Unable to instantiate historical reader: {e}"))?;
-                historical.sync_manifest_once().await?;
 
                 let live = live_url
                     .map(|url| create_remote_store_client(url, Default::default(), timeout_secs))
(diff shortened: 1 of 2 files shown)


samuel-rufi (approved):
Awesome @sergiupopescu199 , looks good to me!

tomxey (approved):
lgtm

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -117,7 +117,6 @@ impl RemoteStore {
                 };
                 let historical = HistoricalReader::new(config)
                     .inspect_err(|e| error!("Unable to instantiate historical reader: {e}"))?;
-                historical.sync_manifest_once().await?;
```
Thoralf-M, line 120: This is also not needed with timeout?

sergiupopescu199: I removed it because when spawning the `HisotricalReader` internally it fetches the Manifest in a `tokio` task.

Thoralf-M (approved):

PR #8101, opened 2025-08-05
feat(iota-core,iota-types): Add AuthorityCapabilities endpoint 

### LINKED ISSUE #7992: Implement an endpoint to publish AuthorityCapabilities
As part of #7814 we need to know AuthorityCapabilities not only of Committee nodes but all Active Validators that can form a committee in the next epoch. Currently it's only published through `consensus` by Committee validators and non-committee validators don't have the possibility to do this. 
To enable them this possibility, we need to create a new endpoint of committee validators which will receive signed AuthorityCapabilities from other non-committee validators and insert it into `consensus` so that all the nodes have the same view of the capabilities.

piotrm50:
# Description of change

This PR implements the authority capabilities advertisement endpoint for non-committee validators, enabling them to notify committee validators about their capabilities and supported protocol versions. This feature lays the groundwork for improved validator coordination and protocol version management across the network that will be implemented in subsequent PRs.

Key Features

- New message format for non-committee active validators to advertise their capabilities.
- Signature Verification: Enhanced signature verifier to validate capability notifications from authorized non-committee active validators
- gRPC Endpoint: New handle_capability_notification_v1 endpoint for receiving capability advertisements
- Epoch State V2: Extended epoch start state to include non-committee active validator information
- Comprehensive Testing: Added unit tests to verify proper rejection of unauthorized capability notifications

Resolves #7992 

- Unit tests verify proper rejection of capability notifications from unauthorized authorities
- Further tests will be implemented in subsequent PRs as end-to-end tests. 

## Change checklist

- [x] I have followed the [contribution guidelines](../../CONTRIBUTING.md) for this project
- [ ] I have made corresponding changes to the documentation

### Release Notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates.

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Adds gRPC endpoint to allow non-committee active validators to forward their signed AuthorityCapabilities message to committee members. This will enable selecting new committee only among validators supporting new protocol version when advancing epoch. This endpoint is only enabled on the devnet. 
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/benches/batch_verification_bench.rs
+++ b/crates/iota-core/benches/batch_verification_bench.rs
@@ -78,7 +78,7 @@ fn async_verifier_bench(c: &mut Criterion) {
                         .unwrap();
                     let batch_verifier = Arc::new(SignatureVerifier::new_with_batch_size(
                         committee.clone(),
-                        BTreeMap::new(),
+                        BTreeSet::new(),
                         batch_size,
                         metrics.clone(),
                         ZkLoginEnv::Test,
--- a/crates/iota-core/benches/batch_verification_bench.rs
+++ b/crates/iota-core/benches/batch_verification_bench.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeSet, sync::Arc};
 
 use criterion::*;
 use fastcrypto_zkp::bn254::zk_login_api::ZkLoginEnv;
--- a/crates/iota-core/benches/batch_verification_bench.rs
+++ b/crates/iota-core/benches/batch_verification_bench.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
 
 use criterion::*;
 use fastcrypto_zkp::bn254::zk_login_api::ZkLoginEnv;
@@ -78,6 +78,7 @@ fn async_verifier_bench(c: &mut Criterion) {
                         .unwrap();
                     let batch_verifier = Arc::new(SignatureVerifier::new_with_batch_size(
                         committee.clone(),
+                        BTreeMap::new(),
                         batch_size,
                         metrics.clone(),
                         ZkLoginEnv::Test,
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -2334,7 +2334,7 @@ impl AuthorityPerEpochStore {
 
     /// Record most recently advertised capabilities of all authorities
     pub fn record_capabilities_v1(&self, capabilities: &AuthorityCapabilitiesV1) -> IotaResult {
-        info!("received capabilities v1 {:?}", capabilities);
+        info!("received capabilities v1 {capabilities:?}");
         let authority = &capabilities.authority;
         let tables = self.tables()?;
 
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -36,9 +36,7 @@ use iota_types::{
         TransactionDigest,
     },
     committee::{Committee, CommitteeTrait},
-    crypto::{
-        AuthorityPublicKey, AuthoritySignInfo, AuthorityStrongQuorumSignInfo, RandomnessRound,
-    },
+    crypto::{AuthoritySignInfo, AuthorityStrongQuorumSignInfo, RandomnessRound},
     digests::{ChainIdentifier, TransactionEffectsDigest},
     effects::TransactionEffects,
     error::{IotaError, IotaResult},
@@ -52,7 +50,8 @@ use iota_types::{
     },
     messages_consensus::{
         AuthorityCapabilitiesV1, ConsensusTransaction, ConsensusTransactionKey,
-        ConsensusTransactionKind, TimestampMs, VersionedDkgConfirmation, check_total_jwk_size,
+        ConsensusTransactionKind, SignedAuthorityCapabilitiesV1, TimestampMs,
+        VerifiedAuthorityCapabilitiesV1, VersionedDkgConfirmation, check_total_jwk_size,
     },
     signature::GenericSignature,
     storage::{BackingPackageStore, InputKey, ObjectStore},
@@ -907,7 +906,9 @@ impl AuthorityPerEpochStore {
 
         let signature_verifier = SignatureVerifier::new(
             committee.clone(),
-            epoch_start_configuration.non_committee_validators(),
+            epoch_start_configuration
+                .epoch_start_state()
+                .get_non_committee_validators(),
             signature_verifier_metrics,
             zklogin_env,
             protocol_config.accept_zklogin_in_multisig(),
@@ -2326,7 +2327,7 @@ impl AuthorityPerEpochStore {
 
     /// Record most recently advertised capabilities of all authorities
     pub fn record_capabilities_v1(&self, capabilities: &AuthorityCapabilitiesV1) -> IotaResult {
-        info!("received capabilities v2 {:?}", capabilities);
+        info!("received capabilities v1 {:?}", capabilities);
         let authority = &capabilities.authority;
         let tables = self.tables()?;
 
@@ -2560,6 +2561,16 @@ impl AuthorityPerEpochStore {
         self.signature_verifier
             .verify_tx(tx.data())
             .map(|_| VerifiedTransaction::new_from_verified(tx))
+    }
+
+    #[instrument(level = "trace", skip_all)]
+    pub fn verify_authority_capabilities(
+        &self,
+        authority_capabilities: SignedAuthorityCapabilitiesV1,
+    ) -> IotaResult<VerifiedAuthorityCapabilitiesV1> {
+        self.signature_verifier
+            .verify_authority_capabilities(&authority_capabilities)
+            .map(|_| VerifiedAuthorityCapabilitiesV1::new_from_verified(authority_capabilities))
     }
 
     /// Verifies transaction signatures and other data
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -904,11 +904,18 @@ impl AuthorityPerEpochStore {
             _ => ZkLoginEnv::Test,
         };
 
+        // Get all active validators and filter out committee members to get
+        // non-committee validators
+        let non_committee_validators: BTreeMap<_, _> = epoch_start_configuration
+            .epoch_start_state()
+            .get_active_validators()
+            .into_iter()
+            .filter(|(authority_name, _)| !committee.authority_exists(authority_name))
+            .collect();
+
         let signature_verifier = SignatureVerifier::new(
             committee.clone(),
-            epoch_start_configuration
-                .epoch_start_state()
-                .get_non_committee_validators(),
+            non_committee_validators,
             signature_verifier_metrics,
             zklogin_env,
             protocol_config.accept_zklogin_in_multisig(),
@@ -2638,7 +2645,7 @@ impl AuthorityPerEpochStore {
                 ..
             }) => {
                 // TODO: this needs to be modified as committee validators might be different
-                // than the actual authority
+                //  than the actual authority sending the notification.
                 if transaction.sender_authority() != *authority {
                     warn!(
                         "CapabilityNotificationV1 authority {} does not match its author from consensus {}",
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -1118,6 +1118,10 @@ impl AuthorityPerEpochStore {
         &self.committee
     }
 
+    pub fn committee(&self) -> &Arc<Committee> {
+        &self.committee
+    }
+
     pub fn protocol_config(&self) -> &ProtocolConfig {
         &self.protocol_config
     }
@@ -2623,6 +2627,8 @@ impl AuthorityPerEpochStore {
                     }),
                 ..
             }) => {
+                // TODO: this needs to be modified as committee validators might be different
+                // than the actual authority
                 if transaction.sender_authority() != *authority {
                     warn!(
                         "CapabilityNotificationV1 authority {} does not match its author from consensus {}",
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -36,7 +36,9 @@ use iota_types::{
         TransactionDigest,
     },
     committee::{Committee, CommitteeTrait},
-    crypto::{AuthoritySignInfo, AuthorityStrongQuorumSignInfo, RandomnessRound},
+    crypto::{
+        AuthorityPublicKey, AuthoritySignInfo, AuthorityStrongQuorumSignInfo, RandomnessRound,
+    },
     digests::{ChainIdentifier, TransactionEffectsDigest},
     effects::TransactionEffects,
     error::{IotaError, IotaResult},
@@ -905,6 +907,7 @@ impl AuthorityPerEpochStore {
 
         let signature_verifier = SignatureVerifier::new(
             committee.clone(),
+            epoch_start_configuration.non_committee_validators(),
             signature_verifier_metrics,
             zklogin_env,
             protocol_config.accept_zklogin_in_multisig(),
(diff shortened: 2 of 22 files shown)


crates/iota-types/src/messages_grpc.rs
```diff
@@ -272,3 +273,12 @@ pub struct HandleSoftBundleCertificatesRequestV1 {
...
     pub include_auxiliary_data: bool,
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HandleCapabilityNotificationV1Request {
+    pub message: SignedAuthorityCapabilitiesV1,
+}
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HandleCapabilityNotificationV1Response {
+    // Currently empty, may be extended in the future
```
Thoralf-M: Breaking changes are fine here or how would this be extended in the future?

piotrm50: Good point, I guess changing this in the future would be problematic. I'll remove the comment and I'll double check if something needs to be returned here.

crates/iota-core/src/signature_verifier.rs
```diff
@@ -409,8 +425,55 @@ impl SignatureVerifier {
         )
     }
 
+    pub fn verify_authority_capabilities(
+        &self,
+        signed_authority_capabilities: &SignedAuthorityCapabilitiesV1,
+    ) -> IotaResult {
+        self.authority_capability_cache.is_verified(
+            *signed_authority_capabilities.digest(),
```
semenov-vladyslav, line 432: This is a potentially dangerous spot. The `digest()` does not include the epoch, while epoch is appended to the serialized capabilities for signature verification below. So valid capabilities in epoch1 will be cached and will be valid in epoch2 which should not be the case. Of course, `SignatureVerifier` is only used in `AuthorityPerEpochStore` so there's a new cache in each new epoch and thus can't be exploited. However, it would be great to not introduce a weakness in the first place (in case `SignatureVerifier` is reused elsewhere). It makes sense to me to include epoch id in the message itself so that it is counted in `digest()`.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2592,6 +2606,8 @@ impl AuthorityPerEpochStore {
                     }),
                 ..
             }) => {
+                // TODO: this needs to be modified as committee validators might be different
+                // than the actual authority
```
semenov-vladyslav: Unclear comment: what is "committee validators" here, what is the "actual" authority? There's only `transaction.sender_authority()` and "capability holder" `authority`.

crates/iota-core/src/unit_tests/server_tests.rs
```diff
@@ -44,3 +54,109 @@ async fn test_simple_request() {
...
+    // This should also be rejected - authorities should not accept capability
+    // notifications from themselves or other committee members
+    assert!(
+        validator_service
+            .handle_capability_notification_v1(make_tonic_request_for_testing(request2))
+            .await
+            .is_err(),
+        "Expected capability notification from authority itself to be rejected"
+    );
+}
```
semenov-vladyslav, line 170: Please, add a test case for a bad epoch.

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
```diff
@@ -274,6 +309,209 @@ impl EpochStartSystemStateTrait for EpochStartSystemStateV1 {
...
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct EpochStartSystemStateV2 {
+    epoch: EpochId,
+    protocol_version: u64,
+    reference_gas_price: u64,
+    safe_mode: bool,
+    epoch_start_timestamp_ms: u64,
+    epoch_duration_ms: u64,
+    committee_validators: Vec<EpochStartValidatorInfoV1>,
+    non_committee_validators: Option<Vec<EpochStartValidatorInfoV1>>,
```
semenov-vladyslav: Does it have to be `Option`? Will empty `Vec` not work for `None`?

semenov-vladyslav (changes requested):
`SignatureVerifier::verify_authority_capabilities` is potentially dangerous if misused.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: This will be resolved in a subsequent PR. The problem is that we currently assume that the AuthorityNotification author is also the author of the consensus block that contains it, however we want to allow non-committee validators to also issue those so that they can be considerred during committee selection. This needs to be extended and the verification needs to be updated.

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
piotrm50: Replaced it with a Vec without an Option wrapper and changed the semantic of the field as well.

crates/iota-core/src/unit_tests/server_tests.rs
piotrm50: This will be added as end-to-end test as setting up non-committee validators in those tests would require a lot of changes to genesis creation etc.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2592,6 +2613,8 @@ impl AuthorityPerEpochStore {
                     }),
                 ..
             }) => {
+                // TODO: this needs to be modified as committee validators might be different
+                //  than the actual authority sending the notification.
```
Thoralf-M, line 2651: Should this be done as part of this PR?

piotrm50: It's gonna be extended by a subsequent PR. I divided it into mulitple PRs for easier and more focused review.

crates/iota-core/src/signature_verifier.rs
piotrm50: In order to avoid implementing a whole new SignedAuthorityCapabilitiesV2 struct, I added a cache_digest method that takes epoch as a parameter and includes it in digest calculation that is only used in SignatureVerifier's cache object. Is that ok?

crates/iota-core/src/authority_server.rs
```diff
@@ -197,133 +200,147 @@ impl ValidatorServiceMetrics {
                 "Number of transaction signature errors",
                 registry,
             )
-            .unwrap(),
+                .unwrap(),
```
bingyanglin: Do we need these spaces changes? Same for the following `unwrap()`.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2296,7 +2307,7 @@ impl AuthorityPerEpochStore {
 
     /// Record most recently advertised capabilities of all authorities
     pub fn record_capabilities_v1(&self, capabilities: &AuthorityCapabilitiesV1) -> IotaResult {
-        info!("received capabilities v2 {:?}", capabilities);
+        info!("received capabilities v1 {:?}", capabilities);
```
bingyanglin:
```suggestion
        info!("received capabilities v1 {capabilities:?}");
```

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
```diff
@@ -303,6 +338,191 @@ impl EpochStartSystemStateTrait for EpochStartSystemStateV1 {
...
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct EpochStartSystemStateV2 {
+    epoch: EpochId,
+    protocol_version: u64,
+    reference_gas_price: u64,
+    safe_mode: bool,
+    epoch_start_timestamp_ms: u64,
+    epoch_duration_ms: u64,
+    committee_validators: Vec<EpochStartValidatorInfoV1>,
```
semenov-vladyslav: Why is `EpochStartSystemStateV1` not reused for these fields? That would prevent code duplication/copy-paste, reduce the code size, the differences between V1 and V2 would be more pronounced.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1315,6 +1331,14 @@ impl ProtocolConfig {
     pub fn additional_multisig_checks(&self) -> bool {
         self.feature_flags.additional_multisig_checks
     }
+
+    pub fn select_committee_supporting_protocol_version(&self) -> bool {
```
semenov-vladyslav: Should it be `fn select_committee_from_eligible_validators`?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -316,6 +320,18 @@ struct FeatureFlags {
     // If true enable additional multisig checks.
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
+
+    // If true, use ChangeEpochV3 for epoch change to pass an additional eligible_active_validators
```
semenov-vladyslav, line 325: What is `ChangeEpochV3`?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -316,6 +320,18 @@ struct FeatureFlags {
...
+    // parameter to IotaSystem's advance_epoch call. This should only be enabled when on-chain
+    // IotaSystem objects are updated as well.
+    #[serde(skip_serializing_if = "is_false")]
+    select_committee_from_eligible_validators: bool,
```
semenov-vladyslav, line 329: Where is this flag used?

semenov-vladyslav: Why are the comments to `select_committee_from_eligible_validators` and `track_non_committee_eligible_validators` the same?

crates/iota-core/src/authority_server.rs
piotrm50: Not necessary. I'm not sure why it's formatted like that and why rustfmt doesn't format it at all.

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
```diff
@@ -38,6 +39,7 @@ pub trait EpochStartSystemStateTrait {
     fn get_validator_as_p2p_peers(&self, excluding_self: AuthorityName) -> Vec<PeerInfo>;
     fn get_authority_names_to_peer_ids(&self) -> HashMap<AuthorityName, PeerId>;
     fn get_authority_names_to_hostnames(&self) -> HashMap<AuthorityName, String>;
+    fn get_active_validators(&self) -> Vec<(AuthorityName, AuthorityPublicKey)>;
```
semenov-vladyslav: `(AuthorityName, AuthorityPublicKey)` doesn't make sense to me. `AuthorityName` is a synonym for `AuthorityPublicKeyBytes` which can be converted to and from `AuthorityPublicKey` (unless for some reason it can't be converted). Just return `Vec<AuthorityName>` (or `Vec<AuthorityPublicKey>`) or `impl Iterator<Item = AuthorityName>` even better.

crates/iota-core/src/signature_verifier.rs
```diff
@@ -94,8 +98,11 @@ impl CertBuffer {
 /// - User signed data - caching.
 pub struct SignatureVerifier {
     committee: Arc<Committee>,
+    non_committee_validators: BTreeMap<AuthorityName, AuthorityPublicKey>,
```
semenov-vladyslav: Do you really need a `BTreeMap` here? Why not just `Vec<AuthorityPublicKey>`?

crates/iota-core/src/signature_verifier.rs
```diff
@@ -409,8 +425,55 @@ impl SignatureVerifier {
...
+            *signed_authority_capabilities.digest(),
+            || {
+                // Check if authority exists in non-committee validators
+                let authority_name = signed_authority_capabilities.data().authority;
+                let authority_key = self
+                    .non_committee_validators
+                    .get(&authority_name)
+                    .ok_or_else(|| IotaError::IncorrectSigner {
+                        error: "Signer must be part of non-committee active validators".to_string(),
+                    })?;
```
semenov-vladyslav: Why do you need to lookup `authority_key` in a map while you can just convert `authority_name` to the corresponding `BLS12381PublicKey`? Of course, you still need to verify that the authority is a non committee validator.

crates/iota-core/src/signature_verifier.rs
semenov-vladyslav: That would be OK if it was implemented. There's no `cache_digest` method taking `epoch` as a parameter. Maybe, I was looking in a wrong place?

semenov-vladyslav: You can also just add a comment stating that "`self` stays constant between calls to this function, thus epoch stays constant, so the verification results stay consistent even if cache is invalidated. Epoch is important since the capabilities message is signed for one specific epoch."

semenov-vladyslav (changes requested):
A few minor nits could be improved.

crates/iota-core/src/signature_verifier.rs
semenov-vladyslav: Found `cache_digest` in #8127.

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
piotrm50: I tried it, but the serialized format was different. Now that I think about it it doesn't really matter as we don't exchange it over the wire with other nodes but only store it in the database. We don't have to worry about the byte layout following the same format as V1 then.

crates/iota-protocol-config/src/lib.rs
piotrm50: It's a new transaction type introduced in a subsequent PR that is issued based on this flag.

crates/iota-protocol-config/src/lib.rs
piotrm50: This flag is used in subsequent PRs. Fixed the comment for track_non_committee_eligible_validators so it correctly describes the feature.

crates/iota-protocol-config/src/lib.rs
piotrm50: good catch!

crates/iota-core/src/signature_verifier.rs
piotrm50: I think it's better to use BTreeSet/HashSet instead of a Vector as the main purpose of it then is  to test for existence in the set.

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
piotrm50: Changed to `Vec<AuthorityPublicKey>`

crates/iota-core/src/signature_verifier.rs
piotrm50: Changed to only test for existence of the key in the set.

cyberphysic4l (approved):

cyberphysic4l (approved):
Reviewed changes in iota-core crate

crates/iota-types/src/iota_system_state/epoch_start_iota_system_state.rs
semenov-vladyslav: Even if there's a need to preserve serialized format, it's always possible to override serde serialize/deserialize methods.

crates/iota-core/src/signature_verifier.rs
semenov-vladyslav: Perhaps, but at least it's not a Map. Also, for small vectors just plain search would be faster. Or you can just sort the vector and use binary search.

crates/iota-core/benches/batch_verification_bench.rs
```diff
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
```
semenov-vladyslav:
```suggestion
use std::{collections::BTreeSet, sync::Arc};
```

crates/iota-core/src/unit_tests/batch_verification_tests.rs
```diff
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
```
semenov-vladyslav:
```suggestion
use std::{collections::BTreeSet, sync::Arc};
```

semenov-vladyslav (changes requested):
Typos in crates/iota-core/benches/batch_verification_bench.rs, crates/iota-core/src/unit_tests/batch_verification_tests.rs.
Consider using sorted Vec instead of BTreeSet, might be more efficient (I am not insisting).

semenov-vladyslav (approved):

filipdulic (approved):

Issue #7994, opened 2025-07-25
Use AuthorityCapabilities when selecing committee for a new epoch


piotrm50:
Follow up for #7992 and #7993. When all the necessary information is available, we can use that to only select validators that support new protocol version during committee selection in Move.

PR #8127, opened 2025-08-06
feat(iota-core,iota-types,iota-framework): Process AuthorityCapabilities from non-committee validators

### LINKED ISSUE #7993: Track AuthorityCapabilities for all active validators
A follow-up task after #7992 is implemented. The scope of this task is to track AuthorityCapabilities on the node side to be able to pass them to the `advance_epoch` function when advancing epochs (this will be a separate tasks).

piotrm50:
# Description of change

This PR updates the way committee selection is done by limiting eligible active validators only to those that support the ProtocolVersion of the upcoming epoch. This involves introducing ChangeEpochV3 which contains indices of the active validators supporting upcoming ProtocolVersion. 
There is a fallback to select among all active validators if eligible_active_validators set does not contain a quorum of current committee.

Resolves #7993 #7994 

Full suite of tests and CI fixes will be implemented in the following PR. 

### Release Notes

- [x] Protocol: Improved committee selection where committee is selected only from the list of vector of eligible active validators. This will be enabled on all networks with the same protocol version. By default all active validators are considered eligible, however when `track_non_committee_eligible_validators` feature flag is enabled, then the list of eligible validators is composed of committee and non-committee active validators that have issued AuthorityCapabilities to signal their readiness to become a committee member as well as announce that they support the next epoch's protocol version. 
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-types/src/transaction.rs
```diff
@@ -466,13 +537,30 @@ impl EndOfEpochTransactionKind {
...
                         "protocol defined base fee required".to_string(),
                     ));
                 }
+                if config.select_committee_supporting_protocol_version() {
+                    return Err(UserInputError::Unsupported(
+                        "selecting committee only among validators supporting the protocol version not supported".to_string(),
+                    ));
+                }
+            }
+            Self::ChangeEpochV3(_) => {
```
Thoralf-M, line 558: Is the `if !config.protocol_defined_base_fee() {` closure for previous versions not needed here?

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1686,6 +1733,7 @@ pub struct IotaEndOfEpochTransaction {
 pub enum IotaEndOfEpochTransactionKind {
     ChangeEpoch(IotaChangeEpoch),
     ChangeEpochV2(IotaChangeEpochV2),
+    ChangeEpochV3(IotaChangeEpochV3),
```
kodemartin: Adding an enum variant is a breaking change. Instead you could just extend `IotaChangeEpochV2` to take the additional `eligible_active_validators` and `serde(skip_serializing_if = "Vec::is_empty")` this will be handled gracefully by all sdks.

crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
```diff
@@ -44,6 +45,7 @@ pub(crate) struct EndOfEpochTransaction {
 pub(crate) enum EndOfEpochTransactionKind {
     ChangeEpoch(ChangeEpochTransaction),
     ChangeEpochV2(ChangeEpochTransactionV2),
+    ChangeEpochV3(ChangeEpochTransactionV3),
```
kodemartin: Here too, adding an enum variant  might break some client implementations. I suggest instead extending `ChangeEpochTransactionV2` with the `eligible_active_validators` field by adding it as an `Option`.

kodemartin (changes requested):
The new feature can be accommodated by both the JSON-RPC and the GraphQL API in a more lenient way, so as to keep backwards compatibility. See more specific suggestions.

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -621,6 +621,29 @@ impl From<crate::transaction::EndOfEpochTransactionKind> for EndOfEpochTransacti
...
+                EndOfEpochTransactionKind::ChangeEpochV3(ChangeEpochV3 {
+                    epoch: change_epoch_v3.epoch,
+                    protocol_version: change_epoch_v3.protocol_version.as_u64(),
+                    storage_charge: change_epoch_v3.storage_charge,
+                    computation_charge: change_epoch_v3.computation_charge,
+                    computation_charge_burned: change_epoch_v3.computation_charge_burned,
+                    storage_rebate: change_epoch_v3.storage_rebate,
+                    non_refundable_storage_fee: change_epoch_v3.non_refundable_storage_fee,
+                    epoch_start_timestamp_ms: change_epoch_v3.epoch_start_timestamp_ms,
+                    // eligible_active_validators: change_epoch_v3.eligible_active_validators,
```
roman1e2f5p8s:
```suggestion
```

crates/iota-types/src/transaction.rs
```diff
@@ -236,6 +236,40 @@ pub struct ChangeEpochV2 {
     pub system_packages: Vec<(SequenceNumber, Vec<Vec<u8>>, Vec<ObjectID>)>,
 }
 
+// System transaction for advancing the epoch.
+// This version includes active validator indices that are eligible
+// to take part in committee selection based on protocol version support.
```
roman1e2f5p8s:
```suggestion
/// System transaction for advancing the epoch.
/// This version includes active validator indices that are eligible
/// to take part in committee selection based on protocol version support.
```

crates/iota-types/src/transaction.rs
roman1e2f5p8s: I have the same question.

roman1e2f5p8s (commented):
Left some minor suggestions.
Overall, files owned by core-ptotocol and research look good to me.
The only thing that needs to be double-checked is https://github.com/iotaledger/iota/pull/8127/files#r2310098221.

crates/iota-types/src/transaction.rs
piotrm50: Well, we could add it, but I was thinking that we can't revert the protocol_defined_base_fee flag anymore so it's unnecessary. But I guess it's better to have it so I'll add it.

crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
piotrm50: I don't think I understand the point - `ChangeEpochTransactionV2` is an alias for `ChangeEpochV2`, so we can't really extend ity with an option because this would change the byte layout of the serialized `ChangeEpochV2` transaction - that's the main reason for adding V3 rather than extending the struct like we do for  `AdvanceEpochParams` which does not have this constraint.

crates/iota-json-rpc-types/src/iota_transaction.rs
piotrm50: 1. Why wasn't it done for V2 as well? It seems like it also only added `computation_charge_burned: u64` field. 
2. Wouldn't it become super confusing after a while that both `ChangeEpochV2` and `ChangeEpochV3` transaction is represented as `ChangeEpochV2` in JSON RPC types? This off-by-one would stay with us forever.
3. Given the above, I think clients will have to update anyway at some point to handle the additional field, while the inconsistent versioning of those structs will create a technical debt that will stay with us much longer and will be harder to maintain in the long run.

crates/iota-core/src/authority.rs
```diff
@@ -4705,6 +4688,65 @@ impl AuthorityState {
         (next_protocol_version, system_packages)
     }
 
+    /// Returns the indices of validators that support the given protocol
+    /// version. This includes both committee and non-committee validators
+    /// based on their capabilities. Uses active validators instead of committee
+    /// indices.
+    fn get_validators_supporting_protocol_version(
+        target_protocol_version: ProtocolVersion,
+        active_validators: Vec<(AuthorityName, AuthorityPublicKey)>,
```
semenov-vladyslav:
```suggestion
        active_validators: &[(AuthorityName, AuthorityPublicKey)],
```

crates/iota-core/src/authority.rs
```diff
@@ -4705,6 +4688,65 @@ impl AuthorityState {
...
+                .supported_protocol_versions
+                .get_version_digest(target_protocol_version)
+                .is_some()
+            {
+                // Find the validator's index in the active validators list
+                if let Some(index) = active_validators
+                    .iter()
+                    .position(|(name, _)| *name == capability.authority)
+                {
+                    eligible_validators.push(index as u64);
```
semenov-vladyslav, line 4743: Looks like items in `eligible_validators` could be duplicated. Is that intended?

crates/iota-core/src/authority.rs
```diff
@@ -4705,6 +4688,65 @@ impl AuthorityState {
...
+            // Check if this validator supports the target protocol version and digest
+            if capability
+                .supported_protocol_versions
+                .get_version_digest(target_protocol_version)
+                .is_some()
+            {
+                // Find the validator's index in the active validators list
+                if let Some(index) = active_validators
+                    .iter()
+                    .position(|(name, _)| *name == capability.authority)
```
semenov-vladyslav: `position` returns the first index, not all of them. Is that intended?

crates/iota-core/src/authority.rs
semenov-vladyslav: Or even better:
```suggestion
        active_validators: &[AuthorityName],
```

crates/iota-core/src/authority.rs
```diff
@@ -4705,6 +4688,65 @@ impl AuthorityState {
...
+    /// to committee members to get their weights.
+    fn calculate_eligible_validators_weight(
+        eligible_validator_indices: &[u64],
+        active_validators: &[(AuthorityName, AuthorityPublicKey)],
+        committee: &Committee,
+    ) -> u64 {
+        let mut total_weight = 0u64;
+
+        for &index in eligible_validator_indices {
+            if let Some((authority_name, _)) = active_validators.get(index as usize) {
```
semenov-vladyslav: Why not just use index operator? The indices must be consistent with the list of validators:
```suggestion
            let (authority_name, _) = active_validators[index as usize];
```

crates/iota-core/src/authority.rs
```diff
@@ -4811,9 +4853,77 @@ impl AuthorityState {
...
+        if config.select_committee_supporting_protocol_version() {
+            // Get the list of eligible validators that support the target protocol version
+            let active_validators = epoch_store.epoch_start_state().get_active_validators();
+
+            //
+            let mut eligible_active_validators = active_validators
+                .iter()
+                .enumerate()
+                .map(|(i, _)| i as u64)
+                .collect();
```
semenov-vladyslav:
```suggestion
            let mut eligible_active_validators = (0..active_validators.len() as u64).collect();
```

crates/iota-core/src/authority.rs
```diff
@@ -4811,9 +4853,77 @@ impl AuthorityState {
...
+                        This could indicate a bug in validator selection logic or inconsistency with protocol version decision.",
+                    );
+                    // Pass all active validator indices as eligible validators
+                    // to perform selection among all of
+                    // them.
+                    eligible_active_validators = active_validators
+                        .iter()
+                        .enumerate()
+                        .map(|(i, _)| i as u64)
+                        .collect();
```
semenov-vladyslav:
```suggestion
                    eligible_active_validators = (0..active_validators.len() as u64).collect();
```

crates/iota-core/src/authority_server.rs
```diff
@@ -1160,7 +1176,7 @@ impl ValidatorService {
         let signed_authority_capabilities = request.message;
         // Verify the message signature
         let verified_authority_capabilities = epoch_store
-            .verify_authority_capabilities(signed_authority_capabilities)
+            .verify_authority_capabilities(signed_authority_capabilities.clone())
```
semenov-vladyslav: Can't `verify_authority_capabilities` take a reference instead?

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -438,6 +441,12 @@ public(package) fun request_set_commission_rate(
 ///   3. Process pending stake deposits, and withdraws.
 ///   4. Process pending validator application and withdraws.
 ///   5. At the end, we calculate the total stake for the new epoch.
+///
+/// IMPORTANT: With the new authority capability notification system, newly activated validators
+/// cannot immediately join the committee. They must wait one epoch after activation to:
+/// 1. Notify their AuthorityCapabilities to the network  
+/// 2. Show that they support the correct ProtocolVersion
+/// This means validators activated in epoch N can only become committee members in epoch N+2.
```
semenov-vladyslav, line 449: Are there tests where:
- validators are activated in epoch N and then some are deactivated in epoch N+1;
- some validators are activated in epoch N, and some in epoch N+1?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2145,9 +2145,9 @@ impl ProtocolConfig {
                     // changes
                 }
                 12 => {
+                    // Enable selecting committee based on eligible active validators in devnet.
```
semenov-vladyslav:
```suggestion
                    // Enable selecting committee based on eligible active validators.
```

kodemartin (approved):
ltgm following [diff](https://github.com/iotaledger/iota/pull/8127/files/15d67194dbe7d45dce75074de5fa2bb8a6edf405..d75230abfd3450e0182980b49ee68998bc559f5b)

semenov-vladyslav (changes requested):
~~Some weird logic with `eligible_validators`.~~ The logic is OK.

crates/iota-framework/packages/iota-system/sources/validator_set.move
piotrm50: Yes, this is covered in [reconfiguration_tests.rs](https://github.com/iotaledger/iota/blob/0cb36503d26114969e81805bb2543a794c1304e2/crates/iota-e2e-tests/tests/reconfiguration_tests.rs)

crates/iota-core/src/authority.rs
piotrm50: Done

crates/iota-core/src/authority.rs
piotrm50: How can it be duplicated? Vector of capabilities should contain one AuthorityCapabilities elem per authority in the first place.

crates/iota-core/src/authority.rs
piotrm50: done

crates/iota-core/src/authority.rs
piotrm50: done

crates/iota-core/src/authority.rs
piotrm50: done

crates/iota-core/src/authority_server.rs
piotrm50: not really, but I modified it so that we can safely move it inside.

crates/iota-core/src/authority.rs
piotrm50: yes, there can only be one active validator with a given AuthorityName

crates/iota-protocol-config/src/lib.rs
piotrm50: Fixed, good catch!

roman1e2f5p8s (approved):
Approving core-protocol and research owned files.

crates/iota-core/src/authority.rs
semenov-vladyslav: Fair enough, `record_capabilities_v1` handles capabilities duplicates.

polinikita (approved):

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -1202,6 +1231,37 @@ fun calculate_total_committee_stakes(
...
+        return vector::tabulate!(active_validators.length(), |i| i)
+    };
+
+    // Calculate total voting power of eligible validators
+    let mut eligible_total_voting_power = 0;
+    eligible_active_validators.do_ref!(|idx| {
+        // Validate index bounds
+        assert!(*idx < active_validators.length(), EInvalidEligibleValidatorIndex);
+        eligible_total_voting_power = eligible_total_voting_power + active_validators[*idx].voting_power();
+    });
```
semenov-vladyslav:
```suggestion
    let eligible_total_voting_power = active_validators.take_fold_ref!(&eligible_active_validators, 0, |sum, v| sum + v.voting_power());
```

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -1458,22 +1514,60 @@ public(package) fun committee_validator_addresses(self: &ValidatorSetV2): vector
...
+
+    // Convert eligible validator indices into current active_validators indices, independent of the changes in active_validators.
+    let mut current_eligible_indices = vector[];
+
+    eligible_active_validators.do_ref!(|idx| {
+        // Validate that the index is within bounds of prev_active_validator_addresses
+        assert!(*idx < prev_active_validator_addresses.length(), EInvalidEligibleValidatorIndex);
+        
+        // Get address from prev_active_validator_addresses using the old index
+        let addr = prev_active_validator_addresses[*idx];
```
semenov-vladyslav:
```suggestion
    prev_active_validator_addresses.take_do_ref!(&eligible_active_validators, |addr| {
```

semenov-vladyslav (approved):
There are `take_do_ref` and `take_fold_ref` macros in `vector`.

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -1202,6 +1231,37 @@ fun calculate_total_committee_stakes(
...
+    });
+
+    // If eligible validators don't have enough voting power, fallback to all validators.
+    // This should never happen under normal circumstances, but we include this
+    // safeguard to ensure the committee selection can always proceed in a safe manner.
+    if (eligible_total_voting_power < voting_power::quorum_threshold()) {
+        vector::tabulate!(active_validators.length(), |i| i)
+    } else {
+        eligible_active_validators
+    }
```
miker83z:
```suggestion
    if (!eligible_active_validators.is_empty()) {
        // Calculate total voting power of eligible validators
        let mut eligible_total_voting_power = 0;
        eligible_active_validators.do_ref!(|idx| {
            // Validate index bounds
            assert!(*idx < active_validators.length(), EInvalidEligibleValidatorIndex);
            eligible_total_voting_power =
                eligible_total_voting_power + active_validators[*idx].voting_power();
        });

        // If eligible validators don't have enough voting power, fallback to all validators.
        // This should never happen under normal circumstances, but we include this
        // safeguard to ensure the committee selection can always proceed in a safe manner.
        if (eligible_total_voting_power >= voting_power::quorum_threshold()) {
            return eligible_active_validators
        }
    };
    // fallback
    vector::tabulate!(active_validators.length(), |i| i)
```

miker83z: Just a nit in this case. But if you don't like this version, still the code should be formatted (use the mysten's Move Formatter VS code extension)

piotrm50: what is more Move idiomatic? I still have the habit of returning early I got from Go :laughing: 

I formatted other files as well. I also noticed that files like `crates/iota-e2e-tests/tests/framework_upgrades/mock_iota_systems/base/sources/iota_system.move` are not formatted at all, but this should be done in a separate PR.

miker83z: I guess Move idiomatic = Rust idiomatic, but even there I am not sure I am following it 100% 😄 

Regarding the format, we have many unsolved cases unfortunately! But up to now we only cared about the files within `crates/iota-framework` and maybe some example.

piotrm50: then let's leave the early return as it is. The files should be now formatted properly so please take another look.

miker83z (approved):

crates/iota-framework/packages/iota-system/sources/validator_set.move
piotrm50: This makes it hard to cover a case when the indices are out of bounds.

crates/iota-framework/packages/iota-system/sources/validator_set.move
piotrm50: This makes it hard to cover a case when the indices are out of bounds.

filipdulic (approved):

Issue #5727, opened 2025-03-03
Port upstream node team related changes [mainnet-v1.43.1..mainnet-v1.44.3).


bingyanglin:

bingyanglin:
Note that we need to investigate the crates owned by both `@iotaledger/core-protocol` and `@iotaledger/node`
(ref: [CODEOWNERS](https://github.com/iotaledger/iota/blob/develop/.github/CODEOWNERS))

Check the missing upstream changes for  `@iotaledger/core-protocol` since `v.1.32.2`

bingyanglin:
The range upstream PR #8125  can be merged after the CI/CID passed.

jkrvivian:
Merged in #8125

PR #8132, opened 2025-08-07
upstream(core): Remove indirect object functionality

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/5739b6d0e9359c40137f497a3509fca892d22811

The default value is `0`, so this functionality was never used. We can safely port this commit to remove it.

Part of #5727 

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): 
   - Remove  `indirect-objects-threshold` from node config
- [ ] Indexer:
- [ ] JSON-RPC: 
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


muXxer:
Can you please do the additional tests I suggested yesterday?

1. Does a node start if this option is still in the config file?
2. Does a node start if the database was created with a version without that change, and then started with this change?

muXxer (changes requested):
Delete the file crates/typed-store/src/rocks/util.rs, it's not used anymore.

jkrvivian:
> Can you please do the additional tests I suggested yesterday?
> 
>     1. Does a node start if this option is still in the config file?
>     2. Does a node start if the database was created with a version without that change, and then started with this change?

1. Yes, serde will ignore the parameter if it still existed.
2. Yes, it works!

muXxer (approved):

semenov-vladyslav (approved):

PR #8138, opened 2025-08-07
feat(iota-types, iota-tx-checks, iota-protocol-config): new struct move_authenticator and dedicated functions

### LINKED ISSUE #8118: [AA][auth] Add the `MoveAuthenticator` as new variant of the `GenericSignature`
- The MoveAuthenticator type has a field inputs of type Vec<CallArg>; this field must be constructed in the same way the inputs field of ProgrammableTransactionBlock is constructed.
- Based on the above new type, a new method check_authenticator_input is needed in the iota-transaction-checks
    - this method will be similar to checked::check_objects
    - it must check that all the authenticator input objects are valid shared or immutable objects
    - it must also create an IotaGasStatus similarly to check_transaction_input
    - it must return IotaResult<(IotaGasStatus, CheckedInputObjects)> 
    - the IotaGasStatus must take into account a new protocol variable which is indicating the max budget for authentication
      - currently we use the protocol config max_tx_gas for transactions.
      - this check needs a new max_auth_gas.

Dkwcs:
# Description of change

- Added `MoveAuthenticator` new variant for `GenericSignature` that implements `AuthenticatorTrait` trait.
- new iota-transaction-check function - `check_auth_inputs` that checks gas status and verifies auth input objects are valid shared or immutable objects. This function returns `IotaResult<(IotaGasStatus, CheckedInputObjects)>`.
- new protocol variable `max_auth_gas` has been provided and set as 1000000000.
- openrpc and protocol snapshots have been updated.

Fixes #8118 .

Check snapshots - `cargo insta review`

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -317,7 +317,7 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
 
-    // Enable passkey auth (SIP-9)
+    // Enable move auth
     #[serde(skip_serializing_if = "is_false")]
     move_auth: bool,
 }
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -520,6 +520,9 @@ pub struct ProtocolConfig {
     /// Maximum gas budget in NANOS that a transaction can use.
     max_tx_gas: Option<u64>,
 
+    /// Maximum gas budget in NANOS that a authentification transaction can use.
+    max_auth_gas: Option<u64>,
+
     /// Maximum amount of the proposed gas price in NANOS (defined in the
     /// transaction).
     max_gas_price: Option<u64>,
@@ -1494,6 +1497,8 @@ impl ProtocolConfig {
             max_publish_or_upgrade_per_ptb: Some(5),
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
+            // max gas budget is in NANOS and an absolute value 50IOTA
+            max_auth_gas: Some(50_000_000_000),
             max_gas_price: Some(100_000),
             max_gas_computation_bucket: Some(5_000_000),
             max_loop_depth: Some(5),
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -316,6 +316,10 @@ struct FeatureFlags {
     // If true enable additional multisig checks.
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
+
+    // Enable passkey auth (SIP-9)
+    #[serde(skip_serializing_if = "is_false")]
+    move_auth: bool,
 }
 
 fn is_true(b: &bool) -> bool {
@@ -1191,6 +1195,10 @@ impl ProtocolConfig {
         self.feature_flags.passkey_auth
     }
 
+    pub fn move_auth(&self) -> bool {
+        self.feature_flags.move_auth
+    }
+
     pub fn max_transaction_size_bytes(&self) -> u64 {
         // Provide a default value if protocol config version is too low.
         self.consensus_max_transaction_size_bytes
@@ -1498,7 +1506,7 @@ impl ProtocolConfig {
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
             // max gas budget is in NANOS and an absolute value 50IOTA
-            max_auth_gas: Some(50_000_000_000),
+            max_auth_gas: Some(1_000_000_000),
             max_gas_price: Some(100_000),
             max_gas_computation_bucket: Some(5_000_000),
             max_loop_depth: Some(5),
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -524,7 +524,7 @@ pub struct ProtocolConfig {
     /// Maximum gas budget in NANOS that a transaction can use.
     max_tx_gas: Option<u64>,
 
-    /// Maximum gas budget in NANOS that a authentification transaction can use.
+    /// Maximum gas budget in NANOS that a authentication transaction can use.
     max_auth_gas: Option<u64>,
 
     /// Maximum amount of the proposed gas price in NANOS (defined in the
@@ -1505,7 +1505,7 @@ impl ProtocolConfig {
             max_publish_or_upgrade_per_ptb: Some(5),
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
-            // max gas budget is in NANOS and an absolute value 50IOTA
+            // max auth gas budget is in NANOS and an absolute value 1IOTA
             max_auth_gas: Some(1_000_000_000),
             max_gas_price: Some(100_000),
             max_gas_computation_bucket: Some(5_000_000),
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -2135,9 +2135,12 @@ impl ProtocolConfig {
                     // changes
                 }
                 12 => {
-                    // max auth gas budget is in NANOS and an absolute value 1IOTA
-                    cfg.max_auth_gas = Some(1_000_000_000);
-                    cfg.feature_flags.move_auth = true;
+                    // TODO: determine required networks
+                    if chain != Chain::Mainnet && chain != Chain::Testnet {
+                        // max auth gas budget is in NANOS and an absolute value 1IOTA
+                        cfg.max_auth_gas = Some(1_000_000_000);
+                        cfg.feature_flags.move_auth = true;
+                    }
                 }
                 // Use this template when making changes:
                 //
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -19,7 +19,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-pub const MAX_PROTOCOL_VERSION: u64 = 11;
+pub const MAX_PROTOCOL_VERSION: u64 = 12;
 
 // Record history of protocol version allocations here:
 //
@@ -71,6 +71,7 @@ pub const MAX_PROTOCOL_VERSION: u64 = 11;
 //             Add additional signature checks
 //             Add additional linkage checks
 // Version 11: Framework fix regarding candidate validator commission rate.
+// Version 12: Max authentication gas budget property.
 
 #[derive(Copy, Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ProtocolVersion(u64);
@@ -1503,10 +1504,10 @@ impl ProtocolConfig {
             max_move_object_size: Some(250 * 1024),
             max_move_package_size: Some(100 * 1024),
             max_publish_or_upgrade_per_ptb: Some(5),
+            // max auth gas budget
+            max_auth_gas: None,
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
-            // max auth gas budget is in NANOS and an absolute value 1IOTA
-            max_auth_gas: Some(1_000_000_000),
             max_gas_price: Some(100_000),
             max_gas_computation_bucket: Some(5_000_000),
             max_loop_depth: Some(5),
@@ -2132,6 +2133,11 @@ impl ProtocolConfig {
                 11 => {
                     // version 11 is a new framework version but with no config
                     // changes
+                }
+                12 => {
+                    // max auth gas budget is in NANOS and an absolute value 1IOTA
+                    cfg.max_auth_gas = Some(1_000_000_000);
+                    cfg.feature_flags.move_auth = true;
                 }
                 // Use this template when making changes:
                 //
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -317,7 +317,7 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
 
-    // Enable move auth
+    // If true, enables the authentication of account using Move code.
     #[serde(skip_serializing_if = "is_false")]
     move_auth: bool,
 }
--- a/crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_1.snap
+++ b/crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_1.snap
@@ -275,3 +275,4 @@ checkpoint_summary_version_specific_data: 1
 max_soft_bundle_size: 5
 bridge_should_try_to_finalize_committee: true
 max_accumulated_txn_cost_per_object_in_mysticeti_commit: 10
+
--- a/crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_1.snap
+++ b/crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_1.snap
@@ -45,6 +45,7 @@ max_move_object_size: 256000
 max_move_package_size: 102400
 max_publish_or_upgrade_per_ptb: 5
 max_tx_gas: 50000000000
+max_auth_gas: 1000000000
 max_gas_price: 100000
 max_gas_computation_bucket: 5000000
 gas_rounding_step: 1000
@@ -275,4 +276,3 @@ checkpoint_summary_version_specific_data: 1
 max_soft_bundle_size: 5
 bridge_should_try_to_finalize_committee: true
 max_accumulated_txn_cost_per_object_in_mysticeti_commit: 10
-
--- a/crates/iota-transaction-checks/src/lib.rs
+++ b/crates/iota-transaction-checks/src/lib.rs
@@ -73,8 +73,20 @@ mod checked {
         input_objects: InputObjects,
     ) -> IotaResult<(IotaGasStatus, CheckedInputObjects)> {
         for object in input_objects.iter() {
-            if object.is_mutable() {
-                panic!("Mutable objects are not allowed") // TODO: Error handling
+            let Some(object) = object.as_object() else {
+                // Skip deleted/cancelled objects
+                continue;
+            };
+
+            match object.owner {
+                Owner::Immutable | Owner::Shared { .. } => {
+                    // valid objects
+                }
+                _ => {
+                    return Err(IotaError::UserInput {
+                        error: UserInputError::ImmutableOrSharedObjectsExpected,
+                    });
+                }
             }
         }
 
--- a/crates/iota-transaction-checks/src/lib.rs
+++ b/crates/iota-transaction-checks/src/lib.rs
@@ -54,44 +54,15 @@ mod checked {
         reference_gas_price: u64,
         transaction: &TransactionData,
     ) -> IotaResult<IotaGasStatus> {
-        if transaction.is_system_tx() {
-            Ok(IotaGasStatus::new_unmetered())
-        } else {
-            check_gas(
-                objects,
-                protocol_config,
-                reference_gas_price,
-                gas,
-                transaction.gas_budget(),
-                transaction.gas_price(),
-            )
-        }
-    }
-
-    #[instrument(level = "trace", skip_all)]
-    pub fn check_auth_inputs(
-        protocol_config: &ProtocolConfig,
-        sender: IotaAddress,
-        input_objects: InputObjects,
-        reference_gas_price: u64,
-        gas_owner: IotaAddress,
-        gas_budget: u64,
-        gas_price: u64,
-        gas: &[ObjectRef],
-    ) -> IotaResult<(IotaGasStatus, CheckedInputObjects)> {
-        // TODO: modify the check_gas in order to support the auth budget
-        let gas_status = check_gas(
-            &input_objects,
+        check_gas(
+            objects,
             protocol_config,
             reference_gas_price,
             gas,
-            gas_budget,
-            gas_price,
-        )?;
-
-        check_auth_objects(sender, gas_owner, gas, &input_objects)?;
-
-        Ok((gas_status, input_objects.into_checked()))
+            transaction.gas_budget(),
+            transaction.gas_price(),
+            transaction.kind(),
+        )
     }
 
     #[instrument(level = "trace", skip_all)]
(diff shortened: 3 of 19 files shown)


crates/iota-types/src/base_types.rs
```diff
@@ -851,6 +851,8 @@ impl TryFrom<&GenericSignature> for IotaAddress {
                 IotaAddress::try_from_unpadded(&zklogin.inputs)
             }
             GenericSignature::PasskeyAuthenticator(s) => Ok(IotaAddress::from(&s.get_pk()?)),
+            // TODO: handle this
+            GenericSignature::MoveAuthenticator(move_authenticator) => todo!(),
```
miker83z: Let's add a field `address: ObjectId` to `MoveAuthenticator` and then here we can call something like `move_authenticator.address.into()`

crates/iota-types/src/crypto.rs
```diff
@@ -1753,6 +1756,9 @@ pub struct ZkLoginAuthenticatorAsBytes(#[schemars(with = "Base64")] pub Vec<u8>)
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
 pub struct PasskeyAuthenticatorAsBytes(#[schemars(with = "Base64")] pub Vec<u8>);
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
+pub struct MoveAuthentificatorAsBytes(#[schemars(with = "Base64")] pub Vec<u8>);
```
miker83z:
```suggestion
pub struct MoveAuthenticatorAsBytes(#[schemars(with = "Base64")] pub Vec<u8>);
```

crates/iota-types/src/multisig.rs
```diff
@@ -266,6 +266,8 @@ impl AuthenticatorTrait for MultiSig {
                         )
                         .map_err(|e| FastCryptoError::GeneralError(e.to_string()))
                 }
+                // TODO: handle this case
+                CompressedSignature::Move(move_authentificator_as_bytes) => todo!(),
```
miker83z: Create an error for this case, IotaError::InvalidSignature "Invalid Move Authenticator signature; it cannot be used for multisig"

crates/iota-types/src/transaction.rs
```diff
@@ -2269,6 +2270,8 @@ impl SenderSignedData {
                     }
                 }
                 GenericSignature::Signature(_) | GenericSignature::MultiSig(_) => (),
+                // TODO: handle this
+                GenericSignature::MoveAuthenticator(move_authenticator) => (),
```
miker83z: You could create a flag `move_auth: bool` in the protocol and fix this.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -65,6 +65,30 @@ mod checked {
...
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_auth_inputs(
+        protocol_config: &ProtocolConfig,
+        reference_gas_price: u64,
+        transaction: &TransactionData,
+        input_objects: InputObjects,
+    ) -> IotaResult<(IotaGasStatus, CheckedInputObjects)> {
+        for object in input_objects.iter() {
+            if object.is_mutable() {
+                panic!("Mutable objects are not allowed") // TODO: Error handling
```
miker83z: Yes please use an appropriate one or create a new one in `UserInputError`

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1494,6 +1497,8 @@ impl ProtocolConfig {
             max_publish_or_upgrade_per_ptb: Some(5),
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
+            // max gas budget is in NANOS and an absolute value 50IOTA
+            max_auth_gas: Some(50_000_000_000),
```
miker83z: make it 1_000_000_000

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -65,6 +65,30 @@ mod checked {
...
+            }
+        }
+
+        let gas_status = get_gas_status(
```
miker83z: Add a `//TODO: modify the check_gas in order to support the auth budget`, we'll do it in Part 4.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -65,6 +65,30 @@ mod checked {
...
 
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_auth_inputs(
+        protocol_config: &ProtocolConfig,
+        reference_gas_price: u64,
+        transaction: &TransactionData,
+        input_objects: InputObjects,
+    ) -> IotaResult<(IotaGasStatus, CheckedInputObjects)> {
+        for object in input_objects.iter() {
+            if object.is_mutable() {
```
miker83z: We need something like `check_auth_objects` similar to `check_objects`. Then move this check inside that new function.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -65,6 +65,30 @@ mod checked {
         )
     }
 
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_auth_inputs(
+        protocol_config: &ProtocolConfig,
+        reference_gas_price: u64,
+        transaction: &TransactionData,
```
miker83z: instead of transaction, you can pass just:
```
gas_budget: u64, // tx.gas_budget()
gas_price: u64, // tx.gas_price()
gas: &[ObjectRef], // tx.gas()            
```

then modify `check_gas(...,tx_kind: Option<&TransactionKind>)`
and instead of callyng `get_gas_status` in this `check_auth_inputs`, you can call `check_gas` directly.

miker83z: You need these as well, to be passed to `check_auth_objects`
```
sender: IotaAddress, // tx.sender()
sponsor: Option<IotaAddress>, // tx.gas_owner()
```

In `check_auth_objects` you can skip the genesis check and call
```
check_one_object(
    &owner_address,
    input_object_kind,
    object,
    false, //system_transaction
)?;
```

crates/iota-transaction-checks/src/lib.rs
Dkwcs: It’s obsolete already

crates/iota-types/src/base_types.rs
Dkwcs: Done

crates/iota-types/src/crypto.rs
Dkwcs: Done

crates/iota-types/src/multisig.rs
Dkwcs: Done

crates/iota-types/src/transaction.rs
Dkwcs: Done

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Done

crates/iota-protocol-config/src/lib.rs
Dkwcs: Done

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Done

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Done

crates/iota-protocol-config/src/lib.rs
```diff
@@ -316,6 +316,10 @@ struct FeatureFlags {
     // If true enable additional multisig checks.
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
+
+    // Enable passkey auth (SIP-9)
```
miker83z: Please fix the comment here

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -61,10 +61,37 @@ mod checked {
             gas,
             transaction.gas_budget(),
             transaction.gas_price(),
-            transaction.kind(),
+            Some(transaction.kind()),
```
miker83z:
```suggestion
        if transaction.is_system_tx() {
            Ok(IotaGasStatus::new_unmetered())
        } else {
            check_gas(
                objects,
                protocol_config,
                reference_gas_price,
                gas,
                transaction.gas_budget(),
                transaction.gas_price(),
            )
```
let's move this check here, so that you don't need to pass tx_kind to `check_gas`

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -61,10 +61,37 @@ mod checked {
...
         )
     }
 
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_auth_inputs(
+        sender: IotaAddress,
+        sponsor: IotaAddress,
```
miker83z:
```suggestion
        gas_owner: IotaAddress,
```

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -346,31 +373,77 @@ mod checked {
...
-            gas_status.check_gas_balance(&gas_objects, gas_budget)?;
-            Ok(gas_status)
         }
     }
 
+    /// Check all the objects used in authentication
+    #[instrument(level = "trace", skip_all)]
+    fn check_auth_objects(
+        sender: IotaAddress,
+        sponsor: IotaAddress,
```
miker83z:
```suggestion
        gas_owner: IotaAddress,
```

crates/iota-types/src/signature.rs
```diff
@@ -247,6 +252,10 @@ impl ToFromBytes for GenericSignature {
                     let passkey = PasskeyAuthenticator::from_bytes(bytes)?;
                     Ok(GenericSignature::PasskeyAuthenticator(passkey))
                 }
+                SignatureScheme::MoveAuthenticator => {
+                    let passkey = MoveAuthenticator::from_bytes(bytes)?;
```
miker83z: fix "passkey" name

crates/iota-types/src/transaction.rs
```diff
@@ -2269,6 +2270,15 @@ impl SenderSignedData {
                     }
                 }
                 GenericSignature::Signature(_) | GenericSignature::MultiSig(_) => (),
+                GenericSignature::MoveAuthenticator(_) => {
```
miker83z, line 2272: pls move it above `GenericSignature::Signature(_) | GenericSignature::MultiSig(_) => (),`

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,109 @@
...
+    signature::{AuthenticatorTrait, VerifyParams},
+    signature_verification::VerifiedDigestCache,
+    transaction::CallArg,
+};
+
+#[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
+pub struct MoveAuthenticator {
+    /// Input objects or primitive values
+    inputs: Vec<CallArg>,
+    object_id: ObjectID,
```
miker83z:
```suggestion
    object_to_authenticate: ObjectID,
```

crates/iota-transaction-checks/src/lib.rs
miker83z: Given this, then make all gas parameters together, so in order:
- protocol_config: &ProtocolConfig,
- sender: IotaAddress,
- input_objects: InputObjects,
- reference_gas_price: u64,
- gas_owner: IotaAddress,
- gas_budget: u64,
- gas_price: u64,
- gas: &[ObjectRef],

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Done

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Done

crates/iota-types/src/signature.rs
Dkwcs: Done

crates/iota-types/src/transaction.rs
Dkwcs: Done

crates/iota-protocol-config/src/lib.rs
Dkwcs: Done

crates/iota-protocol-config/src/lib.rs
```diff
@@ -316,6 +316,10 @@ struct FeatureFlags {
     // If true enable additional multisig checks.
     #[serde(skip_serializing_if = "is_false")]
     additional_multisig_checks: bool,
+
+    // Enable move auth
```
miker83z: Check the other inline comments above, these are more expressive. I suggest you something like this: "If true, enables the authentication of account using Move code."

miker83z (approved):

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,109 @@
...
+    signature::{AuthenticatorTrait, VerifyParams},
+    signature_verification::VerifiedDigestCache,
+    transaction::CallArg,
+};
+
+#[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
+pub struct MoveAuthenticator {
+    /// Input objects or primitive values
+    inputs: Vec<CallArg>,
+    object_to_authenticate: ObjectID,
```
miker83z: Could you make it an `ObjectRef` instead of a `ObjectID`? The ref is needed for later in the authority

Dkwcs: Done

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,109 @@
...
+    signature::{AuthenticatorTrait, VerifyParams},
+    signature_verification::VerifiedDigestCache,
+    transaction::CallArg,
+};
+
+#[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
+pub struct MoveAuthenticator {
+    /// Input objects or primitive values
+    inputs: Vec<CallArg>,
+    object_ref_to_authenticate: ObjectRef,
```
miker83z:
```suggestion
    /// The reference to the object that this authenticates. This object represents the account being the sender of the transaction.
    object_to_authenticate: ObjectRef,
```

crates/iota-types/src/move_authenticator.rs
```diff
@@ -29,7 +29,7 @@ pub struct MoveAuthenticator {
     inputs: Vec<CallArg>,
     /// The reference to the object that this authenticates. This object
     /// represents the account being the sender of the transaction.
-    object_to_authenticate: ObjectRef,
+    object_ref_to_authenticate: ObjectRef,
```
miker83z: Please revert! `object_to_authenticate` is more appropriate as a name from the point of view of the authenticator.

Dkwcs: Okay
Done

crates/iota-protocol-config/src/lib.rs
```diff
@@ -520,6 +524,9 @@ pub struct ProtocolConfig {
     /// Maximum gas budget in NANOS that a transaction can use.
     max_tx_gas: Option<u64>,
 
+    /// Maximum gas budget in NANOS that a authentification transaction can use.
```
TheMrAI: Typo: `authentification`
Maybe `authentication` ?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1494,6 +1505,8 @@ impl ProtocolConfig {
             max_publish_or_upgrade_per_ptb: Some(5),
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
+            // max gas budget is in NANOS and an absolute value 50IOTA
```
TheMrAI: This was meant to say `10 IOTA`?

crates/iota-types/src/error.rs
```diff
@@ -310,6 +310,9 @@ pub enum UserInputError {
 
     #[error("Invalid identifier found in the transaction: {error}")]
     InvalidIdentifier { error: String },
+
+    #[error("Authenticator input objects should be shared or immutable")]
+    ImmutableOrSharedObjectsExpected,
```
TheMrAI: Could this perhaps give some more context? Which input was faulty?
Consider referring to `Authenticator` as `Move Authenticator`. Making it very clear what is going on.
This way, later on we can write user docs easier by referring to this concept as `Move Authenticator`.

crates/iota-types/src/transaction.rs
```diff
@@ -2268,6 +2269,15 @@ impl SenderSignedData {
                         });
                     }
                 }
+                GenericSignature::MoveAuthenticator(_) => {
+                    if !config.move_auth() {
+                        return Err(IotaError::UserInput {
+                            error: UserInputError::Unsupported(
+                                "Move authentication is not enabled on this network".to_string(),
```
TheMrAI: What do you think?
Could it help the user understand the error better if instead of
`"Move authentication is not enabled on this network"`
it was
`"'Move authenticator' is not enabled on this network"`

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,111 @@
...
+    crypto::SignatureScheme,
+    digests::ZKLoginInputsDigest,
+    error::IotaResult,
+    signature::{AuthenticatorTrait, VerifyParams},
+    signature_verification::VerifiedDigestCache,
+    transaction::CallArg,
+};
+
+#[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
+pub struct MoveAuthenticator {
```
TheMrAI, line 31: Could you describe in more detail what this struct is meant to do?
Now that every information is fresh we should write down all the info.
For example, that is it part of the abstract account implementation, that it is meant to describe what inputs should the function receive that has been claimed from the object_to_authenticate.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1494,6 +1505,8 @@ impl ProtocolConfig {
...
+            max_auth_gas: Some(1_000_000_000),
```
Thoralf-M:
```suggestion
            // max auth gas budget is in NANOS and an absolute value 1IOTA
            max_auth_gas: Some(1_000_000_000),
```

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Input objects or primitive values
+    inputs: Vec<CallArg>,
+    /// The reference to the object that this authenticates. This object
+    /// represents the account being the sender of the transaction.
+    object_to_authenticate: ObjectRef,
+    /// A bytes representation of [struct MoveAuthenticator]. This helps with
+    /// implementing [trait AsRef<[u8]>].
+    #[serde(skip)]
+    bytes: OnceCell<Vec<u8>>,
```
TheMrAI, line 40: Sheesh this is nasty. Not your fault, I see it is done like this in other places just crying inside.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,111 @@
...
+        _zklogin_inputs_cache: Arc<VerifiedDigestCache<ZKLoginInputsDigest>>,
+    ) -> IotaResult
+    where
+        T: Serialize,
+    {
+        Ok(())
+    }
+}
+
+/// Necessary trait for [struct SenderSignedData].
```
TheMrAI: If you want to properly refer to this type using rustdoc, you have two options I know of:
[crate::transaction::SenderSignedData] => You need to specify the whole path, but this will make it appear in the doc as `crate::transaction::SenderSignedData`, which is ugly as heck.
`[SenderSignerData](crate::transaction::SenderSignedData)` => This way it will appear in the doc as `SenderSignerData`.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -0,0 +1,111 @@
...
+
+#[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
+pub struct MoveAuthenticator {
+    /// Input objects or primitive values
+    inputs: Vec<CallArg>,
+    /// The reference to the object that this authenticates. This object
+    /// represents the account being the sender of the transaction.
+    object_to_authenticate: ObjectRef,
+    /// A bytes representation of [struct MoveAuthenticator]. This helps with
+    /// implementing [trait AsRef<[u8]>].
```
TheMrAI: In rustdoc I don't think it is necessary to write either `trait` or `struct` in any linkpoint.
Instead you need to only watch out that what you are referring to is accessible from the current scope, or you have specified the proper path to it (see example below).
Here if you want to refer to the specific trait implementation on `MoveAuthenticator` you should write it like so: `[MoveAuthenticator::as_ref<u8>]`.
If you were to refer to the general AsRef template you could write it like `[std::convert::AsRef]`.

When in doubt run `cargo doc` on your given code and check the result.
Or if you have `rust analyser` it should highlight the appropriately set links as green.

TheMrAI (changes requested):
LGTM, small typos and some documentation links to fix, but I am very happy you wrote some rustdocs! The more the better!

crates/iota-protocol-config/src/lib.rs
Dkwcs: Done

crates/iota-protocol-config/src/lib.rs
Dkwcs: Yes, thanks

crates/iota-protocol-config/src/lib.rs
Dkwcs: Done

crates/iota-types/src/error.rs
Dkwcs: Please take a look on update

crates/iota-types/src/transaction.rs
Dkwcs: For me it looks the same, but we can improve it as you proposed

crates/iota-types/src/move_authenticator.rs
Dkwcs: Comment has been added

crates/iota-types/src/move_authenticator.rs
Dkwcs: Good point, done.

crates/iota-types/src/move_authenticator.rs
Dkwcs: Updated

crates/iota-types/src/transaction.rs
TheMrAI: It is the same, but now we are highlight that we are talking about the 'Move authenticate' concept, so minimally it helps the unassuming reader connecting the dots.
They see the code `MoveAuthenticator`, they see this error, ahh so this is connected to this.

When you say **Move authentication** that is general enough that it can get confusing.
Admittedly even this change may be just marginally helpful, that is why I just proposed it for consideration.

crates/iota-types/src/move_authenticator.rs
TheMrAI: Leave it as is, just pointing it out for future, but this is extremely minor.
```
/// MoveAuthenticator is a new GenericSignature variant that enables a new
/// method of authentication through Move code.
```
There concept of **new** doesn't exist in this context. By the time this code reacher prod it will be old for us, new for someone seeing it for the first time.
```
/// MoveAuthenticator is a [GenericSignature] variant that enables
/// transaction authentication through Move code.
```
Timeless :)

TheMrAI (approved):
Thanks for the updates!

crates/iota-types/src/transaction.rs
Dkwcs: Added

crates/iota-types/src/move_authenticator.rs
Dkwcs: You’re completely right I agree.
It has been updated

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -54,15 +54,44 @@ mod checked {
...
+                reference_gas_price,
+                gas,
+                transaction.gas_budget(),
+                transaction.gas_price(),
+            )
+        }
+    }
+
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_auth_inputs(
```
valeriyr: @miker83z @Dkwcs I propose to revert the changes in `iota-transaction-checks` because a new version is introduced in the following PR:
https://github.com/iotaledger/iota/pull/8167
It will be changed there if it is necessary.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1494,6 +1505,8 @@ impl ProtocolConfig {
             max_publish_or_upgrade_per_ptb: Some(5),
             // max gas budget is in NANOS and an absolute value 50IOTA
             max_tx_gas: Some(50_000_000_000),
+            // max auth gas budget is in NANOS and an absolute value 1IOTA
+            max_auth_gas: Some(1_000_000_000),
```
valeriyr: This initialization needs to be moved to the expected(last) protocol version block.

valeriyr (changes requested):

crates/iota-protocol-config/src/lib.rs
Dkwcs: Yeah, I moved it

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_1.snap
```diff
@@ -275,4 +275,3 @@ checkpoint_summary_version_specific_data: 1
 max_soft_bundle_size: 5
 bridge_should_try_to_finalize_committee: true
 max_accumulated_txn_cost_per_object_in_mysticeti_commit: 10
-
```
valeriyr: Despite only the line ending is changed, I would revert this file; It is not the best practice to have changes in the released snapshots, and it doesn't make any sense in the PR.

valeriyr (approved):

PR #8139, opened 2025-08-07
feat(explorer): add validator address alias

### LINKED ISSUE #7240: Add labeled names to validator addresses
Just like in #7233, we should add labels and images to validator's addresses. Maybe control it with a feature in apps backend

Image

VmMad:

Fixes #7240 

Launch apps-backend along with the explorer to see the aliases


apps/apps-backend/src/features/constants/validatorAlias.constants.ts
```diff
@@ -0,0 +1,46 @@
...
+export type ValidatorAliasMap = Record<string, string>;
+
+export const MAINNET_VALIDATOR_ALIASES: ValidatorAliasMap = {
+    '0x864c651958094732a1227134cf7cab7587f05a399398804552553fbc01dba4e7': 'IOTA 1',
+    '0x8e2e13c2ecfda356f07d008885b7bb82befb4d602245c3bff98ad59863162dd8': 'IOTA 2',
+    '0xde58c0e225e01f890f05b0d2c19ae15b7c337faae539545bfd3a5d8676d62a87': 'IOTA 3',
+};
+
+export const TESTNET_VALIDATOR_ALIASES: ValidatorAliasMap = {
+    '0x392316417a23198afeeb80d9fec314c65162ab5ad18f8a4c3375d31deab29670': 'IOTA Foundation 1',
```
alexsporn: Why are you hardcoding addresses here? This can all be taken from the system state.
For example these validators got renamed on Testnet and now you would be using wrong values.

marc2332 (approved):

apps/core/src/hooks/useAddressAliasLookup.ts
```diff
@@ -1,44 +1,49 @@
...
-                alias: addressAlias,
-            };
-        },
-        [knownAddressesFeature],
+    const knownAddresses = useFeatureValue<KnownAddressAliasesFeature>(
+        Feature.KnownAddressAlias,
+        ADDRESSES_ALIAS_FALLBACK,
+    );
+
+    const { data: systemState } = useIotaClientQuery('getLatestIotaSystemState');
```
alexsporn, line 27: does this automatically call the v2?

alexsporn (approved):
Left a small question, but lgtm

apps/core/src/hooks/useAddressAliasLookup.ts
marc2332: Yep, `getLatestIotaSystemState` will always call whatever is the latest system state version based on the protocol config of the node that is being used. So, yes, it will call v2 if its available :)

If it happens to support up to v1 for example, then some properties of the return value will simply be missing.

msarcev (approved):

PR #8144, opened 2025-08-08
upstream(node): Remove keys reference in typed-store

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/7a079dc8569d6beabf9bfbd08bdd62fa27b90e82

NOTE: We keep the `test_db.rs` to test the remaining functionalities.

Part of #5727 


crates/typed-store/src/test_db.rs
```diff
@@ -1,841 +0,0 @@
-// Copyright (c) Mysten Labs, Inc.
```
muXxer: Sorry that I didn't state it before, but I didn't see that this file get's removed. IMO we should keep it, and only remove the keys and value and iterator parts that might not work anymore because they got removed by the 3 PRs. 
The iterator parts maybe can be replaced the same way it is done in the rest of the code.

muXxer (changes requested):

muXxer (approved):

bingyanglin (approved):

PR #8157, opened 2025-08-09
feat(starfish): improving transaction synchronizer

### LINKED ISSUE #8147: starfish: change the logic and parameters for transaction synchronizer
After a restart, a node could have many missing transactions which are all attempted to be fetched. It could be that the transaction synchronizer attempts to fetch too many transaction data and the locks are never released leading to a problem of non-synced committed subdag

polinikita:
# Description of change

In this PR, we improve the transaction synchronizer in Starfish. Main changes:
- we join fetching and processing transactions into one method `fetch_and_process_transactions_from_authorities` to reuse this for live and periodic syncing
- we make at most one fetch request to a given authority in one call of `fetch_and_process_transactions_from_authorities`
- we track the number of concurrent requests to each authority and limit that value using a new struct in synchronizer called `active_requests`
- we process transactions after fetching for each authority individually. While this creates much more calls to the core thread, this improves the syncing process significantly since it was slowed down by the slowest response in the previous implementation. Later on we can add one/a few active thread(s) that will take all fetched bytes and process them, removing duplicates and sending processed transactions to the core thread on a regular basis.
- we changed some parameters for fetching transactions, but they might require further changes once the protocol is tested on a larger network with longer discrepancies. 

Fixes #8147 

### Release Notes

- [x] Protocol: Improvements in Starfish tx synchronizer
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/transactions_synchronizer.rs
+++ b/crates/starfish/core/src/transactions_synchronizer.rs
@@ -451,7 +451,11 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
 
         loop {
             // Wait for a permit asynchronously
-            let permit = semaphore.clone().acquire_owned().await.expect("We expect semaphore to be valid");
+            let permit = semaphore
+                .clone()
+                .acquire_owned()
+                .await
+                .expect("We expect semaphore to be valid");
 
             match receiver.recv().await {
                 Some(missing_transactions_block_refs) => {
@@ -475,7 +479,7 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
                             dag_state,
                             SyncMethod::Live,
                         )
-                            .await;
+                        .await;
 
                         // Release the permit when done
                         drop(permit);
--- a/crates/starfish/core/src/transactions_synchronizer.rs
+++ b/crates/starfish/core/src/transactions_synchronizer.rs
@@ -65,21 +65,36 @@ const FETCH_AND_PROCESS_FROM_PEERS_TIMEOUT: Duration = Duration::from_millis(700
 /// given block ref.
 const MAX_AUTHORITIES_TO_FETCH_PER_TRANSACTION: usize = 3;
 
+#[derive(Debug,Clone,Copy,Ord,Eq,PartialOrd,PartialEq)]
+enum SyncMethod {
+    Live,
+    Periodic,
+}
+impl SyncMethod {
+    fn get_string(&self) -> String {
+        match self {
+            SyncMethod::Live => "live",
+            SyncMethod::Periodic => "periodic",
+        }
+        .to_string()
+    }
+}
+
 struct ActiveRequestGuard {
     authority: AuthorityIndex,
-    sync_method: String,
-    active_requests: Arc<Mutex<BTreeMap<(AuthorityIndex, String), usize>>>,
+    sync_method: SyncMethod,
+    active_requests: Arc<Mutex<BTreeMap<(AuthorityIndex, SyncMethod), usize>>>,
 }
 
 impl ActiveRequestGuard {
     fn new(
         authority: AuthorityIndex,
-        sync_method: String,
-        active_requests: Arc<Mutex<BTreeMap<(AuthorityIndex, String), usize>>>,
+        sync_method: SyncMethod,
+        active_requests: Arc<Mutex<BTreeMap<(AuthorityIndex, SyncMethod), usize>>>,
     ) -> Self {
         {
             let mut map = active_requests.lock();
-            *map.entry((authority, sync_method.clone())).or_insert(0) += 1;
+            *map.entry((authority, sync_method)).or_insert(0) += 1;
         }
         Self {
             authority,
@@ -92,7 +107,7 @@ impl ActiveRequestGuard {
 impl Drop for ActiveRequestGuard {
     fn drop(&mut self) {
         let mut map = self.active_requests.lock();
-        if let Some(val) = map.get_mut(&(self.authority, self.sync_method.clone())) {
+        if let Some(val) = map.get_mut(&(self.authority, self.sync_method)) {
             *val = val.saturating_sub(1);
         }
     }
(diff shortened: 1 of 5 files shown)


crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -133,19 +144,33 @@ impl InflightTransactionsMap {
...
+                }
             }
         }
     }
 
+    /// Cleans up the old locks that are older than `max_age`. This is useful
+    /// to remove the locks that are not used anymore. The latter could happen
+    /// when the timeout occurred and the locks are forgotten to be
+    /// unlocked.
+    fn cleanup_old_locks(self: &Arc<Self>, max_age: Duration) {
```
piotrm50: How can it happen that it's forgotten? If we remove it from the map here, is `drop()` called at some later point in time? If not then this could indicate that we might have memory leak as the non-dropped object is still held in memory. Can't we drop the TransactionGuard when the request times out instead of cleaning the whole map periodically?

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -865,61 +894,59 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
...
+            let selected_block_refs: BTreeSet<_> = authority_block_refs
+                .iter()
+                .filter(|b| {
+                    assigned_block_refs_with_multiplicity
+                        .get(*b)
+                        .copied()
+                        .unwrap_or(0)
+                        < MAX_ASSIGNMENTS_PER_BLOCK_REF_PER_RUN
+                })
+                .take(context.parameters.max_transactions_per_fetch)
```
piotrm50: Here we will always take the first BlockRefs. Shouldn't we randomize it to avoid situation where the first elements are fetched from maximum number of peers while the last ones might not be fetched in extreme scenario?

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -865,61 +894,59 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
...
-                    .collect::<BTreeSet<_>>();
 
-                if selected_block_refs.is_empty() {
-                    break;
-                }
+            // Filter: keep only blocks that have not reached the per-block limit
+            let selected_block_refs: BTreeSet<_> = authority_block_refs
+                .iter()
+                .filter(|b| {
+                    assigned_block_refs_with_multiplicity
```
piotrm50: `assigned_block_refs_with_multiplicity` is local to the call, so periodic syncer still can send requests above this limit. I think this logic is already covered by `InflightTransactionsMap`, isn't it?

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -928,12 +955,16 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
...
+        let number_of_requests: i64 = request_futures
+            .len()
+            .try_into()
+            .expect("We should expect a correct conversion from request_futures.len() to i64");
         context
             .metrics
             .node_metrics
             .transactions_synchronizer_inflight_requests
-            .set(request_futures.len() as i64);
+            .add(number_of_requests);
```
piotrm50: This should be `set` to reflect the number of inflight requests at a given point in time. Also, looks like we should make sure that this metric is tracked correctly overall in other places as it looks a bit suspicious now that I look at it.

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -970,6 +1001,13 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
             }
         }
 
+        // Decrease the number of concurrent requests
+        context
```
piotrm50: This sub is definitely not correct because we're **setting** the metric value 10 lines above, so after we subtract we'll definitely end up having negative values.

piotrm50 (changes requested):

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Changed to avoid this local threshold. Selecting is done inside locking now.

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Yes, this is done on purpose. The first ones are more important to progress available commits/checkpoints. For now, I want to ensure that the concept works as it should. I don't like randomization at this stage as I limit the number of fetch request in this PR and the first ones could be not fetched at al.

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Good catch. Yeah, I overlooked the above set. This method is called from live and periodic paths. So, set does not represent the correct value. I change the set values to reflect correctly

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Left only inc() in the internal function and dec() is done automatically whenever we drop the corresponding guard. Having just inc() and dec() might be not good if the outside timeout expires.

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Yeah, changed to the suggested approach

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -43,16 +43,61 @@ use crate::{
...
+struct ActiveRequestGuard {
+    authority: AuthorityIndex,
+    sync_method: String,
+    active_requests: Arc<Mutex<BTreeMap<(AuthorityIndex, String), usize>>>,
+}
+
+impl ActiveRequestGuard {
+    fn new(
+        authority: AuthorityIndex,
+        sync_method: String,
```
VorobyevIlya: would it be better to use enum instead of string?

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -375,55 +423,33 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
...
-                            "live",
-                       )
-                       .await
-                       {
-                           warn!(
-                               "Error occurred while processing fetched blocks from peer {peer}: {err}"
-                           );
-                       }
-                    };
+                Some(_) = requests.next() => {
```
piotrm50: If we're not using the result returned from the future in any way, do we even need to handle it? Maybe we could get rid of the `tokio::select` macro if we don't care about the result.

VorobyevIlya (approved):

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Good idea. This will improve the overall readability. I have addressed this in the next commit

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Replaced the logic to Semaphore with a limited capacity

piotrm50 (approved):

PR #8131, opened 2025-08-07
feat(iota-execution): add a `validate_transaction` function

valeriyr:
# Description of change

A new function has been added to `Executor` that implements smart account transaction verification by calling the related Move authenticator.

It is intended to be called during the processing of a smart account transaction to unlock the corresponding smart account and allow the transaction to be executed.

This function is integrated in the following PR: https://github.com/iotaledger/iota/pull/8167

## Links to any relevant issues

fixes #8119

PR #8167, opened 2025-08-11
feat: validate a transaction with a `MoveAuthenticator` signature schema

### LINKED ISSUE #8120: [AA][auth] Add a validate_transaction() in AuthorityState and connect checks and execution
* `validate_transaction()` must be implemented within `AuthorityState` since it requires access to on-chain objects.
* `signature_verifier::verify_tx` invokes `GenericSignature::verify_authenticator`, which must always succeed for `MoveAuthenticator`.
* Prior to entering `AuthorityState`, in `ValidatorService::handle_transaction` and `ValidatorService::handle_certificates`, the system must detect if the signature is a `MoveAuthenticator` and pass this information downstream.
* In `AuthorityState::handle_transaction` and `AuthorityState::process_certificate`, transaction input objects are read from state and passed to `iota-transaction-checks`. This same flow must be extended to `MoveAuthenticator`.
* From the `MoveAuthenticator`, inputs must be extracted similarly to `tx_data.input_objects()?`, followed by loading those via `self.input_loader.read_objects_for_signing`.
* These loaded inputs are then passed to `iota_transaction_checks::check_authenticator_input` (only input objects are relevant, not receiving ones).
* Before proceeding, the sender must be verified as an `Account` object. This means confirming the object with `object_id == sender` has a dynamic field keyed with `IOTA_AUTHENTICATION`, holding an `AuthenticatorInfoV1`.
* After successful checks, the data is passed to `Executor::validate_transaction` for evaluation.
* An IotaGasStatus is created during the checks in iota-transaction-checks; this must be considered for the execution of the auth and then the TX.
  - after the IotaGasStatus instance has been created for the execution of authenticate(), then keep track of the paid gas cost and destroy the instance; when iota_transaction_checks::check_transaction_input is executed for the main TX then a new instance of IotaGasStatus is created; from this it must be deducted the authentication gas cost.
  - the IotaGasStatusAPI can be extended with a function charge_authentication (similar to charge_storage_read). This can be then be called when the main TX instance of IotaGasStatus is created, before checking the gas coin (should be around [here](https://github.com/iotaledger/iota/blob/9ab424d780ea84940d3b5048a786881cc145db91/crates/iota-transaction-checks/src/lib.rs#L356)).

valeriyr:
# Description of change

Integrated the `Executor:: validate_transaction` function implemented in the following PR: https://github.com/iotaledger/iota/pull/8131

fixes #8120


crates/iota-transaction-checks/src/lib.rs
```diff
@@ -190,6 +193,46 @@ mod checked {
         Ok(input_objects.into_checked())
     }
 
+    #[instrument(level = "trace", skip_all)]
+    pub fn check_move_authenticator_input(
+        protocol_config: &ProtocolConfig,
+        reference_gas_price: u64,
+        transaction: &TransactionData,
```
miker83z: Let's pass the single elements instead of the TransactionData, this is more clear for the caller:
```suggestion
        gas_owner: IotaAddress,
        gas_budget: u64,
        gas_price: u64,
        gas: &[ObjectRef],
```

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -570,6 +608,188 @@ mod checked {
...
+    /// Check one `MoveAuthenticator` input object.
+    fn check_one_move_authenticator_object(
+        object_kind: InputObjectKind,
+        object: &Object,
+        gas_owner: IotaAddress,
+        is_gas_object: bool,
+    ) -> UserInputResult {
+        match object_kind {
+            InputObjectKind::MovePackage(package_id) => {
+                // TODO: Is this check correct?
```
miker83z: Yes

crates/iota-types/src/move_authenticator.rs
```diff
@@ -69,6 +69,17 @@ impl MoveAuthenticator {
         &self.type_arguments
     }
 
+    pub fn object_to_authenticate(&self) -> &ObjectRef {
+        &self.object_to_authenticate
```
miker83z, line 74: `object_to_authenticate` should be a `CallArg`.

crates/iota-core/src/authority.rs
```diff
@@ -4979,6 +5181,158 @@ impl AuthorityState {
         Ok(new_epoch_store)
     }
 
+    /// Checks if `account` is a valid smart account and returns the
+    /// account-related `AuthenticatorInfo` object.
+    fn check_smart_account(&self, account: &ObjectRef) -> IotaResult<AuthenticatorInfo> {
+        let account_object_id = account.0;
```
miker83z: After [this change](https://github.com/iotaledger/iota/pull/8167/files#r2285549600) we have the following checks:
- If the CallArg is `Pure` abort
- If the CallArg is `ObjectArg::Receiving(_)` abort
- If the CallArg is `ObjectArg::SharedObject(obj)` then check
    - that obj is mutable
    - that the retrieved object Exists, is_shared, initial_shared_version matches.
- If the CallArg is `ObjectArg::ImmOrOwnedObject(obj_ref)` then check
    - that the retrieved object Exists, !is_shared, !is_immutable
    - then version and digest matches

Could this be moved to iota-transaction-checks? The `object_to_authenticate` is an input of the MoveAuthenticator but it is, maybe, not used in the authenticate() function.

crates/iota-core/src/authority.rs
```diff
@@ -4979,6 +5161,176 @@ impl AuthorityState {
...
+    ) -> IotaResult<AuthenticatorInfo> {
+        let (account_object_id, account_object_seq_number, account_object_digest) =
+            authenticator.object_to_authenticate_components()?;
+
+        let account_object = self.get_object_read(&account_object_id)?;
+
+        match account_object {
+            ObjectRead::Exists(_, object, _) => {
+                fp_ensure!(
+                    object.is_shared() || object.is_immutable(),
```
valeriyr: If we want to allow owned objects to be accounts, we can do this here.

valeriyr: @miker83z, where and how is it guaranteed that the object ID here is the transaction sender? Should we check it before signing?

crates/iota-types/src/transaction.rs
```diff
@@ -2342,6 +2370,23 @@ impl SenderSignedData {
 
         Ok(tx_size)
     }
+
+    // TODO: A temporary created function. Needs to be replaced with a proper check.
+    pub fn move_authenticator(&self) -> Option<&MoveAuthenticator> {
+        let signatures = self.tx_signatures();
+
+        if signatures.len() == 1 && signatures[0].is_move_authenticator() {
+            match &signatures[0] {
```
TheMrAI: Are the other cases important? It doesn't seems so, you could replace
the whole `match` expression with a 
            
```suggestion
if let GenericSignature::MoveAuthenticator(move_authenticator) = &signatures[0] {
    Some(move_authenticator)
} else {
    None
}
```

In any case the unreacheable is suspect. From the authority I see it being called without any additional checks.

crates/iota-core/src/authority.rs
```diff
@@ -1207,9 +1261,19 @@ impl AuthorityState {
...
             return Ok((effects, None));
         }
-        let input_objects =
+
+        let tx_input_objects =
             self.read_objects_for_execution(tx_guard.as_lock_guard(), certificate, epoch_store)?;
 
+        // TODO: Since the authenticator input is loaded with a dedicated call, we
+        // can load the same object twice if it is presented in the transaction and
+        // authenticator inputs.
```
TheMrAI, line 1269: Is that an issue? Isn't there some type of caching preventing it actually getting loaded twice?

crates/iota-types/src/account.rs
```diff
@@ -2,12 +2,69 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
```
TheMrAI: Shouldn't this be `Copyright (c) IOTA Stiftung`?

crates/iota-types/src/gas.rs
```diff
@@ -108,6 +125,32 @@ pub mod checked {
                 Self::V1(status) => status.check_gas_balance(gas_objs, gas_budget),
             }
         }
+
+        fn check_gas_preconditions(
```
TheMrAI, line 132: Nice refactor!

crates/iota-types/src/gas_model/tables.rs
```diff
@@ -79,10 +79,21 @@ impl GasStatus {
     /// Charge for every operation and fail when there is no more gas to pay for
     /// operations. This is the instantiation that must be used when
     /// executing a user script.
-    pub fn new(cost_table: CostTable, budget: u64, gas_price: u64, gas_model_version: u64) -> Self {
+    pub fn new(
+        cost_table: CostTable,
+        budget: u64,
+        gas_spent: u64,
```
TheMrAI: Couldn't we just call this with `budget-gas_spent`, instead of the `gas_spent` being passed?
It looks so alien.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -570,6 +645,133 @@ mod checked {
...
+                // reading it.
+                ObjectReadResultKind::CancelledTransactionSharedObject(_) => (),
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Check one `MoveAuthenticator` input object.
+    fn check_one_move_authenticator_object(
```
TheMrAI, line 673: Can we refactor around here?
`check_one_object` seems to be mostly the same.
It might not make sense though, not sure what we are doing.

crates/iota-types/src/error.rs
```diff
@@ -311,10 +316,56 @@ pub enum UserInputError {
     #[error("Invalid identifier found in the transaction: {error}")]
     InvalidIdentifier { error: String },
 
+    // `MoveAuthenticator` related errors
+    #[error("Account object {object_id:?} not found")]
```
TheMrAI: I very much like your errors.

crates/iota-core/src/authority.rs
```diff
@@ -4979,6 +5161,176 @@ impl AuthorityState {
...
 
+    /// Checks if `authenticator` unlocks a valid smart account and returns the
+    /// account-related `AuthenticatorInfo` object.
+    fn check_smart_account(
+        &self,
+        authenticator: &MoveAuthenticator,
+    ) -> IotaResult<AuthenticatorInfo> {
+        let (account_object_id, account_object_seq_number, account_object_digest) =
+            authenticator.object_to_authenticate_components()?;
+
```
miker83z, line 5239: It should check here that `account_object_id == sender`

crates/iota-core/src/authority.rs
miker83z: https://github.com/iotaledger/iota/pull/8167/files#r2293973953

TheMrAI (commented):
Honestly this is a bit beyond me.
I see that authority is generally extremely messy and a number of checks we have to duplicate as well.

The gas check is what I don't like for sure. That somehow feels very wrong, but I can't say if we could properly integrate it into Tx or if there is a better way.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -69,12 +70,69 @@ impl MoveAuthenticator {
         &self.type_arguments
     }
 
+    pub fn object_to_authenticate(&self) -> &CallArg {
+        &self.object_to_authenticate
+    }
+
+    pub fn object_to_authenticate_components(
+        &self,
+    ) -> IotaResult<(ObjectID, SequenceNumber, Option<ObjectDigest>)> {
```
miker83z:
```suggestion
    ) -> IotaResult<(ObjectID, Option<SequenceNumber>, Option<ObjectDigest>)> {
```
The `initial_shared_version` and the object.version() for shared objects are two different versions. So this should return (ObjectId, None, None) when it matches `ObjectArg::SharedObject`.

crates/iota-core/src/authority.rs
```diff
@@ -4979,6 +5161,176 @@ impl AuthorityState {
...
+                fp_ensure!(
+                    object.is_shared() || object.is_immutable(),
+                    UserInputError::AccountObjectNotSupported {
+                        object_id: account_object_id
+                    }
+                    .into()
+                );
+
+                fp_ensure!(
+                    object.version() == account_object_seq_number,
```
miker83z: This must be checked only in the case of ImmOrOwned object, not for shared objects.

miker83z: See https://github.com/iotaledger/iota/pull/8167/files#r2294001138

crates/iota-core/src/authority.rs
```diff
@@ -4979,6 +5161,176 @@ impl AuthorityState {
         Ok(new_epoch_store)
     }
 
+    /// Checks if `authenticator` unlocks a valid smart account and returns the
+    /// account-related `AuthenticatorInfo` object.
+    fn check_smart_account(
```
miker83z:
```suggestion
    /// Checks if `authenticator` unlocks a valid account and returns the
    /// account-related `AuthenticatorInfo` object.
    fn check_account(
```

crates/iota-core/src/authority.rs
```diff
@@ -863,57 +865,109 @@ impl AuthorityState {
...
+
+            if let Err(validation_error) = validation_result {
+                return Err(IotaError::MoveAuthenticatorExecutionFailure {
+                    error: validation_error.to_string(),
+                });
+            }
+        }
+
+        let (_gas_status, tx_checked_input_objects) =
             iota_transaction_checks::check_transaction_input(
```
miker83z, line 949: Could you add a todo that says that this should be changed in the future to take into account the fact that the gas coins shouldn't be checked again if the GenericSignature is of type MoveAuthenticator?

crates/iota-core/src/authority/shared_object_version_manager.rs
```diff
@@ -153,8 +153,21 @@ impl SharedObjVerManager {
...
+                move_authenticator.shared_objects().into_iter()
+            } else {
+                Vec::new().into_iter()
+            };
+
         // Make an iterator to update the locks of the transaction's shared objects.
-        let shared_input_objects: Vec<_> = cert.shared_input_objects().collect();
+        let shared_input_objects: Vec<_> = cert
+            .shared_input_objects()
+            .chain(authenticator_shared_objects)
```
miker83z: In here, it must remove duplicates found in both `shared_input_objects` and `authenticator_shared_objects`

crates/iota-core/src/authority/shared_object_version_manager.rs
```diff
@@ -153,8 +153,21 @@ impl SharedObjVerManager {
             };
         let txn_cancelled = cancellation_info.is_some();
 
+        // Add the Move authenticator shared objects if any.
+        //
+        // TODO: Make sure this is correct.
+        let authenticator_shared_objects =
+            if let Some(move_authenticator) = cert.move_authenticator() {
+                move_authenticator.shared_objects().into_iter()
```
miker83z: I think this should also include the `object_to_authenticate` if it is a shared object.

crates/iota-core/src/authority.rs
valeriyr: If you don't like the name "smart", lets use another name, it would be nice to emphesize that it is not just an account but a programmable one

crates/iota-core/src/authority.rs
valeriyr: added here 9471576965

crates/iota-core/src/authority.rs
valeriyr: added here 9471576965

crates/iota-types/src/transaction.rs
valeriyr: I didn't pay a lot of attention here because I suppose this function is temporary and will be removed.
But I like your option, it fixed here: 6132898210

crates/iota-types/src/account.rs
valeriyr: Good catch! Yes, you are right, it is also a temporarily created file that needs to be fully removed once the real type is merged.
Just in case, I fixed it here: 6132898210

crates/iota-core/src/authority.rs
valeriyr: I am not 100% sure about this. For example, the function that loads objects for signing has a `_tx_digest_for_caching` parameter that is supposed to be used for caching:
https://github.com/iotaledger/iota/blob/6132898210d122605f2d37b70d4c07501be3caf4/crates/iota-core/src/transaction_input_loader.rs#L44
but it is not implemented:
```rust
/// Read the inputs for a transaction that the validator was asked to sign.
///
/// tx_digest is provided so that the inputs can be cached with the
/// tx_digest and returned with a single hash map lookup when
/// notify_read_objects_for_execution is called later. TODO: implement
/// this caching
pub fn read_objects_for_signing(
        &self,
        _tx_digest_for_caching: Option<&TransactionDigest>,
        input_object_kinds: &[InputObjectKind],
        receiving_objects: &[ObjectRef],
        epoch_id: EpochId,
    ) -> IotaResult<(InputObjects, ReceivingObjects)> { ... }
```

In the [read_objects_for_execution](https://github.com/iotaledger/iota/blob/6132898210d122605f2d37b70d4c07501be3caf4/crates/iota-core/src/transaction_input_loader.rs#L140) function, it is not implemented as well according to the comment, but it looks outdated:

```rust
    /// Read the inputs for a transaction that is ready to be executed.
    ///
    /// epoch_store is used to resolve the versions of any shared input objects.
    ///
    /// This function panics if any inputs are not available, as
    /// TransactionManager should already have verified that the transaction
    /// is ready to be executed.
    ///
    /// The tx_digest is provided here to support the following optimization
    /// (not yet implemented): All the owned input objects will likely have
    /// been loaded during transaction signing, and can be stored as a group
    /// with the transaction_digest as the key, allowing the lookup to
    /// proceed with only a single hash map lookup. (additional lookups may be
    /// necessary for shared inputs, since the versions are not known at
    /// signing time). Receiving objects could be cached, but only with
    /// appropriate invalidation logic for when an object is received by a
    /// different tx first.
    pub fn read_objects_for_execution(
        &self,
        tx_key: &TransactionKey,
        // Important to hold the _tx_lock, otherwise it would be possible for a concurrent
        // execution of the same tx to enter this point after the first execution has
        // finished and the shared locks have been deleted.
        _tx_lock: &CertLockGuard,
        input_object_kinds: &[InputObjectKind],
        epoch_id: EpochId,
    ) -> IotaResult<InputObjects> { ... }
```
`tx_key` is used for logging and resolving versions of shared objects used in the related transaction.

valeriyr: So I left this TODO to check it later when it is possible to run a transaction with an authenticator.

crates/iota-types/src/gas_model/tables.rs
valeriyr: It was the first thing I tried to do, but the `budget` is stored as a dedicated field [initial_budget](https://github.com/iotaledger/iota/blob/6132898210d122605f2d37b70d4c07501be3caf4/crates/iota-types/src/gas_model/tables.rs#L107) that will be different from the transaction budget in this case.

So I decided to create an `IotaGasStatus` instance like this, which means that gas has been charged for something before execution, and we don't change the context in which the transaction is executed. But I found one more problem with this, the following assert will fail:
https://github.com/iotaledger/iota/blob/6132898210d122605f2d37b70d4c07501be3caf4/iota-execution/latest/iota-adapter/src/execution_engine.rs#L540
I will try to fix the problem since and we can check if we like this approach or if we would like to change it to something else.

One more option here to create a [GasCharger](https://github.com/iotaledger/iota/blob/6132898210d122605f2d37b70d4c07501be3caf4/iota-execution/latest/iota-adapter/src/gas_charger.rs#L36) and charge the gas using it, but it is a versioned(cutted) internal `iota-adapter` type that is not supposed to be used here.

crates/iota-transaction-checks/src/lib.rs
valeriyr: They have some common elements; maybe we can consider refactoring them later, when the solution is ready.
For now, it is better to leave them separated because in this case, we can change them independently if we need.

crates/iota-core/src/authority.rs
valeriyr: fixed in e2e1b7beb7

crates/iota-types/src/move_authenticator.rs
valeriyr: fixed in e2e1b7beb7

crates/iota-core/src/authority.rs
TheMrAI: Maybe `check_move_account`, `check_abstract_account`?
I think either would work as we use both denominations for the abstract account implementation.
I might go with the first one, because that signals that the account is in move, but meh.

valeriyr: @miker83z @TheMrAI I will follow your decision. I don't really have a strong opinion about this. I think we need to have a name that we follow in all parts related to this topic.

We used the following description [here](https://github.com/iotaledger/iota/blob/e2e1b7beb7ba54057f36f5a97e4c885b3634371d/iota-execution/latest/iota-adapter/src/execution_engine.rs#L245):
```rust
/// This function implements an abstracted IOTA account transaction
/// validation. This validation checks that the authentication method used
/// for the account is valid. It prepares a `MoveAuthenticator` PTB with
/// a single move call for execution, then executes it through an inner
/// execution method. The `MoveAuthenticator` provides the inputs to use for
/// the authentication function found in `AuthenticatorInfo`, that is
/// retrieved from the abstracted IOTA account.
```
In this description is not really clear to me: From what is an IOTA account abstracted?
`abstract` seems to be too abstract.

Maybe a good name can be: `check_iota_move_account`, for example.

crates/iota-core/src/authority.rs
valeriyr: added d9eebed088

crates/iota-core/src/authority/shared_object_version_manager.rs
valeriyr: implemented here: 8881a77275

`unique` requires `SharedInputObject` implement `Clone + Eq + Hash`, so I needed to add additional traits that looks ok.

crates/iota-core/src/authority/shared_object_version_manager.rs
valeriyr: I am not sure.
These versions are used for inputs loading [here](https://github.com/iotaledger/iota/blob/8881a77275879eefd6121898223d1d665eb20261/crates/iota-core/src/transaction_input_loader.rs#L140), but we load `object_to_authenticate` differently, actually we don't check a shared object version there. `object_to_authenticate` is not really included in the transaction inputs if it is not added explicitly.

If we add it as an input, we need to load it as an input(for signing and executing) as well, right?

crates/iota-core/src/authority/shared_object_version_manager.rs
valeriyr: I found several more places where the shared objects are taken from a certificate, so I moved this logic inside the related function: 89fa315345

It looks ok, because certificate shared objects logically contain the authenticator ones.

But at the same time, I found a lot of places where `TransactionData::shared_input_objects` is used, so probably we will need to check them.

crates/iota-core/src/authority.rs
TheMrAI: I would go towards the direction of calling it `move account`. As you say abstract doesn't tell us much.
The only thing I am against is adding `iota` as a prefix, suffinx, infix to anything.
The code is chock full of it and in the overwhelming majority of cases it is just noise.

TheMrAI (approved):

crates/iota-core/src/authority.rs
valeriyr: Ok, I renamed it here: 60a6ff1700
Let me know if you have any other preferences for the name.

crates/iota-types/src/gas_model/tables.rs
valeriyr: I improved the implementation here: c3569e19f3

crates/iota-core/src/authority.rs
miker83z: Ok, later we can rename these if we come up with a fancy PR name!

crates/iota-core/src/authority/shared_object_version_manager.rs
miker83z: @valeriyr 
> If we add it as an input, we need to load it as an input(for signing and executing) as well, right?

If the `MoveAuthenticator` explicitly makes it an input then it is fine. If not, we should implicitly treat it as input (that won't be used during execution) because we need to be sure that all the validators lock the same version of the account shared object.

valeriyr: And it means that the `check_move_account` function must load the specific account object version(now, it loads the latest version by ID), because otherwise, we can use a wrong `AuthenticatorInfo` instance.

miker83z: Yes, I think in the case of `process_certificate` the function `read_authenticator_objects_for_execution` must return the specific account object version. Maybe `check_move_authenticator_inputs_for_signing` must do it in the other case.

valeriyr: @miker83z the implementation is improved in 057c8d0de2
Please let me know your thoughts.

crates/iota-core/src/authority.rs
```diff
@@ -5171,11 +5197,36 @@ impl AuthorityState {
...
+        account_object: Option<ObjectReadResult>,
         signer: &IotaAddress,
     ) -> IotaResult<AuthenticatorInfo> {
+        // Check basic `object_to_authenticate` preconditions and get its components.
         let (account_object_id, account_object_seq_number, account_object_digest) =
             authenticator.object_to_authenticate_components()?;
 
+        // Since the `object_to_authenticate` components are provided, it is supposed
+        // that the account object is loaded.
+        let account_object = account_object.expect("Account object must be provided");
```
miker83z: Looks nice the new flow! Just a minor detail here: it can be avoided to pass an Option to `check_move_account` and have this check "Account object must be provided" before calling `check_move_account`.

miker83z (approved):

crates/iota-core/src/authority.rs
valeriyr: The main idea here is that we can check the loaded object only after a `object_to_authenticate_components` call.
I pushed a new, a bit refactored version, so you can check which one you like more: 35441df5c6

PR #8169, opened 2025-08-11
feat(iota-framework)!: Add create_auth_info_v1 function

### LINKED ISSUE #8117: [AA][auth] Create an Account pointing to an `authenticate()` function
This part involves the development of an `AuthenticatorInfo` Move struct, a `create_auth_info` native function and the creation of a new iota-verifier check.

- The `verify_authenticate_function` function
  - The verifier is enriched with a new check dedicated to the authenticate function
  - The check has as input a Compiled module and a function name (or similar)
  - It checks that the function:
    - must only have inputs that are read-only object references or pure types (checking if an object is shared or owned or immutable is not in the semantics of the verifier, thus probably this check cannot happen here)
    - must have no return type 
    - must have an input of type AuthContext as the second to last parameter (see #8121, but it shouldn't block this task)
    - must have an input of type TxContext as last parameter.
- The `create_auth_info_v1` function allows any developer to create any kind of type for representing an Account, by checking the passed inputs through the `verify_authenticate_function` and then packing an `AuthenticatorInfoV1`.
- The `AuthenticatorInfoV1` type
  - In version 1 it uses 3 fields to point exactly to an on-chain function
  ```
  public struct AuthenticatorInfoV1 has store {
    package id: address,
    module: String,
    function name: String,
  } 
  ```

### LINKED ISSUE #8415: Provide a `create_auth_info_self` function
`create_auth_info_v1` has the current signature:
```
  package: ID, 
  module_name: String,
  function_name: String): AuthenticatorInfoV1
```
where `package` ID is the **Storage ID** of the package the function should be found.

This function cannot be used in cases where the user wants to define an `authenticate` function within the same package they are currently writing.
There is no compiler support for this, and may not be worth implementing as the address of the package is only generated before publishing, and we would have to edit the either the source or some additional data that would describe this relationship.

To support this pattern we will introduce a function called
`create_auth_info_self` with the signature:
```
public fun create_auth_info_self( 
  module_name: String,
  function_name: String): AuthenticatorInfoV1
```

This function will have the exact same requirements as the non self version except that it will look for the module in the current package/module.
This may not be entirely trivial to do, further investigation of package/module loading is warranted.

TheMrAI:
# Description of change

Introduce `create_auth_info_v1` and `create_auth_info_self_v1` native functions.
These functions provide the capability of creating an `AuthenticatorInfoV1` pointing to a desired `authenticate`
function.

With the temporary `account_auth_example` unit tests:

```
cargo run -p iota -- move test -p account_auth_example
```

Unfortunately the features aren't fully tested as it isn't entirely clear how they could be. The newly introduced
verifier is more or less fully tested, with `create_auth_self_v1`.

But even this is a limited test as it doesn't mimic proper deployment scenarios.
Both `create_auth_info_v1` and `create_auth_info_self_v1` may be only tested in e2e tests as the currently available testing frameworks aren't necessarily prepared for these scenarios.

fixes #8117 
fixes #8415 

### Release Notes

- [x] Protocol: Introduce new move native function `create_auth_info_v1` and `create_auth_info_self_v1` with its gas costs.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -0,0 +1,24 @@
...
+
+use std::ascii::String;
+
+// const AUTHENTICATOR_ID: vector<u8> = b"IOTA_AUTHENTICATION";
+
+#[allow(unused_field)]
+public struct AuthenticatorInfoV1 has store, drop {
+  // this is actually storage id
+  package: address,
+  module_name: String, // `module` is a keyword in move and the parser unfortunately isn't smart enough to realize that one couldn't start a module declaration within a structure, or function argument etc.
```
miker83z: Alright, could you then also make it coherent for the function? I.e., `function_name`?

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -0,0 +1,24 @@
...
+module iota::account;
+
+use std::ascii::String;
+
+// const AUTHENTICATOR_ID: vector<u8> = b"IOTA_AUTHENTICATION";
+
+#[allow(unused_field)]
+public struct AuthenticatorInfoV1 has store, drop {
+  // this is actually storage id
+  package: address,
```
miker83z: What are the implications of making this an `ID` instead of an `address`?

TheMrAI: Haven't considered it. Followed the specifications so far.

crates/iota-framework/packages/iota-framework/sources/account.move
TheMrAI: Of course of course, do not a worry. We experimenting here.

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: Could you please check? We experimenting here. 😉

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,141 @@
...
+        )));
+    };
+
+    if function_definition.is_entry {
+        return Err(verification_failure(format!(
+            "Authenticator function '{function_identifier}' cannot be marked as `entry`"
+        )));
+    }
+
+    if function_definition.visibility != Visibility::Public {
```
miker83z, line 63: Not for this first iteration of development, but we should consider whether it's worth relax this constraint to allow public and private:
- we can execute private functions from the rust side
- a dev, setting this function as private, means that this is declared as not taking part to composability with other `authenticate()` (same logic as using just `entry` for normal functions)

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    // Apart from AuthContext and TxContext we only require that the arguments are
+    // not mutable references. They can be mutable values, as their mutability
+    // cannot affect outside state.
+    for token in function_signature
+        .0
+        .iter()
+        .take(function_signature.len() - 2)
+    {
+        if token.is_mutable_reference() {
```
miker83z: It is a bit more complex. It must be either a primitive type or just an immutable reference.
- `is_mutable_reference()` blocks `MutableReference` but not `Datatype`, `DatatypeInstantiation` and `TypeParameter` of objects -> by value.
- something like checking if the token is of a primitive value first and then, if not, it can only be a `Reference` is needed https://github.com/iotaledger/iota/blob/4660e9505235ecd54e9a5e02fdaade071622cfaf/crates/iota-json/src/lib.rs#L698

crates/iota-framework/packages/iota-framework/sources/account.move
valeriyr: I don't think it looks better, but Move has a feature that allows using a keyword as an identifier. Something like this:
```
`module`: String,
```

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+        return Err(verification_failure(format!(
+            "Authenticator function '{function_identifier}' can only receive 'AuthContext' as immutable reference"
+        )));
+    }
+
+    // TxContext can only be an immutable reference. Passing it as mutable would
+    // allow `authenticate` functions to create objects, which would be
+    // problematic.
+    if let SignatureToken::Reference(referent) = tx_context {
```
miker83z: We currently have a `TxContext::kind()` function similar to this, but allowing also mutable reference. It would be nice to add a new function there just for immutable ref and then copy both for the `AuthContext`.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: I was thinking about this as well, but couldn't really figure out what would make sense, so landed on leaving it public.
Didn't want to potentially break some move invariants. For example if I allow private functions to be called, I need to somehow also enforce that
the `create_auth_info` call originates from the same module as the targeted `authenticate` is in.

On your first point are you referring to dev mode Tx execution or what exactly? Then it could make sense.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Ahh, very good catch thank you! I remember seeing those, but forgot to investigate further.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Uhh very good thank you!

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Created a sub issue to track this: https://github.com/iotaledger/iota/issues/8215

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Fixed in a659ac6109d027024f1995e0e0d2eb08bec0b805

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Fixed in a659ac6109d027024f1995e0e0d2eb08bec0b805 
with the caveat that I didn't restrict it as much at all.
Made a ticket for figuring this out: https://github.com/iotaledger/iota/issues/8216
I suspect we can only really inspect the consequences when we have a running setup.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,133 @@
...
+    let auth_context = &function_signature.0[function_signature.len() - 2];
+    let tx_context = &function_signature.0[function_signature.len() - 1];
+
+    // AuthContext could potentially passed as value, but that opens up the
+    // possibility for the `authenticate` function to receive it as mutable
+    // value, from which it could mutate before passing it to further `authenticate`
+    // functions, so similarly to TxContext, it is simply not allowed.
+    if !matches!(
+        AuthContext::kind(module, auth_context),
+        AuthContextKind::Immutable
```
miker83z, line 102: To provide more detailed errors, this could match all cases and throw 2 different errors for `None` and `Mutable`. Same for TxContext below.

TheMrAI: If this errors, that means that the code failed during execution.
The user had to write some code, deploy it, then call it from some transaction (which is probably another deployed contract).
If it fails at this point, I am not sure how they are helped with more detail.

Ideally what should happen is that during verification the code collects all errors and reports them as one,
so the user may fix all of them at once.
But this is more of a compiler level function not a runtime/pre-runtime bytecode safety check as others also seem to be.

Here at the verifier stage the user is beyond easy help. We don't know which line of code failed, what spans etc.

Given that for the `view` function the council wants compiler level support for user experience, I think we should consider
instead implementing the the feature there perhaps and sidestepping many problems (while also introducing some serious other complications).

We can add more detailed errors, but I think it will make it jus more complicated.

crates/iota-types/src/base_types.rs
```diff
@@ -132,6 +132,9 @@ pub trait ConciseableName<'a> {
 
 #[serde_as]
 #[derive(Eq, PartialEq, Clone, Copy, PartialOrd, Ord, Hash, Serialize, Deserialize, JsonSchema)]
+/// `Storage ID` of any shared Object
+/// 
+/// The ID by which the object can be identified in global storage.
```
miker83z: Storage ID as a specific context from which the object ID alone may be unrelated. I'd keep that out from the description of this type.
```suggestion
/// The ID by which the object can be identified in global storage.
```

crates/iota-types/src/move_package.rs
```diff
@@ -2,6 +2,37 @@
...
+//!
+//! This module contains the [MovePackage] and types necessary for describing
+//! its update behavior and linkage information for module resolution during
+//! execution.
+//!
+//! The code contains terminology that may be confusing for the uninitiated,
+//! like `Module ID`, `Package ID`, `Storage ID` and `Runtime ID`. For avoidance
+//! of doubt these concepts are defined as so:
+//! - `Package ID` is the [ObjectID] representing the address by which the given
+//!   package may be found in storage.
```
miker83z, line 19: Maybe this to make it different from the storage id?
```suggestion
//! - `Package ID` is the [ObjectID] representing the address of an immutable object which
//!   contains as data the package bytes.
```

crates/iota-types/src/move_package.rs
```diff
@@ -2,6 +2,37 @@
...
+//! - `Storage ID` represents the [ObjectID] by which an object may be found in
+//!   storage. Be that a [MovePackage], [CompiledModule] or just any shared
+//!   object.
```
miker83z: I guess also immutable and owned objects.
```suggestion
//!   storage. Be that a [MovePackage], [CompiledModule] or just any object.
```

crates/iota-types/src/move_package.rs
```diff
@@ -2,6 +2,37 @@
...
+//! - `Runtime ID` will always mean the `Package ID`/`Storage ID` of the
+//!   initially published package. For a non upgradeable package this will
+//!   always be the `Storage ID`. For an upgradeable package, it will be the
+//!   `Storage ID` of the package's first deployed version.
```
miker83z:
```suggestion
//!   always be equal to the `Storage ID`. For an upgradeable package, it will be equal to the
//!   `Storage ID` of the package's first deployed version.
```

TheMrAI: Fixed in: 859e18f20cc0c938824bd54b1fc5524b77a37aae

crates/iota-types/src/move_package.rs
TheMrAI: Fixed in: 859e18f20cc0c938824bd54b1fc5524b77a37aae
Moved them into ObjectId for even better docs.

crates/iota-types/src/move_package.rs
TheMrAI: Fixed in: 859e18f20cc0c938824bd54b1fc5524b77a37aae
Well, it is the same as `Package ID`, so instead I highlighted it more. 
Unless I misunderstood you somewhere.

crates/iota-types/src/base_types.rs
TheMrAI: Fixed in: 859e18f20cc0c938824bd54b1fc5524b77a37aae

crates/iota-types/src/base_types.rs
```diff
@@ -130,6 +130,11 @@ pub trait ConciseableName<'a> {
     fn concise_owned(&self) -> Self::ConciseType;
 }
 
+/// The ID by which the object can be identified in global storage
+///
+/// Be that a [MovePackage](crate::move_package::MovePackage),
+/// [CompiledModule](move_binary_format::CompiledModule) or just
+/// any shared/immutable object.
```
miker83z: This is mixing ownership and object data type. [Object `Data`](https://github.com/iotaledger/iota/blob/005587cead6cbfe071886e1acecd36f42b10d743/crates/iota-types/src/object.rs#L373) can be `MoveObject` or `MovePackage`. [Object `Owner`](https://github.com/iotaledger/iota/blob/005587cead6cbfe071886e1acecd36f42b10d743/crates/iota-types/src/object.rs#L450) can be `AddressOwner`, `ObjectOwner`, `Shared`, `Immutable`.

But I don't see why specifying those object properties for the ObjectID.

TheMrAI: It is not intending to mix anything, but help place this actual type into a mental model.
When one says Object that can mean a lot of things. An instance of a Struct is an object, regardless of the Move language. Even just a piece of memory can be described as an object if you know it represents a type.

This sentence is highlighting the conceptual connection to the reader that the Object that they read about in external Move docs, describing that a package is an object is the same as this MovePackage. 
I think a CompiledModule is not an Object now that I think about it.

| or any shared/immutable object
Is not focusing on the ownership, just expanding the scope of the sentence.

Here one specific "package"  example and btw this also means in general "any shared/immutable object".

I also notice the sentence was a little hastily moved.
How about this:
```
/// The ID by which the `Move object` can be identified in global storage
///
/// This could be a [MovePackage](crate::move_package::MovePackage) or
/// any other shared/immutable `Move object`.
```
But I can remove `shared/immutable` part etc. Not sure how it would make more sense.

miker83z: How about this?
```
/// The ID by which the `Move object` can be identified in global storage
///
/// This could be a [MovePackage](crate::move_package::MovePackage) or
/// any other [MoveObject](crate::object::MoveObject).
```

TheMrAI: Fixed in: 5b0592dc47e2d9e289a1e4a956b3609a89f7b841

iota-execution/latest/iota-move-natives/src/account.rs
```diff
@@ -0,0 +1,172 @@
...
+                ),
+            ),
+        );
+    };
+
+    if let Err(execution_error) =
+        account_auth_verifier::verify_authenticate_func(&compiled_module, function_identifier)
+    {
+        return Err(
+            PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
```
Dkwcs, line 85: I guess in future we have to standardize error codes for the main failure cases - for instance(E_MODULE_NOT_FOUND, E_FUNC_NOT_FOUND, E_SIG_MISMATCH, etc)
Just reasoning

iota-execution/latest/iota-move-natives/src/account.rs
```diff
@@ -0,0 +1,172 @@
...
+    let account_create_auth_info_v1_impl_params = context
+        .extensions_mut()
+        .get::<NativesCostTable>()
+        .account_create_auth_info_v1_impl_params;
+    native_charge_gas_early_exit!(
+        context,
+        account_create_auth_info_v1_impl_params.create_auth_info_v1_cost_base
+    );
+
+    let function_name_bytes = pop_arg!(args, VectorRef);
```
Dkwcs, line 47: Just a refactoring suggestion we can extract a single or few functions that encapsulate resolution + validation

```
context
        .extensions_mut()
        .get::<NativesCostTable>()
        .account_create_auth_info_v1_impl_params;
    native_charge_gas_early_exit!(
        context,
        account_create_auth_info_v1_impl_params.create_auth_info_v1_cost_base
    );

    let function_name_bytes = pop_arg!(args, VectorRef);
    
    let function_name = String::from(unsafe {
        std::str::from_utf8_unchecked(function_name_bytes.as_bytes_ref().as_slice())
    });
    let function_identifier = Identifier::new(function_name.clone()).unwrap();

    let module_name_bytes = pop_arg!(args, VectorRef);
    let module_name = String::from(unsafe {
        std::str::from_utf8_unchecked(module_name_bytes.as_bytes_ref().as_slice())
    });
    let module_identifier = Identifier::new(module_name.clone()).unwrap();

    let package = pop_arg!(args, AccountAddress);
    let package_id = ObjectID::from(package);
```

etc

in order to have something like 

```
struct AuthenticateData<‘a> {
    module: CompiledModule,
    func_handle: FunctionHandleIndex,
    is_public_entry: bool,
    sig_ok: bool,
}

fn resolve_auth_data(
    ctx: &mut NativeContext,
    package_id: Option<ObjectID>,    // None => derive from active frame (self)
    module_name: &str,
    function_name: &str,
) -> VMResult<AuthenticateData<'_>> {
    let (pkg_id, module_ident) = resolve_package_and_module(ctx, package_id, module_name)?;
    let module = load_compiled_module(ctx, pkg_id, &module_ident)?; // via RawModuleLoader or active frame
    let (fh, vis) = lookup_function(&module, function_name)?;
    enforce_public_entry(vis)?;
    enforce_authenticate_signature(&module, fh)?;
    Ok(ResolvedAuthenticate { module, func_handle: fh, is_public_entry: true, sig_ok: true })
}

```

TheMrAI:
This PR will be split, do not review.

iota-execution/latest/iota-move-natives/src/account.rs
TheMrAI: Good points! I was hesitant at this stage in refactoring as I am still unsure if it would actually work.

iota-execution/latest/iota-move-natives/src/account.rs
TheMrAI: TBH I forgot about the Error, just put in something to have it. Have to double check if there is something more appropriate.

PR #8184, opened 2025-08-12
feat(transaction-builder): Add `move_view_call_tx_kind` with checks

### LINKED ISSUE #8145: feat(transaction-builder): add move_view_call_tx_kind method
## Description

Create a new method in `TransactionBuilder::move_view_call_tx_kind` which duplicates the logic of the respective `move_call_tx_kind` method with the following additional requirement:

* The method should verifies that the move function signature complies to the move view function specification. More specifically:
   * Verifies that the function is read only, and that
   * It has a return type

This follows [#8079](https://github.com/iotaledger/iota/issues/8079#issuecomment-3163815525)_

miker83z:
# Description of change

- Added a `move_view_call_tx_kind` function to the `TransactionBuilder`; this works similarly to `move_call_tx_kind`.
- It invokes `single_move_view_call`, which creates a PTB with a single move call to a Move View Function;
- The main logic for checking the compliance with a Move View Function is in the newly created function `resolve_and_checks_view_function_json_args`
- This function checks that in the signature of the function passed as input: 
  - at least one return type is present  
  - all parameters are immutable references or pure bytes. 
- Added an example in the `iota-sdk` which creates a move view call tx kind PTB and then using the dev inspect read API (both public and private functions scenarios are tested).

Fixes #8145 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Added move_view_call_tx_kind and single_move_view_call to the TransactionBuilder
- [ ] REST API:

### DIFF
--- a/crates/iota-sdk/examples/read_api/move_view_function_call.rs
+++ b/crates/iota-sdk/examples/read_api/move_view_function_call.rs
@@ -1,0 +1,108 @@
+// Copyright (c) 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! This example shows how to update a PTB with a move view function call.
+//!
+//! cargo run --example move_view_function_call
+
+#[path = "../utils.rs"]
+mod utils;
+
+use iota_json::IotaJsonValue;
+use iota_json_rpc_types::{DevInspectResults, IotaTypeTag};
+use iota_sdk::{
+    IotaClient, types::programmable_transaction_builder::ProgrammableTransactionBuilder,
+};
+use iota_types::{
+    base_types::{IotaAddress, ObjectID},
+    transaction::{ProgrammableTransaction, TransactionKind},
+};
+use serde_json::json;
+use utils::setup_for_read;
+
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let (client, sender) = setup_for_read().await?;
+
+    // Move view function call to a public function: get the current timestamp in
+    // milliseconds.
+    let public_call_results = move_view_function_dev_inspect(
+        sender,
+        &client,
+        "0x2".parse()?,
+        "clock",
+        "timestamp_ms",
+        vec![],
+        vec![IotaJsonValue::new(json!(iota_types::IOTA_CLOCK_OBJECT_ID))?],
+    )
+    .await?
+    .results;
+    println!("{:?}", public_call_results);
+
+    // Move view function call to a private function: get the current timestamp in
+    // milliseconds.
+    let private_call_results = move_view_function_dev_inspect(
+        sender,
+        &client,
+        "0x2".parse()?,
+        "random",
+        "load_inner",
+        vec![],
+        vec![IotaJsonValue::new(json!(
+            iota_types::IOTA_RANDOMNESS_STATE_OBJECT_ID
+        ))?],
+    )
+    .await?
+    .results;
+    println!("{:?}", private_call_results);
+
+    Ok(())
+}
+
+async fn move_view_function_dev_inspect(
+    sender: IotaAddress,
+    client: &IotaClient,
+    package_id: ObjectID,
+    module_name: &str,
+    function_name: &str,
+    type_args: Vec<IotaTypeTag>,
+    args: Vec<IotaJsonValue>,
+) -> Result<DevInspectResults, anyhow::Error> {
+    let pt = construct_move_view_function_call(
+        client,
+        package_id,
+        module_name,
+        function_name,
+        type_args,
+        args,
+    )
+    .await?;
+
+    Ok(client
+        .read_api()
+        .dev_inspect_transaction_block(sender, TransactionKind::programmable(pt), None, None, None)
+        .await?)
+}
+
+async fn construct_move_view_function_call(
+    client: &IotaClient,
+    package_id: ObjectID,
+    module_name: &str,
+    function_name: &str,
+    type_args: Vec<IotaTypeTag>,
+    args: Vec<IotaJsonValue>,
+) -> Result<ProgrammableTransaction, anyhow::Error> {
+    let mut ptb = ProgrammableTransactionBuilder::new();
+    client
+        .transaction_builder()
+        .single_move_view_call(
+            &mut ptb,
+            package_id,
+            module_name,
+            function_name,
+            type_args,
+            args,
+        )
+        .await?;
+    Ok(ptb.finish())
+}
--- a/crates/iota-sdk/examples/read_api/move_view_function_call.rs
+++ b/crates/iota-sdk/examples/read_api/move_view_function_call.rs
@@ -33,7 +33,7 @@ async fn main() -> Result<(), anyhow::Error> {
     )
     .await?
     .results;
-    println!("{:?}", public_call_results);
+    println!("{public_call_results:?}");
 
     // Move view function call to a private function: get the current timestamp in
     // milliseconds.
@@ -50,7 +50,7 @@ async fn main() -> Result<(), anyhow::Error> {
     )
     .await?
     .results;
-    println!("{:?}", private_call_results);
+    println!("{private_call_results:?}");
 
     Ok(())
 }
--- a/crates/iota-sdk/examples/read_api/move_view_function_call.rs
+++ b/crates/iota-sdk/examples/read_api/move_view_function_call.rs
@@ -10,13 +10,8 @@ mod utils;
 
 use iota_json::IotaJsonValue;
 use iota_json_rpc_types::{DevInspectResults, IotaTypeTag};
-use iota_sdk::{
-    IotaClient, types::programmable_transaction_builder::ProgrammableTransactionBuilder,
-};
-use iota_types::{
-    base_types::{IotaAddress, ObjectID},
-    transaction::{ProgrammableTransaction, TransactionKind},
-};
+use iota_sdk::IotaClient;
+use iota_types::base_types::{IotaAddress, ObjectID};
 use serde_json::json;
 use utils::setup_for_read;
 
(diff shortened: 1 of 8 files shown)


crates/iota-transaction-builder/src/utils.rs
```diff
@@ -198,6 +183,97 @@ impl TransactionBuilder {
...
+        json_args: Vec<IotaJsonValue>,
+    ) -> Result<Vec<Argument>, anyhow::Error> {
+        // Fetch the move package for the given package ID.
+        let package = self.fetch_move_package(package_id).await?;
+
+        // Extract the expected function signature and check the return type.
+        // If the function is a view function, it MUST return at least a value.
+        let module = package.deserialize_module(module_ident, &BinaryConfig::standard())?;
+        let function_str = function_ident.as_ident_str();
+        let fdef = module
```
TheMrAI: Here you could use the `CompiledModule::find_function_def_by_name` function.

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -198,6 +183,97 @@ impl TransactionBuilder {
...
+                            false,
+                            &module,
+                            &expected_type,
+                        )
+                        .await?,
+                    ))
+                }
+                // Move View Functions can not accept vector of object by value (this case).
+                ResolvedCallArg::ObjVec(_) => Err(UserInputError::InvalidMoveViewFunction {
+                    error: "Found a function parameter which is a vactor of objects".to_owned(),
```
TheMrAI: Vactor! :D
Sorry it just made me laugh.

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -198,6 +183,97 @@ impl TransactionBuilder {
...
+            function_ident.clone(),
+            type_args,
+            json_args,
+        )?;
+
+        // Finally construct the input arguments for the builder, while checking that
+        // none of these are mutable references.
+        let mut args = Vec::new();
+        let mut objects = BTreeMap::new();
+        for (arg, expected_type) in json_args_and_tokens {
```
TheMrAI, line 213: Given that `fetch_move_package` was extracted from `resolve_and_checks_json_args`, would it make sense to also extract the snippet of code constructing the input arguments as well?
```
        let mut args = Vec::new();
        let mut objects = BTreeMap::new();
        for (arg, expected_type) in json_args_and_tokens {
            ....
        }
```
into a function called `args_to_builder_input` or something similar. As far as I can tell that would be the same code for both of the above functions.
Then here you could just verify before the call, that the inputs are good and that could make it cleaner.

crates/iota-transaction-builder/src/lib.rs
```diff
@@ -466,6 +492,38 @@ impl TransactionBuilder {
...
+        let module = Identifier::from_str(module)?;
+        let function = Identifier::from_str(function)?;
+
+        let type_args = type_args
+            .into_iter()
+            .map(|ty| ty.try_into())
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let call_args = self
+            .resolve_and_checks_view_function_json_args(
```
TheMrAI: Not sure what are our exact restrictions.
If we want to allow ourselves to refactor code, this might be also a good candidate as only this function is different with single move call.
If not, then the other refactor in utils.rs should be reverted as well.

I personally would start leaning on the side of refactoring even if it pains us when pulling, but not my business to decide.

crates/iota-sdk/examples/read_api/move_view_function_call.rs
```diff
@@ -0,0 +1,108 @@
...
+async fn move_view_function_dev_inspect(
+    sender: IotaAddress,
+    client: &IotaClient,
+    package_id: ObjectID,
+    module_name: &str,
+    function_name: &str,
+    type_args: Vec<IotaTypeTag>,
+    args: Vec<IotaJsonValue>,
+) -> Result<DevInspectResults, anyhow::Error> {
+    let pt = construct_move_view_function_call(
```
TheMrAI: Wouldn't it be simpler to just call `move_view_call_tx_kind`?

TheMrAI (approved):
Raised some questions, but see no breaking issues.

TheMrAI:
Does this modify the `Rust SDK`?
Maybe the release note should be updated:  `Rust SDK: `

crates/iota-transaction-builder/src/utils.rs
miker83z: Nice, thanks! [27171c5](https://github.com/iotaledger/iota/pull/8184/commits/27171c5b92bd3ebc0b317ba17946bc8e8b53a7ea)

crates/iota-transaction-builder/src/utils.rs
miker83z: Pls check this version [91f9462](https://github.com/iotaledger/iota/pull/8184/commits/91f9462f469481a40648c40f9da9a4415d4fb77b)

crates/iota-transaction-builder/src/lib.rs
miker83z: [91f9462](https://github.com/iotaledger/iota/pull/8184/commits/91f9462f469481a40648c40f9da9a4415d4fb77b)

crates/iota-sdk/examples/read_api/move_view_function_call.rs
miker83z: Yes right [b5cfebf](https://github.com/iotaledger/iota/pull/8184/commits/b5cfebf2e809a120b49920d4e657d00d5469f5e8)

crates/iota-sdk/examples/read_api/move_view_function_call.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! This example shows how to update a PTB with a move view function call.
```
Thoralf-M: I don't see any "update" on a PTB

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -117,85 +117,133 @@ impl TransactionBuilder {
...
         &self,
         builder: &mut ProgrammableTransactionBuilder,
         package_id: ObjectID,
-        module: &Identifier,
-        function: &Identifier,
+        module_ident: &Identifier,
+        function_ident: &Identifier,
         type_args: &[TypeTag],
         json_args: Vec<IotaJsonValue>,
+        is_view: bool,
```
Thoralf-M: That's a breaking change, needs to be in the release notes at least

piotrm50 (approved):
Approving the changes in `crates/iota-types/src/error.rs` as it's the only thing I co-own.

crates/iota-sdk/examples/read_api/move_view_function_call.rs
miker83z: [c3e5881](https://github.com/iotaledger/iota/pull/8184/commits/c3e58812579b620f3ccb4d1c718fb280d88cd592)

crates/iota-transaction-builder/src/utils.rs
kodemartin: Following this comment, I would suggest to create a method dedicated to move view functions instead of adding the `is_view` argument.

You already did some work with this patch to split the inner logic into smaller function calls, rather than inlining, make it more ergonomic to add a dedicated method.

This will avoid the breaking change and refactoring on the downstream (e.g. in `iota-indexer` and `rpc-tests`)

miker83z: reverted [e6d980d](https://github.com/iotaledger/iota/pull/8184/commits/e6d980dc75b3302f3080a6831c15319e9a0440c0)

crates/iota-sdk/examples/read_api/move_view_function_call.rs
```diff
@@ -0,0 +1,76 @@
...
+        "random",
+        "load_inner",
+        vec![],
+        vec![IotaJsonValue::new(json!(
+            iota_types::IOTA_RANDOMNESS_STATE_OBJECT_ID
+        ))?],
+    )
+    .await?
+    .results;
+    println!("{:?}", private_call_results);
```
Thoralf-M:
```suggestion
    println!("{public_call_results:?}");

    // Move view function call to a private function: get the current timestamp in
    // milliseconds.
    let private_call_results = move_view_function_dev_inspect(
        sender,
        &client,
        "0x2".parse()?,
        "random",
        "load_inner",
        vec![],
        vec![IotaJsonValue::new(json!(
            iota_types::IOTA_RANDOMNESS_STATE_OBJECT_ID
        ))?],
    )
    .await?
    .results;
    println!("{private_call_results:?}");
```

crates/iota-transaction-builder/src/lib.rs
```diff
@@ -466,6 +492,38 @@ impl TransactionBuilder {
         Ok(())
     }
 
+    /// Add a single move call to the provided
+    /// [`ProgrammableTransactionBuilder`]. If `is_view` is true, check that the
```
Thoralf-M: `is_view` doesn't exist anymore

crates/iota-types/src/error.rs
```diff
@@ -203,6 +203,8 @@ pub enum UserInputError {
     BlockedMoveFunction,
     #[error("Empty input coins for Pay related transaction")]
     EmptyInputCoins,
+    #[error("Invalid Move View Function call: {error:?}.")]
```
Thoralf-M: All the other errors don't end with a dot

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -220,4 +268,53 @@ impl TransactionBuilder {
...
+    async fn fetch_move_package(&self, package_id: ObjectID) -> Result<MovePackage, anyhow::Error> {
+        let object = self
+            .0
+            .get_object_with_options(package_id, IotaObjectDataOptions::bcs_lossless())
+            .await?
+            .into_object()?;
+        let Some(IotaRawData::Package(package)) = object.bcs else {
+            bail!(
+                "Bcs field in object [{}] is missing or not a package.",
+                package_id
```
Thoralf-M:
```suggestion
                "Bcs field in object [{package_id}] is missing or not a package."                
```

kodemartin (approved):
lgtm

Thoralf-M (approved):

PR #8244, opened 2025-08-18
fix(starfish): eviction acks from tracker

### LINKED ISSUE #8243: starfish: fix bug with eviction from transactions tracker
Currently we evict from the transaction tracker at the moment when the commited leader is going forward. However, some transactions could be unavailable and the tracker is also used to get information who acknowledged availability of that data. This makes impossible for transactions synchronizer to get the required information

polinikita:
# Description of change

Right now eviction from ack tracker in linearizer is done whenever we progress pending commits. However, we should keep this information until the commit gets solid, i.e., all transactions committed to it are locally available. Otherwise, the transaction synchronizer, which relies on the ack tracker, will not get information about authorities that marked the data as available.

Fixes #8243 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/commit_observer.rs
+++ b/crates/starfish/core/src/commit_observer.rs
@@ -143,6 +143,17 @@ impl CommitObserver {
             sent_sub_dags.push(solid_sub_dag);
         }
         self.report_metrics(&pending_sub_dags, &solid_sub_dags);
+
+        // Evict the ack tracker using the information from the latest solid subdag
+        if !solid_sub_dags.is_empty() {
+            let max_solid_commit_leader_round = solid_sub_dags
+                .last()
+                .expect("There should be at least one solid subdag")
+                .leader
+                .round;
+            self.commit_interpreter
+                .evict_old_acknowledgments(max_solid_commit_leader_round);
+        }
         tracing::trace!("Committed & sent {sent_sub_dags:#?}");
 
         Ok((pending_sub_dags, missing_transaction_acknowledgers))
--- a/crates/starfish/core/src/dag_state.rs
+++ b/crates/starfish/core/src/dag_state.rs
@@ -93,7 +93,7 @@ pub(crate) struct DagState {
     /// transactions. Does not persist across restarts and after recovery.
     /// All transactions below this round minus MAX_TRANSACTIONS_ACK_DEPTH
     /// minus MAX_LINEARIZER_DEPTH are evicted from memory.
-    last_available_commit_leader_round: Option<Round>,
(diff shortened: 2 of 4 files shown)


crates/starfish/core/src/commit_observer.rs
```diff
@@ -143,6 +143,17 @@ impl CommitObserver {
...
+        // Evict the ack tracker using the information from the latest solid subdag
+        if !solid_sub_dags.is_empty() {
+            let max_solid_commit_leader_round = solid_sub_dags
+                .last()
+                .expect("There should be at least one solid subdag")
+                .leader
+                .round;
+            self.commit_interpreter
+                .evict_old_acknowledgments(max_solid_commit_leader_round);
+        }
```
filipdulic, line 156: Nit: You could simplify this block using if let Some(...), which avoids is_empty() + expect and makes the intent a bit clearer IMO.
```suggestion
        if let Some(last_solid_sub_dag) = solid_sub_dags.last() {
            self.commit_interpreter
                .evict_old_acknowledgments(last_solid_sub_dag.leader.round);
        }
```

filipdulic (approved):

filipdulic:
Great catch :+1:

crates/starfish/core/src/dag_state.rs
```diff
@@ -340,21 +340,18 @@ impl DagState {
...
         info!(
-            "Last commit with available transactions has leader at round {}; last commit leader round was {}",
-            last_available_commit_leader_round, max_commit_round
+            "Last solid commit has leader at round {}; last pending commit has leader at round {}",
+            last_solid_commit_leader_round, max_commit_round
         );
-        self.last_available_commit_leader_round = Some(last_available_commit_leader_round);
-        let gap = last_available_commit_leader_round.saturating_sub(*max_commit_round);
+        self.last_solid_commit_leader_round = Some(last_solid_commit_leader_round);
+        let gap = (*max_commit_round).saturating_sub(last_solid_commit_leader_round);
```
VorobyevIlya, line 354: I'm a little confused. Why did we subtract max_commit_round before, and now we subtract from max_commit_round? Was it a bug before?

VorobyevIlya (approved):

crates/starfish/core/src/dag_state.rs
polinikita: Yes! We overlooked that. It was constantly zero

PR #8245, opened 2025-08-18
chore(iota): Convert some `panic!`s to `bail!`

### LINKED ISSUE #5506: Harmonize `bail!` vs `panic!` in the CLI
We can definitely `bail`, but I do see some panics in `validator_commands` already. For example

https://github.com/iotaledger/iota/blob/466ac8285237eca2ab06a9dcf4721c61d5cf9af2/crates/iota/src/validator_commands.rs#L309-L312

So whatever you think is more consistent for the API. cc: @thibault-martinez

_Originally posted by @kodemartin in https://github.com/iotaledger/iota/pull/5502#discussion_r1963058240_

DaughterOfMars:
## Description 

Converts a few panics to bails which may inform the user that they may need to update.

Closes #5506

### DIFF
--- a/crates/iota/src/validator_commands.rs
+++ b/crates/iota/src/validator_commands.rs
@@ -407,7 +407,9 @@ impl IotaValidatorCommand {
                     IotaSystemStateSummary::V2(v2) => {
                         (v2.active_validators, Some(v2.committee_members))
                     }
-                    _ => panic!("unsupported IotaSystemStateSummary"),
+                    _ => bail!(
+                        "Unsupported IotaSystemStateSummary found. You may need to upgrade your iota binary."
+                    ),
                 };
 
                 let mut entries = Vec::new();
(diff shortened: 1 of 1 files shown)


crates/iota/src/validator_commands.rs
```diff
@@ -503,7 +505,7 @@ async fn get_cap_object_ref(
         let owner = resp.owner().unwrap();
         let cap_obj_ref = resp
             .object_ref_if_exists()
-            .unwrap_or_else(|| panic!("OperationCap {cap_object_id} shall exist."));
+            .unwrap_or_else(|| panic!("OperationCap {cap_object_id} does not exist"));
```
marc2332, line 508: Just passing by, don't mind me :P
```suggestion
            .with_context(|| format!("OperationCap {cap_object_id} does not exist"))?;
```

DaughterOfMars: The only reason I don't like this is that it won't print the stack trace, and in the case that this happens something is very wrong

thibault-martinez:
I know zklogin is currently disabled but do you want to still replace the panics there for consistency? Even if it doesn't really compile.

DaughterOfMars:
> I know zklogin is currently disabled but do you want to still replace the panics there for consistency? Even if it doesn't really compile.

Not really.

Thoralf-M (approved):

thibault-martinez (approved):

Issue #8234, opened 2025-08-18
node: port the "data-quarantine" changes


muXxer:

jkrvivian:
Merged in #8247

Issue #7909, opened 2025-07-21
Investigate node-team related code could be simplified


muXxer:
We should check if certain functions in the node-related code can be simplified by removing error returns, in case we panic anyways if those functions fail.

This should only be done for internal functions that are not part of traits.
Traits should always be fallible, so the implementor can decide to return errors.

PR #8247, opened 2025-08-18
upstream(core): Add consensus data quarantining

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/e4db4a07c31bfcbcada149cc8f0ef6a73c261b13
- Description:
  - All consensus output is held in memory until all checkpoints built from that output have been committed. This ensures that the epoch db is not corrupted in the event of a fork.
  - Consensus therefore becomes the primary replay log for all crash recovery.
  - Additionally this means there are several tables that we no longer need to write to, and their contents can live in memory.

Part of #8234

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): - Add metrics `consensus_quarantine_queue_size` and `shared_object_assignments_size`
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -25,7 +25,7 @@ use fastcrypto::{
     hash::MultisetHash,
 };
 use iota_archival::reader::ArchiveReaderBalancer;
-use iota_common::debug_fatal;
+use iota_common::{debug_fatal, fatal};
 use iota_config::{
     NodeConfig,
     genesis::Genesis,
@@ -3137,6 +3137,7 @@ impl AuthorityState {
         accumulator: Arc<StateAccumulator>,
         expensive_safety_check_config: &ExpensiveSafetyCheckConfig,
         epoch_supply_change: i64,
+        epoch_last_checkpoint: CheckpointSequenceNumber,
     ) -> IotaResult<Arc<AuthorityPerEpochStore>> {
         Self::check_protocol_version(
             supported_protocol_versions,
@@ -3152,6 +3153,31 @@ impl AuthorityState {
 
         // Terminate all epoch-specific tasks (those started with within_alive_epoch).
         cur_epoch_store.epoch_terminated().await;
+
+        let highest_locally_built_checkpoint_seq = self
+            .checkpoint_store
+            .get_latest_locally_computed_checkpoint()
+            .map(|c| *c.sequence_number())
+            .unwrap_or(0);
+
+        assert!(
+            epoch_last_checkpoint >= highest_locally_built_checkpoint_seq,
+            "{epoch_last_checkpoint} >= {highest_locally_built_checkpoint_seq}"
+        );
+        if highest_locally_built_checkpoint_seq == epoch_last_checkpoint {
+            // if we built the last checkpoint locally (as opposed to receiving it from a
+            // peer), then all shared_version_assignments except the one for the
+            // ChangeEpoch transaction should have been removed
+            let num_shared_version_assignments = cur_epoch_store.num_shared_version_assignments();
+            // Note that while 1 is the typical value, 0 is possible if the node restarts
+            // after committing the last checkpoint but before reconfiguring.
+            if num_shared_version_assignments > 1 {
+                // If this happens in prod, we have a memory leak, but not a correctness issue.
+                debug_fatal!(
+                    "all shared_version_assignments should have been removed (num_shared_version_assignments: {num_shared_version_assignments})"
+                );
+            }
+        }
 
         // Safe to reconfigure now. No transactions are being executed,
         // and no epoch-specific tasks are running.
@@ -3200,6 +3226,7 @@ impl AuthorityState {
                 new_committee,
                 epoch_start_configuration,
                 expensive_safety_check_config,
+                epoch_last_checkpoint,
             )
             .await?;
         assert_eq!(new_epoch_store.epoch(), new_epoch);
@@ -3232,6 +3259,11 @@ impl AuthorityState {
             self.get_backing_package_store().clone(),
             self.get_object_store().clone(),
             &self.config.expensive_safety_check_config,
+            self.checkpoint_store
+                .get_epoch_last_checkpoint(epoch_store.epoch())
+                .unwrap()
+                .map(|c| *c.sequence_number())
+                .unwrap_or_default(),
         );
         let new_epoch = new_epoch_store.epoch();
         self.transaction_manager.reconfigure(new_epoch);
@@ -4950,6 +4982,7 @@ impl AuthorityState {
         new_committee: Committee,
         epoch_start_configuration: EpochStartConfiguration,
         expensive_safety_check_config: &ExpensiveSafetyCheckConfig,
+        epoch_last_checkpoint: CheckpointSequenceNumber,
     ) -> IotaResult<Arc<AuthorityPerEpochStore>> {
         let new_epoch = new_committee.epoch;
         info!(new_epoch = ?new_epoch, "re-opening AuthorityEpochTables for new epoch");
@@ -4966,6 +4999,7 @@ impl AuthorityState {
             self.get_object_store().clone(),
             expensive_safety_check_config,
             cur_epoch_store.get_chain_identifier(),
+            epoch_last_checkpoint,
         );
         self.epoch_store.store(new_epoch_store.clone());
         Ok(new_epoch_store)
@@ -5060,6 +5094,14 @@ impl RandomnessRoundReceiver {
         let transaction = VerifiedExecutableTransaction::new_system(transaction, epoch);
         let digest = *transaction.digest();
 
+        // Randomness state updates contain the full bls signature for the random round,
+        // which cannot necessarily be reconstructed again later. Therefore we must
+        // immediately persist this transaction. If we crash before its outputs
+        // are committed, this ensures we will be able to re-execute it.
+        self.authority_state
+            .get_cache_commit()
+            .persist_transactions(&transaction);
+
         // Send transaction to TransactionManager for execution.
         self.authority_state
             .transaction_manager()
@@ -5106,7 +5148,7 @@ impl RandomnessRoundReceiver {
             let mut effects = result.unwrap_or_else(|_| panic!("failed to get effects for randomness state update transaction at epoch {epoch}, round {round}"));
             let effects = effects.pop().expect("should return effects");
             if *effects.status() != ExecutionStatus::Success {
-                panic!(
+                fatal!(
                     "failed to execute randomness state update transaction at epoch {epoch}, round {round}: {effects:?}"
                 );
             }
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -3156,7 +3156,7 @@ impl AuthorityState {
 
         let highest_locally_built_checkpoint_seq = self
             .checkpoint_store
-            .get_latest_locally_computed_checkpoint()
+            .get_latest_locally_computed_checkpoint()?
             .map(|c| *c.sequence_number())
             .unwrap_or(0);
 
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -5100,7 +5100,7 @@ impl RandomnessRoundReceiver {
         // are committed, this ensures we will be able to re-execute it.
         self.authority_state
             .get_cache_commit()
-            .persist_transactions(&transaction);
+            .persist_transaction(&transaction);
 
         // Send transaction to TransactionManager for execution.
         self.authority_state
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -3162,7 +3162,7 @@ impl AuthorityState {
 
         assert!(
             epoch_last_checkpoint >= highest_locally_built_checkpoint_seq,
-            "{epoch_last_checkpoint} >= {highest_locally_built_checkpoint_seq}"
+            "expected {epoch_last_checkpoint} >= {highest_locally_built_checkpoint_seq}"
         );
         if highest_locally_built_checkpoint_seq == epoch_last_checkpoint {
             // if we built the last checkpoint locally (as opposed to receiving it from a
@@ -3174,7 +3174,8 @@ impl AuthorityState {
             if num_shared_version_assignments > 1 {
                 // If this happens in prod, we have a memory leak, but not a correctness issue.
                 debug_fatal!(
-                    "all shared_version_assignments should have been removed (num_shared_version_assignments: {num_shared_version_assignments})"
+                    "all shared_version_assignments should have been removed \
+                    (num_shared_version_assignments: {num_shared_version_assignments})"
                 );
             }
         }
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -70,7 +70,7 @@ use tap::TapOptional;
 use tokio::{sync::OnceCell, time::Instant};
 use tracing::{debug, error, info, instrument, trace, warn};
 use typed_store::{
-    DBMapUtils, Map, TypedStoreError,
+    DBMapUtils, Map,
     rocks::{
         DBBatch, DBMap, DBOptions, MetricConf, ReadWriteOptions, default_db_options,
         read_size_from_env,
@@ -2470,8 +2470,7 @@ impl AuthorityPerEpochStore {
             // for the same tx digest
             assert!(
                 user_sigs.insert(digest, sigs).is_none(),
-                "duplicate user signatures for transaction digest: {:?}",
-                digest
+                "duplicate user signatures for transaction digest: {digest:?}"
             );
         }
     }
@@ -3974,11 +3973,11 @@ impl AuthorityPerEpochStore {
         {
             // Reading from the db table is only needed when upgrading to data quarantining
             // for the first time.
+            let tables = self.tables()?;
             let db_iter = tables
                 .pending_checkpoints
                 .safe_iter_with_bounds(last.map(|height| height + 1), None);
-            db_iter
-                .collect::<Result<Vec<(CheckpointHeight, PendingCheckpoint)>, _>>()?
+            db_iter.collect::<Result<Vec<(CheckpointHeight, PendingCheckpoint)>, _>>()?
         } else {
             vec![]
         };
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -50,7 +50,7 @@ use iota_types::{
     },
     messages_consensus::{
         AuthorityCapabilitiesV1, ConsensusTransaction, ConsensusTransactionKey,
-        ConsensusTransactionKind, VersionedDkgConfirmation, check_total_jwk_size,
+        ConsensusTransactionKind, TimestampMs, VersionedDkgConfirmation, check_total_jwk_size,
     },
     signature::GenericSignature,
     storage::{BackingPackageStore, InputKey, ObjectStore},
@@ -107,12 +107,13 @@ use crate::{
     epoch::{
         epoch_metrics::EpochMetrics,
         randomness::{
-            CommitTimestampMs, DkgStatus, RandomnessManager, RandomnessReporter,
+            CommitTimestampMs, DkgStatus, RandomnessManager, RandomnessReporter, SINGLETON_KEY,
             VersionedProcessedMessage, VersionedUsedProcessedMessages,
         },
         reconfiguration::ReconfigState,
     },
-    execution_cache::ObjectCacheRead,
+    execution_cache::{ObjectCacheRead, TransactionCacheRead, cache_types::CacheResult},
+    fallback_fetch::do_fallback_lookup,
     module_cache_metrics::ResolverMetrics,
     post_consensus_tx_reorder::PostConsensusTxReorder,
     signature_verifier::*,
@@ -134,7 +135,9 @@ pub(crate) type EncG = bls12381::G2Element;
 #[path = "consensus_quarantine.rs"]
 pub(crate) mod consensus_quarantine;
 
-use consensus_quarantine::ConsensusCommitOutput;
+use consensus_quarantine::{
+    ConsensusCommitOutput, ConsensusOutputCache, ConsensusOutputQuarantine,
+};
 
 // CertLockGuard and CertTxGuard are functionally identical right now, but we
 // retain a distinction anyway. If we need to support distributed object
@@ -407,6 +410,14 @@ pub struct AuthorityPerEpochStore {
     /// and it needs to be cleared at the end of the epoch.
     tables: ArcSwapOption<AuthorityEpochTables>,
 
+    /// Holds the outputs of both consensus handler and checkpoint builder in
+    /// memory until they are proven not to have forked by a certified
+    /// checkpoint.
+    consensus_quarantine: RwLock<ConsensusOutputQuarantine>,
+    /// Holds variouis data from consensus_quarantine in a more easily
+    /// accessible form.
+    consensus_output_cache: ConsensusOutputCache,
+
     protocol_config: ProtocolConfig,
 
     // needed for re-opening epoch db.
@@ -525,39 +536,12 @@ pub struct AuthorityEpochTables {
     /// Transactions that were executed in the current epoch.
     executed_in_epoch: DBMap<TransactionDigest, ()>,
 
-    /// The tables below manage shared object locks / versions. There are three
-    /// ways they can be updated:
-    /// 1. (validators only): Upon receiving a certified transaction from
-    ///    consensus, the authority assigns the next version to each shared
-    ///    object of the transaction. The next versions of the shared objects
-    ///    are updated as well.
-    /// 2. (validators only): Upon receiving a new consensus commit, the
-    ///    authority assigns the next version of the randomness state object to
-    ///    an expected future transaction to be generated after the next random
-    ///    value is available. The next version of the randomness state object
-    ///    is updated as well.
-    /// 3. (fullnodes + validators): Upon receiving a certified effect from
-    ///    state sync, or transaction orchestrator fast execution path, the node
-    ///    assigns the shared object versions from the transaction effect. Next
-    ///    object versions are not updated.
-    ///
-    /// REQUIRED: all authorities must assign the same shared object versions
-    /// for each transaction.
+    #[allow(dead_code)]
     assigned_shared_object_versions: DBMap<TransactionKey, Vec<(ObjectID, SequenceNumber)>>,
+    /// Next available shared object versions for each shared object.
     next_shared_object_versions: DBMap<ObjectID, SequenceNumber>,
 
-    /// Certificates that have been received from clients or received from
-    /// consensus, but not yet executed. Entries are cleared after
-    /// execution. This table is critical for crash recovery, because
-    /// usually the consensus output progress is updated after a certificate
-    /// is committed into this table.
-    ///
-    /// In theory, this table may be superseded by storing consensus and
-    /// checkpoint execution progress. But it is more complex, because it
-    /// would be necessary to track inflight executions not ordered by
-    /// indices. For now, tracking inflight certificates as a map
-    /// seems easier.
-    #[default_options_override_fn = "pending_execution_table_default_config"]
+    // TODO: delete after DQ is rolled out
     pub(crate) pending_execution: DBMap<TransactionDigest, TrustedExecutableTransaction>,
 
     /// Track which transactions have been processed in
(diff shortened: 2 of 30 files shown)


crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -1364,23 +1298,13 @@ impl ExecutionCacheCommit for WritebackCache {
         WritebackCache::commit_transaction_outputs(self, epoch, digests).boxed()
     }
 
-    fn try_persist_transactions<'a>(
-        &'a self,
-        digests: &'a [TransactionDigest],
-    ) -> BoxFuture<'a, IotaResult> {
-        WritebackCache::persist_transactions(self, digests).boxed()
+    fn try_persist_transactions(&self, tx: &VerifiedExecutableTransaction) -> IotaResult {
```
muXxer:
```suggestion
    fn try_persist_transaction(&self, tx: &VerifiedExecutableTransaction) -> IotaResult {
```

muXxer: We also need to change that in all the other implementations.
Also take care that sometimes we called it "digest" instead of "tx" in the function signature.

crates/iota-core/src/execution_cache.rs
```diff
@@ -176,41 +177,17 @@ pub trait ExecutionCacheCommit: Send + Sync {
...
-    /// re-processing consensus commits and pending_consensus_transactions,
-    /// and this method can be removed.
-    fn try_persist_transactions<'a>(
-        &'a self,
-        digests: &'a [TransactionDigest],
-    ) -> BoxFuture<'a, IotaResult>;
+    /// Durably commit a transaction to the database. Used to store any
+    /// transactions that cannot be reconstructed at start-up by consensus
+    /// replay. Currently the only case of this is RandomnessStateUpdate.
+    fn try_persist_transactions(&self, transaction: &VerifiedExecutableTransaction) -> IotaResult;
```
muXxer:
```suggestion
    fn try_persist_transaction(&self, transaction: &VerifiedExecutableTransaction) -> IotaResult;
```

crates/iota-core/src/execution_cache.rs
```diff
@@ -176,41 +177,17 @@ pub trait ExecutionCacheCommit: Send + Sync {
...
 
     /// Non-fallible version of `try_persist_transactions`.
-    fn persist_transactions<'a>(&'a self, digests: &'a [TransactionDigest]) -> BoxFuture<'a, ()> {
-        Box::pin(async move {
-            self.try_persist_transactions(digests)
-                .await
-                .expect("storage access failed")
-        })
+    fn persist_transactions(&self, digests: &VerifiedExecutableTransaction) {
```
muXxer:
```suggestion
    fn persist_transaction(&self, transaction: &VerifiedExecutableTransaction) {
```

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -748,6 +748,9 @@ impl CheckpointStore {
         Ok(())
     }
 
+    /// TODO: this is only needed while upgrading from non-dataquarantine to
```
muXxer, line 875: we should make a tracking issue out of it, that we remove that later.

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -1239,24 +1250,15 @@ impl CheckpointBuilder {
...
-            .await?;
-
         batch.write()?;
 
+        // Send all checkpoint sigs to consensus.
+        for (summary, content) in &new_checkpoints {
+            self.output
+                .checkpoint_created(summary, content, &self.epoch_store, &self.tables)
+                .await?;
+        }
```
muXxer:
```suggestion
        // Send all checkpoint sigs to consensus.
        for (summary, contents) in &new_checkpoints {
            self.output
                .checkpoint_created(summary, contents, &self.epoch_store, &self.tables)
                .await?;
        }
```

crates/iota-core/src/unit_tests/authority_tests.rs
```diff
@@ -4802,7 +4802,6 @@ async fn test_shared_object_transaction_ok() {
         .epoch_store_for_testing()
         .get_assigned_shared_object_versions(&certificate.key())
         .expect("Reading shared version assignments should not fail")
```
muXxer: You removed the wrong except, the "Reading shared version assignments should not fail" was the correct one.

crates/iota-core/src/unit_tests/authority_tests.rs
```diff
@@ -4913,7 +4912,6 @@ async fn test_consensus_commit_prologue_generation() {
             .epoch_store_for_testing()
             .get_assigned_shared_object_versions(txn_key)
             .unwrap()
```
muXxer: Instead of the "except", the "unwrap" should have been removed.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -37,17 +42,18 @@ pub(crate) struct ConsensusCommitOutput {
     end_of_publish: BTreeSet<AuthorityName>,
     reconfig_state: Option<ReconfigState>,
     consensus_commit_stats: Option<ExecutionIndicesWithStats>,
-    pending_execution: Vec<VerifiedExecutableTransaction>,
 
     // transaction scheduling state
-    shared_object_versions: Option<(AssignedTxAndVersions, HashMap<ObjectID, SequenceNumber>)>,
+    // shared_object_versions: Option<(AssignedTxAndVersions, HashMap<ObjectID, SequenceNumber>)>,
```
muXxer:
```suggestion
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +279,610 @@ impl ConsensusCommitOutput {
...
+        tables: &AuthorityEpochTables,
+        metrics: Arc<EpochMetrics>,
+    ) -> Self {
+        let deferred_transactions = tables
+            .get_all_deferred_transactions()
+            .expect("load deferred transactions cannot fail");
+
+        let deferred_transactions_v2 = tables
+            .get_all_deferred_transactions_v2()
+            .expect("load deferred transactions cannot fail");
```
muXxer:
```suggestion
            .expect("load deferred transactions v2 cannot fail");
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +279,610 @@ impl ConsensusCommitOutput {
...
+            processed_consensus_messages: RefCountedHashMap::new(),
+            metrics: authority_metrics,
+        }
+    }
+}
+
+// Write methods - all methods in this block insert new data into the
+// quarantine. There are only two sources! ConsensusHandler and
+// CheckpointBuilder.
+impl ConsensusOutputQuarantine {
```
muXxer:
```suggestion

    // Write methods - all methods in this block insert new data into the
    // quarantine. There are only two sources! ConsensusHandler and
    // CheckpointBuilder.
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +279,610 @@ impl ConsensusCommitOutput {
...
+    ) {
+        debug!(?sequence_number, "inserting builder summary {:?}", summary);
+        for tx in contents.iter() {
+            self.builder_digest_to_checkpoint
+                .insert(tx.transaction, sequence_number);
+        }
+        self.builder_checkpoint_summary
+            .insert(sequence_number, (summary, contents));
+    }
+
```
muXxer, line 510:
```suggestion

    // Commit methods.

```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1854,14 +1819,70 @@ impl AuthorityPerEpochStore {
...
+        }
+    }
+
     pub fn get_all_deferred_transactions_for_test(
         &self,
-    ) -> IotaResult<Vec<(DeferralKey, Vec<DeferredTransaction>)>> {
-        Ok(self
-            .tables()?
+    ) -> Vec<(DeferralKey, Vec<DeferredTransaction>)> {
+        self.consensus_output_cache
```
muXxer: Shouldn't we apply similar logic here?
```rust
if self
            .protocol_config
            .congestion_control_gas_price_feedback_mechanism()
        {
            self.load_from_deferred_transactions_v2(&mut keys, &mut txns, min, max);
        } else {
            self.load_from_deferred_transactions_v1(&mut keys, &mut txns, min, max);
        }
 ```

**Review by muXxer — CHANGES_REQUESTED** [review]

File: crates/iota-core/src/authority/consensus_quarantine.rs (hunk 8 above)
Comment (jkrvivian) [reply crates/iota-core/src/authority/consensus_quarantine.rs]: I went through the changes multiple times, still missed that 😅

File: crates/iota-core/src/authority/authority_per_epoch_store.rs (hunk 12 above)
Comment (jkrvivian) [reply crates/iota-core/src/authority/authority_per_epoch_store.rs]: I didn't add this logic bc only v2 table is used in our old changes as well. 😁

File: crates/iota-core/src/checkpoints/mod.rs (hunk 4 above)
Comment (jkrvivian) [reply crates/iota-core/src/checkpoints/mod.rs:875]: Added in https://github.com/iotaledger/iota/pull/8247/commits/b74e6ce1713913d73a65a028396340ad53b8ae8e

File: crates/iota-core/src/authority/authority_per_epoch_store.rs (hunk 12 above)
Comment (jkrvivian) [reply crates/iota-core/src/authority/authority_per_epoch_store.rs]: Fixed in https://github.com/iotaledger/iota/pull/8247/commits/54d953e679abd42895864cab9ce06bd4d9b9220e

File: crates/iota-core/src/unit_tests/authority_tests.rs (hunk 6 above)
Comment (jkrvivian) [reply crates/iota-core/src/unit_tests/authority_tests.rs]: Fixe in https://github.com/iotaledger/iota/pull/8247/commits/fa4231b2b12ab40a812207c65a7dc5e760b7c372

File: crates/iota-core/src/unit_tests/authority_tests.rs (hunk 7 above)
Comment (jkrvivian) [reply crates/iota-core/src/unit_tests/authority_tests.rs]: Fixed in https://github.com/iotaledger/iota/pull/8247/commits/fa4231b2b12ab40a812207c65a7dc5e760b7c372

File: crates/iota-core/src/execution_cache/writeback_cache.rs (hunk 1 above)
Comment (jkrvivian) [reply crates/iota-core/src/execution_cache/writeback_cache.rs]: Fixed in https://github.com/iotaledger/iota/pull/8247/commits/1ad6af37e1c1f35faea7fac77cd64c38ba5f0e41

File: crates/iota-core/src/execution_cache.rs (hunk 2 above)
Comment (jkrvivian) [reply crates/iota-core/src/execution_cache.rs]: Fixed in https://github.com/iotaledger/iota/pull/8247/commits/1ad6af37e1c1f35faea7fac77cd64c38ba5f0e41

File: crates/iota-core/src/execution_cache.rs (hunk 3 above)
Comment (jkrvivian) [reply crates/iota-core/src/execution_cache.rs]: Fixed in https://github.com/iotaledger/iota/pull/8247/commits/1ad6af37e1c1f35faea7fac77cd64c38ba5f0e41

File: crates/iota-core/src/authority/authority_per_epoch_store.rs [hunk 13]
```diff
@@ -1854,14 +1819,88 @@ impl AuthorityPerEpochStore {
...
+                transactions.len(),
+                key
+            );
+            keys.push(*key);
+            txns.push((
+                *key,
+                transactions
+                    .clone()
+                    .into_iter()
+                    .map(|tx| DeferredTransaction::new(tx, None))
```
semenov-vladyslav:
```suggestion
                    .iter()
                    .map(|tx| DeferredTransaction::new(tx.clone(), None))
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3323,10 +3406,36 @@ impl AuthorityPerEpochStore {
...
+                let mut deferred_transactions =
+                    self.consensus_output_cache.deferred_transactions.lock();
+
+                for (key, txns) in deferred_txns.into_iter() {
+                    total_deferred_txns += txns.len();
+                    deferred_transactions.insert(
+                        key,
+                        txns.clone()
+                            .into_iter()
+                            .map(|tx| tx.transaction)
```
semenov-vladyslav:
```suggestion
                        txns.iter()
                            .map(|tx| tx.transaction.clone())
```

muXxer (approved):

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,610 @@ impl ConsensusCommitOutput {
...
+        tables: &AuthorityEpochTables,
+    ) -> Vec<(TransactionKey, Vec<(ObjectID, SequenceNumber)>)> {
+        tables
+            .assigned_shared_object_versions
+            .safe_iter()
+            .collect::<Result<Vec<_>, _>>()
+            .expect("db error")
+            .into_iter()
+            .map(|(key, value)| (key, value.into_iter().collect::<Vec<_>>()))
+            .collect()
```
semenov-vladyslav:
```suggestion
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,610 @@ impl ConsensusCommitOutput {
...
+        }
+    }
+
+    pub fn insert(&mut self, key: K, value: V) {
+        let entry = self.map.entry(key);
+        match entry {
+            hash_map::Entry::Occupied(mut entry) => {
+                let (ref_count, v) = entry.get_mut();
+                *ref_count += 1;
+                *v = value;
```
semenov-vladyslav, line 851: What happens when `*v != value`? Should it be an UB (eg. guarded with an `assert_eq!(*v, value)`, and in this case there's no need to reassign `*v` to `value`)? Or should `ref_count` be reset to 1 instead of being incremented? The current implementation is confusing.

semenov-vladyslav (changes requested):

crates/iota-core/src/authority/consensus_quarantine.rs
jkrvivian:
```
// If there are N inserts for the same key, the key will not be removed until
// there are N removes.
//
// It is intended to track the *latest* value for a given key, so duplicate
// inserts are intended to overwrite any prior value.
```
It will overwrite the value regardless.

semenov-vladyslav: The comment was too far away

semenov-vladyslav (approved):

piotrm50 (approved):

crates/typed-store/src/test_db.rs
```diff
@@ -144,7 +144,7 @@ impl<'a, K: Serialize, V> TestDBIter<'a, K, V> {
     pub fn skip_to_last(mut self) -> Self {
         self.with_mut(|fields| {
             // `last` instead of `next_back` because we actually want to consume `iter`
```
TheMrAI: Could you update this comment please otherwise all good.?

TheMrAI (approved):
iota-types all good.

crates/typed-store/src/test_db.rs
jkrvivian: Thanks for pointing this out! I'll revert this change, it's irrelevant to this PR, plus I have no idea how this change sneaked in 🤔

crates/iota-types/src/messages_consensus.rs
```diff
@@ -31,6 +31,9 @@ use crate::{
     transaction::CertifiedTransaction,
 };
 
+/// Non-decreasing timestamp produced by consensus in ms.
+pub type TimestampMs = u64;
+
```
roman1e2f5p8s, line 34: Shouldn't this be named `CommitTimestamp`?
We already have `CheckpointTimestamp`, which is also an alias for `u64`, and which, by the way, should be replaced by `TimestampMs` or `CommitTimestamp` in many places since checkpoint != commit.

crates/iota-single-node-benchmark/src/mock_storage.rs
```diff
@@ -71,8 +71,7 @@ impl InMemoryObjectStore {
                         .get_or_init(|| {
                             epoch_store
                                 .get_assigned_shared_object_versions(tx_key)
-                                .expect("get_assigned_shared_object_versions should not fail")
-                                .map(|l| l.into_iter().collect())
+                                .map(|versions| versions.into_iter().collect::<HashMap<_, _>>())
```
roman1e2f5p8s:
```suggestion
                                .map(|versions| versions.into_iter().collect())
```
The target type can be inferred by the compiler here, so the turbofish is not needed.

crates/iota-node/src/lib.rs
```diff
@@ -584,6 +584,10 @@ impl IotaNode {
             signature_verifier_metrics,
             &config.expensive_safety_check_config,
             ChainIdentifier::from(*genesis.checkpoint().digest()),
+            checkpoint_store
+                .get_highest_executed_checkpoint_seq_number()
+                .expect("checkpoint store read cannot fail")
```
roman1e2f5p8s, line 589:
```suggestion
                .get_highest_executed_checkpoint_seq_number()?
```
Why not error propagation here?

crates/iota-core/src/fallback_fetch.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_types::error::IotaResult;
+
+use crate::execution_cache::cache_types::CacheResult;
+
+/// do_fallback_lookup is a helper function for multi-get operations.
```
roman1e2f5p8s:
```suggestion
/// `do_fallback_lookup` is a helper function for multi-get operations.
```

crates/iota-core/src/fallback_fetch.rs
```diff
@@ -0,0 +1,64 @@
...
+/// It takes a list of keys and first attempts to look up each key in the cache.
+/// The cache can return a hit, a miss, or a negative hit (if the object is
+/// known to not exist). Any keys that result in a miss are then looked up in
+/// the store.
+///
+/// The "get from cache" and "get from store" behavior are implemented by the
+/// caller and provided via the get_cached_key and multiget_fallback functions.
```
roman1e2f5p8s:
```suggestion
/// caller and provided via the `get_cached_key` and `multiget_fallback`
/// functions.
```

crates/iota-core/src/fallback_fetch.rs
```diff
@@ -0,0 +1,64 @@
...
+    assert_eq!(fallback_results.len(), fallback_indices.len());
+    assert_eq!(fallback_results.len(), fallback_keys.len());
+
+    for (i, result) in fallback_indices
+        .into_iter()
+        .zip(fallback_results.into_iter())
+    {
+        results[i] = result;
+    }
+    Ok(results)
```
roman1e2f5p8s, line 67:
```suggestion
    
    Ok(results)
```

crates/iota-core/src/fallback_fetch.rs
```diff
@@ -0,0 +1,64 @@
...
+) -> Vec<V> {
+    do_fallback_lookup_fallible(
+        keys,
+        |key| Ok(get_cached_key(key)),
+        |keys| Ok(multiget_fallback(keys)),
+    )
+    .expect("cannot fail")
+}
+
+pub fn do_fallback_lookup_fallible<K: Clone, V: Default + Clone>(
```
roman1e2f5p8s: Maybe rename this to `try_do_fallback_lookup`?

Also, I think the current doc-string for `do_fallback_lookup` should be removed from there and added here instead. For `do_fallback_lookup`, the doc-string could be something like "Non-fallible version of `try_do_fallback_lookup`".

crates/iota-core/src/fallback_fetch.rs
```diff
@@ -0,0 +1,64 @@
...
+    keys: &[K],
+    get_cached_key: impl Fn(&K) -> CacheResult<V>,
+    multiget_fallback: impl Fn(&[K]) -> Vec<V>,
+) -> Vec<V> {
+    do_fallback_lookup_fallible(
+        keys,
+        |key| Ok(get_cached_key(key)),
+        |keys| Ok(multiget_fallback(keys)),
+    )
+    .expect("cannot fail")
```
roman1e2f5p8s:
```suggestion
    .expect("`try_do_fallback_lookup` should not fail")
```

crates/iota-core/src/epoch/randomness.rs
```diff
@@ -697,12 +697,11 @@ impl RandomnessManager {
...
         let epoch_store = self.epoch_store()?;
-        let tables = epoch_store.tables()?;
 
-        let last_round_timestamp = tables
-            .randomness_last_round_timestamp
-            .get(&SINGLETON_KEY)
-            .expect("typed_store should not fail");
+        let last_round_timestamp = epoch_store
+            .get_randomness_last_round_timestamp()
+            .expect("read should not fail");
```
roman1e2f5p8s, line 703:
```suggestion
        let last_round_timestamp = epoch_store.get_randomness_last_round_timestamp()?;
```
Maybe error propagation is fine here?

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -1317,36 +1334,32 @@ impl CheckpointBuilder {
...
+                .locally_computed_checkpoints
+                .get(&summary.sequence_number)?
+            {
+                if previously_computed_summary != *summary {
+                    // Panic so that we don't send out an equivocating checkpoint sig.
+                    fatal!(
+                        "Checkpoint {} was previously built with a different result: {:?} vs {:?}",
+                        summary.sequence_number,
+                        previously_computed_summary,
+                        summary
```
roman1e2f5p8s:
```suggestion
                        "Checkpoint {} was previously built with a different result: \
                            {previously_computed_summary:?} vs {summary:?}",
                        summary.sequence_number,
```

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -2529,12 +2549,17 @@ impl CheckpointService {
     pub async fn wait_for_rebuilt_checkpoints(&self) {
         let highest_previously_built_seq = self.highest_previously_built_seq;
         let mut rx = self.highest_currently_built_seq_tx.subscribe();
+        let mut highest_currently_built_seq = *rx.borrow_and_update();
+        info!(
+            "Waiting for checkpoints to be rebuilt, previously built seq: {highest_previously_built_seq}, currently built seq: {highest_currently_built_seq}"
```
roman1e2f5p8s:
```suggestion
            "Waiting for checkpoints to be rebuilt, previously built seq: \
            {highest_previously_built_seq}, currently built seq: {highest_currently_built_seq}"
```

crates/iota-core/src/authority.rs
```diff
@@ -3153,6 +3154,31 @@ impl AuthorityState {
...
 
+        let highest_locally_built_checkpoint_seq = self
+            .checkpoint_store
+            .get_latest_locally_computed_checkpoint()?
+            .map(|c| *c.sequence_number())
+            .unwrap_or(0);
+
+        assert!(
+            epoch_last_checkpoint >= highest_locally_built_checkpoint_seq,
+            "{epoch_last_checkpoint} >= {highest_locally_built_checkpoint_seq}"
```
roman1e2f5p8s:
```suggestion
            "expected {epoch_last_checkpoint} >= {highest_locally_built_checkpoint_seq}"
```

crates/iota-core/src/authority.rs
```diff
@@ -3153,6 +3154,31 @@ impl AuthorityState {
...
+            // if we built the last checkpoint locally (as opposed to receiving it from a
+            // peer), then all shared_version_assignments except the one for the
+            // ChangeEpoch transaction should have been removed
+            let num_shared_version_assignments = cur_epoch_store.num_shared_version_assignments();
+            // Note that while 1 is the typical value, 0 is possible if the node restarts
+            // after committing the last checkpoint but before reconfiguring.
+            if num_shared_version_assignments > 1 {
+                // If this happens in prod, we have a memory leak, but not a correctness issue.
+                debug_fatal!(
+                    "all shared_version_assignments should have been removed (num_shared_version_assignments: {num_shared_version_assignments})"
```
roman1e2f5p8s:
```suggestion
                    "all shared_version_assignments should have been removed \
                    (num_shared_version_assignments: {num_shared_version_assignments})"
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -407,6 +410,14 @@ pub struct AuthorityPerEpochStore {
     /// and it needs to be cleared at the end of the epoch.
     tables: ArcSwapOption<AuthorityEpochTables>,
 
+    /// Holds the outputs of both consensus handler and checkpoint builder in
+    /// memory until they are proven not to have forked by a certified
+    /// checkpoint.
+    consensus_quarantine: RwLock<ConsensusOutputQuarantine>,
+    /// Holds variouis data from consensus_quarantine in a more easily
```
```suggestion
    /// Holds various data from consensus_quarantine in a more easily
```
typo

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1839,14 +1804,87 @@ impl AuthorityPerEpochStore {
         Ok(txns)
     }
 
+    fn load_from_deferred_transactions_v2(
```
roman1e2f5p8s: Would rename it to `load_deferred_transactions_v2`, and maybe return `keys` and `txns` directly from the function?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1839,14 +1804,87 @@ impl AuthorityPerEpochStore {
...
+            keys.push(*key);
+            txns.push((*key, transactions.clone()));
+        }
+
+        for key in keys {
+            deferred_transactions.remove(key);
+        }
+    }
+
+    fn load_from_deferred_transactions_v1(
```
roman1e2f5p8s: Would rename it to `load_deferred_transactions_v1`, and maybe return `keys` and `txns` directly from the function?

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,606 @@ impl ConsensusCommitOutput {
...
 
         Ok(())
     }
 }
+
+/// ConsensusOutputCache holds outputs of consensus processing that do not need
+/// to be committed to disk. Data quarantining guarantees that all of this data
+/// will be used (e.g. for building checkpoints) before the consensus commit
+/// from which it originated is marked as processed. Therefore we can rely
+/// on replay of consensus commits to recover this data.
```
```suggestion
/// `ConsensusOutputCache` holds outputs of consensus processing that do not
/// need to be committed to disk. Data quarantining guarantees that all of this
/// data will be used (e.g. for building checkpoints) before the consensus
/// commit from which it originated is marked as processed. Therefore, we can
/// rely on replay of consensus commits to recover this data.
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,606 @@ impl ConsensusCommitOutput {
...
+    ) -> Vec<(TransactionKey, Vec<(ObjectID, SequenceNumber)>)> {
+        tables
+            .assigned_shared_object_versions
+            .safe_iter()
+            .collect::<Result<Vec<_>, _>>()
+            .expect("db error")
+    }
+}
+
+/// ConsensusOutputQuarantine holds outputs of consensus processing in memory
```
```suggestion
/// `ConsensusOutputQuarantine` holds outputs of consensus processing in memory
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,606 @@ impl ConsensusCommitOutput {
...
+                        "Shared object next version not found in quarantine: {:?}",
+                        object_id
+                    );
+                }
+            }
+        }
+    }
+
+    // Read methods - all methods in this block return data from the quarantine
+    // which would otherwise be found in the database.
```
roman1e2f5p8s, line 673:
```suggestion
    // which would otherwise be found in the database.
    
```

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -266,9 +278,606 @@ impl ConsensusCommitOutput {
...
+}
+
+// A wrapper around HashMap that uses refcounts to keep entries alive until
+// they are no longer needed.
+//
+// If there are N inserts for the same key, the key will not be removed until
+// there are N removes.
+//
+// It is intended to track the *latest* value for a given key, so duplicate
+// inserts are intended to overwrite any prior value.
```
```suggestion
/// A wrapper around `HashMap` that uses ref counts to keep entries alive until
/// they are no longer needed.
///
/// If there are `N` inserts for the same key, the key will not be removed until
/// there are `N` removes.
///
/// It is intended to track the *latest* value for a given key, so duplicate
/// inserts are intended to overwrite any prior value.
```

roman1e2f5p8s (approved):
Looks good to me. Left many comments and suggestions, but they are not critical, so feel free to address them before the merge.

crates/iota-types/src/messages_consensus.rs
jkrvivian: I'll keep the naming `TimestampMs` here, and replace `CheckpointTimestamp` with `TimestampMs` in this file now. I think it's better to open a tracking issue to do the timestamp cleanup.

jkrvivian: https://github.com/iotaledger/iota/issues/8391

crates/iota-node/src/lib.rs
jkrvivian: I'll keep this now, and further clean up in https://github.com/iotaledger/iota/issues/7909

PR #8250, opened 2025-08-19
feat(iota_account): implementation of a Move package to represent a basic abstracted IOTA account

### LINKED ISSUE #8238: [AA][move] Implementation of a Move package to represent a basic abstracted IOTA account
Depends on all sub-issues of #8116.  

## General Requirements

- It **MUST NOT** be part of the `iota-framework`.

## 1. `IOTAccount` Struct

The `IOTAccount` is a simple struct that allows creating a plain shared object.

```move
public struct IOTAccount has key {
  id: UID,
}
````

* It must have **no fields** except for the `UID` needed for being an object.
* It must have only the **key ability**; no `store` ability.
* It must have fields only in the form of `dynamic_fields`.
* It must have a field with key `b"IOTACCOUNT_OWNER"` and value being a valid public key, representing the **owner of the account**.

### Account Creation

It must be instantiated using a unique function with a similar implementation:

```move
public entry fun create(pubkey: vector<u8>, scheme: SignatureScheme, ctx: &mut TxContext) {
    let mut iota_account = IOTAccount { id: object::new(ctx) };

    // check the flag in pubkey is the same as the input scheme
    assert!(check_scheme(pubkey, scheme));
    
    // add the owner field
    dynamic_field::add(&mut iota_account.id, b"IOTACCOUNT_OWNER", pubkey);

    let auth_info = account::create_auth_info_v1_self(select_authenticate(scheme));
    dynamic_field::add(&mut object.id, b"IOTA_AUTHENTICATION", auth_info);
  
    // share
    transfer::public_share(iota_account);
}
```

### Authentication Selector
The module must declare an enum:
```move
public enum SignatureScheme {
  ED25519,
  Secp256k1,
  Secp256r1,
}

fun select_authenticate(scheme: SignatureScheme): String {
  match (scheme) {
    SignatureScheme::ED25519 => ascii::string(b"authenticate_ed25519"),
    SignatureScheme::Secp256k1 => ascii::string(b"authenticate_secp256k1"),
    SignatureScheme::Secp256r1 => ascii::string(b"authenticate_secp256r1"),
  } 
}
```

### Add / Remove / Modify Fields

It must have a function to allow the `IOTAccount` to add new fields:

```move
public fun add_field<Name: copy + drop + store, Value: store>(
    self: &mut IOTAccount,
    name: Name,
    value: Value,
    ctx: &TxContext,
){
    let mut id = parent.id;
    // Check that the sender of this TX is the account
    assert!(id.uid_to_address() == ctx.sender);

    // add a new field
    dynamic_field::add(&mut id, name, value);
}
```

* It must also have functions to **remove** and **modify** fields (similar to the one above).

### Authentication

It must implement 3 different authentication mechanisms with 3 functions:

```move
fun authenticate_ed25519(self: &IOTAccount, signature: vector<u8>, _: &AuthContext, ctx: &TxContext);
fun authenticate_secp256k1(self: &IOTAccount, signature: vector<u8>, _: &AuthContext, ctx: &TxContext);
fun authenticate_secp256r1(self: &IOTAccount, signature: vector<u8>, _: &AuthContext, ctx: &TxContext);
```

These must:

* Use the framework crypto natives (`ed25519`, `ecdsa256k1`, `ecdsa256r1`) to verify the signature.
* Read `ctx.digest()` (the digest of the transaction stored in `TxContext`) and use it as the message in `ed25519_verify`, `secp256k1_verify`, and `secp256r1_verify`.
* Use `dynamic_field::borrow(&self.id, b"IOTACCOUNT_OWNER")` to get the public key bytes that are passed to the verify functions.

Dkwcs:
# Description of change

A new Move type has been introduced:

```
public struct IOTAccount has key {
  id: UID,
}
```

What can we do with the introduced account:
- Create an IOTAccount instance.
- Use account dynamic fields.
- Rotate the public key handled by the IOTAccount struct.
- Use basic authenticator functions.

Fixes #8238 .


examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
...
+    let account_id = &mut account.id;
+
+    // Add the account owner public key as a dynamic field.
+    dynamic_field::add(account_id, IOTACCOUNT_OWNER_PUBLIC_KEY_DF_NAME, public_key);
+
+    // Add the authenticator info as a dynamic field.
+    dynamic_field::add(account_id, account::authenticator_df_name(), authenticator);
+
+    // Turn the account object into a mutable shared object.
+    iota::transfer::share_object(account);
```
miker83z: More succint?
```suggestion
    let mut id = object::new(ctx);

    // Add the account owner public key as a dynamic field.
    dynamic_field::add(&mut id, IOTACCOUNT_OWNER_PUBLIC_KEY_DF_NAME, public_key);

    // Add the authenticator info as a dynamic field.
    dynamic_field::add(&mut id, account::authenticator_df_name(), authenticator);

    // Turn the account object into a mutable shared object.
    iota::transfer::share_object(IOTAccount { id });
```

examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
...
+public fun remove_field<Name: copy + drop + store, Value: store>(
+    self: &mut IOTAccount,
+    name: Name,
+    ctx: &TxContext,
+): Value {
+    // Check that the sender of this transaction is the account.
+    ensure_tx_sender_is_account(self, ctx);
+
+    // Remove a new field and return it.
+    dynamic_field::remove(&mut self.id, name)
```
miker83z, line 99:
```suggestion
    let a: Value = dynamic_field::remove(&mut self.id, name);
    let name_type = std::type_name::get<Value>().into_string().into_bytes();
    let aUTHENTICATOR_TYPE_NAME: vector<u8> =
        b"0000000000000000000000000000000000000000000000000000000000000002::account::AuthenticatorInfoV1";
    assert!(name_type != aUTHENTICATOR_TYPE_NAME, 0);
```
Not sure 100% this works, but if yes then it would make it more secure.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
...
+
+use iota::account::{Self, AuthenticatorInfoV1};
+use iota::auth_context::AuthContext;
+use iota::dynamic_field;
+use iota::ecdsa_k1;
+use iota::ecdsa_r1;
+use iota::ed25519;
+
+/// A dynamic field name for the account owner public key.
+const IOTACCOUNT_OWNER_PUBLIC_KEY_DF_NAME: vector<u8> = b"IOTACCOUNT_OWNER_PUBLIC_KEY";
```
miker83z: We should use the pattern used in kiosk for keys. So, not hardcoded constants but types such as:
```move
public struct OwnerPubKey has copy, drop, store { }
```

examples/move/iota_account/sources/iota_account.move
miker83z: Then also `OwnerPubKey`
https://github.com/iotaledger/iota/pull/8250/files#r2334190844
(but in this case we need to truncate the address part because we don't know it)

examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
...
+public fun borrow_field_mut<Name: copy + drop + store, Value: store>(
+    self: &mut IOTAccount,
+    name: Name,
+    ctx: &TxContext,
+): &mut Value {
+    // Check that the sender of this transaction is the account.
+    ensure_tx_sender_is_account(self, ctx);
+
+    // Borrow the related dynamic field.
+    dynamic_field::borrow_mut(&mut self.id, name)
```
miker83z, line 126: same checks as remove ^

examples/move/iota_account/Move.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "iota_account"
+edition = "2024.beta" # edition = "legacy" to use legacy (pre-2024) Move
```
miker83z:
```suggestion
edition = "2024"
```

TheMrAI (approved):
lgtm

examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
...
+// --------------------------------------- Utilities ---------------------------------------
+
+/// An utility function to borrow the account-related public key.
+fun borrow_public_key(self: &IOTAccount): &vector<u8> {
+    dynamic_field::borrow(&self.id, IOTACCOUNT_OWNER_PUBLIC_KEY_DF_NAME)
+}
+
+/// Checks that the sender of this transaction is the account.
+fun ensure_tx_sender_is_account(self: &IOTAccount, ctx: &TxContext) {
+    assert!(self.id.uid_to_address() == ctx.sender());
```
theiari: Probably I'm missing something, but I guess this assert is enough and we don't need any signature verification, right?

examples/move/iota_account/sources/iota_account.move
```diff
@@ -0,0 +1,185 @@
+module iota_account::iota_account;
```
theiari, line 4: Missing copyrights (?)

theiari (commented):
Looks good to me!
I only left two small comments that are more like personal doubts.

examples/move/iota_account/sources/iota_account.move
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/sources/iota_account.move
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/sources/iota_account.move
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/sources/iota_account.move
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/Move.toml
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/sources/iota_account.move
valeriyr: fixed here: 9b2750bc33

examples/move/iota_account/sources/iota_account.move
valeriyr: No, this is a utility function just to check the sender address, nothing more.

miker83z (approved):

PR #8277, opened 2025-08-19
feat(consensus): Misbehavior reports and verification

cyberphysic4l:
# Description of change

This PR deals with adding misbehavior reports for equivocation and provably faulty blocks to consensus blocks so we can get consensus on the associated misbehaviors. This PR does not deal with using the counts of these misbehaviors to take any actions, but they will be used to modify validator rewards in a future PR. 

## Equivocation reports
One type of misbehavior report implemented here is an equivocation report. Equivocation is the act of an authority signing and disseminating more than one block in a single consensus round which is a form of misbehavior. When an authority witnesses equivocation, they can report it by providing `BlockRef`s for two equivocating blocks in a `MisbehaviorReport` in the `misbehavior_reports` field of their next block. This misbehavior report can be thought of as similar to simple ancestors, except labelled to indicate misbehavior rather than providing approval for those blocks. 
## Provably faulty block reports
Another type of misbehavior report implemented here is `InvalidBlock` which reports a provably faulty block. A provably faulty block is one which passes signature verification, proving that it was created by that stated authority, but fails later verification check. Prior to this PR, such blocks were discarded upon receipt, but now they must be retained to be able to handle the reporting. As such, provably faulty blocks are now
stored in the DAG state. Provably faulty block reports consist simply of a `BlockRef` for the faulty block and are also included in the `misbehavior_reports` field of the block.

## 'Proving' misbehavior with misbehavior reports
Misbehavior reports do not really contain proof of the misbehavior itself, they are just a reference to the actual misbehavior and a statement that the issuer of the block containing the report claims to have witnessed the misbehavior. All references in the reports must be checked before ancestor selection, so it follows that anyone approving the block containing reports also claims to have witnessed those misbehaviors. To achieve this, when dealing with a block containing a misbehavior report, the `BlockRef`s in the report must be seen before accepting the block, and missing blocks from these refs must be retrieved before using that block for ancestor selection. This works in the same way as regular ancestors, ensuring that if a block containing an equivocation report is finalised, that means a quorum of validators witnessed the equivocation.

## Future use of misbehavior reports
Once the reports are present in the the DAG, they will be used to calculate a score for validators that will be used to adjust their rewards. The exact amount that the misbehaviors will affect rewards is yet to be determined.

### Release Notes

- [x] Protocol: Adds misbehavior reporting to consensus blocks for provably faulty blocks and equivocations.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


consensus/core/src/authority_service.rs
```diff
@@ -395,8 +436,18 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+                        if let GetBlockResult::VerifiedBlock(b) = b {
+                            let mut ancestors = b.ancestors().to_vec();
+                            ancestors.extend(
+                                b.misbehavior_reports().iter().flat_map(|r| r.references()),
+                            );
+                            Some(ancestors)
+                        } else {
+                            None
+                        }
+                    })
```
polinikita, line 424: We already don't go inside the if `if !self.context.protocol_config.consensus_batched_block_sync()` even on the Mainnet, so I would remove this part as the whole not if section should be deprecated.

consensus/core/src/block.rs
```diff
@@ -620,15 +688,37 @@ impl TestBlock {
 /// A block can attach reports of misbehavior by other authorities.
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct MisbehaviorReport {
-    pub target: AuthorityIndex,
-    pub proof: MisbehaviorProof,
+    target: AuthorityIndex,
+    proof: MisbehaviorProof,
```
polinikita, line 692: Not sure that this can be called a proof, but rather a hint as BlockRef is a digest+author+round.

consensus/core/src/block.rs
```diff
@@ -620,15 +688,37 @@ impl TestBlock {
 /// A block can attach reports of misbehavior by other authorities.
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct MisbehaviorReport {
-    pub target: AuthorityIndex,
-    pub proof: MisbehaviorProof,
+    target: AuthorityIndex,
```
polinikita, line 691: I am wondering whether target needs to be there. Would it be more natural to extract it from the BlockRef?

polinikita:
Could you please write the scope of this PR? When arriving through the block stream. we are not adding blocks to the local DAG if they are provably faulty, but we do that if they arrive through the block synchronizer or commit syncer.  Also, what are the next planned PRs to make this misbehaviour mechanism fully working?

consensus/core/src/block.rs
cyberphysic4l: I would agree that it would be more natural to simply include a BlockRef, but my reason for using the existing struct is to avoid introducing a entirely new block version as `BlockV1` already includes a `Vec<MisbehaviorReport>` field.

consensus/core/src/block.rs
cyberphysic4l: Same comment as above, I agree about naming but would have to introduce a new block version in order to change this I think.

cyberphysic4l (commented):
> Could you please write the scope of this PR? 

I have improved the PR description to explain the scope of the PR. 

> When arriving through the block stream. we are not adding blocks to the local DAG if they are provably faulty, but we do that if they arrive through the block synchronizer or commit syncer.  

What do you mean by the block stream? Faulty blocks arriving via `handle_send_block` are added to the local DAG and used in they are added to the reports for the next block during smart ancestor selection. The faulty blocks also must be present in the DAG state in order to accept a block that reports them during the `handle_send_block`. 

> Also, what are the next planned PRs to make this misbehaviour mechanism fully working? 

This is now explained in the PR description.

consensus/core/src/core.rs
```diff
@@ -571,7 +586,7 @@ impl Core {
                 .protocol_config
                 .consensus_smart_ancestor_selection()
```
polinikita, line 586: This is off for all networks. We decided to not use this feature and later on, it will not be available in Starfish. So, one needs to handle the else branch to include misbehaviour proofs

cyberphysic4l: Thanks for spotting this, misbehavior reports are now dealt with in the regular ancestor selection aswell.

consensus/core/src/synchronizer.rs
```diff
@@ -625,28 +634,51 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
...
+        // take care of that.
         let missing_blocks = core_dispatcher
-            .add_blocks(blocks)
+            .add_blocks(verified_blocks)
             .await
             .map_err(|_| ConsensusError::Shutdown)?;
+        // Add any provably faulty blocks to core for processing
+        for provably_faulty_block in provably_faulty_blocks {
+            core_dispatcher
+                .add_provably_faulty_block(provably_faulty_block)
```
polinikita: Could you please change the method to use as argument a vector of faulty blocks to align with add_blocks method.

cyberphysic4l: Done

consensus/core/src/block_manager.rs
```diff
@@ -301,11 +311,13 @@ impl BlockManager {
...
+    // TODO: remove timestamp part of checks once timestamping is refactored to the
+    // new approach.
+    // Verifies each block's timestamp based on its ancestors, and
+    // then verifies any misbehavior reports in the block and persists in store
     // all the valid blocks that should be accepted. Method returns the accepted
     // and persisted blocks.
-    fn verify_block_timestamps_and_accept(
+    fn verify_referenced_blocks_and_accept(
         &mut self,
         unsuspended_blocks: impl IntoIterator<Item = VerifiedBlock>,
```
polinikita: could you please rename unsuspended_blocks to just accepted_or_unsuspended blocks as we extend accepted blocks with unsuspended at the previous step in the main flow

consensus/core/src/block_manager.rs
```diff
@@ -369,6 +384,32 @@ impl BlockManager {
...
+                            proof_blocks.push(report_proof_blocks);
+                        }
+                    }
+                }
+                if let Err(e) = self
+                    .block_verifier
+                    .verify_misbehavior_reports(&b, &proof_blocks)
+                {
+                    warn!("Block {:?} failed to verify misbehavior reports: {}", b, e);
+                    blocks_to_reject.insert(b.reference(), b);
```
polinikita: This part seems to be problematic if the blocks are received through the commit syncer. We will not fetch these proof blocks through the synchroniser and the node will stuck forever. I suggest to add a flag `committed` from the previously called method and not check misbehaviour proofs in such a case.

consensus/core/src/authority_service.rs
cyberphysic4l: I have to update this part because of the new `GetBlockResult` struct returned, so I think it's no harm to add the misbehavior reports there too. I have made sure to take care of it in the other branch too.

consensus/core/src/storage/mem_store.rs
```diff
@@ -208,6 +260,11 @@ impl Store for MemStore {
         Ok(commits)
     }
 
+    fn scan_misbehavior_reports(&self) -> ConsensusResult<HashSet<MisbehaviorReport>> {
+        let inner = self.inner.read();
+        Ok(inner.misbehavior_reports.clone())
```
oliviasaa, line 265: checkk if this is not too much

consensus/core/src/block.rs
```diff
@@ -618,17 +686,39 @@ impl TestBlock {
...
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, Hash)]
 pub struct MisbehaviorReport {
-    pub target: AuthorityIndex,
-    pub proof: MisbehaviorProof,
+    target: AuthorityIndex,
+    proof: MisbehaviorProof,
+}
+
+impl MisbehaviorReport {
+    pub fn new(target: AuthorityIndex, proof: MisbehaviorProof) -> Self {
```
oliviasaa, line 696: why passing the target?

consensus/core/src/authority_service.rs
```diff
@@ -108,21 +111,34 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+                    .node_metrics
+                    .invalid_blocks
+                    .with_label_values(&[
+                        peer_hostname.as_str(),
+                        "handle_send_block",
+                        e.clone().name(),
+                    ])
+                    .inc();
+                info!("Unprovably faulty block from {}: {}", peer, e);
+                return Err(e);
```
oliviasaa, line 127: why inside?

consensus/core/src/authority_service.rs
```diff
@@ -427,31 +456,43 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        let block_bytes = if commit_sync_handle {
             // For commit sync, we respond with all blocks from the store
             self.dag_state
                 .read()
                 .get_blocks(&block_refs)
                 .into_iter()
-                .flatten()
-                .collect()
+                .filter_map(|b| match b {
+                    GetBlockResult::VerifiedBlock(block) => Some(block.serialized().clone()),
```
oliviasaa, line 466: why serialized?

consensus/core/src/authority_service.rs
```diff
@@ -473,31 +514,30 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
-                    self.context
-                        .parameters
-                        .max_blocks_per_sync
-                        .saturating_sub(blocks.len()),
-                );
-                blocks.extend(missing_blocks);
-                if blocks.len() >= self.context.parameters.max_blocks_per_sync {
-                    blocks.truncate(self.context.parameters.max_blocks_per_sync);
+                let missing_blocks = dag_state
+                    .get_cached_blocks_in_range(
```
oliviasaa, line 518: why not adding the faulty ones too?

consensus/core/src/block_manager.rs
```diff
@@ -369,9 +385,38 @@ impl BlockManager {
...
+                    let mut proof_blocks: Vec<Vec<GetBlockResult>> = vec![];
+                    for report in b.misbehavior_reports() {
+                        // get references and remove none values.
+                        match report.proof() {
+                            MisbehaviorProof::InvalidBlock(block_ref) => {
+                                proof_blocks.push(vec![self.dag_state.read().get_block(block_ref)]);
+                            }
+                            MisbehaviorProof::Equivocation { .. } => {
+                                let report_proof_blocks =
+                                    (self.dag_state.read().get_blocks(&report.references()))
```
oliviasaa, line 403: what if you don't find it?

consensus/core/src/authority_service.rs
cyberphysic4l: Just because we don't store cached faulty blocks by authority and round as we do with verified blocks, so it is not possible at present to retrieve the missing ones in this way.

oliviasaa:
I deleted my old comments because I pushed them by mistake, they were just local notes to myself.
I'm going to start now by adding short comments about small parts of the code

consensus/core/src/storage/mem_store.rs
oliviasaa: During recovery, not everything is scanned, in general. I am not sure if reading all `MisbehaviorReport` from storage is doable. I did something similar, initially, for the unprovable metrics, and got reviews saying that this should not be done.

oliviasaa: Same comment in rocksdb_store.rs

consensus/core/src/authority_service.rs
```diff
@@ -395,8 +411,18 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
             if !commit_sync_handle {
                 let all_ancestors = blocks
                     .iter()
+                    .filter_map(|b| {
+                        if let GetBlockResult::VerifiedBlock(b) = b {
```
oliviasaa, line 415: another minor comment: `GetBlockResult` sounds odd to me, but I still don't have a better suggestion. I'll get back to this when I think of something

consensus/core/src/authority_service.rs
```diff
@@ -427,31 +456,43 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
             block_refs.dedup();
             let dag_state = self.dag_state.read();
-            let mut blocks = dag_state
-                .get_blocks(&block_refs)
+            let get_blocks_results = dag_state.get_blocks(&block_refs);
+            let mut bytes = get_blocks_results
+                .iter()
+                .filter_map(|b| match b {
+                    GetBlockResult::VerifiedBlock(block) => Some(block.serialized().clone()),
+                    GetBlockResult::ProvablyFaultyBlock(block) => Some(block.serialized.clone()),
```
oliviasaa, line 482: This should be checked with consensus, but, as I mentioned to you during a call, I wonder how much consensus would be impacted if we don't look for `ProvablyFaultyBlock` in storage for requests that came from the synchronizer (and only return blocks from cache). This would probably not work with the current logic and would require some changes, I think

consensus/core/src/authority_service.rs
```diff
@@ -473,31 +514,30 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
-                    blocks.truncate(self.context.parameters.max_blocks_per_sync);
+                let missing_blocks = dag_state
+                    .get_cached_blocks_in_range(
+                        authority,
+                        highest_accepted_round + 1,
+                        lowest_missing_round,
+                        self.context
+                            .parameters
+                            .max_blocks_per_sync
+                            .saturating_sub(verified_blocks.len()),
```
oliviasaa: I think here maybe it should be `bytes.len()`, no?

consensus/core/src/block.rs
```diff
@@ -618,17 +686,50 @@ impl TestBlock {
...
+        let target = block_refs[0].author;
+        block_refs.sort();
+        let proof = MisbehaviorProof::Equivocation(block_refs);
+        Self { target, proof }
+    }
+
+    pub fn new_invalid_block_report(block_ref: BlockRef) -> Self {
+        let target = block_ref.author;
+        let proof = MisbehaviorProof::InvalidBlock(block_ref);
+        Self { target, proof }
```
oliviasaa, line 708: You have the target inside the `MisbehaviorProof`, no? Why passing it then?
Also, I might have missed where you do this, but I think the `target` in `MisbehaviorReport` is never verified agains the one in `MisbehaviorProof`

consensus/core/src/block_verifier.rs
```diff
@@ -237,6 +247,68 @@ impl BlockVerifier for SignedBlockVerifier {
...
+        for (index, (misbehavior_report, proof_blocks)) in block
+            .misbehavior_reports()
+            .iter()
+            .zip(proof_blocks)
+            .enumerate()
+        {
+            match misbehavior_report.proof() {
+                MisbehaviorProof::InvalidBlock(_block_ref) => {
+                    // We just need to check that a provably faulty block was provided.
+                    if let Some(GetBlockResult::ProvablyFaultyBlock(_)) = proof_blocks.first() {
```
oliviasaa, line 265: I don't understand this line. You got a block with a `MisbehaviorReport` and `MisbehaviorProof` inside it, which means that it was deserialized as `MisbehaviorProof::InvalidBlock(something)`. But where do you verify that `something` is actually a `BlockRef` of an invalid block?

consensus/core/src/block_verifier.rs
```diff
@@ -237,6 +247,68 @@ impl BlockVerifier for SignedBlockVerifier {
...
+                            || block_ref.author != block_refs[0].author
+                        {
+                            return Err(ConsensusError::InvalidEquivocationReport(index));
+                        }
+                    }
+                    // Check that all proof blocks are VerifiedBlock and from the current epoch.
+                    for proof_block in proof_blocks {
+                        if let GetBlockResult::VerifiedBlock(proof_block) = proof_block {
+                            if proof_block.epoch() != self.context.committee.epoch() {
+                                return Err(ConsensusError::InvalidEquivocationReport(index));
```
oliviasaa, line 301: I think you can remove this line :)

consensus/core/src/block_verifier.rs
```diff
@@ -237,6 +247,68 @@ impl BlockVerifier for SignedBlockVerifier {
         }
         Ok(())
     }
+
+    fn verify_misbehavior_reports(
```
oliviasaa, line 251: I'm missing the `target` verification and the ordering between `first` and `second` in the equivocation part

oliviasaa:
A general question, because sometimes it is a bit difficult to follow the whole flow:
A block containing an invalid `MisbehaviorReport` is also provably faulty, right?

consensus/core/src/authority_service.rs
cyberphysic4l: How about `BlockRetrievalResult`?

consensus/core/src/block_verifier.rs
cyberphysic4l: The `proof_blocks` are retrieved from the local DAG state, so the `GetBlockResult::ProvablyFaultyBlock(_)` being returned by the dag state means that the block corresponding to this `BlockRef` in the dag state is saved as provably faulty. That means the block itself was received locally and failed post-signature verification, i.e. it is provably faulty.

cyberphysic4l: I will elaborate on this in the comment to make it more clear if that now makes sense.

cyberphysic4l:
> A general question, because sometimes it is a bit difficult to follow the whole flow: A block containing an invalid `MisbehaviorReport` is also provably faulty, right?

Actually no, it is treated like a block with rejected ancestors, it will be rejected itself, but it is not provably faulty. The `verify()` function that determines if a block is provably faulty does not involve checking the ancestors (or references in misbehavior reports) because that requires retrieving those blocks.

cyberphysic4l:
Provable misbehavior reporting is deferred until we have a stable version of starfish 
See relevant issue https://github.com/iotaledger/iota-private/issues/180

PR #8279, opened 2025-08-19
feat(starfish): update tests for synchronizer

### LINKED ISSUE #8123: starfish: Review and modify tests in synchronizer as necessary.

VorobyevIlya:
# Description of change

All tests in synchronizer.rs were fixed and/or improved for the Starfish protocol. 
One new test with future timestamps was added. There are also a lot of renamings, so the actual change is not as big as it may seem.
Changes outside of the synchronizer file are also just renamings

Fixes #8123

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/synchronizer.rs
+++ b/crates/starfish/core/src/synchronizer.rs
@@ -268,7 +268,7 @@ pub(crate) struct Synchronizer<C: NetworkClient, V: BlockVerifier, D: CoreThread
     commit_vote_monitor: Arc<CommitVoteMonitor>,
     dag_state: Arc<RwLock<DagState>>,
     transactions_synchronizer: Arc<TransactionsSynchronizerHandle>,
-    fetch_blocks_scheduler_task: JoinSet<()>,
+    fetch_block_headers_scheduler_task: JoinSet<()>,
     fetch_own_last_header_task: JoinSet<()>,
     network_client: Arc<C>,
     block_verifier: Arc<V>,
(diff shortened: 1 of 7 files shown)


crates/starfish/core/src/synchronizer.rs
```diff
@@ -2548,10 +2673,88 @@ mod tests {
...
-        let expected5 = missing_verified_block_headers[0..known_number_block_headers]
+        let expected5 = all_verified_block_headers[0..known_number_block_headers]
             .iter()
             .map(|vb| vb.serialized().clone())
             .collect::<Vec<_>>();
         assert_eq!(bytes5, &expected5);
     }
+
+    #[tokio::test]
+    async fn test_process_fetched_headers_with_future_timestamp() {
```
piotrm50, line 2684: Please add more comments describing what is tested and the expected outcome, as it's a bit unclear. If we wait a bit longer then would the other block with the future timestamp be added as well? Or is it ignored completely? The comments in the test suggest that both blocks are added (` // THEN ensure those ended up in Core` ) but we assert that only one has been added.

piotrm50 (changes requested):

crates/starfish/core/src/synchronizer.rs
VorobyevIlya: Fixed. Comments were wrong before, only normal header should be added, the header with a future timestamp should be ignored.

piotrm50 (approved):

polinikita (approved):

PR #8143, opened 2025-08-08
upstream(core): Deprecate skip_to/skip_prior_to/skip_to_last/reverse methods

jkrvivian:
# Description of change

- Upstream range: [v1.43.1, v1.44.3)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/0269b42498befb8f4cf628dd16b277d2e9d250dc

## Links to any relevant issues
Part of #5727 

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

Issue #6153, opened 2025-03-31
Port upstream node team related changes [mainnet-v1.44.3..mainnet-v1.45.3)


muXxer:

muXxer:
Closed by #9767

PR #8282, opened 2025-08-20
upstream(core): Fix expensive callsites of iterator

jkrvivian:
# Description of change

- Upstream range: [v1.44.3, v1.45.3)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/b40bce37d61ea57859a02da816c3c240660df7ae
- Description:
This fixes expensive call sites caused by https://github.com/MystenLabs/sui/commit/0269b42498befb8f4cf628dd16b277d2e9d250dc ([typed store] iterators: deprecate skip_to/skip_prior_to/skip_to_last/reverse methods), #8143

Part of #6153 

### DIFF
--- a/crates/iota-core/src/jsonrpc_index.rs
+++ b/crates/iota-core/src/jsonrpc_index.rs
@@ -15,6 +15,7 @@ use std::{
     },
 };
 
+use iota_common::try_iterator_ext::TryIteratorExt;
 use iota_json_rpc_types::{IotaObjectDataFilter, TransactionFilter};
 use iota_storage::{mutex_table::MutexTable, sharded_lru::ShardedLruCache};
 use iota_types::{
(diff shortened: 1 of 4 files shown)


muXxer (approved):

muXxer:
We should merge that in the former range, because the performance downgrade is introduced there.

piotrm50 (approved):

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -733,14 +733,17 @@ impl IndexStore {
                 )?
                 // skip one more if exclusive cursor is Some
                 .skip(usize::from(cursor.is_some()))
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .take_while(|((id, _), _)| *id == key)
-                .map(|(_, digest)| digest);
+                .take_while(|result| {
```
semenov-vladyslav: The semantics changed:

- before `.collect::<Result<Vec<_>, _>>()?` would return `Err` early even if a single item was `Err`;
- now `.take_while(|result| {..})` would take all items until the first `Err`.

Is this really intended?

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -869,18 +872,21 @@ impl IndexStore {
                 .reversed_safe_iter_with_bounds(None, Some(key))?
                 // skip one more if exclusive cursor is Some
                 .skip(usize::from(cursor.is_some()))
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .take_while(|((id, m, f, _), _)| {
-                    *id == package
-                        && module.as_ref().map(|x| x == m).unwrap_or(true)
-                        && function.as_ref().map(|x| x == f).unwrap_or(true)
+                .take_while(|result| {
```
semenov-vladyslav: Same pattern here.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -971,14 +977,19 @@ impl IndexStore {
             self.tables
                 .event_order
                 .reversed_safe_iter_with_bounds(None, Some((min(tx_seq, seq), event_seq)))?
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .take_while(|((tx, _), _)| tx == &seq)
+                .take_while(|result| {
```
semenov-vladyslav: And here.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1003,14 +1014,14 @@ impl IndexStore {
         Ok(if descending {
             index
                 .reversed_safe_iter_with_bounds(None, Some((key.clone(), (tx_seq, event_seq))))?
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .take_while(|((m, _), _)| m == key)
+                .take_while(|result| result.as_ref().map(|((m, _), _)| m == key).unwrap_or(false))
```
semenov-vladyslav: Here too.

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -1108,14 +1119,19 @@ impl IndexStore {
             self.tables
                 .event_by_time
                 .reversed_safe_iter_with_bounds(None, Some((end_time, (tx_seq, event_seq))))?
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .take_while(|((m, _), _)| m >= &start_time)
+                .take_while(|result| {
```
semenov-vladyslav: The last one.

semenov-vladyslav (commented):
Weird how they got rid of `.collect::<Result<Vec<_>, _>>()?.into_iter()` and changed the semantics.

crates/iota-core/src/jsonrpc_index.rs
jkrvivian: I didn't find the reason to change the semantics, but the changes remain now. 👀

crates/iota-core/src/jsonrpc_index.rs
```diff
@@ -725,39 +727,22 @@ impl IndexStore {
...
-                .take_while(|((id, _), _)| *id == key)
-                .map(|(_, digest)| digest);
-            if let Some(limit) = limit {
-                iter.take(limit).collect()
-            } else {
-                iter.collect()
-            }
+        let iter = if reverse {
+            Either::Left(index.reversed_safe_iter_with_bounds(
+                None,
```
semenov-vladyslav, line 732: Shall we set both limits?

semenov-vladyslav: There's a [doc](https://github.com/facebook/rocksdb/wiki/Iterator#iterating-upper-bound-and-lower-bound) explaining the importance of both bounds.

semenov-vladyslav (approved):
Setting both bounds may be worth.

PR #8295, opened 2025-08-20
feat(verifier): add verify_authenticate_param_type

miker83z:
A suggestion to deal with `TypeParameter`. In particular, we can just check that the type has no key ability, i.e., accept abilities just for primitive types https://github.com/iotaledger/iota/blob/dceb64b9ccf364b0438015b0cc3a1540bd9eeeb7/external-crates/move/crates/move-binary-format/src/file_format.rs#L850-L851

Moreover, I added new test cases. Is this now an exhaustive set of tests?

### DIFF
--- a/account_auth_example/tests/account_auth_example_tests.move
+++ b/account_auth_example/tests/account_auth_example_tests.move
@@ -1,118 +1,673 @@
-
 #[test_only]
 module account_auth_example::account_auth_example_tests;
 
-use account_auth_example::m;
-use std::ascii;
+use account_auth_example::main_m;
+use account_auth_example::object_m;
+use account_auth_example::option_m;
+use account_auth_example::template_m;
+use account_auth_example::vector_m;
 use iota::account;
+use std::ascii;
 
 #[test]
 fun bind_to_test_module() {
     // This is necessary to force the move compiler in including the
     // account_auth_example::m in the test module.
-    // Without this call, the main module won't be available from the test module
+    // Without this call, the main_m module won't be available from the test module
     // making all the test fail outright, because the functions weren't found.
-    m::bind_to_test_module();
+    main_m::bind_to_test_module();
+    object_m::bind_to_test_module();
+    template_m::bind_to_test_module();
+    vector_m::bind_to_test_module();
+    option_m::bind_to_test_module();
 }
 
 #[test]
 fun test_minimally_viable_auth_function() {
-    let account_info = account::create_auth_info_v1(@0x0, ascii::string(b"m"), ascii::string(b"minimally_viable_auth_function"));
+    let account_info = account::create_auth_info_v1(
+        @0x0,
+        ascii::string(b"main_m"),
+        ascii::string(b"minimally_viable_auth_function"),
+    );
     account::drop_auth_info_v1(account_info);
 }
 
 #[test, expected_failure]
 fun has_to_be_public_auth_function() {
-    let account_info = account::create_auth_info_v1(@0x0, ascii::string(b"m"), ascii::string(b"has_to_be_public_auth_function"));
+    let account_info = account::create_auth_info_v1(
+        @0x0,
+        ascii::string(b"main_m"),
+        ascii::string(b"has_to_be_public_auth_function"),
+    );
     account::drop_auth_info_v1(account_info);
 }
 
 #[test, expected_failure]
 fun crate_auth_at_least_two_args() {
-    let account_info = account::create_auth_info_v1(@0x0, ascii::string(b"m"), ascii::string(b"crate_auth_at_least_two_args"));
+    let account_info = account::create_auth_info_v1(
+        @0x0,
+        ascii::string(b"main_m"),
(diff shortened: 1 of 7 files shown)


iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -203,30 +198,79 @@ fn primitive(st: &SignatureToken) -> bool {
...
 
-    match st {
-        st if primitive(st) => true,
+    match param {
+        U8 | U16 | U32 | U64 | U128 | U256 | Bool | Address => Ok(()),
+        Vector(inner) => verify_pure_type(module, function_type_args, inner),
         Datatype(handle_index) => {
             let resolved_struct = resolve_struct(module, *handle_index);
-            resolved_struct == RESOLVED_ASCII_STR
+            if resolved_struct == RESOLVED_ASCII_STR
```
TheMrAI, line 194: Isn't this now unnecessary?
If we allow in general `TypeParameter` for which the only requirement is that it doesn't have `key` then we don't need to check for regular structs these either.
Similar to `DatatypeInstantiation`

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -203,30 +198,79 @@ fn primitive(st: &SignatureToken) -> bool {
 /// On a similar notion a simple `vector<T>` and an `Option<T>` are both also
 /// acceptable as they are built-in move types with rust side counterpart as
 /// long as `T` is recursively `pure` as well.
-#[warn(dead_code)]
-fn pure_type(module: &CompiledModule, st: &SignatureToken) -> bool {
+fn verify_pure_type(
```
TheMrAI: So these will not be "pure". types.
I propose that we simply do away with the name and just call it `verify_type`.
Don't even try to define, what it can be.

account_auth_example/tests/account_auth_example_tests.move
```diff
@@ -1,118 +1,673 @@
-
 #[test_only]
 module account_auth_example::account_auth_example_tests;
 
-use account_auth_example::m;
-use std::ascii;
+use account_auth_example::main_m;
```
TheMrAI, line 4: What is the purpose of the `_m`?
If it is to just signal that it is a module isn't it a bit redundant?
Or otherwise we conflict on `object`, `vector`, `option` words?

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: If we agree that it is unnecessary. Then I think we can also simplify the tests as well.

Essentially what we say is that on the top level you may have an immutable reference to anything (except signer). Maybe something else?
If you have a value however than it has to be recursively `verify_type` type.
Anything else is not acceptable. Then we don't have to test separately for option, vector etc. because we don't distinguish.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -203,30 +198,79 @@ fn primitive(st: &SignatureToken) -> bool {
...
+
+fn verify_authenticate_param_type(
+    module: &CompiledModule,
+    function_type_args: &[AbilitySet],
+    param: &SignatureToken,
+) -> Result<(), String> {
+    use SignatureToken::*;
+
+    match param {
+        Reference(_) => Ok(()),
```
TheMrAI, line 255: It can't be a `signer`, maybe something else? I can't think of anything else.

TheMrAI (commented):
Yes, I shall consume this.

account_auth_example/tests/account_auth_example_tests.move
miker83z: Yes we conflict on object, vector, option words. There could be a nicer addition rather than `_m`, feel free to suggest!

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
miker83z: Why do you think this is not pure as in pure input?

miker83z: Clarified here [eae0840](https://github.com/iotaledger/iota/pull/8295/commits/eae08404b54286945ac784e26eacaaf6f8387086)

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
miker83z: Clarified here [eae0840](https://github.com/iotaledger/iota/pull/8295/commits/eae08404b54286945ac784e26eacaaf6f8387086)

TheMrAI: Good, good, with the clarification I am happy! Thank you sir!

TheMrAI (commented):
Good hack!

PR #8296, opened 2025-08-21
feat(iota-core,iota-node): Capability notification sending for non-committee validators

piotrm50:
# Description of change

This PR introduces ability to send CapabilityNotification by non-committee validators to be taken into consideration during the next committee selection process. 

Part of #

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Ability to send CapabilityNotification by non-committee active validators to be taken into consideration during the next committee selection process.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


roman1e2f5p8s (approved):
Approving files owned by core-protocol and research.

crates/iota-core/src/authority_aggregator.rs
```diff
@@ -1849,6 +1880,132 @@ where
...
+            errors: Vec<(IotaError, Vec<AuthorityName>, StakeUnit)>,
+        }
+
+        let validity_threshold = self.committee.validity_threshold();
+        let quorum_threshold = self.committee.quorum_threshold();
+        let validator_display_names = self.validator_display_names.clone();
+
+        debug!(
+            "Sending capability notification to committee validators with validity threshold: {}",
+            validity_threshold
```
bingyanglin, line 1906:
```suggestion
            "Sending capability notification to committee validators with validity threshold: {validity_threshold}"
```

crates/iota-e2e-tests/tests/reconfiguration_tests.rs
```diff
@@ -989,6 +1124,152 @@ async fn safe_mode_reconfig_test() {
...
+
+    // Should fail with non-retryable error
+    assert!(matches!(
+        result,
+        Err(AggregatorSendCapabilityNotificationError::NonRetryableNotification { .. })
+    ));
+}
+
+#[sim_test]
+async fn test_authority_capabilities_incorrect_epoch_rejection() {
```
semenov-vladyslav, line 1199: Nice, thanks.

crates/iota-node/src/lib.rs
```diff
@@ -2132,6 +2151,96 @@ impl IotaNode {
...
+            self.config.authority_key_pair(),
+        );
+
+        let request = HandleCapabilityNotificationRequestV1 {
+            message: SignedAuthorityCapabilitiesV1::new_from_data_and_sig(capabilities, signature),
+        };
+
+        let mut retry_interval = Duration::from_secs(INITIAL_RETRY_INTERVAL_SECS);
+
+        loop {
```
semenov-vladyslav, line 2246: Should there be a bound on the number of retries? Otherwise this loop can be infinite.

semenov-vladyslav (approved):

TheMrAI (approved):
Parts owned by vm-lang look okay.

crates/iota-node/src/lib.rs
piotrm50: It retries every 5 minutes maximum, and only if the errors are retryable. On the receiving end there is also logic to check if this was already processed (either by receiving through gRPC or from consensus), so as long as one node manages to put it into consensus, other honest nodes should return successful response early. If error is not retryable  (bad signature etc.) then we exit the loop immediately.

filipdulic (approved):

kodemartin (approved):
lgtm

PR #8314, opened 2025-08-22
feat(json-rpc): add iota_view method

### LINKED ISSUE #8249: [indexer,json-rpc]: Support move view function calls
## Description

Following #8145 and #8184 we want to support the move view function calls on the JSON-RPC API by adding a new method `iota_view`.

kodemartin:
# Description of change

Adds the `iota_view` JSON-RPC method following IIP5.

Given the smart contract under `crates/iota-indexer/tests/data/wat_counter`:

```bash
curl -X POST http://localhost:9000 \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "method": "iota_view",
    "params": {
      "function_name": "0xf4faac08dc010872a17a94552f8160bd8e2b1ca72ae9285dd1b641ec9f08fe6d::wat_counter::get_counter",
      "typeArgs": [],
      "callArgs": [
        "0xdf9dd2c49ef82d423c74c20f23c5baac989843a5f15ab1e8cee647ca2109df8f"
      ]
    },
    "id": 1
  }'
```
gives
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "functionReturnValues": [
      "10"
    ]
  }
}
```

and 

```bash
$ curl -X POST http://localhost:9000 \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "method": "iota_view",
    "params": {
      "function_name": "0xf4faac08dc010872a17a94552f8160bd8e2b1ca72ae9285dd1b641ec9f08fe6d::wat_counter::get_wat_object",
      "typeArgs": [],
      "callArgs": [
        "0xdf9dd2c49ef82d423c74c20f23c5baac989843a5f15ab1e8cee647ca2109df8f"
      ]
    },
    "id": 1
  }'
```

gives

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "functionReturnValues": [
      {
        "type": "0xf4faac08dc010872a17a94552f8160bd8e2b1ca72ae9285dd1b641ec9f08fe6d::wat_counter::Wat",
        "fields": {
          "counter": "10",
          "id": {
            "id": "0xdf9dd2c49ef82d423c74c20f23c5baac989843a5f15ab1e8cee647ca2109df8f"
          }
        }
      }
    ]
  }
}

```

Fixes #8249 

### Release Notes

- [x] JSON-RPC: Adds support for move view function calls through a new `iota_view` RPC method.


crates/iota-open-rpc/spec/openrpc.json
```diff
@@ -3201,6 +3201,52 @@
...
+          "name": "function_name",
+          "description": "The fully qualified function name `<package_id>::<module_name>::<function_name>`. E.g.  `0x3::iota_system::get_total_iota_supply`.",
+          "required": true,
+          "schema": {
+            "type": "string"
+          }
+        },
+        {
+          "name": "type_args",
+          "required": true,
```
Thoralf-M: I know it's listed as required in https://github.com/iotaledger/IIPs/blob/main/iips/IIP-0005/iip-0005.md, but wouldn't it make more sense to have this optional and default to an empty array, considering that many functions don't need this?

miker83z: I think the same. However, I followed the specification of `unsafe_moveCall` for writing that part in the iip. In there, `type_arguments` is required. So, I guess, it is a matter of deciding to keep the two methods equivalent or not.

crates/iota-json-rpc-types/src/iota_move.rs
```diff
@@ -129,6 +131,51 @@ pub struct IotaMoveModuleId {
...
+    }
+}
+
+impl TryFrom<ParsedFqName> for MoveFunctionName {
+    type Error = UserInputError;
+
+    fn try_from(parsed: ParsedFqName) -> Result<Self, Self::Error> {
+        let package = match parsed.module.address {
+            ParsedAddress::Numerical(addr) => ObjectID::from(addr.into_inner()),
+            _ => {
```
miker83z: Did you check `iota_types::parse_iota_fq_name`? In there it accepts `ParsedAddress::Named` for known framework packages. I guess `into_fq_name(&resolve_address)` could be used equivalently here, and `parse_iota_fq_name` could be used above in the `FromStr for MoveFunctionName`.

crates/iota-json-rpc-types/src/iota_move.rs
```diff
@@ -129,6 +131,51 @@ pub struct IotaMoveModuleId {
     name: String,
 }
 
+/// Identifies a Move function.
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MoveFunctionName {
```
miker83z, line 137: I prefer `MoveFunctionName` as well, it is clearer. However, given the standard used in this file, something like `IotaMoveFqName` would be more in line.

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -300,3 +302,82 @@ fn execute_transaction_block() {
...
+                Some(NANOS_PER_IOTA),
+                address,
+            )
+            .await;
+        indexer_wait_for_object(client, gas_id, gas_version).await;
+        let ((package_id, version, _), transaction_response) =
+            publish_test_move_package(client, address, &keypair, "wat_counter")
+                .await
+                .unwrap();
+        indexer_wait_for_object(client, package_id, version).await;
```
tomxey, line 335: Since we publish through the indexer and the optimistic branch is merged, this line should not be needed anymore :partying_face:

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -300,3 +302,82 @@ fn execute_transaction_block() {
...
+            })
+            .unwrap();
+
+        // Test u64 return value, which is cast to string.
+        let fn_name = format!("{package_id}::wat_counter::get_counter");
+        let view_results = client
+            .view_function_call(
+                fn_name,
+                vec![],
+                vec![review_id.to_string().as_str().parse().unwrap()],
```
tomxey: Seems a bit not elegant to use string as intermediate form during the conversion.
Would it be possible to convert it directly?

Would the `call_args!` macro help here?

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -300,3 +302,82 @@ fn execute_transaction_block() {
...
+            )
+            .await
+            .unwrap();
+        assert!(view_results.error().is_none(), "{view_results:?}");
+        let return_values = view_results.into_return_values();
+        assert_eq!(return_values.len(), 1);
```
tomxey, line 370: Did you consider asserting the whole value, like: `assert_eq!(return_values, vec![...])` ?
Or for some reason it was not practical/possible?

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1264,6 +1265,80 @@ pub struct IotaExecutionResult {
...
+    /// Execution error from executing the move view call
+    #[serde(rename = "executionError")]
+    Error(String),
+    /// The return values of the move view function
+    #[serde(rename = "functionReturnValues")]
+    Results(Vec<IotaMoveValue>),
+}
+
+impl IotaMoveViewCallResults {
+    /// Process the dev-inspect results to produce the response
```
tomxey: nit: should be third person, so "Processes", according to official rust guidelines.

crates/iota-json-rpc-api/src/write.rs
```diff
@@ -43,6 +44,17 @@ pub trait WriteApi {
         request_type: Option<ExecuteTransactionRequestType>,
     ) -> RpcResult<IotaTransactionBlockResponse>;
 
+    /// Calls a move view function.
+    #[rustfmt::skip]
+    #[method(name = "view")]
+     async fn view_function_call(
```
miker83z, line 50: I think it should go under the Read API, also in the IIP we talk about "read api". Any practical reason for this to be under write api?

crates/iota-json-rpc/src/transaction_execution_api.rs
```diff
@@ -42,12 +47,15 @@ use crate::{
...
     logger::FutureWithTracing,
+    transaction_builder_api::AuthorityStateDataReader,
 };
 
+#[derive(Clone)]
 pub struct TransactionExecutionApi {
     state: Arc<dyn StateRead>,
     transaction_orchestrator: Arc<TransactionOrchestrator<NetworkAuthorityClient>>,
     metrics: Arc<JsonRpcMetrics>,
+    transaction_builder: TransactionBuilder,
```
miker83z, line 58: We are now merging `TransactionExecution` and `TransactionBuilder`. Maybe, instead, it could be worth to have a dedicated `view_api.rs` with that merge and keep the original semantic of transaction execution here? Just a suggestion, I am not sure about the necessity of possible additional cloned code.

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1264,6 +1265,80 @@ pub struct IotaExecutionResult {
...
+            return Ok(Self::Error("function call returned no values".into()));
+        };
+        let Some(execution_results) = tx_execution_results.pop() else {
+            return Ok(Self::Error(
+                "no results from move view function call".into(),
+            ));
+        };
+        if !tx_execution_results.is_empty() {
+            return Ok(Self::Error("multiple transactions executed".into()));
+        }
```
tomxey, line 1314: Could we make it more clear with [exactly_one](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one) ?

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1264,6 +1265,80 @@ pub struct IotaExecutionResult {
...
+        let mut move_call_results = Vec::with_capacity(execution_results.return_values.len());
+        let package_resolver = Resolver::new(package_store);
+        let mut execution_results =
+            execution_results.into_stream_return_value_layouts(&package_resolver);
+        while let Some(result) = execution_results.next().await {
+            let (bytes, move_type_layout) = result?;
+            let move_value = BoundedVisitor::deserialize_value(&bytes, &move_type_layout)?;
+            move_call_results.push(IotaMoveValue::from(move_value));
+        }
```
tomxey, line 1323: Have you considered calling map on execution_results, instead of looping?
As for error handling it should be possible to collect iterator with results to single result.

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1264,6 +1265,80 @@ pub struct IotaExecutionResult {
...
+        Ok(Self::Results(move_call_results))
+    }
+
+    pub fn into_return_values(self) -> Vec<IotaMoveValue> {
+        match self {
+            IotaMoveViewCallResults::Error(_) => Default::default(),
```
tomxey, line 1329: Isn't this hiding the error in case it happens? Could we return Result<Vec<_>> or Option<Vec<_>> from this fn?

crates/iota-open-rpc/spec/openrpc.json
kodemartin: It makes sense to me as well. Why should the signatures of the two methods be coupled?

crates/iota-json-rpc-types/src/iota_move.rs
kodemartin: Good one! Implementing.

crates/iota-json-rpc-types/src/iota_move.rs
kodemartin: The prefix `Iota` is used for types to differentiate from equivalent non-prefixed types in other crates, like the external crates or iota types. 

This is not the case here however.  `MoveFunctionName` was relocated and refactored with this patch in a more proper scope, and there are other types that are not prefixed by `Iota`. This should be constraining at any case I reckon.

crates/iota-json-rpc-api/src/write.rs
kodemartin: Since this is a wrapper of `dev_inspect` it seems intuitive that they both belong in the same scope, represented through modules in the json-rpc implementation.

We might want to restructure the API modules but this extends beyond the scope of this PR.

crates/iota-json-rpc/src/transaction_execution_api.rs
kodemartin: Here I honestly see no conflict using the `TransactionBuilder` internally. Grouping together the methods is dictated by the scope, and as I argued above `view` should be together with `dev_inspect` that wraps it.

How the latter is wrapped is an implementation detail that is flexible.

crates/iota-indexer/tests/rpc-tests/write_api.rs
kodemartin: Yes it would, thanks for the suggestion!

crates/iota-indexer/tests/rpc-tests/write_api.rs
kodemartin: I would have to construct the entire value, plus I get two different assertions that is more granular. One for length, one for the value itself.

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: Yes, it would be more verbose and imply more nesting. I chose this as a clearer implementation.

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: An empty vector implies that there are no return values. Empty return values imply that it is an error, as a move view function must return values if successfully executed.

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: I don't see the benefit. Here we handle two cases: Empty results, and more results than expected.

The checks are thus made explicit without the need for another library.

crates/iota-json-rpc-types/src/iota_move.rs
kodemartin: done with c1f96033a4

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: done with b611a1e396

crates/iota-indexer/tests/rpc-tests/write_api.rs
kodemartin: done with 0e7e8e6501

used `call_arg!` instead to avoid the implicit `?`

crates/iota-indexer/tests/rpc-tests/write_api.rs
kodemartin: Can be removed after rebasing. But I think it's ok for the moment.

crates/iota-open-rpc/spec/openrpc.json
kodemartin: Done with 8a514b6aa2

Thoralf-M (approved):

miker83z (approved):

tomxey (approved):

PR #8350, opened 2025-08-25
feat(indexer): Speed up get_transaction_block

### LINKED ISSUE #8352: [indexer]: Speed up get_transaction_block
get_transaction_block currently uses the same implementation as multi_get_transaction_blocks, which reads data sequentially first from `transactions` table and then from `optimistic_transactions` table.
This can be sped up by reading from 2 tables in parallel

tomxey:
# Description of change

get_transaction_block currently uses the same implementation as multi_get_transaction_blocks, which reads data sequentially first from transactions table and then from optimistic_transactions table.

This PR speeds it up by reading from 2 tables in parallel

fixes https://github.com/iotaledger/iota/issues/8352

Compared performance of get_transaction_block of indexer and node on staging env.

The indexer is currently 1.8× slower than the node for simple read queries, a notable improvement from the earlier measurement of 5× slower, but it happens only for fresh txs that can be returned from optimistic tables.
For old txs where we need to wait for checkpoint tables read the speedup is not that noticeable. Indexer is in such case 4.5x times slower than the node which is similar to the previous situation.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/indexer_reader.rs
+++ b/crates/iota-indexer/src/indexer_reader.rs
@@ -18,6 +18,7 @@ use diesel::{
     sql_types::{BigInt, Bool},
 };
 use fastcrypto::encoding::{Encoding, Hex};
+use futures::stream::{FuturesUnordered, StreamExt};
 use iota_json_rpc_types::{
     AddressMetrics, Balance, CheckpointId, Coin as IotaCoin, DisplayFieldsResponse, EpochInfo,
     EventFilter, IotaCoinMetadata, IotaEvent, IotaMoveValue, IotaObjectDataFilter,
@@ -673,15 +674,11 @@ impl IndexerReader {
         .await
     }
 
-    fn multi_get_transactions(
+    fn get_checkpointed_transactions(
         &self,
-        digests: &[TransactionDigest],
+        digests: &[Vec<u8>],
     ) -> Result<Vec<StoredTransaction>, IndexerError> {
-        let digests = digests
-            .iter()
-            .map(|digest| digest.inner().to_vec())
-            .collect::<HashSet<_>>();
-        let checkpointed_txs = run_query!(&self.pool, |conn| {
+        run_query!(&self.pool, |conn| {
             transactions::table
                 .inner_join(
                     tx_digests::table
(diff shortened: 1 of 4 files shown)


crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -712,10 +708,65 @@ impl IndexerReader {
...
+            let this = self.clone();
+            tokio::task::spawn_blocking(move || {
+                this.get_optimistic_transactions(&[digest_bytes])
+                    .map(|mut txs| txs.pop().map(Into::into))
+            })
+        };
+
+        let mut tx_futures = FuturesUnordered::new();
+        tx_futures.push(checkpointed_tx_future);
+        tx_futures.push(optimistic_tx_future);
```
sergiupopescu199: nit: We could use `FuturesUnordered::from_iter` for a more concise initialization:

```suggestion
        let mut tx_futures =
            FuturesUnordered::from_iter([checkpointed_tx_future, optimistic_tx_future]);
```

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -712,10 +708,65 @@ impl IndexerReader {
...
+
+        if checkpointed_txs.len() == digests.len() {
+            return Ok(checkpointed_txs);
+        }
+
+        let mut missing_digests: HashSet<Vec<u8>> = digests.iter().cloned().collect();
+        for tx in &checkpointed_txs {
+            missing_digests.remove(&tx.transaction_digest);
+        }
+        let missing_digests = missing_digests.into_iter().collect::<Vec<_>>();
```
sergiupopescu199: Currently, we're cloning all digests upfront into a HashSet, then removing matching digests. The suggested approach aims to only clone the missing digests instead of cloning all digests upfront.

```suggestion
        let checkpointed_tx_digests_set = checkpointed_txs
            .iter()
            .map(|tx| &tx.transaction_digest)
            .collect::<HashSet<&Vec<u8>>>();

        let missing_digests = digests
            .iter()
            .filter(|digest| !checkpointed_tx_digests_set.contains(digest))
            .cloned()
            .collect::<Vec<Vec<u8>>>();
```

sergiupopescu199 (approved):
lgtm 🌸

left som optional nits, feel free to close them if you find them redundant.

crates/iota-indexer/src/indexer_reader.rs
```diff
@@ -712,10 +707,102 @@ impl IndexerReader {
...
+        &self,
+        digest: TransactionDigest,
+    ) -> IndexerResult<Option<StoredTransaction>> {
+        let digest_bytes = digest.inner().to_vec();
+        let mut optimistic_tx_future = {
+            let digest_bytes = digest_bytes.clone();
+            Box::pin(self.spawn_blocking(move |this| {
+                this.get_optimistic_transactions_with_cp_info(&[digest_bytes])
+                    .map(|mut txs| txs.pop())
+            }))
```
kodemartin: Nit: consider pinning in the stack with [tokio::pin!](https://docs.rs/tokio/latest/tokio/macro.pin.html)

kodemartin (approved):
lgtm

sergiupopescu199 (approved):

PR #8367, opened 2025-08-26
refactor(move/move-model-2): Switch maps to IndexMap

### LINKED ISSUE #8266: [move-model-2] Switch maps to IndexMap · MystenLabs/sui@2c5490ca
Sui’s commit - https://github.com/MystenLabs/sui/commit/2c5490ca6ed26de2ad37cbf82345375d2e4a8036

Dkwcs:
# Description of change

- Module members are now ordered by their declared+compiled index

According to [changes](https://github.com/MystenLabs/sui/commit/2c5490ca6ed26de2ad37cbf82345375d2e4a8036#diff-75ca1f519c92cce1c3701cb370477da60ef31bccc0ce0810a82bc65c55ed2494)

Fixes #8266 .

### DIFF
--- a/external-crates/move/Cargo.lock
+++ b/external-crates/move/Cargo.lock
@@ -1765,7 +1765,7 @@ name = "move-bytecode-utils"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "indexmap 2.10.0",
+ "indexmap 2.11.0",
  "move-binary-format",
  "move-core-types",
  "petgraph",
@@ -1948,7 +1948,7 @@ dependencies = [
  "clap 4.5.41",
  "codespan",
  "colored",
- "indexmap 2.10.0",
+ "indexmap 2.11.0",
  "move-abstract-interpreter",
  "move-binary-format",
  "move-bytecode-source-map",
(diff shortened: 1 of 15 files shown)


external-crates/move/Cargo.lock
```diff
@@ -2125,6 +2125,7 @@ dependencies = [
  "anyhow",
  "bcs",
  "codespan-reporting",
+ "indexmap 2.10.0",
```
valeriyr: It would be nice to use the same `indexmap` version in both lock files. Now we have `2.10.0` and `2.11.0`. And maybe we will just update it to `2.11.0` in the toml files?

TheMrAI: Something doesn't look right here for me. 
Cargo.toml sets it to 2.8, but then the lock file updates to 2.10?

TheMrAI: If I purge the lock file then it updates it to 2.11.

Dkwcs: I’ve updated it to 2.11

TheMrAI (changes requested):
Please also update `iota/Cargo.toml` indexmap version.

external-crates/move/Cargo.lock
valeriyr: Then you need to update it here as well:
https://github.com/iotaledger/iota/blob/c3569e19f3ff919a4bb42a0252de56825ac25294/Cargo.toml#L285
Thanks!

Dkwcs: Done

TheMrAI (approved):
Yeah, now the indexmap versions look good

valeriyr (approved):

PR #7511, opened 2025-06-24
feat(iota-core): add internal gRPC API for checkpoint fetching (experimental)

bingyanglin:
# Description of change

Currently the indexer syncs the checkpoints via a REST-API via polling, or via filesystem.

This PR adds an internal experimental gRPC API, that supports subscriptions to get new checkpoints. This should reduce the latency between checkpoint creation and indexing.

The profiling results by using a local-network (4 validator, 4 nodes, 2 indexers, 2 Postgres DBs):
image

More details and steps to reproduce the profiling results were documented and discussed at #6995

## Links to any relevant issues

#6995 

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): add experimental "internal" gRPC API
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:
- [x] internal gRPC API: add gRPC streaming for checkpoint data and checkpoint summary

PR #8371, opened 2025-08-27
feat(gRPC): remain original checkpoint processing logic and enhance release_notes

bingyanglin:
# Description of change

- Remain the execution order in `process_executed_checkpoint` as the version before adding gRPC support
- Enhance `release_notes` rule to support `Internal gRPC API`
- Fix lint error due to rustc version update

No

### DIFF
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -524,6 +524,11 @@ impl CheckpointExecutor {
             }
         }
 
+        if let Some(ref checkpoint_data) = checkpoint_data {
+            self.commit_index_updates_and_enqueue_to_subscription_service(checkpoint_data.clone())
+                .await;
+        }
+
         if !checkpoint.is_last_checkpoint_of_epoch() {
             self.accumulator
                 .accumulate_running_root(epoch_store, checkpoint.sequence_number, checkpoint_acc)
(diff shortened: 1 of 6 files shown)


jkrvivian (approved):

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -524,6 +524,11 @@ impl CheckpointExecutor {
             }
         }
 
+        if let Some(ref checkpoint_data) = checkpoint_data {
+            self.commit_index_updates_and_enqueue_to_subscription_service(checkpoint_data.clone())
```
semenov-vladyslav: `commit_index_updates_and_enqueue_to_subscription_service` only needs `CheckpointData` for `checkpoint_summary.sequence_number`. I'd say if the concern is to avoid extra copies, the best solution would be to change `commit_index_updates_and_enqueue_to_subscription_service` to take `CheckpointData` by reference.

scripts/release_notes/release_notes.py
```diff
@@ -63,6 +63,7 @@
     "CLI",
     "Rust SDK",
     "REST API",
+    "Internal gRPC API",
```
semenov-vladyslav, line 66: We are going to add "External gRPC API" later, right? Otherwise why not just "gRPC API"?

semenov-vladyslav (changes requested):
Let's fix `commit_index_updates_and_enqueue_to_subscription_service` a bit.

scripts/release_notes/release_notes.py
bingyanglin: In the future when the gRPC API is not for experimental/internal usage anymore I guess we will rename it. The word `Internal` was added in purpose in the description at https://github.com/iotaledger/iota/pull/7511.

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
bingyanglin: Agreed. Refactored it at [ed68457](https://github.com/iotaledger/iota/pull/8371/commits/ed68457b4befd9824b07085b8f6d2df9d0eab73a).

semenov-vladyslav (approved):
Much better!

PR #8374, opened 2025-08-27
refactor(ts-sdk): rename util back to `parseAmount`

VmMad:
# Description of change

We use this function to parse coins that are not always specifically IOTA coins, so it renames function back to it's old name since the new name is not accurate to all places used


marc2332 (approved):
:godmode:

KeitoTadashi (approved):

PR #8399, opened 2025-08-28
feat(indexer): Prune optimistic transactions table

### LINKED ISSUE #5930: [indexer]: Prune entries in optimistic indexing tables
To be done at least after: https://github.com/iotaledger/iota/issues/5870

The transactions in optimistic tables should all be checkpointed till the end of epoch, which means that at the end of epoch we can clear all the optimistic tables.

Additionally, the `tx_insertion_order` table could be pruned from entries that reference transactions that are pruned from the indexer (this concerns both optimistically indexed, and checkpointed transactions)

tomxey:
# Description of change

This PR introduces pruner for optimistic_transactions table.
The pruner is disabled by default, unless OPTIMISTIC_PRUNER_BATCH_SIZE env var is set to some value.

fixes: https://github.com/iotaledger/iota/issues/5930

### DIFF
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -626,7 +626,7 @@ pub mod deprecated {
                             .unwrap_or_else(|_e| None),
                         optimistic_pruner_batch_size: std::env::var("OPTIMISTIC_PRUNER_BATCH_SIZE")
                             .map(|s| s.parse::<u64>().ok())
-                            .unwrap_or_else(|_e| None),
+                            .unwrap_or_default(),
                         pruning_config_path: None,
                     },
                     reset_db: old_conf.reset_db,
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -242,6 +242,8 @@ pub struct PruningOptions {
     /// Path to TOML file containing configuration for retention policies.
     #[arg(long)]
     pub pruning_config_path: Option<PathBuf>,
+    #[arg(long, env = "OPTIMISTIC_PRUNER_BATCH_SIZE")]
+    pub optimistic_pruner_batch_size: Option<u64>,
 }
 
 /// Represents the default retention policy and overrides for prunable tables.
@@ -622,6 +624,9 @@ pub mod deprecated {
                         epochs_to_keep: std::env::var("EPOCHS_TO_KEEP")
                             .map(|s| s.parse::<u64>().ok())
                             .unwrap_or_else(|_e| None),
+                        optimistic_pruner_batch_size: std::env::var("OPTIMISTIC_PRUNER_BATCH_SIZE")
+                            .map(|s| s.parse::<u64>().ok())
+                            .unwrap_or_else(|_e| None),
                         pruning_config_path: None,
                     },
                     reset_db: old_conf.reset_db,
@@ -738,6 +743,7 @@ mod test {
         let pruning_options = PruningOptions {
             epochs_to_keep: None,
             pruning_config_path: Some(temp_path.clone()),
+            optimistic_pruner_batch_size: None,
         };
         let retention_config = pruning_options.load_from_file().unwrap().unwrap();
 
@@ -788,6 +794,7 @@ mod test {
         let pruning_options = PruningOptions {
             epochs_to_keep: None,
             pruning_config_path: Some(temp_path.clone()),
+            optimistic_pruner_batch_size: None,
         };
         let retention_config = pruning_options.load_from_file().unwrap().unwrap();
 
--- a/crates/iota-indexer/src/handlers/optimistic_pruner.rs
+++ b/crates/iota-indexer/src/handlers/optimistic_pruner.rs
@@ -54,8 +54,8 @@ impl OptimisticPruner {
             }
 
             match self.prune_single_batch().await {
-                Ok(pruning_occured) => {
-                    pruning_in_progress = pruning_occured;
+                Ok(pruning_occurred) => {
+                    pruning_in_progress = pruning_occurred;
                 }
                 Err(err) => {
                     pruning_in_progress = false;
--- a/crates/iota-indexer/src/handlers/optimistic_pruner.rs
+++ b/crates/iota-indexer/src/handlers/optimistic_pruner.rs
@@ -4,7 +4,7 @@
 use std::time::Duration;
 
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, info, warn};
+use tracing::{info, warn};
 
 use super::checkpoint_handler::CheckpointHandler;
 use crate::{
@@ -44,21 +44,17 @@ impl OptimisticPruner {
     pub async fn start(&self, cancel: CancellationToken) -> IndexerResult<()> {
         info!("Starting Optimistic Pruner task...");
         let mut pruning_in_progress = false;
-        let mut last_pruned_id = (-1, -1);
 
         while !cancel.is_cancelled() {
             if !pruning_in_progress {
                 // let's not spam the DB if there's no pruning to be done
-                tokio::time::sleep(Duration::from_secs(1)).await;
+                tokio::time::sleep(Duration::from_secs(3)).await;
             }
             pruning_in_progress = false;
 
-            match self.prune_single_batch(last_pruned_id).await {
-                Ok(pruned_to) => {
-                    if pruned_to > last_pruned_id {
-                        pruning_in_progress = true;
-                    }
-                    last_pruned_id = pruned_to;
+            match self.prune_single_batch().await {
+                Ok(pruning_occured) => {
+                    pruning_in_progress = pruning_occured;
                 }
                 Err(err) => {
                     warn!("Failed to prune optimistic transaction batch: {err}");
@@ -71,15 +67,17 @@ impl OptimisticPruner {
         Ok(())
     }
 
-    async fn prune_single_batch(&self, last_pruned_id: (i64, i64)) -> IndexerResult<(i64, i64)> {
+    async fn prune_single_batch(&self) -> IndexerResult<bool> {
+        let whole_batch_timer = self.metrics.optimistic_pruner_batch_duration.start_timer();
+
         let current_epoch = self
             .store
             .get_latest_epoch_id_in_blocking_worker()
             .await?
             .unwrap_or(0);
         if current_epoch < EPOCHS_TO_KEEP {
-            debug!("No epochs available for pruning");
-            return Ok(last_pruned_id);
+            info!("No epochs available for optimistic pruning");
+            return Ok(false);
         }
 
         let prune_to_epoch = current_epoch.saturating_sub(EPOCHS_TO_KEEP);
@@ -97,37 +95,29 @@ impl OptimisticPruner {
             .store
             .get_global_order_for_tx_seq_in_blocking_worker(epoch_end_tx)
             .await?;
-        let prune_to = self
-            .store
-            .get_nth_smallest_optimistic_tx_global_order_in_blocking_worker(
-                self.optimistic_pruner_batch_size - 1,
-            )
-            .await?;
 
-        debug!(
-            "Last tx for epoch {prune_to_epoch} has global order {epoch_end_global_order:?}, \
-             next pruning batch ends at {prune_to:?}"
+        let rows_pruned = {
+            let _delete_timer = self
+                .metrics
+                .optimistic_pruner_delete_query_duration
+                .start_timer();
+            self.store
+                .prune_optimistic_transactions_up_to_in_blocking_worker(
+                    epoch_end_global_order,
+                    self.optimistic_pruner_batch_size as i64,
+                )
+                .await?
+        };
+        self.metrics
+            .optimistic_pruner_total_rows_pruned
+            .inc_by(rows_pruned as u64);
+        let elapsed = whole_batch_timer.stop_and_record();
+        info!(
+            "Pruned {rows_pruned} optimistic transactions with limit at {epoch_end_global_order:?} in {elapsed:?} seconds"
         );
 
-        if let Some(prune_to) = prune_to {
-            let prune_to = std::cmp::min(prune_to, epoch_end_global_order);
-            if prune_to <= last_pruned_id {
-                debug!(
-                    "No new optimistic transactions to prune, already pruned to {last_pruned_id:?}"
-                );
-                return Ok(last_pruned_id);
-            }
-            let rows_pruned = self
-                .store
-                .prune_optimistic_transactions_up_to_in_blocking_worker(prune_to)
-                .await?;
-            self.metrics
-                .last_pruned_optimistic_global_seq_num
-                .set(prune_to.0);
-            info!("Pruned {rows_pruned} optimistic transactions up to global order {prune_to:?}",);
-            return Ok(prune_to);
-        }
-
-        Ok(last_pruned_id)
+        // brief pause to give DB time to vacuum deleted rows
+        tokio::time::sleep(Duration::from_millis(200)).await;
+        Ok(rows_pruned > 0)
     }
 }
--- a/crates/iota-indexer/src/handlers/optimistic_pruner.rs
+++ b/crates/iota-indexer/src/handlers/optimistic_pruner.rs
@@ -17,8 +17,8 @@ use crate::{
 // Keeping current and previous epoch ensures we will not prune just executed
 // txs on epoch boundary
 const EPOCHS_TO_KEEP: u64 = 2;
-const PRUNING_NOT_IN_PROGRESS_DELAY: Duration = Duration::from_secs(3);
-const DELETE_ROWS_DELAY: Duration = Duration::from_millis(200);
+const CHECK_EPOCH_CHANGE_INTERVAL: Duration = Duration::from_secs(3);
+const DELAY_BETWEEN_BATCHES: Duration = Duration::from_millis(200);
 
 pub struct OptimisticPruner {
     pub store: PgIndexerStore,
@@ -50,7 +50,7 @@ impl OptimisticPruner {
         while !cancel.is_cancelled() {
             if !pruning_in_progress {
                 // let's not spam the DB if there's no pruning to be done
-                tokio::time::sleep(PRUNING_NOT_IN_PROGRESS_DELAY).await;
+                tokio::time::sleep(CHECK_EPOCH_CHANGE_INTERVAL).await;
             }
 
             match self.prune_single_batch().await {
@@ -117,8 +117,8 @@ impl OptimisticPruner {
             "Pruned {rows_pruned} optimistic transactions with limit at {epoch_end_global_order:?} in {elapsed:?} seconds"
         );
 
-        // brief pause to give DB time to vacuum deleted rows
-        tokio::time::sleep(DELETE_ROWS_DELAY).await;
+        // brief pause to relieve the I/O pressure on the DB
+        tokio::time::sleep(DELAY_BETWEEN_BATCHES).await;
         Ok(rows_pruned > 0)
     }
 }
--- a/crates/iota-indexer/src/handlers/optimistic_pruner.rs
+++ b/crates/iota-indexer/src/handlers/optimistic_pruner.rs
@@ -101,7 +101,8 @@ impl OptimisticPruner {
         Ok(())
     }
 
-    /// Prunes optimistic transactions in batches until there are no more rows to delete.
+    /// Prunes optimistic transactions in batches until there are no more rows
+    /// to delete.
     async fn prune_in_batches(
         &self,
         epoch_end_global_order: TxGlobalOrderCursor,
@@ -160,8 +161,8 @@ impl OptimisticPruner {
         Ok(Some(epoch_end_global_order))
     }
 
-    /// Deletes a single batch of optimistic transactions up to the given global order.
-    /// Returns the number of deleted rows.
+    /// Deletes a single batch of optimistic transactions up to the given global
+    /// order. Returns the number of deleted rows.
     async fn prune_single_batch(
         &self,
         epoch_end_global_order: TxGlobalOrderCursor,
--- a/crates/iota-indexer/src/handlers/optimistic_pruner.rs
+++ b/crates/iota-indexer/src/handlers/optimistic_pruner.rs
@@ -10,7 +10,9 @@ use super::checkpoint_handler::CheckpointHandler;
 use crate::{
     errors::IndexerError,
     metrics::IndexerMetrics,
-    store::{IndexerStore, PgIndexerStore, pg_partition_manager::PgPartitionManager},
+    store::{
+        IndexerStore, PgIndexerStore, TxGlobalOrderCursor, pg_partition_manager::PgPartitionManager,
+    },
     types::IndexerResult,
 };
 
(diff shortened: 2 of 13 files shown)


crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,133 @@
...
+                    continue;
+                }
+            };
+        }
+
+        info!("Optimistic Pruner task cancelled.");
+        Ok(())
+    }
+
+    async fn prune_single_batch(&self, last_pruned_id: (i64, i64)) -> IndexerResult<(i64, i64)> {
```
kodemartin: Have you considered the suggested way to delete batches in the [docs](https://www.postgresql.org/docs/current/sql-delete.html)?

> While there is no LIMIT clause for DELETE, it is possible to get a similar effect using the same method described in [the documentation of UPDATE](https://www.postgresql.org/docs/current/sql-update.html#UPDATE-LIMIT):
```sql
WITH delete_batch AS (
  SELECT l.ctid FROM user_logs AS l
    WHERE l.status = 'archived'
    ORDER BY l.creation_date
    FOR UPDATE
    LIMIT 10000
)
DELETE FROM user_logs AS dl
  USING delete_batch AS del
  WHERE dl.ctid = del.ctid;
```

It will simplify and make the implementation more robust.

tomxey: Simplified in e438a3276bb21dd5c076cb818742ab71a6b3d435

crates/iota-indexer/src/config.rs
```diff
@@ -497,6 +499,9 @@ pub mod deprecated {
                         epochs_to_keep: std::env::var("EPOCHS_TO_KEEP")
                             .map(|s| s.parse::<u64>().ok())
                             .unwrap_or_else(|_e| None),
+                        optimistic_pruner_batch_size: std::env::var("OPTIMISTIC_PRUNER_BATCH_SIZE")
+                            .map(|s| s.parse::<u64>().ok())
+                            .unwrap_or_else(|_e| None),
```
sergiupopescu199: Since `pub optimistic_pruner_batch_size: Option<u64>` we could use `unwrap_or_default()` method, it will default to `None` automatically.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub metrics: IndexerMetrics,
+}
+
+impl OptimisticPruner {
+    pub fn new(
+        store: PgIndexerStore,
+        optimistic_pruner_batch_size: u64,
+        metrics: IndexerMetrics,
+    ) -> Result<Self, IndexerError> {
+        let blocking_cp = CheckpointHandler::pg_blocking_cp(store.clone()).unwrap();
```
sergiupopescu199: Since we return a `Result<Self, IndexerError>` we could handle the error avoiding using `unwrap()`.

crates/iota-indexer/src/indexer.rs
```diff
@@ -98,6 +98,19 @@ impl Indexer {
             spawn_monitored_task!(pruner.start(CancellationToken::new()));
         }
 
+        if let Some(optimistic_pruner_batch_size) = pruning_options.optimistic_pruner_batch_size {
+            info!("Starting indexer optimistic tables pruner");
+            let optimistic_pruner: OptimisticPruner = OptimisticPruner::new(
```
sergiupopescu199: I suppose we can remove the type in the variable declaration since the `OptimisticPruner::new` does not return a generic type but `Self`.

```suggestion
            let optimistic_pruner = OptimisticPruner::new(
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -262,7 +272,9 @@ impl PgIndexerStore {
                 .first::<(i64, Option<i64>)>(conn)
                 .map(|(min, max)| (min as u64, max.map(|v| v as u64)))
         })
-        .context("Failed reading checkpoint range from PostgresDB")
+        .context(
+            format!("Failed reading checkpoint range from PostgresDB for epoch {epoch}").as_str(),
```
sergiupopescu199: Lowercase for error messages

```suggestion
            format!("failed reading checkpoint range from PostgresDB for epoch {epoch}").as_str(),
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -279,7 +291,80 @@ impl PgIndexerStore {
                 .first::<(i64, i64)>(conn)
                 .map(|(min, max)| (min as u64, max as u64))
         })
-        .context("Failed reading transaction range from PostgresDB")
+        .context(
+            format!("Failed reading transaction range from PostgresDB for checkpoint {checkpoint}")
```
sergiupopescu199: lowercase for error messages 

```suggestion
            format!("failed reading transaction range from PostgresDB for checkpoint {checkpoint}")
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -279,7 +291,80 @@ impl PgIndexerStore {
...
+            tx_global_order::dsl::tx_global_order
+                .select((
+                    tx_global_order::global_sequence_number,
+                    tx_global_order::optimistic_sequence_number,
+                ))
+                .filter(tx_global_order::chk_tx_sequence_number.eq(tx_seq))
+                .first::<(i64, i64)>(conn)
+        })
+        .context(
+            format!("Failed reading global sequence number from PostgresDB for tx seq {tx_seq}")
```
sergiupopescu199: lowercase for erro messages 

```suggestion
            format!("failed reading global sequence number from PostgresDB for tx seq {tx_seq}")
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -279,7 +291,80 @@ impl PgIndexerStore {
...
+                     WHERE (ot.global_sequence_number, ot.optimistic_sequence_number) = \
+                           (ids_to_delete.global_sequence_number, ids_to_delete.optimistic_sequence_number)";
+                diesel::sql_query(sql)
+                    .bind::<BigInt, _>(to_global_sequence_num)
+                    .bind::<BigInt, _>(to_optimistic_sequence_num)
+                    .bind::<BigInt, _>(limit)
+                    .execute(conn)
+                    .map_err(IndexerError::from)
+                    .context(
+                        format!("Failed to prune optimistic_transactions table to {to:?} with limit {limit}").as_str(),
```
sergiupopescu199: lowercase for erro messages 

```suggestion
                        format!("failed to prune optimistic_transactions table to {to:?} with limit {limit}").as_str(),
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -1581,7 +1666,7 @@ impl PgIndexerStore {
                 .select(epochs::network_total_transactions)
                 .get_result::<Option<i64>>(conn)
         })
-        .context("Failed to get network total transactions in epoch")
+        .context(format!("Failed to get network total transactions in epoch {epoch}").as_str())
```
sergiupopescu199: lowercase for erro messages 

```suggestion
        .context(format!("failed to get network total transactions in epoch {epoch}").as_str())
````

File: crates/iota-indexer/src/handlers/optimistic_pruner.rs [hunk 10]
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    pub async fn start(&self, cancel: CancellationToken) -> IndexerResult<()> {
+        info!("Starting Optimistic Pruner task...");
+        let mut pruning_in_progress = false;
+
+        while !cancel.is_cancelled() {
+            if !pruning_in_progress {
+                // let's not spam the DB if there's no pruning to be done
+                tokio::time::sleep(Duration::from_secs(3)).await;
```
sergiupopescu199: I suppose we could use a constants here. Fell free to choose a better name representing the behavior.
```rust
const PRUNING_NOT_IN_PROGRESS_DELAY: Duration = Duration::from_secs(3);
tokio::time::sleep(PRUNING_NOT_IN_PROGRESS_DELAY)).await;
```

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            match self.prune_single_batch().await {
+                Ok(pruning_occured) => {
+                    pruning_in_progress = pruning_occured;
+                }
+                Err(err) => {
+                    warn!("Failed to prune optimistic transaction batch: {err}");
+                    continue;
+                }
+            };
```
sergiupopescu199: nit: if you like this more, otherwise ignore it.
```suggestion
let Ok(pruning_occured) = self
        .prune_single_batch()
        .await
        .inspect_err(|e| warn!("Failed to prune optimistic transaction batch: {err}"))
    else {
        continue;
    };
    pruning_in_progress = pruning_occured;
```

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub async fn start(&self, cancel: CancellationToken) -> IndexerResult<()> {
+        info!("Starting Optimistic Pruner task...");
+        let mut pruning_in_progress = false;
+
+        while !cancel.is_cancelled() {
+            if !pruning_in_progress {
+                // let's not spam the DB if there's no pruning to be done
+                tokio::time::sleep(Duration::from_secs(3)).await;
+            }
+            pruning_in_progress = false;
```
sergiupopescu199: we could move the ` pruning_in_progress = false;` after the ` tokio::time::sleep(Duration::from_secs(3)).await;` in this way if `self.prune_single_batch().await == false` we don't over write it as `false` again

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.metrics
+            .optimistic_pruner_total_rows_pruned
+            .inc_by(rows_pruned as u64);
+        let elapsed = whole_batch_timer.stop_and_record();
+        info!(
+            "Pruned {rows_pruned} optimistic transactions with limit at {epoch_end_global_order:?} in {elapsed:?} seconds"
+        );
+
+        // brief pause to give DB time to vacuum deleted rows
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
sergiupopescu199: I suppose we could use a constants here. Fell free to choose a better name representing the behavior.

```rust
const VACUUM_DELETE_ROWS_DELAY: Duration = Duration::from_millis(200); 
tokio::time::sleep(VACUUM_DELETE_ROWS_DELAY).await;

```

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+        let partition_manager = PgPartitionManager::new(blocking_cp.clone())?;
+        Ok(Self {
+            store,
+            partition_manager,
+            optimistic_pruner_batch_size,
+            metrics,
+        })
+    }
+
+    pub async fn start(&self, cancel: CancellationToken) -> IndexerResult<()> {
```
sergiupopescu199: Since you handle the errors by logging them do we need to return a `IndexerResult<()>`?

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        let prune_to_epoch = current_epoch.saturating_sub(EPOCHS_TO_KEEP);
+        let total_txs = self
+            .store
+            .get_network_total_transactions_by_end_of_epoch(prune_to_epoch)
+            .await?
+            .ok_or_else(|| {
+                IndexerError::PostgresRead(format!(
+                    "No network total transactions found for epoch {prune_to_epoch}"
```
sergiupopescu199: lowercase for error messages

```suggestion
                    "no network total transactions found for epoch {prune_to_epoch}"
```

sergiupopescu199 (approved):
lgtm! 🌸

Added some nits but nothing critical.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: moved inside of the `match`

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Having `pruning_in_progress=false` in the Err arm now, I think match is a better option currently.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    pub async fn start(&self, cancel: CancellationToken) {
+        info!("Starting Optimistic Pruner task...");
+        let mut pruning_in_progress = false;
+
+        while !cancel.is_cancelled() {
+            if !pruning_in_progress {
+                // let's not spam the DB if there's no pruning to be done
+                tokio::time::sleep(PRUNING_NOT_IN_PROGRESS_DELAY).await;
```
kodemartin: Checking every `PRUNING_NOT_IN_PROGRESS_DELAY = 3` secs if way to frequent for events that occur every 24 hours. 

Every 1hour would be a sufficient resolution, considering also that this is a maintenance operation that should not affect read queries.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+    errors::IndexerError,
+    metrics::IndexerMetrics,
+    store::{IndexerStore, PgIndexerStore, pg_partition_manager::PgPartitionManager},
+    types::IndexerResult,
+};
+
+// Keeping current and previous epoch ensures we will not prune just executed
+// txs on epoch boundary
+const EPOCHS_TO_KEEP: u64 = 2;
+const PRUNING_NOT_IN_PROGRESS_DELAY: Duration = Duration::from_secs(3);
```
kodemartin: Nit: I would go with a name  that is more explicit about the triggered operation.
```suggestion
const CHECK_EPOCH_CHANGE_INTERVAL: Duration = Duration::from_secs(3);
```

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+const DELETE_ROWS_DELAY: Duration = Duration::from_millis(200);
```
kodemartin: Nit: Again I think this can be more specific

```suggestion
const DELAY_BETWEEN_BATCHES: Duration = Duration::from_millis(200);
```

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+                    warn!("Failed to prune optimistic transaction batch: {err}");
+                    continue;
+                }
+            };
+        }
+
+        info!("Optimistic Pruner task cancelled.");
+    }
+
+    async fn prune_single_batch(&self) -> IndexerResult<bool> {
```
kodemartin: This method is overloaded with side effects that are best left in the discretion of the calling scope. More specifically

1. Check for epoch change
2. Resolution of transaction threshold on epoch end
3. Resolution of `pruning_occurred`.

I don't see why (1) and (2) should happen on every delete operation. Instead it seems more intuitive that when we notice the epoch change in the caller we calculate the transaction threshold and then await for all batches to be deleted.

I would suggest to just stick to what the name suggests: Delete  a single batch and return the number of deleted rows.

Returning also the number of deleted rows give more flexibility to the caller to handle the termination logic.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.metrics
+            .optimistic_pruner_total_rows_pruned
+            .inc_by(rows_pruned as u64);
+        let elapsed = whole_batch_timer.stop_and_record();
+        info!(
+            "Pruned {rows_pruned} optimistic transactions with limit at {epoch_end_global_order:?} in {elapsed:?} seconds"
+        );
+
+        // brief pause to give DB time to vacuum deleted rows
+        tokio::time::sleep(DELETE_ROWS_DELAY).await;
```
kodemartin: This too should be done transparently in the caller.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
```diff
@@ -0,0 +1,124 @@
...
+        };
+        self.metrics
+            .optimistic_pruner_total_rows_pruned
+            .inc_by(rows_pruned as u64);
+        let elapsed = whole_batch_timer.stop_and_record();
+        info!(
+            "Pruned {rows_pruned} optimistic transactions with limit at {epoch_end_global_order:?} in {elapsed:?} seconds"
+        );
+
+        // brief pause to give DB time to vacuum deleted rows
```
kodemartin: This is not precise. Autovacuum is trigerred every "autovacuum_vacuum_threshold + autovacuum_vacuum_scale_factor * table_size" which diverges significantly from the batch size based on default values.

I think the main impact here is the decrease of the I/O load on the database.

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -283,7 +291,80 @@ impl PgIndexerStore {
...
+    fn prune_optimistic_transactions_up_to(
+        &self,
+        to: (i64, i64),
+        limit: i64,
+    ) -> IndexerResult<usize> {
+        let (to_global_sequence_num, to_optimistic_sequence_num) = to;
+        transactional_blocking_with_retry!(
+            &self.blocking_cp,
+            |conn| {
+                let sql = "\
```
kodemartin: You can use raw strings `r#"` for strings that span multiple rows. This will avoid the need for the break `\` character and would allow for easy revisions on the query in the future if necessary.


```rust
                let sql = r#"
                <multi-line-query>
                "#
```

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -283,7 +291,80 @@ impl PgIndexerStore {
...
+    ) -> IndexerResult<usize> {
+        self.execute_in_blocking_worker(move |this| {
+            this.prune_optimistic_transactions_up_to(to, limit)
+        })
+        .await
+    }
+
+    fn prune_optimistic_transactions_up_to(
+        &self,
+        to: (i64, i64),
```
kodemartin: I would suggest splitting into two arguments `global_sequence_num` and `optimistic_sequence_num`.

Constructing and deconstructing the tuple is maintenance overhead given that it is an opaque type. That is the caller cannot no through the type itself what each element of the tuple represents.

Alternatively add a newtype `TxGlobalOrderCursor` or something of the sort to add more semantics.

kodemartin (changes requested):
Hey @tomxey. Looks good but I reckon a few revisions are still necessary. Have a look at more specific comments.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: In some networks we may have more frequent epoch changes. In particular in tests we switch epochs very fast, on demand. So at least for test we would need a way to decrease this delay. But given that checking if there is something to prune didn't seem to be an resource intensive operation I decided to go with such small delay here.

Do you think we should make the delay configurable then?

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Refactored in ed9cda4a80b2dd0a976d92ca6e6f2d44d5046558

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Refactored in ed9cda4a80b2dd0a976d92ca6e6f2d44d5046558

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Done in 2b752d8a9241d6c4d5f34a9fc8ee036ed4ecd6e0

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: Done in 2b752d8a9241d6c4d5f34a9fc8ee036ed4ecd6e0

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: Done in 2b752d8a9241d6c4d5f34a9fc8ee036ed4ecd6e0

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Done in 2b752d8a9241d6c4d5f34a9fc8ee036ed4ecd6e0

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Done in 2b752d8a9241d6c4d5f34a9fc8ee036ed4ecd6e0

crates/iota-indexer/src/handlers/optimistic_pruner.rs
kodemartin: The configuration should primarily focus on the production networks with stable state. Networks with smaller epoch duration still don't need to be pruned more frequently than the granularity imposed by the production networks. 

So I don't think this has to be configurable on the application level yet.

To accommodate testing you could add an `Option` argument that would override this value and then start the pruner explicitly in the tests as a separate process.

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -68,6 +71,28 @@ use crate::{
...
+            optimistic_sequence_number,
+        }
+    }
+}
+
+impl From<(i64, i64)> for TxGlobalOrderCursor {
+    fn from((global_sequence_number, optimistic_sequence_number): (i64, i64)) -> Self {
+        Self::new(global_sequence_number, optimistic_sequence_number)
+    }
+}
```
kodemartin: Nit: I think this is error prone, since it still requires the user to care for the order of the `i64` for values in the tuple, so that it correspond to the correct sequence numbers.

I would suggest just relying on the `new` constructor for this one.

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: I would prefer to test the indexer as a whole, and avoid starting the pruner separately.

What would you say about:

```rust
#[cfg(test)]
const CHECK_EPOCH_CHANGE_INTERVAL: Duration = Duration::from_secs(3);
#[cfg(not(test))]
const CHECK_EPOCH_CHANGE_INTERVAL: Duration = Duration::from_hours(1);
```

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: Refactored code to only use the TxGlobalOrderCursor { } construction

crates/iota-indexer/src/handlers/optimistic_pruner.rs
tomxey: Done in 0877825e7c4b95175a95eded9e21b24207a4ca7e

kodemartin (approved):

PR #8432, opened 2025-09-01
fix(ts-sdk): map `Option` correctly in `moveDataToRpcContent`

### LINKED ISSUE #8368: Fix move data layout of `Option`s in graphql-transport
This fails:

```ts

const client = new IotaClient({
    transport: new IotaClientGraphQLTransport({
        url: getNetwork('devnet').graphql,
        fallbackTransportUrl: getNetwork('devnet').url,
        unsupportedMethods: [
            'multiGetObjects',
            'getReferenceGasPrice',
            'getNormalizedMoveFunction',
        ],
    })
});

(async () => {
    let cursor = undefined;
    do {
        const { data: objectResponse, nextCursor } = await client.getOwnedObjects({
            owner: '0xc764fc6cf091d5c000118c99e9d93fb557aa2b1fd4dff2dca400bbb6d6a41770',
            filter: {
                "MatchNone": [
                    {
                        "StructType": "0x2::coin::Coin"
                    },
                    {
                        "StructType": "0x3::staking_pool::StakedIota"
                    },
                    {
                        "StructType": "0x2::iota::IOTA"
                    },
                    {
                        "StructType": "0x3::timelocked_staking::TimelockedStakedIota"
                    },
                    {
                        "StructType": "0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>"
                    },
                    {
                        "StructType": "0xb9d617f24c84826bf660a2f4031951678cc80c264aebc4413459fb2a95ada9ba::name_registration::NameRegistration"
                    },
                    {
                        "StructType": "0xb9d617f24c84826bf660a2f4031951678cc80c264aebc4413459fb2a95ada9ba::subname_registration::SubnameRegistration"
                    }
                ]
            },
            options: {
                showType: true,
                showContent: true,
                showDisplay: true,
            },
            limit: 50,
            cursor,
        });
        cursor = nextCursor;
        console.log(objectResponse)
    } while (cursor);
})()
```

With:
```
Error: Invalid layout for data: {"Vector":[{"Number":"1"},{"Number":"2"},{"Number":"3"},{"Number":"4"}]}}
```

VmMad:
# Description of change

Fixes #8368 


sdk/graphql-transport/src/mappers/move.ts
```diff
@@ -250,14 +250,40 @@ export function moveDataToRpcContent(data: MoveData, layout: MoveTypeLayout): Mo
...
+        const innerData = data.Option;
+
+        if (
+            typeof innerLayout === 'object' &&
+            'vector' in innerLayout &&
+            innerData &&
+            !('Vector' in innerData)
+        ) {
+            const itemLayout = innerLayout.vector;
+            return moveDataToRpcContent(innerData, itemLayout);
```
marc2332, line 283: This cannot be like this, You are supposing all Options will have a Vector inside, which can definitely not happen. Also, we are already handling `Vectors` just above this in L251, so why cannot that be reused?

marc2332 (changes requested):

sdk/graphql-transport/src/mappers/move.ts
VmMad: we need to make a distintion that "vec" in the layout and "Vector" in the data are two different things. We search for a vec field and we handle cases with nested "Vectors" and without

sdk/graphql-transport/src/mappers/move.ts
```diff
@@ -250,14 +250,40 @@ export function moveDataToRpcContent(data: MoveData, layout: MoveTypeLayout): Mo
...
         return data.Vector.map((item) => moveDataToRpcContent(item, itemLayout));
     }
 
     if ('Option' in data) {
-        return data.Option && moveDataToRpcContent(data.Option, layout);
+        if (data.Option === null) {
+            return null;
+        }
+
+        if (typeof layout !== 'object' || !layout || !('struct' in layout)) {
```
panteleymonchuk:
```suggestion
        if (typeof layout !== 'object' || !('struct' in layout)) {
```
should be enough

VmMad: i think it's not necessary to remove this, just in case we get a null or an empty string, just to be safe

marc2332 (approved):
Nice job, its like `extract_option` from `move_value.rs` so it looks correct!

sdk/graphql-transport/src/mappers/move.ts
panteleymonchuk: alright, let's keep it safe

panteleymonchuk (approved):

sdk/graphql-transport/src/mappers/move.ts
panteleymonchuk: Just by types it can't be null

PR #8458, opened 2025-09-03
upstream(indexer): Modify PruningOptions to point to a toml file of epochs_to_keep and optional per-table overrides

### LINKED ISSUE #8457: upstream(indexer): Modify PruningOptions to point to a toml file of epochs_to_keep and optional per-table override
MystenLabs/sui#19637

tomxey:
# Description of change
Modify `PruningOptions` to point to a toml file that must specify `epochs_to_keep` and an optional [overrides] portion of per-table overrides.

pruner.rs file will define prunable tables. the toml file when parsed into Rust config will warn if it tries to name any variants not supported by indexer code. additionally, there's also a check to make sure that prunable tables actually exist in the db

fixes #8457

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -235,7 +235,8 @@ pub enum Command {
 
 #[derive(Args, Default, Debug, Clone)]
 pub struct PruningOptions {
-    // Argument left for backward compatibility, users are encouraged to use pruning_config_path
+    /// Argument left for backward compatibility, users are encouraged to use
+    /// pruning_config_path
     #[arg(long, env = "EPOCHS_TO_KEEP")]
     pub epochs_to_keep: Option<u64>,
     /// Path to TOML file containing configuration for retention policies.
@@ -257,25 +258,26 @@ pub struct RetentionConfig {
 }
 
 impl PruningOptions {
-    /// Load default retention policy and overrides from file.
+    /// Loads default retention policy and overrides from file.
     pub fn load_from_file(&self) -> IndexerResult<Option<RetentionConfig>> {
-        if let Some(epochs_to_keep) = self.epochs_to_keep {
+        let Some(config_path) = self.pruning_config_path.as_ref() else {
+            let Some(epochs_to_keep) = self.epochs_to_keep else {
+                return Ok(None);
+            };
             warn!(
                 "Using the deprecated --epochs-to-keep argument for pruning configuration. \
-                Please use --pruning-config-path to specify a TOML configuration file instead."
-            );
-            assert!(
-                self.pruning_config_path.is_none(),
-                "Both --epochs-to-keep and --pruning-config-path cannot be provided at the same time."
+                 Please use --pruning-config-path to specify a TOML configuration file instead."
             );
             return Ok(Some(RetentionConfig::new(
                 epochs_to_keep,
                 Default::default(),
             )));
-        }
+        };
 
-        let Some(config_path) = self.pruning_config_path.as_ref() else {
-            return Ok(None);
+        if self.epochs_to_keep.is_some() {
+            warn!(
+                "The --epochs-to-keep argument will be ignored since --pruning-config-path is also provided."
+            );
         };
 
         let contents = std::fs::read_to_string(config_path)
@@ -302,10 +304,12 @@ impl PruningOptions {
 }
 
 impl RetentionConfig {
-    /// Create a new `RetentionConfig` with the specified default retention and
-    /// overrides. Call `finalize()` on the instance to update the
-    /// `policies` field with the default retention policy for all tables
-    /// that do not have an override specified.
+    /// Creates a new `RetentionConfig` with the specified default retention and
+    /// overrides.
+    ///
+    /// Call `finalize()` on the instance to update the `policies` field with
+    /// the default retention policy for all tables that do not have an
+    /// override specified.
     pub fn new(epochs_to_keep: u64, overrides: HashMap<PrunableTable, u64>) -> Self {
         Self {
             epochs_to_keep,
@@ -323,11 +327,13 @@ impl RetentionConfig {
         Self::new(epochs_to_keep, HashMap::new())
     }
 
-    /// Consumes this struct to produce a full mapping of every prunable table
-    /// and its retention policy. By default, every prunable table will have
-    /// the default retention policy from `epochs_to_keep`. Some tables like
-    /// `objects_history` will observe a different default retention policy.
-    /// These default values are overridden by any entries in `overrides`.
+    /// Consumes the struct and produces a mapping of every prunable table
+    /// and its retention policy.
+    ///
+    /// By default, every prunable table will have the default retention policy
+    /// from `epochs_to_keep`. Some tables like `objects_history` will
+    /// observe a different default retention policy. These default values
+    /// are overridden by any entries in `overrides`.
     pub fn retention_policies(self) -> HashMap<PrunableTable, u64> {
         let RetentionConfig {
             epochs_to_keep,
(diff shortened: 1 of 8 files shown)


crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +230,96 @@ pub enum Command {
...
+    /// A map of tables to their respective retention policies that will
+    /// override the default. Prunable tables not named here will use the
+    /// default retention policy.
+    #[serde(default)]
+    pub overrides: HashMap<PrunableTable, u64>,
+}
+
+impl PruningOptions {
+    /// Load default retention policy and overrides from file.
+    pub fn load_from_file(&self) -> Option<RetentionConfig> {
```
sergiupopescu199: From my understanding in this function we will never get the `None` variant, manieng that we could change the function signature to:

```rust
pub fn load_from_file(&self) -> IndexerResult<RetentionConfig> {}
```
Possibly replacing panics with `Result::Error` variant for better erro outputs.

Nit: Although `Indexer::start_writer_with_config` accepts an `Option<RetentionConfig>`, we have two options to handle this:

1. **Wrap the result at the call site**: Simply wrap the result in `Some()` when calling the function
2. **Refactor the caller to be more flexible**: Change the function parameter to accept `impl Into<Option<RetentionConfig>>`. This would allow the caller to pass either:
   - `Some(retention_config)` explicitly, or 
   - `retention_config` directly (which would be automatically wrapped in `Some`)
   
   The second approach provides better ergonomics by eliminating the need for manual wrapping while maintaining the same functionality. Should not introduce breaking changes.

crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +230,96 @@ pub enum Command {
...
+    #[arg(long)]
+    pub pruning_config_path: Option<PathBuf>,
+}
+
+/// Represents the default retention policy and overrides for prunable tables.
+/// Instantiated only if `PruningOptions` is provided on indexer start.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RetentionConfig {
+    /// Default retention policy for all tables.
+    pub epochs_to_keep: u64,
```
sergiupopescu199, line 252: Mostly a genuine question: Since we accept any value except 0, it could be feasible to use [NonZeroU64](https://doc.rust-lang.org/std/num/type.NonZeroU64.html)?

crates/iota-indexer/src/handlers/pruner.rs
```diff
@@ -63,30 +130,38 @@ impl Pruner {
...
+                        if cancel.is_cancelled() {
+                            info!("Pruner task cancelled.");
+                            return Ok(());
+                        }
+                        self.partition_manager
+                            .drop_table_partition(table_name.clone(), epoch)?;
+                        info!(
+                            "Batch dropped table partition {} epoch {}",
+                            table_name, epoch
+                        );
```
sergiupopescu199:
```suggestion
                        info!("Batch dropped table partition {table_name} epoch {epoch}");
```

sergiupopescu199 (approved):
lgtm ✨

some optional nits, feel free to skip if not feasible.

crates/iota-indexer/src/config.rs
tomxey: Actually the function will return None in case `self.pruning_config_path` is None.

crates/iota-indexer/src/config.rs
tomxey: I think it would be possible, but such change would influence many places in code, and I would like to skip it for now.

crates/iota-indexer/src/config.rs
tomxey: Modified the function to return errors instead of some panics

crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +235,116 @@ pub enum Command {
 
 #[derive(Args, Default, Debug, Clone)]
 pub struct PruningOptions {
+    // Argument left for backward compatibility, users are encouraged to use pruning_config_path
```
kodemartin:
```suggestion
    /// Argument left for backward compatibility, users are encouraged to use pruning_config_path
```

crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +235,116 @@ pub enum Command {
...
+    pub fn load_from_file(&self) -> IndexerResult<Option<RetentionConfig>> {
+        if let Some(epochs_to_keep) = self.epochs_to_keep {
+            warn!(
+                "Using the deprecated --epochs-to-keep argument for pruning configuration. \
+                Please use --pruning-config-path to specify a TOML configuration file instead."
+            );
+            assert!(
+                self.pruning_config_path.is_none(),
+                "Both --epochs-to-keep and --pruning-config-path cannot be provided at the same time."
+            );
```
kodemartin: This doesn't have to panic. A `warn` could be issue that the configuration file will be used instead. Panicking on the application level is not ideal for this benign user-input conflict.

crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +235,116 @@ pub enum Command {
...
+
+        Ok(Some(retention_with_overrides))
+    }
+}
+
+impl RetentionConfig {
+    /// Create a new `RetentionConfig` with the specified default retention and
+    /// overrides. Call `finalize()` on the instance to update the
+    /// `policies` field with the default retention policy for all tables
+    /// that do not have an override specified.
```
kodemartin: Rust doc conventions: Try to comply to [guidelines](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#summary-sentence).

This applies to other doc strings as well.

crates/iota-indexer/src/config.rs
```diff
@@ -224,8 +235,116 @@ pub enum Command {
...
+        );
+
+        Self::new(epochs_to_keep, HashMap::new())
+    }
+
+    /// Consumes this struct to produce a full mapping of every prunable table
+    /// and its retention policy. By default, every prunable table will have
+    /// the default retention policy from `epochs_to_keep`. Some tables like
+    /// `objects_history` will observe a different default retention policy.
+    /// These default values are overridden by any entries in `overrides`.
```
kodemartin: Rust doc guidelines.

crates/iota-indexer/src/handlers/pruner.rs
```diff
@@ -18,26 +20,91 @@ use crate::{
...
+    strum_macros::EnumString,
+    strum_macros::EnumIter,
+    strum_macros::AsRefStr,
+    Hash,
+    Serialize,
+    Deserialize,
+    Clone,
+)]
+#[strum(serialize_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
```
kodemartin, line 45: Add `#[non_exhaustive]`.

crates/iota-indexer/src/indexer.rs
```diff
@@ -33,29 +33,12 @@ use crate::{
...
-        Indexer::start_writer_with_config(
-            config,
-            store,
-            metrics,
-            snapshot_config,
-            PruningOptions::default(),
-            CancellationToken::new(),
-        )
-        .await
-    }
```
kodemartin, line 51: Isn't it used at all?

kodemartin (commented):
Looks good. Left a few suggestions nevertheless that require minor changes.

crates/iota-indexer/src/indexer.rs
tomxey: We only use the `start_writer_with_config`

crates/iota-indexer/src/config.rs
tomxey: Reformatted it a bit.

crates/iota-indexer/src/config.rs
tomxey: Reformatted it a bit.

kodemartin (approved):
lgtm

PR #8472, opened 2025-09-05
feat(iota-faucet): address rate limitting 

shufps:
# Description of change

Introduces rate-limit per address for the faucet.

There are three new parameters:

`--enable-rate-limiting`
Enable per-recipient rate limiting. Disabled by default.

`--max-requests-per-window <N> (default: 12)`
Maximum number of requests allowed per recipient within the sliding window.

`--rate-window-secs <SECONDS> (default: 43200)`
Size of the sliding window, in seconds (e.g. 43200 = 12 hours).

- [X] Tested on Testnet for a 3 months

### Release Notes

- [X] Faucet / `iota-tools`

### DIFF
--- a/crates/iota-faucet/src/faucet/mod.rs
+++ b/crates/iota-faucet/src/faucet/mod.rs
@@ -129,6 +129,19 @@ pub struct FaucetConfig {
 
     #[arg(long, action = clap::ArgAction::Set, default_value_t = false)]
     pub batch_enabled: bool,
+
+    // enable rate limiting
+    #[arg(long, default_value_t = false)]
+    pub enable_rate_limiting: bool,
+
+    // fields for rate limiting
+    // default 12 requests
+    #[arg(long, default_value_t = 12)]
+    pub max_requests_per_window: usize,
+
+    // default: 12 hours
+    #[arg(long, default_value_t = 43200)]
+    pub rate_window_secs: u64,
 }
 
 impl Default for FaucetConfig {
@@ -147,6 +160,9 @@ impl Default for FaucetConfig {
             batch_request_size: 500,
             ttl_expiration: 300,
             batch_enabled: false,
+            enable_rate_limiting: false,
+            max_requests_per_window: 12,
+            rate_window_secs: 12,
         }
     }
 }
--- a/crates/iota-faucet/src/faucet/mod.rs
+++ b/crates/iota-faucet/src/faucet/mod.rs
@@ -80,6 +80,9 @@ pub trait Faucet {
 
     /// Get the status of a batch_send request
     async fn get_batch_send_status(&self, task_id: Uuid) -> Result<BatchSendStatus, FaucetError>;
+
+    /// Apply rate limiting
+    async fn rate_limit(&self, recipient: IotaAddress) -> Result<(), FaucetError>;
 }
 
 pub const DEFAULT_AMOUNT: u64 = 1_000_000_000;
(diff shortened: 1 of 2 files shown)


crates/iota-faucet/src/faucet/simple_faucet.rs
```diff
@@ -898,6 +910,59 @@ impl Faucet for SimpleFaucet {
     ) -> Result<FaucetReceipt, FaucetError> {
         info!(?recipient, uuid = ?id, ?amounts, "Getting faucet requests");
 
+        // If rate limiting is enabled, perform the rate check.
+        if self.enable_rate_limiting {
+            let mut request_times = self.request_times.lock().await;
```
sergiupopescu199: it could be possible to move the code to a separate function in the `Impl SimpleFaucet block`?

```rust
async fn rate_limit(&self, recipient: IotaAddress) -> Result<(), FaucetError> {
        let mut request_times = self.request_times.lock().await;

        // Define the time window based on configuration.
        let window = Duration::from_secs(self.rate_window_secs);
        let now = Instant::now();

        // clean map
        if request_times.len() >= MAX_TRACKED_ADDRESSES {
            request_times.retain(|_, times| {
                times.retain(|&t| now.duration_since(t) < window);
                !times.is_empty()
            });
            if request_times.len() >= MAX_TRACKED_ADDRESSES
                && !request_times.contains_key(&recipient)
            {
                return Err(FaucetError::BatchSendQueueFull);
            }
        }

        let entries = request_times.entry(recipient).or_insert_with(VecDeque::new);

        // Remove timestamps older than the configured window.
        entries.retain(|&timestamp| now.duration_since(timestamp) < window);

        // Check if the number of requests in the window exceeds the configured limit.
        if entries.len() >= self.max_requests_per_window {
            info!(
                "{:?} has {} requests in the past {} seconds; blocking.",
                recipient,
                entries.len(),
                self.rate_window_secs
            );
            return Err(FaucetError::BatchSendQueueFull);
        }

        // Debug: log before recording the new request.
        debug!(
            "Allowing a new request for recipient {:?}; current count: {}",
            recipient,
            entries.len()
        );

        // Record the current request.
        entries.push_back(now);

        Ok(())
    }
```
The manual `drop` is not needed since it will be automatically be dropped once the `rate_limit` method will be out of scope.

The code in `send` method could look like this:

```rust
  // If rate limiting is enabled, perform the rate check.
  if self.enable_rate_limiting {
      self.rate_limit(recipient).await?;
  }
```

crates/iota-faucet/src/faucet/mod.rs
```diff
@@ -129,6 +129,19 @@ pub struct FaucetConfig {
 
     #[arg(long, action = clap::ArgAction::Set, default_value_t = false)]
     pub batch_enabled: bool,
+
+    // enable rate limiting
```
sergiupopescu199: This comments seems redundant since it's the same variable name

crates/iota-faucet/src/faucet/mod.rs
```diff
@@ -129,6 +129,19 @@ pub struct FaucetConfig {
...
+    #[arg(long, default_value_t = false)]
+    pub enable_rate_limiting: bool,
+
+    // fields for rate limiting
+    // default 12 requests
```
sergiupopescu199: Here we also already have all needed info, comment seems redundant

crates/iota-faucet/src/faucet/simple_faucet.rs
```diff
@@ -898,6 +910,59 @@ impl Faucet for SimpleFaucet {
...
+                "Allowing a new request for recipient {:?}; current count: {}",
+                recipient,
+                entries.len()
+            );
+
+            // Record the current request.
+            entries.push_back(now);
+
+            // Release the lock.
+            drop(request_times);
```
sergiupopescu199: no need for manual drop since the MutexGuard will be released automatically and the end of the scope in this case at line 962

sergiupopescu199 (approved):
lgtm 🌸

I've added some nits which could improve the code a bit.

junwei0117 (approved):
LGTM

Issue #8433, opened 2025-09-01
Set an appropriate gas cost for the new native functions


TheMrAI:
`create_auth_info_v1`
`create_auth_info_self_v1` (if it exists)

miker83z:
Closed in favor of #8556

Issue #8416, opened 2025-09-01
[AA] Ensure that the feature integrates with current testing frameworks


TheMrAI:
(This issue is deliberately left a little vague as we don't know the exact constraints.)

The `create_auth_info` and `AuthContext` or `authenticate` feature must be properly integrated into the current testing infrastructure. Both on the user and developer side.

The list of currently known scenarios:
- built in move unit testing framework https://intro.sui-book.com/unit-three/lessons/7_unit_testing.html
- tests initiated by `move test` (likely similar/same to the first case)
- the iota-transactional-adapter-tests
- simulator tests

What we need to pay utmost attention to is that the packages we are trying to load for the given function may not be deployed, may not have an address, may have to be loaded from some other source.
This could become tricky depending on how the implementations are done.

One case for consideration:
Upgradeable package A has an `authenticate` function and an `elevate_to_account` (which takes an object and elevates it to an account by attaching the `authenticate` function) which uses the `authenticate` function in this version.
Then package A is upgraded and `elevate_to_account` is modified such that it will refer to the first version of the `authenticate` function.

Issue #8215, opened 2025-08-15
Investigate the options for private `authenticate` functions


TheMrAI:
At the moment `iota-execution/latest/iota-verifier/src/account_auth_verifier.rs` only accepts
`authenticate` functions which are public, but we may need to alleviate this restriction.

```
Not for this first iteration of development, but we should consider whether it's worth relax this constraint to allow public and private:

we can execute private functions from the rust side
a dev, setting this function as private, means that this is declared as not taking part to composability with other authenticate() (same logic as using just entry for normal functions)
```

```
I was thinking about this as well, but couldn't really figure out what would make sense, so landed on leaving it public.
Didn't want to potentially break some move invariants. For example if I allow private functions to be called, I need to somehow also enforce that
the create_auth_info call originates from the same module as the targeted authenticate is in.
```

PR #8477, opened 2025-09-05
feat(iota-framework): Add create_auth_info_v1 native function

### LINKED ISSUE #8117: [AA][auth] Create an Account pointing to an `authenticate()` function

### LINKED ISSUE #8216: Figure out how much `authenticate` function arguments must be restricted
An `authenticate` function for AA has to look like:
```
public fun name(...., auth: &AuthContext, txc: &TxContext)  
```

The dotted argument section can contain any number of arguments, but it isn't entirely clear exactly which types.
They can't be mutable references, that is clear or problems may arise.

But what else can't they be and more importantly why?
SignatureToken (`external-crates/move/crates/move-binary-format/src/file_format.rs`) contains a number of other types.

TheMrAI:
# Description of change

Introduce create_auth_info_v1 native functions.
This function provides the capability of creating an AuthenticatorInfoV1 pointing to a desired authenticate
function.

The gas costs for the native function haven't been identified yet. Tracking issue: #8433
The native function is not yet tested or integrated into the testing frameworks. Tracking issue: #8416
The possibility of enabling the use of private `authenticate` function will be investigated here: #8215

fixes #8117
fixes #8216 

The `create_auth_info_v1` feature hasn't been tested yet, but the verifier was in a separate PR. 

### Release Notes

- [x] Protocol: Introduce new move native function create_auth_info_v1 and create_auth_info_self_v1 with its gas costs.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-move/src/unit_test.rs
```diff
@@ -125,6 +125,9 @@ fn new_testing_object_and_natives_cost_runtime(ext: &mut NativeContextExtensions
     let metrics = Arc::new(LimitsMetrics::new(&registry));
     let store = Lazy::force(&TEST_STORE);
 
+    // If you this list needs to be updated you likely need to update
```
Thoralf-M:
```suggestion
    // If this list needs to be updated you likely need to update
```

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+            "Invalid pure type. Signer is not a pure type, offending argument: {:?}",
+            param
+        )),
+        Reference(_) => Err(format!(
+            "Invalid pure type. Reference is not a pure type, offending argument: {:?}",
+            param
+        )),
+        MutableReference(_) => Err(format!(
+            "Invalid pure type. MutableReference is not a pure type, offending argument: {:?}",
+            param
```
Thoralf-M: Can do that in all the places in this file
```suggestion
            "Invalid pure type. MutableReference is not a pure type, offending argument: {param:?}"            
```

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -0,0 +1,50 @@
+module iota::account;
```
Thoralf-M, line 4: Missing copyright header

crates/iota-framework-snapshot/tests/compatibility_tests.rs
```diff
@@ -88,6 +88,7 @@ mod compatibility_tests {
     /// match the metadata in the `Move.toml` files in the repo for each
     /// revision.
     #[test]
+    #[ignore = "let me see that CI pass without this nonsense"]
```
valeriyr: Just not to forget to revert this. Are we going to revert this, right?

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -0,0 +1,50 @@
...
+/// In fact package A may have a dependency on package B version 1, but can still refer to an `authenticate`
+/// function defined in package B version 2.
+/// Refiring to an `authenticate` function with `create_auth_info_v1` is a strictly runtime dependency and
+/// it does not collide with any compile time restrictions.
+///
+/// The only scenario which cannot be handled by this function is that of referring to an `authenticate`
+/// function from the current version of the package. Simply because the current package address won't be know
+/// before publishing, thus the user cannot specify it in code.
+/// If an `authenticate` function should be used from the current version of the package please use
+/// `create_auth_info_self_v1` instead.
```
valeriyr: I would add this in the PR related to `create_auth_info_self_v1`.
And I would change the description here, because we can't refer to an unpublished package; it doesn't matter if it is `self` or not.
If we want to refer to a package where the authenticator is defined, we can do this with a dedicated Move function call after the package is published.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2140,6 +2148,10 @@ impl ProtocolConfig {
                         // max auth gas budget is in NANOS and an absolute value 1IOTA
                         cfg.max_auth_gas = Some(1_000_000_000);
                         cfg.feature_flags.move_auth = true;
+                        // === Native Function Costs ===
+                        // `account` module
+                        cfg.create_auth_info_v1_cost_base = Some(1000); /* TODO how do you decide on a good gas * 
+                        price? */
```
valeriyr: Let's remove this TODO and create a dedicated issue(if it is not created)

crates/iota-types/src/account.rs
```diff
@@ -17,16 +17,16 @@ use crate::{
 pub const AUTHENTICATOR_DF_NAME: &str = "IOTA_AUTHENTICATION";
 
 pub const AUTHENTICATOR_INFO_MODULE_NAME: &IdentStr = ident_str!("account");
-pub const AUTHENTICATOR_INFO_STRUCT_NAME: &IdentStr = ident_str!("AuthenticatorInfo");
+pub const AUTHENTICATOR_INFO_STRUCT_NAME: &IdentStr = ident_str!("AuthenticatorInfoV1");
```
valeriyr: I think we need to add `V1` to all these names.
Because we suppose `AuthenticatorInfoV2` can be created, in this case, we will need to add new constants for it.

crates/iota-types/src/account.rs
```diff
@@ -17,16 +17,16 @@ use crate::{
...
```
valeriyr, line 18: Remove the following comment, please:
```rust
// Temporary created structures.
// This part will be removed once the real types are implemented.
```

crates/iota-types/src/account.rs
```diff
@@ -38,32 +38,31 @@ impl AuthenticatorInfo {
...
+            error: format!("Unable to deserialize AuthenticatorInfoV1 object: {err}"),
         })
     }
 
-    // TODO: Needs to be moved to MoveObjectType.
-    pub fn is_authenticator_info(other: &MoveObjectType) -> bool {
-        other.address() == IOTA_FRAMEWORK_ADDRESS
-            && other.module() == AUTHENTICATOR_INFO_MODULE_NAME
-            && other.name() == AUTHENTICATOR_INFO_STRUCT_NAME
+    pub fn is_authenticator_info(tag: &StructTag) -> bool {
```
valeriyr: Should this function be called `is_authenticator_info_v1`? To be clearer about what it does.

crates/iota-types/src/auth_context.rs
```diff
@@ -70,4 +77,42 @@ impl AuthContext {
...
+
+        let S::Datatype(idx) = &**token else {
+            return AuthContextKind::None;
+        };
+
+        let (module_addr, module_name, struct_name) = resolve_struct(module, *idx);
+
+        let is_tx_context_type = module_name == AUTH_CONTEXT_MODULE_NAME
+            && module_addr == &IOTA_FRAMEWORK_ADDRESS
+            && struct_name == AUTH_CONTEXT_STRUCT_NAME;
```
valeriyr: This part can be extracted in a dedicated function, like it is done for `AuthenticatorInfoV1`

crates/iota-types/src/auth_context.rs
```diff
@@ -70,4 +77,42 @@ impl AuthContext {
     pub fn to_bcs_bytes(&self) -> Vec<u8> {
         bcs::to_bytes(&self).unwrap()
     }
+
+    /// Returns whether the type signature is &mut TxContext, &TxContext, or
+    /// none of the above.
```
valeriyr:
```suggestion
    /// Returns whether the type signature is &mut AuthContext, &AuthContext, or
    /// none of the above.
```

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: I can, but I don't see why this pattern is any good in general.
Doesn't work in the general case, where you don't have a clean variable name, but an expression `something.get_value()`.

external-crates/move/crates/move-binary-format/src/file_format.rs
```diff
@@ -1077,14 +1077,27 @@ pub enum SignatureToken {
...
+    /// public struct Data {}
+    /// ```
     Datatype(DatatypeHandleIndex),
+    /// User defined type with template parameters.
+    /// ```
+    /// public struct Data<T: drop> has drop {
+    ///     v: T
+    /// }
+    /// ```
     DatatypeInstantiation(Box<(DatatypeHandleIndex, Vec<SignatureToken>)>),
```
valeriyr, line 1091: I am not 100% sure, it would be nice to check the information in this comment.
Looks like these two descriptions are a bit misleading. These enum elements represent not a type declaration, but a type when it is used somewhere.
Something like this:
```rust
public struct A { ... }

public struct B {
    x: A, // <- `A` is Datatype
    y: Balance<A>, // <- `Balance<A>` is DatatypeInstantiation
}

public fun with_datatype(x: A) {  // <- `A` is Datatype
    ...
}

public fun with_datatype_instantiation(x: Balance<A>) {  // <- `Balance<A>` is DatatypeInstantiation
    ...
}
```

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
Thoralf-M: Yes doesn't work in all cases, but here are only clean variable names
There is also a (pedantic) clippy lint that suggests that
https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+
+/// Verify if a given function can be used as an `authenticate` function
+///
+/// A function is an authenticate function if:
+/// - only has read-only inputs (immutable owned/shared references or pure
+///   types)
+/// - has no return type
+/// - the last two arguments in order are AuthContext and TxContext
+/// - AuthContext has to be an immutable reference
+/// - TxContext hat to be an immutable reference
```
valeriyr:
```suggestion
/// - TxContext has to be an immutable reference
```

valeriyr (commented):
Looks good!
The way you updated the documentation comments related to IDs can be used as an example for our future PRs :+1:

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
TheMrAI: Ehh, there is some merit to it. Will add it, but it would be better if we just enabled the clippy checks and let that handle this outside of PRs.

crates/iota-framework-snapshot/tests/compatibility_tests.rs
TheMrAI: https://github.com/iotaledger/iota/pull/8477/commits/9da63dedf2d18077048974f2ea34b07cefbe58c4 
with small correction. I will only revert it if the PR is approved. I want to see the CI pass, when all requests have been fixed.

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
Thoralf-M: Fine by me, was just a suggestion :smile:

crates/iota-types/src/auth_context.rs
```diff
@@ -70,4 +77,42 @@ impl AuthContext {
...
+        if is_tx_context_type {
+            kind
+        } else {
+            AuthContextKind::None
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub enum AuthContextKind {
```
theiari, line 111: shouldn't this enum be placed near the top of the file? I think it's better readable in this way (?)

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -0,0 +1,50 @@
...
+    create_auth_info_v1_impl(package, module_name.as_bytes(), function_name.as_bytes())
+}
+
+native fun create_auth_info_v1_impl(
+    package: address,
+    module_name: &vector<u8>,
+    function_name: &vector<u8>,
+): AuthenticatorInfoV1;
+
+public fun drop_auth_info_v1(auth_info: AuthenticatorInfoV1) {
```
miker83z: This function can be avoided by adding the `drop` ability to `AuthenticatorInfoV1`. It can be added because the inner types all have `drop`. Moreover, I guess we could also add the `copy` ability as there could be use cases in which you might want to use the same authenticator without the need to re-validate it.

miker83z (approved):

iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+        verify_authenticate_param_type(module, &function_handle.type_parameters, param)
+            .map_err(verification_failure)?;
+    }
+
+    // Check type of AuthContext and TxContext, they both must be structs with the
+    // appropriate names, addresses and access
+    let auth_context = &function_signature.0[function_signature.len() - 2];
+    let tx_context = &function_signature.0[function_signature.len() - 1];
+
+    // AuthContext could potentially passed as value, but that opens up the
```
theiari:
```suggestion
    // AuthContext could potentially be passed as a value, but that opens up the
```

crates/iota-framework/packages/iota-framework/sources/account.move
TheMrAI: I don't understand what you are looking for.
The section you highlighted describes that you can't refer to the current package directly from the source itself, because of the described restriction. What is the issue with this section? It merely describes a pattern that they might want to use initially but cant and would wonder why not.

I will move the final sentence as that doesn't make sense: 
```
/// If an `authenticate` function should be used from the current version of the package please use
/// `create_auth_info_self_v1` instead.
```

| And I would change the description here
But you don't describe how you want it to be changed. I can't tell if you want me to make an example, want to describe to the user what you made the description for, or change something else.
What would you like to see?

crates/iota-types/src/account.rs
TheMrAI: Only if you insist but I prefer not to.
The code is already chock full of completely needless redundancy.
In this case the call-site looks like:
```
 MoveObjectType_::Other(s) => AuthenticatorInfoV1::is_authenticator_info(s),
 ```
 In fact even this is very redundant, but didn't want to change it too much.
 
 With your request it would become:
 ```
 MoveObjectType_::Other(s) => AuthenticatorInfoV1::is_authenticator_info_v1(s),
 ```

File: crates/iota-framework/packages/iota-framework/sources/account.move (hunk 15 above)
Comment (TheMrAI) [reply crates/iota-framework/packages/iota-framework/sources/account.move]: The `drop` ability wasn't added as per specification.
Currently without `drop` and `copy` the user may only create exactly one instance of `AuthenticatorInfoV1` if I understand move properly.
This provides a protection against weird patterns.
Lets say I have an `authenticate` function:
```
public fn authenticate(another: AuthenticatorInfoV1, auth: AuthenticatorInfoV1, ctx: TxContext) {
    ...
    arbitrary_package::authenticate(another, ctx)
    ...
}
```
I can brake the initial rules we set out for ourselves.
This can be a pattern you are looking for, if you are sure, I can do as you request.

Edit: Not sure exactly if you can do the above or not or what else can be broken if the restrictions are lifted.

File: crates/iota-types/src/auth_context.rs (hunk 14 above)
Comment (TheMrAI) [reply crates/iota-types/src/auth_context.rs:111]: In general no.
Structs, enums etc don't belong to the top of the source file out of principle.
They belong closest to the point they are used in (if the language allows, after the place of use).
This way you can read the code top to bottom, always finding the new things you are reading right below the piece of code you just inspected.

Now of course this pattern can fall apart very quickly as well, if not maintained, but it is still better than separating the use from the definition.

In this case it works very well as this enum is not used for anything else, but for that function only.

File: crates/iota-types/src/auth_context.rs (hunk 10 above)
Comment (TheMrAI) [reply crates/iota-types/src/auth_context.rs]: As you noticed from the erroneously unmodified comment this function comes from https://github.com/iotaledger/iota/blob/baecdc08d14d047575bed78bb8c8c1dce1a64204/crates/iota-types/src/base_types.rs#L1063 

I would avoid doing so, unless you feel strongly about this.
If I remember correctly we don't check these anywhere else for `AuthContext` only in the verifier which aims to follow the same pattern as done for `TxContext`.

File: crates/iota-types/src/account.rs (hunk 9 above)
Comment (valeriyr) [reply crates/iota-types/src/account.rs]: Yes, I understand this, but we don't have a struct `AuthenticationInfo`, it is not clear what `is_authenticator_info` means if it is versioned. But I don't insist here, if you think it is better without the version let's leave it as is.

File: crates/iota-framework/packages/iota-framework/sources/account.move (hunk 5 above)
Comment (valeriyr) [reply crates/iota-framework/packages/iota-framework/sources/account.move]: You can refer to the `self` package if you pass the address as a parameter to an entry Move function. You will have such an entry function anyway. The only place where you don't have it is the package `init` function, it is called automatically during publishing(we need to check if it works in this case).

```
The only scenario that cannot be handled by this function is that of referring to an `authenticate` function from the current version of the package.
```

What does `current` version mean? If you can call a function on-chain, you have the address. The use case of this function is to omit passing the address as a parameter, but in this case, we need to clarify better what `self` is.

File: crates/iota-types/src/auth_context.rs (hunk 10 above)
Comment (valeriyr) [reply crates/iota-types/src/auth_context.rs]: I don't see a reason not to create a dedicated function. It is absolutely logical to have such, and the `kind` function will be shorter/cleaner. We can use it later somewhere if it is necessary.

btw: there is a typo `is_tx_context_type` -> `is_auth_context_type`

File: external-crates/move/crates/move-binary-format/src/file_format.rs (hunk 12 above)
Comment (TheMrAI) [reply external-crates/move/crates/move-binary-format/src/file_format.rs:1091]: You may be right, I don't know. I tried to untangle what it meant by writing example code and dumping the type information.
In the comment, you see the example code I used in my tests.

File: crates/iota-framework/packages/iota-framework/sources/account.move (hunk 5 above)
Comment (TheMrAI) [reply crates/iota-framework/packages/iota-framework/sources/account.move]: Alright, I don't follow what is that you are looking for, but from what I gather if I remove the section you will be happy so I will do so.

File: crates/iota-types/src/account.rs (hunk 7 above)
Comment (TheMrAI) [reply crates/iota-types/src/account.rs]: Could you be please be more specific?
Do you want to update only `AUTHENTICATOR_INFO_STRUCT_NAME`?
Or do you want to update: `AUTHENTICATOR_INFO_STRUCT_NAME`, `AUTHENTICATOR_INFO_MODULE_NAME`?
How about the module name `account`?

If we make a new version do we want to put it into the same module?

File: crates/iota-move/src/unit_test.rs (hunk 1 above)
Comment (TheMrAI) [reply crates/iota-move/src/unit_test.rs]: Fixed in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-framework/packages/iota-framework/sources/account.move (hunk 3 above)
Comment (TheMrAI) [reply crates/iota-framework/packages/iota-framework/sources/account.move:4]: Good catch! I thought this was auto checked, but it isn't for `.move` files. :(
Fixed in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-framework/packages/iota-framework/sources/account.move (hunk 5 above)
Comment (TheMrAI) [reply crates/iota-framework/packages/iota-framework/sources/account.move]: Removed in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-protocol-config/src/lib.rs (hunk 6 above)
Comment (TheMrAI) [reply crates/iota-protocol-config/src/lib.rs]: Removed in: 0765331206ba40def3c666c3b4fc76f4eb4acada
The ticket was there https://github.com/iotaledger/iota/issues/8433 just the comment was left in.

File: crates/iota-types/src/account.rs (hunk 8 above)
Comment (TheMrAI) [reply crates/iota-types/src/account.rs:18]: Good catch, removed in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-types/src/account.rs (hunk 9 above)
Comment (TheMrAI) [reply crates/iota-types/src/account.rs]: Well, I think it is considerably nastier but updated in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-types/src/auth_context.rs (hunk 11 above)
Comment (TheMrAI) [reply crates/iota-types/src/auth_context.rs]: Good catch, updated in: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: iota-execution/latest/iota-verifier/src/account_auth_verifier.rs (hunk 13 above)
Comment (TheMrAI) [reply iota-execution/latest/iota-verifier/src/account_auth_verifier.rs]: Updated: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: iota-execution/latest/iota-verifier/src/account_auth_verifier.rs (hunk 16 above)
Comment (TheMrAI) [reply iota-execution/latest/iota-verifier/src/account_auth_verifier.rs]: Ehh, missed the `a`, but good enough: 0765331206ba40def3c666c3b4fc76f4eb4acada

File: crates/iota-types/src/account.rs (hunk 7 above)
Comment (valeriyr) [reply crates/iota-types/src/account.rs]: I meant constants, we can use the following names, for example:
```rust
// We can leave it as is or rename to `AUTHENTICATOR_V1_DF_NAME`. Depends on what we want to say with this name.
pub const AUTHENTICATOR_DF_NAME: &str = "IOTA_AUTHENTICATION";

// Change the name of the constant. 
pub const ACCOUNT_MODULE_NAME: &IdentStr = ident_str!("account");
// Change the name of the constant
pub const AUTHENTICATOR_INFO_V1_STRUCT_NAME: &IdentStr = ident_str!("AuthenticatorInfoV1");
```

crates/iota-types/src/auth_context.rs
TheMrAI: Regardless, this function is different to the pattern you are referring to in `AuthenticatorInfoV1`.
There you merely check a StructTag, which comes from an Object. Some input type validation, not entirely sure.

Here we are collecting the data from the CompiledModule and then confirming if it is as we expect.
I can add the StructTag support to this one and then `is_auth_context` function can be extracted.
Is that what you are looking for?

crates/iota-types/src/account.rs
valeriyr: I also thought about the module name, probably `account` is not the best option. We can have a directory that we can call `account` or `account_abstraction` and put all the related modules inside. But it can be done in a dedicated PR.

valeriyr: If we leave the `account` module with `AuthenticatorInfoV1`, we will create `AuthenticatorInfoV2` inside it as well

crates/iota-types/src/auth_context.rs
valeriyr: Ah, yes, it is not a StructTag. But we can create a function anyway. The code will be like this:
```rust
let (module_addr, module_name, struct_name) = resolve_struct(view, *idx);

if is_auth_context(module_addr, module_name, struct_name) {
    kind
} else {
    TxContextKind::None
}
```
but, ok, it is not really so important. let's leave it as is.

valeriyr (approved):

theiari (approved):

crates/iota-framework/packages/iota-framework/sources/account.move
TheMrAI: Snippet from Slack discussion:
The example you made is not correct. Adding copy won't allow you to do that. Instead, it would "just" allow you to create an instance of AuthenticatorInfoV1 without invoking create_auth_info_v1_impl. This allows to create a new instance by copying from one already existing. So, if I am not overseeing some additional particular implications, it should be safe because:
1. You can create an AuthenticatorInfoV1 instance using create_auth_info_v1_impl -> safe
2. You can create an AuthenticatorInfoV1 instance by copying from another an AuthenticatorInfoV1 instance; this other one was either copied (recursion to this point 2) or created using create_auth_info_v1_impl (point 1) -> safe by inheritance

crates/iota-types/src/auth_context.rs
TheMrAI: I fail to see how that would help in any meaningful way.
At the moment we are following a pattern that is the same as in TxContext.
At the moment we don't need to extract the functionality.
But if this struct needs to later support the StructTag, then we would have to again resolve the duplication or sphagettify further.

crates/iota-types/src/account.rs
TheMrAI: How about this?
I will make a ticket describing that this can be investigated and update the 'AUTHENTICATOR_INFO_V1_STRUCT_NAME' for now.

crates/iota-framework/packages/iota-framework/sources/account.move
TheMrAI: Fixed in: ea9c587d3b333d57d57ee16647ec0048c45d65e5

crates/iota-types/src/account.rs
TheMrAI: Fixed in: 2a17efb5abdce284c2fcac42d2b66f3b64c25faf
Ticket: https://github.com/iotaledger/iota/issues/8498

PR #8487, opened 2025-09-08
feat(iota-framework): Add create_auth_info_self_v1

### LINKED ISSUE #8415: Provide a `create_auth_info_self` function

TheMrAI:
# Description of change

This function relies on the currently active stack frame from the
Interpreter for identifying in which package it is in.
A few preconditions are used here. First we know that within a
given package all module references are resolved during runtime.
The stack frames contain the runtime id's of each module, which
together with the resolved package linkage uniquely identify
the module source. Allowing the function to reclaim the appropriate
CompiledModule.
    
To allow this with minimal modification DataStorage was propagated
into NativeContext.
This makes the code more confusing, because the Loader does not store
DataStorage at all, but always receives it.
If we decide there is no need for create_auth_info_self, then we
can remove this dependency.

fixes #8415 

### Release Notes

- [x] Protocol: Introduce new move native function create_auth_info_self_v1 with its gas costs.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


external-crates/move/crates/move-vm-runtime/src/native_functions.rs
```diff
@@ -104,6 +106,7 @@ pub struct NativeContext<'a, 'b> {
     extensions: &'a mut NativeContextExtensions<'b>,
     gas_left: RefCell<InternalGas>,
     gas_budget: InternalGas,
+    data_store: &'a dyn DataStore,
```
miker83z, line 109: Overall it's a good solution! I am only a bit hesitant just about this change.

PR #8492, opened 2025-09-08
feat(dapp-kit): add names support to account switcher

### LINKED ISSUE #8480: Add name support to the dapp-kit wallet switcher
Add name support to the dapp-kit wallet switcher

Image

KeitoTadashi:
# Description of change

add names support to dapp-kit

fixes #8480

### DIFF
--- a/sdk/dapp-kit/package.json
+++ b/sdk/dapp-kit/package.json
@@ -75,6 +75,7 @@
         "vitest": "^2.1.9"
     },
     "dependencies": {
+        "@iota/iota-names-sdk": "^0.3.0",
         "@iota/iota-sdk": "workspace:*",
         "@iota/wallet-standard": "workspace:*",
         "@radix-ui/react-dialog": "^1.1.1",
--- a/sdk/dapp-kit/src/components/AccountDropdownMenu.tsx
+++ b/sdk/dapp-kit/src/components/AccountDropdownMenu.tsx
@@ -16,15 +16,24 @@ import { ChevronIcon } from './icons/ChevronIcon.js';
 import { StyleMarker } from './styling/StyleMarker.js';
 import { Button } from './ui/Button.js';
 import { Text } from './ui/Text.js';
+import { useGetDefaultIotaName } from '../hooks/useGetDefaultIotaName.js';
 
 type AccountDropdownMenuProps = {
     currentAccount: WalletAccount;
     size?: React.ComponentProps<typeof Button>['size'];
+    iotaNamesEnabled: boolean;
 };
 
-export function AccountDropdownMenu({ currentAccount, size = 'lg' }: AccountDropdownMenuProps) {
+export function AccountDropdownMenu({
+    currentAccount,
+    size = 'lg',
+    iotaNamesEnabled,
+}: AccountDropdownMenuProps) {
     const { mutate: disconnectWallet } = useDisconnectWallet();
     const accounts = useAccounts();
+    const { data: iotaName } = useGetDefaultIotaName(currentAccount.address, iotaNamesEnabled);
+    const displayAccount =
+        iotaName ?? currentAccount.label ?? formatAddress(currentAccount.address);
 
     return (
         <DropdownMenu.Root modal={false}>
@@ -32,7 +41,7 @@ export function AccountDropdownMenu({ currentAccount, size = 'lg' }: AccountDrop
                 <DropdownMenu.Trigger asChild>
                     <Button size={size} className={styles.connectedAccount}>
                         <Text mono weight="bold">
-                            {currentAccount.label ?? formatAddress(currentAccount.address)}
+                            {displayAccount}
                         </Text>
                         <ChevronIcon />
                     </Button>
@@ -47,6 +56,7 @@ export function AccountDropdownMenu({ currentAccount, size = 'lg' }: AccountDrop
                                     key={account.address}
                                     account={account}
                                     active={currentAccount.address === account.address}
+                                    iotaNamesEnabled={iotaNamesEnabled}
                                 />
                             ))}
                         </div>
@@ -67,18 +77,22 @@ export function AccountDropdownMenu({ currentAccount, size = 'lg' }: AccountDrop
 export function AccountDropdownMenuItem({
     account,
     active,
+    iotaNamesEnabled,
 }: {
     account: WalletAccount;
     active?: boolean;
+    iotaNamesEnabled: boolean;
 }) {
     const { mutate: switchAccount } = useSwitchAccount();
+    const { data: iotaName } = useGetDefaultIotaName(account.address, iotaNamesEnabled);
+    const displayAccount = iotaName ?? account.label ?? formatAddress(account.address);
 
     return (
         <DropdownMenu.Item
             className={clsx(styles.menuItem, styles.switchAccountMenuItem)}
             onSelect={() => switchAccount({ account })}
         >
-            <Text mono>{account.label ?? formatAddress(account.address)}</Text>
+            <Text mono>{displayAccount}</Text>
             {active ? <CheckIcon /> : null}
         </DropdownMenu.Item>
     );
--- a/sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
+++ b/sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
@@ -1,0 +1,43 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { getNetwork } from '@iota/iota-sdk/client';
+import { useQuery } from '@tanstack/react-query';
+import { IotaNamesClient, normalizeIotaName } from '@iota/iota-names-sdk';
+import { useIotaClientContext } from './useIotaClient.js';
+import { useMemo } from 'react';
+import { IotaGraphQLClient } from '@iota/iota-sdk/graphql';
+
+export function useGetDefaultIotaName(
+    address: string | null | undefined,
+    iotaNamesEnabled: boolean,
+) {
+    const iotaContext = useIotaClientContext();
+    const network = getNetwork(iotaContext.network);
+
+    const iotaNamesClient = useMemo(() => {
+        const iotaGraphQLClient = new IotaGraphQLClient({
+            url: network.graphql!,
+        });
+
+        return new IotaNamesClient({
+            graphQlClient: iotaGraphQLClient,
+            network: network.id,
+        });
+    }, [network.id]);
+
+    const iotaNamesSupported = !!iotaNamesClient.config;
+
+    return useQuery({
+        queryKey: ['iota-name', 'default-name', address],
+        queryFn: async () => {
+            if (!address) return null;
+
+            const name = await iotaNamesClient?.getDefaultName(address);
+
+            return name ? normalizeIotaName(name) : name;
+        },
+        enabled: !!iotaNamesClient && !!address && iotaNamesEnabled && iotaNamesSupported,
+        staleTime: 1000 * 60 * 10,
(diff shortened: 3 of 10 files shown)


sdk/dapp-kit/src/hooks/useFeatureEnabledByNetwork.ts
marc2332, line 1: We can't put this hook in the dapp kit

sdk/dapp-kit/src/enums/features.enums.ts
marc2332, line 1: Neither this

sdk/dapp-kit/src/hooks/useNetwork.ts
marc2332, line 1: Neither this

sdk/dapp-kit/src/hooks/useFeatureEnabledByNetwork.ts
marc2332, line 1: Or this

sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
```diff
@@ -2,23 +2,22 @@
...
+import { useFeatureEnabledByNetwork } from './useFeatureEnabledByNetwork.js';
+import { Feature } from '../enums/features.enums.js';
 import { useQuery } from '@tanstack/react-query';
 import { normalizeIotaName } from '@iota/iota-names-sdk';
 
 export function useGetDefaultIotaName(
     address: string | null | undefined,
     normalized: boolean = true,
 ) {
+    const { iotaNamesClient } = useIotaNamesClient();
```
marc2332: To be honest, Instead of pulling the graphql client context and the iota names client context here you could simple create a new iota names client on the fly without using those contexts at all, these clients are stateless, so they are cheap to create and destroy, to create a name client you only A: graphql client and B: the network id, something like this:

```ts
const iotaNamesClient = useMemo(() => {
    const iotaGraphQLClient = new IotaGraphQLClient({
        url: network.graphql!,
    });

    return new IotaNamesClient({
        graphQlClient: iotaGraphQLClient,
        network: network.id,
    });
}, [network.id]);
```

You can get the network id using the iota client context and with `getNetwork` you can get the full network object

```ts

const iotaContext = useIotaClientContext();
const network = getNetwork(iotaContext.network);
```

I haven't tested this but it should work, let me know if you have any doubts

sdk/dapp-kit/src/contexts/IotaNamesClientContext.tsx
marc2332, line 1: Move it back

sdk/dapp-kit/src/contexts/IotaGraphQLClientContext.tsx
marc2332, line 1: Move it back

sdk/dapp-kit/src/components/AccountDropdownMenu.tsx
```diff
@@ -72,13 +76,16 @@ export function AccountDropdownMenuItem({
     active?: boolean;
 }) {
     const { mutate: switchAccount } = useSwitchAccount();
-
+    const { data: iotaName } = useGetDefaultIotaName(account.address);
+    console.log('account', account.address);
```
marc2332: debris

sdk/dapp-kit/src/hooks/useGetIotaNameRecord.ts
marc2332, line 1: This was not necessary to move at all I think

sdk/dapp-kit/src/utils/validation/names.ts
marc2332, line 1: can be moved back

sdk/dapp-kit/package.json
```diff
@@ -75,6 +75,8 @@
         "vitest": "^2.1.9"
     },
     "dependencies": {
+        "@growthbook/growthbook-react": "^1.0.0",
```
marc2332: Can be reverted

marc2332 (changes requested):

sdk/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -47,6 +47,9 @@ export type WalletProviderProps = {
     /** The theme to use for styling UI components. Defaults to using the light theme. */
     theme?: Theme | null;
 
+    /** Key to store if iota names feature is enabled */
```
marc2332:
```suggestion
    /** Enable IOTA Names resolution */
```

marc2332 (changes requested):
Does it actually work? 🤔 

Image

Can you also check why the CI is failing?

apps/wallet-dashboard/providers/AppProviders.tsx
```diff
@@ -14,65 +14,76 @@ import {
...
 
 growthbook.init();
 
-export function AppProviders({ children }: React.PropsWithChildren) {
+function InnerProviders({ children }: React.PropsWithChildren) {
     const [queryClient] = useState(() => new QueryClient());
     const allNetworks = getAllNetworks();
     const defaultNetwork = getDefaultNetwork();
     const [persistedNetwork] = useLocalStorage<string>('network_iota-dashboard', defaultNetwork);
+    const isNameResolutionEnabled = useFeatureEnabledByNetwork(Feature.IotaNames, defaultNetwork);
```
marc2332:
```suggestion
    const isNameResolutionEnabled = useFeatureEnabledByNetwork(Feature.IotaNames, persistedNetwork);
```

marc2332 (changes requested):

marc2332:
Seems like if you dont remove the old localstorage `iotaNames` wont evet get enabled, maybe you can look into "forcing" the value to be there? Maybe using the wallet context was a bad idea, perhaps another simpler context is just better.

KeitoTadashi:
> Seems like if you dont remove the old localstorage `iotaNames` wont evet get enabled, maybe you can look into "forcing" the value to be there? Maybe using the wallet context was a bad idea, perhaps another simpler context is just better.

if Feature is enabled now we update localstorage

apps/wallet/src/ui/app/hooks/useGetDefaultIotaName.ts
marc2332, line 1: Why dont you use the one from the dappkit?

marc2332 (changes requested):
Seems like the wallet got broken somehow

apps/wallet/src/ui/app/hooks/useGetDefaultIotaName.ts
KeitoTadashi: because wallet does not use WalletProvider, where we are using `iotaNames={isNameResolutionEnabled}`

sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function useGetDefaultIotaName(
+    address: string | null | undefined,
+    normalized: boolean = true,
+) {
+    const iotaContext = useIotaClientContext();
+    const network = getNetwork(iotaContext.network);
+    const iotaNamesEnabled = useIotaNamesEnabled();
+
+    const iotaNamesClient = useMemo(() => {
```
panteleymonchuk, line 18: If we'll reuse other parts of iotaNamesClient, I'd create `useIotaNamesClient` hook

sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useQuery } from '@tanstack/react-query';
+import { IotaNamesClient, normalizeIotaName } from '@iota/iota-names-sdk';
+import { useIotaClientContext } from './useIotaClient.js';
+import { useMemo } from 'react';
+import { IotaGraphQLClient } from '@iota/iota-sdk/graphql';
+import { useIotaNamesEnabled } from './wallet/useIotaNamesEnabled.js';
+
+export function useGetDefaultIotaName(
+    address: string | null | undefined,
+    normalized: boolean = true,
```
panteleymonchuk: I'd rename this property.
`normalized` define fact.

Imo better to use:
- `normalize`
- `shouldNormalize`
- `applyNormalization`

apps/wallet/src/ui/app/hooks/useGetDefaultIotaName.ts
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Feature, useFeatureEnabledByNetwork, useIotaNamesClient, useNetwork } from '@iota/core';
+import { normalizeIotaName } from '@iota/iota-names-sdk';
+import { getNetwork } from '@iota/iota-sdk/client';
+import { useQuery } from '@tanstack/react-query';
+
+export function useGetDefaultIotaName(address: string | undefined) {
+    const { iotaNamesClient } = useIotaNamesClient();
+    const networkName = useNetwork();
```
panteleymonchuk: why we don't reuse hook from dapp here?

KeitoTadashi: because wallet does not use WalletProvider, where we are using iotaNames={isNameResolutionEnabled}

marc2332 (dismissed):

panteleymonchuk (dismissed):

sdk/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -47,6 +48,9 @@ export type WalletProviderProps = {
     /** The theme to use for styling UI components. Defaults to using the light theme. */
     theme?: Theme | null;
 
+    /** Enable IOTA Names resolution */
+    iotaNames?: boolean;
```
VmMad: use descriptive names for props
```suggestion
    isIotaNamesEnabled?: boolean;
```

sdk/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -60,8 +64,13 @@ export function WalletProvider({
     enableUnsafeBurner = false,
     autoConnect = false,
     theme = lightTheme,
+    iotaNames = false,
     children,
 }: WalletProviderProps) {
+    const [names, setNames] = useState({ enable: iotaNames });
+    if (iotaNames != names.enable) {
+        setNames({ enable: iotaNames });
+    }
```
VmMad: Why is this necessary?

sdk/dapp-kit/src/hooks/useGetDefaultIotaName.ts
VmMad, line 1: I don't think this should live here. This is a specific hook for iota names that is mostly not related to the dapp kit

VmMad (changes requested):
I don't think this makes sense as it is right now. At least not for an external dApp. They will probably just pass a hardcoded `true` or `false` to the `iotaNames` prop that you defined and based on that, the default names will try to be fetched. But what if a dapp has a network switcher? Or if it's a mainnet dapp? It will start throwing errors, and even more if the `useGetDefaultIotaName` is exported from the `dapp-kit` and we want to avoid that.

What i recommend is:
1. Take `useGetDefaultIotaName` out of the dapp-kit
2. Create a standalone `useGetDefaultIotaName` in the dapp kit but not export it outside of the library
3. When creating the iotaNamesClient in the `useGetDefaultIotaName`, check if there's `config`, and based on that enable the query or not. This way we avoid relying on a prop that we know it's uncertain if it will make it work or not. We can still pass a prop to enable/disable the names fetching in the WalletProvider

VmMad (approved):

marc2332 (approved):

PR #8473, opened 2025-09-05
feat(wallet): Import Keystone accounts

marc2332:
Closes https://github.com/iotaledger/iota/issues/8466

image

PR #8494, opened 2025-09-09
feat(wallet): Keystone Account Sources

### LINKED ISSUE #8478: Add (if possible) a keystone account source
This way we might allow multiple keystone wallets to be used.

Maybe the master fingerprint can be used.

marc2332:
Based off https://github.com/iotaledger/iota/pull/8473

Closes https://github.com/iotaledger/iota/issues/8478


apps/wallet/src/background/account-sources/accountSource.ts
```diff
@@ -14,6 +14,7 @@ import {
 export enum AccountSourceType {
     Mnemonic = 'mnemonic',
     Seed = 'seed',
+    Keystone = 'keystone',
 }
```
Thoralf-M, line 18: Btw why is ledger not listed here?

marc2332: Because ledger doesnt have an account source, just accounts

marc2332: because I dont think we could differentiate multiple ledgers or something like that

apps/wallet/src/background/account-sources/keystoneAccountSource.ts
```diff
@@ -0,0 +1,130 @@
...
+    }: {
+        password: string;
+        masterFingerprint: string;
+    }) {
+        // This way the UI can always know the source id by just knowing the master fingerprint
+        const id = `keystone-${masterFingerprint}`;
+        const dataSerialized: KeystoneAccountSourceSerialized = {
+            id,
+            type: AccountSourceType.Keystone,
+            encryptedData: await encrypt(password, {}),
```
msarcev, line 45: Why is there no payload that is encrypted here ?

marc2332: Because there is simply nothing to encrypt, so I kinda left it there in case we need to add something in the future. Same happens in LedgerAccounts

PR #8500, opened 2025-09-09
upstream(node): Delay SpawnOnce completion until server bind() completes

jkrvivian:
# Description of change

- Upstream range: [v1.44.3, v1.45.3)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/a2e9d75ce69fbed2561b13f4b1787ad4d0357039
- Description:

Apparently bind() is what initiates request handling, so it must be delayed.
This should fix client-side panics caused by validators serving requests after restart, but before they have replayed all pending consensus transactions.

The `tokio::spawn` handle was removed since it was unused and not the correct handle for shutting down the server. We now retain only the `server.handle()` within `SpawnOnce`, which is the proper handle for shutdown.

We also rewrite the logic to remove `SpawnOnce` and `oneshot::channel`, and make it more straightforward.

Part of #6153 

### DIFF
--- a/crates/iota-node/src/lib.rs
+++ b/crates/iota-node/src/lib.rs
@@ -140,7 +140,7 @@ use simulator::*;
 use tap::tap::TapFallible;
 use tokio::{
     runtime::Handle,
-    sync::{Mutex, broadcast, mpsc, watch},
+    sync::{Mutex, broadcast, mpsc, oneshot, watch},
     task::{JoinHandle, JoinSet},
 };
 use tokio_util::sync::CancellationToken;
@@ -897,7 +897,7 @@ impl IotaNode {
 
             // Start the gRPC server
             components.validator_server_spawn_handle =
-                components.validator_server_spawn_handle.start();
+                components.validator_server_spawn_handle.start().await;
 
             Some(components)
         } else {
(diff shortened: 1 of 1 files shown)


semenov-vladyslav (changes requested):
This probably works. However, the design is overcomplicated IMO, we don't need SpawnOnce thing. This should be simplified/refactored.

crates/iota-node/src/lib.rs
```diff
@@ -2167,44 +2165,58 @@ impl IotaNode {
...
-    #[allow(unused)]
-    Started(iota_http::ServerHandle),
+type ServerFuture =
+    Box<dyn Fn(iota_network_stack::server::Server) -> BoxFuture<'static, Result<()>> + Send + Sync>;
+
+struct ValidatorServerHandle {
+    // Mutex is only needed to make ValidatorServerHandle Send
+    bind_future: Option<Mutex<BoxFuture<'static, Result<iota_network_stack::server::Server>>>>,
+    serve_future: Option<ServerFuture>,
+    server_handle: Option<iota_http::ServerHandle>,
```
bingyanglin: Do we need `bind_future` and `server_future` in the `ValidatorServerHandle`? Or can we just have `server_handle` in the structure, and make the `bind_future` and `serve_future` passed in the start() function?

jkrvivian: No, both futures are constructed during `ValidatorComponent` initialization. And `start()` can only be called after the `ValidatorComponent` is ready. I don't see a way to have both futures outside and send it to `start()` currently.

bingyanglin (approved):

semenov-vladyslav (approved):

PR #8529, opened 2025-09-11
feat(iota-core, iota-types): Introduction of EndOfPublishV2 transaction 

oliviasaa:
```mermaid
---
config:
  gitGraph:
    {parallelCommits: true,
    mainBranchName: 'Feature',
    rotateCommitLabel: true}
---
gitGraph
   commit id: "Initial commit"
   branch Scorer
   commit id: "Add Scorer type"
   checkout Feature
    branch EndOfPublishV2
   commit id: "Add EndOfPublishV2 type"
   checkout Feature
    branch ChangeEpochV4
   commit id: "Add ChangeEpochV4 type"
   checkout Feature
   branch mergedbranch
   merge ChangeEpochV4 id: "Merge ChangeEpochV4"
   merge EndOfPublishV2 id: "Merge EndOfPublishV2" 
    merge Scorer id: "Merge Scorer"
   commit id: "Add protocol configs"
      branch process-EOPV2
   commit id: "EndOfPublishV2 processing"
    checkout mergedbranch
    branch advance-epoch
   commit id: "add advance_epoch_v4"
    checkout mergedbranch
    branch send-EOPV2
   commit id: "EndOfPublishV2 dissemination"
checkout Feature
    merge EndOfPublishV2 tag: "THIS PR"

```

# Description of change

We introduce a new type of end of publish message (`EndOfPublishV2`) containing a vector of `u64`. This vector represents the local view of the authority about all validators' partial scores

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Introduces EndOfPublishV2
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-types/src/messages_consensus.rs
+++ b/crates/iota-types/src/messages_consensus.rs
@@ -192,7 +192,7 @@ impl AuthorityCapabilitiesV1 {
 pub enum ConsensusTransactionKind {
     CertifiedTransaction(Box<CertifiedTransaction>),
     CheckpointSignature(Box<CheckpointSignatureMessage>),
-    EndOfPublishV1(AuthorityName),
+    EndOfPublish(AuthorityName),
     EndOfPublishV2(AuthorityName, Vec<u64>),
 
     CapabilityNotificationV1(AuthorityCapabilitiesV1),
@@ -326,7 +326,7 @@ impl ConsensusTransaction {
         let tracking_id = hasher.finish().to_le_bytes();
         Self {
             tracking_id,
-            kind: ConsensusTransactionKind::EndOfPublishV1(authority),
+            kind: ConsensusTransactionKind::EndOfPublish(authority),
         }
     }
 
@@ -423,7 +423,7 @@ impl ConsensusTransaction {
                     data.summary.sequence_number,
                 )
             }
-            ConsensusTransactionKind::EndOfPublishV1(authority)
+            ConsensusTransactionKind::EndOfPublish(authority)
             | ConsensusTransactionKind::EndOfPublishV2(authority, _) => {
                 ConsensusTransactionKey::EndOfPublish(*authority)
             }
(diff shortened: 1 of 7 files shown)


crates/iota-types/src/messages_consensus.rs
```diff
@@ -192,7 +192,8 @@ impl AuthorityCapabilitiesV1 {
 pub enum ConsensusTransactionKind {
     CertifiedTransaction(Box<CertifiedTransaction>),
     CheckpointSignature(Box<CheckpointSignatureMessage>),
-    EndOfPublish(AuthorityName),
+    EndOfPublishV1(AuthorityName),
```
cyberphysic4l: I'm not sure we can rename this, we may have to leave it as EndOfPublish for the sake of backward compatibility so nodes running older software versions can still deserialise it. I still struggle to understand backwards compatibility sometimes so not sure and we should check this.

oliviasaa: I'll summon @piotrm50 here.

My reasoning was that this message type is something "internal" to an epoch, and that if someone wanted to replay   the chain, it wouldn't even use or see any consensus EndOfPublish message. I could be wrong, of course.

cyberphysic4l: The part I think is problematic is not about replaying but when validator A has latest software that works up to protocol version X+1, but validator B has an older software capable of max protocol version X, and protocol version X is active. Validator A should be sending `EndOfPublish` transactions for as long as protocol version is X so that validator B can deserialise it, as validator B will not know about any new types on their old software version.

oliviasaa: ok, got it!

VorobyevIlya (commented):
Everything is fine, except for the problem with renaming Andrew mentioned above

crates/iota-types/src/messages_consensus.rs
polinikita: Just consider an edge case when there is not enough votes for a new version of the protocol. Say, 50% of validators stay on on old software and send EndOfPublish at the end of the epoc, while the remaining 50% of validators are upgraded to a new software and send EndOfPublishV1 messages. Validators from different groups will not deserialise these consensus messages and no part will advance to the next epoch. 

So, the approach should be to keep EndOfPublish and add a new EndOfPublishV2 such that sending and parsing consensus messages depends on the protocol config parameter.

oliviasaa: yes, I got it! I'm pushing those changes right now

PR #8537, opened 2025-09-11
feat(iota-core): improve gas price prediction function

vekkiokonio:
# Description of change

Previously, gas price estimation was based on the sum of hotness weights of all shared objects in a transaction. This approach made the prediction strictly dependent on the number of objects, which is not necessarily an accurate reflection of transaction cost.

Based on experimental results, the following improvements were made:
- Optimized learning (`HOTNESS_ADJUSTMENT_FACTOR`) and decay (`HOTNESS_DECAY_FACTOR`) parameters.
- Updated gas price prediction to use the maximum hotness among the objects in a transaction, instead of the sum.
- Fixed tests to validate the updated behavior.

These changes are important because gas price prediction directly influences the update rule for object hotness. By providing better feedback, the system can maintain more accurate hotness values over time.

**EDIT**

Additionally, we modify the hotness update logic for successfully executed transactions. Specifically, we now account for potential overpayment estimation. Since recomputing the sequencing to evaluate overpayment is too costly, we propose approximating this by leveraging `lowest_clearing_gas_price` which is already existing in the `CongestionTracker` module.

fixes https://github.com/iotaledger/iota-private/issues/217.
fixes https://github.com/iotaledger/iota-private/issues/224.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Improved accuracy in the computation of object hotness and gas price prediction.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-core/src/congestion_tracker.rs
```diff
@@ -378,16 +406,45 @@ impl CongestionTracker {
...
+                    .and_modify(|info| info.hotness += hotness_adjustment);
+
+                objects_with_mutated_hotness.push(max_object_id);
+            }
+        }
+
+        // Decay hotness for objects that were not updated in this checkpoint
+        for (object_id, info) in congestion_info_map.iter_mut() {
+            if !objects_with_mutated_hotness.contains(object_id) {
+                info.hotness = f64::MAX;
```
muXxer, line 435: So the hotness is set to maximum value? Can we add a comment here why it is the maximum?

vekkiokonio: This value represents the amount that will be subtructed from the current hotness of the object -- see `update_hotness` function. Here, `info.hotness` is set to maximum meaning that we decrease the current hotness as max as possible because the object was not present in the current checkpoint. The maximum hotness decrease is bounded by `MAX_DECAY_FACTOR`. I updated the description.

muXxer (approved):

piotrm50 (approved):

piotrm50 (approved):
I assume that none of those changes are protocol-breaking in a way that would make two nodes order transaction differently if one runs with this change and another runs the old version within the same epoch?

vekkiokonio:
> I assume that none of those changes are protocol-breaking in a way that would make two nodes order transaction differently if one runs with this change and another runs the old version within the same epoch?

Nope, the change is local to the computation of hotness inside `congestion_tracker`, which will be used by the `get_suggested_gas_price_with_ogd` function to predict gas price of shared-object transactions. Hence, the change doesn't affect any transaction ordering or sequencing.

PR #8539, opened 2025-09-11
feat(iota-e2e-tests): Provide e2e tests for create_auth_info feature 

### LINKED ISSUE #8414: Provide e2e tests for `create_auth_info` feature
There are multiple distinct components in the way of easily testing `create_auth_info`.

The verifier we have to run is "runtime" not "compile time" as every other verifier currently in Move. In other terms this means that while other verifiers run on publishing/loading a package this verifier runs when executing code.

There are multiple testing frameworks for Move, each with their distinctive hacks, especially concerning verifiers and where the currently compiled packages are posted and what their address will be.
`create_auth_info` relies on knowing exactly what address the target package is at, which is a pattern directly not supported by any testing framework, so we must use e2e tests.

### LINKED ISSUE #8525: Investigate why vector tests fail in e2e tests
The added e2e tests in: crates/iota-e2e-tests/tests/authenticate_verifier_tests.rs
seem to work, except for a lot (probably all actually, just the failing are expected to fail) **vector** for some reason.

They produce error like so:
```
iota_adapter_latest::execution_engine::checked: iota-execution/latest/iota-adapter/src/execution_engine.rs:163: INVARIANT VIOLATION! Source: Some(VMError { major_status: UNKNOWN_INVARIANT_VIOLATION_ERROR, sub_status: None, message: Some("Referenced module:\n    e364f602ea1cdb65bb6f567da18ec125011902ebc26e4359faa3bc29c21c3ace::vector unavailable"), exec_state: None, location: Module(ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000002, name: Identifier("account") }), indices: [], offsets: [(FunctionDefinitionIndex(1), 0)] }) kind=VMInvariantViolation tx_digest=TransactionDigest(6YvzopE3x3EcxFt5CR4oJuwUCUAYdvLDV9moJ5soNVSW)
```
Not sure how this can happen. `vector` is built in `std` library, always available.
According to this error it seems that it is looking for `vector` on address `e364f602ea1cdb65bb6f567da18ec125011902ebc26e4359faa3bc29c21c3ace` instead of the expected

TheMrAI:
# Description of change

Added a number of e2e tests for the `iota-execution/latest/iota-verifier/src/account_auth_verifier.rs`.

fixes #8414 
fixes #8525 


crates/iota-e2e-tests/tests/authenticate_verifier_tests.rs
miker83z, line 1: LGTM! Even if something about the setup/creation of AuthenticateInfo might change later, I think these tests help to understand the logic behind the verifier (which is a component stable enough imho).

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/option.move
```diff
@@ -0,0 +1,156 @@
+module authenticate::option;
+
+use iota::auth_context::AuthContext;
+
+public struct Object has key, store {
+    id: iota::object::UID,
+}
+
+#[allow(unused_field)]
+public struct ObjectTemplated<T: key + store> has copy, drop, store {
```
theiari, line 13:
```suggestion
public struct ObjectTemplated<T: copy + drop+ store> has copy, drop, store {
```

I'm not so familiar with generic types, but _maybe_ there should be a `copy` ability in here , if we want `ObjectTemplated` to have `copy` aswell [Iota docs](https://docs.iota.org/developer/references/move/abilities#conditional-abilities-and-generic-types)(which'd make this struct pointless since it'd be equal to `NonObjectTemplated`).


The other way maybe is something like:
```suggestion
public struct ObjectTemplated<T: key + store> has store {
```

TheMrAI: > I'm not so familiar with generic types, but _maybe_ there should be a `copy` ability in here , if we want `ObjectTemplated` to have `copy` aswell [Iota docs](https://docs.iota.org/developer/references/move/abilities#conditional-abilities-and-generic-types)(which'd make this struct pointless since it'd be equal to `NonObjectTemplated`).
> 
> The other way maybe is something like:

Yes, you are right. `ObjectTemplate` `T` has to have the `copy` and `drop` ability, given the restrictions on the struct itself. You don't need to write them out though, as the compiler enforces the restriction as you wrote.
The important ability here is the `key`. This is also the main difference between `ObjectTemplated` and `NonObjectTemplated` test cases as the verifier will behave differently to them: [verify_pure_input_type](https://github.com/iotaledger/iota/blob/48ad690eb12609c525b41ffe875d1efa8e7f4f60/iota-execution/latest/iota-verifier/src/account_auth_verifier.rs#L217)
(We could write it out, but it isn't necessary and here you we get into ideologies, who likes to write more or less code.)

The second example wouldn't work as we are testing `std::option` which requires `copy, drop, store`.

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
```diff
@@ -0,0 +1,156 @@
+module authenticate::m;
```
miker83z:
```suggestion
module authenticate::m_vector;
```
Or something similar to fix #8525

crates/iota-e2e-tests/tests/authenticate_verifier_tests.rs
```diff
@@ -0,0 +1,295 @@
...
+    templated_non_object_mutable_ref,
+    templated_non_object_by_value,
+    object_by_value,
+    object_mutable_ref,
+    templated_object_by_value,
+    templated_object_mutable_ref
+});
+
+tests_in_module!(
+vector;
```
miker83z, line 204: This mismatches with the module name https://github.com/iotaledger/iota/pull/8539/files#r2391473098

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
miker83z: https://github.com/iotaledger/iota/pull/8539/files#r2391485029

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/object.move
```diff
@@ -0,0 +1,26 @@
...
+
+use iota::auth_context::AuthContext;
+
+// Object
+
+public struct Object has key, store {
+    id: iota::object::UID,
+}
+
+public fun immutable_ref(
```
miker83z, line 15: Since these test modules will become a sort of reference for understanding whether a function signature is correct or not for an authenticate fn, could you add a comment for each function indicating ‘//pass’ or ‘//fail’, so that it is sufficient to just check the Move code (and not look for the fn in the e2e test)?

miker83z (approved):
LGTM!

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/object.move
TheMrAI: Fixed in: c4d58f528ed17edeb931b8581b7dbd42633d97e5

Dkwcs (commented):
LGTM, however should we ensure that all new files have ?

```
```

Dkwcs (approved):

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/signature.move
```diff
@@ -0,0 +1,40 @@
+module authenticate::signature;
+
+use iota::auth_context::AuthContext;
+
+// PASS
+public fun minimally_viable_auth_function(_auth_ctx: &AuthContext, _ctx: &TxContext) {}
+
+// FAIL
+#[allow(unused_function)]
+fun has_to_be_public_auth_function(_auth_ctx: &AuthContext, _ctx: &TxContext) {}
```
valeriyr, line 13: It would be nice to add the following function test cases to cover all the visibility options:
- `entry`
- `entry public`
- `public(package)`
- `entry public(package)`
- `public(friend)`?
- `entry public(friend)`?

TheMrAI:
> LGTM, however should we ensure that all new files have ?
> 
> ```
> // Copyright (c) 2025 IOTA Stiftung
> // SPDX-License-Identifier: Apache-2.0
> ```

Fixed in: 86a3bfb363282504d9fd2a977824b7b7241f2386

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
```diff
@@ -0,0 +1,176 @@
...
+
+// PASS
+public fun template_object_immutable_reference<T: key>(
+    _objects: &vector<T>,
+    _auth_ctx: &AuthContext,
+    _ctx: &TxContext,
+) {}
+
+// FAIL
+public fun template_object_by_value<T: key + store>(
```
valeriyr, line 140:
```suggestion
public fun template_object_by_value<T: key>(
```
Should it be like this?

crates/iota-e2e-tests/tests/authenticate_verifier_tests.rs
```diff
@@ -0,0 +1,285 @@
...
+    () => {};
+    ($module_name: ident::$function_name: ident) => {
+        #[sim_test]
+        async fn $function_name() {
+            let env = TestEnvironment::new().await;
+            let (effects, _) = env
+                .create_auth_info_using("signature", "at_least_two_args")
+                .await
+                .unwrap();
+            assert!(effects.status().is_err());
```
valeriyr, line 121: For future improvements, it would be nice to check the error here.
Probably convert these tests to ones with the snapshots.
It is important not only to check that we fail, but also the error message that we provide to a user. It should contain a reason why a function can't be used as an authenticator.

valeriyr (approved):
I left several comments for consideration; however, it looks good to me overall :+1:

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
TheMrAI: Nope, tried it, public_share_object needs `store` as well.

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/signature.move
TheMrAI: Added tracking ticket: https://github.com/iotaledger/iota/issues/8721

crates/iota-e2e-tests/tests/authenticate_verifier_tests.rs
TheMrAI: Added tracking ticket: https://github.com/iotaledger/iota/issues/8721

PR #8541, opened 2025-09-12
fix(CI): update cargo deny

Thoralf-M:
# Description of change

Ignore RUSTSEC-2025-0056 because the adler crate is unmaintained and one of our sub dependencies we can't update ourselves


DaughterOfMars (commented):
Do we need external too?

Thoralf-M:
> Do we need external too?

Seems so

DaughterOfMars (approved):

miker83z (approved):

PR #8543, opened 2025-09-12
feat(docs): add an aa example using the rust sdk

### LINKED ISSUE #8538: [AA] Write some examples using the Rust SDK and test the alpha version
Depends on #8116 and #8238. The aim is to validate the implementation of those two.

miker83z:
# Description of change

This example shows how to create an abstracted account and send a TX through it using a Move-based Ed25519 authentication.

Here’s what it does step by step:
1. **Setup**
 **IMPORTANT**: this example needs to run a local network with a faucet manually (see instructions below for the exact command)
   * Connects to a local IOTA network using `IotaClientBuilder`.
   * Initializes an in-memory keystore and imports a test account from a fixed mnemonic.
   * Requests test tokens (faucet) for the sender.

3. **Create an IOTAccount**

   * Publishes the **IOTAccount Move package** (which uses an Account Abstraction logic) to the network.
   * Uses the published package to create a new `IOTAccount` shared object.
   * Requests faucet tokens for the new account.

4. **Build a Simple Transaction**

   * Prepares a sample PTB (`iota_account_simple_tx`) that calls `iota_account::add_field` with dummy values.
   * Uses the `IOTAccount` as the sender.

5. **Sign & Authenticate**

   * Signs the transaction data with the `IOTAccount` owner’s Ed25519 key.
   * Encodes the signature and wraps it in a `MoveAuthenticator` so it can be verified by the on-chain authenticator in Move.

6. **Execute Transaction**

   * Submits the signed transaction with the Move-based authenticator to the network.

Fixes #8538

- start a local network 
`cargo run --bin iota start --force-regenesis --with-faucet --faucet-amount 100000000000000 --committee-size 2 --epoch-duration-ms 3600000`
- `cd docs/examples/rust`
- `cargo run --example iotaccount`
- The explorer can be helpful (runs only with some browser configurations) 
https://explorer.iota.org/?network=http%3A%2F%2F127.0.0.1%3A9000


crates/iota-types/src/move_authenticator.rs
```diff
@@ -53,6 +53,19 @@ impl Hash for MoveAuthenticator {
 }
 
 impl MoveAuthenticator {
+    pub fn new_for_testing(
```
TheMrAI, line 56: Could we restrict this function into a test module scope?
```
#[cfg(test)]
mod test {
    ...
}
```

TheMrAI (approved):

valeriyr (approved):

Dkwcs (approved):
I’ve testes the flow on my machine locally, everything works well.

crates/iota-types/src/move_authenticator.rs
miker83z: It could be done, but it will need additional handling for the example to run.

PR #8552, opened 2025-09-12
feat(gRPC): Add transaction and read/write services for internal usage

### LINKED ISSUE #7724: gRPC: Replace all REST endpoints used in the tests with gRPC ones
We want to get rid of the REST API, so we also need to replace all occurrences in our tests.
In case we can already replace the REST endpoint with a JSON-RPC call, we should do that instead for now.

bingyanglin:
# Description of change

- Added transaction filters for client subscription
  - WASP team has provided the feedback that the transaction filter subscriber works as expected
- Added node service including `GetLatestCheckpoint`, `ExecuteTransaction`, and `GetObject`.
  - The `GetLatestCheckpoint` support get the latest checkpoint data or checkpoint summary
  - The `ExecuteTransaction` is fully aligned with the current JSON-RPC design in IOTA and SUI (for the request, see the function `execute_transaction_block` in `pub trait WriteApi`; for the response, see `IOTATransactionBlockResponse` and `SuiTransactionBlockResponse`, respectively.).
    -  For `iota-rest-api`, there is an additional field `EffectsFinality` in the response, but it is not used in our codebase, so we do not implement this field.
  - The `GetObject` is also mostly aligned with the current JSON-RPC design in IOTA and SUI (see the function `get_object` in the `pub trait ReadApi` for more details). The only difference is that I removed `display` field which was for UI rendering, and removed the `content` field which is the parsed/JSON representation of Move object or package content
- Implemented `WriteService` and `ReadService` to ease of future extension of more API calls.
- Removed the `GrpcStateReader` and `RestStateReaderAdapter` in `iota-grpc-api` crate, and leveraged the existing REST API infrastructure (including the `RestStateReader` and `RestReadStore` four our use cases). The plan is after REST API is deprecated, we can directly rename them to be `Grpc*`, so as to avoid repetitive work in the future.
- Enhanced `RestStateReader` and `RestReadStore` to access `AuthorityState` and `AuthorityPerEpochStore` for usage in `ReadGrpcService` and `WriteGrpcServce`. This is required to return more information like what JSON-RPC does.
- Added test utils for cleaner code.
- **NOTE THAT the document will be enhanced/updated after more APIs are added and the design pattern is finalized, otherwise we might need to modify the document back and forth.** (see #8695)

fixes #7724 

### Release Notes

- [x] Nodes (Validators and Full nodes): For gPRC, added transaction filters for client subscription, and added 3 APIs including `GetLatestCheckpoint`, `ExecuteTransaction`, and `GetObject`.


crates/iota-grpc-api/proto/common.proto
```diff
@@ -15,7 +15,54 @@ message Address {
...
+  Address package_id = 1;         // Package ID
+  optional string module = 2;     // Module name (optional)
+  optional string function = 3;   // Function name (optional)
+}
+
+// Object reference (ID + version + digest)
+message ObjectRef {
+  Address object_id = 1;
+  uint64 version = 2;
+  Digest digest = 3;
```
Thoralf-M: Is this really interesting with version and digest? Just the object id alone seems more useful to me

bingyanglin: Thanks for pointing this out. Removed it at [8d40b0b](https://github.com/iotaledger/iota/pull/8552/commits/8d40b0b22241b4d3b4993300fdce471f353b01f3)

This was used to align with the following type 
```rust
pub type ObjectRef = (ObjectID, SequenceNumber, ObjectDigest);
```
defined in `iota-types`, which is used by `iota-json-rpc`. But after refactoring, for the related response types of gRPC we just use json_data (string), so it can be safely removed now. And for the `object_id` we can directly use `Address object_id` without additional wrapper.

Thoralf-M: The linked commit just changes something with `TransactionEffectsV1`?

bingyanglin: Uh sorry missed the file, updated. Please check, thanks!

crates/iota-types/src/storage/read_store.rs
```diff
@@ -814,6 +814,9 @@ pub trait RestStateReader: ObjectStore + ReadStore + Send + Sync {
 
     // Get a handle to an instance of the RpcIndexes
     fn indexes(&self) -> Option<&dyn RestIndexes>;
+
+    /// Enable downcasting to concrete types for enhanced functionality
+    fn as_any(&self) -> &dyn std::any::Any;
```
semenov-vladyslav: A weird method for a trait. Should `RestStateReader` be extended with `pub fn authority_state(&self) -> Arc<AuthorityState>` and `pub fn load_epoch_store_one_call_per_task(&self) -> Arc<AuthorityPerEpochStore>` methods instead?

crates/iota-core/src/storage.rs
```diff
@@ -380,6 +380,16 @@ impl RestReadStore {
...
+
+    /// Get access to the underlying AuthorityState
+    pub fn authority_state(&self) -> &Arc<AuthorityState> {
+        &self.state
+    }
+
+    /// Load epoch store for transaction processing
+    pub fn load_epoch_store_one_call_per_task(&self) -> Arc<AuthorityPerEpochStore> {
+        self.state.load_epoch_store_one_call_per_task().clone()
+    }
```
semenov-vladyslav, line 395: Why not move these methods and make them a part of `RestStateReader` trait?

crates/iota-core/src/subscription_handler.rs
```diff
@@ -114,24 +112,17 @@ impl SubscriptionHandler {
         Ok(())
     }
 
-    pub fn subscribe_events(&self, filter: EventFilter) -> impl Stream<Item = IotaEvent> {
-        self.event_streamer.subscribe(filter)
+    pub fn subscribe_events(
+        &self,
+        filter: EventFilter,
+    ) -> Box<dyn futures::Stream<Item = IotaEvent> + Send + Unpin> {
```
semenov-vladyslav: Should it be `BoxStream<'_, IotaEvent>`? Or is `Unpin` important here?

crates/iota-core/src/subscription_handler.rs
```diff
@@ -114,24 +112,17 @@ impl SubscriptionHandler {
...
+        Box::new(Box::pin(self.event_streamer.subscribe(filter)))
```
semenov-vladyslav: Double `Box`? Doesn't look right. With `BoxStream` we can just use one `Box::pin`.

crates/iota-core/src/subscription_handler.rs
```diff
@@ -114,24 +112,17 @@ impl SubscriptionHandler {
...
-
-// Implement EventSubscriber trait for gRPC integration
-impl EventSubscriber for SubscriptionHandler {
-    fn subscribe_events(
-        &self,
-        filter: EventFilter,
-    ) -> Box<dyn futures::Stream<Item = IotaEvent> + Send + Unpin> {
-        Box::new(Box::pin(self.event_streamer.subscribe(filter)))
+    ) -> Box<dyn futures::Stream<Item = IotaTransactionBlockEffects> + Send + Unpin> {
+        Box::new(Box::pin(self.transaction_streamer.subscribe(filter)))
```
semenov-vladyslav: Same here.

crates/iota-grpc-server/src/read_service.rs
```diff
@@ -0,0 +1,237 @@
...
+        });
+
+        let grpc_reader = self.grpc_reader.clone();
+        let object_read = spawn_monitored_task!(async move {
+            grpc_reader.get_object_read(&object_id).map_err(|e| {
+                warn!(?object_id, "Failed to get object: {:?}", e);
+                Status::internal(format!("Failed to read object from storage: {e}"))
+            })
+        })
+        .await
```
semenov-vladyslav, line 50: Why does it have to be a task? `grpc_reader.get_object_read` is not even an async fn.

crates/iota-grpc-api/src/checkpoint_service.rs
```diff
@@ -75,7 +75,7 @@ impl CheckpointGrpcService {
 #[tonic::async_trait]
 impl CheckpointService for CheckpointGrpcService {
     type StreamCheckpointsStream =
-        Pin<Box<dyn futures::Stream<Item = Result<crate::checkpoint::Checkpoint, Status>> + Send>>;
+        Pin<Box<dyn futures::Stream<Item = Result<Checkpoint, Status>> + Send>>;
```
semenov-vladyslav:
```suggestion
        futures::BoxStream<Result<Checkpoint, Status>>;
```

crates/iota-grpc-api/src/checkpoint_service.rs
```diff
@@ -130,13 +130,63 @@ impl CheckpointService for CheckpointGrpcService {
...
         Ok(Response::new(CheckpointSequenceNumberResponse {
             sequence_number,
         }))
     }
+
+    async fn get_latest_checkpoint(
+        &self,
+        request: Request<GetLatestCheckpointRequest>,
+    ) -> Result<Response<Checkpoint>, Status> {
+        debug!("get_latest_checkpoint called");
```
semenov-vladyslav:
```suggestion
```

crates/iota-grpc-api/src/read_service.rs
```diff
@@ -0,0 +1,237 @@
...
+// service must implement.
+#[tonic::async_trait]
+impl ReadService for ReadGrpcService {
+    #[instrument(skip(self))]
+    async fn get_object(
+        &self,
+        request: Request<GetObjectRequest>,
+    ) -> Result<Response<GetObjectResponse>, Status> {
+        let req = request.into_inner();
+        debug!("get_object called");
```
semenov-vladyslav:
```suggestion
```

crates/iota-grpc-api/src/transaction_service.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+// The `TransactionService` is the auto-generated trait from the protobuf
+// definition. It's generated by tonic/protobuf and defines the interface that
+// any gRPC transaction service must implement.
+#[tonic::async_trait]
+impl TransactionService for TransactionGrpcService {
+    type StreamTransactionsStream =
+        std::pin::Pin<Box<dyn futures::Stream<Item = Result<Transaction, Status>> + Send>>;
```
semenov-vladyslav:
```suggestion
        futures::BoxStream<Result<Transaction, Status>>;
```

crates/iota-grpc-api/src/types.rs
```diff
@@ -204,107 +190,138 @@ impl BcsData {
...
-    fn get_checkpoint_data(&self, seq: u64) -> Option<CheckpointData> {
-        let summary = self.inner.get_checkpoint_by_sequence_number(seq)?;
-        let contents = self.inner.get_checkpoint_contents_by_sequence_number(seq)?;
-        Some(self.inner.get_checkpoint_data(summary, contents))
+    /// Load epoch store for transaction processing with graceful fallback
+    pub fn load_epoch_store_one_call_per_task(&self) -> Option<Arc<AuthorityPerEpochStore>> {
+        // Downcast to RestReadStore for enhanced functionality
+        self.state_reader
+            .as_any()
+            .downcast_ref::<RestReadStore>()
```
semenov-vladyslav: This downcast suggests that the only type for the dyn object `state_reader` is `RestReadStore`, all other types are "fallback" or mock.

semenov-vladyslav (commented):
Overall looks good. There are a few nits to improve.

piotrm50 (approved):
approving files owned by `consensus`

crates/iota-types/src/storage/read_store.rs
bingyanglin: Unfortunately we cannot do this, because the `AuthorityState` and `AuthorityPerEpochStore` are defined in `iota-core` crate, and `iota-core` depends on `iota-types` (where the `RestStateReader` is defined), so we cannot use the `iota-core` types in `iota-type` crate, which will introduce circular dependency problem.

To avoid confusing, I changed the `as_any` to be `authority_state_any`, and added more comets in https://github.com/iotaledger/iota/pull/8552/commits/65759d67703b221c24156a8ec9c90b38b7be5b40 to make it clear, hopefully it helps.

crates/iota-core/src/storage.rs
bingyanglin: Same reason of the previous comment. Refactored in [d19ed18](https://github.com/iotaledger/iota/pull/8552/commits/d19ed18505f96c37053b0902220f04ae3b857c4b).

crates/iota-grpc-api/src/types.rs
bingyanglin: Due to the reason in the first comment of the circular dependency issue, I am now using the `authority_state_any()` to explicitly what we are going to downcast in https://github.com/iotaledger/iota/pull/8552/commits/65759d67703b221c24156a8ec9c90b38b7be5b40. This might be used for simulacrum as well in the future. Hopefully this is clean and clear.

crates/iota-grpc-api/src/transaction_service.rs
bingyanglin: Thanks for pointing this out, updated in https://github.com/iotaledger/iota/pull/8552/commits/21bd6127396a78df8b59e78051b9bfc471b206b4.

crates/iota-grpc-api/src/checkpoint_service.rs
bingyanglin: Thanks for pointing this out, updated in https://github.com/iotaledger/iota/pull/8552/commits/21bd6127396a78df8b59e78051b9bfc471b206b4.

crates/iota-core/src/subscription_handler.rs
bingyanglin: Thanks for pointing this out, updated in https://github.com/iotaledger/iota/pull/8552/commits/21bd6127396a78df8b59e78051b9bfc471b206b4.

crates/iota-core/src/subscription_handler.rs
bingyanglin: Thanks for pointing this out, updated in https://github.com/iotaledger/iota/pull/8552/commits/21bd6127396a78df8b59e78051b9bfc471b206b4.

crates/iota-core/src/subscription_handler.rs
bingyanglin: We should use `BoxStream` so the code is cleaner. Thanks for pointing this out, updated in https://github.com/iotaledger/iota/pull/8552/commits/21bd6127396a78df8b59e78051b9bfc471b206b4.

crates/iota-grpc-server/src/read_service.rs
bingyanglin: It's for observability purpose. We wrap it in `spawn_monitored_task!` to have structured tracing/logging, and this is the same pattern as the `get_object` in `iota-json-rpc` crate (see file `crates/iota-json-rpc/src/read_api.rs`. Do you think it makes sense? In the future when we have more powerful tracing/monitoring system maybe we can enhance this.

crates/iota-core/src/storage.rs
semenov-vladyslav: This method is somewhat dangerous, see the comment to `AuthorityState::load_epoch_store_one_call_per_task`.

crates/iota-grpc-api/src/write_service.rs
```diff
@@ -0,0 +1,577 @@
...
+        sender: IotaAddress,
+        core_transaction_data: Option<TransactionData>,
+    ) -> Result<Response<ExecuteTransactionResponse>, Status> {
+        // Store core events for BCS serialization before converting to JSON type
+        let core_events = response.events.clone();
+
+        let events = if opts.show_events {
+            tracing::trace!("Resolving events");
+            if let (Some(epoch_store), Some(authority_state)) = (
+                self.grpc_reader.load_epoch_store_one_call_per_task(),
```
semenov-vladyslav: There's a reason for such long name. Epoch store may change during task execution.

semenov-vladyslav (changes requested):
The use of `load_epoch_store_one_call_per_task()` doesn't look safe. `<WriteGrpcService as WriteService>::execute_transaction` loads epoch store twice.

crates/iota-grpc-api/src/write_service.rs
bingyanglin: I agree, just this is the way to get the `epoch_store`, and both https://github.com/iotaledger/iota/blob/develop/crates/iota-json-rpc/src/transaction_execution_api.rs#L198 and https://github.com/MystenLabs/sui/blob/main/crates/sui-json-rpc/src/transaction_execution_api.rs#L187 do the same thing. I guess this is already the last part of the `execute_transaction` which mostly does the type casting and should be quick, epoch store change during this period maybe a rare corner case. If the user encountered the corner case, the easiest way is to just retry in the client side.

crates/iota-core/src/storage.rs
bingyanglin: Yeah we use the same implementation as what `json-rpc-api` and the upstream does now, where we need the this store in many places.

bingyanglin: Added code comments in [c2c1b59](https://github.com/iotaledger/iota/pull/8552/commits/c2c1b59b9cf334122c8d7eb542e6a36a59637070) to make it clearer.

crates/iota-grpc-api/src/write_service.rs
bingyanglin: As our discussion, updated the code behavior to load the `epoch_store` it in the beginning and pass the same arced epoch store to the functions in the `execute_transaction` which use it in [c2c1b59](https://github.com/iotaledger/iota/pull/8552/commits/c2c1b59b9cf334122c8d7eb542e6a36a59637070).

semenov-vladyslav (approved):

crates/iota-grpc-types/proto/iota/grpc/v0/write.proto
kodemartin, line 1: This implementation of  `execute_transaction` complies with the JSON-RPC counterpart, but it doesn't match the response contents of the REST-API.

Appart from the finality returned by the REST-API, a major difference is that it returns also the full contents of input and output objects.

This is important to us because the write API of the indexer relies on these data.

In addition, isn't the information about local execution obsolete by now?

kodemartin: IMHO the write API should match the REST API counterpart, and not the JSON-RPC. Or at least provide the option to get the same response data.

bingyanglin: Hi @kodemartin , thanks for porting this out!
- We are refining the gRPC API design so it will be low-level and provide BCS data directly in most cases, and are preparing a draft PR which will contain the proto files so easier to discuss.
  - We will return the full contents of input and output objects for `execute_transaction` in the updated design. 
- The local execution obsolete PR was just merged so we will remove the field yes.

muXxer:
This can be closed because it was done in #9335, #9328, #9314 and #9233

PR #8553, opened 2025-09-12
fix: loading an authenticator info in `check_move_account`

### LINKED ISSUE #8545: [AA][core,execution] Fix AA flow in validator
- [x] Issue 1: 
  During the execution of a move authentication (before spawning the VM) this error is raised: "Index store not available on this Fullnode". It seems to come from:
  `let authenticator_id = self.get_dynamic_field_object_id` in `check_move_account`
  The error should happen because that part of the code is not executed by a full node, so probably the solution could be to fetch the dynamic object using another way (i.e., not using the AuthorityState `indexes`).

- [x] Issue 2: 
  ```
  ErrorObject { code: ServerError(-32002), message: "Transaction execution failed due to issues with transaction inputs, please review the errors and try again:\n- Failed to execute the Move authenticator, reason: \"ExecutionError: ExecutionError { inner: ExecutionErrorInner { kind: CommandArgumentError { arg_idx: 2, kind: InvalidUsageOfPureArg }, source: Some(\\\"Non-primitive argument at index 2. If it is an object, it must be populated by an object\\\"), command: Some(0) } }\".", data: None }
  ```
  It is referring to the `AuthContext`, it checks this in `check_param_type`.

- [x] Issue 3: 
  ```
  ErrorObject { code: ServerError(-32002), message: "Transaction execution failed due to issues with transaction inputs, please review the errors and try again:\n- Failed to execute the Move authenticator, reason: \"ExecutionError: ExecutionError { inner: ExecutionErrorInner { kind: VMVerificationOrDeserializationError, source: Some(VMError { major_status: FAILED_TO_DESERIALIZE_ARGUMENT, sub_status: None, message: None, exec_state: None, location: Undefined, indices: [], offsets: [] }), command: Some(0) } }\".", data: None }
  ```
   Still referring to the `AuthContext`. In this case we need a special serialization because on the Move side we use `TypeName` instead of `InputType`.

valeriyr:
# Description of change

Fixed `AuthenticatorInfoV1` dynamic field ID generation when validating an account object.

fixes #8545


TheMrAI (approved):
Looks good.

crates/iota-core/src/authority.rs
```diff
@@ -5287,33 +5287,41 @@ impl AuthorityState {
...
-                    account_object_id: auth_account_object_id,
-                    account_object_version: auth_account_object_seq_number,
-                }
-                .into())
-            }
+        if let Some(authenticator_info_field_obj) = authenticator_info_field {
+            let field_move_object = authenticator_info_field_obj
+                .data
+                .try_as_move()
+                .expect("dynamic field should never have be a package object");
```
Thoralf-M: Something went wrong here, should it be like this?
```suggestion
                .expect("dynamic field should never be a package object");
```

valeriyr: fixed in beaa7a6b00

miker83z (approved):

PR #8555, opened 2025-09-12
feat(indexer): add streaming support for events and transactions

### LINKED ISSUE #8315: [indexer]: add streaming for transactions and events
## Description

We would like to support streaming on APIs built on top of the Indexer database.

To this end we should

- [x] analyze the feasibility and requirements for implementing the feature
- [x] proceed with the implementation in `iota-indexer`

sergiupopescu199:
# Description of change

This PR is the first iteration attempt to provide streaming support for the `iota-indexer` as a library componente. Currently it supports streaming of `StoredEvents` and `StoredTransactions`. It also supports filters.

fixes #8315 

created a separated binary which used the `iota-indexer` as a library, used `testnet` network as source of checkpoints for synchronization, subscribed to incoming events, applied different filter to assert that they were respected.

### Release Notes
- [x] Indexer: Add streaming support for `events` and `transactions`.

### DIFF
--- a/crates/iota-indexer/src/stream.rs
+++ b/crates/iota-indexer/src/stream.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{BTreeMap, VecDeque},
+    collections::BTreeMap,
     str::FromStr,
     sync::{Arc, Mutex},
     time::Instant,
@@ -17,7 +17,7 @@ use iota_types::{
     transaction::TransactionDataAPI,
 };
 use serde::{Deserialize, Serialize};
-use tokio::sync::mpsc::{self, error::TrySendError};
+use tokio::sync::mpsc::{self};
 use tokio_postgres::{AsyncMessage, Client, Config, Connection, NoTls, Socket, tls::NoTlsStream};
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, error, warn};
@@ -31,18 +31,22 @@ use crate::{
     },
 };
 
+/// Buffer size used in the mpsc bounded channels.
 const BUFFER_SIZE: usize = 1000;
+/// Postgres Notify channel name.
 const CHANNEL_NAME: &str = "checkpoint_committed";
 
+/// Active subscribers held in memory.
 type Subscribers<T, F> = Arc<Mutex<BTreeMap<String, Subscriber<T, F>>>>;
 
+/// A single subscriber to which we can send data based on provided filters.
 struct Subscriber<T, F> {
-    sender: mpsc::Sender<T>,
-    lag_buffer: VecDeque<T>,
+    sender: mpsc::Sender<StreamEvent<T>>,
     filter: F,
 }
 
-#[derive(Clone, Debug, Default)]
+/// Filter returned [`StoredEvent`] form the stream.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StreamEventFilter {
     #[default]
     All,
@@ -67,13 +71,14 @@ impl Filter<StoredEvent> for StreamEventFilter {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Module {
     name: String,
     function: Option<String>,
 }
 
-#[derive(Clone, Debug, Default)]
+/// Filter returned [`StoredTransaction`] form the stream.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StreamTransactionFilter {
     #[default]
     All,
@@ -115,42 +120,69 @@ impl Filter<StoredTransaction> for StreamTransactionFilter {
     }
 }
 
+/// Represents a notification about a checkpoint commit in Indexer Database, it
+/// acts as a trigger to request transaction related to that particular
+/// checkpoint.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 struct CheckpointCommitNotification {
+    /// Committed checkpoint sequence number.
     checkpoint_sequence_number: i64,
+    /// the minimum transaction sequence number.
     min_tx_sequence_number: i64,
+    /// the maximum transaction sequence number.
     max_tx_sequence_number: i64,
+}
+
+/// Represents events that can be sent to stream subscribers in a broadcasting
+/// system.
+///
+/// This enum encapsulates the different types of messages that subscribers can
+/// receive when listening to a stream. It handles both normal message delivery
+/// and backpressure scenarios where slow subscribers need to be managed.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum StreamEvent<T> {
+    /// A regular message delivered to an active subscriber.
+    ///
+    /// This variant indicates normal operation where the subscriber is keeping
+    /// up with the message flow and can continue receiving subsequent
+    /// messages. The wrapped data represents the actual payload being
+    /// streamed.
+    Message(T),
+    /// The final message sent to a subscriber that is lagging behind.
+    ///
+    /// This variant is sent when the server detects that a subscriber cannot
+    /// process messages fast enough to keep up with the stream. After sending
+    /// this message, the subscriber will be automatically removed from the
+    /// broadcasting system to prevent memory buildup and maintain overall
+    /// system performance.
+    ///
+    /// The wrapped data represents the last message the subscriber will
+    /// receive, allowing for graceful cleanup or final processing before
+    /// disconnection.
+    ClosingDueToLag(T),
 }
 
 /// A lightweight in-memory fan-out broadcaster that distributes data to
 /// multiple filtered subscribers.
 ///
-/// The `Broadcaster` provides a publish-subscribe pattern where a single data
-/// producer can efficiently distribute items to multiple consumers, each with
-/// their own filtering criteria.
+/// It provides a publish-subscribe pattern where a single data producer can
+/// efficiently distribute items to multiple consumers, each with their own
+/// filtering criteria.
 ///
 /// # Backpressure Handling
+/// The "slow receiver" problem is handled through capacity-based monitoring:
 ///
-/// This broadcaster handles the "slow receiver" problem similarly to tokio's
-/// broadcast channels. When subscribers cannot keep up with the message rate,
-/// the system maintains bounded memory usage while gracefully handling lag.
-///
-/// ## Channel Send Behavior
-/// - **Success**: Message is delivered immediately to the subscriber's channel
-/// - **Channel Full**: Message is added to the subscriber's local lag buffer
-/// - **Channel Closed**: Subscriber is automatically removed from the
-///   subscribers list
-///
-/// ## Lag Buffer
-/// Each subscriber has a local buffer that retains messages when their channel
-/// is full:
-/// - When a subscriber's channel is full, messages are queued in their lag
-///   buffer
-/// - The lag buffer has a hard upper bound on capacity (`BUFFER_SIZE`)
-/// - When the buffer is at capacity and a new message arrives, the **oldest**
-///   message is released to free up space for the new message
-/// - On the next broadcast iteration, buffered messages are sent first
-///   (oldest-to-newest) before new messages
+/// - **Normal Operation**: When a subscriber's channel has capacity > 1,
+///   messages are sent as [`StreamEvent::Message`] allowing continued
+///   subscription
+/// - **Backpressure Detection**: When capacity drops to 1 (channel nearly
+///   full), the subscriber is marked as lagging behind the message production
+///   rate
+/// - **Graceful Termination**: Lagging subscribers receive their final message
+///   wrapped in [`StreamEvent::ClosingDueToLag`] and are automatically removed
+///   from the subscriber list
+/// - **System Protection**: This prevents unbounded memory growth and maintains
+///   overall broadcaster performance by avoiding blocking on slow consumers
 ///
 /// # Filtering
 ///
@@ -169,9 +201,10 @@ where
     /// Creates a new broadcaster that consumes items from the provided
     /// receiver.
     ///
-    /// This method spawns a background task that continuously reads from `rx`
-    /// and distributes items to all registered subscribers. The broadcaster
-    /// begins processing immediately upon creation.
+    /// This method spawns a background task that continuously reads from
+    /// [`mpsc::Receiver`] and distributes items to all registered
+    /// subscribers. The broadcaster begins processing immediately upon
+    /// creation.
     fn new(mut rx: mpsc::Receiver<T>) -> Self {
         let streamer = Self {
             subscribers: Default::default(),
@@ -197,28 +230,32 @@ where
             });
 
             for (id, subscriber) in subscribers_snapshot.iter_mut() {
+                if subscriber.sender.is_closed() {
+                    warn!(
+                        subscription_id = id,
+                        "Subscriber closed connection, removing"
+                    );
+                    to_remove.push(id.clone());
+                    continue;
+                }
+
                 if !(subscriber.filter.matches(&data)) {
                     continue;
                 }
-                Self::push_bounded(&mut subscriber.lag_buffer, data.clone());
-                while let Some(data) = subscriber.lag_buffer.pop_front() {
-                    match subscriber.sender.try_send(data) {
-                        Ok(_) => {
-                            debug!(subscription_id = id, "Streaming data to subscriber.");
-                        }
-                        Err(TrySendError::Full(returned_data)) => {
-                            warn!(subscription_id = id, "Lagging behind");
-                            Self::push_bounded(&mut subscriber.lag_buffer, returned_data);
-                        }
-                        Err(TrySendError::Closed(_)) => {
-                            warn!(
-                                subscription_id = id,
-                                "Sender half dropped, removing subscriber"
-                            );
-                            to_remove.push(id.clone());
-                        }
-                    }
-                }
+
+                let data = if subscriber.sender.capacity() > 1 {
+                    StreamEvent::Message(data.clone())
+                } else {
+                    warn!(
+                        subscription_id = id,
+                        "Lagging behind, sending last message and removing"
+                    );
+                    to_remove.push(id.clone());
+                    StreamEvent::ClosingDueToLag(data.clone())
+                };
+
+                // since we checked the capacity we can ignore the result
+                let _ = subscriber.sender.try_send(data);
             }
             to_remove
         };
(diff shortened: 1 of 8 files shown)


crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+    indexer_reader::IndexerReader,
+    models::{
+        events::StoredEvent,
+        transactions::{StoredTransaction, stored_events_to_events},
+    },
+    types::IndexerResult,
+};
+
+const BUFFER_SIZE: usize = 1000;
+const MAX_SUBSCRIBERS: usize = 100;
```
kodemartin: We can skip the upper limit for the moment. If the need arises we can implement some batching logic while fanning out the data.

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+/// their own filtering criteria.
+///
+/// # Backpressure Handling
+///
+/// The broadcaster implements automatic subscriber pruning for backpressure
+/// management:
+/// - Each subscriber has a bounded channel with `BUFFER_SIZE` capacity
+/// - Maximum of `MAX_SUBSCRIBERS` concurrent subscribers allowed
+/// - When a subscriber's channel is full, `try_send` fails immediately
+/// - Failed subscribers are automatically removed from the subscriber list
```
kodemartin: Let's mimic https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging instead of removing

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,516 @@
...
+        });
+        streamer
+    }
+
+    async fn send_to_all_subscribers(subscribers: Subscribers<T, F>, data: T) {
+        let to_remove = {
+            let mut to_remove = vec![];
+            let subscribers_snapshot = subscribers
+                .read()
+                .expect("failed to acquire read subscribers lock");
```
kodemartin: Why `expect`? You have calls to `write` in other parts of the code.

crates/iota-indexer/src/stream.rs
kodemartin, line 1: Please revisit visibility. Keep `pub` only whatever necessary.

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9fa1030e69afa7238fc46945cbe7d33fe5e809ef

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9fa1030e69afa7238fc46945cbe7d33fe5e809ef

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9fa1030e69afa7238fc46945cbe7d33fe5e809ef

to mimic the Ring buffer behavior of the broadcast channel I've added a new data structure, due to this since I always needed write access I replaced the `RwLock` with a `Mutex`

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9fa1030e69afa7238fc46945cbe7d33fe5e809ef

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+}
+
+/// Represents events that can be sent to stream subscribers in a broadcasting
+/// system.
+///
+/// This enum encapsulates the different types of messages that subscribers can
+/// receive when listening to a stream. It handles both normal message delivery
+/// and backpressure scenarios where slow subscribers need to be managed.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum StreamEvent<T> {
```
kodemartin: The name "event" is used mostly to refer to transaction events, which makes it confusing to use it for the stream data. Another iteration on the name is thus warranted.

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+    },
+};
+
+/// Buffer size used in the mpsc bounded channels.
+const BUFFER_SIZE: usize = 1000;
+/// Postgres Notify channel name.
+const CHANNEL_NAME: &str = "checkpoint_committed";
+
+/// Active subscribers held in memory.
+type Subscribers<T, F> = Arc<Mutex<BTreeMap<String, Subscriber<T, F>>>>;
```
kodemartin: Why using `BTreeMap`? Do we care about order? `ObjectID::random()` for new keys implies otherwise.

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+    /// ```
+    fn subscribe(&self, filter: F) -> impl Stream<Item = StreamEvent<T>> {
+        let mut subscribers = self.subscribers.lock().unwrap_or_else(|poisoned| {
+            error!("Subscribers mutex poisoned, recovering...");
+            self.subscribers.clear_poison();
+            poisoned.into_inner()
+        });
+        let (tx, rx) = mpsc::channel::<StreamEvent<T>>(BUFFER_SIZE);
+        subscribers.insert(
+            ObjectID::random().to_string(),
```
kodemartin: Using `ObjectID` for subscriber ids seems off, despite being practical. You can maintain reuseability by implementing a new-type `SubsriberId(ObjectID)` with a `new` method that wraps `ObjectID::random`. This way we will have the flexibility to refactor `new` at will in the future to use a lighter than 32bytes id. What do you think?

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+                            break;
+                        }
+                    }
+                    Err(e) => {
+                        error!("connection error: {e}");
+                        break;
+                    }
+                }
+            }
+        };
```
kodemartin: Why don't you consume directly the `stream` below 507 and use an `mpsc` channel to synchronize?

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+        mut connection: Connection<Socket, NoTlsStream>,
+        indexer_reader: IndexerReader,
+        event_tx: mpsc::Sender<StoredEvent>,
+        transaction_tx: mpsc::Sender<StoredTransaction>,
+    ) -> Result<(), IndexerError> {
+        // Create a channel for receiving messages
+        let (tx, rx) = mpsc::channel(BUFFER_SIZE);
+
+        // Create a stream from the connection that forwards messages to the channel
+        let stream = stream::poll_fn(move |cx| connection.poll_message(cx));
```
kodemartin: Nit:

```rust
        let mut stream = stream::poll_fn(move |cx| connection.poll_message(cx));
```

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+        // Listen for notifications on a specific channel
+        client
+            .execute(&format!("LISTEN {CHANNEL_NAME};"), &[])
+            .await
+            .context("failed to listen to channel notifications")?;
+
+        let mut notification_stream = ReceiverStream::new(rx).ready_chunks(BUFFER_SIZE);
+
+        while let Some(messages) = notification_stream.next().await {
+            if let Some((first, last)) = Self::first_and_last_checkpoint_notifications(&messages) {
```
kodemartin: Is there a guarantee that `first` and `last` correspond to the min and max checkpoints sequence numbers in the sequence?

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,609 @@
...
+            match msg {
+                AsyncMessage::Notification(n) => {
+                    match serde_json::from_str::<CheckpointCommitNotification>(n.payload()) {
+                        Ok(parsed) => {
+                            first.get_or_insert(parsed.clone());
+                            last = Some(parsed);
+                        }
+                        Err(e) => {
+                            error!("failed parsing checkpoint notification: {e}");
+                        }
```
kodemartin: This assumes that all notifications are checkpoint notifications.

kodemartin: Nit: Seems a bit verbose. Why not `messages.filter_map` to retain only `Notiification` messages that are parsed successfully into `CheckpointCommitNotification` and then let the caller take `next()` as the first and `Iterator::last` as the last notification?

Then you could split this method into `filter_checkpoint_notifications` with no other side effects for `Notice` or different types of notifications, and delegate processing of the filter to the caller.

kodemartin (commented):
Looks good, but I left a few more comments and suggestions to discuss.

crates/iota-indexer/src/stream.rs
sergiupopescu199: In the checkpoint committer we receive checkpoint in order, this is guaranteed by the `iota-data-ingestion-core`, checking how `IndexedCheckpoint` is created I see this logic for calculating the `min` and `max` transactions sequence number. 

It does follow correctly the transaction sequence numbers contained in the checkpoint.

```rust
let tx_digests = contents.iter().map(|t| t.transaction).collect::<Vec<_>>();
let max_tx_sequence_number = checkpoint.network_total_transactions - 1;
// NOTE: + 1u64 first to avoid subtraction with overflow
let min_tx_sequence_number = max_tx_sequence_number + 1u64 - tx_digests.len() as u64;
```

source: https://github.com/iotaledger/iota/blob/develop/crates/iota-indexer/src/types.rs#L58

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/8933896ec099295b01179c46735258a42044f347

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    async fn process_checkpoint_notifications(
+        mut connection: Connection<Socket, NoTlsStream>,
+        indexer_reader: IndexerReader,
+        event_tx: broadcast::Sender<StoredEvent>,
+        transaction_tx: broadcast::Sender<StoredTransaction>,
+    ) -> Result<(), IndexerError> {
+        // create a stream from the connection that forwards messages to the channel.
+        let mut stream = stream::poll_fn(move |cx| connection.poll_message(cx)).ready_chunks(10);
```
kodemartin: Let's make the chunk size a const parameter, instead of hardcoding. E.g.
```rust
        let mut stream = stream::poll_fn(move |cx| connection.poll_message(cx)).ready_chunks(CHECKPOINT_CHUNK_SIZE);
```

kodemartin (approved):
Looks good with a last minor suggestion.

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/745088e4a385cb3dbe196449a80ee04cab52037c

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Module {
+    pub name: String,
+    pub function: Option<String>,
+}
```
tomxey, line 91: Feels a bit strange that function name is part of module

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+impl IndexerStreamer {
+    /// Creates a new `IndexerStreamer` instance.
+    ///
+    /// This method establishes a connection to PostgreSQL, sets up the
+    /// notification listener, and spawns the background task that processes
+    /// checkpoint notifications.
+    pub async fn new(db_url: &str, indexer_reader: IndexerReader) -> Result<Self, IndexerError> {
+        let (client, connection) = Config::from_str(db_url)
+            .map_err(|e| IndexerError::Generic(format!("failed to parse Postgresdb url: {e}")))?
+            .connect(NoTls)
```
tomxey, line 189: Does it mean we are using unencrypted connection?

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// ```
+    pub fn subscribe_events(
+        &self,
+    ) -> impl Stream<Item = Result<StoredEvent, BroadcastStreamRecvError>> {
+        BroadcastStream::new(self.event_tx.subscribe())
+    }
+
+    /// Subscribe to a stream of [`StoredTransaction`].
+    ///
+    /// By default all events are received, it's possible to filter on client
```
tomxey: nit: did you mean: all *transactions* are received?

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+        transaction_tx: &broadcast::Sender<StoredTransaction>,
+    ) -> Result<(), IndexerError> {
+        let instant = Instant::now();
+        let transactions: Vec<StoredTransaction> = indexer_reader
+            .spawn_blocking(move |this| {
+                this.multi_get_transactions_by_sequence_numbers_range(start, end)
+            })
+            .await?;
+
+        debug!(
```
tomxey, line 358: Have you considered to introduce some prometheus metrics, to easier monitor usage (number of subscribers) and performance (query duration, tx/s throughput, lag? or whatever makes sense to check if system is not choking)

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+        messages.into_iter().filter_map(|msg_result| {
+            match msg_result {
+                Ok(AsyncMessage::Notification(n)) => {
+                    match serde_json::from_str::<CheckpointCommitNotification>(n.payload()) {
+                        Ok(notification) => Some(Ok(notification)),
+                        Err(_) => None,
+                    }
+                }
+                Ok(_) => None, // Not a notification message, skip
+                Err(e) => Some(Err(IndexerError::Generic(format!(
```
tomxey: Nit, maybe IndexerError::PostgresRead would fit better here

crates/iota-indexer/src/stream.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// [`CheckpointCommitNotification`] from PostgreSQL messages.
+    fn filter_checkpoint_notifications(
+        messages: Vec<Result<AsyncMessage, tokio_postgres::Error>>,
+    ) -> impl Iterator<Item = Result<CheckpointCommitNotification, IndexerError>> {
+        messages.into_iter().filter_map(|msg_result| {
+            match msg_result {
+                Ok(AsyncMessage::Notification(n)) => {
+                    match serde_json::from_str::<CheckpointCommitNotification>(n.payload()) {
+                        Ok(notification) => Some(Ok(notification)),
+                        Err(_) => None,
```
tomxey, line 397: I wonder, what other notifications we may receive here?

tomxey (approved):
Leaving some nitpicks, but the solution looks good overall. LGTM

crates/iota-indexer/src/stream.rs
sergiupopescu199: While implementing the streaming filters I also looked on already available implementation of JSON-RPC subscribers filters, based on this implementation I constructed the new filter: https://github.com/iotaledger/iota/blob/develop/crates/iota-indexer/src/indexer_reader.rs#L1137-1164

I saw that if user provided the package & function name it will result in an error, so instead of returning it I designed the type in a way to guide the user on possible data to insert.

crates/iota-indexer/src/stream.rs
sergiupopescu199: Yes we do, currently the indexer does not connect so a encrypted connection to postgres trough diesel, it came automatic to do the same here, when we'll decide to use secure connection it will be easy to refactor it in the future.

crates/iota-indexer/src/stream.rs
sergiupopescu199: I've considered it but The PR is already quite big, I suppose it could be part of the next issue regarding the optimization of the actual implementation to support multiple subscribers, since that part will include more testing in term of performance it could be perfect to add metrics also.

crates/iota-indexer/src/stream.rs
sergiupopescu199: the `tokio-postgres` crate provides this [enum](https://docs.rs/tokio-postgres/latest/tokio_postgres/enum.AsyncMessage.html), so for now it has only two variants, the notice one is not an error per se so we can skip it for now,  our goal here is only to filter valid message notifications

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9a2cd5e4abc4362252d830d8b922e533d1c3a451

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9a2cd5e4abc4362252d830d8b922e533d1c3a451

crates/iota-indexer/src/stream.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/9a2cd5e4abc4362252d830d8b922e533d1c3a451

crates/iota-indexer/src/stream.rs
tomxey: The type design is good I think. Only the naming is a bit confusing.
Maybe sth like:

```rust
pub struct ModuleFunction {
    pub module: String,
    pub function: Option<String>,
}
```

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8555/commits/d2e3769ff7850ba8905f5d9434111c0fd40519f5

PR #8580, opened 2025-09-16
feat(verifier): Alternative creation of auth info that does not involve verification at runtime

### LINKED ISSUE #8579: [aa][authenticator] Investigate the alternative of a compile time validation

miker83z:
[run-ci]
# Description of change

Experimental feature which avoids the verification of an `authenticate` function at runtime. 
Based on the feature introduced in #8477:
- the `iota-verifier`'s `verify_authenticate_func` is re-used as is;
- what changes is that, this `verify_authenticate_func` is not executed at runtime through a native function, but instead during the "canonical" verification of a module (see changes to the `iota_verify_module_metered` function in the `iota-verifier`'s lib);
- this means that every time the verifier is run, e.g., during a package build, `authenticate` functions are checked.
- How does the verifier know whether a function is an `authenticate` function? (see `iota-execution/latest/iota-verifier/src/authenticate_functions_verifier.rs`)
  - it checks in the init of a module;
  - if it does NOT invoke `0x2::account::public_authenticate_registry`, then it is considered to not use authenticate functions and thus the module is considered valid;
  - if it invokes `0x2::account::public_authenticate_registry`, then it must conform to the rules:
    - the input of such function MUST be a constant representing a list of authenticate functions;
    - the constant must be of type `vector<vector<u8>>`; 
    - each inner vector<u8> must be a valid UTF-8 string that can be serialized into the name of a function in the module;
    - such function must also conform to the rules for `authenticate` functions.
- The `0x2::account::public_authenticate_registry` function: 
  - needs an OTW parameter, so it can be called only during the publishing of a package;
  - uses the OTW to derive the package id and module name of an `authenticate` function;
  - creates a registry of `AuthenticatorInfoV1`s in the form of an immutable object which can be used by anyone. 

Fixes #8579 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


account_auth_example/sources/account_auth_example.move
```diff
@@ -0,0 +1,32 @@
...
+
+// temporary placeholder
+public struct AuthContext has drop {}
+
+public struct MAIN_M has drop {}
+
+const AUTHENTICATE_FUNCTIONS: vector<vector<u8>> = vector[b"arg_value"];
+
+fun init(witness: MAIN_M, ctx: &mut TxContext) {
+    public_authenticate_registry(
```
TheMrAI: Do I understand correctly that `init` is only called on the initial deployment of the contract and it is not re-run ever again during package upgrades?
If so won't this pattern lock in the authenticate functions for the whole lifetime of the contract without any option of changing them?

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -60,3 +63,48 @@ public fun create_auth_info_v1_for_testing(
...
+/// can only be called once for a specific `T`.
+public fun public_authenticate_registry<OTW: drop>(
+    witness: &OTW,
+    function_names: vector<vector<u8>>,
+    ctx: &mut TxContext,
+) {
+    // Verify that the type is an OTW
+    assert!(types::is_one_time_witness(witness), ENotOneTimeWitness);
+
+    let type_name = type_name::get_with_original_ids<OTW>();
```
TheMrAI, line 79: Does this mean that we rely on the location of the one time witness in the source to identify the `authenticate` function?
What happens if the `authenticate` function and the otw are in different modules?

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -60,3 +63,48 @@ public fun create_auth_info_v1_for_testing(
...
+    // Share the record globally
+    transfer::freeze_object(AuthenticateRegistry {
+        id: object::new(ctx),
+        package: type_name.get_address(),
+        module_name: type_name.get_module(),
+        function_names,
+    });
+}
+
+public fun create_auth_info_v2(
```
TheMrAI: If the registry is frozen, could the registry just contain the `AuthenticatorInfoV1`'s in it?
Perhaps no need to create them.

TheMrAI (commented):
Can we try to find a name for this "compile time".
Because this isn't compile time, we are after it, but before publish/runtime.
I am not aware of similar patterns, maybe "verification time" or "static verification time" since we see that there may be "runtime verifications" in the future.

I don't dislike the idea, but seems a bit hackier then the runtime version. Which I don't like either, but it makes a bit more conceptual sense to me. Even though in terms of usability this could be user friendlier.

If I am correct with my observed limitations above, can we remove them somehow?

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: Yes, this would be a good optimization!

account_auth_example/sources/account_auth_example.move
miker83z: Yes exactly, currently you can't add new functions to the registry once it is created. There could be the possibility to upgrade the logic maybe. Something like this could be of help: https://github.com/MystenLabs/sui/pull/23212

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: Indeed. Each module would have its own `OTW` and `init` function in which `public_authenticate_registry` is invoked.

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: [30bb9eb](https://github.com/iotaledger/iota/pull/8580/commits/30bb9eb566127c8da19fb0677538e882a84bac07)

iota-execution/latest/iota-verifier/src/authenticate_functions_verifier.rs
```diff
@@ -0,0 +1,207 @@
...
+/// has a call instruction to the `0x2::account::publish_authenticate_registry`
+/// function within the `init` function.
+///
+/// If the module does not have such call instruction, then it is considered to
+/// not use authenticate functions and thus the module is considered valid.
+/// If the module does have such call instruction, then it must conform to the
+/// rules.
+pub fn verify_module(view: &CompiledModule) -> Result<(), ExecutionError> {
+    // For each function in the module, look for the `init` function
+    for fdef in &view.function_defs {
```
TheMrAI, line 42: Can there be multiple init functions in a module?
Can we exit after finding one?

iota-execution/latest/iota-verifier/src/authenticate_functions_verifier.rs
```diff
@@ -0,0 +1,207 @@
...
+
+/// Checks if the module conforms to the authenticate functions rules only if it
+/// has a call instruction to the `0x2::account::publish_authenticate_registry`
+/// function within the `init` function.
+///
+/// If the module does not have such call instruction, then it is considered to
+/// not use authenticate functions and thus the module is considered valid.
+/// If the module does have such call instruction, then it must conform to the
+/// rules.
+pub fn verify_module(view: &CompiledModule) -> Result<(), ExecutionError> {
```
TheMrAI, line 40: Not entirely certain, so just nothing my concerns here with this approach here.
Verifiers by design are called on a single CompiledModule and I don't know of a simple way to accessing other CompiledModules from the package.

The following code seems to rely on the assumption that the OTW, the init function, the const AUTHENTICATE_FUNCTIONS and the referred authenticate functions all reside in the
same module.
Do I see it correctly? If so, this may be a very big restriction, and I am not sure how we could sidestep it.

iota-execution/latest/iota-verifier/src/authenticate_functions_verifier.rs
miker83z: Yes, it should exit after finding the init function

iota-execution/latest/iota-verifier/src/authenticate_functions_verifier.rs
miker83z: The assumption of this solution is that there will be an OTW and a `publish_authenticate_registry` for each module that wants to have/declare some authenticate functions. 

So yes, this means possibly multiple registries for each package.

PR #8585, opened 2025-09-17
feat(consensus): compress ancestor and ack references into a single vector

### LINKED ISSUE #8151: compress  `acknowledgments` and `ancestors` references as they overlap most of the time.
`BlockHeaderV1` `acknowledgments` and `ancestors` fields in practice contain most of the same references, a data structure should be used that is more efficient then two vectors.

filipdulic:
# Description of change
Compresses the ancestor and acknowledgements vectors into a single reference vector with overlap indices. Using the following layout.

```
references: Vec<BlockRef>, // layout: |ancestors|overlap|acks|ancestor[0]?|
overlap_start_index:u8, // bounded by committee size (<=256)
overlap_end_index: u8, // same bound as above
```
`ancestors -> &references[..overlap_end_index]`
`acks -> &references[overlap_start_index..]`
- ancestor[0] is the first ancestor supplied during block header creation and is also the first  reference in references.
  - it is not added in the overlap part of references, but is added to the end of the vectors if it is acknowledged.
- ancestors - non overlapping ancestors excluding ancestor 0
- overlap - ancestor and ack overlap without ancestor 0
- acks - non overlapping acks

This implementation adds a bit of computational overhead during block header creation. But compresses to a single shared reference vector with 2 bytes for the indices as overhead and a single BlockRef duplication. It keeps access to `ancestors` / `acknowledgements` in a single continuous array in memory.

Closes #8151 

### DIFF
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    collections::HashSet,
     fmt,
     hash::{Hash, Hasher},
     ops::Deref,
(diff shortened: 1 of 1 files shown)


crates/starfish/core/src/block_header.rs
```diff
@@ -114,26 +115,79 @@ impl BlockHeaderV1 {
...
+            references,
+            overlap_start_index,
+            overlap_end_index,
             transactions_commitment,
             commit_votes,
         }
     }
+    /// Compresses ancestors and acknowledgments into a single references
+    /// vector, and returns the overlap indices. The first ancestor is
+    /// always the first reference (ref0). If it is also in acknowledgments,
```
piotrm50, line 134: Can it ever happen that ref0 is an ancestor but not an ack though?

piotrm50 (approved):
Not a big fan of this header layout tbh, especially with ref0 being potentially duplicated. Perhaps we can always treat ref0 as an ack since it's produced by the same authority? I understand that the alternative approach with a tuple was more complicated because we would have to clone those ancestors, but how many times would that happen per block? I don't like that the main argument for changing the protocol in such an ugly way (byte layout sent over-the-wire) to optimize implementation with less copying. 

I approve this regardless as it will reduce header size significantly and we can think about possible alternative approaches later.

polinikita (approved):

PR #8593, opened 2025-09-18
feat(json-rpc): add new method isTransactionIndexedOnNode

### LINKED ISSUE #8526: feat(json-rpc): query existence of transaction on the node from the json-rpc
## Description

Implement `iota_isTransactionIndexedOnNode`.

kodemartin:
# Description of change

This patch adds a new endpoint `iota_isTransactionIndexedOnNode`

Fixes #8526

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: A new method `iota_isTransactionIndexedOnNode` is added that allows clients and dapps to optionally poll the fullnode to verify that a transaction has been indexed. The latter is a necessary condition so that any dependent transaction is executed successfully. The new method complements the recent advancements in the Indexer JSON-RPC API that  indexes transactions after execution without waiting for a checkpoint to include them. 
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-json-rpc-tests/tests/read_api.rs
+++ b/crates/iota-json-rpc-tests/tests/read_api.rs
@@ -551,6 +551,27 @@ async fn get_transaction_block() {
 }
 
 #[sim_test]
+async fn is_transaction_present() {
+    let cluster = TestClusterBuilder::new().build().await;
+    let address = cluster.get_address_0();
+
+    let (object_ids, gas) = get_objects_to_mutate(&cluster, address).await;
+
+    let transaction = cluster
+        .transfer_object(address, address, object_ids[0], gas, None)
+        .await
+        .unwrap();
+
+    assert!(
+        cluster
+            .rpc_client()
+            .is_transaction_indexed_on_node(transaction.digest)
+            .await
+            .unwrap()
+    );
+}
+
+#[sim_test]
 async fn get_transaction_block_with_full_content() {
     get_transaction_block_with_options(IotaTransactionBlockResponseOptions::full_content()).await;
 }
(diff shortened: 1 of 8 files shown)


crates/iota-json-rpc/src/read_api.rs
```diff
@@ -727,6 +727,12 @@ impl ReadApiServer for ReadApi {
         .await
     }
 
+    #[instrument(skip(self))]
+    async fn is_transaction_indexed_on_node(&self, digest: TransactionDigest) -> RpcResult<bool> {
+        let transaction = self.get_transaction_block(digest, None).await?;
+        Ok(transaction.digest == digest)
```
tomxey: The behaviour seems a bit strange to me. The function name suggests that it returns if transactions is indexed on the node, so true if it is, and false if it's not.
In other words: if user asks a question IF transaction is indexed, he should expect yes or no as normal answers, so I think there should be no error returned in such standard cases.

Have you considered implementing it like that, or was it problematic for some other reasons?

crates/iota-json-rpc-tests/tests/read_api.rs
```diff
@@ -550,6 +550,27 @@ async fn get_transaction_block() {
     get_transaction_block_with_options(IotaTransactionBlockResponseOptions::default()).await;
 }
 
+#[sim_test]
+async fn is_transaction_present() {
```
tomxey, line 570: could we also test negative case with tx not indexed?

crates/iota-json-rpc/src/read_api.rs
kodemartin: That was an oversight. Good catch! Fixed with 111203a2be

crates/iota-json-rpc-tests/tests/read_api.rs
kodemartin: 111203a2be

Thoralf-M (approved):

tomxey (approved):

PR #8594, opened 2025-09-18
feat(graphql): add isIndexedOnNode field for transactions

### LINKED ISSUE #8592: feat(graphql): transaction indexed on node
## Description

Add a new `query.Transaction.indexedOnNode`  field.

kodemartin:
# Description of change

This patch adds a new `bool` field `isIndexedOnNode` on `TransactionBlock`.

Fixes #8592 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL: Adds a new `Query.isTransactionIndexedOnNode` method that can be used to poll the backing fullnode
to verify that a transaction is indexed locally. This patch also introduces an additional field `TransactionBlock.indexedOnNode` that can be used in a similar way to ask for the index status of transactions just executed.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
@@ -88,6 +88,16 @@ pub(crate) enum TransactionBlockInner {
     },
 }
 
+impl TransactionBlockInner {
+    /// Returns if the transaction is included in a checkpoint.
+    fn is_checkpointed(&self) -> bool {
+        let TransactionBlockInner::Stored { stored_tx, .. } = &self else {
+            return false;
+        };
+        stored_tx.checkpoint_sequence_number >= 0
+    }
+}
+
 /// An input filter selecting for either system or programmable transactions.
 #[derive(Enum, Copy, Clone, Eq, PartialEq, Debug)]
 pub(crate) enum TransactionBlockKindInput {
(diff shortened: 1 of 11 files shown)


crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -231,6 +232,21 @@ impl TransactionBlock {
...
+    async fn indexed_on_node(&self, ctx: &Context<'_>) -> Result<Option<bool>> {
+        let fullnode_client = get_fullnode_client(ctx)?;
+        let Some(digest) = self.native_signed_data().map(|d| d.digest()) else {
+            // dry-run transactions are never indexed
+            return Ok(Some(false));
+        };
+        Ok(Some(
+            fullnode_client
+                .http()
+                .is_transaction_indexed_on_node(digest)
```
tomxey, line 270: I'm thinking a bit about performance here.
Is it possible that someone will request this field for a page of transactions, resulting in a separate request to fullnode for every item on the page?

Is this a consideration? Maybe we should look at some optimizations. (single request for many digests? or maybe we could keep in memory checkpoint id of last tx indexed for given fullnode, so we know that everything that happened earlier is already indexed, without hitting the fullnode, for querying historical data this would practically mean we don't have to query the fullnode at all)

kodemartin: You are right. We should definitely avoid calling the node in the `Query.TransactionBlocks` path. But to this end can take advantage of the invariant that this returns consistent results, i.e. referring to consistent views up to a certain checkpoint. 

I introduced a method `is_checkpointed` with 258abf282e that enables this logic.

The original intention was to use the intuitive `TransactionBlock` path to get the index status on the node. However, upon revisiting this it occurred to me that this path is premised on fetching data from the db, and this implies additional latency.

So I added a direct query `isTransactionIndexedOnNode` with 9f0bcfd331 to enable a more optimal query path.

`TransactionBlock.isIndexedOnNode` still makes sense for  `mutation.executeTransactionBlock` queries though to directly ask for the index status on the node.

tomxey: It makes sense.
Could we also have a test for `isTransactionIndexedOnNode`?

kodemartin: Test added: https://github.com/iotaledger/iota/pull/8594/commits/3e8b3ee0cc9be70effcc84bde96989b0657e28c0

tomxey (approved):

Thoralf-M (approved):

Issue #8517, opened 2025-09-10
Open in a new tab if camera permission is not allowed


marc2332:
Otherwise the browser wont prompt the user to allow the camera. In the other hand, if there is no permission error then we can assume the camera works just fine and so a new tab is not needed

PR #8596, opened 2025-09-18
feat(wallet): Check camera permissions on keystone flow

msarcev:
# Description of change

Closes https://github.com/iotaledger/iota/issues/8517


marc2332 (changes requested):
Also need to check when importing accounts

KeitoTadashi (approved):

marc2332 (approved):

PR #8597, opened 2025-09-18
feat(explorer): add address existence check in getResultsForAddress for names

panteleymonchuk:


apps/explorer/src/hooks/useSearch.ts
```diff
@@ -120,6 +122,35 @@ const getResultsForAddress = async (
     ];
 };
 
+async function addressExists(client: IotaClient, address: string): Promise<boolean> {
```
marc2332:
```suggestion
async function isAddress(client: IotaClient, address: string): Promise<boolean> {
```

apps/explorer/src/hooks/useSearch.ts
```diff
@@ -84,11 +84,13 @@ const getResultsForAddress = async (
 
         if (!nameRecord || !nameRecord.targetAddress) return null;
 
+        const exists = await addressExists(client, nameRecord.targetAddress);
```
marc2332:
```suggestion
        const _isAddress = await isAddress(client, nameRecord.targetAddress);
```

apps/explorer/src/hooks/useSearch.ts
```diff
@@ -120,6 +122,35 @@ const getResultsForAddress = async (
...
+            filter: { FromAddress: normalized },
+            limit: 1,
+        });
+        if (fromTransactions.data.length > 0) return true;
+
+        const toTransactions = await client.queryTransactionBlocks({
+            filter: { ToAddress: normalized },
+            limit: 1,
+        });
+        if (toTransactions.data.length > 0) return true;
```
marc2332: Wouldn't this also work by doing:

```suggestion
const transactions = await client.queryTransactionBlocks({
    filter: { FromOrToAddress: { addr: normalized } },
    limit: 1,
});
if (toTransactions.data.length > 0) return true;
```

?

marc2332 (changes requested):

msarcev (approved):

marc2332 (approved):

PR #8602, opened 2025-09-19
feat(examples): Add "Time locked account" account abstraction example

### LINKED ISSUE #8534: Implement "Time locked account" using account abstraction
Example prototype:
https://iotafoundation.atlassian.net/wiki/spaces/LSC/pages/2479718716/Time+Locked+Account

TheMrAI:
# Description of change

Implement "Time locked account" example for abstract accounts.
This example explores the possibility of utilizing the authentication logic.

Each authenticator type can have a number of "reserved" dynamic fields and provide basic functions for attaching, detaching, rotating, checking and borrowing that type.
Properly using these functions is the responsibility of the implementer, but given the language restrictions this is a very flexible pattern.

fixes #8534 


examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+
+/// Set the `AuthenticatorInfo` for the account.
+///
+/// The `AuthenticatorInfo` will be attached as a dynamic field with key provided by: `account::authenticator_df_name()`
+/// This function must be called, otherwise the builder will fail.
+public fun set_authenticator<Authenticator: copy + drop + store>(
+    self: &mut IOTAccountBuilder,
+    authenticator: Authenticator,
+) {
+    dynamic_field::add(&mut self.account.id, account::authenticator_df_name(), authenticator);
```
TheMrAI: If we refactor `account::authenticator_df_name()` to return a `public struct AuthenticatorKey has copy, drop, store {}` instead, this whole function can me removed and the interface simplified further.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+/// Return `true` if and only if `self` has a dynamic field with the specified `name`.
+public fun has_field<Name: copy + drop + store>(self: &IOTAccount, name: Name): bool {
+    dynamic_field::exists_(&self.id, name)
+}
+
+/// Rotate a reserved dynamic field.
+///
+/// Only the account itself can call this function and the dynamic field must refer be a
+/// reserved one.
+public fun rotate_reserved<Name: copy + drop + store, Value: drop + store>(
```
TheMrAI: I am a little on the fence about this function, because one may want to rotate all of them at once, in which case these repeated checks are a waste of cycle, but not sure how that could be extracted.

examples/move/account_template/tests/account_template_tests.move
```diff
@@ -0,0 +1,121 @@
...
+        // check the ReservedKey value as well
+        assert!(account.has_field(reserved_df_example));
+        assert_eq(*account.borrow_field(reserved_df_example), 6);
+
+        test_scenario::return_shared(account);
+    };
+
+    test_scenario::end(scenario_val);
+}
+
```
TheMrAI: Not testing any functionality that is tested in the `iota_account`examples.
If this abstraction is acceptable, in a later PR we can refactor the examples appropriately.

examples/move/time_locked/sources/unlock_time.move
```diff
@@ -0,0 +1,83 @@
...
+    ctx: &TxContext,
+) {
+    let public_key: &vector<u8> = self.borrow_field(OwnerPublicKey {});
+    // Check the signature.
+    assert!(
+        ed25519::ed25519_verify(&signature, public_key, ctx.digest()),
+        EEd25519VerificationFailed,
+    );
+}
+
```
TheMrAI, line 97: With the abstraction of `account_template`, the user will not have to test basic/common functionality when implementing an account.

examples/move/time_locked/sources/time_locked.move
```diff
@@ -0,0 +1,83 @@
...
+    self: &IOTAccount,
+    clock: &Clock,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+) {
+    account_template::ensure_tx_sender_is_account(self, ctx);
+
+    authenticate_time(self, clock, auth_ctx, ctx);
+    authenticate_ed25519(self, signature, auth_ctx, ctx);
```
TheMrAI: Ideally I would want to use the implementation in the `iota_account` example, but didn't want to touch it as long as we are not sure of this approach.
If this is okay, then in a later PR this can be cleaned up.

In any case it shows in however limited capacity that we can kinda build authenticate functionality on top of each other.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+module account_template::account_template;
+
+use iota::account;
+use iota::bcs;
+use iota::dynamic_field;
+
+// The following template provides the basic tools for a developer for implementing an abstract account,
+// with the necessary authentication functions, error codes and expectations.
+
+#[error(code = 0)]
```
TheMrAI: Just a reminder here if you know by any chance if this could be a problem.
Not sure if there could be an issue with error codes colliding or becoming very unwieldy to the developer when making an account implementation.
For example they use **account_template**, **some_authenticator** in their **best_account** package.
All there having error codes at 0, 1, 2, 3 etc.
Are you aware this possibly causing an issue somewhere?

examples/move/time_locked/tests/time_locked_tests.move
```diff
@@ -0,0 +1,165 @@
...
+    };
+
+    test_scenario::end(scenario_val);
+}
+
+#[test]
+fun account_unlocked() {
+    let mut scenario_val = test_scenario::begin(@0x0);
+    let scenario = &mut scenario_val;
+    let public_key = x"cc62332e34bb2d5cd69f60efbb2a36cb916c7eb458301ea36636c4dbb012bd88";
```
Dkwcs, line 145: Where did you get those parameters like pub_key, digest and signature?
Could you add a comment or something that explains it in short manner ?

TheMrAI: To be honest I stole it from @valeriyr  :D

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+const EAuthenticatorNotSet: vector<u8> =
+    b"An 'AuthenticatorInfo' must be set for account creation.";
+#[error(code = 1)]
+const EReservedDynamicFieldsNotSet: vector<u8> =
+    b"An account must have at least one restricted dynamic field set.";
+#[error(code = 2)]
+const EReservedDynamicFieldsListCannotBeSet: vector<u8> =
+    b"'ReservedDfNames' field cannot be set by the user.";
```
Dkwcs: We already discussed it with Davo, just leaving here a reminder 
The error is pretty clear, however we can add a small clarification like why we need this

examples/move/account_template/tests/account_template_tests.move
```diff
@@ -0,0 +1,121 @@
...
+    let ctx = test_scenario::ctx(scenario);
+
+    let mut builder = account_template::init_account_builder(ctx);
+
+    let authenticator = create_authenticator_info_v1_for_testing();
+    builder.set_authenticator(authenticator);
+    // Any field value can be set as a reserved, and for the purposes of this test
+    // the exact value doesn't matter.
+    let reserved_df_example = ReservedDfExample {};
+    builder.add_reserved_field(reserved_df_example, 6);
```
Dkwcs: Why it’s not failing, while it does

```
    builder.add_reserved_field(
        account_template::get_reserved_df_names(),
        vector<std::type_name::TypeName>[],
    );
```

examples/move/account_template/sources/account_template.move
Dkwcs: Why is it needed?

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+#[error(code = 3)]
+const ETransactionSenderIsNotTheAccount: vector<u8> = b"Transaction must be signed by the account.";
+/// It should be emitted every time when an attempt at modifying a restricted dynamic field was made in
+/// an inappropriate scope scope. This scope will generally be defined by the account implementers.
```
Dkwcs: I guess 1 scope is extra

examples/move/time_locked/sources/time_locked.move
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/// Verify if the account has passed the unlock time.
+///
+/// Looks like an `authenticate` function, but it isn't as it is private. Nor would it provide
+/// satisfactory access protection for the account itself.
+fun authenticate_time(self: &IOTAccount, clock: &Clock, _auth_ctx: &AuthContext, _ctx: &TxContext) {
```
Dkwcs: Is it usable only in authenticate function? 
I mean I’m just trying to get the main purpose - so we can authenticate only if amount of time we set is passed?

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+    has_to_be_reserved: bool,
+) {
+    let reserved_df_names: &vector<std::type_name::TypeName> = dynamic_field::borrow(
+        &self.id,
+        ReservedDfNames {},
+    );
+    let reserved_found =
+        reserved_df_names.any!(|reserved| reserved == std::type_name::get<Name>()) || is_authenticate(name);
+
+    if (has_to_be_reserved) {
```
Dkwcs: It’s a bit tricky for me
Do we have some documentation with the requirements on confluence that explain this flow?

examples/move/account_template/sources/account_template.move
TheMrAI: The account_template provides safety for not accidentally modifying the reserved dynamic fields.
But in practice this is problematic if you would want to rotate your public keys or any data that you need for authentication.
So you need a way to update them somehow, to give some flexibility to account handling.
Now this isn't perfect, but I didn't have a better idea. 

The pattern is still breakable if one rotates out the reserved fields to something else, but I see no real way of properly restricting it in a flexible manner.

examples/move/account_template/tests/account_template_tests.move
TheMrAI: Not sure I follow. Could you elaborate?
Maybe you are referring to the
`get_reserved_df_names()` => `ReservedDfNames`
And that here the example uses `ReservedDfExample` ?

examples/move/time_locked/sources/time_locked.move
TheMrAI: It is like a gate that only opens after a certain point in time.
The unlock_time is a point in time, not an amount of time.
So this opens once and it will stay open.

Yep, this function is only for authentication purposes. Anything that receives an `AuthContext` can only be called in authentication scenarios/within authenticate functions, because `AuthContext` can't be made, only by the system (at least this is how I think the design stands).

examples/move/account_template/sources/account_template.move
TheMrAI: Nope, as far as I know there is none, only this pseudocode: https://iotafoundation.atlassian.net/wiki/spaces/LSC/pages/2479718716/Time+Locked+Account

But the patterns used here come from: https://github.com/iotaledger/iota/tree/95fa86891d118a8e65f5265a88bef510233171cf/examples/move/iota_account

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+///
+/// The builder is entirely temporary. It cannot be copied, stored or dropped.
+/// Account implementations are expected to call the builder in a single function call,
+/// add the desired `AuthenticatorInfo` and all reserved dynamic fields necessary for the
+/// operation of the account authentication logic.
+/// Reserved fields can't be removed later, only their values may be replaced. For this reason
+/// adding any dynamic fields that are not vital for the specified authenticator should be avoided.
+public struct IOTAccountBuilder {
+    account: IOTAccount,
+    reserved_fields: vector<std::type_name::TypeName>,
```
valeriyr: As we discussed already, using just `std::type_name::TypeName` is not enough to define a dynamic field name.
We should use the bytes representation as well, or implement a new utility function in the `dynamic_field` module to calculate the name value by type.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/// Construct an IOTAccountBuilder.
+public fun init_account_builder(ctx: &mut TxContext): IOTAccountBuilder {
```
valeriyr: we can rename this function to `builder` or `create_builder`, it will be called like `account_template::builder` which makes it clear that it is related to `account_template`.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/// Set the `AuthenticatorInfo` for the account.
+///
+/// The `AuthenticatorInfo` will be attached as a dynamic field with key provided by: `account::authenticator_df_name()`
+/// This function must be called, otherwise the builder will fail.
+public fun set_authenticator<Authenticator: copy + drop + store>(
+    self: &mut IOTAccountBuilder,
+    authenticator: Authenticator,
+) {
```
valeriyr: I propose to refactor all the builder functions to return the vuilder instance:
```
public fun set_authenticator<Authenticator: copy + drop + store>(
    self: IOTAccountBuilder,
    authenticator: Authenticator,
): IOTAccountBuilder { ... }
```

examples/move/account_template/sources/account_template.move
valeriyr: As I see, we can treat the authenticator as a reserved field, so we can just call `add_reserved_field` here.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+    dynamic_field::add(&mut self.account.id, account::authenticator_df_name(), authenticator);
+}
+
+/// Attach a `Value` as a dynamic field to the builder.
+public fun add_reserved_field<Name: copy + drop + store, Value: store>(
```
valeriyr: We need to add a function that allows tus o add non-reserved dynamic fields. It is necessary because dynamic fields can be added only by the account itself that is impossible while creating.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+    self: &IOTAccount,
+    name: &Name,
+    has_to_be_reserved: bool,
+) {
+    let reserved_df_names: &vector<std::type_name::TypeName> = dynamic_field::borrow(
+        &self.id,
+        ReservedDfNames {},
+    );
+    let reserved_found =
+        reserved_df_names.any!(|reserved| reserved == std::type_name::get<Name>()) || is_authenticate(name);
```
TheMrAI: Don't forget to also check the actual value in the Name. The type is not enough in case built-in/framework types are used like a String.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/// Finish building the `IOTAccount` and share the object.
+///
+/// The call fails if no authenticator or reserved dynamic fields were set.
+public fun finish_and_share(self: IOTAccountBuilder) {
+    let IOTAccountBuilder { mut account, reserved_fields } = self;
+
+    let authenticator_set = dynamic_field::exists_(&account.id, account::authenticator_df_name());
+    assert!(authenticator_set, EAuthenticatorNotSet);
```
valeriyr: If we want a user force to set the authenticator, we can pass it to the `build` function and remove `set_authenticator`.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+    );
+
+    dynamic_field::add(&mut self.account.id, name, value);
+    self.reserved_fields.push_back(type_name);
+}
+
+/// Finish building the `IOTAccount` and share the object.
+///
+/// The call fails if no authenticator or reserved dynamic fields were set.
+public fun finish_and_share(self: IOTAccountBuilder) {
```
valeriyr: I propose returning the IOTAccount here without sharing and adding the following function:
```rust
public fun share(self: IOTAccount) {
    iota::transfer::share_object(account);
}
```
It makes it possible to do something with the account instance before sharing, for example, get the address.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+    let IOTAccountBuilder { mut account, reserved_fields } = self;
+
+    let authenticator_set = dynamic_field::exists_(&account.id, account::authenticator_df_name());
+    assert!(authenticator_set, EAuthenticatorNotSet);
+
+    assert!(!reserved_fields.is_empty(), EReservedDynamicFieldsNotSet);
+    dynamic_field::add(&mut account.id, ReservedDfNames {}, reserved_fields);
```
valeriyr: It looks to me that this collection can be empty; we can add only the authenticator here.
It is useful when you don't need any data for the authenticator.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+/// Only the account itself can call this function and the dynamic field must refer be a
+/// reserved one.
+public fun rotate_reserved<Name: copy + drop + store, Value: drop + store>(
+    self: &mut IOTAccount,
+    name: Name,
+    value: Value,
+    ctx: &TxContext,
+) {
+    ensure_tx_sender_is_account(self, ctx);
+    check_df<Name>(self, &name, true);
```
valeriyr: Instead of having a bool param, we can add two functions, one checks that the name is reserved, and another that it is not reserved. It will be easier to read.

examples/move/account_template/sources/account_template.move
```diff
@@ -0,0 +1,251 @@
...
+
+    let account_id = &mut self.id;
+    dynamic_field::remove<_, Value>(account_id, name);
+    dynamic_field::add(account_id, name, value);
```
valeriyr: We can add an utility function like this:
```
/// Updates a dynamic field value and returns the previous one.
/// It is supposed that the dynamic field with the given name already exists.
fun update_dynamic_field<Name: copy + drop + store, Value: store>(
    account_id: &mut UID,
    name: Name,
    value: Value
): Value {
    let previous_value = dynamic_field::remove(account_id, name);
    dynamic_field::add(account_id, name, value);
    previous_value
}
```
It makes code cleaner; it is not necessary to explicitly set the value type.

examples/move/account_template/sources/account_template.move
valeriyr: if treat the authenticator field as reserved, we don't need to check `is_authenticate(name)` here.

valeriyr (commented):
I personally would prefer to create a dedicated PR for `account_template`, because it is an independent feature from the `time_locked` example.

valeriyr:
@TheMrAI I see that the `account_template` file contains `IOTAccount` implementation. Are you going to refactor the original `IOTAccount`? I think it would be better to refactor it and not add a new file.

examples/move/account_template/tests/account_template_tests.move
Dkwcs: Oh gosh, yeah, that’s what I was asking, thanks

examples/move/account_template/sources/account_template.move
TheMrAI: Hehe I just commented it above :D

examples/move/account_template/sources/account_template.move
TheMrAI: I am hesitant with such a change. Until we don't need to update it often replacing two lines for 10 just adds noise.
There is exactly one call site at the moment.
When there is more, we can refactor it.

examples/move/account_template/sources/account_template.move
TheMrAI: Good point! This was some legacy oversight.

examples/move/account_template/sources/account_template.move
TheMrAI: After our offline discussion I will update it.

examples/move/account_template/sources/account_template.move
miker83z: I think we can consider these codes only relative to the module itself. Check this in the sui framework: https://github.com/MystenLabs/sui/blob/0b1a1d351aeb441cea7b1717c725c97dc0d1df6e/crates/sui-framework/packages/sui-framework/sources/transfer.move#L44

examples/move/time_locked/sources/time_locked.move
```diff
@@ -0,0 +1,83 @@
+module time_locked::time_locked;
```
Dkwcs: I guess that is missing 

TheMrAI: Dang it. I despise those copyright things XD

examples/move/account_template/sources/account_template.move
TheMrAI: Not yet, need to refactor authenticator_df_name first, but that is the idea. First I will fix everything in this PR, then when splitting do that refactor as well.

examples/move/account_template/sources/account_template.move
TheMrAI: You are right. I just couldn't conceive of such a scenario, so far so opted to keep it requiring at least one.

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 2e4a0b426a

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 2e4a0b426a

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 2e4a0b426a

examples/move/time_locked/sources/time_locked.move
TheMrAI: Fixed in: 2e4a0b426a

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 2e4a0b426a

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 73999ed61195ac8d64b388646f77dc30fc6d3943

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 73999ed61195ac8d64b388646f77dc30fc6d3943

examples/move/account_template/sources/account_template.move
TheMrAI: Fixed in: 73999ed61195ac8d64b388646f77dc30fc6d3943

examples/move/account_template/sources/account_template.move
TheMrAI: Closing it, before the re-review.

examples/move/account_template/sources/account_template.move
TheMrAI: Closing it, before the re-review.

examples/move/account_template/sources/account_template.move
TheMrAI: Closing it, before the re-review.

examples/move/account_template/sources/account_template.move
TheMrAI: Closing it, before the re-review.

examples/move/iotaccount/sources/basic_keyed_account.move
```diff
@@ -133,3 +138,45 @@ public fun borrow_public_key(account: &IOTAccount): &vector<u8> {
...
+
+// Had to suffix with _id to avoid collision
+public fun borrow_public_key_id(account_id: &UID): &vector<u8> {
+    dynamic_field::borrow(account_id, OwnerPublicKey {})
+}
+
+public fun authenticate_ed25519_signature(
+    account_id: &UID,
+    signature: vector<u8>,
+    digest: &vector<u8>,
```
valeriyr: This field can be called `message` because, in this case, it reflects the `ed25519_verify` function parameter.
It is not an obligation to sign the digest.

examples/move/iotaccount/sources/basic_keyed_account.move
```diff
@@ -19,6 +20,10 @@ const EEd25519VerificationFailed: vector<u8> = b"Ed25519 authenticator verificat
 const ESecp256k1VerificationFailed: vector<u8> = b"Secp256k1 authenticator verification failed.";
 #[error(code = 2)]
 const ESecp256r1VerificationFailed: vector<u8> = b"Secp256r1 authenticator verification failed.";
+#[error(code = 4)]
+const EPublicKeyAttached: vector<u8> = b"Public key already attached.";
+#[error(code = 5)]
+const EPublicKeyMissing: vector<u8> = b"Public key missing.";
```
valeriyr: 3 and 4?

examples/move/iotaccount/sources/iotaccount.move
```diff
@@ -188,6 +188,10 @@ public fun ensure_tx_sender_is_account(self: &IOTAccount, ctx: &TxContext) {
     assert!(self.id.uid_to_address() == ctx.sender(), ETransactionSenderIsNotTheAccount);
 }
 
+public fun ensure_tx_sender_is_account_id(account_address: &address, ctx: &TxContext) {
```
valeriyr: I would propose to pass `UID` here instead of `address`. `UID` identifies that you call this function from a protected part of code, while `address` can be passed from everywhere.

examples/move/time_locked/sources/unlock_time.move
```diff
@@ -0,0 +1,122 @@
...
+    ctx: &TxContext,
+) {
+    iotaccount::ensure_tx_sender_is_account_id(&id.to_address(), ctx);
+
+    basic_keyed_account::authenticate_ed25519_signature(id, signature, ctx.digest());
+    let now = clock.timestamp_ms();
+    authenticate_unlock_time(id, now);
+}
+
+public fun authenticate_unlock_time(account_id: &UID, current_time: u64) {
```
valeriyr, line 74: We can consider adding the following two functions:
```rust
public fun authenticate_with_epoch_timestamp(account_id: &UID, ctx: &TxContext) { ... }
public fun authenticate_with_clock(account_id: &UID, clock: &Clock) { ... }
```
They add protection from passing a wrong value.

examples/move/time_locked/sources/time_locked.move
```diff
@@ -0,0 +1,122 @@
...
+    basic_keyed_account::attach_public_key(&mut id, public_key);
+    attach_unlock_time(&mut id, unlock_time);
+
+    let account = TimeLocked { id };
+    iota::transfer::share_object(account);
+}
+
+/// Authenticate access for the `Time locked account`.
+public fun authenticate(
+    id: &UID,
```
valeriyr: This parameter must be changed to `account: & TimeLocked` because it is the account authorization entry point.

examples/move/time_locked/sources/time_locked.move
```diff
@@ -0,0 +1,122 @@
...
+
+// === View Functions ===
+
+public fun account_address(self: &TimeLocked): address {
+    self.id.to_address()
+}
+
+public fun borrow_uid(self: &TimeLocked): &UID {
+    &self.id
+}
```
valeriyr: This function should be removed.
All the functions you expected to be called from `TimeLocked` need to be explicitly exposed to be gated with the following parameter `self: &TimeLocked`.
Even adding it as `test_only` seems unnecessary.
If you want to test, for example, `authenticate_unlock_time`, you can create a dedicated struct for the test module(TestAccount) and test on it. An example can be found [here](https://github.com/iotaledger/iota/blob/a04b77022f5dd29f1878c9bdec5160a45f6286fe/crates/iota-framework/packages/iota-framework/tests/account_tests.move#L13).

examples/move/time_locked/sources/utils.move
```diff
@@ -0,0 +1,122 @@
...
+    dynamic_field::remove(account_id, UnlockTime {})
+}
+
+public fun rotate_unlock_time(account_id: &mut UID, unlock_time: u64): u64 {
+    assert!(has_unlock_time(account_id), EUnlockTimeMissing);
+
+    let prev_unlock_time = dynamic_field::remove(account_id, UnlockTime {});
+    dynamic_field::add(account_id, UnlockTime {}, unlock_time);
+    prev_unlock_time
+}
```
valeriyr: If we consider `UnlockTime` as an extension that can be used by other account types, we need to extract it in a dedicated module.
If we don't, we need to change this first parameter to `self: &TimeLocked`. `TimeLocked` shouldn't expose `UID` to users.

examples/move/iotaccount/sources/basic_keyed_account.move
TheMrAI: Fixed in: 0c89d589e2

examples/move/iotaccount/sources/basic_keyed_account.move
TheMrAI: Fixed in: 0c89d589e2

examples/move/iotaccount/sources/iotaccount.move
TheMrAI: Fixed in: 0c89d589e2

examples/move/time_locked/sources/time_locked.move
TheMrAI: Fixed in: 0c89d589e2

examples/move/time_locked/sources/utils.move
TheMrAI: Fixed in: 67be434a92

examples/move/time_locked/sources/time_locked.move
TheMrAI: Fixed in: 17a5ce07fc

examples/move/time_locked/sources/unlock_time.move
TheMrAI: Fixed in: 65c5767dd8

examples/move/time_locked/sources/account.move
```diff
@@ -0,0 +1,81 @@
...
+    clock: &Clock,
+    signature: vector<u8>,
+    _auth_ctx: &AuthContext,
+    ctx: &TxContext,
+) {
+    iotaccount::ensure_tx_sender_is_account_id(&account.id, ctx);
+
+    owner_public_key::authenticate_ed25519(&account.id, signature, ctx.digest());
+    let now = clock.timestamp_ms();
+    unlock_time::authenticate_unlock_time(&account.id, now);
```
valeriyr: Let's use a new function `authenticate_with_clock` here.

examples/move/time_locked/sources/unlock_time.move
```diff
@@ -0,0 +1,82 @@
...
+    prev_unlock_time
+}
+
+public fun authenticate_with_epoch_timestamp(account_id: &UID, ctx: &TxContext) {
+    authenticate_unlock_time(account_id, ctx.epoch_timestamp_ms())
+}
+
+public fun authenticate_with_clock(account_id: &UID, clock: &Clock) {
+    authenticate_unlock_time(account_id, clock.timestamp_ms())
+}
```
valeriyr, line 70: It would be nice to add some tests for these functions.

valeriyr (approved):
I left a few comments, but overall it looks good to me!

examples/move/time_locked/sources/unlock_time.move
TheMrAI: I was considering it originally, but it seems a bit excessive as both call `authenticate_unlock_time` directly which is fully tested.
Wdyt?

examples/move/time_locked/sources/account.move
TheMrAI: Fixed in: 3d2d1da8bc

examples/move/time_locked/sources/unlock_time.move
TheMrAI: Fixed in: 00d607e124

miker83z (approved):
LGTM!

PR #219, opened 2024-05-15
refactor(explorer, wallet, sdk): Unified, extensible and private configuration for Networks

marc2332:
# Description of change

- Unifies all network-related environment variables and hard-coded values into 1 single JSON served as an env var to the apps and the SDK
- The `IOTA_NETWORKS` env var contains all the networks configurations
- The `DEFAULT_NETWORK` env var contains the default network to be used
- I removed `dotenv-webpack` because it doesn't support multi-line `.env` files
- Any change to the `.env` file should invalidate the cache of turborepo so it should run just fine on CI/CD

Example using the public SUI endpoints

```env
DEFAULT_NETWORK = 'testnet'

IOTA_NETWORKS = '
{
    "mainnet": {
        "id": "mainnet",
        "name": "Mainnet",
        "url": "https://sui-mainnet.mystenlabs.com/json-rpc",
        "explorer": "https://suiexplorer.com"
    },
    "testnet": {
        "id": "testnet",
        "name": "Testnet",
        "url": "https://sui-testnet.mystenlabs.com/json-rpc",
        "explorer": "https://suiexplorer.com",
        "faucet": "https://faucet.testnet.sui.io/gas"
    }
}
'
```

## But Why?
- Values are not hardcoded
   - Private and harder to leak
   - Easier to change as you only need to change the env var in whatever deploy platform it's being deployed and deploy again, instead of having to change the code itself
- Code is easier to understand and scale as all the values come from the same place   

## Links to any relevant issues

Closes https://github.com/iotaledger/kinesis/issues/198

## Type of change

- Refactor

## How the change has been tested

Running the Wallet and the Explorer

## Change checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

PR #8619, opened 2025-09-23
fix(wallet): Use current chain for wallet dry runs when requesting txs signings

### LINKED ISSUE #8428: Wallet ignores chain parameter for transaction signing
## Bug description

When requesting the web wallet to sign a transaction, there is a chain parameter to specify the network like `iota:mainnet`. This should be used to do the dry run in the correct network, right now it's ignored and always uses the network selected in the wallet itself

https://github.com/iotaledger/iota/blob/abf4d2b61667d1ebf7fc71385bb188aa619b2ddb/sdk/wallet-standard/src/features/iotaSignTransaction.ts#L31

## Steps To reproduce the bug

1. call signTransaction() with the chain parameter set to a different network than what's selected in the browser wallet
2. dry run fails

## Expected behaviour

Wallet should do the dry run with the node for the corresponding chain parameter and then the dry run should not fail

## Actual behaviour

Dry run fails

marc2332:
Closes https://github.com/iotaledger/iota/issues/8428

Dapps only need to add the new `chain` property in `WalletProvider` https://github.com/iotaledger/iota/blob/533e0b08eab50592c01980ff4e0c1a1ecd6547be/apps/wallet-dashboard/providers/AppProviders.tsx#L62 for this to work, that or pass `chain` every time `useSignAndExecuteTransaction` is called.


apps/wallet/src/shared/messaging/messages/payloads/transactions/approvalRequest.ts
```diff
@@ -13,6 +13,7 @@ export type TransactionDataType = {
     type: 'transaction';
     data: string;
     account: string;
+    chain?: ChainType;
```
brancoder, line 16: should we use the `IdentifierString` type from wallet standard here? and also possibly remove out ChainType type, because now we have two types for the same thing across projects?
seams we have added the CoinType in this PR https://github.com/iotaledger/iota/pull/219/files#diff-05477b1fd3a3bfb3af4614271e1969cc43f1677991e10d40fe5b6d6bce7deac0

brancoder (changes requested):
in walletStandardInterface.ts `#signTransaction`  we spread the input `...input,` so the chain property gets sent to approval request:
image

but in the `#signAndExecuteTransaction` we dont propagate the chain property. we could add it also so that we don't have the same issue when using `#signAndExecuteTransaction:`
image

to test this out qickly i just changed the `signTransaction` method for `signAndExecuteTransaction` method in `useSignAndExecuteTransaction` hook
image

apps/wallet/src/shared/messaging/messages/payloads/transactions/approvalRequest.ts
marc2332: ChainType is also used in the sdk so its weird to make it depend on the wallet standard no? its just a type alias anyway

apps/wallet/src/ui/app/walletSigner.ts
```diff
@@ -28,20 +32,34 @@ export abstract class WalletSigner {
         this.client = client;
     }
 
+    protected getClient(chain?: ChainType): IotaClient {
+        const requestedNetwork = Object.values(getAllNetworks() ?? {}).find(
+            (network) => network.chain === chain,
+        ) as NetworkId | undefined;
+        const client = requestedNetwork
+            ? new IotaClient({ url: getFullnodeUrl(requestedNetwork) })
```
brancoder: we can access the url from the found requestedNetwork directly. no need to call `getFullnodeUrl` because it just calls -> getNetwork() -> getAllNetworks().
ex.
```
 const requestedNetwork = Object.values(getAllNetworks() ?? {}).find(
            (network) => network.chain === chain,
        );
        const client = requestedNetwork
            ? new IotaClient({ url: requestedNetwork.url })
            : this.client;
        return client;
```

brancoder (changes requested):
LG 🚀  just a nit about client url

brancoder (approved):
LGTM Nice job 🚀

brancoder (changes requested):
its good, just conflicts 🙏

brancoder (approved):
LGTM

msarcev (approved):

PR #8638, opened 2025-09-23
feat: first dynamic multisig account implementation

### LINKED ISSUE #8621: Implement "Dynamic Multisig Account" using account abstraction

valeriyr:
# Description of change

This PR contains a `Dynamic Multisig Account` example implementation.

fixes #8621

### DIFF
--- a/examples/move/dynamic_multisig_account/sources/account.move
+++ b/examples/move/dynamic_multisig_account/sources/account.move
@@ -110,6 +110,8 @@ public fun remove_transaction(self: &mut DynamicMultisigAccount, transaction_dig
 
 /// Updates the account data: members information, threshold and authenticator.
 /// Can be called only by the account itself, that means that this call must be approved by the account members.
+/// The transactions that are proposed but not yet executed can have approves from members
+/// who are not in the new members list. These approves will be ignored when checking if the transaction is approved.
 public fun update_account_data(
     self: &mut DynamicMultisigAccount,
     members_addresses: vector<address>,
--- a/examples/move/dynamic_multisig_account/sources/account.move
+++ b/examples/move/dynamic_multisig_account/sources/account.move
@@ -1,0 +1,216 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+module dynamic_multisig_account::account;
+
+use dynamic_multisig_account::members::{Self, Members};
+use dynamic_multisig_account::transactions::{Self, Transactions};
+
+use iota::account::{Self, AuthenticatorInfoV1};
+use iota::auth_context::AuthContext;
+use iota::dynamic_field;
+
+// --------------------------------------- Errors ---------------------------------------
+
+#[error(code = 0)]
+const ETotalMembersWeightLessThenThreshold: vector<u8> = b"The members weight is less then the threshold.";
+#[error(code = 1)]
+const EThresholdIsZero: vector<u8> = b"The threshold can not be equal to 0.";
+#[error(code = 2)]
+const ETransactionSenderIsNotTheAccount: vector<u8> = b"The user who signed the transaction is not the account.";
+#[error(code = 3)]
+const ETransactionDoesNotHaveSufficientApprovals: vector<u8> = b"The transaction does not have sufficient approvals.";
+
+// -------------------------------- Dynamic Field Names --------------------------------
+
+/// A dynamic field key for the account members.
+public struct MembersKey has copy, drop, store {}
+/// A dynamic field key for the threshold.
+public struct ThresholdKey has copy, drop, store {}
+/// A dynamic field key for the transactions.
+public struct TransactionsKey has copy, drop, store {}
+
+// ---------------------------------- Data Structures ----------------------------------
+
+/// This struct represents a dynamic multisig account.
+public struct DynamicMultisigAccount has key {
+    id: UID,
+}
+
+// -------------------------------------- Creation --------------------------------------
+
+/// Creates a new `DynamicMultisigAccount` instance as a shared object with the given
+/// members, threshold and authenticator.
+public fun create(
+    members_addresses: vector<address>,
+    members_weights: vector<u64>,
+    threshold: u64,
+    authenticator: AuthenticatorInfoV1,
+    ctx: &mut TxContext
+) {
+    // Create a `Members` instance.
+    let members = members::create(members_addresses, members_weights);
+
+    // Verify the provided data consistency.
+    verify_threshold(&members, threshold);
+
+    // Create a UID for an account object.
+    let mut id = object::new(ctx);
+
+    // Add all the data as dynamic fields.
+    dynamic_field::add(&mut id, members_key(), members);
+    dynamic_field::add(&mut id, threshold_key(), threshold);
+    dynamic_field::add(&mut id, account::authenticator_df_name(), authenticator);
+    dynamic_field::add(&mut id, transactions_key(), transactions::create(ctx));
+
+    // Create a mutable shared account object.
+    iota::transfer::share_object(DynamicMultisigAccount { id });
+}
+
+// --------------------------------------- Transactions ---------------------------------------
+
+/// Proposes a new transaction to be approved by the account members.
+/// The member who proposes the transaction is added as the first approver.
+public fun propose_transaction(
+    self: &mut DynamicMultisigAccount,
+    transaction_digest: vector<u8>,
+    ctx: &TxContext
+) {
+    // Get the member who proposed the transaction.
+    let member_address = *self.borrow_members().borrow_member(ctx.sender()).borrow_address();
+
+    // Store the transaction.
+    self.borrow_transactions_mut().add_transaction(transaction_digest, member_address);
+}
+
+/// Approves a proposed transaction.
+public fun approve_transaction(self: &mut DynamicMultisigAccount, transaction_digest: vector<u8>, ctx: &TxContext) {
+    // Get the member who proposed the transaction.
+    let member_address = *self.borrow_members().borrow_member(ctx.sender()).borrow_address();
+
+    // Get the transaction.
+    let transaction = self.borrow_transactions_mut().borrow_transaction_mut(transaction_digest);
+
+    // Approve the transaction.
+    transaction.add_approval(member_address);
+}
+
+// --------------------------------------- Authentication ---------------------------------------
+
+/// Updates the account data: members information, threshold and authenticator.
+public fun update_account_data(
+    self: &mut DynamicMultisigAccount,
+    members_addresses: vector<address>,
+    members_weights: vector<u64>,
+    threshold: u64,
+    authenticator: AuthenticatorInfoV1,
+    ctx: &TxContext
+) {
+    // Check that the sender of this transaction is the account.
+    ensure_tx_sender_is_account(self, ctx);
+
+    // Create a `Members` instance.
+    let members = members::create(members_addresses, members_weights);
+
+    // Verify the provided data consistency.
+    verify_threshold(&members, threshold);
+
+    let account_id = &mut self.id;
+
+    // Update the dynamic fields. It is expected that the fields already exist.
+    update_dynamic_field(account_id, members_key(), members);
+    update_dynamic_field(account_id, threshold_key(), threshold);
+    update_dynamic_field(account_id, account::authenticator_df_name(), authenticator);
+}
+
+/// A transaction authenticator.
+/// Checks that the sender of this transaction is the account and that the transaction is approved.
+public fun authenticate(self: &DynamicMultisigAccount, _: &AuthContext, ctx: &TxContext) {
+    // Check that the sender of this transaction is the account.
+    ensure_tx_sender_is_account(self, ctx);
+
+    // Check that the transaction is approved.
+    ensure_tx_is_approved(self, ctx);
+}
+
+// --------------------------------------- Utilities ---------------------------------------
+
+/// Checks that the sender of this transaction is the account.
+fun ensure_tx_sender_is_account(self: &DynamicMultisigAccount, ctx: &TxContext) {
+    assert!(self.id.uid_to_address() == ctx.sender(), ETransactionSenderIsNotTheAccount);
+}
+
+/// Checks that the transaction is approved.
+/// The total weight of the members who approved the transaction must be greater than or equal to the threshold.
+/// If the members list is changed after the transaction proposal, only the members who are still in the list
+/// are considered for the approval. Their weights are taken from the current members list.
+fun ensure_tx_is_approved(self: &DynamicMultisigAccount, ctx: &TxContext) {
+    let members = self.borrow_members();
+    let transaction = self.borrow_transactions().borrow_transaction(*ctx.digest());
+    let threshold = self.borrow_threshold();
+
+    let mut total_approves = 0;
+
+    transaction.approves().do_ref!(|addr| {
+        if (members.has_member(*addr)) {
+            total_approves = total_approves + members.borrow_member(*addr).weight();
+        }
+    });
+
+    assert!(total_approves >= threshold, ETransactionDoesNotHaveSufficientApprovals);
+}
+
+/// Returns the dynamic field name used to store the members information.
+fun members_key(): MembersKey {
+    MembersKey{}
+}
+
+/// Returns the dynamic field name used to store the threshold.
+fun threshold_key(): ThresholdKey {
+    ThresholdKey{}
+}
+
+/// Returns the dynamic field name used to store the transactions.
+fun transactions_key(): TransactionsKey {
+    TransactionsKey{}
+}
+
+/// Immutably borrows the account members.
+fun borrow_members(self: &DynamicMultisigAccount): &Members {
+    dynamic_field::borrow(&self.id, members_key())
+}
+
+/// Immutably borrows the account transactions.
+fun borrow_transactions(self: &DynamicMultisigAccount): &Transactions {
+    dynamic_field::borrow(&self.id, transactions_key())
+}
+
+/// Mutably borrows the account transactions.
+fun borrow_transactions_mut(self: &mut DynamicMultisigAccount): &mut Transactions {
+    dynamic_field::borrow_mut(&mut self.id, transactions_key())
+}
+
+/// Borrows the account threshold.
+fun borrow_threshold(self: &DynamicMultisigAccount): u64 {
+    *dynamic_field::borrow(&self.id, threshold_key())
+}
+
+/// Verifies the threshold.
+fun verify_threshold(members: &Members, threshold: u64 ) {
+    // Check that the threshold is not zero.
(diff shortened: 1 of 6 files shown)


examples/move/dynamic_multisig_account/sources/members.move
```diff
@@ -0,0 +1,92 @@
...
+public struct Members has drop, store {
+    /// The members collection.
+    list: vector<Member>,
+}
+
+// --------------------------------------- Creation ---------------------------------------
+
+/// Creates a `Members` struct from the given vectors of addresses and weights.
+/// The vectors must have the same length.
+public(package) fun create(addresses: vector<address>, weights: vector<u64>): Members {
```
TheMrAI, line 36: Do we need to ensure that each member can appear at most once in the list of addresses?

examples/move/dynamic_multisig_account/sources/account.move
```diff
@@ -0,0 +1,216 @@
...
+    let member_address = *self.members().member(ctx.sender()).addr();
+
+    // Store the transaction.
+    self.transactions_mut().add_transaction(transaction_digest, member_address);
+}
+
+/// Approves a proposed transaction.
+public fun approve_transaction(self: &mut DynamicMultisigAccount, transaction_digest: vector<u8>, ctx: &TxContext) {
+    // Get the member who proposed the transaction.
+    let member_address = *self.members().member(ctx.sender()).addr();
```
TheMrAI: This confused me a bit.
Won't it be cleaner to say
```
let member_address = ctx.sender(); 
assert!(member_address, some_error_msg)
```

examples/move/dynamic_multisig_account/sources/transactions.move
```diff
@@ -0,0 +1,78 @@
...
+    /// The transaction digest.
+    digest: vector<u8>,
+    /// The members who approved the transaction.
+    approves: vector<address>,
+}
+
+/// Holds the information about the account transactions.
+public struct Transactions has store {
+    /// The members collection.
+    bag: Bag,
```
TheMrAI: What happens to Transactions that never get approved?
How do we eject them? Maybe I am confused, but Bag itself knows nothing about what it contains, it only tracks how many things were added to it.

examples/move/dynamic_multisig_account/sources/transactions.move
```diff
@@ -0,0 +1,78 @@
...
+/// Mutably borrows the account transaction with the provided digest.
+public(package) fun transaction_mut(self: &mut Transactions, digest: vector<u8>): &mut Transaction {
+    self.bag.borrow_mut(digest)
+}
+
+/// Adds a new transaction to the account.
+public(package) fun add_transaction(self: &mut Transactions, digest: vector<u8>, member: address) {
+    self.bag.add(digest, Transaction{digest, approves: vector[ member ]});
+}
+
```
TheMrAI, line 65: How are approved transactions removed after they are executed?

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,216 @@
...
+    // Verify the provided data consistency.
+    verify_threshold(&members, threshold);
+
+    let account_id = &mut self.id;
+
+    // Update the dynamic fields. It is expected that the fields already exist.
+    update_dynamic_field(account_id, members_key(), members);
+    update_dynamic_field(account_id, threshold_key(), threshold);
+    update_dynamic_field(account_id, account::authenticator_df_name(), authenticator);
+}
```
TheMrAI, line 185: Shouldn't we require that all remaining transactions have been resolved/removed?
I can't think of a way it could cause a problem, just asking.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,216 @@
...
+    dynamic_field::borrow_mut(&mut self.id, transactions_key())
+}
+
+/// Borrows the account threshold.
+fun threshold(self: &DynamicMultisigAccount): u64 {
+    *dynamic_field::borrow(&self.id, threshold_key())
+}
+
+/// Verifies the threshold.
+fun verify_threshold(members: &Members, threshold: u64 ) {
```
TheMrAI, line 229: Would it make sense to do this verification while constructing the members?

TheMrAI (commented):
Nice clean code.
Not sure you want to write unit tests now or later?
I think this could also use the `account_template`.
It will provide read access to the dynamic fields for everyone, which you said isn't needed, but it could be a nice "view" function feature.

examples/move/dynamic_multisig_account/sources/account.move
```diff
@@ -0,0 +1,216 @@
...
+}
+
+/// Immutably borrows the account transactions.
+fun transactions(self: &DynamicMultisigAccount): &Transactions {
+    dynamic_field::borrow(&self.id, transactions_key())
+}
+
+/// Mutably borrows the account transactions.
+fun transactions_mut(self: &mut DynamicMultisigAccount): &mut Transactions {
+    dynamic_field::borrow_mut(&mut self.id, transactions_key())
```
Dkwcs, line 190: In this case we don’t have a possibility to store more than 1 transaction, do we?

examples/move/dynamic_multisig_account/sources/account.move
```diff
@@ -0,0 +1,216 @@
...
+
+/// Proposes a new transaction to be approved by the account members.
+/// The member who proposes the transaction is added as the first approver.
+public fun propose_transaction(
+    self: &mut DynamicMultisigAccount,
+    transaction_digest: vector<u8>,
+    ctx: &TxContext
+) {
+    // Get the member who proposed the transaction.
+    let member_address = *self.members().member(ctx.sender()).addr();
```
Dkwcs: The same as Dave mentioned 

> This confused me a bit.
> Won't it be cleaner to say
> 
> let member_address = ctx.sender(); 
> assert!(member_address, some_error_msg)

examples/move/dynamic_multisig_account/sources/account.move
TheMrAI: Apologies I messed up my example I meant something like this:
```
let member_address = ctx.sender(); 
assert!(self.members().has_member(member_address), some_error_msg)
```

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,216 @@
...
+/// Approves a proposed transaction.
+public fun approve_transaction(self: &mut DynamicMultisigAccount, transaction_digest: vector<u8>, ctx: &TxContext) {
+    // Get the member who proposed the transaction.
+    let member_address = *self.members().member(ctx.sender()).addr();
+
+    // Get the transaction.
+    let transaction = self.transactions_mut().transaction_mut(transaction_digest);
+
+    // Approve the transaction.
+    transaction.add_approval(member_address);
```
Dkwcs, line 142: Can we calculate that transaction is approved on this stage? Then have additional field like approved: bool for tx data.
Would it be more efficient, just brain food to consider

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,216 @@
...
+    members_weights: vector<u64>,
+    threshold: u64,
+    authenticator: AuthenticatorInfoV1,
+    ctx: &mut TxContext
+) {
+    // Create a `Members` instance.
+    let members = members::create(members_addresses, members_weights);
+
+    // Verify the provided data consistency.
+    verify_threshold(&members, threshold);
```
Dkwcs, line 55: Should this check be before ?

`let members = members::create(members_addresses, members_weights);
`


I guess this verification is unnecessary, because we can easily check it during the `members` creation.
`verify_threshold(&members, threshold);`

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,216 @@
...
+    ctx: &TxContext
+) {
+    // Check that the sender of this transaction is the account.
+    ensure_tx_sender_is_account(self, ctx);
+
+    // Create a `Members` instance.
+    let members = members::create(members_addresses, members_weights);
+
+    // Verify the provided data consistency.
+    verify_threshold(&members, threshold);
```
Dkwcs, line 177: The same - https://github.com/iotaledger/iota/pull/8638/files#r2379207401

examples/move/dynamic_multisig_account/sources/account.move
valeriyr: The `member` function checks this condition inside and returns a reference.
Dereferencing is in this line to make the borrow checker happy.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: In this implementation, `Members` knows nothing about the threshold, so I think checking this condition is outside its responsibility.

examples/move/dynamic_multisig_account/sources/account.move
TheMrAI: What I am trying to say that in this line:
```
let member_address = *self.members().member(ctx.sender()).addr();
```
It isn't entirely clear that this will be the sender address, the same address that I am searching with. It looks like as if we searched for something else and got the address of that.

valeriyr: Yes, because logically it looks like this:
```
let member = self.members().member(ctx.sender());

self.transactions_mut().add_transaction(transaction_digest, *member.addr());
```
The function operates with a member, not the sender, and it is absolutely correct. Theoretically, it could not be the same; this depends on the implementation.

examples/move/dynamic_multisig_account/sources/transactions.move
valeriyr: added a function that removes a transaction
1faa63d76e

examples/move/dynamic_multisig_account/sources/transactions.move
valeriyr: added a function that removes a transaction
https://github.com/iotaledger/iota/commit/1faa63d76e7f0adab9586377798843c805123bad

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: I don't think so, because if a member is compromised, you want to update the members as fast as possible.
We shouldn't wait until the transactions are executed.

examples/move/dynamic_multisig_account/sources/members.move
valeriyr: added here: 1d5bd22ece

examples/move/dynamic_multisig_account/sources/account.move
valeriyr: The discussion is here:
https://github.com/iotaledger/iota/pull/8638#discussion_r2378888066

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
TheMrAI: Hmm, hmm, makes sense. Could you put it in as a comment, because in 2 weeks I won't remember for sure. :)

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: Theoretically, we can; it will be a different implementation where we have a similar problem.
What if a member who approved a transaction is compromised? Should we drop their approval or execute the transaction?

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: The members list knows nothing about the threshold; checking this condition is outside its responsibility.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: Added a comment here: 28a7cc1159
Check, please, if you like it.

valeriyr:
> Not sure you want to write unit tests now or later?

yep, the PR is still a draft.

> I think this could also use the `account_template`.

We can refactor it once the template is ready.

> It will provide read access to the dynamic fields for everyone, which you said isn't needed, but it could be a nice "view" function feature.

I expect the dynamic fields functions will be implemented by the `account_template`, so we don't need to do this for each example.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
Dkwcs: What is the mechanism for checking the approver validity or is it still undefined? I guess it depends on it. I mean if the malicious approver issue still remains in both of the implementations, would it be logically better to check the approval during the each approve? Anyway it’s just for thinking

valeriyr: We need to answer the question:
If it is logically correct to execute a transaction that is approved by an address that is not a member.
It can potentially lead to security issues.

In general, this case looks like a corner one, because in a normal situation, everything should work as expected.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,230 @@
...
+public struct DynamicMultisigAccount has key {
+    id: UID,
+}
+
+// -------------------------------------- Creation --------------------------------------
+
+/// Creates a new `DynamicMultisigAccount` instance as a shared object with the given
+/// members, threshold and authenticator.
+public fun create(
+    members_addresses: vector<address>,
```
miker83z, line 45: Can it be generalized a bit? For instance, have a Member enum (which for instance can be Simple(address) or PubKey(KeyType)); then, move out the authenticate to a dedicated module. 
This would allow to add more authenticate implementations later.

examples/move/dynamic_multisig_account/sources/account.move
```diff
@@ -0,0 +1,230 @@
...
+use dynamic_multisig_account::transactions::{Self, Transactions};
+
+use iota::account::{Self, AuthenticatorInfoV1};
+use iota::auth_context::AuthContext;
+use iota::dynamic_field;
+
+// --------------------------------------- Errors ---------------------------------------
+
+#[error(code = 0)]
+const ETotalMembersWeightLessThenThreshold: vector<u8> = b"The members weight is less then the threshold.";
```
theiari:
```suggestion
const ETotalMembersWeightLessThanThreshold: vector<u8> = b"The members weight is less than the threshold.";
```

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,230 @@
...
+    verify_threshold(&members, threshold);
+
+    // Create a UID for an account object.
+    let mut id = object::new(ctx);
+
+    // Add all the data as dynamic fields.
+    dynamic_field::add(&mut id, members_key(), members);
+    dynamic_field::add(&mut id, threshold_key(), threshold);
+    dynamic_field::add(&mut id, account::authenticator_df_name(), authenticator);
+    dynamic_field::add(&mut id, transactions_key(), transactions::create(ctx));
```
theiari, line 64: Most likely I'm wrong here (or if it's possible to do), but wouldn't be better to have a helper function for this that groups all dynamic field operations together to minimize gas costs?

examples/move/dynamic_multisig_account/sources/transactions.move
```diff
@@ -0,0 +1,91 @@
...
+    self.bag.borrow(digest)
+}
+
+/// Mutably borrows the account transaction with the provided digest.
+public(package) fun transaction_mut(self: &mut Transactions, digest: vector<u8>): &mut Transaction {
+    self.bag.borrow_mut(digest)
+}
+
+/// Adds a new transaction to the account.
+public(package) fun add_transaction(self: &mut Transactions, digest: vector<u8>, member: address) {
```
theiari:
```suggestion
public(package) fun add_transaction(self: &mut Transactions, digest: vector<u8>, proposer: address) {
```

Nothing fancy but I assume the first member to add a transaction is also the proposer of it (?)

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: What is the reason to add this to an example?
We don't create a dApp here, so extra complexity is not really required.
This example shows the feasibility of implementing the concept; any additional abstraction adds complexity that makes it harder to understand the initial idea.

valeriyr: And adds the effort that we need to put into the related tests.

examples/move/dynamic_multisig_account/sources/transactions.move
valeriyr: It would be beneficial if the proposer were represented as a Member structure instance instead of an address.
I wanted to emphasize that the proposer is a member.

examples/move/dynamic_multisig_account/sources/account.move
valeriyr: fixed in c8046a4d3b

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: This is theoretically possible if we add such a function to the framework.
We can consider this idea to be implemented in a dedicated PR.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
TheMrAI: Yes, I would wait with any potential additional abstractions as well. Let us see first how all the examples look like, and then thonk about it.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
TheMrAI: But wouldn't that be something that the compiler/optimizer should be able to do instead?

TheMrAI (approved):
Very nice. Stealing the macro idea from the tests!

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: I am not sure that I caught your idea, but the compiler knows nothing about the logic behind native calls.
Without such a context, it is impossible to optimize something.

miker83z (approved):

PR #8646, opened 2025-09-24
feat(wallet): add restore passkey account flow

### LINKED ISSUE #8609: Investigate and implement account restoring flow
## Description

it is possible to restore passkey account by signing the test message twice and use findCommonPublicKey to deduce the keypari to sign the transaction (look at the hook `useRestoreWallet`).

- figure out what causes the account to be out of sync and needs to be restored
- implement triggering restore flow
- the restored account needs to be able to use standard signer after being restored

brancoder:
# Description of change

resolves [#8609 ](https://github.com/iotaledger/iota/issues/8609)


VmMad (changes requested):
We decided to remove the passkey context so this PR will probably change

VmMad (approved):
lgtm

PR #8660, opened 2025-09-26
fix(node): Fix node peer discovery

muXxer:
# Description of change

There are two problems in the node discovery:
1. Only the first 200 entries of a `GetKnownPeers` response are evaluated, but the `own_info` of a node is pushed as the last element, resulting in it being cut from the updates of other nodes. Nodes are never able to update their own info in the whole network if there are more than 200 known peers.
2. New entries are not checked for valid addresses, so the list might contain old entries.

This PR adds a check for valid addresses, before adding the peers to the known peers list.

### DIFF
--- a/crates/iota-network/src/discovery/mod.rs
+++ b/crates/iota-network/src/discovery/mod.rs
@@ -802,8 +802,9 @@ fn verify_peer_infos(
                             true // Cooldown expired
                         } else {
                             debug!(
-                                "Peer {} is in verification failure cooldown (failed {:.1}s ago, cooldown: {:.1}s)",
+                                "Peer {} ({}) is in verification failure cooldown (failed {:.1}s ago, cooldown: {:.1}s)",
                                 peer_info.peer_id,
+                                peer_info.addresses.iter().map(|a| a.to_string()).collect::<Vec<_>>().join(", "),
                                 time_since_failure.as_secs_f32(),
                                 cooldown_duration.as_secs_f32()
                             );
@@ -905,7 +906,7 @@ async fn verify_addresses_of_peers(
     config: &DiscoveryConfig,
 ) -> (
     Vec<(VerifiedEnvelope<NodeInfo, Ed25519Signature>, Vec<Multiaddr>)>,
-    Vec<PeerId>,
+    Vec<NodeInfo>,
 ) {
     let peers_count = peers.len();
     let verification_stream = futures::stream::iter(peers.into_iter().map(|verified_peer_info| {
@@ -952,10 +953,16 @@ async fn verify_addresses_of_peers(
     for (verified_peer_info, verified_addresses) in address_verification_results {
         if verified_addresses.is_empty() {
             info!(
-                "Rejecting peer {} due to failed address verification for all addresses",
-                verified_peer_info.peer_id
+                "Rejecting peer {} ({}) due to failed address verification for all addresses",
+                verified_peer_info.peer_id,
+                verified_peer_info
+                    .addresses
+                    .iter()
+                    .map(|a| a.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ")
             );
-            failed_peers.push(verified_peer_info.peer_id);
+            failed_peers.push(verified_peer_info.data().to_owned());
         } else {
             verified_peers.push((verified_peer_info, verified_addresses));
         }
@@ -1052,7 +1059,7 @@ async fn update_known_peers(
     }
 
     // Verify addresses for peers that need verification
-    let (mut verified_peers, failed_peer_ids) = match peers_with_addresses_to_verify.is_empty() {
+    let (mut verified_peers, failed_peer_infos) = match peers_with_addresses_to_verify.is_empty() {
         false => verify_addresses_of_peers(network, peers_with_addresses_to_verify, config).await,
         true => Default::default(),
     };
@@ -1061,7 +1068,7 @@ async fn update_known_peers(
     // peers, we're done
     if verified_peers.is_empty()
         && peers_to_update_directly.is_empty()
-        && failed_peer_ids.is_empty()
+        && failed_peer_infos.is_empty()
     {
         return;
     }
@@ -1097,13 +1104,19 @@ async fn update_known_peers(
 
         // Add all failed peers to verification failure cooldown (if enabled)
         if config.is_address_verification_cooldown_enabled() {
-            for peer_id in failed_peer_ids {
+            for peer_info in failed_peer_infos {
                 state_guard
                     .address_verification_cooldown
-                    .insert(peer_id, now_instant);
+                    .insert(peer_info.peer_id, now_instant);
                 debug!(
-                    "Added peer {} to verification failure cooldown for {:.1}s",
-                    peer_id,
+                    "Added peer {} ({}) to verification failure cooldown for {:.1}s",
+                    peer_info.peer_id,
+                    peer_info
+                        .addresses
+                        .iter()
+                        .map(|a| a.to_string())
+                        .collect::<Vec<_>>()
+                        .join(", "),
                     config.address_verification_failure_cooldown().as_secs_f32()
                 );
             }
--- a/crates/iota-network/src/discovery/mod.rs
+++ b/crates/iota-network/src/discovery/mod.rs
@@ -5,7 +5,7 @@
 use std::{
     collections::{HashMap, HashSet, hash_map::Entry},
     sync::{Arc, RwLock},
-    time::{Duration, Instant},
+    time::Instant,
 };
 
 use anemo::{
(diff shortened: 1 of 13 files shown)


semenov-vladyslav:
The purpose of peer discovery is to keep the network connected even in the presence of malicious nodes.

The problem there is no reliable way to distinguish malicious and honest peers. Checking for connectivity is not very reliable: a malicious peer can listen to multiple ports on the same ip address pretending to be honest, and an honest peer can go offline because of external factors (power outage). Thus we define the main **requirement**: the peer discovery protocol must ensure **connectedness** of honest peers with high probability in presence of (minority of) malicious nodes (ie. malicious known peers do not propagate entirely).

## Peer Ratings

To achieve this the following improvements seem reasonable:
- introduce the notion of peer rating:
  - the higher the rating of a source peer, the higher the rating of its known peers,
  - the higher the rating of a known peer, the higher probability it will be selected for connection,
- ratings of source and known peers are separate and aggregate a number of factors;
- rating factors of a source peer:
  - validity of known peers message (too large response, duplicates, validity of signatures, invalid addresses, etc.),
  - the number/ration of known peers advertised by the source peer,
  - the history of connections to known peers advertised by the source peer,
- rating factors of a known peer:
  - connection metrics (success rate or number of connect failures, rtt, duration) (if there was an attempt to connect to it),
  - the number of times in was advertised by different peers taking into account rating of those peers,
  - freshness of its certificate,
- when selecting a source peer to query known peers its rating must be taken into account,
- when selecting a known peer to connect its rating must be taken into account,
- the rules for rating calculation should be determined empirically.

### Intuition behind Rating

- interaction with a peer is probabilistic based on its rating,
- initially a rating is neutral allowing new nodes to connect / be advertised,
- when failures accumulate the rating lowers thus preventing malicious known peers to propagate,
- when successes accumulate the rating grows indicating the node is honest and should be preferred,
- in a fully honest network the ratings should be roughly equal.

### Adversary Models

1. a (single) malicious node "spams" with "fake" known peer certificates (new and random per each query) without running fake nodes, to verify a known peer certificate we would have to connect to the address, otherwise it seems legit;
2. same as in 1., but "fake" known peer ids are persisted;
3. there is a (potentially large, approx 60000) number of malicious nodes, the malicious nodes only accept incoming QUIC connections; the main malicious node runs the peer discovery protocol;
4. same as in 3., but all the malicious nodes run the peer discovery protocol and advertise (a selected group of) other malicious nodes;
5. same as in 4., but first the malicious node behaves as an honest one earning a high rating and only then it starts behaving maliciously.

### Network Liveness Metric

1. metric per node: ratio of connected honest nodes to connected malicious nodes (a "healthy" node metric should be greater than a threshold) -- the node should eventually be "healthy" with high probability;
2. network overall metric: distribution of per node metrics (number of "healthy" nodes should be greater than a threshold) -- the network should be "healthy" with high probability;
3. new honest node:
  - in a running network (with malicious nodes and already populated known peer tables) an honest node tries to connect;
  - the node becomes "healthy" once it is connected to a (threshold) number of honest nodes;
  - estimate the distribution (mean value, variance) of time period since this new node becomes online until it is "healthy";
  - the mean value and the variance should be reasonably low.

The peer discovery protocol is **robust** if the three metric estimates are admissible.

## Peer Discovery in Anemo

On another note, it is worth considering embedding the implementation of peer discovery into `anemo` crate itself:
- anemo already maintains structures and reactor (ConnectionManager) used in peer discovery: known peers, connected peers, certificate verifier, async event handling loop, ; those should only be extended for peer discovery needs;
- anemo network layer operates upon QUIC protocol which uses TLS for authentication; since there is no way to introduce a rigid notion of trust in a decentralized network, the rating system proposed above can serve as its replacement;
- TLS certificates can serve as known peers certificates: PeerId is SubjectPublicKey, Address is SubjectAltName extension, timestamp is Validity period;
- there should be a two step certificate verification process:
  - preliminary -- without connecting to the peer (can't verify ownership of a network address);
  - TLS/QUIC -- verification during connecting to the peer (verifies ownership of a network address); TLS protocol already deals with verifying certificates (including validating Subject Alt Name extension);
- anemo does not expose the address of a connected peer; since multiple addresses may be controlled by the same (malicious) peer, the current limitation is quite inconvenient while keeping track of known peer certificate validity.

jkrvivian (approved):

kodemartin (approved):

crates/iota-network/src/discovery/mod.rs
```diff
@@ -503,15 +568,58 @@ async fn try_to_connect_to_seed_peers(
     .await;
 }
 
+/// Deduplicates peer infos based on their serialized representation (including
+/// signature). First groups by PeerId for efficiency, then uses BCS
+/// serialization only when there are multiple entries for the same peer.
+fn deduplicate_peers(peers: Vec<SignedNodeInfo>) -> Vec<SignedNodeInfo> {
```
semenov-vladyslav: Why not just put all SignedNodeInfo from Vec into a HashSet? We can add a wrapper for SignedNodeInfo implementing Hash. This whole function will be just `peers.into_iter().collect()`.

crates/iota-network/src/discovery/mod.rs
```diff
@@ -528,8 +636,24 @@ async fn query_peer_for_their_known_peers(
...
         );
+
     if let Some(found_peers) = found_peers {
-        update_known_peers(state, metrics, found_peers, allowlisted_peers);
+        // Limit each client to MAX_PEERS_TO_SEND (+1 because of the own_info from the
+        // peer)
+        let limited_peers = found_peers
+            .into_iter()
+            .take(MAX_PEERS_TO_SEND + 1)
+            .collect::<Vec<_>>();
```
semenov-vladyslav:
```suggestion
        found_peers.truncate(MAX_PEERS_TO_SEND + 1);
```

crates/iota-network/src/discovery/mod.rs
```diff
@@ -569,7 +695,12 @@ async fn query_connected_peers_for_their_known_peers(
                         if !own_info.addresses.is_empty() {
                             known_peers.push(own_info)
                         }
+                        // Limit each client to MAX_PEERS_TO_SEND (+1 because of the own_info from
+                        // the peer)
                         known_peers
+                            .into_iter()
```
semenov-vladyslav: There is a better way to `.truncate()` a Vec.

crates/iota-network/src/discovery/mod.rs
```diff
@@ -569,7 +695,12 @@ async fn query_connected_peers_for_their_known_peers(
                         if !own_info.addresses.is_empty() {
                             known_peers.push(own_info)
                         }
+                        // Limit each client to MAX_PEERS_TO_SEND (+1 because of the own_info from
```
semenov-vladyslav: Should we truncate known_peers before pushing own_info? Otherwise we may lose own_info, although we do not handle it in any way different from other known_peers.

crates/iota-network/src/discovery/mod.rs
```diff
@@ -451,6 +456,67 @@ impl DiscoveryEventLoop {
...
+        .map(|address| {
+            let network = network.clone();
+            let peer_id = peer_info.peer_id;
+            let address = address.clone();
+            Box::pin(async move {
+                if let Ok(anemo_address) = address.to_anemo_address() {
+                    // Try to connect to the claimed address with the claimed peer_id
+                    match tokio::time::timeout(
+                        config.address_verification_timeout(),
+                        network.connect_with_peer_id(anemo_address, peer_id),
```
semenov-vladyslav, line 494: We really need to be careful here. What happens if we are already connected (or almost connected, since the process of connecting in anemo network is asynchronous) to this peer (a peer with `peer_id`)? There will be `simultaneous_dial_tie_breaking`. So only one connection will survive.

crates/iota-network/src/discovery/mod.rs
```diff
@@ -451,6 +456,67 @@ impl DiscoveryEventLoop {
...
+                    .await
+                    {
+                        Ok(Ok(_connection)) => {
+                            debug!(
+                                "Address verification succeeded for peer {} at address {}",
+                                peer_id, address
+                            );
+
+                            // Disconnect immediately since this was just for verification
+                            let _ = network.disconnect(peer_id);
```
semenov-vladyslav, line 505: And here we close the remaining connection which may have been initiated earlier. This is not highly probable, but may result in spurious misterious connection drops.

crates/iota-network/src/discovery/mod.rs
```diff
@@ -650,28 +878,343 @@ fn update_known_peers(
...
+    state: Arc<RwLock<State>>,
+    metrics: Metrics,
+    found_peers: Vec<SignedNodeInfo>,
+    allowlisted_peers: Arc<HashMap<PeerId, Option<Multiaddr>>>,
+    config: &DiscoveryConfig,
+) {
+    let now_instant = Instant::now();
+
+    // Clean up old entries from the verification failure cooldown
+    cleanup_verification_failure_cooldown(&state, now_instant, config);
```
semenov-vladyslav: Should it be called here, or in its own loop?

crates/iota-network/src/discovery/mod.rs
```diff
@@ -619,20 +843,24 @@ fn update_known_peers(
...
+        // Verify that all addresses provided are valid anemo addresses
         if !peer_info
             .addresses
             .iter()
             .all(|addr| addr.len() < MAX_ADDRESS_LENGTH && addr.to_anemo_address().is_ok())
         {
             continue;
         }
+
+        // Verify the signature
```
semenov-vladyslav, line 867: I see no reason for `debug_fatal!` two lines below. A malicious node can just span with invalid signatures. A node built in debug mode will crash, a release mode node will report system_invariant_violation although there's no violations here, we just normally verify the public key, and if it's invalid we just skip peer_info, no violation.

semenov-vladyslav (commented):
Overall looks good except for minor comments and a suspicious behaviour during address ownership verification that may result in legit connection drop.

crates/iota-network/src/discovery/mod.rs
muXxer: Yeah I was thinking about the same line. I thought it's impossible to reach this `debug_fatal!` (this comes from upstream), but I guess someone can send signatures that can't be passed. I will make a `debug!` out of it.

crates/iota-network/src/discovery/mod.rs
muXxer: I thought the check for equal peer IDs first might be faster, but I guess you are right. It's simpler and maybe equal efficiency without all those allocations etc.

PR #8679, opened 2025-09-29
chore(docs): extend the wiki to include the real world examples for PA

itsyaasir:
This pull request adds four comprehensive, real-world examples for IOTA hierarchies and IOTA notarization. 

The use cases are as follows. For IOTA hierarchies, we have the first one: a real-world example to showcase university degrees, and the other is Supply Chain Certification.

 For IOTA Notarization, we have two examples: uses of IoT weather stations and legal contracts


eike-hass:
Great content! Please give a quick overview of the changes in the PR description 🙏

eike-hass:
@iotaledger/devrel

Dr-Electron (approved):
This are not really how-tos. But full tutorials. In any case let's merge this we can move it later (maybe in my restructure anyway 👍 )
Other than that: Great stuff

Ginowine (approved):
Thanks for putting this together.

vivekjain23 (approved):
LGTM!

As discussed earlier with the PA team on IOTA notarization real-world examples, I think adding a TypeScript (Web) / WASM-based example here would make the docs even more practical.

We can either request a tested Web example from the PA team, or I can contribute one I already have a working draft from the workshop notarization DApp work. I had faced a WASM initialization timing issue earlier, but handled it with a temp “init earlier” fix, and this may already be properly resolved now or a suitable package.

Overall, this looks good to merge

eike-hass (approved):

PR #8685, opened 2025-09-29
fix(cli): Remove telemetry subscriber for the analyzer command

### LINKED ISSUE #6263: Dangling move analyzer instances
## Bug description

I closed my VS Code and while looking at my system stats I found 3 dangling instances of `iota` running.

```
ps aux | grep iota
alexsporn        96788   0.0  2.9 417960016 1482080   ??  S     9:16AM   1:54.24 iota analyzer
alexsporn        92108   0.0  0.0 416008112  23456   ??  S     8:49AM   0:31.61 iota analyzer
alexsporn        91120   0.0  0.0 414669664  19072   ??  S     8:42AM   0:12.15 iota analyzer
```

These instances are also consuming a lot of RAM

Image

I took a sample of the process and it seems the analyzer is deadlocked while printing a panic.

```
    2389 Thread_2038770   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2389 start  (in dyld) + 6000  [0x18c226b4c]
    +   2389 main  (in iota) + 52  [0x1004ac28c]
    +     2389 std::rt::lang_start_internal::hd55fd6ef897020fb  (in iota) + 1092  [0x10530bfc8]
    +       2389 std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h1721b625deb171ac  (in iota) + 16  [0x1007ca734]
    +         2389 std::sys::backtrace::__rust_begin_short_backtrace::h795a12ddad124717  (in iota) + 12  [0x100679a34]
    +           2389 iota::main::hc280c66313b9039f  (in iota) + 188  [0x1004abfb8]
    +             2389 tokio::runtime::runtime::Runtime::block_on::h13ed94451b6db699  (in iota) + 604  [0x10055fa54]
    +               2389 tokio::runtime::park::CachedParkThread::block_on::hab2e9868feb0c229  (in iota) + 188  [0x1004dceec]
    +                 2389 iota::main::_$u7b$$u7b$closure$u7d$$u7d$::h7988f037ab449000  (in iota) + 968  [0x1002ec020]
    +                   2389 iota::iota_commands::IotaCommand::execute::_$u7b$$u7b$closure$u7d$$u7d$::hf2e763a93018326e  (in iota) + 4828  [0x100299710]
    +                     2389 move_analyzer::analyzer::run::h660f4f2074963c40  (in iota) + 9592  [0x100e1206c]
    +                       2389 std::io::stdio::_eprint::h977fa9d40c2ea967  (in iota) + 172  [0x105314674]
    +                         2389 core::panicking::panic_fmt::ha0b5c5252c2f1706  (in iota) + 32  [0x10544deb8]
    +                           2389 rust_begin_unwind  (in iota) + 32  [0x10531cdfc]
    +                             2389 std::sys::backtrace::__rust_end_short_backtrace::h25049b070cf1a527  (in iota) + 12  [0x10531c1c0]
    +                               2389 std::panicking::begin_panic_handler::_$u7b$$u7b$closure$u7d$$u7d$::h0edcf2e5c8ce9186  (in iota) + 144  [0x10531d144]
    +                                 2389 std::panicking::rust_panic_with_hook::ha68643d4311595e9  (in iota) + 740  [0x10531d5dc]
    +                                   2389 telemetry_subscribers::set_panic_hook::_$u7b$$u7b$closure$u7d$$u7d$::h563c99821663c7ba  (in iota) + 600  [0x1033bd738]
    +                                     2389 _$LT$std..io..stdio..Stdout$u20$as$u20$std..io..Write$GT$::flush::h6f5bc7209bc78883  (in iota) + 24  [0x1053129d8]
    +                                       2389 _$LT$$RF$std..io..stdio..Stdout$u20$as$u20$std..io..Write$GT$::flush::ha7709c257fc5bbe9  (in iota) + 20  [0x105312c30]
    +                                         2389 std::io::stdio::Stderr::lock::h423c69b101e36e45  (in iota) + 160  [0x10531376c]
    +                                           2389 _pthread_mutex_firstfit_lock_slow  (in libsystem_pthread.dylib) + 220  [0x18c5c0840]
    +                                             2389 _pthread_mutex_firstfit_lock_wait  (in libsystem_pthread.dylib) + 84  [0x18c5c2e58]
    +                                               2389 __psynch_mutexwait  (in libsystem_kernel.dylib) + 8  [0x18c58689c]
```

Attached is the full sample.

[sample.txt](https://github.com/user-attachments/files/19601859/sample.txt)

miker83z:
# Description of change

Before analyzing the fix we need a bit of background:
- The [IOTA Move VS Code](https://marketplace.visualstudio.com/items?itemName=iotaledger.iota-move) extension (current ver. 1.2.0) always tries to install a binary when launched/started. 
-  During this installation, it can pick one of these 3 cases:
    1. if the `iota-move.force-bundled` flag is set in the VS code settings 
        - -> force install the `move-analyzer` bin bundled with the extension in the ` ~/.iota/bin/` folder 
        - -> USE `~/.iota/bin/move-analyzer` as the command to run the [LSP server](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification);
    2. else, if the version of `~/.iota/bin/move-analyzer` is `>` than the version of `iota` 
        - -> USE `~/.iota/bin/move-analyzer` as the command to run the LSP server
    3. else -> USE `iota analyzer` as the command to run the LSP server
- The issue described in #6263, i.e., dangling `iota` processes, happens only in the case iii., that is when the `iota analyzer` command is used to run a LSP server.
- Cases i. and ii. are not affected by this. In fact, during tests, `move-analyzer` processes were never left dangling.

Based on the background information, we know that a comparison between the command `move-analyzer` and `iota analyzer` should lead to the same result; however, an experiment shows different results:
commands comparison
- After running both commands, a server is started.
- A way to interact with the server is to hit "Enter"; this will force the initialization of a connection, but it will make the server panic because there is no actual client trying to connect.
- In the case of the `move-analyzer` command, the main thread is panicking and the process terminates.
- In the case of the `iota analyzer` command, the main thread is panicking and but the process does NOT terminate (I needed to force quit with CTRL+C).

Now back to the main issue:
- When a LSP client sends a stop signal to the [analyzer server](https://github.com/iotaledger/iota/blob/9ff49c6d4e85030c06a8f0127d8fb51f374d0564/external-crates/move/crates/move-analyzer/src/analyzer.rs#L342), the `run` is terminated.
- With another experiment, I checked that this signal reaches the server with no issues and no error are returned back, meaning that the `run` method actually terminates.
- So, it seems that, when this `analyzer::run` is invoked as an iota command [here](https://github.com/iotaledger/iota/blob/9ff49c6d4e85030c06a8f0127d8fb51f374d0564/crates/iota/src/iota_commands.rs#L601), it gets stuck. Whilst, when it is invoked in the [`move-analyzer` bin](https://github.com/iotaledger/iota/blob/9ff49c6d4e85030c06a8f0127d8fb51f374d0564/external-crates/move/crates/move-analyzer/src/bin/move-analyzer.rs#L19) it goes through. 

Solution:
- using `std::panic::set_hook` in the iota command matched arm fixes the issue
- @iotaledger/dev-tools might have a more efficient solution though, or a better explanation about the different behaviors

(additionally, this is broken also in Sui:)
sui

Fixes #6263 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Resolved a dangling IOTA process issue in the IOTA Move VS Code extension by removing the telemetry subscriber from the analyzer CLT command. For affected users, please update the installed `iota` binary on your system to `v1.9.x` and above.
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota/src/iota_commands.rs
+++ b/crates/iota/src/iota_commands.rs
@@ -598,6 +598,10 @@ impl IotaCommand {
             }
             IotaCommand::FireDrill { fire_drill } => run_fire_drill(fire_drill).await,
             IotaCommand::Analyzer => {
+                std::panic::set_hook(Box::new(|panic_info| {
+                    eprintln!("Fatal panic: {panic_info}");
+                    std::process::exit(1);
+                }));
                 analyzer::run(implicit_deps(latest_system_packages()));
                 Ok(())
             }
--- a/crates/iota/src/iota_commands.rs
+++ b/crates/iota/src/iota_commands.rs
@@ -598,10 +598,6 @@ impl IotaCommand {
             }
             IotaCommand::FireDrill { fire_drill } => run_fire_drill(fire_drill).await,
             IotaCommand::Analyzer => {
-                std::panic::set_hook(Box::new(|panic_info| {
-                    eprintln!("Fatal panic: {panic_info}");
-                    std::process::exit(1);
-                }));
                 analyzer::run(implicit_deps(latest_system_packages()));
                 Ok(())
             }
--- a/crates/iota/src/main.rs
+++ b/crates/iota/src/main.rs
@@ -34,19 +34,21 @@ async fn main() {
 
     let args = Args::parse();
     let _guard = match args.command {
-        IotaCommand::KeyTool { .. } | IotaCommand::Move { .. } => {
+        IotaCommand::KeyTool { .. } | IotaCommand::Move { .. } => Some(
             telemetry_subscribers::TelemetryConfig::new()
                 .with_log_level("error")
                 .with_env()
-                .init()
-        }
+                .init(),
+        ),
         IotaCommand::Client {
             cmd: Some(ReplayBatch { .. }),
             ..
-        } => telemetry_subscribers::TelemetryConfig::new()
-            .with_log_level("info")
-            .with_env()
-            .init(),
+        } => Some(
+            telemetry_subscribers::TelemetryConfig::new()
+                .with_log_level("info")
+                .with_env()
+                .init(),
+        ),
 
         IotaCommand::Client {
             cmd: Some(ReplayTransaction {
@@ -63,25 +65,32 @@ async fn main() {
             if ptb_info {
                 config = config.with_trace_target("replay_ptb_info");
             }
-            config.init()
+            Some(config.init())
         }
         IotaCommand::Client {
             cmd: Some(ProfileTransaction { .. }),
             ..
         } => {
             // enable full logging for ProfileTransaction and ReplayTransaction
-            telemetry_subscribers::TelemetryConfig::new()
-                .with_env()
-                .init()
+            Some(
+                telemetry_subscribers::TelemetryConfig::new()
+                    .with_env()
+                    .init(),
+            )
         }
-        IotaCommand::Start { .. } => telemetry_subscribers::TelemetryConfig::new()
-            .with_log_level("info")
-            .with_env()
-            .init(),
-        _ => telemetry_subscribers::TelemetryConfig::new()
-            .with_log_level("error")
-            .with_env()
-            .init(),
+        IotaCommand::Start { .. } => Some(
+            telemetry_subscribers::TelemetryConfig::new()
+                .with_log_level("info")
+                .with_env()
+                .init(),
+        ),
+        IotaCommand::Analyzer { .. } => None,
+        _ => Some(
+            telemetry_subscribers::TelemetryConfig::new()
+                .with_log_level("error")
+                .with_env()
+                .init(),
+        ),
     };
     debug!("IOTA CLI version: {VERSION}");
     exit_main!(args.command.execute().await);
--- a/crates/iota/src/main.rs
+++ b/crates/iota/src/main.rs
@@ -84,7 +84,7 @@ async fn main() {
                 .with_env()
                 .init(),
         ),
-        IotaCommand::Analyzer { .. } => None,
+        IotaCommand::Analyzer => None,
         _ => Some(
             telemetry_subscribers::TelemetryConfig::new()
                 .with_log_level("error")


Thoralf-M (approved):

crates/iota/src/iota_commands.rs
```diff
@@ -598,6 +598,10 @@ impl IotaCommand {
             }
             IotaCommand::FireDrill { fire_drill } => run_fire_drill(fire_drill).await,
             IotaCommand::Analyzer => {
+                std::panic::set_hook(Box::new(|panic_info| {
+                    eprintln!("Fatal panic: {panic_info}");
```
TheMrAI: A panic can generally be considered fatal. Except the very special circumstances where abort mechanism is set to [stack unwinding and we are expecting panics that can do](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html).
The panic hook takes effect before the [panic runtime](https://doc.rust-lang.org/std/panic/fn.set_hook.html), there are two runtimes aborting or unwinding, both will result in the thread of execution being terminated.
So exiting inside the panic hook may solve the problem, but the real issue is somewhere else.

Is there a unwind catch set in this 3rd type of execution? How is the thread of execution dispatched? Does anybody watch out for what happened with the thread?

valeriyr: Looks like a problem is in the telemetry subscriber that is initialed [here](https://github.com/iotaledger/iota/blob/47f48427556ee1b7b3f5ef3483dcf44e9d3e454b/crates/iota/src/main.rs#L81).

It sets a panic hook to print information:
https://github.com/iotaledger/iota/blob/47f48427556ee1b7b3f5ef3483dcf44e9d3e454b/crates/telemetry-subscribers/src/lib.rs#L215
and stucks on these lines:
https://github.com/iotaledger/iota/blob/47f48427556ee1b7b3f5ef3483dcf44e9d3e454b/crates/telemetry-subscribers/src/lib.rs#L235-L236

@miker83z, can you please check this on your side?

valeriyr: As I see, `std::io::stdout()` is locked by the `LspServerWriter` thread.

```rust
pub(crate) fn stdio_transport() -> (Sender<Message>, Receiver<Message>, IoThreads) {
    let (writer_sender, writer_receiver) = bounded::<Message>(0);
    let writer = thread::Builder::new()
        .name("LspServerWriter".to_owned())
        .spawn(move || {
            let stdout = stdout();
            let mut stdout = stdout.lock();
            writer_receiver.into_iter().try_for_each(|it| it.write(&mut stdout))
        })
        .unwrap();
    ...
}
```
The code is taken from here: `lsp-server-0.7.7/src/stdio.rs`

muXxer (approved):

valeriyr (changes requested):
I request changes just to have a discussion if we want to leave this fix.
The reason why the application hangs is described in the comments.
The original fix that is introduced in the PR replaces the telemetry subscriber panic hook with a different one, that is why it works.
Replacing the hook also limits the information that is printed in case of panic.

crates/iota/src/iota_commands.rs
miker83z: Yes, that was it! Thanks @valeriyr! 
[ed3339f](https://github.com/iotaledger/iota/pull/8685/commits/ed3339f628680e9e8e2692a35fffa9b36f01e27e)

miker83z: @iotaledger/dev-tools I removed the telemetry for the analyzer here [ed3339f](https://github.com/iotaledger/iota/pull/8685/commits/ed3339f628680e9e8e2692a35fffa9b36f01e27e) but you should have the final say! We know the main issue now, but the solution could be different!

miker83z: cc @Thoralf-M

valeriyr (approved):

PR #8697, opened 2025-09-30
feat(examples/move): function key plugin implementation for IOTAccount

### LINKED ISSUE #8680: Implement “Function Keys”
Enable accounts to attach per-function authorization policies (“Function Keys”) that allow invoking only specific target functions of on-chain modules.

Dkwcs:
# Description of change

This package adds plugin-based execution feature to an IOTAccount, enabling a per-pubkey allow-set of callable functions identified as `(package, module, function)`. The authenticate entrypoint implements a dual flow:

- Owner flow: if the provided pubkey is the stored owner key, only Ed25519 verification is required (no allow-set or command-count checks).
- User flow: otherwise, after Ed25519 verification, the PTB must contain exactly one MoveCall, and that call must be present in the delegated pubkey’s allow-set.

Fixes #8680 .

Run move tests for `function_key` package.
`iota move test`


examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,122 @@
...
+/// - `EUnauthorized` if the call target isn’t allowed.
+public fun authenticate(
+    account: &IOTAccount,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+) {
+    assert!(store_exists(account), EFunctionKeysNotInitialized);
+
+    authenticate_ed25519(account, signature, auth_ctx, ctx);
```
miker83z: Instead of using a `basic_keyed_account`, you could make it so that the public key is stored in the `FunctionKeysStore`
In [Near's Function-call](https://docs.near.org/protocol/access-keys#function-call-keys) keys we have a store that maps a key to a set of `method_names` (and also `allowance`, but we skip the allowance part for this PR). 

So, here it could be:
```move
public struct FunctionKeysStore has store {
    function_keys: Table<vector<u8>,vector<FunctionKey>>,
}
```
(where the table key is a pub_key)

and instead of using `authenticate_ed25519`, here:
```move
public fun authenticate(
    account: &IOTAccount,
    pub_key: vector<u8>,
    signature: vector<u8>,
    auth_ctx: &AuthContext,
    ctx: &TxContext,
) {
    assert!(store_exists(account), EFunctionKeysNotInitialized);

    
    // Check the signature.
    assert!(
        ed25519::ed25519_verify(&decode(signature), pub_key, ctx.digest()),
        EEd25519VerificationFailed,
    );

    // PTB MUST contain exactly one command.
    assert!(auth_ctx.tx_commands().length() == 1, EInvalidAmountOfCommands);

    let func_key = extract_func_key(&auth_ctx.tx_commands()[0]);
    let store = borrow_store(account);

    // Check that in store, the pub_key exists and then the associated vector value contains func_key 
    assert!(store.is_allowed(&pub_key, &func_key), EUnauthorized);
}
```

Dkwcs: Nice point, I thought about it as well.

TheMrAI: So this would allow setting a set of functions that can only be called using key/signature (how you want to look at it) **A** and another set with **B**?

miker83z: @TheMrAI yes, then the use case would be:
- I create N private keys and then derive N pub keys
-  In the `FunctionKeysStore` I store these N pub keys associated to fn `send_exactly_1_IOTA(recv: address)`
- Off-chain, I send 1 private key to you
- You derive the pub key and the address, then invoke `send_exactly_1_IOTA(addr)` (addr is arbitrary, because you could decide to send that coin to some other addr, we don't need to check that addr comes from pub key)
- send_exactly_1_IOTA removes the pub key from the `FunctionKeysStore` (we don't need to implement this aspect in this PR, it is enough to focus on the authenticate)
- replicate this for N - 1 actors -> basically an Airdrop

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,122 @@
...
+/// DF missing (forgot to `create`).
+const EFunctionKeysNotInitialized: u64 = 2;
+/// PTB does not contain **exactly one** command.
+const EInvalidAmountOfCommands: u64 = 3;
+/// Called function not in the allow-set.
+const EUnauthorized: u64 = 4;
+
+/// Initializes the Function Keys store under the given `account`.
+public fun create(account: &mut IOTAccount, ctx: &mut TxContext) {
+    assert!(!account.has_field(fk_store_key()), EFunctionKeysAlreadyInitialized);
```
TheMrAI: If I understood correctly this is not how we intend to use an IOTAccount.
While this works, it means you can't create an account in one go, properly set up.
An IOTAccount has to be made first, than the account itself has to modify itself to be a function keyed account.

TheMrAI (approved):
Checked the code.
I think the outline is good.
It has some minor points of improvement, but I propose not to waste any time in fixing these. How anything is called, that certain asserts aren't necessary are uninportant.

At this stage we just need the example, we can clean it up later, when we have a clearer outline of the needs.

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,122 @@
...
+// --------------------
+
+/// DF already exists (double init).
+const EFunctionKeysAlreadyInitialized: u64 = 1;
+/// DF missing (forgot to `create`).
+const EFunctionKeysNotInitialized: u64 = 2;
+/// PTB does not contain **exactly one** command.
+const EInvalidAmountOfCommands: u64 = 3;
+/// Called function not in the allow-set.
+const EUnauthorized: u64 = 4;
```
valeriyr: It is better to unify the errors format. I propose to use errors with a description, as it is done in `fk_store.move`.

examples/move/function_keys/sources/function_keys.move
Dkwcs: @miker83z here fix ce8aa3c1e1625f426551265409c3fed9e5a1d5c1

examples/move/function_keys/sources/function_keys.move
valeriyr: It is better to fix this; otherwise, it will be harder to test this example using p2p tests.

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,136 @@
...
+
+#[error(code = 1)]
+const EFunctionKeyDoesNotExist: vector<u8> = b"The function key does not exist";
+
+// =========================
+// Types
+// =========================
+
+/// Dynamic-field name for the Function Keys store inside the `IOTAccount`.
+public struct FunctionKeysName has copy, drop, store {}
```
valeriyr: I propose to move this struct to the `function_keys.move` file and make it possible to create its instance only inside the `function_keys`module.
It makes sense because that logic defines how `FunctionKeysStore` is stored.

Or if you want to leave it here, it is better to make it private by making private `fk_store_key ` and add a function to attach the related dynamic field to IOTAccount.

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,136 @@
...
+public struct FunctionKeysStore has store {
+    function_keys: VecSet<FunctionKey>,
+}
+
+// =========================
+// Accessors / helpers
+// =========================
+
+/// Returns the **dynamic field key** used to access the Function Keys store.
+public fun fk_store_key(): FunctionKeysName { FunctionKeysName {} }
```
valeriyr: This function should be private; otherwise, it will be possible to attach an unexpected field type using this key.

examples/move/function_keys/sources/function_keys.move
Dkwcs: > An IOTAccount has to be made first, then the account itself has to modify itself to be a function keyed account.
Exactly, that is the idea behind, to use the account as a plugins foundation. Account has to be created in advance and then can be attached by different “plugins”.

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,122 @@
...
+    assert!(store_exists(account), EFunctionKeysNotInitialized);
+
+    authenticate_ed25519(account, signature, auth_ctx, ctx);
+
+    // PTB MUST contain exactly one command.
+    assert!(auth_ctx.tx_commands().length() == 1, EInvalidAmountOfCommands);
+
+    let func_key = extract_func_key(&auth_ctx.tx_commands()[0]);
+    let store = borrow_store(account);
+    assert!(store.is_allowed(&func_key), EUnauthorized);
```
valeriyr: As I see, an account is created with an empty store, so my question here is how we add the first permission to make it possible to grant or revoke permissions? Since this can be done only by the account itself.

examples/move/function_keys/sources/function_keys.move
valeriyr: I do not insist; we can leave it as is

examples/move/function_keys/sources/function_keys.move
Dkwcs: If get you right:
only the owner of account can grant or revoke permissions, later it allows to authenticate with `public_key` passed in public` fun authenticate(..) {}`

examples/move/function_keys/sources/fk_store.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
Dkwcs: I would prefer second option
> Or if you want to leave it here, it is better to make it private by making private fk_store_key  and add a function to attach this dynamic field to IOTAccount.

examples/move/function_keys/sources/function_keys.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
miker83z: I also vote for the second option.

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,131 @@
...
+
+/// Grants (allows) a `FunctionKey` under a specific `pub_key`.
+/// - Only the account owner may mutate their DF.
+public fun grant_permission(
+    account: &mut IOTAccount,
+    pub_key: vector<u8>,
+    func_key: FunctionKey,
+    ctx: &mut TxContext,
+) {
+    account.ensure_tx_sender_is_account(ctx);
```
miker83z: This will always fail

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,131 @@
...
+///    against the account’s stored public key and the canonical `ctx.digest()`.
+/// 2. Check `auth_ctx.tx_commands().length() == 1` to enforce a single operation PTB.
+/// 3. Convert the sole `Command` into a `FunctionKey` via `extract_func_key(...)`.
+/// 4. Assert the key exists in the account’s allow-set.
+///
+/// Errors:
+/// - `EFunctionKeysNotInitialized` if `create` has not been called.
+/// - `EInvalidAmountOfCommands` if PTB has 0 or > 1 commands.
+/// - `EUnauthorized` if the call target isn’t allowed.
+public fun authenticate(
```
miker83z, line 134: This authenticate function should have two flows:
- One flow is dedicated to the "real" account owner, and must allow to authenticate using an `OwnerPubKey` or similar and no functions checks. This is needed to fix this issue -> https://github.com/iotaledger/iota/pull/8697/files#r2419415087
- A second flow, enacted when the first one fails, is what we currently have in the function body. This allows several users to authenticate if they use a function in the PTB.

A more sophisticated way to implement this could be done, but it is not needed for this PR.

miker83z: Also, add tests for both flows

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,131 @@
...
+/// Called function not in the allow-set.
+#[error(code = 2)]
+const EUnauthorized: vector<u8> = b"Function key is not the allowed set";
+/// Ed225519 verification has failed.
+#[error(code = 3)]
+const EEd25519VerificationFailed: vector<u8> = b"Ed25519 verification has failed";
+
+
+/// Initializes the Function Keys store under the given `account`.
+public fun create(account: &mut IOTAccount, ctx: &mut TxContext) {
```
miker83z: As discussed before, this should be called `attach` and left here as a demonstration. 

Then, add a new function `create` which uses `IOTAccountBuilder`. Only use this new function in tests.

examples/move/function_keys/Move.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "function_keys"
+edition = "2024" # edition = "legacy" to use legacy (pre-2024) Move
+# license = ""           # e.g., "MIT", "GPL", "Apache 2.0"
```
miker83z: Pls remove all comments, it's cleaner

examples/move/function_keys/Move.toml
```diff
@@ -0,0 +1,26 @@
...
+# authors = ["..."]      # e.g., ["Joe Smith (joesmith@noemail.com)", "John Snow (johnsnow@noemail.com)"]
+
+[dependencies]
+Iota = { local = "../../../crates/iota-framework/packages/iota-framework" }
+iotaccount = { local = "../../move/iotaccount" }
```
miker83z: This should work as well
```suggestion
iotaccount = { local = "../iotaccount" }
```

miker83z (changes requested):

examples/move/function_keys/sources/function_keys.move
Dkwcs: Done, so actually as I see that the non-owner can’t grant/revoke permissions because of this flow:
` let fk_store = borrow_fk_store_mut(account, ctx)` - >  `account.borrow_field(fk_store_key())` -> 
 ```
public fun borrow_field_mut<Name: copy + drop + store, Value: store>(
    self: &mut IOTAccount,
    name: Name,
    ctx: &TxContext,
): &mut Value {
    // Check that the sender of this transaction is the account.
    ensure_tx_sender_is_account(self, ctx);  /// this field is preventing from non-owner access 

    // Borrow the related dynamic field.
    dynamic_field::borrow_mut(&mut self.id, name)
}
```

Anyway I’ve separated auth flow into 2 owner/‘user’ and added additional tests

File: examples/move/function_keys/sources/function_keys.move (hunk 9 above)
Comment (Dkwcs) [reply examples/move/function_keys/sources/function_keys.move]: Done

File: examples/move/function_keys/Move.toml (hunk 10 above)
Comment (Dkwcs) [reply examples/move/function_keys/Move.toml]: Done

File: examples/move/function_keys/Move.toml (hunk 11 above)
Comment (Dkwcs) [reply examples/move/function_keys/Move.toml]: Done

File: examples/move/function_keys/sources/function_keys.move [hunk 12]
```diff
@@ -0,0 +1,166 @@
...
+
+/// Initializes the Function Keys store under the given `account`.
+public fun attach(account: &mut IOTAccount, ctx: &mut TxContext) { attach_fk_store(account, ctx); }
+
+/// Creates a new `IOTAccount` as a shared object with the given authenticator.
+public fun create(public_key: vector<u8>, authenticator: AuthenticatorInfoV1, ctx: &mut TxContext) {
+    let account = builder(authenticator, ctx)
+        .add_dynamic_field(OwnerPublicKey {}, public_key)
+        .finish();
+    account.share();
```
miker83z, line 69: During the create, the fk_store should be attached as well, otherwise the account is not complete.
```suggestion
    attach_fk_store(account, ctx);
    account.share();
```

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,140 @@
...
+    FunctionKey { package, module_name, function_name }
+}
+
+// =========================
+// Per-pubkey allow-set ops
+// =========================
+
+/// Ensure a VecSet exists for `pub_key`; if absent, create an empty set.
+/// Returns a &mut to the set.
+public fun ensure_key_entry(
```
miker83z: I don't think it's necessary to have this public.
```suggestion
fun ensure_key_entry(
```

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,140 @@
...
+    pub_key: vector<u8>,
+): &mut VecSet<FunctionKey> {
+    if (!tbl::contains(&store.function_keys, pub_key)) {
+        tbl::add(&mut store.function_keys, pub_key, vec_set::empty());
+    };
+    tbl::borrow_mut(&mut store.function_keys, pub_key)
+}
+
+/// **Allow** a function key for a specific public key.
+public fun allow(store: &mut FunctionKeysStore, pub_key: vector<u8>, fk: FunctionKey) {
```
miker83z: public(package) should be enough
```suggestion
public(package) fun allow(store: &mut FunctionKeysStore, pub_key: vector<u8>, fk: FunctionKey) {
```

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,140 @@
...
+    let entry = ensure_key_entry(store, pub_key);
+    assert!(!entry.contains(&fk), EFunctionKeyAlreadyAdded);
+    entry.insert(fk);
+}
+
+/// **Disallow** a function key for a specific public key.
+public fun disallow(store: &mut FunctionKeysStore, pub_key: vector<u8>, fk: &FunctionKey) {
```
miker83z: public(package) should be enough
```suggestion
public(package) fun disallow(store: &mut FunctionKeysStore, pub_key: vector<u8>, fk: &FunctionKey) {
```

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,166 @@
...
+    } else {
+        // OWNER FLOW: verify against the stored owner public key (bypass restrictions).
+        // If succeeds, we short-circuit.
+        let owner_pk = borrow_public_key(account);
+        assert!(pub_key == owner_pk);
+        assert!(
+            ed25519::ed25519_verify(&sig_bytes, owner_pk, ctx.digest()),
+            EEd25519VerificationFailed,
+        );
+    }
```
miker83z, line 164:
```suggestion
    // Decode signature once for both attempts.
    let sig_bytes = decode(signature);

    // Verify against the stored owner public key.
    let owner_pk = borrow_public_key(account);
    let is_owner = pub_key == owner_pk;
    let is_ed25519_verified = ed25519::ed25519_verify(&sig_bytes, &pub_key, ctx.digest());

    if (is_owner) {
        // OWNER FLOW
        assert!(
            is_ed25519_verified,
            EEd25519VerificationFailed,
        );
    } else {
        // FUNCTION KEY FLOW
        assert!(fk_store_exists(account), EFunctionKeysNotInitialized);

        // Verify delegated signature against provided pub_key.
        assert!(
            is_ed25519_verified,
            EEd25519VerificationFailed,
        );
        
        // Require exactly one command.
        assert!(auth_ctx.tx_commands().length() == 1, EInvalidAmountOfCommands);
        // Extract and check allow-set membership.
        let func_key = extract_func_key(&auth_ctx.tx_commands()[0]);
        let fk_store = borrow_fk_store(account);
        assert!(fk_store.is_allowed(pub_key, &func_key), EUnauthorized);
    }
```

examples/move/function_keys/sources/function_keys.move
miker83z: The semantic of `ctx.sender()` is different. I clarified this concept in slack. See https://github.com/iotaledger/iota/pull/8697/files#r2428858894 for a code suggestion.

miker83z (changes requested):

examples/move/function_keys/sources/function_keys.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
Dkwcs: Done

examples/move/function_keys/sources/function_keys.move
Dkwcs: What I've found that we can't attach the `fk_store` during the creation of iota_account.
IotaAccount `add_field` method verifies that `self.id.uid_to_address() == ctx.sender()` , but during the account creation it's impossible. Sender cannot be equal to the account address, because account just was created

miker83z: Right, then you could create a function `build_fn_keys_store` that builds the `FunctionKeysStore { function_keys: tbl::new<vector<u8>, VecSet<FunctionKey>>(ctx) }`
and then:

```move
let account = builder(authenticator, ctx)
    .add_dynamic_field(OwnerPublicKey {}, public_key)
    .add_dynamic_field(fk_store_key(), build_fn_keys_store(public_key, ctx))
    .finish();
account.share();
```

Dkwcs: `fk_store_key` is private, so we should reconsider our approach for option 2

Dkwcs: Done it here - a2b785c0c9882d07984452f561a183aeddfdf33d
However we can discuss it one more time

miker83z: Should be public(package) then, we need to restrict access to that key.

Dkwcs: Done

examples/move/function_keys/tests/function_keys_tests.move
```diff
@@ -0,0 +1,570 @@
...
+        let mut cmds = vector::empty<Command>();
+        vector::push_back(
+            &mut cmds,
+            make_move_call_for_testing(
+                package_id,
+                b"wallet".to_ascii_string(),
+                b"withdraw".to_ascii_string(),
+            ),
+        );
+        let auth_ctx = create_auth_context_with_commands_for_testing(cmds);
```
miker83z:
```suggestion
    
    // TX 1
    scenario.next_tx(account_address);
    {
        let digest = x"315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3";

        let account = scenario.take_shared<IOTAccount>();
        let ctx = create_tx_context_for_testing(account_address, digest);
        let auth_ctx = create_auth_context_with_commands_for_testing(vector::empty<Command>());
```

miker83z: Owner path does not require fn keys to be set

examples/move/function_keys/tests/function_keys_tests.move
```diff
@@ -0,0 +1,570 @@
...
+use iota::test_scenario::{Self as scen, Scenario};
+use iota::tx_context as txc;
+use iotaccount::iotaccount::{Self, IOTAccount};
+use std::ascii;
+
+// ----------------------------------------------------------------------------
+// Happy path (delegated): attach → grant(pubkey, function) → authenticate OK
+// ----------------------------------------------------------------------------
+#[test]
+fun test_fk_authenticate_happy_path_delegated() {
```
miker83z:
```suggestion
fun test_fk_authenticate_happy_path() {
```

miker83z: Delegation, in terms of having a different sender, does not exist for AA. The sender is always equal to account_address

examples/move/function_keys/tests/function_keys_tests.move
```diff
@@ -0,0 +1,570 @@
...
+
+        let fk = make_func_key(package_id.to_address(), b"wallet", b"withdraw");
+
+        function_keys::grant_permission(&mut account, user_public_key, fk, ctx);
+
+        scen::return_shared(account);
+    };
+    let delegated_sender = @0xA;
+    // TX 2: delegated authenticate (sender != account), exactly one matching MoveCall
+    scenario.next_tx(delegated_sender);
```
miker83z:
```suggestion
    // TX 2: exactly one matching MoveCall
    scenario.next_tx(account_address);
```

miker83z: Same for the other functions below

examples/move/function_keys/sources/function_keys.move
miker83z: I forgot that this should be put at the start of the function:
```move
// Ensure the sender is indeed the account address.
ensure_tx_sender_is_account(account, ctx);
```

examples/move/function_keys/sources/function_keys.move
```diff
@@ -0,0 +1,173 @@
...
+
+/// Creates a new `IOTAccount` as a shared object with the given authenticator, but without
+/// attaching the Function Keys store. This is useful for testing purposes.
+#[test_only]
+public fun create_without_fk_store(public_key: vector<u8>, authenticator: AuthenticatorInfoV1, ctx: &mut TxContext) {
+    let account = builder(authenticator, ctx)
+        .add_dynamic_field(OwnerPublicKey {}, public_key)
+        .finish();
+    account.share();
+}
```
miker83z: This is not formatted

examples/move/function_keys/sources/function_keys.move
Dkwcs: Fixed

examples/move/function_keys/tests/function_keys_tests.move
Dkwcs: Done

examples/move/function_keys/tests/function_keys_tests.move
Dkwcs: Done

examples/move/function_keys/tests/function_keys_tests.move
Dkwcs: Done

examples/move/function_keys/sources/function_keys.move
Dkwcs: Fixed

examples/move/function_keys/tests/function_keys_tests.move
```diff
@@ -0,0 +1,533 @@
...
+    let owner_pk = x"1ea6f0f467574295a2cd5d21a3fd3a712ade354d520d3bd0fe6088d7b7c2e00e";
+    let account_address = create_iotaccount_for_testing_without_fk_store(
+        scenario,
+        option::some(owner_pk),
+    );
+    let package_id = object::id_from_bytes(iota::hash::blake2b256(&b"0x123"));
+
+    scenario.next_tx(account_address);
+    {
+        let mut account = scenario.take_shared<IOTAccount>();
```
miker83z:
```suggestion
        let account = scenario.take_shared<IOTAccount>();
```

examples/move/function_keys/tests/function_keys_tests.move
```diff
@@ -0,0 +1,533 @@
...
+
+        let digest = x"315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3";
+        let mut ctx = create_tx_context_for_testing(account_address, digest);
```
miker83z:
```suggestion
        let ctx = create_tx_context_for_testing(account_address, digest);
```

miker83z (approved):

examples/move/function_keys/tests/function_keys_tests.move
Dkwcs: Done

examples/move/function_keys/tests/function_keys_tests.move
Dkwcs: Done

examples/move/function_keys/sources/fk_store.move
```diff
@@ -0,0 +1,141 @@
...
+    account.has_field(fk_store_key())
+}
+
+public fun borrow_fk_store(account: &IOTAccount): &FunctionKeysStore {
+    account.borrow_field(fk_store_key())
+}
+
+public fun borrow_fk_store_mut(account: &mut IOTAccount, ctx: &TxContext): &mut FunctionKeysStore {
+    account.borrow_field_mut(fk_store_key(), ctx)
+}
```
valeriyr: if `fk_store_key` is `public(package)` looks like it makes sense to remove all these functions and move `FunctionKeysName` to `function_keys.move`. Use it in the same way as `OwnerPublicKey`. I am not insisting and leaving this for your consideration.

valeriyr (approved):
I left one comment for consideration, but in general looks good to me!

examples/move/function_keys/sources/fk_store.move
Dkwcs: Fixed according your comment

PR #8713, opened 2025-09-30
feat(starfish): send shard and proof in a block bundle

### LINKED ISSUE #8672: Store own shard+proof and send them in block bundle
1. When we receive a block and check its transactions' commitment, save our own shard and its proof. 
2. Store shard and proof in dag_state and storage in the same way as we store transactions.
3. Update block bundle structure to send a vector of shards and proofs.
4. Send shards and proofs in a block bundle. Keep track of which shards are needed for which validator by remembering which shards have already been sent.
5. Add logic in the authority service to verify and store these additional shards from block bundles.

VorobyevIlya:
# Description of change

In this PR, we start sending shards in block bundles.

In more detail, the following changes are introduced.
1. When we receive a block bundle, our shard with its proof is extracted from the transaction data of the received block. Shard and its proof are sent to the core to be stored in dag_state.
2. When we create a block bundle to send to validator **v**, we add to the bundle shards and their proofs that we know but haven't yet sent to validator **v**.
3. When we receive a block bundle, we extract those shards and proofs and verify proofs against the transaction commitment. Then the shards with correct proofs are collected to be sent to decoders. The decoding logic will be implemented in a subsequent PR.
4. Shard eviction from dag_state is implemented.

Fixes #8672 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -26,8 +26,8 @@ use tracing::{debug, info, warn};
 use crate::{
     CommitIndex, Round, Transaction, VerifiedBlockHeader,
     block_header::{
-        BlockHeaderAPI, BlockHeaderDigest, BlockRef, GENESIS_ROUND, SignedBlockHeader,
-        TransactionsCommitment, VerifiedBlock, VerifiedTransactions,
+        BlockHeaderAPI, BlockHeaderDigest, BlockRef, GENESIS_ROUND, ShardWithProof,
+        SignedBlockHeader, TransactionsCommitment, VerifiedBlock, VerifiedTransactions,
     },
     block_verifier::BlockVerifier,
     commit::{CommitAPI as _, CommitRange, TrustedCommit},
@@ -38,8 +38,8 @@ use crate::{
     encoder::ShardEncoder,
     error::{ConsensusError, ConsensusResult},
     network::{
-        BlockBundle, BlockBundleStream, NetworkService, SerializedBlock, SerializedBlockAndHeaders,
-        SerializedBlockBundle, SerializedHeaderAndTransactions, SerializedTransactions,
+        BlockBundle, BlockBundleStream, NetworkService, SerializedBlock, SerializedBlockBundle,
+        SerializedBlockBundleParts, SerializedHeaderAndTransactions, SerializedTransactions,
     },
     stake_aggregator::{QuorumThreshold, StakeAggregator},
     storage::Store,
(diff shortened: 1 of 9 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -464,9 +484,19 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
 
         missing_ancestors.extend(missing_block_ancestors);
         missing_committed_txns.extend(missing_block_committed_transactions);
+        // 11. Add our shard from the received block and its proof to the dag_state
+        self.core_dispatcher
+            .add_shards(vec![ShardWithProof {
+                shard: our_shard,
+                transaction_commitment,
+                proof: proof_for_shard,
+                block_ref,
```
polinikita: I would create the shard with proof outside the call of the method. Also, it makes sense to make it verified shard and store in memory in addition serialization of this struct such that when sending the shard it is sufficient to copy bytes only.

crates/starfish/core/src/authority_service.rs
```diff
@@ -445,7 +443,29 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+                count: serialized_block_bundle_parts.serialized_shards.len(),
+                limit: MAX_SHARDS_PER_BUNDLE,
+            });
+        }
+        // TODO: use correct type
+        let mut shards_for_decoder: Vec<ShardWithProof> = vec![];
+        for serialized_shard in serialized_block_bundle_parts.serialized_shards.iter() {
+            let shard: ShardWithProof =
+                bcs::from_bytes(serialized_shard).map_err(ConsensusError::MalformedShard)?;
+            if TransactionsCommitment::check_merkle_proof(
```
polinikita: I think that one needs to output a consensus error about incorrect proof here

crates/starfish/core/src/authority_service.rs
```diff
@@ -464,9 +484,23 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+            shard: our_shard,
+            transaction_commitment,
+            proof: proof_for_shard,
+            block_ref,
+        };
+        let serialized_shard_for_core: Bytes = bcs::to_bytes(&shard_for_core)
+            .map_err(ConsensusError::SerializationFailure)?
+            .into();
+        self.core_dispatcher
+            .add_shards(vec![(block_ref, serialized_shard_for_core)])
```
polinikita: I would join this tuple (block_ref, serialized_shard_for_core) into a new type, e.g. VerifiedOwnShard or something

polinikita (changes requested):
Added a few comments to improve error handling and readiness

crates/starfish/core/src/block_header.rs
```diff
@@ -417,13 +417,47 @@ impl AsRef<[u8]> for BlockHeaderDigest {
...
+pub(crate) struct ShardWithProof {
+    pub(crate) shard: Shard,
+    pub(crate) transaction_commitment: TransactionsCommitment,
+    pub(crate) proof: MerkleProofBytes,
+    pub(crate) block_ref: BlockRef,
+}
+
+impl PartialEq for ShardWithProof {
+    fn eq(&self, other: &Self) -> bool {
+        self.transaction_commitment == other.transaction_commitment
```
filipdulic: ShardWithProof is a single shard of many for a TransactionCommitment, so 2 different shards of the same transaction_commitment should not be compared through the transaction_commitment field.

filipdulic (approved):

crates/starfish/core/src/authority_service.rs

crates/starfish/core/src/block_header.rs
VorobyevIlya: You are right, removed completely, since we don't need it anymore

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Done. Actually, it helped me to find a bug, where I passed the wrong peer index to the proof checker. So all proofs were wrong, and shards had been skipped. Now it is correct.

polinikita (approved):

PR #8723, opened 2025-10-01
refactor(indexer): insert object deletion and mutation chunks concurrently

### LINKED ISSUE #8676: upstream(indexer): execute objects mutation & deletion in parallel
https://github.com/MystenLabs/sui/pull/19766

kodemartin:
# Description of change

This patch makes a minor optimization in the iota-indexer following the respective upstream changes.

Resolves #8676

No further QA tests were deemed necessary as this patch only makes the insertion of deletion and mutations concurrent, on the guarantee that all objects are unique and have the highest version among a set.

### DIFF
--- a/crates/iota-indexer/src/types.rs
+++ b/crates/iota-indexer/src/types.rs
@@ -23,6 +23,7 @@ use iota_types::{
     transaction::SenderSignedData,
 };
 use move_core_types::language_storage::StructTag;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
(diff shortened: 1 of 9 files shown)


crates/iota-indexer/src/types.rs
```diff
@@ -427,13 +428,45 @@ impl IndexedObject {
     }
 }
 
+#[cfg(any(feature = "pg_integration", feature = "shared_test_runtime", test))]
+impl IndexedObject {
+    pub fn random() -> Self {
```
tomxey, line 434: nit: could we maybe name it `random_for_testing_only`, analogous to the  method name in `IotaAddress`? Since I believe it should not be used in prod code

crates/iota-indexer/src/types.rs
```diff
@@ -427,13 +428,45 @@ impl IndexedObject {
...
 #[derive(Clone, Debug)]
 pub struct IndexedDeletedObject {
     pub object_id: ObjectID,
     pub object_version: u64,
     pub checkpoint_sequence_number: u64,
 }
 
+#[cfg(any(feature = "pg_integration", feature = "shared_test_runtime", test))]
+impl IndexedDeletedObject {
+    pub fn random() -> Self {
```
tomxey, line 461: nit: could we maybe name it `random_for_testing_only`, analogous to the  method name in `IotaAddress`? Since I believe it should not be used in prod code

crates/iota-indexer/src/types.rs
tomxey: Okay, now I see that it's only compiled for test feature flags.

crates/iota-indexer/src/types.rs
tomxey: Okay, now I see that it's only compiled for test feature flags.

tomxey (approved):

sergiupopescu199 (approved):
lgtm 🦀

PR #8733, opened 2025-10-01
fix(wallet): stake flow, amplitude send event with correct amount

### LINKED ISSUE #8704: Fix staking event bug (reset before event sent)
When staking via wallet, the form resets before amplitude receives the event, causing all staking events to show 0. Fix event timing so the correct values are sent.

panteleymonchuk:
Closes #8704


apps/wallet/src/ui/app/staking/stake/StakeForm.tsx
```diff
@@ -134,8 +135,7 @@ export function StakeFormComponent({ validatorAddress, epoch, onSuccess }: Stake
 
     const handleSubmit = async (_: FormValues, formikHelpers: FormikHelpers<FormValues>) => {
         try {
-            const response = await stakeTokenMutateAsync(formikHelpers);
-            onSuccess(response);
+            await stakeTokenMutateAsync(formikHelpers);
```
VmMad: you can also pass a onSuccess here, so it's not necessary to pass the formik helpers all the way down, just do this:

```suggestion
            await stakeTokenMutateAsync(undefined, {
                onSuccess: () => {
                    formikHelpers.resetForm()
                }
            });
```

VmMad (changes requested):

KeitoTadashi (approved):

VmMad (approved):

PR #8736, opened 2025-10-01
feat: investigate a solution where immutable objects are created when a package is published

### LINKED ISSUE #8719: [aa][authenticator] Investigate a solution where immutable objects are created when a package is published
When a developer publishes a module that includes some functions that could be used as `authenticate function` the AuthenticatorInfoV1 related to that function is published as immutable object.
- Consider the alternative of having a broader "PackageMetadata" object.
- In order for the authenticate function to be recognizable the [authenticator] attribute can be used, however, this part should not be part of this investigation.

valeriyr:
# Description of change

This PR shows how we can implement the idea of creating `AuthenticatorInfoV1` instances during package publishing and updating.

It shouldn't be considered as a solution; it is just a prototype.

fixes #8719

### DIFF
--- a/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
+++ b/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
@@ -15,7 +15,9 @@ mod checked {
     use iota_move_natives::object_runtime::ObjectRuntime;
     use iota_protocol_config::ProtocolConfig;
     use iota_types::{
-        IOTA_FRAMEWORK_ADDRESS, auth_context,
+        IOTA_FRAMEWORK_ADDRESS,
+        account::AuthenticatorInfoV1,
+        auth_context,
         base_types::{
             IotaAddress, MoveObjectType, ObjectID, RESOLVED_ASCII_STR, RESOLVED_STD_OPTION,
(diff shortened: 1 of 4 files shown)


iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -96,26 +99,56 @@ mod checked {
...
+                            move_authenticator_info_object,
+                            Owner::Immutable,
+                            previous_transaction,
+                        );
+
+                        authenticator_info_objects.push(authenticator_info_object);
+                    }
+                }
+                Err(err) => {
+                    let object_runtime: &ObjectRuntime = context.object_runtime();
```
Dkwcs, line 142: To be honest, it’s not fully clear for me, why we’re loading objects when error occurs.
I’m still getting familiar with [doc](https://iotafoundation.atlassian.net/wiki/spaces/LSC/pages/2531164180/Part+1+V2+-+create+auth+info), probably I’ll find an answer there, but it would be nice if you add minor clarification of this flow or something like (regular flow - this flow).

miker83z:
Closed in favor of #9296

PR #8737, opened 2025-10-01
fix(dashboard): update validator address and amount in stake transaction logging

### LINKED ISSUE #8707: Add validator property to staking events in dashboard
Staking events from the wallet dashboard are missing the validator property (though timelock staking has it). Add validator info to standard staking events.

### LINKED ISSUE #8709: Fix validator property in timelockStake event
In the timelockStake event, we are incorrectly sending the sender address instead of the selected validator. Update to send the correct validator.

panteleymonchuk:
Closes #8707
Closes #8709

### DIFF
--- a/apps/wallet-dashboard/components/dialogs/staking/views/EnterAmountView.tsx
+++ b/apps/wallet-dashboard/components/dialogs/staking/views/EnterAmountView.tsx
@@ -9,7 +9,7 @@ import {
     getGasBudgetErrorMessage,
     NO_BALANCE_GENERIC_MESSAGE,
 } from '@iota/core';
-import { CoinFormat, IOTA_DECIMALS, IOTA_TYPE_ARG, parseAmount } from '@iota/iota-sdk/utils';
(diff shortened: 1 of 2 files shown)


apps/wallet-dashboard/components/dialogs/staking/views/EnterAmountView.tsx
```diff
@@ -94,7 +94,8 @@ export function EnterAmountView({
                     onSuccess(tx.digest);
                     toast.success('Stake transaction has been sent');
                     ampli.stakedIota({
-                        stakedAmount: Number(parseAmount(values.amount, IOTA_DECIMALS)),
+                        stakedAmount: Number(values.amount),
```
msarcev, line 97: Is it safe to remove `parseAmount` here ?

panteleymonchuk: Here we just provide number what user put to input. We have validation of the form, so it should be safe.

KeitoTadashi (approved):

VmMad (approved):

PR #8738, opened 2025-10-01
feat(e2e-tests): add abstract account create and issuing TX tests

### LINKED ISSUE #8670: [AA] Add e2e test for IOTAccount creation and TX issuing

miker83z:
# Description of change

This PR adds a new set of e2e tests for the Account Abstraction feature:
- create an abstract account
- issue a TX using it
- issue a sponsored TX using it

This is based on an implementation of a Move module that supports basic operations for an Abstract Account (it could be modified later).

This is also the base for a series of additional e2e tests in the future.

Fixes #8670 


Dkwcs (commented):
Maybe I missed that part we discussed it, however I’d like to clarify it one more time - `abstract_account.move` is nearly a minimal re-implementation of the `IOTAccount` surface. We’re using it for the test purpose, but also at the same time we have to consider that any changes in “original” AA account should be reflected here and vice versa, how we will deal with it? Why we don’t use existing implementation?

In case of test cases we can add more negative scenarios like:
Wrong signature → expect `EEd25519VerificationFailed`
Wrong sender for a mutating call → expect `ETransactionSenderIsNotTheAccount`
etc
Maybe they can be considered as unit test cases, but at least ETransactionSenderIsNotTheAccount looks quite global

miker83z:
@Dkwcs 
> Maybe I missed that part we discussed it, however I’d like to clarify it one more time - `abstract_account.move` is nearly a minimal re-implementation of the `IOTAccount` surface. We’re using it for the test purpose, but also at the same time we have to consider that any changes in “original” AA account should be reflected here and vice versa, how we will deal with it? Why we don’t use existing implementation?

Having a dedicated package for the test allows us to modify it more easily in case we need some particular types or functions for the e2e tests. In this case it is still very similar, in fact I am waiting for #8744 to be finished to copy the structure of the account implementation. But, still, the idea is to decouple the two implementations because they serve different purposes.

> In case of test cases we can add more negative scenarios like: Wrong signature → expect `EEd25519VerificationFailed` Wrong sender for a mutating call → expect `ETransactionSenderIsNotTheAccount` etc Maybe they can be considered as unit test cases, but at least ETransactionSenderIsNotTheAccount looks quite global

Indeed, we need to create more tests! (with new PRs)

valeriyr (approved):

Dkwcs (approved):

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -0,0 +1,365 @@
...
+const ABSTRACTACCOUNT_AUTHENTICATE_FN_NAME_FREE_ACCESS: &str = "authenticate_free_access";
+
+#[sim_test]
+async fn test_abstract_account_creation_and_issue_tx() -> Result<(), anyhow::Error> {
+    telemetry_subscribers::init_for_testing();
+
+    // Build a test environment and create an abstract account
+    let mut test_env = TestEnvironment::new().await;
+    test_env
+        .setup_abstract_account(ABSTRACTACCOUNT_AUTHENTICATE_FN_NAME_ED25519)
```
TheMrAI: Could we shorten all `abstractaccount` strings to just `aa`?

TheMrAI (approved):
Not sure I entirely like the duplication, but I assume, you know what you are doing.

Perhaps add some minimal description to the `crates/iota-e2e-tests/tests/abstract_account_tests.rs` describing that the duplicated code is temporary until we stabilizie/figure out/etc.
So when others stumble upon it, won't be confused.

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: [d0e493d](https://github.com/iotaledger/iota/pull/8738/commits/d0e493d0d108142697394277fe4d41561fb42755)

PR #8739, opened 2025-10-02
feat(iota-json-rpc-types): support results in batchTransaction

Thoralf-M:
# Description of change

Add support to reference results of previous commands in the transaction like this `"arguments": [{"Result": 0}]`

Fixes https://github.com/iotaledger/iota/issues/8763

```bash
iota start --force-regenesis --with-faucet

iota client switch --env localnet
iota client faucet
ACTIVE_ADDRESS=$(iota client active-address)
read GAS_COIN COIN_TO_SPLIT < <(iota client gas --json | jq -r '[.[0].gasCoinId, .[1].gasCoinId] | @sh' | tr -d \')
AMOUNT_TO_SPLIT=1000
# just sending to own address
RECIPIENT_ADDRESS=$(iota client active-address)
GAS_BUDGET=100000000

# without gas coin
JSON_DATA='{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "unsafe_batchTransaction",
  "params": {
    "signer": "'$SENDER_ADDRESS'",
    "single_transaction_params": [
      {
        "moveCallRequestParams": {
          "packageObjectId": "0x2",
          "module": "coin",
          "function": "split",
          "typeArguments": ["0x2::iota::IOTA"],
          "arguments": ["'$COIN_TO_SPLIT'", "'$AMOUNT_TO_SPLIT'"]
        }
      },
      {
        "moveCallRequestParamsV2": {
          "packageObjectId": "0x2",
          "module": "transfer",
          "function": "public_transfer",
          "typeArguments": ["0x2::coin::Coin<0x2::iota::IOTA>"],
          "arguments": [{"Result": 0}, {"Pure": "'$RECIPIENT_ADDRESS'"}]
        }
      }
    ],
    "gas": "'$GAS_COIN'",
    "gas_budget": "'$GAS_BUDGET'"
  }
}'

echo "$JSON_DATA"

curl http://localhost:9000 \
--header 'content-type: application/json' \
--data "$JSON_DATA" | jq .
```

```JSON
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "txBytes": "AAADAQA3td2Z0qo0perjxpLvkeal2gagyNzh4eRAIjOcdpmXmwIAAAAAAAAAIOoHM85GUzDO3pf8HMApM8E1KnnOlVWrhhgYquKUOzsjAAjoAwAAAAAAAAAgERERERUE6TUOY11lzTjM0sApQ0xqOkgNiUepumoVshUCAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACBGNvaW4Fc3BsaXQBBwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACBGlvdGEESU9UQQACAQAAAQEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACCHRyYW5zZmVyD3B1YmxpY190cmFuc2ZlcgEHAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIEY29pbgRDb2luAQcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgRpb3RhBElPVEEAAgIAAAECABEREREVBOk1DmNdZc04zNLAKUNMajpIDYlHqbpqFbIVARfhNmCqOECu1WCaq7Kt8z25MkA/BI2KrRRMrcIqrY2+AgAAAAAAAAAgtGDVq6PjTxcJ9KQdJp0LhXtlUyX23syRmSgTNHADnTgRERERFQTpNQ5jXWXNOMzSwClDTGo6SA2JR6m6ahWyFegDAAAAAAAAAOH1BQAAAAAA",
    "gas": [
      {
        "objectId": "0x17e13660aa3840aed5609aabb2adf33db932403f048d8aad144cadc22aad8dbe",
        "version": 2,
        "digest": "D982YfFE89pNogyCPUZDgGSBAGz29KSdRoymEYoqXzsq"
      }
    ],
    "inputObjects": [
      {
        "ImmOrOwnedMoveObject": {
          "objectId": "0x37b5dd99d2aa34a5eae3c692ef91e6a5da06a0c8dce1e1e44022339c7699979b",
          "version": 2,
          "digest": "GkYmiw1Mqbk9iNaY5Z241BRttw7RM71CXitotVTuMBxe"
        }
      },
      {
        "MovePackage": "0x0000000000000000000000000000000000000000000000000000000000000002"
      },
      {
        "ImmOrOwnedMoveObject": {
          "objectId": "0x17e13660aa3840aed5609aabb2adf33db932403f048d8aad144cadc22aad8dbe",
          "version": 2,
          "digest": "D982YfFE89pNogyCPUZDgGSBAGz29KSdRoymEYoqXzsq"
        }
      }
    ]
  }
}
```

Some tests fail because the graphql snapshots and TS SDK isn't updated in this PR, they should be updated in another one

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: Added PtbInput to support referencing results in unsafe_batchTransaction
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Added `single_move_call_with_ptb_inputs()` to support referencing results of previous commands
- [ ] REST API:

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
@@ -176,6 +176,7 @@ pub(crate) enum TransactionArgument {
     GasCoin(GasCoin),
     Input(Input),
     Result(TxResult),
+    Pure(PureArg),
 }
 
 /// Access to the gas inputs, after they have been smashed into one coin. The
@@ -207,6 +208,13 @@ pub(crate) struct TxResult {
     /// the individual result among the multiple results from that command
     /// (also 0-indexed).
     ix: Option<u16>,
+}
+
+/// A pure value argument.
+#[derive(SimpleObject, Clone, Debug, Eq, PartialEq)]
+pub(crate) struct PureArg {
+    /// The JSON representation of the pure value.
+    value: String,
 }
 
 /// A user transaction that allows the interleaving of native commands (like
@@ -437,3 +445,19 @@ impl From<IotaArgument> for TransactionArgument {
         }
     }
 }
+
+// impl From<IotaArgumentV2> for TransactionArgument {
+//     fn from(argument: IotaArgumentV2) -> Self {
+//         use IotaArgumentV2 as S;
+//         use TransactionArgument as A;
+//         match argument {
+//             S::GasCoin => A::GasCoin(GasCoin { dummy: None }),
+//             S::Input(ix) => A::Input(Input { ix }),
+//             S::Result(cmd) => A::Result(TxResult { cmd, ix: None }),
+//             S::NestedResult(cmd, ix) => A::Result(TxResult { cmd, ix:
+// Some(ix) }),             S::Pure(v) => A::Pure(PureArg {
+//                 value: format!("{v:?}"),
+//             }),
+//         }
+//     }
+// }
--- a/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
@@ -176,7 +176,6 @@ pub(crate) enum TransactionArgument {
     GasCoin(GasCoin),
     Input(Input),
     Result(TxResult),
-    Pure(PureArg),
 }
 
 /// Access to the gas inputs, after they have been smashed into one coin. The
@@ -208,13 +207,6 @@ pub(crate) struct TxResult {
     /// the individual result among the multiple results from that command
     /// (also 0-indexed).
     ix: Option<u16>,
-}
-
-/// A pure value argument.
-#[derive(SimpleObject, Clone, Debug, Eq, PartialEq)]
-pub(crate) struct PureArg {
-    /// The JSON representation of the pure value.
-    value: String,
 }
 
 /// A user transaction that allows the interleaving of native commands (like
--- a/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
@@ -445,19 +445,3 @@ impl From<IotaArgument> for TransactionArgument {
         }
     }
 }
-
-// impl From<IotaArgumentV2> for TransactionArgument {
-//     fn from(argument: IotaArgumentV2) -> Self {
-//         use IotaArgumentV2 as S;
-//         use TransactionArgument as A;
-//         match argument {
-//             S::GasCoin => A::GasCoin(GasCoin { dummy: None }),
-//             S::Input(ix) => A::Input(Input { ix }),
-//             S::Result(cmd) => A::Result(TxResult { cmd, ix: None }),
-//             S::NestedResult(cmd, ix) => A::Result(TxResult { cmd, ix:
-// Some(ix) }),             S::Pure(v) => A::Pure(PureArg {
-//                 value: format!("{v:?}"),
-//             }),
-//         }
-//     }
-// }
--- a/crates/iota-json-rpc-types/src/displays/transaction_displays.rs
+++ b/crates/iota-json-rpc-types/src/displays/transaction_displays.rs
@@ -11,8 +11,8 @@ use tabled::{
 };
 
 use crate::{
-    IotaArgument, IotaCallArg, IotaCommand, IotaObjectArg, IotaProgrammableMoveCall,
-    IotaProgrammableTransactionBlock, displays::Pretty,
+    IotaArgument, IotaArgumentV2, IotaCallArg, IotaCommand, IotaObjectArg,
+    IotaProgrammableMoveCall, IotaProgrammableTransactionBlock, displays::Pretty,
 };
 
 impl Display for Pretty<'_, IotaProgrammableTransactionBlock> {
@@ -179,6 +179,21 @@ impl Display for Pretty<'_, IotaArgument> {
             IotaArgument::Input(i) => format!("Input  {i}"),
             IotaArgument::Result(i) => format!("Result {i}"),
             IotaArgument::NestedResult(j, k) => format!("Nested Result {j}: {k}"),
+        };
+        write!(f, "{output}")
+    }
+}
+
+impl Display for Pretty<'_, IotaArgumentV2> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let Pretty(argument) = self;
+
+        let output = match argument {
+            IotaArgumentV2::GasCoin => "GasCoin".to_string(),
+            IotaArgumentV2::Input(i) => format!("Input  {i}"),
+            IotaArgumentV2::Result(i) => format!("Result {i}"),
+            IotaArgumentV2::NestedResult(j, k) => format!("Nested Result {j}: {k}"),
+            IotaArgumentV2::Pure(v) => format!("Pure {v:?}"),
         };
         write!(f, "{output}")
     }
--- a/crates/iota-json-rpc-types/src/displays/transaction_displays.rs
+++ b/crates/iota-json-rpc-types/src/displays/transaction_displays.rs
@@ -190,7 +190,7 @@ impl Display for Pretty<'_, IotaArgumentV2> {
 
         let output = match argument {
             IotaArgumentV2::GasCoin => "GasCoin".to_string(),
-            IotaArgumentV2::Input(i) => format!("Input  {i}"),
+            IotaArgumentV2::Input(i) => format!("Input {i}"),
             IotaArgumentV2::Result(i) => format!("Result {i}"),
             IotaArgumentV2::NestedResult(j, k) => format!("Nested Result {j}: {k}"),
             IotaArgumentV2::Pure(v) => format!("Pure {v:?}"),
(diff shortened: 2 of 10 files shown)


crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
```diff
@@ -437,3 +445,19 @@ impl From<IotaArgument> for TransactionArgument {
...
+//             S::GasCoin => A::GasCoin(GasCoin { dummy: None }),
+//             S::Input(ix) => A::Input(Input { ix }),
+//             S::Result(cmd) => A::Result(TxResult { cmd, ix: None }),
+//             S::NestedResult(cmd, ix) => A::Result(TxResult { cmd, ix:
+// Some(ix) }),             S::Pure(v) => A::Pure(PureArg {
+//                 value: format!("{v:?}"),
+//             }),
+//         }
+//     }
+// }
```
DaughterOfMars: 👀

crates/iota-json-rpc-types/src/displays/transaction_displays.rs
```diff
@@ -183,3 +183,18 @@ impl Display for Pretty<'_, IotaArgument> {
...
     }
 }
+
+impl Display for Pretty<'_, IotaArgumentV2> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let Pretty(argument) = self;
+
+        let output = match argument {
+            IotaArgumentV2::GasCoin => "GasCoin".to_string(),
+            IotaArgumentV2::Input(i) => format!("Input  {i}"),
```
DaughterOfMars: was two spaces intentional here?

Thoralf-M: No, good catch

Thoralf-M: https://github.com/iotaledger/iota/pull/8739/commits/9e14b4e95677439fcac01073f246a7e29ad7d4cf

crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
Thoralf-M: Forgot to remove it, did it now in https://github.com/iotaledger/iota/pull/8739/commits/9e14b4e95677439fcac01073f246a7e29ad7d4cf

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -198,6 +181,115 @@ impl TransactionBuilder {
...
+        module: &Identifier,
+        function: &Identifier,
+        type_args: &[TypeTag],
+        call_args: Vec<IotaArgumentV2>,
+    ) -> Result<Vec<Argument>, anyhow::Error> {
+        let package = self.load_move_package(package_id).await?;
+
+        let module_compiled = package.deserialize_module(module, &BinaryConfig::standard())?;
+        let function_str = function.as_ident_str();
+        let fdef = module_compiled
```
DaughterOfMars: Maybe call this `function_def`?

Thoralf-M: Sure

DaughterOfMars:
So I see that the big change is adding the Pure variant to the arguments, but how is that related to referencing results of previous commands?

crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
```diff
@@ -176,6 +176,7 @@ pub(crate) enum TransactionArgument {
     GasCoin(GasCoin),
     Input(Input),
     Result(TxResult),
+    Pure(PureArg),
```
kodemartin: The scope of this variant is unclear, as its introduction is isolated from the rest of the API. I see there is a `Pure` type already that could be extended to accommodate such use cases.

But at any case, I suggest you delegate the graphql changes into a separate issue with a description of the requirements for this feature.

Thoralf-M:
> So I see that the big change is adding the Pure variant to the arguments, but how is that related to referencing results of previous commands?

`MoveCallParamsV2` uses `pub arguments: Vec<IotaArgumentV2>` instead of `pub arguments: Vec<IotaJsonValue>` in `MoveCallParams`, this difference makes it possible provide Results. I didn't modify `IotaJsonValue` as this seems to be limited to actual MoveValues (there is `.to_move_value()` for example), adding results there would probably be unexpected in other cases

DaughterOfMars (approved):

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2049,6 +2049,49 @@ impl From<Argument> for IotaArgument {
     }
 }
 
+/// An argument to a transaction in a programmable transaction block
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+pub enum IotaArgumentV2 {
```
kodemartin: The premise here is that It is ok to break the Rust API as long as we are backward compatible on the web interface. For example this is satisfied by the addition of the new `RPCTransactionRequestParam`.

On that basis we could think of adding the following type instead 
```rust
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
#[serde(untagged)]
pub enum PtbInput {
     CallArg(IotaJsonValue)
     PtbRef(IotaArgument)
}
```
This is backward compatible with `IotaJsonValue` with respect to deserialization, and could be used in `MoveCallParams` directly. So without the need for a `MoveCallParamsV2`. And it would not require `v2` methods in the `transaction-builder` either.

What do you think?

crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
Thoralf-M: Good catch, this is actually not needed at all, since there is no API currently to build txs with graphql
Reverted here https://github.com/iotaledger/iota/pull/8739/commits/f13fa85890d1b8ec82ce0e3a74d02bc876d8b20e

crates/iota-json-rpc-types/src/iota_transaction.rs
Thoralf-M: Would definitely clean some things up, but would be a breaking change in the transaction builder crate, so not sure if we want that
Draft PR with `PtbInput` change https://github.com/iotaledger/iota/pull/8779
And total diff with all changes:
https://github.com/iotaledger/iota/compare/develop...dev-tools/PtbInput

kodemartin: Then you can limit the introduction of new methods only in the transaction builder. With a `single_move_call_with_ptb_inputs` method (similar to the `_v2` variant).

Thoralf-M: Thanks for the suggestion, I made the change and like it much more now!

DaughterOfMars (approved):

crates/iota-json/src/lib.rs
```diff
@@ -789,6 +789,27 @@ pub fn resolve_move_function_args(
...
+        type_args,
+        combined_args_json,
+        true,
+    )
+}
+
+/// Resolve the JSON args of a function into the expected formats to make them
+/// usable by Move call. This version only optionally checks the argument count
+/// to allow using it in combination with result arguments.
+pub fn resolve_move_function_args_inner(
```
kodemartin: Does this have to be `pub`?

crates/iota-transaction-builder/src/lib.rs
```diff
@@ -466,6 +466,39 @@ impl TransactionBuilder {
         Ok(())
     }
 
+    /// Add a single move call to the provided
+    /// [`ProgrammableTransactionBuilder`].
+    /// Accepting PtbInput so one can also provide results from previous move
+    /// calls.
```
kodemartin: Nit: To comply with the [summary-sentence guidelines](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#summary-sentence)

```suggestion
    /// Adds a single move call to the provided
    /// [`ProgrammableTransactionBuilder`].
    ///
    /// Accepting [`PtbInput`] so one can also provide results from previous move
    /// calls.
```

kodemartin (commented):
Looks good, but you need to regenerate the open-rpc spec.

```
$ cargo run -p iota-open-rpc --example generate-json-rpc-spec
```

crates/iota-json/src/lib.rs
Thoralf-M: No, can be made private

Thoralf-M: But it's actually not needed anymore

Thoralf-M:
> Looks good, but you need to regenerate the open-rpc spec.

Should this be done in this PR? I thought the generated files should be done separated

crates/iota-json/src/lib.rs
Thoralf-M: Undid the change to add it here https://github.com/iotaledger/iota/pull/8739/commits/4335c9b92e538536585d3bb552dd021b3417d4d8

kodemartin:
> > Looks good, but you need to regenerate the open-rpc spec.
> 
> Should this be done in this PR? I thought the generated files should be done separated

Yes, yes. We are targeting a feature branch.

kodemartin (approved):
lgtm 🎈

PR #8744, opened 2025-10-02
refactor(examples): Make iota_account reusable

### LINKED ISSUE #8710: Refactor examples/move/iota_account
We have identified that the basic functionality of the IOTAccount can and should be extracted in [Time locked example](https://github.com/iotaledger/iota/pull/8602) PR.
(That PR will be rebased and only contain the `time_locked` example, so it won't work for future references.)
For posterity the idea is that an IOTAccount is essentially nothing but an ID and some dynamic field management logic.
We need to ensure that there are some reserved dynamic fields which can only be rotated in special cases and that there are regular dynamic fields which can hold data. This provides all the flexibility we need for handling an abstract account.

Follow this naming:
- iotaccount::IOTAccount the base struct, can't be transferred(shared) without having set an authenticate function
- iotaccount::IOTAccountBuilder makes so that external packages/PTBs can create IOTAccounts, i.e., can set authenticators
- iotaccount::basic_keyed_account contains the basic cryptographic authenticate functions and the function that calls IOTAccountBuilder  and sets the pub keys as DF (and rotate logic)

TheMrAI:
# Description of change

Refactor `iota_account` example to make it more reusable by extracting its construction and protection aspects.
`iota_account` has its tests fully reworked, now they should cover 100% all expected behavior.
`basic_keyed_account` has been left in the `iota_account` package for now, for simplicity, but in the future we may extract it further. The tests for this module have only been ported with no further modifications. Some of it can/could be simplified, because with the current setup certain scenarios are already forbidden, such as when non-accounts attempt to modify state.

fixes #8710 


examples/move/iota_account/sources/basic_keyed_account.move
```diff
@@ -0,0 +1,128 @@
...
+use iota::ed25519;
+use iota::hex::decode;
+use iota_account::iota_account::{Self, IOTAccount, ensure_tx_sender_is_account};
+
+#[error(code = 10)]
+const EEd25519VerificationFailed: vector<u8> = b"Ed25519 authenticator verification failed.";
+#[error(code = 11)]
+const ESecp256k1VerificationFailed: vector<u8> = b"Secp256k1 authenticator verification failed.";
+#[error(code = 12)]
+const ESecp256r1VerificationFailed: vector<u8> = b"Secp256r1 authenticator verification failed.";
```
valeriyr: We can have the codes equal to 0, 1, 2

examples/move/iota_account/sources/basic_keyed_account.move
```diff
@@ -0,0 +1,128 @@
...
+    let authenticator_df_name = account::authenticator_df_name();
+    account.rotate(authenticator_df_name, authenticator, ctx);
+}
+
+// --------------------------------------- Utilities ---------------------------------------
+
+/// An utility function to borrow the account-related public key.
+fun borrow_public_key(account: &IOTAccount): &vector<u8> {
+    account.borrow_field(OwnerPublicKey {})
+}
```
valeriyr: I suppose we want to have access to the public key from the package outside.

There are two options to do this:
1. Make this function public, which looks more preferable to me.
2. Make a public function like this: `create_owner_public_key`. In this case, it is possible to add dynamic fields with this key from the package outside.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
+// --------------------------------------- IOTAccountBuilder ---------------------------------------
 
-/// A dynamic field key for the account owner public key.
-public struct OwnerPublicKey has copy, drop, store {}
+/// Safely construct an IOTAccount.
+///
+/// The builder is entirely temporary. It cannot be copied, stored or dropped.
+///
+/// Account implementations are expected to call the builder in a single function call,
+/// add the desired `AuthenticatorInfo` and all reserved dynamic fields necessary for the
```
valeriyr:
```suggestion
/// add the desired `AuthenticatorInfoV1` and all reserved dynamic fields necessary for the
```

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
 
-/// This struct represents an IOTA account on-chain.
-/// It holds all the related data as dynamic fields to simplify updates, migrations and extensions.
-public struct IOTAccount has key {
-    id: UID,
+/// Construct an IOTAccountBuilder and set the Authenticator.
+///
+/// The `AuthenticatorInfo` will be attached as a dynamic field under key provided by:
+/// `account::authenticator_df_name()`.
+public fun builder<Authenticator: copy + drop + store>(
```
valeriyr: I caught the idea, but I am not sure that we should pass the authenticator as a generic parameter.
Because in this case, we can pass a type which is not an authenticator implementation, `String`, for example.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
+        value_bytes: bcs::to_bytes(&key),
+    }
+}
+
+/// The key by which the list of reserved dynamic fields,
+/// can be queried from an `IOTAccount`.
+public struct ReservedDynamicFields has copy, drop, store {}
+
+/// Create the key for accessing reserved dynamic fields.
+public fun get_reserved_dynamic_fields(): ReservedDynamicFields {
```
valeriyr: I propose to remove/make private this function and add something like this:
```rust
public fun borrow_reserved_dynamic_fields(self: &IOTAccount): vector<DfKey> { ... }
```
This protects us from adding a dynamic field with this name by a user.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
-    let mut id = object::new(ctx);
+/// They aren't meant to be used by callers/developers as `dynamic_field`
+/// already handles differentiation better. Only necessary for our internally
+/// managed `ReservedDynamicFields`.
+public struct DfKey has copy, drop, store {
+    type_name: std::type_name::TypeName,
+    value_bytes: vector<u8>,
+}
+
+public(package) fun make_key<KeyType: copy + drop + store>(key: KeyType): DfKey {
```
valeriyr:
```suggestion
public(package) fun make_df_key<KeyType: copy + drop + store>(key: KeyType): DfKey {
```

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
-// --------------------------------------- Creation ---------------------------------------
+/// Attach a `Value` as a reserved dynamic field to the builder.
+public fun add_reserved_field<Name: copy + drop + store, Value: store>(
+    mut self: IOTAccountBuilder,
+    name: Name,
+    value: Value,
+): IOTAccountBuilder {
+    let field_key = make_key(name);
+    let reserved_dynamic_fields_key = get_reserved_dynamic_fields_key();
+    assert!(field_key != reserved_dynamic_fields_key, EReservedDynamicFieldsListCannotBeSet);
```
valeriyr: We can remove this check if it is impossible for a user to create a `ReservedDynamicFields` instance.

examples/move/iotaccount/sources/iotaccount.move
```diff
@@ -3,70 +3,157 @@
...
+    let reserved_dynamic_fields_key = get_reserved_dynamic_fields_key();
+    assert!(field_key != reserved_dynamic_fields_key, EReservedDynamicFieldsListCannotBeSet);
+
+    dynamic_field::add(&mut self.account.id, name, value);
+    let reserved_keys: &mut vector<DfKey> = dynamic_field::borrow_mut(
+        &mut self.account.id,
+        get_reserved_dynamic_fields(),
+    );
+    // No need to check for duplicates, because dynamic_field::add above would fail on colliding keys
+    // and in the builder one can only add fields.
```
valeriyr: This is true; it makes sense if we want to improve error handling by having a dedicated error. It makes errors clearer to a user.

examples/move/iotaccount/sources/iotaccount.move
```diff
@@ -3,70 +3,157 @@
...
+    let reserved_dynamic_fields_key = get_reserved_dynamic_fields_key();
+    assert!(field_key != reserved_dynamic_fields_key, EReservedDynamicFieldsListCannotBeSet);
+
+    dynamic_field::add(&mut self.account.id, name, value);
+
+    self
+}
+
+/// Finish building the `IOTAccount` and share the object.
+public fun finish(self: IOTAccountBuilder): IOTAccount {
```
valeriyr, line 78: nit: I would prefer the name `build`.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
+
+/// Attach a `Value` as a regular dynamic field to the builder.
+public fun add_regular_field<Name: copy + drop + store, Value: store>(
+    mut self: IOTAccountBuilder,
+    name: Name,
+    value: Value,
+): IOTAccountBuilder {
+    let field_key = make_key(name);
+    let reserved_dynamic_fields_key = get_reserved_dynamic_fields_key();
+    assert!(field_key != reserved_dynamic_fields_key, EReservedDynamicFieldsListCannotBeSet);
```
valeriyr: This can be remove as well

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
-/// - `authenticate_ed25519`
-/// - `authenticate_secp256k1`
-/// - `authenticate_secp256r1`
-public fun create(public_key: vector<u8>, authenticator: AuthenticatorInfoV1, ctx: &mut TxContext) {
-    // Create a UID for an account object.
-    let mut id = object::new(ctx);
+/// They aren't meant to be used by callers/developers as `dynamic_field`
+/// already handles differentiation better. Only necessary for our internally
+/// managed `ReservedDynamicFields`.
+public struct DfKey has copy, drop, store {
```
valeriyr:
```suggestion
public struct DynamicFieldKey has copy, drop, store {
```

examples/move/iota_account/sources/iota_account.move
valeriyr: This function can be private

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
+public fun get_reserved_dynamic_fields(): ReservedDynamicFields {
+    ReservedDynamicFields {}
+}
 
-    // Add the account owner public key as a dynamic field.
-    dynamic_field::add(&mut id, OwnerPublicKey{}, public_key);
+/// Create a `DfKey` for `ReservedDynamicFields`.
+public(package) fun get_reserved_dynamic_fields_key(): DfKey {
+    make_key(ReservedDynamicFields {})
+}
```
valeriyr: This function can be removed if we remove the related asserts.

examples/move/iota_account/sources/iota_account.move
```diff
@@ -77,14 +164,16 @@ public fun add_field<Name: copy + drop + store, Value: store>(
     ensure_tx_sender_is_account(self, ctx);
 
     // Check if `name` is allowed to be used.
-    check_reserved_df_name(&name);
+    check_reserved_df<Name>(self, &name);
```
valeriyr:
```suggestion
    check_reserved_df(self, &name);
```

examples/move/iota_account/sources/iota_account.move
```diff
@@ -94,13 +183,14 @@ public fun remove_field<Name: copy + drop + store, Value: store>(
...
```
valeriyr:
```suggestion
    check_reserved_df(self, &name);
```

examples/move/iota_account/sources/iota_account.move
```diff
@@ -132,121 +224,79 @@ public fun has_field<Name: copy + drop + store>(self: &IOTAccount, name: Name):
...
-}
+// --------------------------------------- Test Utilities ---------------------------------------
 
-/// Checks that the sender of this transaction is the account.
-fun ensure_tx_sender_is_account(self: &IOTAccount, ctx: &TxContext) {
-    assert!(self.id.uid_to_address() == ctx.sender(), ETransactionSenderIsNotTheAccount);
+#[test_only]
+public fun get_address(self: &IOTAccount): address {
+    self.id.to_address()
 }
```
valeriyr: We can have this function as a simple public function, should be useful

examples/move/iota_account/sources/iota_account.move
```diff
@@ -132,121 +224,79 @@ public fun has_field<Name: copy + drop + store>(self: &IOTAccount, name: Name):
...
 
-/// Checks if `name` is allowed to be used for a user-defined dynamic field.
-fun check_reserved_df_name<Name: copy + drop + store>(name: &Name) {
-    // Check that `Name` is not `OwnerPublicKey`.
-    assert!(std::type_name::get<Name>() != std::type_name::get<OwnerPublicKey>(), EOwnerPublicKeyCannotBeUsed);
+#[test_only]
+public(package) fun create_iotaccount_for_testing(
```
valeriyr: It is better to add the following functions to the tests directory.
`create_iotaccount_for_testing`
`create_authenticator_info_v1_for_testing`
They don't seem to be generic enough to be used in external tests.

examples/move/iota_account/sources/basic_keyed_account.move
TheMrAI: Exactly! The moment I tried to refactor time_locked I immediately realized this :D

examples/move/iota_account/sources/iota_account.move
TheMrAI: I deliberately left it without version. I think this can be version agnostic. What do you think?

examples/move/iota_account/sources/iota_account.move
TheMrAI: Using them for both iota_accout tests and iota_basic_keyed tests. Didn't want to duplicate.

examples/move/iota_account/sources/iota_account.move
TheMrAI: Good point.

examples/move/iota_account/sources/iota_account.move
valeriyr: You can create a new file for common utilities in the test folder

examples/move/iota_account/sources/iota_account.move
valeriyr: It looks like a struct reference when it is written in quotation marks. If you don't want to use a versioned name you can just write like this:
```
add a desired authenticator info and ...
```

Dkwcs:
Good improvements overall! I just want to highlight a few points that could make the structure and API clearer and more convenient to use:
1. Module naming:
- The current `basic_keyed_account.move` name may suggest an account type rather than an authenticator implementation. For the reusable library, we could use more descriptive names like:
`iota_account::auth_ed25519`, `iota_account::auth_k1`, `iota_account::auth_r1` (one module per authenticator), or
- A single `iota_account::authenticators` module (e.g. `account_authenticators.move`).
  Additionally, we might consider moving the create function and some utility methods directly into iota_account.move for better discoverability.

2. Builder usage:
The create method could be turned into a thin example that only delegates to the builder. Common logic such as reserved field addition and object sharing could live inside the builder’s fluent API.
For example, instead of manually calling `.share()` after `.finish()`, the builder could provide `.finish_shared()` to streamline the flow.

3. Builder improvements:
- We can consider whether we truly need the builder abstraction. If we keep it, integrating` .share()` into `.finish()` (or adding .`finish_shared()`) would make the flow more natural, since we don’t expose the IOTAccount itself to the outside world.
- Add a `builder.with_authenticator_v1()` method to construct and attach an AuthenticatorInfoV1 automatically. This would reduce repeated usage of authenticator_df_name in multiple places.
- Similarly, a .`with_public_key`(vec<u8>) method that internally uses the `OwnerPublicKey` dynamic field key would improve clarity and reduce boilerplate.

4. Naming consistency:
Minor but important: while we can keep the struct `OwnerPublicKey` type, renaming it to something like `OwnerPublicKeyDfKey` or `OwnerPublicKeyName` would make its purpose as a dynamic field key clearer and more consistent with naming patterns used elsewhere.

Dkwcs (commented):
Good improvements overall! I just want to highlight a few points that could make the structure and API clearer and more convenient to use:
1. Module naming:
- The current `basic_keyed_account.move` name may suggest an account type rather than an authenticator implementation. For the reusable library, we could use more descriptive names like:
`iota_account::auth_ed25519`, `iota_account::auth_k1`, `iota_account::auth_r1` (one module per authenticator), or
- A single `iota_account::authenticators` module (e.g. `account_authenticators.move`).
  Additionally, we might consider moving the create function and some utility methods directly into iota_account.move for better discoverability.

2. Builder usage:
The create method could be turned into a thin example that only delegates to the builder. Common logic such as reserved field addition and object sharing could live inside the builder’s fluent API.
For example, instead of manually calling `.share()` after `.finish()`, the builder could provide `.finish_shared()` to streamline the flow.

3. Builder improvements:
- We can consider whether we truly need the builder abstraction. If we keep it, integrating` .share()` into `.finish()` (or adding .`finish_shared()`) would make the flow more natural, since we don’t expose the IOTAccount itself to the outside world.
- Add a `builder.with_authenticator_v1()` method to construct and attach an AuthenticatorInfoV1 automatically. This would reduce repeated usage of authenticator_df_name in multiple places.
- Similarly, a .`with_public_key`(vec<u8>) method that internally uses the `OwnerPublicKey` dynamic field key would improve clarity and reduce boilerplate.

4. Naming consistency:
Minor but important: while we can keep the struct `OwnerPublicKey` type, renaming it to something like `OwnerPublicKeyDfKey` or `OwnerPublicKeyName` would make its purpose as a dynamic field key clearer and more consistent with naming patterns used elsewhere.

5. `borrow_public_key(account)` appears in `basic_keyed_account`. We can promote a single some kind of helper (e.g., iota_account::get_owner_public_key(&IOTAccount)) in the reusable lib so examples don’t redefine it.

TheMrAI:
> Good improvements overall! I just want to highlight a few points that could make the structure and API clearer and more convenient to use:
> 
> 1. Module naming:
> 
> * The current `basic_keyed_account.move` name may suggest an account type rather than an authenticator implementation. For the reusable library, we could use more descriptive names like:
>   `iota_account::auth_ed25519`, `iota_account::auth_k1`, `iota_account::auth_r1` (one module per authenticator), or
> * A single `iota_account::authenticators` module (e.g. `account_authenticators.move`).
>   Additionally, we might consider moving the create function and some utility methods directly into iota_account.move for better discoverability.
> 
> 2. Builder usage:
>    The create method could be turned into a thin example that only delegates to the builder. Common logic such as reserved field addition and object sharing could live inside the builder’s fluent API.
>    For example, instead of manually calling `.share()` after `.finish()`, the builder could provide `.finish_shared()` to streamline the flow.
> 3. Builder improvements:
> 
> * We can consider whether we truly need the builder abstraction. If we keep it, integrating` .share()` into `.finish()` (or adding .`finish_shared()`) would make the flow more natural, since we don’t expose the IOTAccount itself to the outside world.
> * Add a `builder.with_authenticator_v1()` method to construct and attach an AuthenticatorInfoV1 automatically. This would reduce repeated usage of authenticator_df_name in multiple places.
> * Similarly, a .`with_public_key`(vec) method that internally uses the `OwnerPublicKey` dynamic field key would improve clarity and reduce boilerplate.
> 
> 4. Naming consistency:
>    Minor but important: while we can keep the struct `OwnerPublicKey` type, renaming it to something like `OwnerPublicKeyDfKey` or `OwnerPublicKeyName` would make its purpose as a dynamic field key clearer and more consistent with naming patterns used elsewhere.
> 5. `borrow_public_key(account)` appears in `basic_keyed_account`. We can promote a single some kind of helper (e.g., iota_account::get_owner_public_key(&IOTAccount)) in the reusable lib so examples don’t redefine it.

1. Module naming:

- See the PR description. The name suggests it is an account because that is what it is.
- Yes, we could, but there

examples/move/iota_account/sources/iota_account.move
```diff
@@ -132,121 +224,79 @@ public fun has_field<Name: copy + drop + store>(self: &IOTAccount, name: Name):
...
-
-    dynamic_field::remove<_, vector<u8>>(account_id, owner_public_key);
-    dynamic_field::add(account_id, owner_public_key, public_key);
-
-    // Update the account owner public key dynamic field. It is expected that the field already exists.
-    let authenticator_df_name = account::authenticator_df_name();
-
-    dynamic_field::remove<_, AuthenticatorInfoV1>(account_id, authenticator_df_name);
-    dynamic_field::add(account_id, authenticator_df_name, authenticator);
+    dynamic_field::remove<_, Value>(account_id, name);
```
miker83z: It can return the value, so we don't need the drop constraint.

examples/move/iotaccount/sources/iotaccount.move
```diff
@@ -3,70 +3,157 @@
...
+/// The builder is entirely temporary. It cannot be copied, stored or dropped.
+///
+/// Account implementations are expected to call the builder in a single function call,
+/// add the desired `AuthenticatorInfo` and all reserved dynamic fields necessary for the
+/// operation of the account authentication logic.
+/// All reserved field `Name`s will be stored under key `ReservedDynamicFields` which is managed
+/// by the builder.
+///
+/// For convenience the regular dynamic fields may be added at this stage as well.
+public struct IOTAccountBuilder {
```
miker83z, line 26: As you suggested, I'd also create a dedicated module for the Builder

examples/move/iota_account/sources/iota_account.move
miker83z: I'd also change the name to `address()`

examples/move/iota_account/sources/iota_account.move
```diff
@@ -3,70 +3,157 @@
...
-/// - `authenticate_secp256k1`
-/// - `authenticate_secp256r1`
-public fun create(public_key: vector<u8>, authenticator: AuthenticatorInfoV1, ctx: &mut TxContext) {
-    // Create a UID for an account object.
-    let mut id = object::new(ctx);
+/// They aren't meant to be used by callers/developers as `dynamic_field`
+/// already handles differentiation better. Only necessary for our internally
+/// managed `ReservedDynamicFields`.
+public struct DfKey has copy, drop, store {
+    type_name: std::type_name::TypeName,
```
miker83z: I confirm what I thought during the meeting, in here we better use just the generic `T`. The string here can be as long as "0000000000000000000000000000000a::module_name1::type_name1<0000000000000000000000000000000a::module_name2::type_name2<u64>>" and even more. So using a type directly will make the storage gas usage more efficient.

miker83z: Something along this line should work (I guess this would require a dedicated module as well): 
```move
public struct DfKey<phantom Key: copy + drop + store> has copy, drop, store {
    value_bytes: vector<u8>,
}

public(package) fun make_key<Key: copy + drop + store>(key: &Key): DfKey<Key> {
    DfKey {
        value_bytes: bcs::to_bytes(key),
    }
}

public struct ReservedDynamicFieldKeys has store {
    /// This bag maps the type T from DfKey<T> to a vector of DfKey<T> instances.
    /// For instance, with DfKey<vector<u8>>, the Bag key will be "vector<u8>" and the value will be
    /// a vector of DfKey<vector<u8>> instances all with different bcs bytes representation.
    reserved_keys: Bag,
}

public fun add_reserved_df_key<Key: copy + drop + store>(
    mut self: ReservedDynamicFieldKeys,
    key: Key,
) {
    let df_key = make_key(&key);

    if (!self.reserved_keys.contains(key)) {
        self.reserved_keys.add(key, vector[df_key]);
    } else {
        let reserved_df_keys: &mut vector<DfKey<Key>> = self.reserved_keys.borrow_mut(key);
        assert!(!reserved_df_keys.contains(&df_key), 123);
        reserved_df_keys.push_back(df_key);
    }
}

public fun check_reserved_df<Key: copy + drop + store>(self: &ReservedDynamicFieldKeys, key: Key) {
    let df_key = make_key(&key);
    if (!self.reserved_keys.contains(key)) {
        return;
    };
    let reserved_df_keys: &vector<DfKey<Key>> = self.reserved_keys.borrow(key);
    let reserved_found = reserved_df_keys.any!(|reserved| reserved == &df_key);

    assert!(!reserved_found, ECantModifyReservedDynamicField);
}
```

miker83z: An instance of `ReservedDynamicFieldKeys` is then attached to the IOTAccount instead of the `vector<DfKey<Key>>`

miker83z: (I used a mixed naming convention, it is not what I suggest it should be)

TheMrAI:
> Good improvements overall! I just want to highlight a few points that could make the structure and API clearer and more convenient to use:
> 
> 1. Module naming:
> 
> * The current `basic_keyed_account.move` name may suggest an account type rather than an authenticator implementation. For the reusable library, we could use more descriptive names like:
>   `iota_account::auth_ed25519`, `iota_account::auth_k1`, `iota_account::auth_r1` (one module per authenticator), or
> * A single `iota_account::authenticators` module (e.g. `account_authenticators.move`).
>   Additionally, we might consider moving the create function and some utility methods directly into iota_account.move for better discoverability.
> 
> 2. Builder usage:
>    The create method could be turned into a thin example that only delegates to the builder. Common logic such as reserved field addition and object sharing could live inside the builder’s fluent API.
>    For example, instead of manually calling `.share()` after `.finish()`, the builder could provide `.finish_shared()` to streamline the flow.
> 3. Builder improvements:
> 
> * We can consider whether we truly need the builder abstraction. If we keep it, integrating` .share()` into `.finish()` (or adding .`finish_shared()`) would make the flow more natural, since we don’t expose the IOTAccount itself to the outside world.
> * Add a `builder.with_authenticator_v1()` method to construct and attach an AuthenticatorInfoV1 automatically. This would reduce repeated usage of authenticator_df_name in multiple places.
> * Similarly, a .`with_public_key`(vec) method that internally uses the `OwnerPublicKey` dynamic field key would improve clarity and reduce boilerplate.
> 
> 4. Naming consistency:
>    Minor but important: while we can keep the struct `OwnerPublicKey` type, renaming it to something like `OwnerPublicKeyDfKey` or `OwnerPublicKeyName` would make its purpose as a dynamic field key clearer and more consistent with naming patterns used elsewhere.
> 5. `borrow_public_key(account)` appears in `basic_keyed_account`. We can promote a single some kind of helper (e.g., iota_account::get_owner_public_key(&IOTAccount)) in the reusable lib so examples don’t redefine it.

1. Please read the PR description and the linked ticket. It is an account, it was left there to limit the scope of the change to what is necessary, but I will extract it then.
2. I don't understand what you are describing with the builder. Originally only `.finish_shared()` was provided, but @valeriyr rightly asked for it to be changed to make it more flexible. I can add the `finish_shared()` as well.
3. I don't understand what you are describing here again. No need for builder but should add`builder.with_authenticator_v1()`?
4. I will follow @miker83z suggestion on the matter and remove all descriptors from the names. It will be simply `OwnerPublicKey` ( in this case the key only remains because that variable is a public key not because it is used as a key).
5. It is hard to follow what you exactly mean by this, because your comment wasn't attached to a source section, and I don't remember.

In the future please avoid this walls of texts as feedback. With this I have to manually track the issues myself, track down what section of the code you were even referring to, and then manually track resolving them as well.
The change is about refactoring a piece of code, not making it perfect or satisfying potential future needs.
We can never know how exactly would something be good. The best we can do if provide a flexible interface and observe the emerging usage patterns.
Please don't respond to this thread and make proper comments for the relevant sections of the code with the suggestions.

examples/move/iota_account/sources/iota_account.move
TheMrAI: I don't understand exactly the whole thing, but I think I get what you mean. Will try to implement.

examples/move/iotaccount/sources/iotaccount.move
TheMrAI: Please don't write any `nit` comments. If you feel strongly about something, make a case for it. Nearly all of naming is personal preference anyways. Especially comments that you already feel so insignificant that you feel the need to preface it with `nit`. Bad practice in general.

examples/move/iotaccount/sources/iotaccount.move
valeriyr: If we have the IOTAccount and IOTAccountBuilder in the same file, IOTAccountBuilder has access to all the module internals. It is easier to create an account instance, add dynamic fields without limitations, and have access to internal dynamic field names(ReservedDynamicFieldKeys), errors, etc. Otherwise, we will need to expose more functions outside the module

Dkwcs (approved):

examples/move/iota_account/sources/iota_account.move
valeriyr: Don't we want to add a function to the framework that returns a dynamic field address and use it?
We can replace a noticeable part of this code by just using an `address`.

examples/move/iota_account/sources/basic_keyed_account.move
TheMrAI: Fixed in: 10c51f964dba22d2229ad74a48b36c9aa58b28e9

examples/move/iota_account/sources/basic_keyed_account.move
TheMrAI: Fixed in: 322487e12009cd5cca1d19727cddbb0f96001637

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: a401a51547bc820b7138c42fd847ad7f3f04f5a4

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 24977d29e7c8c55bcb5f649730171d78244facea

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 9ce4208609218b0eec7cbab902c45fd2f9538974

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 9ce4208609218b0eec7cbab902c45fd2f9538974

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 9ce4208609218b0eec7cbab902c45fd2f9538974

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 95d4bdc59a4b760cf97e5e2f83d06ebe53e69bc9

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: https://github.com/iotaledger/iota/commit/95d4bdc59a4b760cf97e5e2f83d06ebe53e69bc9

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: https://github.com/iotaledger/iota/commit/9ce4208609218b0eec7cbab902c45fd2f9538974

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 59a021601c07e5a3b5200f99bf54802869528aa3.

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 59a021601c07e5a3b5200f99bf54802869528aa3.

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: f9ba4aff34c162d5f5890276b3433146bf084e8e

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 909049792dbc30d47256123637526cddf1de6486

examples/move/iota_account/sources/iota_account.move
TheMrAI: Fixed in: 48caa01df898977b1e88916a99b19997760ce94e

TheMrAI:
Applied the change requested here: https://github.com/iotaledger/iota/issues/8710#issuecomment-3366016215

examples/move/iotaccount/sources/basic_keyed_account.move
```diff
@@ -0,0 +1,144 @@
...
+// === Package Functions ===
+
+// === Private Functions ===
+
+// === Test Functions ===
+
+#[test_only]
+public fun create_owner_public_key_for_testing(): OwnerPublicKey {
+    OwnerPublicKey {}
+}
```
valeriyr: This function can be removed. `borrow_public_key` can be used in the tests instead.
It is not supposed that `OwnerPublicKey` is used by a user, so using the public interface for testing is preferable.

examples/move/iotaccount/sources/basic_keyed_account.move
```diff
@@ -0,0 +1,144 @@
...
+    assert!(
+        ecdsa_r1::secp256r1_verify(&decode(signature), borrow_public_key(account), ctx.digest(), 0),
+        ESecp256r1VerificationFailed,
+    );
+}
+
+/// Rotates the account owner public key to a new one as well as the authenticator.
+/// Once this function is called, the previous public key and authenticator are no longer valid.
+/// Only the account itself can call this function.
+public fun rotate_public_key(
```
valeriyr, line 111: It would be nice to have a test using this function.

examples/move/iotaccount/tests/iotaccount_builder_tests.move
```diff
@@ -0,0 +1,129 @@
...
+
+        test_scenario::return_shared(account);
+    };
+
+    test_scenario::end(scenario_val);
+}
+
+#[test]
+#[expected_failure(abort_code = iota::dynamic_field::EFieldAlreadyExists)]
+fun attempting_to_add_a_field_as_reserved_then_regular() {
```
valeriyr: Looks like the name should be corrected

examples/move/iotaccount/tests/iotaccount_builder_tests.move
```diff
@@ -0,0 +1,129 @@
...
+        .add_dynamic_field(field_name, 3)
+        .finish();
+    account.share();
+
+    test_scenario::end(scenario_val);
+}
+
+#[test]
+#[expected_failure(abort_code = iota::dynamic_field::EFieldAlreadyExists)]
+fun attempting_to_add_a_field_as_regular_then_reserved() {
```
valeriyr: This test looks the same as the previous one; can be removed.

examples/move/iotaccount/tests/test_utils.move
```diff
@@ -0,0 +1,34 @@
...
+        .add_dynamic_field(b"SomeData".to_ascii_string(), 3u8)
+        .finish();
+    let account_address = account.account_address();
+
+    share(account);
+
+    account_address
+}
+
+#[test_only]
```
valeriyr: Since the module is test-only, it looks like we can remove the `#[test_only]` macros from the functions?

examples/move/iotaccount/tests/iotaccount_tests.move
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// --------------------------------------- Test Utilities ---------------------------------------
+
+macro fun non_account_sender($f: |&mut Scenario|) {
+    let mut scenario_val = test_scenario::begin(@0x0);
+    let scenario = &mut scenario_val;
+    create_iotaccount_for_testing(scenario);
+
+    scenario.next_tx(@0x0);
```
valeriyr, line 356: As I understand, this line can be removed because the scenario begins with the same address; it is not necessary to call this function.

valeriyr (approved):
I left several comments to cleanup the code. In general looks good!

examples/move/iotaccount/tests/iotaccount_tests.move
TheMrAI: If I remove it all non_account tests fail with: 
```
│ error[E11001]: test failure
│     ┌─ ./../../../crates/iota-framework/packages/iota-framework/sources/test/test_scenario.move:353:5
│     │
│ 351 │ public fun take_shared<T: key>(scenario: &Scenario): T {
│     │            ----------- In this function in iota::test_scenario
│ 352 │     let id_opt = most_recent_id_shared<T>();
│ 353 │     assert!(id_opt.is_some(), EEmptyInventory);
│     │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Test did not error as expected. Expected test to abort with error constant 'ETransactionSenderIsNotTheAccount' originating in the module iotaccount::iotaccount but instead it aborted with code 3 originating in the module iota::test_scenario rooted here
```

examples/move/iotaccount/sources/basic_keyed_account.move
TheMrAI: Fixed in: 63e49484ec

examples/move/iotaccount/tests/test_utils.move
TheMrAI: Fixed in: 63e49484ec

examples/move/iotaccount/tests/iotaccount_builder_tests.move
TheMrAI: Fixed in: c428e5a0f0

examples/move/iotaccount/tests/iotaccount_builder_tests.move
TheMrAI: Fixed in: c428e5a0f0

examples/move/iotaccount/sources/basic_keyed_account.move
TheMrAI: Apologies, these were lost in translation.
Restored original tests in: 8951839d60

miker83z (approved):

PR #8745, opened 2025-10-02
fix(indexer): Handle errors for task common_handler.start_transform_and_load

### LINKED ISSUE #5780: Handle errors in spawn_monitored_task!(common_handler.start_transform_and_load(receiver, cancel));
> I'm a bit worried about this task as it does not handle errors, if any error will occur the task will exit but the indexer will potentially still run, having some unexpected effects.
> ```rust
> spawn_monitored_task!(common_handler.start_transform_and_load(receiver, cancel));
> ```

_Originally posted by @sergiupopescu199 in https://github.com/iotaledger/iota/pull/5777#pullrequestreview-2660983225_

To be investigated and fixed if needed, considering comments:
 - https://github.com/iotaledger/iota/pull/5777#discussion_r1981357074
 - https://github.com/iotaledger/iota/pull/5777#discussion_r1981438509

tomxey:
# Description of change

Handle errors for task common_handler.start_transform_and_load

fixes https://github.com/iotaledger/iota/issues/5780

I've ran custom checks, to verify that on failure graceful shutdown of the other branch works.

### DIFF
--- a/crates/iota-indexer/src/indexer.rs
+++ b/crates/iota-indexer/src/indexer.rs
@@ -4,7 +4,7 @@
 
 use std::{collections::HashMap, env};
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use iota_data_ingestion_core::{
     DataIngestionMetrics, IndexerExecutor, ProgressStore, ReaderOptions, WorkerPool,
@@ -64,13 +64,14 @@ impl Indexer {
 
         // Start objects snapshot processor, which is a separate pipeline with its
         // ingestion pipeline.
-        let (object_snapshot_worker, object_snapshot_watermark) = start_objects_snapshot_handler(
-            store.clone(),
-            metrics.clone(),
-            snapshot_config,
-            cancel.clone(),
-        )
-        .await?;
+        let (object_snapshot_worker, object_snapshot_watermark, mut object_snapshot_task_handle) =
+            start_objects_snapshot_handler(
+                store.clone(),
+                metrics.clone(),
+                snapshot_config,
+                cancel.clone(),
+            )
+            .await?;
 
         if let Some(retention_config) = retention_config {
             let pruner = Pruner::new(store.clone(), retention_config, metrics.clone())?;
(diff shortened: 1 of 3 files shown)


sergiupopescu199 (approved):

crates/iota-indexer/src/indexer.rs
```diff
@@ -140,8 +141,26 @@ impl Indexer {
...
-            )
-            .await?;
+            ),
+        );
+
+        tokio::select! {
+            executor_result = &mut executor_handle => {
+                // Executor completed first - cancel snapshot task and check result
+                cancel.cancel();
+                let snapshot_result = object_snapshot_task_handle.await;
```
kodemartin: Shouldn't we [abort](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort) here? 

Is there a guarantee that executor completion guarantees snapshot completion so that to await? Otherwise this might hang.

crates/iota-indexer/src/indexer.rs
```diff
@@ -140,8 +141,26 @@ impl Indexer {
...
+                executor_result.context("failed to join data ingestion executor")?.context("data ingestion executor failed")?;
+                snapshot_result.context("failed to join snapshot task during shutdown")?.context("snapshot task failed during shutdown")?;
+            },
+            snapshot_result = &mut object_snapshot_task_handle => {
+                // Snapshot task completed first - cancel executor and check result
+                cancel.cancel();
+                let executor_result = executor_handle.await;
```
kodemartin: Same here: shouldn't we abort instead?

crates/iota-indexer/src/indexer.rs
tomxey: Proceeded with a timeout: a37b2cba6eb75f43b290f379416b77f28e168a3c

crates/iota-indexer/src/indexer.rs
tomxey: Proceeded with a timeout: a37b2cba6eb75f43b290f379416b77f28e168a3c

crates/iota-indexer/src/indexer.rs
```diff
@@ -15,6 +15,9 @@ use prometheus::Registry;
 use tokio_util::sync::CancellationToken;
 use tracing::info;
 
+// Timeout for waiting for tasks to shutdown gracefully after cancellation
```
sergiupopescu199:
```suggestion
/// Timeout for waiting for tasks to shutdown gracefully after cancellation
```

tomxey: fixed

kodemartin (approved):
lgtm

PR #8751, opened 2025-10-03
feat(consensus): add reconstruction from shards components to Starfish

### LINKED ISSUE #8673: Add decoder
1. Create shard accumulators for each block_ref
2. Send shards from each connection task to one reconstruction task, which accumulates the respected shards.
3. Create a given number of decoders in the reconstruction task.
4. Spawn decoding when a shard accumulator is ready for reconstruction
5. Output the reconstructed transactions to the core periodically.

polinikita:
# Description of change

This PR introduces a new component to Starfish consensus, called **ShardReconstructor**.  
Its handle becomes part of an authority node.

Upon receiving a new block bundle via streaming and verifying its components, the authority service component sends **transaction messages** to the ShardReconstructor:
- `ShardMessage` – a single shard of encoded transaction data corresponding to some block reference and transaction commitment
- `FullTransactionMessage` – a full transaction, that indicates that these transaction data was verified against the block header.

The ShardReconstructor:
- Collects shards per block (`ShardAccumulator`). 
- Filters shards for blocks that are already processed (e.g. with `FullTransactionMessage`)
- Detects when enough shards (`info_length`) are available to reconstruct
- Offloads reconstruction jobs to a specified number async workers
- Verifies commitments against reconstructed transactions
- Forwards verified transactions to the core (`CoreThreadDispatcher::add_transactions`) if the respected headers are already part of the dag state
- Periodically evicts old state based on the DAG’s GC round

Unit and integration tests cover:
- Reconstruction only after `info_length` shards
- Stopping shard collection when a full transaction message is received
- Multi-round recovery where one authority is blocked, verifying reconstruction from other authorities

Fixes #8673 

### Release Notes

- [x] Protocol: Starfish consensus allows for decoding of the transaction data from partial shards.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/block_manager/mod.rs
+++ b/crates/starfish/core/src/block_manager/mod.rs
@@ -88,7 +88,7 @@ impl BlockManager {
             if exists[i] {
                 self.dag_state
                     .write()
-                    .add_transactions(block.verified_transactions);
+                    .add_transactions(block.verified_transactions, "block streaming");
             } else {
                 self.suspended_blocks.insert(block.reference(), block);
             }
@@ -149,7 +149,7 @@ impl BlockManager {
                 // for this accepted header we already have a block, so we add it to dag_state
                 self.dag_state
                     .write()
-                    .add_transactions(block.verified_transactions);
+                    .add_transactions(block.verified_transactions, "block streaming");
             }
         }
 
--- a/crates/starfish/core/src/block_manager/mod.rs
+++ b/crates/starfish/core/src/block_manager/mod.rs
@@ -88,7 +88,7 @@ impl BlockManager {
             if exists[i] {
                 self.dag_state
                     .write()
-                    .add_transactions(block.verified_transactions, "block streaming");
+                    .add_transactions(block.verified_transactions, "Block streaming");
             } else {
                 self.suspended_blocks.insert(block.reference(), block);
             }
--- a/crates/starfish/core/src/shard_reconstructor.rs
+++ b/crates/starfish/core/src/shard_reconstructor.rs
@@ -19,19 +19,11 @@ use tokio::{
 };
 use tracing::{debug, warn};
 
-use crate::{
-    BlockRef, Transaction, VerifiedBlockHeader,
-    block_header::{
-        BlockHeaderDigest, Shard, ShardWithProof, TransactionsCommitment, VerifiedBlock,
-        VerifiedTransactions,
-    },
-    context::Context,
-    core_thread::CoreThreadDispatcher,
-    dag_state::DagState,
-    decoder::{ShardsDecoder, create_decoder},
-    encoder::{ShardEncoder, create_encoder},
-    error::{ConsensusError, ConsensusResult},
-};
+use crate::{BlockRef, Transaction, VerifiedBlockHeader, block_header::{
+    BlockHeaderDigest, Shard, ShardWithProof, TransactionsCommitment, VerifiedBlock,
+    VerifiedTransactions,
+}, context::Context, core_thread::CoreThreadDispatcher, dag_state::DagState, decoder::{ShardsDecoder, create_decoder}, encoder::{ShardEncoder, create_encoder}, error::{ConsensusError, ConsensusResult}, Round};
+use crate::block_header::GENESIS_ROUND;
 
 #[derive(Clone)]
 pub struct ShardAccumulator {
@@ -311,6 +303,8 @@ impl<C: CoreThreadDispatcher + 'static> ShardReconstructor<C> {
 }
 
 pub struct ShardReconstructor<C: CoreThreadDispatcher> {
+    /// Shards below this round will not be collected
+    transaction_gc_round: Round,
     info_length: usize,
     total_length: usize,
     context: Arc<Context>,
@@ -347,6 +341,7 @@ impl<C: CoreThreadDispatcher> ShardReconstructor<C> {
             context: context.clone(),
             core_dispatcher,
             dag_state,
+            transaction_gc_round: GENESIS_ROUND,
             reconstruction_queue: BTreeSet::new(),
             ready_to_reconstruct_sender: ready_sender,
             ready_to_reconstruct_receiver: Arc::new(Mutex::new(ready_receiver)),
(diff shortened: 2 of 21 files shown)


crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1144 @@
...
+            block_ref,
+            transactions_commitment: commitment,
+            collected_shards,
+            number_shards: 0,
+            header_verified: true,
+        }
+    }
+
+    /// Update the accumulator with a new shard
+    fn update_with_shard(&mut self, msg: ShardMessage) {
```
VorobyevIlya, line 148: mayve add assert that commitments are the same

crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1144 @@
...
+            self.collected_shards.clone(),
+        )?;
+
+        let serialized =
+            Transaction::serialize(&transactions).expect("We should expect serialization to work");
+
+        // Verify the commitment
+        // TODO: remove this verification since the header was verified before the
+        // reconstruction and f+1 shards guarantee that one honest validator already
+        // checked the alignment between the commitment and the header.
```
VorobyevIlya: so, the commitment verification should remain only if the header_verified flag in ShardAccumulator is false?

crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1134 @@
...
+    total_length: usize,
+    context: Arc<Context>,
+    /// Already processed transaction either by authority service or by shard
+    /// reconstructor
+    processed_transactions: BTreeSet<BlockRef>,
+    /// A cache of reconstructed transactions that will be periodically sent in
+    /// the core
+    reconstructed_transactions: BTreeMap<BlockRef, VerifiedTransactions>,
+    /// A map of all shard accumulators. Periodically evicted. Keyed by a pair
+    /// (BlockRef, TransactionsCommitment) since it is possible to m
```
VorobyevIlya: unfinished description

crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1134 @@
...
+                    match job {
+                        Some(shard_accumulator) => {
+                            metrics.node_metrics.reconstruction_jobs_started.inc();
+                            match shard_accumulator.decode_by_codec(&mut codec) {
+                                Ok(verified_transactions) => {
+                                    debug!(
+                                        "Successfully reconstructed transactions for block {:?}",
+                                        shard_accumulator.block_ref
+                                    );
+                                    let _ = result_tx.send(verified_transactions).await;
```
VorobyevIlya: Should we process the possible error here? or at least log it?

crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1134 @@
...
+
+    async fn get_transactions_with_headers_in_dag_state(&mut self) -> Vec<VerifiedTransactions> {
+        let transactions_map = std::mem::take(&mut self.reconstructed_transactions);
+        let block_refs: Vec<BlockRef> = transactions_map.keys().copied().collect();
+        // In most cases, all reconstructed transactions will go to the core
+        let mut ready_to_be_sent_transactions = Vec::new();
+        let mut to_stay_transactions = BTreeMap::new();
+
+        // We introduce a check about the existence of block headers to ensure that for
+        // every transaction, we have the respected header in the dag state
```
VorobyevIlya, line 501: do we need to check the commitment here that it is the same as in the header?

VorobyevIlya (commented):
left some comments

crates/starfish/core/src/shard_reconstructor.rs
polinikita: Let us add the assert. In theory, it is not needed as one of `f+1` was honest and had same block_ref and transaction_commitment, but better to be safe

crates/starfish/core/src/shard_reconstructor.rs
polinikita: Hm, accumulators are keyed by (block_ref and transctions_commitment) when reading the state. When updating with shard, we are sure that both fields are the same
```
let key = (msg.block_ref(), msg.transactions_commitment());
        let total_length = self.total_length;

        match msg {
            TransactionMessage::Shard(shard_msg) => match self.shard_accumulators.entry(key) {
                Entry::Vacant(v) => {
                    v.insert(ShardAccumulator::new_with_shard(shard_msg, total_length));
                }
                Entry::Occupied(mut o) => {
                    o.get_mut().update_with_shard(shard_msg);
                }
            },
```

crates/starfish/core/src/shard_reconstructor.rs
polinikita: thanks

crates/starfish/core/src/shard_reconstructor.rs
polinikita: Added a log

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -88,7 +88,7 @@ impl BlockManager {
             if exists[i] {
                 self.dag_state
                     .write()
-                    .add_transactions(block.verified_transactions);
+                    .add_transactions(block.verified_transactions, "Block streaming");
```
VorobyevIlya, line 91: sometimes it is 'Block', and sometimes it is 'block', should be consistent

crates/starfish/core/src/shard_reconstructor.rs
VorobyevIlya: I know, it is fine here, but we may use the function in another place.

VorobyevIlya (approved):

crates/starfish/core/src/shard_reconstructor.rs
```diff
@@ -0,0 +1,1151 @@
...
+            self.info_length,
+            &key,
+        )
+        .await?;
+
+        Ok(())
+    }
+
+    /// If the accumulator for the given key is ready to reconstruct, remove it
+    /// from the map and enqueue it for reconstruction
```
filipdulic, line 611: is there a reason why `enqueue_if_ready` is implemented as an associated function instead of as a method (&self, key).

filipdulic (approved):

crates/starfish/core/src/shard_reconstructor.rs
polinikita: fixed, thanks for noting. There was a reason, but that idea was later abonded

PR #8755, opened 2025-10-03
feat(graphql-rpc): support move-view function calls

### LINKED ISSUE #8346: [graphql]: support move view functions

kodemartin:
# Description of change

This patch adds support for move-view function calls in the GraphQL API.

The output is fully compatible with the JSON-RPC:

This query
```graphql
{
  moveViewCall(
    functionName: "0x2d7a8e89feadf4d346643cb69cac861d177891ab4c82b2ff7666324b772a53ac::wat_counter::get_counter"
    arguments: ["0x6a574d438156bd05ec9cd4b569fe50313b16d43eaab4833cc358bf4f26e9fe7a"]
  ) {
    error
    results
  }
}
```
returns this
```json
{
  "data": {
    "moveViewCall": {
      "error": null,
      "results": [
        "10"
      ]
    }
  }
}
```

While this query

```graphql
{
  moveViewCall(
    functionName: "0x2d7a8e89feadf4d346643cb69cac861d177891ab4c82b2ff7666324b772a53ac::wat_counter::get_wat_object"
    arguments: ["0x6a574d438156bd05ec9cd4b569fe50313b16d43eaab4833cc358bf4f26e9fe7a"]
  ) {
    error
    results
  }
}
```

Returns this
```json
{
  "data": {
    "moveViewCall": {
      "error": null,
      "results": [
        {
          "type": "0x2d7a8e89feadf4d346643cb69cac861d177891ab4c82b2ff7666324b772a53ac::wat_counter::Wat",
          "fields": {
            "counter": "10",
            "id": {
              "id": "0x6a574d438156bd05ec9cd4b569fe50313b16d43eaab4833cc358bf4f26e9fe7a"
            }
          }
        }
      ]
    }
  }
}
```

Fixes #8346 

Manual tests have been performed additionally.

### Release Notes

- [x] GraphQL: Introducing a new `moveViewCall` query path to call move-view functions. :rotating_light:  The server now requires that the fullnode JSON-RPC url is specified in the configuration, to be able to run `mutation.executeTransactionBlock`, `query.{moveViewCall,dryRunTransactionBlock}`. :rotating_light:

### DIFF
--- a/crates/iota-graphql-rpc/src/server/builder.rs
+++ b/crates/iota-graphql-rpc/src/server/builder.rs
@@ -37,7 +37,6 @@ use iota_json_rpc_api::CLIENT_SDK_TYPE_HEADER;
 use iota_metrics::spawn_monitored_task;
 use iota_network_stack::callback::{CallbackLayer, MakeCallbackHandler, ResponseHandler};
 use iota_package_resolver::{PackageStoreWithLruCache, Resolver};
-use iota_sdk::IotaClientBuilder;
 use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
 use tokio::{join, net::TcpListener, sync::OnceCell};
 use tokio_util::sync::CancellationToken;
@@ -467,18 +466,7 @@ impl ServerBuilder {
 
         // SDK for talking to fullnode. Used for executing transactions only
         // TODO: fail fast if no url, once we enable mutations fully
-        let (iota_sdk_client, write_api) = if let Some(url) = &config.tx_exec_full_node.node_rpc_url
-        {
-            let iota_sdk_client = IotaClientBuilder::default()
-                .request_timeout(RPC_TIMEOUT_ERR_SLEEP_RETRY_PERIOD)
-                .max_concurrent_requests(MAX_CONCURRENT_REQUESTS)
-                .build(url)
-                .await
-                .map_err(|e| {
-                    Error::Internal(format!(
-                        "Failed to connect to fullnode {e}. Is the node server running?"
-                    ))
-                })?;
+        let write_api = if let Some(url) = &config.tx_exec_full_node.node_rpc_url {
             let json_rpc_client = get_json_rpc_client(url)?;
             let indexer_store = PgIndexerStore::new(reader.get_pool(), indexer_metrics.clone());
             let optimistic_tx_executor = OptimisticTransactionExecutor::new(
@@ -491,12 +479,12 @@ impl ServerBuilder {
                 WriteApi::new(json_rpc_client, reader),
                 optimistic_tx_executor,
             );
-            (Some(iota_sdk_client), Some(write_api))
+            Some(write_api)
         } else {
             warn!(
                 "No fullnode url found in config. `dryRunTransactionBlock` and `executeTransactionBlock` will not work"
             );
-            (None, None)
+            None
         };
 
         builder = builder
@@ -505,7 +493,6 @@ impl ServerBuilder {
             .context_data(db)
             .context_data(pg_conn_pool)
             .context_data(resolver)
-            .context_data(iota_sdk_client)
             .context_data(write_api)
             .context_data(iota_names_config)
             .context_data(zklogin_config)
--- a/crates/iota-graphql-rpc/src/server/builder.rs
+++ b/crates/iota-graphql-rpc/src/server/builder.rs
@@ -29,6 +29,7 @@ use iota_graphql_rpc_headers::LIMITS_HEADER;
 use iota_indexer::{
     apis::{OptimisticWriteApi, WriteApi},
     db::{get_pool_connection, setup_postgres::check_db_migration_consistency},
+    indexer_reader::IndexerReader,
     metrics::IndexerMetrics,
     optimistic_indexing::OptimisticTransactionExecutor,
     store::PgIndexerStore,
@@ -445,7 +446,7 @@ impl ServerBuilder {
             // time).
             config.service.limits.request_timeout_ms.into(),
         )
-        .map_err(|e| Error::Internal(format!("Failed to create pg connection pool: {e}")))?;
+        .map_err(|e| Error::ServerInit(format!("Failed to create pg connection pool: {e}")))?;
 
         // DB
         let db = Db::new(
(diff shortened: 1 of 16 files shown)


crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -586,6 +580,35 @@ async fn graphql_handler(
     (extensions, result.into())
 }
 
+pub(crate) fn get_write_api<'ctx>(
+    ctx: &'ctx Context<'_>,
+) -> Result<&'ctx OptimisticWriteApi, Error> {
+    let write_api: &Option<OptimisticWriteApi> = ctx
+        .data()
+        .map_err(|_| Error::Internal("Unable to fetch IOTA SDK client".to_string()))?;
```
tomxey: Error messages in this function should be adjusted, this is no longer an SDK client.

crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -83,6 +85,29 @@ impl Query {
             .extend()
     }
 
+    async fn move_view_call(
```
tomxey, line 88: Could we have some automated test for this?

crates/iota-graphql-rpc/src/types/move_view_result.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use async_graphql::*;
+use iota_json_rpc_types::{IotaMoveValue, IotaMoveViewCallResults};
+use serde::Serialize;
+
+use crate::{error::Error, types::json::Json};
+
+/// The result of a move-view function call.
```
tomxey, line 10: Would it make sense to briefly describe here how the fields would be filled depending on execution results? (I would expect that either error or results will be filled, but not both, but we could make it explicit in the doc)

kodemartin: 7bca8228c0

crates/iota-graphql-rpc/src/server/builder.rs
kodemartin: 7bca8228c0

crates/iota-graphql-rpc/src/types/query.rs
kodemartin: Sadly, `iota-graphql-e2e-tests` doesn't support queries that pass through in the json-rpc. This would make it straight-forward to test.

Then to test this e2e would require duplicating a lot of the logic we have in place in the indexer, which seems like an overkill in this case. Given that this is a wrapper of the `OptimisticWriteApi`, which is tested in the indexer, and there is already a test for its construction while initializing the graphql server, we can then rely on the type system for correctness.

crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -586,6 +580,35 @@ async fn graphql_handler(
...
+    headers.insert(CLIENT_SDK_TYPE_HEADER, HeaderValue::from_static("graphql"));
+
+    let builder = HttpClientBuilder::default()
+        .max_request_size(2 << 30)
+        .set_headers(headers.clone())
+        .request_timeout(RPC_TIMEOUT_ERR_SLEEP_RETRY_PERIOD)
+        .max_concurrent_requests(MAX_CONCURRENT_REQUESTS);
+
+    builder.build(rpc_client_url).map_err(|e| {
+        warn!("Failed to get new Http client with error: {:?}", e);
```
sergiupopescu199:
```suggestion
        warn!("Failed to get new Http client with error: {e:?}");
```

crates/iota-graphql-rpc/src/types/move_view_result.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl TryFrom<IotaMoveValue> for Json {
+    type Error = Error;
+
+    fn try_from(value: IotaMoveValue) -> Result<Self, Error> {
+        let json = serde_json::to_value(value).map_err(|e| Error::Internal(e.to_string()))?;
+
+        Ok(Self(
+            Value::try_from(json).map_err(|e| Error::Internal(e.to_string()))?,
+        ))
```
sergiupopescu199:
````suggestion
        Value::try_from(json)
            .map_err(|e| Error::Internal(e.to_string()))
            .map(Self)
````

File: crates/iota-graphql-rpc/src/server/builder.rs [hunk 6]
```diff
@@ -586,6 +580,35 @@ async fn graphql_handler(
...
+) -> Result<&'ctx OptimisticWriteApi, Error> {
+    let write_api: &Option<OptimisticWriteApi> = ctx
+        .data()
+        .map_err(|_| Error::Internal("Unable to get node execution interface".to_string()))?;
+    write_api
+        .as_ref()
+        .ok_or_else(|| Error::Internal("Node execution interface not initialized".to_string()))
+}
+
+fn get_json_rpc_client(rpc_client_url: &str) -> Result<HttpClient, Error> {
```
sergiupopescu199: nit: Maybe we could rename it to: `new_json_rpc_client` or `build_json_rpc_client`? The `get_` prefix is a bit misleading since we're not getting an existing value.

crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -463,33 +466,25 @@ impl ServerBuilder {
...
+                optimistic_tx_executor,
+            );
+            Some(write_api)
         } else {
             warn!(
                 "No fullnode url found in config. `dryRunTransactionBlock` and `executeTransactionBlock` will not work"
             );
-            (None, None)
+            None
         };
```
sergiupopescu199: I'm wondering if we could add the `context_data` only if we have the `write_api`, instead of wrapping it in an `Option`.

something like this:
```rust
if let Some(url) = &config.tx_exec_full_node.node_rpc_url {
            let json_rpc_client = get_json_rpc_client(url)?;
            let indexer_store = PgIndexerStore::new(reader.get_pool(), indexer_metrics.clone());
            let optimistic_tx_executor = OptimisticTransactionExecutor::new(
                url,
                reader.clone(),
                indexer_store,
                indexer_metrics.clone(),
            );
            let write_api = OptimisticWriteApi::new(
                WriteApi::new(json_rpc_client, reader),
                optimistic_tx_executor,
            );

            builder = builder.context_data(write_api)
        } else {
            warn!(
                "No fullnode url found in config. `dryRunTransactionBlock` and `executeTransactionBlock` will not work"
            );
        };
```

and the `get_write_api` could look like this:

```rust
pub(crate) fn get_write_api<'ctx>(
    ctx: &'ctx Context<'_>,
) -> Result<&'ctx OptimisticWriteApi, Error> {
    ctx.data_opt()
        .ok_or_else(|| Error::Internal("Node execution interface not initialized".to_string()))
}
```

sergiupopescu199 (approved):
lgtm! ✨

left some  optional suggestions and nits.

crates/iota-graphql-rpc/src/types/query.rs
tomxey: If it's too problematic then let's leave it as is.
Thanks for getting into this though :pray:

tomxey (approved):

crates/iota-graphql-rpc/src/server/builder.rs
kodemartin: These are both very good suggestions. I integrated them with 8d1cd55e9a.

I diverged only in requiring the fullnode url and thus keeping separated the constructors from the `context_data` insertion side-effects.

crates/iota-graphql-rpc/src/server/builder.rs
kodemartin: Done with 8d1cd55e9a

Issue #8364, opened 2025-08-26
mysticeti: remove mononicity enforcement for timestamps


polinikita:
We can clean the code by removing all the checks around monotonicity for timestamps (e.g. authority_service, block_manager). These checks create more problems than they solve. So, we should keep timestamps. The commit timestamp should be deterministically determined as a median across the parents and the monotonicity should be enforced in this case only.

polinikita:
Relevant upstream PRs include https://github.com/MystenLabs/sui/pull/21347, https://github.com/MystenLabs/sui/pull/21530,
https://github.com/MystenLabs/sui/pull/22392,
https://github.com/MystenLabs/sui/pull/22392,
https://github.com/MystenLabs/sui/pull/21792,
https://github.com/MystenLabs/sui/pull/21704

PR #8756, opened 2025-10-03
upstream(consensus): Block and commit timestamp refactor

piotrm50:
# Description of change

Porting upstream change: https://github.com/MystenLabs/sui/pull/21347 and https://github.com/MystenLabs/sui/pull/21530

> This PR is doing the following:
> * Is dropping the timestamp monotonicity block ancestor checks. We'll no longer check whether a block's ancestors timestamps are strictly lower than the block in question.
> * Calculating the commit's timestamp by using the leader's strong ancestors timestamps median. The calculation is stake based.

Partly covers #8364 

### Release Notes

- [x] Protocol: Dropping the timestamp monotonicity block ancestor checks and performing stake-based calculation of the commit's timestamp by using the leader's strong ancestors timestamps median. Those changes are enabled only on the devnet. 
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


consensus/core/src/block_manager.rs
```diff
@@ -301,13 +301,39 @@ impl BlockManager {
...
+        // Insert the accepted blocks into DAG state so future blocks including them as
+        // ancestors do not get suspended.
+        self.dag_state
+            .write()
+            .accept_blocks(blocks_to_accept.clone());
+
+        blocks_to_accept
+    }
+
+    // TODO: remove once timestamping is refactored to the new approach.
```
Thoralf-M, line 331: Isn't this PR the refactor to the new approach?

consensus/core/src/context.rs
```diff
@@ -102,23 +102,43 @@ impl Context {
 /// `[Clock]` cloneable to ensure that a single instance is shared behind an
 /// `[Arc]` wherever is needed in order to make sure that consecutive calls to
 /// receive the system timestamp will remain monotonically increasing.
-pub(crate) struct Clock {
+pub struct Clock {
     initial_instant: Instant,
     initial_system_time: SystemTime,
+    // `clock_drift` should be used only for testing
```
filipdulic, line 108: if clock drift is only used for testing, then use `#[cfg(test)]` for this field, as well as `#[cfg(not(test))]` for the standard `Default` implementation as well as `#[cfg(test)]` for the test `default` implementation

filipdulic (approved):

consensus/core/src/block_manager.rs
Thoralf-M: Or will there be another PR which addresses this?

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_13.snap
```diff
@@ -34,6 +34,7 @@ feature_flags:
   normalize_ptb_arguments: true
   select_committee_from_eligible_validators: true
   track_non_committee_eligible_validators: true
+  consensus_median_based_commit_timestamp: true
```
jkrvivian: What about other snapshot files? `iota_protocol_config__test__Testnet_version_13.snap` and `iota_protocol_config__test__Mainnet_version_13.snap`

jkrvivian (changes requested):

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_13.snap
piotrm50: on those networks the feature will not be enabled yet.

consensus/core/src/context.rs
piotrm50: done.

consensus/core/src/block_manager.rs
piotrm50: yes, but this is until all networks have fully switched to use the new approach, we can't remove the old code.

jkrvivian (approved):

PR #8757, opened 2025-10-03
upstream(iota-core): Force checkpoint timestamp to be non decreasing

piotrm50:
# Description of change

Porting upstream change: https://github.com/MystenLabs/sui/pull/21792

> First consensus commit in an epoch does not have knowledge of the last consensus commit of the previous epoch right now. So the first commit timestamp may be less than the previous commit timestamp. Adding this logic to ensure checkpoints have non decreasing timestamps.

Partly covers #8364 

### Release Notes

- [x] Protocol: Force checkpoint timestamp to be non decreasing. This is only enabled on the devnet.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


polinikita (approved):
While two feature flags serve for different purposes, I believe one can use only one as we plan to integrate them simultaneously at all stages.

filipdulic (approved):

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_13.snap
```diff
@@ -35,6 +35,7 @@ feature_flags:
   select_committee_from_eligible_validators: true
   track_non_committee_eligible_validators: true
   consensus_median_based_commit_timestamp: true
+  enforce_checkpoint_timestamp_monotonicity: true
```
jkrvivian: What about snapshot files `iota_protocol_config__test__Testnet_version_13.snap` and `iota_protocol_config__test__Mainnet_version_13.snap`?

jkrvivian (changes requested):

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_13.snap
piotrm50: feature is not yet enabled on those networks

Thoralf-M (approved):

jkrvivian (approved):

PR #8759, opened 2025-10-03
feat: add `AuthenticatorInfoV1Key` instead of `AUTHENTICATOR_DF_NAME`

### LINKED ISSUE #8754: [AA] Replace the AUTHENTICATOR_DF_NAME constant with a type
In `iota::account` the string AUTHENTICATOR_DF_NAME is used as key to look for `AuthenticatorInfoV1` instances in objects' dynamic fields. 
This should be changed, and a proper type should be used instead:
```move
public struct AuthenticatorInfoV1Key has copy, drop, store {}
```
(name versioning will be re-evaluated later here #8498)

This issue tracks also all the required refactoring needed in `iota-core`, `iota-types` and execution crates. It also tracks changes to examples found in the `vm-lang/aa-auth/8116-feature-branch` branch #8533 and e2e tests #8669.

valeriyr:
# Description of change

In `iota::account`, the string `AUTHENTICATOR_DF_NAME` is replaced with a new type `AuthenticatorInfoV1Key`.

fixes #8754


examples/move/iota_account/sources/iota_account.move
```diff
@@ -227,16 +228,9 @@ fun ensure_tx_sender_is_account(self: &IOTAccount, ctx: &TxContext) {
 }
 
 /// Checks if `name` is allowed to be used for a user-defined dynamic field.
-fun check_reserved_df_name<Name: copy + drop + store>(name: &Name) {
+fun check_reserved_df_name<Name: copy + drop + store>() {
```
valeriyr: This function can be removed because a user can't create an `OwnerPublicKey` instance to use it as a df name.

valeriyr: removed in abc37b1409

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -42,17 +49,56 @@ public fun create_auth_info_v1(
...
+/// Attach the `authenticator` instance to the account.
+/// It will be added as a dynamic field specified by the `AuthenticatorInfoV1Key` name.
+public fun attach_auth_info_v1(account_id: &mut UID, authenticator: AuthenticatorInfoV1) {
+    assert!(!has_auth_info_v1(account_id), EAuthenticatorInfoV1AlreadyAttached);
+    dynamic_field::add(account_id, auth_info_v1_key(), authenticator);
+}
+
+/// Detach the account-related authenticator from the account.
+/// The dynamic field specified by the `AuthenticatorInfoV1Key` name will be removed.
+public fun detach_auth_info_v1(account_id: &mut UID): AuthenticatorInfoV1 {
```
miker83z: At the higher feature level, we might not want to allow to detach an auth info and make an account locked forever. Is there a use case where this function might be needed? Otherwise I'd drop it

TheMrAI (approved):

examples/move/iota_account/sources/iota_account.move
```diff
@@ -132,6 +118,20 @@ public fun has_field<Name: copy + drop + store>(self: &IOTAccount, name: Name):
     dynamic_field::exists_(&self.id, name)
 }
 
+/// Borrows a reference to the attached `AuthenticatorInfoV1` instance.
+/// This function is not gated to be called only by the account,
+/// anybody can call it to read the account dynamic fields.
+public fun borrow_auth_info_v1(self: &IOTAccount): &AuthenticatorInfoV1 {
+    account::borrow_auth_info_v1(&self.id)
+}
```
miker83z: We don't need this, we can use `iota::object::borrow_uid()` on the object and then call `iota::account::borrow_auth_info_v1()` on the returned value

valeriyr: `iota::object::borrow_uid()` is a private function; it seems to be impossible.

crates/iota-framework/packages/iota-framework/sources/account.move
valeriyr: removed 78b7a421ad

examples/move/iota_account/sources/iota_account.move
valeriyr: Just to share the idea for reviewers, the `account::*_auth_info_v1` functions are practically gated to be called in the module where an account is defined by having `UID` as the first parameter. 
Usually, `UID` is not exposed to the outside through the module public interface because it allows changing the dynamic fields, or receiving `Receiving` objects, etc.
So if you expose the `UID`, you are ok that your account internals can be changed.

miker83z: Yes, right. So it might be safe to just expose the UID as readonly ref:
```move
public fun uid(self: &IOTAccount): &UID {
    &self.id
}
``` 
Like in kiosk: https://github.com/iotaledger/iota/blob/89a0f0e5f1c080baa7f1d7735d2c2d50ab83e9ef/crates/iota-framework/packages/iota-framework/sources/kiosk/kiosk.move#L528

miker83z (approved):

PR #8760, opened 2025-10-03
feat(consensus): track useful authorities during block bundle exchange

### LINKED ISSUE #8411: starfish: improve bandwidth usage by implementing enhanced cordial dissemination
When receiving a BlockBundle via streaming from a Peer X, we can track block headers from which authorities were useful or missing when including into the local DAG.

Then when streaming our next blocks to this peer, we can send this additional information such that the peer X sends only useful blocks to us.

filipdulic:
# Description of change

This change introduces tracking of useful authorities during block bundle exchange

- Added `useful_authorities_from_peer` and `useful_authorities_to_peer` to `AuthorityService` for tracking which authorities are considered useful when exchanging bundles with peers.
- Extended `BlockBundle` with a new `useful_authorities` field and updated serialization (`SerializedBlockBundleParts`) to encode/decode this via a [u64; 4] bitmask.
- Refactored DagState to expose:
  - take_unknown_block_refs_for_authority
  - take_unknown_headers_for_authority
  - take_useful_headers_for_authority
- Updated network handling to:
  - Cache peer-reported useful authorities.
  - Record locally observed useful authorities when receiving bundles.
  - Prefer sending only useful headers when peers specify them.
- Added a test to validate bitmask to/from `BTreeSet` conversion.

fixes #8411

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{BTreeMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     pin::Pin,
     sync::Arc,
     time::Duration,
@@ -114,6 +114,14 @@ pub(crate) struct AuthorityService<C: CoreThreadDispatcher> {
     received_block_headers: FilterForHeaders,
     /// Sender to send received transaction messages to the shard reconstructor
     transaction_message_sender: Sender<Vec<TransactionMessage>>,
+    /// For each peer, the set of authorities whose block headers the local node
+    /// considered useful when receiving a block bundle from that peer.
+    /// Keyed by the peer’s AuthorityIndex.
+    useful_authorities_from_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
+    /// For each peer, the set of local authorities that the peer reported as
+    /// useful to them (communicated inside their block bundles).
+    /// Keyed by the peer’s AuthorityIndex.
+    useful_authorities_to_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
 }
 
 impl<C: CoreThreadDispatcher> AuthorityService<C> {
@@ -146,6 +154,8 @@ impl<C: CoreThreadDispatcher> AuthorityService<C> {
             dag_state,
             store,
             received_block_headers: FilterForHeaders::new(),
+            useful_authorities_from_peer: Arc::new(RwLock::new(BTreeMap::new())),
+            useful_authorities_to_peer: Arc::new(RwLock::new(BTreeMap::new())),
             transaction_message_sender,
         }
     }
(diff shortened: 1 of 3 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -157,9 +167,17 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         fail_point_async!("consensus-rpc-response");
 
         let peer_hostname = &self.context.committee.authority(peer).hostname;
-        // 1. Create a verified block and make some preliminary checks
         let serialized_block_bundle_parts =
             SerializedBlockBundleParts::try_from(serialized_block_bundle)?;
+
+        // Cache authorities this peer finds useful for cordial
```
polinikita, line 178: The sentence seems not finished

crates/starfish/core/src/dag_state.rs
```diff
@@ -1277,6 +1285,51 @@ impl DagState {
...
+    /// references into full `VerifiedBlockHeader`s.
+    ///
+    /// Panics if any of the block headers are missing from `DagState` or disk.
+    pub(crate) fn take_useful_headers_for_authority(
+        &mut self,
+        authority_index: AuthorityIndex,
+        round_upper_bound_exclusive: Round,
+        useful_authorities: BTreeSet<AuthorityIndex>,
+    ) -> Vec<VerifiedBlockHeader> {
+        let block_refs = self
```
VorobyevIlya: we apply upper bound MAX_HEADERS_PER_BUNDLE before filtering out useless headers, so the number of headers included in a bundle can be significantly smaller than the bound.

crates/starfish/core/src/authority_service.rs
```diff
@@ -111,6 +111,14 @@ pub(crate) struct AuthorityService<C: CoreThreadDispatcher> {
...
     /// evicted when the threshold is exceeded
     received_block_headers: FilterForHeaders,
+    /// For each peer, the set of authorities whose block headers the local node
+    /// considered useful when receiving a block bundle from that peer.
+    /// Keyed by the peer’s AuthorityIndex.
+    useful_authorities_from_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
+    /// For each peer, the set of local authorities that the peer reported as
+    /// useful to them (communicated inside their block bundles).
+    /// Keyed by the peer’s AuthorityIndex.
+    useful_authorities_to_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
```
VorobyevIlya, line 124: As far as I understand, these two structures of useful authorities can only grow. I think we need to have some timers for eviction. Like, if we needed header 1 from validator 2 in round 100, but then we didn't really need it for the next 100 rounds, we can stop asking for it.

crates/starfish/core/src/authority_service.rs
```diff
@@ -548,8 +565,22 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
             .map_err(|_| ConsensusError::Shutdown)?;
 
+        // 12. update `useful_authorities_from_peer`
+        // create a set of authority indexes from the `additional_block_headers`
+        // and `missing_ancestors`
+        let useful_authorities = additional_block_headers
+            .iter()
+            .map(|block_header| block_header.author())
+            .chain(missing_ancestors.iter().map(|block_ref| block_ref.author))
+            .collect::<BTreeSet<_>>();
```
VorobyevIlya, line 594: I don't think that we need to include missing headers in such a way. Imagine that validator 1 sends his block B to everybody except validator 2. Then validator 2 received blocks referencing block B from everyone, so block B will be in missing_ancestors for all peers, and then validator 2 starts to request headers of validator 1 blocks from everyone.
Maybe with eviction it won't be a problem

crates/starfish/core/src/authority_service.rs
filipdulic: they are caped at Committee size ^ 2. 
The number of keys is at max the size of the committee, and each set can at maximum be committee size.

crates/starfish/core/src/authority_service.rs
filipdulic: but at some point, someone will send him the block header he's been missing, from Validator 1, which he will process, and in the same round someone else will send the same block header, which he will filter out as already processed and find not usefull, and thus will not be finally added to the `additional_block_headers` or `missing_ancestors`

VorobyevIlya: oh, I misunderstand something, this is fine

VorobyevIlya (approved):

polinikita (approved):

PR #8762, opened 2025-10-03
upstream(consensus,iota-core): Start epoch from last checkpoint timestamp

### LINKED ISSUE #8364: mysticeti: remove mononicity enforcement for timestamps

piotrm50:
# Description of change

Porting upstream PR: https://github.com/MystenLabs/sui/pull/21804

>This makes the 1st consensus commit timestamp of an epoch the same as the last checkpoint timestamp of the previous epoch.
>
> Reusing the `enforce_checkpoint_timestamp_monotonicity` feature flag, since both changes should roll out with the same schedule.

Fixes #8364 


consensus/core/src/dag_state.rs
```diff
@@ -118,7 +118,7 @@ impl DagState {
         let cached_rounds = context.parameters.dag_state_cached_rounds as Round;
         let num_authorities = context.committee.size();
 
-        let genesis = genesis_blocks(context.clone())
+        let genesis = genesis_blocks(context.as_ref())
```
filipdulic: consistency:
in the file `core.rs` you used `&context` and here you use `context.as_ref()`.

filipdulic: same in `test_dag.rs` and `test_dag_builder.rs`

filipdulic (approved):

jkrvivian (approved):

Thoralf-M (approved):

jkrvivian (approved):

PR #8765, opened 2025-10-06
fix(wallet): construct correct url when opening a new tab

### LINKED ISSUE #8684: Send token form with passkey throws an error
In wallet popup mode, when clicking the send button a new tab is opened. In this new tab, for some reason, the amount input always throws an error `Amount must be less than 0`. Probably some query state not being sent.,

Image

VmMad:
# Description of change

What happened was a wrong construction of the url when opening a new tab. For the Send Token Page we receive a `type` query param with the `type` of the coin (`IOTA_TYPE_ARG`) but when opening a new tab, another query param was also passed, `type=tab` which was intended to replace the `type=popup` param. This PR fixes the placement of the `type` param so that it doesn't conflict with the coin type param

Fixes #8684


brancoder (changes requested):
we need to add new tab for creating profile flow

KeitoTadashi (approved):

brancoder (approved):

PR #8780, opened 2025-10-06
feat(wallet): add amplitude event for Passkey click

### LINKED ISSUE #8628: Add amplitude events for passkeys
## Description

add passkeys to amplitude for tracking users events

Juligs:
fixes #8628


msarcev (approved):

msarcev (changes requested):
Conflicts. Probably a rebase is good.

apps/wallet/src/shared/analytics/ampli/index.ts
```diff
@@ -356,6 +374,12 @@ export interface UnpinnedCoinProperties {
 }
 
 export interface UnstakedIotaProperties {
+    /**
+     * | Rule | Value |
+     * |---|---|
+     * | Type | number |
+     */
+    stakedAmount?: number;
```
panteleymonchuk: we might have a conflicts because of this property. Better to incluce to the PR only chances that you need

apps/wallet/src/shared/analytics/ampli/index.ts
```diff
@@ -158,6 +167,15 @@ export interface ClickedImportPrivateKeyProperties {
     sourceFlow?: string;
 }
 
+export interface ClickedImportSeedProperties {
```
panteleymonchuk: same here

panteleymonchuk (changes requested):

apps/wallet/src/ui/app/helpers/openInNewTab.ts
brancoder, line 1: we should probably delete this file. i think it slipped in when resolving conflicts

brancoder (changes requested):
LG just a file to remove

PR #8787, opened 2025-10-07
refactor: use `IOTAccount` in `dynamic_multisig_account`

valeriyr:
# Description of change

`IOTAccount` is used in the `dynamic_multisig_account` example.


examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -111,38 +94,33 @@ public fun total_approves(self: &DynamicMultisigAccount, transaction_digest: vec
...
     transaction_digest: vector<u8>,
     ctx: &TxContext,
 ) {
     // Get the member who proposed the transaction.
-    let member_address = *self.members().borrow(ctx.sender()).addr();
+    let member_address = *members(self).borrow(ctx.sender()).addr();
 
     // Store the transaction.
-    self.transactions_mut().add(transaction_digest, member_address);
+    transactions_mut(self, ctx).add(transaction_digest, member_address);
```
valeriyr, line 110: A transaction can be proposed by any account's member. In the current `IOTAccount` implementation, it is impossible; we can borrow mutable dynamic fields only when the sender is the account itself.

miker83z:
Will be continuing in a new PR

PR #8788, opened 2025-10-07
feat(iota-protocol-config,iota-core): Add feature flag for selecting committee supporting next epoch version

piotrm50:
# Description of change

Add a separate feature flag for selecting committee supporting next epoch version instead of functionality being covered by `track_non_committee_eligible_validators` flag that controls sending/processing of AuthorityCapabilities notifications from non-committee active validators. 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -4909,9 +4909,9 @@ impl AuthorityState {
             let mut eligible_active_validators = (0..active_validators.len() as u64).collect();
 
             // Use validators supporting the target protocol version as eligible validators
-            // in the next version if track_non_committee_eligible_validators feature flag
-            // is set to true.
-            if config.track_non_committee_eligible_validators() {
+            // in the next version if select_committee_supporting_next_epoch_version feature
+            // flag is set to true.
+            if config.select_committee_supporting_next_epoch_version() {
                 eligible_active_validators = Self::get_validators_supporting_protocol_version(
                     next_epoch_protocol_version,
                     next_epoch_protocol_digest,
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -335,11 +335,21 @@ struct FeatureFlags {
     select_committee_from_eligible_validators: bool,
 
     // If true, non-committee active validators will sign and send AuthorityCapabilitiesV1 to the
-    // committee. The committee will track them and include them in the epoch change. If this is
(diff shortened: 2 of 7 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -4909,9 +4909,9 @@ impl AuthorityState {
             let mut eligible_active_validators = (0..active_validators.len() as u64).collect();
 
             // Use validators supporting the target protocol version as eligible validators
-            // in the next version if track_non_committee_eligible_validators feature flag
-            // is set to true.
-            if config.track_non_committee_eligible_validators() {
+            // in the next version if select_committee_supporting_next_epoch_version feature
+            // flag is set to true.
+            if config.select_committee_supporting_next_epoch_version() {
```
Thoralf-M, line 4914: This change is correct? Shouldn't it check for both of the config flags? Or can these flags actually be merged to a single one?

alexsporn: What about reading the PR description 😝

alexsporn: And the getter for this expects the other to be true (but not the other way round)

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1363,6 +1373,18 @@ impl ProtocolConfig {
...
     }
+
+    pub fn select_committee_supporting_next_epoch_version(&self) -> bool {
+        let res = self
+            .feature_flags
+            .select_committee_supporting_next_epoch_version;
+        assert!(
+            !res || (self.track_non_committee_eligible_validators()
+                && self.select_committee_from_eligible_validators()),
+            "select_committee_from_eligible_validators_with_buffer requires track_non_committee_eligible_validators to be set"
```
muXxer: is this comment correct? what is `select_committee_from_eligible_validators_with_buffer` ?

crates/iota-core/src/authority.rs
Thoralf-M: > What about reading the PR description 😝

:runner:

Thoralf-M (approved):

muXxer (approved):

polinikita (approved):

crates/iota-protocol-config/src/lib.rs
piotrm50: good catch, it's some ai hallucination that I missed because the first part of the param was the same...

piotrm50: good catch, it's some ai hallucination that I missed because the first part of the param was the same...

PR #8792, opened 2025-10-07
feat(consensus): move `MAX_ACKNOWLEDGMENTS_PER_BLOCK` to `iota-protocol-config`

### LINKED ISSUE #8378: Move `MAX_ACKNOWLEDGMENTS_PER_BLOCK` to protocol params crate
Use `MAX_ACKNOWLEDGMENTS_PER_BLOCK`  in block verification.

filipdulic:
# Description of change

Moved `MAX_ACKNOWLEDGMENTS_PER_BLOCK` parameter used in starfish consensus to the `iota-protocol-config`.
- Increased max protocol version to 13.
- added `consensus_max_acknowledgments_per_block` parameter to `iota-protocol-config`
- added `consensus_max_acknowledgments_per_block_or_default` getter method to default to the currently used value.
- replaced usage of `MAX_ACKNOWLEDGMENTS_PER_BLOCK` in starfish to `protocol_config.consensus_max_acknowledgments_per_block_or_default`
- updated snapshots.

fixes #8378.

### Release Notes

- [x] Protocol:
  - Increased max protocol version to 13.
  - added `consensus_max_acknowledgments_per_block` parameter used in starfish consensus
  - added `consensus_max_acknowledgments_per_block_or_default` getter method to default to the currently used value.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -1108,6 +1108,12 @@ pub struct ProtocolConfig {
     /// Configures the garbage collection depth for consensus. When is unset or
     /// `0` then the garbage collection is disabled.
     consensus_gc_depth: Option<u32>,
+
+    /// Configures the maximum number of acknowledgments to be included in a
+    /// block. It must be reasonably larger than the number of validators
+    /// because not all validators create their blocks at the same pace.
+    /// Applicable only to `starfish` consensus.
+    consensus_max_acknowledgments_per_block: Option<u32>,
 }
 
 // feature flags
@@ -1268,6 +1274,10 @@ impl ProtocolConfig {
             "The consensus linearize sub dag V2 requires GC to be enabled"
         );
         res
+    }
+
+    pub fn consensus_max_acknowledgments_per_block_or_default(&self) -> u32 {
+        self.consensus_max_acknowledgments_per_block.unwrap_or(400)
     }
 
     pub fn variant_nodes(&self) -> bool {
@@ -1888,6 +1898,8 @@ impl ProtocolConfig {
             max_committee_members_count: None,
 
             consensus_gc_depth: None,
+
+            consensus_max_acknowledgments_per_block: None,
             // When adding a new constant, set it to None in the earliest version, like this:
             // new_constant: None,
         };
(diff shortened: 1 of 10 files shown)


crates/iota-protocol-config/src/lib.rs
```diff
@@ -73,6 +73,7 @@ pub const MAX_PROTOCOL_VERSION: u64 = 12;
 // Version 11: Framework fix regarding candidate validator commission rate.
 // Version 12: Enable the gas price feedback mechanism in all networks.
 //             Enable the normalization of PTB arguments.
+// Version 13: Add `consensus_max_acknowledgments_per_block` to config
```
piotrm50: Improve description so it's clear that it's for Starfish.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2146,6 +2158,9 @@ impl ProtocolConfig {
                     // Enable normalization of PTB arguments in all networks.
                     cfg.feature_flags.normalize_ptb_arguments = true;
                 }
+                13 => {
+                    cfg.consensus_max_acknowledgments_per_block = Some(400u32);
```
piotrm50: let's make sure that this is indeed the value that we want because changing this will require a new protocol version. @polinikita

crates/iota-protocol-config/src/lib.rs
filipdulic: added  `/// Applicable only to `starfish` consensus.` line to description.

jkrvivian (approved):

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2146,6 +2159,9 @@ impl ProtocolConfig {
                     // Enable normalization of PTB arguments in all networks.
                     cfg.feature_flags.normalize_ptb_arguments = true;
                 }
+                13 => {
```
piotrm50: We don't need this part because the default value is 400 anyway. We will include it in this section once we need to change the value from the default. And we cannot change the default value.

PR #8801, opened 2025-10-08
feat(starfish): send only useful shards

### LINKED ISSUE #8767: starfish: implement hinting mechanism for cordial dissemination of shards

VorobyevIlya:
# Description of change

PR tracks shards of which authors could be sent to other validators.

Shards for encoded transaction data of validator $v$ are considered as useful for validator $u$ if validator $u$ is ready to include an additional block header of validator $v$ after filtering those blocks. In particular, this means that
the potential sequenced transaction data comes from validator $v$ indirectly.

If the shard of the block created by the validator $v$ was useful, then the shards from the blocks created by $v$ will be sent for some number of rounds. In other words, there is some inertia; we don't stop sending shards as soon as one of them is useless.

In practice, there are two prerequisites for the efficiency of this mechanism:
1) block bundle streaming is actively working (e.g. this mechanism wouldn't work if a validator is catching up and doesn't produce blocks)
2) cordial dissemination of block headers is enabled (without additional block headers, validators won't understand which shards are useful for them)

fixes #8767 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -118,18 +118,19 @@ pub(crate) struct AuthorityService<C: CoreThreadDispatcher> {
     /// For each peer, the set of authorities whose block headers the local node
     /// considered useful when receiving a block bundle from that peer.
     /// Keyed by the peer’s AuthorityIndex.
-    useful_authorities_from_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
+    useful_headers_authors_from_peer:
+        Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
     /// For each peer, the set of local authorities that the peer reported as
     /// useful to them (communicated inside their block bundles).
     /// Keyed by the peer’s AuthorityIndex.
-    useful_authorities_to_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
+    useful_headers_authors_to_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
     /// For each peer, stores the latest round in which a received block bundle
     /// from any peer included a useful shard originating from a block
     /// created by authority `i`.
-    last_round_with_useful_shards_from_peer: Arc<RwLock<Vec<Round>>>,
+    last_round_with_useful_shards_by_block_author: Arc<RwLock<Vec<Round>>>,
     /// For each peer `i`, stores a set of authority indices representing the
     /// authors of blocks whose shards were reported as useful by peer `i`.
-    peers_reporting_useful_shards: Arc<RwLock<Vec<BTreeSet<AuthorityIndex>>>>,
+    useful_shards_authors_to_peer: Arc<RwLock<Vec<BTreeSet<AuthorityIndex>>>>,
 }
 
 impl<C: CoreThreadDispatcher> AuthorityService<C> {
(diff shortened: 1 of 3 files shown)


crates/starfish/core/src/dag_state.rs
```diff
@@ -41,6 +41,7 @@ use crate::{
 pub(crate) const MAX_TRANSACTIONS_ACK_DEPTH: Round = 50;
 pub(crate) const MAX_HEADERS_PER_BUNDLE: usize = 150;
 pub(crate) const MAX_SHARDS_PER_BUNDLE: usize = 150;
+const MAX_ROUND_GAP_FOR_USEFUL_SHARDS: usize = 5;
```
piotrm50, line 39: Please describe what this constant controls.

crates/starfish/core/src/dag_state.rs
```diff
@@ -1347,6 +1348,36 @@ impl DagState {
...
+                break;
+            }
+            if useful_authorities.contains(&block_ref.author) {
+                to_take.push(*block_ref);
+            }
+        }
+
+        // Remove the references from the set and update cordial knowledge
+        for block_ref in &to_take {
+            set.remove(block_ref);
```
piotrm50, line 1375: just to make sure: if the blocks are not removed from the `set` here then they will be evicted at some later point?

crates/starfish/core/src/dag_state.rs
```diff
@@ -1395,34 +1426,39 @@ impl DagState {
...
+        block_round: Round,
+    ) -> BTreeSet<AuthorityIndex> {
+        let mut useful_authorities = BTreeSet::new();
+        for (i, round) in last_useful_round.iter().enumerate() {
+            if block_round - round < MAX_ROUND_GAP_FOR_USEFUL_SHARDS as u32 {
+                useful_authorities.insert(AuthorityIndex::from(i as u8));
             }
         }
-        shards
+        useful_authorities
```
piotrm50: imo `useful_shards_authors` is better understandable than `useful_authorities`. wdyt? maybe we could also change it for headers?  Especially when a method is called `take_useful_shards_for_authority` and a param is called `useful_authorities` - `useful_shards_authors` would be more descriptive.

crates/starfish/core/src/network/mod.rs
```diff
@@ -255,21 +256,40 @@ pub(crate) struct SerializedBlockBundleParts {
...
+            let bit = bits.trailing_zeros() as usize;
+            set.insert(AuthorityIndex::from((base + bit) as u8));
+            bits &= bits - 1;
+        }
+    }
+    set
 }
 
 impl SerializedBlockBundleParts {
     pub(crate) fn useful_authorities(&self) -> BTreeSet<AuthorityIndex> {
```
piotrm50: better rename to `useful_headers_authors`

crates/starfish/core/src/authority_service.rs
```diff
@@ -122,6 +123,13 @@ pub(crate) struct AuthorityService<C: CoreThreadDispatcher> {
     /// useful to them (communicated inside their block bundles).
     /// Keyed by the peer’s AuthorityIndex.
     useful_authorities_to_peer: Arc<RwLock<BTreeMap<AuthorityIndex, BTreeSet<AuthorityIndex>>>>,
+    /// For each peer, stores the latest round in which a received block bundle
```
piotrm50, line 127: here also rename the header-related fields to be in line with the shard ones

crates/starfish/core/src/authority_service.rs
```diff
@@ -597,6 +620,20 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 self.useful_authorities_from_peer.write();
             useful_authorities_from_peer_write_guard.insert(peer, useful_authorities);
         }
+        {
+            let useful_shard_authors = additional_block_headers
```
piotrm50, line 643: variable is called `useful_shard_authors` but we iterate over `additional_block_headers`? is this correct?

Also, add a comment describing what this block of code does, or update comment for step 13. above.

piotrm50 (changes requested):
Please update header-related variables so the naming is header-specific rather than general `useful_authorities`.

crates/starfish/core/src/dag_state.rs
VorobyevIlya: yes, they are removed in evict_cordial_knowledge function

crates/starfish/core/src/dag_state.rs
VorobyevIlya: done

crates/starfish/core/src/dag_state.rs
VorobyevIlya: yes, useful_shard_authors is better. Changed it.

crates/starfish/core/src/network/mod.rs
VorobyevIlya: done

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Yes, these headers went through filters and will now be added to dag_state without transaction data, so we request their shards.
Added comment

crates/starfish/core/src/authority_service.rs
VorobyevIlya: did renaming

polinikita (approved):

piotrm50 (approved):

PR #8821, opened 2025-10-09
feat(wallet): export public keys in passkeys accounts

### LINKED ISSUE #8820: Export public account key in passkeys accounts
Image

passkeys accounts should not be able to export account keys

KeitoTadashi:
# Description of change

disable export account keys in passkeys accounts

fixes #8820 


VmMad (changes requested):
I think it's better to do changes in `apps/wallet/src/ui/app/pages/accounts/ExportAccountPage.tsx`, as you should be able to see the public key there, just like with the ledger

brancoder (changes requested):
conflicts after rebase 🙏

apps/wallet/src/ui/app/pages/accounts/ExportAccountPage.tsx
```diff
@@ -38,7 +39,12 @@ export function ExportAccountPage() {
         return <Navigate to="/accounts/manage" replace />;
     }
 
-    const publicKey = account?.publicKey ? new Ed25519PublicKey(account.publicKey) : null;
+    const publicKey =
+        account?.publicKey && !isPasskeysAccount
+            ? new Ed25519PublicKey(account.publicKey).toIotaPublicKey()
+            : isPasskeysAccount
+              ? account?.publicKey
+              : undefined;
```
marc2332: Why is it needed to check for passkeys here? Is it because the public key format is different or what?

marc2332 (changes requested):

apps/wallet/src/ui/app/pages/accounts/ExportAccountPage.tsx
KeitoTadashi: Exactly, passkeys is PasskeyPublicKey with size = 33, while Ed25519PublicKey is size = 32, this make passkeys incompatible with Ed25519PublicKey

I could (maybe) improve code with this (lmk wyt):

```ts
    const publicKey = account?.publicKey
        ? isPasskeyAccount
            ? new PasskeyPublicKey(account.publicKey)
            : new Ed25519PublicKey(account.publicKey)
        : null;
```

this block of code returns a pubkey like:
`BgJtB+Mwjc9kKM9rMdtAB2MvNF/y6BrUxatmy9cUqdvY7g==`

while how it is now returns:
`Am0H4zCNz2Qoz2sx20AHYy80X/LoGtTFq2bL1xSp29ju`

marc2332 (approved):

brancoder (approved):
LGTM

PR #8645, opened 2025-09-24
feat(graphql-rpc): Read optimistic transactions in graphql

tomxey:
# Description of change

Transaction effects are now available immediately after execution through `Query.transactionBlock`.
However, other queries that depend on the chain’s indexed state (e.g., address-level balance updates) may still lag until the transaction has been checkpointed.
To confirm that a transaction has been included in a checkpoint, user can query `Query.transactionBlock` and check whether the `effects.checkpoint` field is set (or `null` if not yet checkpointed).

## Links to any relevant issues

fixes https://github.com/iotaledger/iota/issues/8418

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

- [ ] Synchronization of the indexer from genesis for a network including migration objects.
- [ ] Restart of indexer synchronization locally without resetting the database.
- [ ] Restart of indexer synchronization on a production-like database.
- [ ] Deployment of services using Docker.
- [ ] Verification of API backward compatibility.

### Release Notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL:
    Transaction effects are now available immediately after execution via `Query.transactionBlock`.
    Note that other queries relying on the chain’s indexed state (such as address-level balance updates) may still show a delay until the transaction is checkpointed.
    To verify checkpoint inclusion, query `Query.transactionBlock` and check the `effects.checkpoint` field — it will be `null` until the transaction has been checkpointed.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

PR #8828, opened 2025-10-10
feat(ts-sdk): Add `waitMode` in `waitForTransaction`

marc2332:
Based off https://github.com/iotaledger/iota/pull/8645

Allows to specify how to wait the transaction.

[run-ci]

### DIFF
--- a/sdk/graphql-transport/test/compatibility.test.ts
+++ b/sdk/graphql-transport/test/compatibility.test.ts
@@ -59,8 +59,7 @@ describe('GraphQL IotaClient compatibility', () => {
 
         await toolbox.client.waitForTransaction({
             digest: transactionBlockDigest,
-            waitForCheckpoint: true,
-            waitForIndex: true,
+            waitMode: 'checkpoint',
         });
     });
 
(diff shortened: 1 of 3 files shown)


sdk/graphql-transport/test/compatibility.test.ts
```diff
@@ -870,37 +824,28 @@ describe('GraphQL IotaClient compatibility', () => {
...
-
-                // eslint-disable-next-line no-empty
-            } catch (e) {}
-        }
-        expect(result).toEqual(true);
+        transaction = await toolbox.client.waitForTransaction({
+            digest: transaction.digest,
+            waitForIndex: true,
+        });
+        expect(transaction.checkpoint).toBeNull();
```
tomxey: FYI: it's likely, but not guaranteed. so the test may be flaky

sdk/typescript/src/client/client.ts
```diff
@@ -940,7 +946,21 @@ export class IotaClient {
...
+                    const isIndexedOnNode = await this.isTransactionIndexedOnNode({
+                        digest: input.digest
+                    });
+                    if(!isIndexedOnNode) {
+                        continue;
+                    }
+                }
+                const transaction = await this.getTransactionBlock(input);
+
+                if (waitForCheckpoint && transaction.checkpoint) {
```
tomxey: If user requests for "waitForCheckpoint" and "waitForIndex", it would be enough to just wait for checkpoint, waiting for "isTransactionIndexedOnNode" in such case is not needed.

We could save one request that way in some cases.

sdk/graphql-transport/test/compatibility.test.ts
```diff
@@ -870,37 +823,24 @@ describe('GraphQL IotaClient compatibility', () => {
...
-                    });
-                }
-
-                // eslint-disable-next-line no-empty
-            } catch (e) {}
-        }
-        expect(result).toEqual(true);
+        transaction = await toolbox.client.waitForTransaction({
+            digest: transaction.digest,
+            waitMode: 'indexed',
```
tomxey: The endpoint in indexer is called isTransactionIndexedOnNode, and the "OnNode" part is significant. Since the request is about the indexing status of the tx on the fullnode, not in the indexer itself. Just using the term "indexed" may be confusing so I would suggest to somehow include the "OnNode" part in the waitMode name.

marc2332: True, my bad

tomxey (approved):
lgtm 🚀

brancoder (approved):
LGTM

PR #8836, opened 2025-10-10
refactor(iota-data-ingestion-core): use gRPC instead of REST API to sync checkpoints from fullnode

### LINKED ISSUE #7923: [data-ingestion]: support gRPC streams as checkpoint source
## Description

Following [merging of https://github.com/iotaledger/iota/tree/feat/grpc](https://github.com/iotaledger/iota/pull/9703) we need to refactor downstream data-infrastructure libs/apss to use the gRPC streams.

The new implementation should support in a seamless way both the `rest-api` and the `grpc-api`.

Node operators may choose to `config.enable_{rest,grpc}_api`. Anyone that spawned a node to use as a checkpoint source it has `enable_rest_api = true`. Such users should be supported by all APIs in `iota-data-ingestion-core` until the rest-api is deprecated.

This means that we have to introduce a way to handle both cases:
* If `grpc` is enabled, `V2`  should use it, otherwise
* fallback to `rest`

sergiupopescu199:
# Description of change

This PR adds the ability for the ingestion core to receive a stream of checkpoints trough a gRPC connection from fullnode instead of polling from the REST API.

fixes #7923 

- used the`testnet` network  and also started a local network to test the changes, both tests were conducted for several hours.
- the checkpoints are being downloaded (received form the stream) reliably, no gaps are present, in case of connection loss, the stream is terminated, once the connection is available it resumes downloading the checkpoints (start/stop VPN) all done automatically as prior to REST API implementation

> [!NOTE]
> A known issue is regarding downloading checkpoints which exceeds the gRPC limit of 4MB, which is not possible at the moment. The Node team is working on providing a chunking ability to stream large size checkpoints, like the genesis one.

> [!NOTE]
> The following patch does not affect the indexer, it uses the old checkpoint reader impl which uses the REST API to download checkpoints. The patch introduced the changes on the new checkpoint reader (v2).

### Release Notes
- [x] Indexer: `iota-data-ingestion-core` CheckpointReader V2 use gRPC instead of REST API to sync checkpoints from fullnode

### DIFF
--- a/crates/iota-data-ingestion-core/src/reader/v2.rs
+++ b/crates/iota-data-ingestion-core/src/reader/v2.rs
@@ -290,6 +290,25 @@ impl CheckpointReaderActor {
         Ok(())
     }
 
+    /// Fetch checkpoints from the live object store and stream them to a
+    /// channel.
+    async fn fetch_live(
+        &mut self,
+        batch_size: usize,
+        live: &dyn ObjectStore,
+    ) -> IngestionResult<()> {
+        let mut checkpoint_stream = (self.current_checkpoint_number..u64::MAX)
+            .map(|checkpoint_number| fetch_from_object_store(live, checkpoint_number))
+            .pipe(futures::stream::iter)
+            .buffered(batch_size);
+        while let Some(checkpoint_result) = checkpoint_stream.next().await {
+            let (checkpoint, size) = checkpoint_result?;
+            self.send_remote_checkpoint_with_capacity_check(checkpoint, size)
+                .await?;
+        }
+        Ok(())
+    }
+
     /// Fetches checkpoints from the fullnode trough a gRPC streaming connection
     /// and stream them to a channel.
     async fn fetch_from_fullnode(&mut self, client: &mut CheckpointClient) -> IngestionResult<()> {
@@ -329,10 +348,7 @@ impl CheckpointReaderActor {
             return Err(IngestionError::Grpc("checkpoint stream was closed".into()));
         };
 
-        let checkpoint: CheckpointData = genesis_checkpoint
-            .map(GrpcCheckpoint)
-            .map_err(|e| IngestionError::Grpc(e.to_string()))?
-            .try_into()?;
+        let checkpoint: CheckpointData = genesis_checkpoint.map(GrpcCheckpoint)?.try_into()?;
 
         let size = bcs::serialized_size(&checkpoint)?;
         self.send_remote_checkpoint_with_capacity_check(Arc::new(checkpoint), size)
@@ -370,10 +386,7 @@ impl CheckpointReaderActor {
             })?;
 
         while let Some(checkpoint) = checkpoints_stream.next().await {
-            let checkpoint: CheckpointData = checkpoint
-                .map(GrpcCheckpoint)
-                .map_err(|e| IngestionError::Grpc(e.to_string()))?
-                .try_into()?;
+            let checkpoint: CheckpointData = checkpoint.map(GrpcCheckpoint)?.try_into()?;
 
             let size = bcs::serialized_size(&checkpoint)?;
             self.send_remote_checkpoint_with_capacity_check(Arc::new(checkpoint), size)
@@ -400,21 +413,11 @@ impl CheckpointReaderActor {
             }
             RemoteStore::HybridHistoricalStore { historical, live } => {
                 if let Err(err) = self.fetch_historical(historical).await {
+                    // attempt to fetch from live object store if historical store does not contain
+                    // the requested checkpoint.
                     if matches!(err, IngestionError::CheckpointNotAvailableYet) {
                         let live = live.as_ref().ok_or(err)?;
-                        let mut checkpoint_stream = (self.current_checkpoint_number..u64::MAX)
-                            .map(|checkpoint_number| {
-                                fetch_from_object_store(live, checkpoint_number)
-                            })
-                            .pipe(futures::stream::iter)
-                            .buffered(batch_size);
-
-                        while let Some(checkpoint_result) = checkpoint_stream.next().await {
-                            let (checkpoint, size) = checkpoint_result?;
-                            self.send_remote_checkpoint_with_capacity_check(checkpoint, size)
-                                .await?;
-                        }
-                        return Ok(());
+                        return self.fetch_live(batch_size, live).await;
                     }
                     return Err(err);
                 }
--- a/crates/iota-data-ingestion-core/src/reader/v2.rs
+++ b/crates/iota-data-ingestion-core/src/reader/v2.rs
@@ -14,6 +14,7 @@ use iota_config::{
     node::ArchiveReaderConfig,
     object_storage_config::{ObjectStoreConfig, ObjectStoreType},
 };
+use iota_grpc_api::{CheckpointClient, CheckpointContent, NodeClient};
 use iota_metrics::spawn_monitored_task;
 use iota_rest_api::CheckpointData;
 use iota_types::messages_checkpoint::CheckpointSequenceNumber;
@@ -36,7 +37,7 @@ use crate::{
     IngestionError, IngestionResult, MAX_CHECKPOINTS_IN_PROGRESS, create_remote_store_client,
     history::reader::HistoricalReader,
     reader::{
-        fetch::{LocalRead, ReadSource, fetch_from_full_node, fetch_from_object_store},
+        fetch::{LocalRead, ReadSource, fetch_from_object_store},
         v1::{DataLimiter, ReaderOptions},
     },
 };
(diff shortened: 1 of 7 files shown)


crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -284,6 +289,98 @@ impl CheckpointReaderActor {
         Ok(())
     }
 
+    /// Fetches checkpoints form the fullnode trough a gRPC streaming connection
```
tomxey: typo: form -> from
(it also occurs in the PR title, and potentially other places)

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -284,6 +289,98 @@ impl CheckpointReaderActor {
...
+        // checkpoints (e.g. with migration data) may exceed gRPC's ~2GB message
+        // limit, and it will not work, so this logic must be revised to handle chunk
+        // assembly.
+        let Some(genesis_checkpoint) = genesis_checkpoint_stream.next().await else {
+            return Err(IngestionError::Grpc("checkpoint stream was closed".into()));
+        };
+
+        let checkpoint: CheckpointData = genesis_checkpoint
+            .map(GrpcCheckpoint)
+            .map_err(|e| IngestionError::Grpc(e.to_string()))?
```
tomxey: maybe some automated conversion would be possible from Grpc errors to the IngestionError::Grpc variant?

so that this map is not needed and you can just do `?`

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -284,6 +289,98 @@ impl CheckpointReaderActor {
...
+                Some(self.current_checkpoint_number),
+                None, // internally it resolves to u64::MAX
+                show_full_checkpoint,
+            )
+            .await
+            .map_err(|e| {
+                IngestionError::Grpc(format!("failed to initialize the checkpoint stream: {e}"))
+            })?;
+
+        while let Some(genesis_checkpoint) = checkpoints_stream.next().await {
```
tomxey: this is not a genesis checkpoint here, right?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -284,6 +289,98 @@ impl CheckpointReaderActor {
...
+    /// Fetches checkpoints from the fullnode through a gRPC streaming
+    /// connection.
+    ///
+    /// # Note
+    /// It should be used to download checkpoints except genesis one.
+    ///
+    /// For downloading the genesis checkpoint the
+    /// [`fetch_genesis_checkpoint_from_fullnode`](Self::fetch_genesis_checkpoint_from_fullnode)
+    /// method should be used.
+    async fn fetch_checkpoints_from_fullnode(
```
tomxey: would `fetch_nongenesis_checkpoints_from_fullnode` or `fetch_checkpoints_from_fullnode_post_genesis` make sense? or it's too verbose?
I'm thinking how to make this name more different from `fetch_from_fullnode` as they both fetch checkpoints

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -297,16 +394,7 @@ impl CheckpointReaderActor {
...
-                    .map(|checkpoint_number| fetch_from_full_node(client, checkpoint_number))
-                    .pipe(futures::stream::iter)
-                    .buffered(batch_size);
-
-                while let Some(checkpoint_result) = checkpoint_stream.next().await {
-                    let (checkpoint, size) = checkpoint_result?;
-                    self.send_remote_checkpoint_with_capacity_check(checkpoint, size)
-                        .await?;
-                }
+                self.fetch_from_fullnode(&mut client.clone()).await?;
```
tomxey: why the logic of sending to the channel was moved from here?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -575,3 +663,31 @@ impl CheckpointReader {
...
+    }
+}
+
+impl TryFrom<GrpcCheckpoint> for iota_types::full_checkpoint_content::CheckpointData {
+    type Error = IngestionError;
+
+    fn try_from(grpc_data: GrpcCheckpoint) -> Result<Self, Self::Error> {
+        match grpc_data.into_inner() {
+            CheckpointContent::Data(grpc_data) => grpc_data.into_v1().ok_or(IngestionError::Grpc(
+                "expected checkpoint data but received summary".into(),
```
tomxey: in this case it would mean we got sth different than v1, but not a summary. Maybe a match would be more futureproof here?

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: because is used internally in the `fetch_from_fullnode`, since internally we process data from the stream.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: The `iota-grpc-api` does not expose any `Error` type, it relies on `tonic::Status` which is not re-exported. So I opted to not introduce another dependency in the framework to handle the error at the expense of writing an additional line of code. Let me know what you think.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8836/commits/3bd6d31a252a79c20c92eee9308c05311a6b5b9c

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8836/commits/3bd6d31a252a79c20c92eee9308c05311a6b5b9c

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8836/commits/3bd6d31a252a79c20c92eee9308c05311a6b5b9c

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8836/commits/3bd6d31a252a79c20c92eee9308c05311a6b5b9c

crates/iota-data-ingestion-core/src/reader/v2.rs
tomxey: I can see that it was moved to "fetch_from_fullnode", but what was the reason?
It seemed consistent to have "send_remote_checkpoint_with_capacity_check" calls in this method, as that's also how "HybridHistoricalStore" is handled here.
The naming was also reflecting that, so that we have `send_to_channel` in the name if function is not returning the result directly but pushing it to the stream.

crates/iota-data-ingestion-core/src/reader/v2.rs
tomxey: The way I see it is that if we use `iota-grpc-api` and `iota-grpc-api` uses `tonic::Status` then the dependency is there anyway, it's just not directly specified.
I believe specifying it directly shouldn't be an additional cost, or am I wrong?
I'm not going to insist on that though, it was rather a nitpick.
Feel free to resolve the comment.

sergiupopescu199: You're right, I did not consider this case, will add the changes shortly, thank you!

fixed: https://github.com/iotaledger/iota/pull/8836/commits/10a4b06e4507db519b478f19418e64bfa1c55196

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: The `fetch_historical` does internally include the logic to send the items directly to the channel same as for the `fetch_from_fullnode`

The other logic is regarding the live version
```rust
 if matches!(err, IngestionError::CheckpointNotAvailableYet) {
                        let live = live.as_ref().ok_or(err)?;
                        let mut checkpoint_stream = (self.current_checkpoint_number..u64::MAX)
                            .map(|checkpoint_number| {
                                fetch_from_object_store(live, checkpoint_number)
                            })
                            .pipe(futures::stream::iter)
                            .buffered(batch_size);

                        while let Some(checkpoint_result) = checkpoint_stream.next().await {
                            let (checkpoint, size) = checkpoint_result?;
                            self.send_remote_checkpoint_with_capacity_check(checkpoint, size)
                                .await?;
                        }
                        return Ok(());
                    }
```
I've refactored a bit the code to make it more clear.

fixed: https://github.com/iotaledger/iota/pull/8836/commits/10a4b06e4507db519b478f19418e64bfa1c55196

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -284,6 +290,112 @@ impl CheckpointReaderActor {
         Ok(())
     }
 
+    /// Fetch checkpoints from the live object store and stream them to a
```
tomxey: nit: Fetches ... and streams

tomxey (approved):

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8836/commits/dc72d0c31545062f7b11890ba3534b0fed74f522

kodemartin (approved):
lgtm

PR #8837, opened 2025-10-12
chore(rust): 1.89's `mismatched_lifetime_syntaxes`

thibault-martinez:
# Description of change

Addresses rust 1.89's [mismatched_lifetime_syntaxes](https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint/lifetime_syntax/static.MISMATCHED_LIFETIME_SYNTAXES.html) lint.
This is a **compiler** lint, not a **clippy** lint.


external-crates/move/crates/move-compiler/src/shared/matching.rs
```diff
@@ -64,11 +66,11 @@ pub struct ArmResult {
 // Match Context
 //**************************************************************************************************
 
-/// A shared match context trait for use with counterexample generation in Typing and match
-/// compilation in HLIR lowering.
+/// A shared match context trait for use with counterexample generation in
+/// Typing and match compilation in HLIR lowering.
```
Thoralf-M: Format shouldn't be changed here?

thibault-martinez: Actually, it should. Our comment width is 80 and this was above (91).

thibault-martinez: Oh but are we not formatting the external crates? I forgot.

TheMrAI:
I don't know about managerial plans so I am inquiring here.
Could this be postponed until we `vm-lang` (probably others?) pull the relevant changes from sui?

While these changes aren't big, it will break patch application process, which is already time consuming enough.

thibault-martinez:
> Could this be postponed until we `vm-lang` (probably others?) pull the relevant changes from sui?

Correct me if I'm wrong but pulling upstream is a never-ending process, what would be a good time for us to do this then?

external-crates/move/crates/move-compiler/src/shared/matching.rs
thibault-martinez: Will revert format changes to external crates.

TheMrAI:
> > Could this be postponed until we `vm-lang` (probably others?) pull the relevant changes from sui?
> 
> Correct me if I'm wrong but pulling upstream is a never-ending process, what would be a good time for us to do this then?

| ... pull the relevant changes from sui?
When they have applied these changes as well.

thibault-martinez:
> | ... pull the relevant changes from sui? When they have applied these changes as well.

We have never really waited for them to update their toolchain, even if it meant touching the external crates. They are on 1.85, we are on 1.88. I don't think we should start waiting now.

muXxer (approved):

piotrm50 (approved):

miker83z (approved):

external-crates/move/crates/move-compiler/src/shared/matching.rs
thibault-martinez: Reverted the formatting of the external crates

Thoralf-M (approved):

PR #8859, opened 2025-10-13
fix(iota-config): remove `iota-grpc-api` dependency in `iota-config` and `iota-swarm-config`

muXxer:
# Description of change

We introduced the dependency to `iota-grpc-api` in `iota-config`, which causes that everything that depends on `iota-config` also need to install `protoc`, for example the `iota-sdk`. This PR removes this dependency.


semenov-vladyslav (approved):

bingyanglin (approved):

thibault-martinez (approved):
Approving `crates/iota/src/iota_commands.rs`

sergiupopescu199 (approved):

PR #8869, opened 2025-10-14
chore(move-tooling): use mysten dependency and update locks

miker83z:
# Description of change

Use `@mysten/prettier-plugin-move` dependency in prettier-extension and update package-lock


TheMrAI (commented):
Approved! It just doesn't have the approve tick, because as a test I tried to merge it in without it.
Approving it later, won't show it.

Issue #8688, opened 2025-09-30
Implement more endpoints in the internal gRPC interface


muXxer:
These are the endpoints that are required, based on the feedback from the other teams:

```proto
  rpc GetServiceInfo(GetServiceInfoRequest) returns (GetServiceInfoResponse);
  rpc GetObjects(GetObjectsRequest) returns (stream GetObjectsResponse);
  rpc GetTransactions(GetTransactionsRequest) returns (stream GetTransactionsResponse);
  rpc GetCheckpointData (GetCheckpointDataRequest) returns (stream CheckpointData);
  rpc StreamCheckpointData (CheckpointDataStreamRequest) returns (stream CheckpointData);
  rpc GetEpoch(GetEpochRequest) returns (GetEpochResponse);
  rpc ExecuteTransaction(ExecuteTransactionRequest) returns (ExecuteTransactionResponse);
  rpc SimulateTransaction(SimulateTransactionRequest) returns (SimulateTransactionResponse);
```

muXxer:
These endpoints were required after the first feedback from other teams:

- DevInspectTransactionBlock (infra+wasp) #8870 
- DryRunTransaction (infra+wasp) #8870 
- ExecuteTransactionBlock (infra+wasp) #8552
- GetAllBalances (wasp)
- GetAllCoins (wasp)
- GetBalance (wasp)
- GetCoinMetadata (wasp)
- GetCoins (wasp)
- GetDynamicFields (wasp)
- GetLatestIotaSystemState (wasp)
- GetObject (wasp) #8552
- GetOwnedObjects (wasp)
- GetReferenceGasPrice (wasp)
- ContainsTransaction (infra) #8870
- GetTransactionBlock (infra+wasp) #8870
- MergeCoins (wasp)
- Pay (wasp)
- Publish (wasp)
- SignAndExecuteTransaction (wasp)
- TransferIota (wasp)
- TransferObject (wasp)
- TryGetPastObject (wasp)
- GetLatestCheckpointSequenceNumber (infra) #8552

After we started implementing those, we realized that several endpoints should not be implemented in the node, because they require exhaustive indexing, which we don't want in the node.

After a second round of endpoint definitions, we came up with these endpoints:

```proto
  rpc GetServiceInfo(GetServiceInfoRequest) returns (GetServiceInfoResponse);
  rpc GetObjects(GetObjectsRequest) returns (stream GetObjectsResponse);
[...]

PR #8870, opened 2025-10-14
feat(gRPC): add GetTransaction, ContainsTransaction, DevInspectTransaction, and DryRunTransaction

bingyanglin:
# Description of change

- Please check the protobuf change for discussion in https://github.com/iotaledger/iota/pull/8870/commits/efda5cdaec2387a2b3194a17f75975afc134f518.
- Implement the gRPC APIs including `GetTransaction`, `ContainsTransaction`, `DevInspectTransaction`, and `DryRunTransaction` for internal usage.
- Some notes when we compare them to the `iota-json-rpc`
  - `DevInspectTransaction`
    - `_epoch` is not used in the request so we don't include it
    - `effects` (`iota-types` `TransactionEffects`) will always present, which can construct IotaTransactionBlockEffects , so it should always be shown, and both the original `IotaTransactionBlockEffects` and `raw_effects` are replaced by this one. In this way, we only need  `show_txn_data` to decide whether the `txn_data` is shown. We rename it to be `txn_data` without `raw` because the default response types of the gRPC are considered to be from `iota-types`, so we don't need to add `raw` to specify it's from `iota-types`.
  - `DryRunTransactionRequest`
    - Added the `execution_error_source` in the response, which is added in https://github.com/MystenLabs/sui/pull/21375.
- The `message Bcs` is now only placeholder, and will be used in the future to enhance the response type.

Part of #8688

### Release Notes

- [x] Nodes (Validators and Full nodes): For gPRC, added APIs `GetTransaction`, `ContainsTransaction`, `DevInspectTransaction`, and `DryRunTransaction` for internal usage.


muXxer:
This can be closed because it was done in #9437 and #9335

PR #8881, opened 2025-10-14
refactor: add a new `Validation` execution mode and refactor `AuthContext` creation

### LINKED ISSUE #8845: [AA][execution] Add a new Validation execution mode and refactor AuthContext creation
- Add a new Validation execution mode 
- AuthContext is not treated as PTB input value anymore
- Instead of changing the ExecutionContext APIs, this proposal tries to limit it. This because from upstream there will be changes to such interface coming from the Sui PTBv2. So, limiting changes to this would allow us to pull commits more easily.
- We carry around the AuthContext information inside the TemporaryStore and then we extend the iota_types::storage::Storage trait with a method that reads this. This trait is supported by ExecutionContext and it is convenient to change because TemporaryStore is the only implementor. 
- Moreover, TemporaryStore only needs a new optional fields that is set to None during the constructor: Option<AuthContext>. 
- The AuthContext checking and reading from Storage will only be done if Mode matches Validation.

valeriyr:
# Description of change

Added a new Move execution mode named `Validation`. It is allowed to use the `iota::auth_context::AuthContext` type only in this execution mode.
`iota::auth_context::AuthContext` is added to a transaction parameters list in the same way as it is implemented for `iota::tx_context::TxContext`

fixes #8845

### DIFF
--- a/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
+++ b/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
@@ -417,7 +417,7 @@ mod checked {
             is_init,
         )?;
         // build the arguments, storing meta data about by-mut-ref args
-        let (tx_context_kind, by_mut_ref, serialized_arguments) =
+        let (tx_context_kind, has_auth_context, by_mut_ref, serialized_arguments) =
             build_move_args::<Mode>(context, runtime_id, function, kind, &signature, &arguments)?;
         // invoke the VM
         let SerializedReturnValues {
@@ -429,6 +429,7 @@ mod checked {
             function,
             type_arguments,
             tx_context_kind,
+            has_auth_context,
             serialized_arguments,
             trace_builder_opt,
         )?;
@@ -871,9 +872,18 @@ mod checked {
         function: &IdentStr,
         type_arguments: Vec<Type>,
         tx_context_kind: TxContextKind,
+        has_auth_context: bool,
         mut serialized_arguments: Vec<Vec<u8>>,
         trace_builder_opt: &mut Option<MoveTraceBuilder>,
     ) -> Result<SerializedReturnValues, ExecutionError> {
+        if has_auth_context {
+            let auth_context = context.state_view.read_auth_context();
+            assert_invariant!(
+                auth_context.is_some(),
+                "The `iota::auth_context::AuthContext` value is expected to be read from the storage"
+            );
+            serialized_arguments.push(auth_context.unwrap().to_bcs_bytes());
+        }
         match tx_context_kind {
             TxContextKind::None => (),
             TxContextKind::Mutable | TxContextKind::Immutable => {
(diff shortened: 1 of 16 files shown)


iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -1628,7 +1614,12 @@ mod checked {
         context: &ExecutionContext<'_, '_, '_>,
         t: &Type,
     ) -> Result<bool, ExecutionError> {
-        let Type::Datatype(idx) = t else {
+        let inner = match t {
+            Type::Reference(inner) => inner,
+            _ => return Ok(false),
+        };
```
valeriyr, line 1626: Just want to emphasize that we don't double-check here that an `AuthContext` instance is passed by a mutable reference, for example. `AuthContext` will be added to the transaction parameters list only if it is passed by an immutable reference.

TheMrAI: Add a debug assert perhaps or at least a comment?
When this invariant breaks, we have some guiding hand left.

TheMrAI (approved):
I don't follow what is going on in terms of overall design, but I saw no big issues.
Don't know if it breaks anything as I don't have familiarity with this part of the system.

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
valeriyr: In this function, we don't really check if it is possible to pass `AuthContext` by a mutable reference. We check if we need to add it to the arguments list. If it is not added, it will fail with the `ArityMismatch` error, as it is done for `TxContext`.

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -1315,13 +1326,37 @@ mod checked {
             Some(t) => is_tx_context(context, t)?,
             None => TxContextKind::None,
         };
+        // Check if the second last parameter is an auth context.
+        // According to the current design, authenticators must contain `TxContext` as
+        // the last parameter and `AuthContext` as the second last.
+        let has_auth_context = match parameters.iter().rev().nth(1) {
```
Dkwcs, line 1332: Just curious, what if we have auth_context as third last parameter or on any other place?

Dkwcs (approved):
LGTM, anyway Mirko has the final say

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
valeriyr: We will not detect it here and will not add to the arguments list, so that the result will be either `ArityMismatch` in case a user passed fewer parameters than expected, or could be `TypeMismatch` if they passed something as a parameter.

miker83z (approved):
LGTM!

PR #8883, opened 2025-10-14
feat(protocol-research): allow overshoot of congestion limit in the sequencer

### LINKED ISSUE #9085: [Task (core-protocol-research)]: Enable overshoot in the sequencer on Devnet in Protocol Version 15

cyberphysic4l:
# Description of change

Implements "burst mode" for the sequencer by allowing the accumulated execution duration scheduled in a commit to overshoot the set limit. Any overshoot above the limit is recorded as a "debt" and carried over to the next shared object congestion tracker using up the capacity reserved for the next commit.

The amount that we can overshoot the limit is determined by a new protocol parameter - `max_congestion_limit_overshoot_per_commit`.

The feature is an adapted version of Sui's "overage" feature as seen in [this PR](https://github.com/MystenLabs/sui/commit/f1dd33b7f25a3977ab6a73d5e75f17007a09c07c) and subsequent modifications and bug fixes 

closes #9085 

### Release Notes

- [x] Protocol: when a value is set for `max_congestion_limit_overshoot_per_commit`, the number of transactions that can be scheduled for execution for a single commit may increase by this much, but on average this number will be determined by the existing parameter `max_accumulated_txn_cost_per_object_in_mysticeti_commit`.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


thibault-martinez (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -675,6 +677,10 @@ pub struct AuthorityEpochTables {
 
     /// Holds the timestamp of the most recently generated round of randomness.
     pub(crate) randomness_last_round_timestamp: DBMap<u64, CommitTimestampMs>,
+
```
roman1e2f5p8s, line 679:
```suggestion

    // Tables for recording per-object debts for congestion control.

    //
```
There are quite a few tables just above that, labeled as `Tables for recording state for RandomnessManager`. Add this separator to make it clear these two new tables are not related to randomness manager.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1914,14 +1920,20 @@ impl AuthorityPerEpochStore {
...
         // implementation of the shared object congestion tracker uses the term
         // "execution duration" to describe the same concept.
         self.protocol_config()
             .max_accumulated_txn_cost_per_object_in_mysticeti_commit_as_option()
     }
 
+    fn get_max_congestion_limit_overshoot_per_commit(&self) -> u64 {
+        self.protocol_config()
+            .max_congestion_limit_overshoot_per_commit_as_option()
+            .unwrap_or(0)
```
```suggestion
            .unwrap_or_default()
```
This is considered more idiomatic.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -10,6 +10,7 @@ use std::{
 };
 
 use arc_swap::ArcSwapOption;
+use consensus_core::Round;
```
roman1e2f5p8s: I think we should use `CommitRound` instead, as we do for deferrals, scheduling, etc.
And it is already imported in this file.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2786,7 +2799,7 @@ impl AuthorityPerEpochStore {
             }
         }
 
-        let mut output = ConsensusCommitOutput::new();
+        let mut output = ConsensusCommitOutput::new(consensus_commit_info.round as Round);
```
roman1e2f5p8s: Use `CommitRound` to avoid down-casting.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2921,6 +2934,29 @@ impl AuthorityPerEpochStore {
             &mut sequenced_randomness_transactions,
             self.protocol_config.consensus_transaction_ordering(),
         );
+
+        // We track transaction execution duration separately for regular transactions
+        // and transactions using randomness, since they will be in different
+        // checkpoints.
```
roman1e2f5p8s: I think the reason behind having two congestion trackers is that we want a separate scheduling line for txs using randomness.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2921,6 +2934,29 @@ impl AuthorityPerEpochStore {
...
+        let shared_object_congestion_tracker = SharedObjectCongestionTracker::new(
+            self.consensus_quarantine.read().load_initial_object_debts(
+                self,
+                consensus_commit_info.round as Round,
```
roman1e2f5p8s: Use `CommitRound` to avoid down-casting.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2921,6 +2934,29 @@ impl AuthorityPerEpochStore {
...
+                false,
+                &sequenced_transactions,
+            )?,
+            self.protocol_config(),
+        );
+        let shared_object_using_randomness_congestion_tracker = SharedObjectCongestionTracker::new(
+            self.consensus_quarantine.read().load_initial_object_debts(
+                self,
+                consensus_commit_info.round as Round,
```
roman1e2f5p8s: Use `CommitRound` to avoid down-casting.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -38,6 +42,7 @@ use crate::{
 #[derive(Default)]
 pub(crate) struct ConsensusCommitOutput {
     // Consensus and reconfig state
+    consensus_round: Round,
```
roman1e2f5p8s: I think we should use `CommitRound` instead.
`Round` is for blocks, not commits.

Also, `commit_round` would be a better name.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -66,11 +71,18 @@ pub(crate) struct ConsensusCommitOutput {
     // jwk state
     pending_jwks: BTreeSet<(AuthorityName, JwkId, JWK)>,
     active_jwks: BTreeSet<(u64, (JwkId, JWK))>,
+
+    // congestion control state
```
roman1e2f5p8s: There are two fields for deferred txs above, which are also a part of congestion control state.
Makes sense to keep them close under the same comment.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -675,6 +677,10 @@ pub struct AuthorityEpochTables {
 
     /// Holds the timestamp of the most recently generated round of randomness.
     pub(crate) randomness_last_round_timestamp: DBMap<u64, CommitTimestampMs>,
+
+    /// Accumulated per-object debts for congestion control.
+    pub(crate) congestion_control_object_debts: DBMap<ObjectID, CongestionPerObjectDebt>,
+    pub(crate) congestion_control_randomness_object_debts: DBMap<ObjectID, CongestionPerObjectDebt>,
```
```suggestion
    congestion_control_object_debts: DBMap<ObjectID, CongestionPerObjectDebt>,
    
    /// Accumulated per-object debts for randomness congestion control.
    congestion_control_randomness_object_debts: DBMap<ObjectID, CongestionPerObjectDebt>,
```

Few points:

- both fields can be kept private;
- the second field should have doc string as well.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -817,6 +900,85 @@ impl ConsensusOutputQuarantine {
...
+        epoch_store: &AuthorityPerEpochStore,
+        current_round: Round,
+        for_randomness: bool,
+        transactions: &[VerifiedSequencedConsensusTransaction],
+    ) -> IotaResult<impl IntoIterator<Item = (ObjectID, u64)>> {
+        let protocol_config = epoch_store.protocol_config();
+        let tables = epoch_store.tables()?;
+        let default_per_commit_budget = protocol_config
+            .max_accumulated_txn_cost_per_object_in_mysticeti_commit_as_option()
+            .unwrap_or(0);
```
```suggestion
            .unwrap_or_default();
```

This is more idiomatic.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -817,6 +900,85 @@ impl ConsensusOutputQuarantine {
...
+        let (hash_table, db_table, per_commit_budget) = if for_randomness {
+            (
+                &self.congestion_control_randomness_object_debts,
+                &tables.congestion_control_randomness_object_debts,
+                protocol_config
+                    .max_accumulated_txn_cost_per_object_in_mysticeti_commit_as_option()
+                    .unwrap_or(default_per_commit_budget),
```
```suggestion
                default_per_commit_budget,
```

As it is the identical to already computed value above.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -817,6 +900,85 @@ impl ConsensusOutputQuarantine {
...
+        Ok(results
+            .into_iter()
+            .zip(shared_input_object_ids)
+            .filter_map(|(debt, object_id)| debt.map(|debt| (debt, object_id)))
+            .map(move |((round, debt), object_id)| {
+                // Stored debts already account for the budget of the round in which
+                // they were accumulated. Application of budget from future rounds to
+                // the debt is handled here.
+                assert!(current_round > round);
+                let num_rounds = (current_round - round - 1) as u64;
```
roman1e2f5p8s: This casting could also be avoided if `CommitRound` was used instead of `Round`.

roman1e2f5p8s (changes requested):
I reviewed all files except `crates/iota-core/src/authority/shared_object_congestion_tracker.rs`, which I will take a look at tomorrow. Requested some changes.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -4,12 +4,14 @@
 
 use std::{cmp::Ordering, collections::HashMap};
 
-use iota_protocol_config::PerObjectCongestionControlMode;
+use consensus_core::Round;
```
roman1e2f5p8s: As already mentioned, `CommitRound` seems more appropriate, and it is already imported in this file.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -220,13 +222,53 @@ pub(crate) struct SharedObjectCongestionTracker {
...
+        let mut object_execution_slots = HashMap::new();
+        for (object_id, debt) in initial_object_debts {
+            let mut slots = ObjectExecutionSlots::new();
+            if debt > 0 {
+                // If there is an initial debt, remove the occupied slot from time 0 to
+                // debt.
+                slots.remove(ExecutionSlot::new(0, debt));
+            }
+            object_execution_slots.insert(object_id, slots);
+        }
```
roman1e2f5p8s:
```suggestion
        let object_execution_slots = initial_object_debts
            .into_iter()
            .map(|(object_id, debt)| {
                let mut slots = ObjectExecutionSlots::new();
                if debt > 0 {
                    // If there is an initial debt, remove the occupied slot from time 0 to
                    // debt.
                    slots.remove(ExecutionSlot::new(0, debt));
                }

                (object_id, slots)
            })
            .collect::<HashMap<_, _>>();
```

More importantly, this hash map is computed but used nowhere. Did you mean to use it in line 241 instead of `HashMap::new()`? I guess currently we don't have any debts because `max_congestion_limit_overshoot_per_commit` is not "enabled"? But once `max_congestion_limit_overshoot_per_commit` is set, there is a danger of forgetting to use the computed `object_execution_slots` instead of `HashMap::new()` for initializing object execution slots in the tracker.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -220,13 +222,53 @@ pub(crate) struct SharedObjectCongestionTracker {
...
```
roman1e2f5p8s:
```suggestion
        let object_execution_slots = initial_object_debts
            .into_iter()
            .map(|(object_id, debt)| {
                let mut slots = ObjectExecutionSlots::new();
                if debt > 0 {
                    // If there is an initial debt, remove the occupied slot from time 0 to
                    // debt.
                    slots.remove(ExecutionSlot::new(0, debt));
                }

                (object_id, slots)
            })
            .collect();
```

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -489,6 +534,38 @@ impl SharedObjectCongestionTracker {
             .max()
             .unwrap_or(0)
     }
+
+    // Returns accumulated debts for objects whose budgets have been exceeded over
+    // the course of the commit. Consumes the tracker object, since this should
+    // only be called once after all tx have been processed.
```
roman1e2f5p8s:
```suggestion
    /// Returns accumulated debts for objects whose budgets have been exceeded over
    /// the course of the commit. Consumes the tracker object, since this should
    /// only be called once after all txs have been processed.
```

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -489,6 +534,38 @@ impl SharedObjectCongestionTracker {
...
+            .filter_map(|(obj_id, slots)| {
+                let debt = slots
+                    .max_object_occupied_slot_end_time()
+                    .saturating_sub(max_execution_duration_per_commit);
+                if debt > 0 { Some((obj_id, debt)) } else { None }
+            })
+            .collect()
+    }
+}
+
```
roman1e2f5p8s, line 557:
```suggestion

/// Stores per-object debts from a given consensus commit.
```

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -489,6 +534,38 @@ impl SharedObjectCongestionTracker {
...
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+pub enum CongestionPerObjectDebt {
+    V1(Round, u64),
```
roman1e2f5p8s: Use `CommitRound`.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
             .is_none()
         );
     }
+
+    #[rstest]
+
```
roman1e2f5p8s:
```suggestion
```

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -971,10 +1004,10 @@ mod object_cost_tests {
 
         // Set max_execution_duration_per_commit to only allow 1 transaction
         // to go through.
-        let max_execution_duration_per_commit = match mode {
+        let (max_execution_duration_per_commit, max_overshoot_per_commit) = match mode {
             PerObjectCongestionControlMode::None => unreachable!(),
-            PerObjectCongestionControlMode::TotalGasBudget => 12,
-            PerObjectCongestionControlMode::TotalTxCount => 3,
+            PerObjectCongestionControlMode::TotalGasBudget => (12, 0),
+            PerObjectCongestionControlMode::TotalTxCount => (3, 0),
```
roman1e2f5p8s, line 1011:
```suggestion
            PerObjectCongestionControlMode::TotalTxCount => (1, 0),
```
Why do we have this set to 3 instead of 1? The comment says we want to only allow 1 tx to go through. Is the comment misleading? The test passes with 1, btw.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+
+        // Set max_accumulated_txn_cost_per_object_in_commit to only allow 1 transaction
+        // to go through before overage occurs.
+        let max_execution_duration_per_commit = match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => 90,
+            PerObjectCongestionControlMode::TotalTxCount => 2,
+        };
+
+        let current_max_duration_per_object = match mode {
```
roman1e2f5p8s: This name `current_max_duration_per_object` is confusing.
It should be renamed to `initial_object_debt`.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+                )
+            }
+        };
+
+        // Set max_execution_duration_per_commit to allow 1 transaction to
+        // go through before an overshoot occurs.
+        let max_execution_duration_per_commit = match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => tx_gas_budget,
+            PerObjectCongestionControlMode::TotalTxCount => 2,
```
roman1e2f5p8s:
```suggestion
            PerObjectCongestionControlMode::TotalTxCount => 1,
```

Why is it 2 if the comment says "to allow 1 transaction to go through"?

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+        // instantiate the tracker with some initial debts.
+        let shared_object_congestion_tracker = match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => {
+                // Construct object execution cost as following
+                //                199   301
+                // object 0:            |
+                // object 1:      |
+                //
+                // burst limit is 100 + 200 = 300
```
roman1e2f5p8s: I don't understand where 200 comes from?

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+                    assign_min_free_execution_slot,
+                )
+            }
+            PerObjectCongestionControlMode::TotalTxCount => {
+                // Construct object execution cost as following
+                //                3     4
+                // object 0:            |
+                // object 1:      |
+                //
+                // overall limit is 2 + 2 = 4
```
roman1e2f5p8s: Same here, could you please explain these "2 + 2 = 4"?

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+        let shared_obj_1 = ObjectID::random();
+
+        let tx_gas_budget = 100;
+
+        // Set max_accumulated_txn_cost_per_object_in_commit to only allow 1 transaction
+        // to go through before overage occurs.
+        let max_execution_duration_per_commit = match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => 90,
+            PerObjectCongestionControlMode::TotalTxCount => 2,
```
roman1e2f5p8s, line 1813:
```suggestion
            PerObjectCongestionControlMode::TotalTxCount => 1,
```

Why is it 2 if the comment says "to allow 1 transaction to go through"?

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+            assign_min_free_execution_slot,
+        );
+
+        // Verify that accumulated_debts is empty initially.
+        let accumulated_debts = shared_object_congestion_tracker
+            .clone()
+            .accumulated_debts(max_execution_duration_per_commit);
+        assert!(accumulated_debts.is_empty());
+
+        // Simulate a tx on object 0 that exceeds the budget.
```
roman1e2f5p8s: Which "budget"? `max_execution_duration_per_commit`? Use "per-commit limit" instead of "budget". Otherwise, it can be easily confused with `tx_gas_budget`.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1632,213 @@ mod object_cost_tests {
...
+        let accumulated_debts =
+            shared_object_congestion_tracker.accumulated_debts(max_execution_duration_per_commit);
+        assert_eq!(accumulated_debts.len(), 1);
+        match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => {
+                assert_eq!(accumulated_debts[0], (shared_obj_0, 90)); // init 80 + cost 100 - budget 90 = 90
+            }
+            PerObjectCongestionControlMode::TotalTxCount => {
+                assert_eq!(accumulated_debts[0], (shared_obj_0, 1)); // init 2 + 1 tx - budget 2 = 1
```
roman1e2f5p8s: Same here RE "budget", use (per-commit) limit.

roman1e2f5p8s (changes requested):
The flow of the comments and statements in the new tests is not clear. The tests do not look too complicated, but it is difficult to read them. For example, in the `test_try_schedule_allow_overshoot` test, if `max_execution_duration_per_commit` and `max_overshoot_per_commit` were initialized before `shared_object_congestion_tracker`, it would be must easier to understand the comments in where the tracker is instantiated.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: You're right, this should have been used instead of a new HashMap, thanks for catching that. It was set correctly in the `new_for_test` function, so the problem didn't present itself in the tests.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: The reason the limit is 3 is because we also set some initial objects in the tracker when we initialise it below. In the `TotalTxCount` mode, we have a transaction with end time 2 when we initialise it. Then when we bump execution slots below that, we add the next transaction with `start_time` 2, so it's end time is 3. This is the one transaction that can fit. 

The reason it works when you set this value to 1 is that the initialiser and bump function don't care about the congestion limit, so we can still fill the tracker beyond the limit. THe only thing checked in the test is that something gets deferred and returns shared objects.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: Because the initial debts are set to 2, so a single transaction will result in an overshoot.
I rephrased it to highlight the fact that the initial debts play a role as well and made the phrasing more clear.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: Hopefully comments make that clearer now

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: Hopefully comments make that clearer now

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: Made the comments clearer now. You have to take initial object debts into account as well as the limit.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2921,6 +2938,28 @@ impl AuthorityPerEpochStore {
             &mut sequenced_randomness_transactions,
             self.protocol_config.consensus_transaction_ordering(),
         );
+
+        // We track transaction shared object congestion separately for regular
```
cuileri, line 2943: Why did we need to move the creation of congestion trackers here?

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1637,217 @@ mod object_cost_tests {
...
+            assign_min_free_execution_slot,
+        );
+
+        // Verify that accumulated_debts is empty initially.
+        let accumulated_debts = shared_object_congestion_tracker
+            .clone()
+            .accumulated_debts(max_execution_duration_per_commit);
+        assert!(accumulated_debts.is_empty());
+
+        // Simulate a tx on object 0 that exceeds the per-commit limit.
```
cuileri: We actually simulate two transactions here, one read and one write.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1637,217 @@ mod object_cost_tests {
...
+        }
+
+        // Verify that accumulated_debts reports the debt for object 0.
+        let accumulated_debts =
+            shared_object_congestion_tracker.accumulated_debts(max_execution_duration_per_commit);
+        assert_eq!(accumulated_debts.len(), 1);
+        match mode {
+            PerObjectCongestionControlMode::None => unreachable!(),
+            PerObjectCongestionControlMode::TotalGasBudget => {
+                assert_eq!(accumulated_debts[0], (shared_obj_0, 90)); // overshoot = initial_debt (80) + tx_duration (100) - max_execution_duration_per_commit (90) = 90
```
cuileri: Selected numbers make the case difficult to understand. Two different things have the value 90. We could pick different numbers, for example, make the initial debt 70.

cuileri (commented):
A simplified version of the overage idea, tailored for our protocol. No issues observed with the implementation. I only have two considerations: 
- I didn't understand why you needed to change the place of the initialization of the congestion tracker instances. 
- I think we need tests on the behavior of object debts in successive commit rounds. We can add test(s) showing how the debts are decreased/removed. Example below.

#### An example test scenario:
- assume single object, and budget mode
- limit per commit: 80
- overshoot per commit: 100

##### In commit round 1:
  - Tx1 with cost 150 is **scheduled**.
  - Tx2 with cost 40 is **deferred**.

##### In commit round 2: 
- Verify the initial debt is 70
- Tx3 (with a higher gas price than Tx2) with cost 60 is **scheduled**
- Tx4 (with a higher gas price than Tx2 and lower gas price than Tx3) with cost 60 is **deferred**
- Tx2 with cost 40 is **scheduled**
- Verify the debt is 90

##### In commit round 3:
- No new tx on that object
- Tx4 is **scheduled** this time
- Verify the debt is 70

##### In commit round 4:
- No new tx on that object

##### In commit round 5:
- Verify the debt is removed

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
roman1e2f5p8s: Thanks, I took me a while to understand this test, even though it is not directly part of the PR.

Why I found the test flow confusing:

1. First, we have this

   ```rust
        let tx_gas_budget = 5;

        // Set max_execution_duration_per_commit to only allow 1 transaction
        // to go through.
        let (max_execution_duration_per_commit, max_overshoot_per_commit) = match mode {
            PerObjectCongestionControlMode::None => unreachable!(),
            PerObjectCongestionControlMode::TotalGasBudget => (12, 0),
            PerObjectCongestionControlMode::TotalTxCount => (3, 0),
        };
   ```

2. Then, we initialize the tracker with initial values for `object 0` being `9` in gas budget mode and `2` in tx count mode.

3. From this context, it is easy to see that one transaction is allowed to go through (as the comment says) in the tx count mode. But in the gas budget mode, how can a transaction with `tx_gas_budget = 5` go through if the limit is `12` and initial end time is set to `9`? It turns out that `tx_gas_budget = 5` is only used for a transaction which we don't want to go through (w.r.t. `object 0`), while the gas budget of `1` is used for a transaction we want to go through. But this part is not clear until:

   ```rust
        // add a transaction that writes to object 0 and 1.
        let tx = build_transaction(
            &[(shared_obj_0, true), (shared_obj_1, true)],
            1,
            TEST_ONLY_GAS_PRICE,
        );
   ```

Moving the `let tx_gas_budget = 5;` statement closer to where the value is used, adding a comment to it, adding names for "magic" numbers like `9`, `2`, `1` could improve the readability a lot.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
roman1e2f5p8s: Thanks for making it clearer.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
roman1e2f5p8s:
```suggestion
        // Simulate a tx on object 0 that exceeds the per-commit limit,
        // taking into account the initial object debt.
```
In tx count mode, a tx would not exceed the per-commit limit, which is set to `2`, if the initial object debt was lower. In gas budget mode, a tx always exceeds the limit (`90`) as tx gas budget is set to `100`, no matter if there are initial objects debts or not.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1581,4 +1637,217 @@ mod object_cost_tests {
...
+    fn test_accumulated_debts(
+        #[values(
+            PerObjectCongestionControlMode::TotalGasBudget,
+            PerObjectCongestionControlMode::TotalTxCount
+        )]
+        mode: PerObjectCongestionControlMode,
+        #[values(true, false)] assign_min_free_execution_slot: bool,
+    ) {
+        // Creates two shared objects and three transactions that operate on these
+        // objects.
```
roman1e2f5p8s:
```suggestion
        // Creates two shared objects to operate on them in transactions.
```

roman1e2f5p8s (approved):
Thanks for addressing my suggestions and making the tests clearer.

(note that clippy ci is failing)

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: We create it here because we now load the initial object debts which takes as input the list of user transactions that need to pass through the sequencer. This list was not available in the place where the tracker was initialised before because it is combined into a single vector with the system transactions and randomness transactions below. If we kept the constructor call in the same place we would have to pass all three vectors separately to the `process_consensus_transactions` function instead of a single vector combining them.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: That is indeed a bit confusing, I will do as you suggested, naming the numbers where possible and initalising in the place where they're used.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
cyberphysic4l: Good suggestion, changed to 70.

crates/iota-core/src/unit_tests/congestion_control_tests.rs
```diff
@@ -406,3 +438,474 @@ async fn test_congestion_control_execution_cancellation() {
     );
     assert_eq!(&effects, effects_2.data())
 }
+
+// Tests that congestion control and debt tracking work as expected when there
+// is a burst of traffic and overshoot is allowed.
+//
+// Specifically, we test
```
cuileri: Incomplete sentence

crates/iota-core/src/unit_tests/congestion_control_tests.rs
```diff
@@ -406,3 +438,474 @@ async fn test_congestion_control_execution_cancellation() {
...
+    } else {
+        panic!("Unexpected debt stored in consensus quarantine.");
+    }
+
+    // Commit 4: a transaction with gas budget 2.5*default_tx_gas_budget that
+    // touches shared_object_1 and an owned object.
+    // The debt of 1.5*default_gas_budget on shared object 1 from commit 2 should be
+    // reduced to 0.5*default_gas_budget for commit round 4 because round 3 was
+    // skipped, reducing it by the congestion limit of default_gas_budget.
+    // Therefore, this transaction should be executed successfully as the total
```
cuileri, line 671: Therefore, this transaction should be executed successfully as the total
    // overshoot will be 2*default_gas_budget (0.5*default_gas_budget from
    // existing debt and 2.5*default_gas_budget from this transaction).

The number within the parentheses do not add up to 2. I'd add "minus the commit limit of default_gas_budget" within the parentheses.

crates/iota-core/src/unit_tests/congestion_control_tests.rs
```diff
@@ -406,3 +438,474 @@ async fn test_congestion_control_execution_cancellation() {
...
+        assert_eq!(debt, default_tx_gas_budget / 2);
+        assert_eq!(commit_round, 3);
+    } else {
+        panic!("Unexpected debt stored in consensus quarantine.");
+    }
+
+    // Commit 5: a transaction with gas budget of 1.5*default_tx_gas_budget that
+    // touches both shared objects and an owned object. The transaction should be
+    // cancelled because there is a debt of 2*default_tx_gas_budget on shared object
+    // 1, resulting in a total overshoot of 1.5*default_tx_gas_budget.
```
cuileri: > resulting in a total overshoot of 1.5*default_tx_gas_budget
1.5 should be 2.5?

cuileri (approved):
With the addition of new tests for debt tracking, everything seems fine.

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -463,6 +514,8 @@ impl ConsensusOutputQuarantine {
             builder_checkpoint_summary: BTreeMap::new(),
             builder_digest_to_checkpoint: HashMap::new(),
             shared_object_next_versions: RefCountedHashMap::new(),
+            congestion_control_randomness_object_debts: RefCountedHashMap::new(),
+            congestion_control_object_debts: RefCountedHashMap::new(),
```
muXxer:
```suggestion
            congestion_control_object_debts: RefCountedHashMap::new(),
            congestion_control_randomness_object_debts: RefCountedHashMap::new(),
```
not a biggie, but my OCD says it needs to be in order :D

muXxer (approved):
LGTM, at least from what I understood 😸

muXxer:
Isn't it the case, that we don't initialize a new epoch with remaining debt from the former one?
Does that mean at the end of an epoch, we can overshoot, and that debt will never be repaid?

piotrm50 (approved):

PR #8901, opened 2025-10-15
feat(consensus): add `block_headers` and `transactions` to the `dag_state` atomicaly

### LINKED ISSUE #8900: starfish: refactor `BlockManager` to automicly write block_headers and transactions
Current  logic in `BlockManager` acquires the write lock 3 times in `try_accept_blocks`, twice in the called method `try_accept_block_headers_internal`, once to add block headers to be accepted, and the second time to add unsuspended transactions, and the third time to add transactions from new blocks.

The refactor would extract dag state write from `try_accept_block_headers_internal` and return the accepted_blocks and unsuspended transactions to the caller, that would then acquire the dag_state write lock only once.

filipdulic:
# Description of change

This PR refactors the `BlockManager` to streamline how block headers and transactions are written into the DAG state.  
Previously, the logic acquired the DAG state write lock three separate times in `try_accept_blocks` (twice in `try_accept_block_headers_internal` and once more for received blocks transactions).  
The refactor extracts DAG write operations out of the header acceptance logic and consolidates them into a single step, ensuring the lock is acquired only once.

### Key changes
- Renamed `try_accept_block_headers_internal` to `process_block_headers` method and refactored it to perform only classification and validation.
- Introduced `resolve_transactions` suspend/unsuspend transactions from incoming blocks and accepted headers.
- Added `write_block_headers_and_transactions_to_dag_state` to atomically persist both block headers and transactions to the DAG.
- Updated `try_accept_blocks` and `try_accept_block_headers` to use the new processing flow.

This results in cleaner separation of responsibilities, reduced lock contention, and improved overall concurrency behavior within the `BlockManager`.

Fixes #8900

### DIFF
--- a/crates/starfish/core/src/block_manager/mod.rs
+++ b/crates/starfish/core/src/block_manager/mod.rs
@@ -26,6 +26,7 @@ use crate::{
     Round,
     block_header::{
         BlockHeaderAPI, BlockRef, BlockTimestampMs, VerifiedBlock, VerifiedBlockHeader,
+        VerifiedTransactions,
     },
     block_manager::block_suspender::BlockSuspender,
     context::Context,
@@ -76,25 +77,18 @@ impl BlockManager {
             .iter()
             .map(|b| b.verified_block_header.clone())
             .collect();
-        let (accepted_block_headers, missing_block_headers) =
-            self.try_accept_block_headers_internal(block_headers);
+        let (block_headers_to_accept, missing_block_headers) =
+            self.process_block_headers(block_headers);
+        // collect suspended transactions for accepted headers.
+        let accepted_transactions =
+            self.resolve_transactions(&block_headers_to_accept, Some(blocks));
 
-        let block_refs = blocks
-            .iter()
-            .map(|b| b.verified_block_header.reference())
-            .collect();
-        let exists = self.dag_state.read().contains_block_headers(block_refs);
-        for (i, block) in blocks.into_iter().enumerate() {
-            if exists[i] {
-                self.dag_state
-                    .write()
-                    .add_transactions(block.verified_transactions, "Block streaming");
-            } else {
-                self.suspended_blocks.insert(block.reference(), block);
-            }
-        }
+        self.write_block_headers_and_transactions_to_dag_state(
+            block_headers_to_accept.clone(),
+            accepted_transactions,
+        );
 
-        (accepted_block_headers, missing_block_headers)
+        (block_headers_to_accept, missing_block_headers)
     }
 
     /// Tries to accept the provided block headers assuming that all their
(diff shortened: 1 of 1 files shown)


crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -135,26 +139,56 @@ impl BlockManager {
...
+    ) -> Vec<VerifiedTransactions> {
+        let block_refs_to_be_accepted = block_headers_to_be_accepted
+            .iter()
+            .map(|h| h.reference())
+            .collect::<BTreeSet<_>>();
+        let mut accepted_transactions_from_suspended_blocks = vec![];
+        for block_ref in block_refs_to_be_accepted.iter() {
+            if let Some(block) = self.suspended_blocks.remove(block_ref) {
+                // for this accepted header we already have a block, so we add it accepted
+                // transactions
```
piotrm50:
```suggestion
                // for this accepted header we already have a block, so we add it to 
                // accepted transactions
```

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -135,26 +139,56 @@ impl BlockManager {
...
+        if let Some(blocks) = blocks {
+            let mut accepted_transactions_from_blocks = vec![];
+            for block in blocks {
+                if block_refs_to_be_accepted.contains(&block.reference()) {
+                    accepted_transactions_from_blocks.push(block.verified_transactions);
+                } else {
+                    self.suspended_blocks.insert(block.reference(), block);
+                }
+            }
+            accepted_transactions_from_suspended_blocks.extend(accepted_transactions_from_blocks);
```
piotrm50: `accepted_transactions_from_suspended_blocks` this name is a bit confusing if also used here with non-suspended blocks. Perhaps rename to `accepted_transactions` or `all_accepted_transactions`

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -135,26 +139,56 @@ impl BlockManager {
         // Verify block timestamps
         let accepted_block_headers =
             self.verify_block_timestamps_and_accept(processed_block_headers);
+        (accepted_block_headers, ancestors_to_fetch)
+    }
+    fn write_block_headers_and_transactions_to_dag_state(
```
piotrm50, line 145:
```suggestion

    fn write_block_headers_and_transactions_to_dag_state(
```

piotrm50 (approved):
added small nits in the comments

crates/starfish/core/src/block_manager/mod.rs
filipdulic: fixed in latest push.

crates/starfish/core/src/block_manager/mod.rs
filipdulic: fixed in latest push

crates/starfish/core/src/block_manager/mod.rs
filipdulic: renamed to `all_accepted_transactions` in latest push.

polinikita (approved):

PR #8912, opened 2025-10-15
refactor: remove support for view calls in mainnet and testnet

### LINKED ISSUE #8843: [json-rpc,graphql]: restrict usage of the move-view api while in alpha
## Description

We need to restrict the API from being used in `mainnet`, and `testnet` until we reach `beta` version.

kodemartin:
# Description of change

We restrict usage of the move-view calls through JSON-RPC and GraphQL on networks other than `mainnet`, `testnet`. 

Fixes #8843

### Release Notes

- [x] JSON-RPC: Calls to the `iota_view` endpoint are only supported on  `devnet` , as the feature is still in alpha.
- [x] GraphQL: Usage of the `Query.moveViewCall` are only supported on `devnet`, as the feature is still in alpha.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/query.rs
+++ b/crates/iota-graphql-rpc/src/types/query.rs
@@ -100,8 +100,10 @@ impl Query {
     ) -> Result<MoveViewResult> {
         let chain_id_cache: &ChainIdentifierCache = ctx.data_unchecked();
 
+        let db = ctx.data_unchecked();
+        let metrics = ctx.data_unchecked();
         let chain = chain_id_cache
-            .read(ctx.data_unchecked(), ctx.data_unchecked())
+            .read(db, metrics)
             .await
             .extend()?
             .into_inner()
--- a/crates/iota-graphql-rpc/src/types/query.rs
+++ b/crates/iota-graphql-rpc/src/types/query.rs
@@ -12,6 +12,7 @@ use iota_json_rpc_types::{DevInspectArgs, IotaTypeTag};
 use iota_types::{
     TypeTag,
     gas_coin::GAS,
+    supported_protocol_versions::Chain,
     transaction::{TransactionData, TransactionDataAPI, TransactionKind},
 };
 use move_core_types::account_address::AccountAddress;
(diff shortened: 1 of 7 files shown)


sergiupopescu199 (approved):
lgtm 🌟

Having a separate error only for APIs in the indexer, I suppose, is an excellent idea; it would be easier to decouple the `IndexerError` type from API-specific errors.

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -204,6 +206,20 @@ impl WriteApiServer for OptimisticWriteApi {
...
         call_args: Vec<IotaJsonValue>,
     ) -> RpcResult<IotaMoveViewCallResults> {
+        let chain = self
+            .optimistic_tx_executor
+            .read
+            .get_chain_identifier_in_blocking_task()
+            .await?
+            .chain();
+        if !matches!(chain, Chain::Unknown) {
+            return Err(ApiError::UnsupportedFeature(format!(
```
tomxey, line 216: Have you considered using `IndexerError::NotSupported` ?
It's already used for not supported event filters

crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -97,6 +98,21 @@ impl Query {
         type_args: Option<Vec<String>>,
         arguments: Option<Vec<serde_json::Value>>,
     ) -> Result<MoveViewResult> {
+        let chain_id_cache: &ChainIdentifierCache = ctx.data_unchecked();
+
+        let chain = chain_id_cache
+            .read(ctx.data_unchecked(), ctx.data_unchecked())
```
tomxey: nit: would it make sense to assign those 2 arguments to variables first, so they are named and it's more clear what data is passed to read(..)?

tomxey: something like this:
```suggestion
        let db = ctx.data_unchecked();
        let metrics = ctx.data_unchecked();
        let chain = chain_id_cache
            .read(db, metrics)
```

tomxey (approved):

PR #8915, opened 2025-10-16
feat(iota-wallet & wallet-dashboard): show old browser banner

### LINKED ISSUE #6734: Display a "Not compatible" banner for old browsers running the wallet & dashboard
There are people who might be trying to use the wallet or the dashboard with old versions of their browsers, so old that even some APIs are not support (e.g structureClone, refer to Sentry errors for more details)

We decided to display some banner suggesting updating their browsers.

To detect whether a browser is enough up-to-date, we can use the errors we get plus when did the missing APIs get stable, and based on that just render the banner (or any other UI element) if the version is < what is considered up-to-date.

Juligs:
fixes #6734


apps/core/src/utils/isLegacyBrowser.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export function getBrowserCompatibilityMessage(): string | null {
+    if (typeof navigator === 'undefined') return null;
+
+    try {
+        const ua = navigator.userAgent;
+        const version = (re: RegExp) => Number(ua.match(re)?.[1] || 999);
+
+        const isLegacy =
+            version(/Chrome\/(\d+)/) < 98 ||
```
msarcev: Do we expect these breakpoint versions to change over time ? Should we have them in enums so it's slightly easier to find/change them ? Wdyt ?

apps/core/src/utils/isLegacyBrowser.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    useEffect(() => {
+        const msg = getBrowserCompatibilityMessage();
+        if (msg) setMessage(msg);
+    }, []);
+
+    if (!message) return null;
+
+    return (
+        <div className="fixed right-4 top-4 z-[9999] max-w-lg">
```
evavirseda: works great, maybe we can make the banner span the full width of the screen, wdyt?
Image

evavirseda (changes requested):

evavirseda (approved):

apps/core/src/components/buttons/ThemeSwitcher.tsx
```diff
@@ -3,8 +3,8 @@
 
 import { Button, ButtonType } from '@iota/apps-ui-kit';
 import { DarkMode, LightMode } from '@iota/apps-ui-icons';
-import { Theme, ThemePreference } from '../../enums';
-import { useTheme } from '../../hooks';
+import { Theme, ThemePreference } from '../../enums/theme.enums';
```
brancoder: why should we change these imports?  same for `ThemeContext` and `useThemes`? are we moving away from barrel files?

apps/wallet/src/ui/app/components/error-boundary/index.tsx
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) Mysten Labs, Inc.
-// Modifications Copyright (c) 2024 IOTA Stiftung
+// Modifications Copyright (c) 2025 IOTA Stiftung
```
brancoder: should we change to 2026

apps/wallet/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": "./configs/ts/tsconfig.dev",
-    "include": ["src", "configs", "tests"],
+    "include": ["src", "configs", "tests", "../core/src/utils/isLegacyBrowser.ts"],
```
brancoder: why is this add? also isLegacyBrowser has `.tsx ` extension?

apps/wallet/src/ui/app/components/error-boundary/index.tsx
```diff
@@ -8,6 +8,7 @@ import type { ReactNode } from 'react';
 import { ErrorBoundary as ReactErrorBoundary } from 'react-error-boundary';
 import type { FallbackProps } from 'react-error-boundary';
 import { useLocation } from 'react-router-dom';
+import { LegacyBrowserBanner } from '@iota/core/src/utils/isLegacyBrowser';
```
brancoder: as a side note:  in wallet dashboard we have mapping for imports from core.
we could add something like this in the `tsconfig.common.json` where we do all other mappings.
 ```
           "@iota/core/constants/*": ["../core/src/constants/*"],
            "@iota/core/utils/*": ["../core/src/utils/*"],
            "@iota/core/interfaces/*": ["../core/src/interfaces/*"]
```
maybe this can be done in a separate PR, because in wallet-dashboard we import  like this:
```
import {
    getBrowserCompatibilityMessage,
    LegacyBrowserBanner,
} from '@iota/core/utils/isLegacyBrowser';
```
wdyt?

**Review by brancoder — CHANGES_REQUESTED** [review]
LG just a few small comments

File: apps/wallet-dashboard/components/error-boundary/ErrorBoundary.tsx [hunk 7]
```diff
@@ -0,0 +1,46 @@
+'use client';
+// Copyright (c) Mysten Labs, Inc.
```
brancoder: for the new files we create we should use just Iota  header, no need to mention Mysten labs.

brancoder (changes requested):
LG just small nits like revert imports to files that are not part of the task

brancoder (approved):
LGTM

evavirseda (approved):
⭐

PR #8916, opened 2025-10-16
refactor(indexer): Split primary and snapshot executors

### LINKED ISSUE #8841: [indexer]: decouple object snapshots from ingestion pipeline
## Description

When run in `writer` mode, `iota-indexer` launches two concurrent pipelines that rely on checkpoint data: 

1. The main ingestion pipeline: Responsible for updating objects, transactions, and their indices
2. The object snapshots pipeline that captures an enriched snapshot of the ledger in the past. By default the lag of the snapshots from the current checkpoint number is 300 checkpoints.

These pipelines are coupled in using the same buffer of checkpoint data. As a consequence, the buffer size becomes an upper bound of the delay that can be sustained without one pipeline blocking the other. And hence it limits the effective lag that can be configure between snapshots and the current ledger state.

We would like to lift this restraint by using different buffers (different executors in `data-ingestion-core` jargon) for these two pipelines.

tomxey:
# Description of change

This PR splits executors for Primary and Snapshot pipelines in case it's possible (so in situation where user passes the remote_store_url as the source of checkpoints).
The code was refactored to leverage the fact that those are separate pipelines with separate executors.

fixes https://github.com/iotaledger/iota/issues/8841

colocated setup tested with `iota start` command
non colocated setup tested by running `iota start` and starting indexer separately
sync from genesis for mainnet data was NOT tested yet

### Release Notes

- [ ] Protocol:
- [x] Indexer:
   Passing `--remote-store-url` as the source of checkpoints will run the snapshot indexing pipeline using a separate executor, allowing much bigger config parameters for snapshot lag.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-indexer/src/indexer.rs
+++ b/crates/iota-indexer/src/indexer.rs
@@ -2,23 +2,14 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashMap, env, time::Duration};
+use std::env;
 
 use anyhow::{Context, Result};
-use async_trait::async_trait;
-use iota_data_ingestion_core::{
-    DataIngestionMetrics, IndexerExecutor, ProgressStore, ReaderOptions,
-};
+use iota_data_ingestion_core::ReaderOptions;
 use iota_metrics::spawn_monitored_task;
-use iota_types::messages_checkpoint::CheckpointSequenceNumber;
 use prometheus::Registry;
 use tokio_util::sync::CancellationToken;
-use tracing::info;
-
-use crate::ingestion::common::persist::Writer;
-
-/// Timeout for waiting for tasks to shutdown gracefully after cancellation
-const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
+use tracing::{info, warn};
 
 use crate::{
     build_json_rpc_server,
@@ -26,8 +17,7 @@ use crate::{
     db::ConnectionPool,
     errors::IndexerError,
     ingestion::{
-        primary::orchestration::{setup_primary, start_primary_writer_task},
-        snapshot::orchestration::setup_snapshot,
+        primary::orchestration::PrimaryPipeline, snapshot::orchestration::SnapshotPipeline,
     },
     metrics::IndexerMetrics,
     processors::processor_orchestrator::ProcessorOrchestrator,
@@ -54,34 +44,22 @@ impl Indexer {
         );
 
         info!("IOTA Indexer Writer config: {config:?}",);
-
-        let primary_watermark = store
-            .get_latest_checkpoint_sequence_number()
-            .await
-            .expect("failed to get latest tx checkpoint sequence number from DB")
-            .map(|seq| seq + 1)
-            .unwrap_or_default();
         let extra_reader_options = ReaderOptions {
             batch_size: config.checkpoint_download_queue_size,
             timeout_secs: config.checkpoint_download_timeout,
             data_limit: config.checkpoint_download_queue_size_bytes,
             ..Default::default()
         };
-
-        // Start objects snapshot processor, which is a separate pipeline with its
-        // ingestion pipeline.
-        let (object_snapshot_worker_pool, object_snapshot_writer, object_snapshot_receiver) =
-            setup_snapshot(
-                store.clone(),
-                metrics.clone(),
-                snapshot_config,
-                config.checkpoint_download_queue_size,
-            )
-            .await?;
-        let object_snapshot_watermark = object_snapshot_writer
-            .get_watermark_hi()
-            .await?
-            .unwrap_or_default();
+        let data_ingestion_path = config
+            .sources
+            .data_ingestion_path
+            .clone()
+            .unwrap_or(tempfile::tempdir().unwrap().keep());
+        let remote_store_url = config
+            .sources
+            .remote_store_url
+            .as_ref()
+            .map(|url| url.as_str().to_owned());
 
         if let Some(retention_config) = retention_config {
             let pruner = Pruner::new(store.clone(), retention_config, metrics.clone())?;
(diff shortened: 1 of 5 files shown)


crates/iota-indexer/src/ingestion/primary/orchestration.rs
```diff
@@ -4,52 +4,111 @@
...
-    let (indexed_checkpoint_sender, indexed_checkpoint_receiver) =
-        iota_metrics::metered_channel::channel(
-            checkpoint_queue_size,
-            &global_metrics
-                .channel_inflight
-                .with_label_values(&["checkpoint_indexing"]),
+pub(crate) struct PrimaryPipeline {
+    pub primary_executor: IndexerExecutor<ShimIndexerProgressStore>,
+    pub primary_writer: PrimaryWriter,
+    pub primary_watermark: CheckpointSequenceNumber,
```
kodemartin: You can omit the prefixes here as the name of the type can be construed as a namespace.

```rust
    pub executor: IndexerExecutor<ShimIndexerProgressStore>,
    pub writer: PrimaryWriter,
    pub watermark: CheckpointSequenceNumber,
```

crates/iota-indexer/src/ingestion/primary/orchestration.rs
```diff
@@ -4,52 +4,111 @@
...
+            .map(|seq| seq + 1)
+            .unwrap_or_default();
+        let primary_progress_store =
+            ShimIndexerProgressStore::new(vec![("primary".to_string(), primary_watermark)]);
+        let mut primary_executor = IndexerExecutor::new(
+            primary_progress_store,
+            1,
+            DataIngestionMetrics::new(&Registry::new()),
+            cancel.child_token(),
+        );
```
kodemartin: Constructing the progress store and the executor is something we do in various places now. So this warrants a new function, e.g. `common::orchestrator::new_executor` to avoid boilerplate, and have single reference point for maintenance.

crates/iota-indexer/src/ingestion/primary/orchestration.rs
```diff
@@ -4,52 +4,111 @@
...
+            .expect("failed to get latest tx checkpoint sequence number from DB")
+            .map(|seq| seq + 1)
+            .unwrap_or_default();
+        let primary_progress_store =
+            ShimIndexerProgressStore::new(vec![("primary".to_string(), primary_watermark)]);
+        let mut primary_executor = IndexerExecutor::new(
+            primary_progress_store,
+            1,
+            DataIngestionMetrics::new(&Registry::new()),
+            cancel.child_token(),
```
kodemartin: The `primary_` prefixes do not add more value here, as the scope is clearly defined. You can make the code less verbose and easier to read by avoiding the prefixes.

crates/iota-indexer/src/ingestion/primary/orchestration.rs
```diff
@@ -4,52 +4,111 @@
...
+    }
+
+    pub async fn run(
+        self,
+        data_ingestion_path: std::path::PathBuf,
+        remote_store_url: Option<String>,
+        reader_options: iota_data_ingestion_core::ReaderOptions,
+        cancel: CancellationToken,
+    ) -> IndexerResult<(
+        JoinHandle<iota_data_ingestion_core::IngestionResult<impl std::fmt::Debug>>,
```
kodemartin: Use the concrete type: `HashMap<String, CheckpointSequenceNumber>` instead of the opaque `Debug` generic.

crates/iota-indexer/src/ingestion/primary/orchestration.rs
kodemartin, line 1: The `primary_` prefixes to local variables or fields are redundant. Made a few specific comments but this applies to the entire module as well.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
kodemartin, line 1: Same as above: The `object_snapshot` prefixes are redundant making the code unnecessarily verbose.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,206 @@
...
+        metrics: IndexerMetrics,
+        snapshot_config: SnapshotLagConfig,
+        checkpoint_download_queue_size: usize,
+        cancel: CancellationToken,
+    ) -> IndexerResult<SnapshotPipeline> {
+        let writer = ObjectSnapshotWriter::new(state.clone(), metrics.clone(), snapshot_config);
+
+        let object_snapshot_watermark = writer.get_watermark_hi().await?.unwrap_or_default();
+        let mut snapshot_executor = IndexerExecutor::new(
+            ShimIndexerProgressStore::new(Default::default()),
```
kodemartin: This shouldn't be `Default::default`. The progress store requires the inner hashmap to be initialized with a task.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,206 @@
...
+            1,
+            DataIngestionMetrics::new(&Registry::new()),
+            cancel.child_token(),
+        );
```
kodemartin: As suggested above, this block warrants a dedicated constructor method on a higher module.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,206 @@
...
+    pub object_snapshot_receiver:
+        iota_metrics::metered_channel::Receiver<(u64, TransactionObjectChangesToCommit)>,
+}
+
+impl SnapshotPipeline {
+    pub async fn setup(
+        state: PgIndexerStore,
+        metrics: IndexerMetrics,
+        snapshot_config: SnapshotLagConfig,
+        checkpoint_download_queue_size: usize,
```
kodemartin, line 47: These arguments are shared between `setup` and `setup_with_shared_executor`, and I believe this module could benefit from introducing a struct in the line of the [builder](https://rust-lang.github.io/api-guidelines/type-safety.html?highlight=builder#builders-enable-construction-of-complex-values-c-builder) pattern. 

E.g. a `Setup` builder that constructs the writer and then `completes` by handling the two branches (shared/non-shared executors) and creating the `SnapshotPipeline`.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,206 @@
...
+                    info!(
+                        "Max committable checkpoint is {max_committable}, snapshottable data present",
+                    );
+                    break;
+                }
+                Ok(max_committable) => {
+                    info!(
+                        "Max committable checkpoint is {max_committable}, waiting for snapshottable data",
+                    );
+                    sleep(Duration::from_secs(1)).await;
```
kodemartin: warrants a constant `SnapshotLagPoll` instead of the hardcoded `Duration::from_secs(1)`.

crates/iota-indexer/src/indexer.rs
```diff
@@ -110,79 +88,103 @@ impl Indexer {
...
-                    executor_handle
-                ).await
-                .context("timeout waiting for executor to shutdown");
-                snapshot_result.context("failed to join snapshot task")?.context("snapshot task failed")?;
-                executor_result?.context("failed to join data ingestion executor during shutdown")?.context("data ingestion executor failed during shutdown")?;
+            result = &mut primary_executor_handle => {
+                result.context("failed to join primary executor")?.context("primary executor failed")?;
+                return Ok(());
             }
         };
```
kodemartin: This should probably handle the `primary_writer` as well.

But I think that `waiting` for the snapshot pipeline to start could be handled inside the `snapshot::orchestration`. 

I would suggest to reorganize the orchestration a bit, to avoid boilerplate and define different layers of orchestration scope.

* `snapshot::orchestration`: `Pipeline::run` can `wait_for_snapshot_lag` and then handle orchestration between `executor` and `writer` for graceful shutdown. You could for example use an `tokio::select` in `run` after `waiting`. 
* `primary::orchestration`: `Pipeline::run` can handle orchestration between `executor` and `writer` for graceful shutdown. 
* `indexer`: Handle orchestration between `{primary_,snapshot_}pipeline::run` for graceful shutdown.

kodemartin (changes requested):
I think the patch has some nice ideas, but it could use some polishing. Plus there a  couple of small oversights affecting correctness.

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: The progress store is initialized later in `register_on_executor`.

```
        executor
            .update_watermark("object_snapshot".to_string(), watermark)
            .await?;
```

So to keep the initialization in one place I've decided to construct empty progress store here.

kodemartin: Why to rely on a later side-effect of an executor method for the initialization and not do it up front? Seems a bit error prone: Calls to `load` would fail in between.

Plus it is opaque while reading.

tomxey: We need it that way for the "shared executor" case. Since we cannot supply the progress store then as this is done during executor creation, and we are not creating it. We can only modify the progress store in already existing executor. But we do it before starting the executor task, so I believe nothing should fail as the value is not used yet.

kodemartin: But this is not the shared executor you are constructing here. Generalizing `register_on_executor` should not affect this part. And there are multiple options to keep generality for `register_on_executor` and still make this more transparent: (a) passing `Option<u64>` for the watermark, or (b) delegate the `update_watermark` side-effect in the caller if possible.

Error prone does not mean it fails now, but rather than it requires maintainers to care at future iterations on this codebase for the proper initialization of the inner data after construction. Which can be done with if we just initialize here.

crates/iota-indexer/src/ingestion/primary/orchestration.rs
tomxey: Done in 9d360c8442db5a4de30408597f2203c64ad34d24

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: Done in 9d360c8442db5a4de30408597f2203c64ad34d24

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: Done in 9d360c8442db5a4de30408597f2203c64ad34d24

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: Done in ab94816419f6aa76a911a23892953911d581ced0

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: Done in 5cb5d6b96b6c507134f77f044a9c7511eba0ef80

crates/iota-indexer/src/indexer.rs
tomxey: Done in b1243e3dd0e8a7bc442008132172e70030f6e897

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: Done in c54d4be8167c0341b03ce4bb37cbe70b8517b679

crates/iota-indexer/src/ingestion/primary/orchestration.rs
tomxey: Done in c54d4be8167c0341b03ce4bb37cbe70b8517b679

crates/iota-indexer/src/ingestion/primary/orchestration.rs
tomxey: Done in c54d4be8167c0341b03ce4bb37cbe70b8517b679

crates/iota-indexer/src/ingestion/primary/orchestration.rs
tomxey: Done in c54d4be8167c0341b03ce4bb37cbe70b8517b679

crates/iota-indexer/src/ingestion/primary/orchestration.rs
tomxey: Done

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,220 @@
...
         primary::persist::TransactionObjectChangesToCommit,
         snapshot::{persist::ObjectSnapshotWriter, prepare::ObjectsSnapshotWorker},
     },
     metrics::IndexerMetrics,
+    spawn_monitored_task,
     types::IndexerResult,
 };
 
 const OBJECT_SNAPSHOT_CHANNEL_CAPACITY: usize = 600;
+const WAIT_FOR_SNAPSHOTTABLE_DATA_POLL_INTERVAL: Duration = Duration::from_secs(1);
```
kodemartin, line 32: Nit: I thought I had added a comment but apparently I hadn't.

I would like to suggest to rename `wait_for_snapshottable_data` into something that is more proper English and still maintain the meaning. `wait_for_snapshot_lag` could be an idea.

kodemartin (approved):
lgtm

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,220 @@
...
+        cancel: CancellationToken,
+    ) -> JoinHandle<IndexerResult<()>> {
+        spawn_monitored_task!(writer.persist_sequentially(receiver, cancel))
+    }
+
+    pub async fn run(
+        self,
+        remote_store_url: Option<String>,
+        reader_options: ReaderOptions,
+    ) -> IndexerResult<JoinHandle<IndexerResult<()>>> {
```
sergiupopescu199: seems like returning a result is not needed, removing it and returning the JoinHandle<T> does not produce errors

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
```diff
@@ -1,50 +1,220 @@
...
+            }
+
+            Ok(())
+        });
+
+        Ok(handle)
+    }
+}
+
+async fn wait_for_snapshottable_data(
```
sergiupopescu199: Something similar, to make it more transparent.

```suggestion
/// Fetches the latest committed checkpoint by the primary pipeline to determine
/// when the snapshot pipeline is ready to start.
async fn wait_for_snapshot_data(
```

sergiupopescu199 (approved):
lgtm!

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: done

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: done

crates/iota-indexer/src/ingestion/snapshot/orchestration.rs
tomxey: added a docstring

PR #8921, opened 2025-10-16
feat(starfish): cordial knowledge as separate component

### LINKED ISSUE #8850: starfish: isolate cordial knowledge and unknown headers/shards from the dag_state
To improve the performance, one needs to have only one thread for writing to the dag_state. The latest changes for improving the bandwidth introduced more time spent while writing to dag_state. This needs to be isolated such that authority_service component can read and write to this.

polinikita:
# Description of change

The new `CordialKnowledge` module centralizes and synchronizes knowledge about:
- which headers and shards are locally available, and  
- which ones are known or unknown to each peer.

It acts as a hub between the `DagState`, `AuthorityService`, and per-peer `ConnectionKnowledge` tasks.

### Flow Description

1. **DAG → CordialKnowledge**
When a new header or shard is produced or received:
- `DagState` sends a message (`NewHeader` or `NewShard`) to `CordialKnowledge`.
- `CordialKnowledge` updates its internal DAG knowledge and propagates the relevant info to each `ConnectionKnowledge` instance:
  - Which headers/shards are new.
  - Which should be removed as “unknown” for certain peers.

2. **CordialKnowledge → ConnectionKnowledge**
Each `ConnectionKnowledge` task (one per peer):
- Tracks what *that peer* knows or doesn’t know.
- Receives batched updates from the central `CordialKnowledge` loop through a bounded channel.
- Periodically receives “useful authors information” updates — rounds and authors that are still relevant for block dissemination.

3. **AuthorityService ↔ ConnectionKnowledge**
- When preparing or receiving block bundles`AuthorityService` asks a specific `ConnectionKnowledge` for **additional parts** (headers/shards/useful authors for shards/headers) to include.
- After receiving a bundle from some peer, `AuthorityService` sends an update back (via `UsefulAuthors` messages) indicating which peers should provide additional useful data. Those updates are forwarded both:
  - locally to the respected `ConnectionKnowledge`, and
  - globally via `CordialKnowledge` (for shards as `f+1` shards are needed for reconstruction).

4. **Eviction**
- When old DAG rounds are pruned, `DagState` notifies `CordialKnowledge` using`EvictBelow` message.
- `CordialKnowledge` updates metrics and notifies all connections to prune their local maps.

This refactor eliminates direct writes to `DagState` from multiple async contexts.  
All DAG updates now flow through the **Core Thread** via message passing, reducing contention.

One existing test was updated and two new tested were added:
1. `authority_service::tests::test_handle_subscribe_block_bundles_request` was updated to check that the headers and shards are pushed in the bundles as expected according the CordialKnowledge logic
2. `cordial_knowledge::tests::test_cordial_knowledge_bundle_with_byzantine` was added to check that the headers will be pushed to the peer in case a Byzantine node does not disseminate blocks properly.
3. `cordial_knowledge::tests::test_connection_knowledge_take_additional_parts` was added to check that the headers are taken to the bundle only once (evicted upon taking) and useful authors for shards and headers are calculated correctly using a constant `MAX_ROUND_GAP_FOR_USEFUL_PARTS`.

The change was successfully tested using a local network with some connections between peers being blocked.

Fixes #8850 

### DIFF
--- a/crates/starfish/core/src/cordial_knowledge.rs
+++ b/crates/starfish/core/src/cordial_knowledge.rs
@@ -356,7 +356,10 @@ impl CordialKnowledge {
 
     /// Called when a new own shard (created locally) is added to dag state.
     async fn handle_new_shard(&mut self, block_ref: BlockRef) {
-        for tx in &self.connections {
+        for (index, tx) in self.connections.iter().enumerate() {
+            if index == block_ref.author.value() || index == self.context.own_index.value() {
+                continue;
+            }
             let msg = ConnectionKnowledgeMessage::NewShard { block_ref };
             let _ = tx.send(vec![msg]).await;
         }
@@ -660,12 +663,18 @@ impl ConnectionKnowledge {
         let num_authorities = context.committee.size();
         let headers_not_known = vec![BTreeMap::new(); num_authorities];
         let shards_not_known = vec![BTreeMap::new(); num_authorities];
+        let mut last_useful_headers_from_peer_round =  vec![GENESIS_ROUND; num_authorities];
+        // We should not request headers for own blocks
+        last_useful_headers_from_peer_round[context.own_index] = Round::MAX;
+        let mut last_useful_shards_from_peer_round = vec![GENESIS_ROUND; num_authorities];
+        // We should not request shards for own blocks
+        last_useful_shards_from_peer_round[context.own_index] = Round::MAX;
         Self {
             dag_state,
             last_useful_headers_to_peer_round: vec![GENESIS_ROUND; num_authorities],
             last_useful_shards_to_peer_round: vec![GENESIS_ROUND; num_authorities],
-            last_useful_headers_from_peer_round: vec![GENESIS_ROUND; num_authorities],
-            last_useful_shards_from_peer_round: vec![GENESIS_ROUND; num_authorities],
+            last_useful_headers_from_peer_round,
+            last_useful_shards_from_peer_round,
             context,
             peer_index,
             headers_not_known,
@@ -1072,6 +1081,7 @@ mod tests {
         storage::mem_store::MemStore,
         test_dag_parser::parse_dag,
     };
+    use crate::test_dag_builder::DagBuilder;
 
     #[tokio::test]
     async fn test_cordial_knowledge_bundle_with_byzantine() {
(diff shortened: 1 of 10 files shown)


crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -0,0 +1,1405 @@
...
+            connection_handles: Mutex::new(connection_handles),
+            join_handle: Mutex::new(Some(join_handle)),
+        })
+    }
+
+    /// Main async loop: receives high-level updates (headers, shards,
+    /// evictions) from DAG state and updates global knowledge + notifies
+    /// per-connection tasks.
+    pub async fn run(mut self) {
+        fail_point_async!("consensus-rpc-response");
```
piotrm50: Why is this fail point registered here? It's not a `consensus-rpc-response` so it might not be valid to put it here.

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -0,0 +1,1405 @@
...
+    /// Processes a single high-level cordial knowledge message.
+    async fn process_message(&mut self, cordial_knowledge_message: CordialKnowledgeMessage) {
+        debug!(
+            "Processing Cordial Message: {:?}",
+            cordial_knowledge_message
+        );
+        let message_type: &'static str = match cordial_knowledge_message {
+            CordialKnowledgeMessage::NewHeader(header) => {
+                self.handle_new_header(header).await;
+                "header"
```
piotrm50: I think it would be cleaner to implement a method on the `CordialKnowledgeMessage` to return those names, just like we recently refactored the Transactions source to use an enum.

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -0,0 +1,1405 @@
...
+                useful_shards_to_peer: BTreeMap::new(),
+            };
+            if let Err(e) = connection_sender.send(vec![msg]).await {
+                warn!("Failed to send useful info to connection task: {}", e);
+            }
+        }
+    }
+
+    /// Called when a new verified block header is received.
+    async fn handle_new_header(&mut self, header: VerifiedBlockHeader) {
```
piotrm50: Do we need this method if it only calls another method inside?

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -0,0 +1,1405 @@
...
+        self.update_cordial_knowledge(&header).await;
+    }
+
+    /// Called when a new own shard (created locally) is added to dag state.
+    async fn handle_new_shard(&mut self, block_ref: BlockRef) {
+        for (index, tx) in self.connections.iter().enumerate() {
+            if index == block_ref.author.value() || index == self.context.own_index.value() {
```
piotrm50, line 454:
```suggestion
            // don't send shard to self nor to the author of the block
            if index == block_ref.author.value() || index == self.context.own_index.value() {
```

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Removed and left only internal one

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Changed

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Removed

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Implemented

piotrm50 (approved):

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -0,0 +1,1403 @@
...
+            if !msg.is_empty() {
+                let _ = self.connections[index].send(msg).await;
+            }
+        }
+    }
+
+    /// Report from Authority Service useful information about headers and
+    /// shards to global knowledge and connection knowledge.
+    pub async fn report_useful_info(
+        connection_knowledge_sender: &Sender<Vec<ConnectionKnowledgeMessage>>,
```
filipdulic: Nit: `CordialKnowledge` already has all the `connection_knowledge_senders` in it's internal structure, perhaps we can only pass the peer index here, and retrieve the sender from the internal structure.

filipdulic (approved):

PR #8927, opened 2025-10-16
feat(authority): support failing validation effects commit post-consensus

### LINKED ISSUE #8908: [AA] Handle MoveAuthenticator inputs the same as a the TX inputs
- Pre-consensus: input objects are read from the MoveAuthenticator in the same way any other input object is read from a TX/PTB. If validation fails, do NOT bump the sequence number, just drop everything. (This is in line with current implementation). 
- Post-consensus: input objects are read from the MoveAuthenticator in the same way any other input object is read from a TX/PTB. If validation fails, bump the sequence number of all input objects (coming from MoveAuthenticator and TX), deduct the validation gas computation from the TX gas object and return a TX execution error “Validation failed post-consensus“.
- Additionally, check any time tx.shared_input_objects() is invoked from the TransactionData and adapt this call to support MoveAuthenticator inputs when needed.

miker83z:
# Description of change
This makes so that a move authentication passing pre-consensus (validators sign the TX) and failing post-consensus produces some effects (gas charging).

I think starting from `test_abstract_account_post_consensus_failure()` is a good way to understand the logic behind the change. 
Basically, what is needed, is to make it so that this can happen:
1) Create an AA TX and make the majority of validators to sign it; this means that the Move authentication based on the AA shared object state is CURRENTLY successful.
2) Then, tamper with the AA shared object state by altering the state of the AA shared object (e.g., create and send for execution a second TX  that changes a field of the AA shared object).
3) Finally, submit for execution the original certificate, i.e., the one obtained from the signature of the majority of validators. This PR changes the behavior at this step: 
    - before this PR, the TX execution simply didn't go through but just raised a Move authentication error within the iota-core;
    - whilst, after this PR, the Move authentication error is treated as execution error and thus producing effects; the result is that all input objects are considered when creating the transaction effects and the gas coin is reduced with the authentication computation gas cost execution.

This PR also merges input objects in the case of a successful execution.

In order to produce effects, the method `authenticate_then_execute_transaction_to_effects()` was added to the execution engine. This new method executes all the steps necessary to create the TransactionEffects after the authentication execution. This means that, after running `authenticate_transaction_inner()` and failing (as well as succeeding), the execution continues with `execute_transaction_to_effects_inner()` in order to finish the job, i.e., creating effects for a failing (or successful) transaction. 

Fixes #8908 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1837,7 +1837,7 @@ impl AuthorityState {
             reference_gas_price,
             move_authentication_result
                 .as_ref()
-                .map(|(authentication_cost, _, _)| *authentication_cost)
+                .map(|(computation_cost, _, _)| *computation_cost)
                 .unwrap_or(0),
         )?;
 
@@ -1870,8 +1870,9 @@ impl AuthorityState {
                     )
                 }
                 Some((_, Err(move_authentication_error), authentication_checked_input_objects)) => {
-                    let (store, gas_status, effects, error) =
-                        epoch_store.executor().invalid_transaction_to_effects(
+                    let (store, gas_status, effects, error) = epoch_store
+                        .executor()
+                        .produce_effects_for_invalid_authentication(
                             backing_store,
                             protocol_config,
                             self.metrics.limits_metrics.clone(),
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1809,7 +1809,7 @@ impl AuthorityState {
                 // `max_auth_gas` is used here as a Move authenticator gas budget until it is
                 // not a part of the transaction data.
                 let authenticator_gas_budget = protocol_config.max_auth_gas();
-                let (gas_status, authentication_checked_input_objects, tx_checked_input_objects) =
+                let (gas_status, authenticator_checked_input_objects, authenticator_and_tx_checked_input_objects) =
                     iota_transaction_checks::check_certificate_and_move_authenticator_input(
                         certificate,
                         tx_input_objects.clone(), 
@@ -1819,7 +1819,7 @@ impl AuthorityState {
                         reference_gas_price,
                     )?;
 
-                let owned_object_refs = tx_checked_input_objects.inner().filter_owned_objects();
+                let owned_object_refs = authenticator_and_tx_checked_input_objects.inner().filter_owned_objects();
                 self.check_owned_locks(&owned_object_refs)?;
 
                 epoch_store
@@ -1838,11 +1838,11 @@ impl AuthorityState {
                         gas,
                         move_authenticator.to_owned(),
                         authenticator_info,
-                        authentication_checked_input_objects,
+                        authenticator_checked_input_objects,
+                        authenticator_and_tx_checked_input_objects,
                         kind,
                         signer,
                         tx_digest,
-                        tx_checked_input_objects,
                         &mut None,
                     )
             } else {
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1746,7 +1746,7 @@ impl AuthorityState {
         certificate: &VerifiedExecutableTransaction,
         tx_input_objects: InputObjects,
         account_object: Option<ObjectReadResult>,
-        authenticator_input_objects: Option<InputObjects>,
+        move_authenticator_input_objects: Option<InputObjects>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<(
         InnerTemporaryStore,
@@ -1778,68 +1778,51 @@ impl AuthorityState {
 
         let (kind, signer, gas) = tx_data.execution_parts();
 
-        let (authenticator_computation_cost, validation_result, authenticator_input_objects) =
+        // Check if the sender needs to be authenticated in Move and, if so, execute the
+        // authentication.
+        let move_authentication_result =
             if let Some(move_authenticator) = certificate.move_authenticator() {
-                // It is supposed that `Move authentication` availability is checked in
-                // `SenderSignedData::validity_check`.
-
-                // Check basic `object_to_authenticate` preconditions and get its components.
+                // Check the `MoveAuthenticator` input objects.
                 let (
-                    auth_account_object_id,
-                    auth_account_object_seq_number,
-                    auth_account_object_digest,
-                ) = move_authenticator.object_to_authenticate_components()?;
-
-                // Since the `object_to_authenticate` components are provided, it is supposed
-                // that the account object is loaded.
-                let account_object = account_object.expect("Account object must be provided");
-
-                let authenticator_info = self.check_move_account(
-                    auth_account_object_id,
-                    auth_account_object_seq_number,
-                    auth_account_object_digest,
+                    authentication_gas_status,
+                    authentication_checked_input_objects,
+                    authenticator_info,
+                ) = self.check_move_authenticator_inputs_for_executing(
+                    move_authenticator,
+                    move_authenticator_input_objects,
                     account_object,
-                    &signer,
+                    protocol_config,
+                    reference_gas_price,
+                    tx_data,
+                    &tx_input_objects,
+                    signer,
                 )?;
 
-                let authenticator_input_objects = authenticator_input_objects.expect(
-                "In case of a `MoveAuthenticator` signature, the authenticator input objects must be provided",
-            );
-
-                // Check the `MoveAuthenticator` input objects.
-                let (authenticator_gas_status, authenticator_input_objects) =
-                    Self::check_move_authenticator_inputs_for_executing(
-                        protocol_config,
-                        reference_gas_price,
-                        tx_data,
-                        authenticator_input_objects,
-                        &tx_input_objects,
-                    )?;
-
-                // Execute the Move authenticator.
-                let (authenticator_computation_cost, validation_result) =
+                // Execute the Move authentication.
+                let (authentication_computation_cost, authentication_result) =
                     epoch_store.executor().validate_transaction(
                         backing_store,
                         protocol_config,
                         self.metrics.limits_metrics.clone(),
                         &epoch_id,
                         epoch_start_timestamp,
-                        authenticator_gas_status,
+                        authentication_gas_status,
                         move_authenticator.to_owned(),
                         authenticator_info,
-                        authenticator_input_objects.clone(),
+                        authentication_checked_input_objects.clone(),
                         kind.clone(),
                         signer,
                         tx_digest,
                         &mut None,
                     );
-                (
-                    authenticator_computation_cost,
-                    validation_result,
-                    Some(authenticator_input_objects),
-                )
+
+                Some((
+                    authentication_computation_cost,
+                    authentication_result,
+                    authentication_checked_input_objects,
+                ))
             } else {
-                (0, Ok(()), None)
+                None
             };
 
         // The cost of partially re-auditing a transaction before execution is
@@ -1852,60 +1835,64 @@ impl AuthorityState {
             tx_input_objects,
             protocol_config,
             reference_gas_price,
-            authenticator_computation_cost,
+            move_authentication_result
+                .as_ref()
+                .map(|(authentication_cost, _, _)| *authentication_cost)
+                .unwrap_or(0),
         )?;
 
         let owned_object_refs = tx_input_objects.inner().filter_owned_objects();
         self.check_owned_locks(&owned_object_refs)?;
 
         #[cfg_attr(not(any(msim, fail_points)), expect(unused_mut))]
-        let (inner_temp_store, _, mut effects, execution_error_opt) = match validation_result {
-            Ok(_) => epoch_store.executor().execute_transaction_to_effects(
-                backing_store,
-                protocol_config,
-                self.metrics.limits_metrics.clone(),
-                // TODO: would be nice to pass the whole NodeConfig here, but it creates a
-                // cyclic dependency w/ iota-adapter
-                self.config
-                    .expensive_safety_check_config
-                    .enable_deep_per_tx_iota_conservation_check(),
-                self.config.certificate_deny_config.certificate_deny_set(),
-                &epoch_id,
-                epoch_start_timestamp,
-                tx_input_objects,
-                gas,
-                tx_gas_status,
-                kind,
-                signer,
-                tx_digest,
-                &mut None,
-            ),
-            Err(validation_error) => {
-                let authenticator_input_objects = authenticator_input_objects
-                    .expect("Inputs must be present if validation failed");
-                let (inner_temp_store, gas_status, mut effects, execution_error) =
-                    epoch_store.executor().invalid_transaction_to_effects(
+        let (inner_temp_store, _, mut effects, execution_error_opt) =
+            match move_authentication_result {
+                None | Some((_, Ok(()), _)) => {
+                    epoch_store.executor().execute_transaction_to_effects(
                         backing_store,
                         protocol_config,
                         self.metrics.limits_metrics.clone(),
+                        // TODO: would be nice to pass the whole NodeConfig here, but it creates a
+                        // cyclic dependency w/ iota-adapter
                         self.config
                             .expensive_safety_check_config
                             .enable_deep_per_tx_iota_conservation_check(),
                         self.config.certificate_deny_config.certificate_deny_set(),
                         &epoch_id,
                         epoch_start_timestamp,
-                        tx_gas_status,
-                        gas,
-                        validation_error,
-                        authenticator_input_objects,
                         tx_input_objects,
+                        gas,
+                        tx_gas_status,
                         kind,
                         signer,
                         tx_digest,
-                    );
-                (inner_temp_store, gas_status, effects, Err(execution_error))
-            }
-        };
+                        &mut None,
+                    )
+                }
+                Some((_, Err(move_authentication_error), authentication_checked_input_objects)) => {
+                    let (store, gas_status, effects, error) =
+                        epoch_store.executor().invalid_transaction_to_effects(
+                            backing_store,
+                            protocol_config,
+                            self.metrics.limits_metrics.clone(),
+                            self.config
+                                .expensive_safety_check_config
+                                .enable_deep_per_tx_iota_conservation_check(),
+                            self.config.certificate_deny_config.certificate_deny_set(),
+                            &epoch_id,
+                            epoch_start_timestamp,
+                            tx_gas_status,
+                            gas,
+                            move_authentication_error,
+                            authentication_checked_input_objects,
+                            tx_input_objects,
+                            kind,
+                            signer,
+                            tx_digest,
+                        );
+                    (store, gas_status, effects, Err(error))
+                }
+            };
 
         fail_point_if!("cp_execution_nondeterminism", || {
             #[cfg(msim)]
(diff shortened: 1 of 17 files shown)


iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -257,7 +257,7 @@ mod checked {
     /// happens since we cannot charge it separately in the current
     /// implementation.
     #[instrument(name = "tx_validate", level = "debug", skip_all)]
-    pub fn validate_transaction(
+    pub fn validate_transaction<Mode: ExecutionMode>(
```
valeriyr: We could leave it here, but I didn't add `Mode` here because this function works only with the `Validation` mode. Using any other one will lead to an error.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -405,7 +409,153 @@ mod checked {
...
+    }
+
+    /// This function produces transaction effects for a transaction that failed
+    /// the Move authentication. It charges gas for the failed execution of the
+    /// authentication and produces transaction effects with the appropriate
+    /// error status. It combines the input objects from both the failed
+    /// authentication and the original transaction and updates their sequence
+    /// numbers.
+    #[instrument(name = "produce_effects_for_invalid_auth", level = "debug", skip_all)]
+    pub fn produce_effects_for_invalid_authentication<Mode: ExecutionMode>(
```
valeriyr: I am not sure that I see a reason for adding such a function in the executor.
At the moment, `validate_transaction` is separated from the related transaction execution, but if we need to produce results in the case of a failed authenticator, it would be better to perform a refactoring and combine `validate_transaction` and `execute_transaction_to_effects`, and call a new function like `execute_transaction_with_move_authenticator`, or even integrate the authenticator logic inside `execute_transaction_to_effects`. In this case, all the related logic can be handled inside the executor in a single call.

valeriyr (changes requested):
In this PR, we change the conception of how a transaction with a move authenticator is executed. Since an authenticator can produce the result, I think we need to refactor the way it is called. We need a stronger connection between an authenticator and the transaction, and don't consider the authenticator call as something independent.

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: Removed

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: [15a69df](https://github.com/iotaledger/iota/pull/8927/commits/15a69df392f58bf499c14f98199e7ddfe35886c0)

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -171,6 +143,119 @@ async fn test_abstract_account_issues_sponsored_tx() -> Result<(), anyhow::Error
...
+        .authority_aggregator()
+        .process_certificate(
+            HandleCertificateRequestV1::new(cert).with_events(),
+            Some(client_ip),
+        )
+        .await
+        .unwrap();
+    let status = effects_cert.status().clone();
+
+    assert!(status.is_err(), "Expected the TX execution to fail");
```
TheMrAI: Shouldn't this test check that some gas was charged, given the failure?

crates/iota-types/src/gas_model/tables.rs
```diff
@@ -81,28 +79,10 @@ impl GasStatus {
...
     /// operations. This is the instantiation that must be used when
     /// executing a user script.
-    pub fn new(
-        cost_table: CostTable,
-        budget: u64,
-        gas_spent_for_authentication: u64,
-        gas_price: u64,
-        gas_model_version: u64,
-    ) -> Self {
+    pub fn new(cost_table: CostTable, budget: u64, gas_price: u64, gas_model_version: u64) -> Self {
```
TheMrAI, line 82: I see that the gas calculations have been changed back, I suppose to previous version.
What I don't understand why?
I remember talking with @valeriyr about the previous one, and having some vague sense that it made sense, but forgot since then.

crates/iota-types/src/transaction.rs
```diff
@@ -3299,6 +3304,15 @@ impl InputObjects {
         self.objects.push(object);
     }
 
+    pub fn extend_no_duplicates(&mut self, other: Self) {
```
TheMrAI: This will be an O(n^2) operation in terms of input objects. Is that okay?
Not familiar with this part of the code, just wondering if this is necessary at all.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -89,16 +85,13 @@ mod checked {
...
-        let authenticator_computation_cost = 0;
-
         let gas_status = check_transaction_input_inner(
             protocol_config,
             reference_gas_price,
-            authenticator_computation_cost,
             transaction,
             &input_objects,
             &[],
+            0, // authentication_gas_budget
```
TheMrAI: If you prefer to have the variable named, then keeping the original format is much more expressive.
Otherwise adding the comment of what the value represents only adds clutter to the code. One more thing that needs to be
maintained.
IDE's give you this information and while some extreme minority of people may not use an IDE, they generally don't have to
as they simply have the capacity to keep all this in mind.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -254,15 +233,72 @@ mod checked {
...
+        )?;
+
+        Ok((
+            gas_status,
+            authenticator_input_objects.into_checked(),
+            tx_input_objects.into_checked(),
+        ))
+    }
+
+    fn check_certificate_input_inner(
```
TheMrAI: Is this function really necessary? It seems to be only sphagettifying the code further.
For example above:
```
let gas_status = check_certificate_input_inner(
            cert,
            &tx_input_objects,
            protocol_config,
            reference_gas_price,
            authenticator_gas_budget,
        )?;
```
Could simply be:
```
let gas_status = check_transaction_input_inner(
            protocol_config,
            reference_gas_price,
            cert.data().transaction_data(),
            &tx_input_objects,
            &[],
            authenticator_gas_budget,
        )?;
```

Overall less mental load. One has to understand/remember one function, instead of two.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -242,6 +246,94 @@ mod checked {
...
+            authenticated_transaction_signer,
+            authenticated_transaction_digest,
+            trace_builder_opt,
+            move_vm,
+        );
+
+        // Merge input objects coming from the transaction together with
+        // those coming from the authenticator execution.
+        let mut input_objects = authenticated_transaction_input_objects;
+        input_objects.extend_no_duplicates(authenticator_input_objects);
```
TheMrAI: You can avoid introducing mutable variables into this scope by:
```
    let input_objects = {
        authenticated_transaction_input_objects.extend_no_duplicates(authenticator_input_objects);
        authenticated_transaction_input_objects
    };
```
But for this you have to make the authenticated_transaction_input_objects mutable on the function scope. This also avoids a copy.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -250,14 +342,10 @@ mod checked {
     /// the authentication function found in `AuthenticatorInfo`, that is
     /// retrieved from the abstracted IOTA account.
     ///
-    /// Returns an error if it happens or the move authenticator computation gas
-    /// cost without bucketing. It is different from the
-    /// `execute_transaction_to_effects` return type, because, in this case, we
-    /// do not need to return the gas value used for execution when an error
-    /// happens since we cannot charge it separately in the current
-    /// implementation.
+    /// Returns an error if it happens or and the gas status that can be later
```
TheMrAI: Some minor typo at `or and`

crates/iota-core/src/authority.rs
```diff
@@ -1778,116 +1778,109 @@ impl AuthorityState {
...
-
-            // Check the `MoveAuthenticator` input objects.
-            let (authenticator_gas_status, authenticator_input_objects) =
-                Self::check_move_authenticator_inputs_for_executing(
-                    protocol_config,
-                    reference_gas_price,
-                    tx_data,
-                    authenticator_input_objects,
-                    &tx_input_objects,
+        #[cfg_attr(not(any(msim, fail_points)), expect(unused_mut))]
```
TheMrAI, line 1781: What is the need for the conditional compilation flags?

TheMrAI (approved):
I checked the code mostly from technical perspective. Can't really comment on the design choices, will trust @valery to handle that.

Only thing concerning was the O(n^2) object merger, but lack the deep knowledge on the domain to suggest something better.
With the above disclaimers I approve.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -542,7 +637,7 @@ mod checked {
 
         // At this point no charges have been applied yet
         debug_assert!(
-            gas_charger.no_charges_for_execution(),
+            move_authentication_result_opt.is_some() || gas_charger.no_charges_for_execution(),
```
valeriyr: The `no_charges_for_execution` function can be deleted; `no_charges` checks the same condition.

iota-execution/latest/iota-adapter/src/execution_engine.rs
valeriyr: I propose to consider an approach where `execute_transaction_to_effects` is not called if `authenticate_transaction` fails. It looks logically more correct.
Additionally, it appears suspicious that we add the authenticator inputs to the transaction inputs list; we don't expect them to be used in a transaction call, so it looks unexpected, and it is a potential security issue, the transaction scope should be limited as much as possible.

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: Added in [c508e9b](https://github.com/iotaledger/iota/pull/8927/commits/c508e9b5bad9d2fff164571cf7d9d0dc252bea2b)

crates/iota-types/src/gas_model/tables.rs
miker83z: Previous changes assumed we used a distinct IotaGasStatus for both executions (validation and normal). In that case the second IotaGasStatus (used for normal execution) was init with a debt coming from the previous execution (validation). 

In this PR, the assumption is that we can either use the same IotaGasStatus for both executions ([350aed5](https://github.com/iotaledger/iota/pull/8927/commits/350aed5099504929aa031e6d1685e4b67c7a4afe)), or init the second IotaGasStatus without debt but charging the validation used computation gas during the normal execution ([15a69df](https://github.com/iotaledger/iota/pull/8927/commits/15a69df392f58bf499c14f98199e7ddfe35886c0)). Either one of the 2 should work well in our case (currently the first one is implemented)

piotrm50 (approved):
Approving files owned by consensus - `iota-core`, `iota-transaction-checks` and `iota-types`

crates/iota-types/src/execution.rs
```diff
@@ -33,7 +33,7 @@ pub type DeletedSharedObjectInfo = (ObjectID, SequenceNumber, bool, TransactionD
 /// inputs.
 pub type DeletedSharedObjects = Vec<DeletedSharedObjectInfo>;
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
valeriyr: Do we still need `Hash` trait?

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -250,14 +383,10 @@ mod checked {
...
-    /// cost without bucketing. It is different from the
-    /// `execute_transaction_to_effects` return type, because, in this case, we
-    /// do not need to return the gas value used for execution when an error
-    /// happens since we cannot charge it separately in the current
-    /// implementation.
+    /// Returns an error if it happens and the gas status that can be later
+    /// used to check the computation cost.
     #[instrument(name = "tx_validate", level = "debug", skip_all)]
-    pub fn validate_transaction(
+    pub fn authenticate_transaction(
```
valeriyr, line 419: Should we change the mode name as well? To `MoveAuthentication`, for example.

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -171,6 +142,129 @@ async fn test_abstract_account_issues_sponsored_tx() -> Result<(), anyhow::Error
...
+            HandleCertificateRequestV1::new(cert).with_events(),
+            Some(client_ip),
+        )
+        .await
+        .unwrap();
+    let summary = effects_cert.summary_for_debug();
+
+    assert!(summary.status.is_err(), "Expected the TX execution to fail");
+    assert!(
+        summary.gas_used.gas_used() > 0
```
valeriyr: Should we check the amount of gas here to pay attention in case it changes?

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -171,6 +142,129 @@ async fn test_abstract_account_issues_sponsored_tx() -> Result<(), anyhow::Error
...
+            && summary.unwrapped_object_count == 0
+            && summary.deleted_object_count == 0
+            && summary.wrapped_object_count == 0,
+        "Expected gas to be used in the failed transaction and that only the gas object was mutated and the TX input object was bumped in version",
+    );
+
+    assert!(
+        matches!(
+            summary.status.unwrap_err().0,
+            ExecutionFailureStatus::MoveAbort(MoveLocation { module, function_name, .. }, _)
```
valeriyr: Can we check the error code as well? In case of a clever error, it can be done like this, as I see:
https://github.com/iotaledger/iota/blob/e915cb4837a8bbb70e80126412464313a26db9b7/crates/iota/src/clever_error_rendering.rs#L203
It would be nice to create an issue to extend the `MoveAbort` error with additional information in this case. It is important to see the code when an authenticator fails.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -254,15 +238,55 @@ mod checked {
...
+        let gas_status = check_transaction_input_inner(
+            protocol_config,
+            reference_gas_price,
+            transaction,
+            &tx_input_objects,
+            &[],
+            authenticator_gas_budget,
+        )?;
+
+        check_and_extend_input_objects(&mut tx_input_objects, &authenticator_input_objects)?;
```
valeriyr: It would be nice to shortly explain why we merge the inputs here; in a comment ot in the function doc.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -770,6 +782,81 @@ mod checked {
...
+                                    // input_mutable is false, then swap
+                                    if additional_mutable && !input_mutable {
+                                        existing_object.input_object_kind =
+                                            InputObjectKind::SharedMoveObject {
+                                                id: input_id,
+                                                initial_shared_version:
+                                                    input_initial_shared_version,
+                                                mutable: true,
+                                            };
+                                    }
```
valeriyr: What is a use case for this?
Authenticator inputs are always immutable, so it might make sense to create a more specific function to extend transaction inputs with authenticator ones?
Can we consider this function(`check_and_extend_input_objects`) as a method of the `InputObjects` struct? It looks generic enough to be added as `extend_no_duplicates`, like it was done before.

crates/iota-types/src/transaction.rs
```diff
@@ -2900,13 +2900,13 @@ impl InputObjectKind {
...
 /// ObjectReadResultKind::Deleted is returned.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct ObjectReadResult {
     pub input_object_kind: InputObjectKind,
     pub object: ObjectReadResultKind,
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
```
valeriyr, line 2940: Just to double-check if we still need these traits.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -106,27 +106,18 @@ mod checked {
...
         let receiving_objects = transaction_kind.receiving_objects();
-        let mut transaction_dependencies = input_objects.transaction_dependencies();
-        let contains_deleted_input = input_objects.contains_deleted_objects();
-        let cancelled_objects = input_objects.get_cancelled_objects();
 
-        let mut temporary_store = TemporaryStore::new(
+        let temporary_store = TemporaryStore::new(
             store,
-            input_objects,
+            input_objects.clone(),
```
valeriyr: Maybe it makes sense to add a comment that after creating a `TemporaryStore` instance, `input_objects` can be used for validation/verification only?
Or split them into components before `TemporaryStore` creation with a utility function, and not to clone? `Clone` trait can be removed from `CheckedInputObjects` in this case.

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: Yes, `Authentication` in the `iota-execution` context might suffice

crates/iota-transaction-checks/src/lib.rs
valeriyr: In this PR, we will need to push the account object into the list as well: https://github.com/iotaledger/iota/pull/8986

crates/iota-types/src/transaction.rs
miker83z: Yes, these are needed for the check_and_extend_input_objects

crates/iota-transaction-checks/src/lib.rs
miker83z: About the `if additional_mutable && !input_mutable` check:
- this extensions mutates the left side (input) and reads from the right side (additional);
- where we now call it, we now that the left side is tx_inputs and right side is auth_inputs;
- however, in a more general case, if we call `check_and_extend_input_objects` we assume the left and right have been checked already to be good for their cases (now that I am thinking about it, the function signature should be (input_objects: &mut CheckedInputObjects,additional_objects: &CheckedInputObjects), changed in [a5589fc](https://github.com/iotaledger/iota/pull/8927/commits/a5589fc1eb54ae6898a09048c3f20e111e353e04));
- this means that the right side, might have a shared object which is mutable (maybe in the authenticatorV2?) and we are good with it;
- then, from the point of view of the check_and_extend_input_objects function, if one of the sides has `mutable:true`, then in the final result the mutable must be true; this because keeping `mutable:false` in the final result when one of the two sides had `mutable:true` will break the check that happened before calling `check_and_extend_input_objects`.

miker83z: About the rest of the comments: yes, we need to push the account object as well and potentially also the sponsor authenticator input objects; this I think it could be better handled in a dedicated issue.

crates/iota-types/src/transaction.rs
miker83z: Removed from `ObjectReadResult `

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: reverted in [b3fad83](https://github.com/iotaledger/iota/pull/8927/commits/b3fad83b6687f66b2035a34258a50d5e0da40040)

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: [53fa2b7](https://github.com/iotaledger/iota/pull/8927/commits/53fa2b70370a3ba235700ecbdb82bfe61613cc1d)

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: [53fa2b7](https://github.com/iotaledger/iota/pull/8927/commits/53fa2b70370a3ba235700ecbdb82bfe61613cc1d)

miker83z: @valeriyr 
> It would be nice to create an issue to extend the MoveAbort error with additional information in this case. It is important to see the code when an authenticator fails.

What do you mean? Could you make an example?

crates/iota-core/src/authority.rs
miker83z: It was already there; the unused mut is for the `effects`, but the rest I haven't checked.

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: Removed in later commits.

crates/iota-transaction-checks/src/lib.rs
miker83z: Changed in later commits, now the implementations of `check_certificate_and_move_authenticator_input` and `check_certificate_input` are different.

crates/iota-transaction-checks/src/lib.rs
miker83z: Removed

crates/iota-types/src/transaction.rs
miker83z: This function has been removed, but `checked_input_objects_union` has a similar implementation. Time complexity is still O(n^2), but in this case it will be more complex to refactor the types so that we can use ordered sets or hashes. Currently this operation is capped to `max_input_objects` x `max_function_parameters` = `2048 x 148`

crates/iota-e2e-tests/tests/abstract_account_tests.rs
valeriyr: In case of a clever error, MoveAbort contains an error value that needs to be decoded:
```
 abort_code: 9223372105574252543
```
In this case, we can add a decoded optional user-friendly field to the error itself, for example:
```
 clever_abort_code: Some(
    ErrorBitset {
        version: 8,
        reserved: 0,
        error_code: 0,
        line_number: 15,
        identifier_index: 65535,
        constant_index: 65535,
    },
)
```

crates/iota-transaction-checks/src/lib.rs
valeriyr: idk, for me this logic is misleading, because in our case, if the right side(auth inputs) contains mutable shared objects, it is an error which shouldn't happen. We should have at least a debug assert here, since we don't really expect such inputs here. We can remove this assert when it is needed.

valeriyr (approved):

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: #9037

PR #8935, opened 2025-10-17
fix(graphql-rpc): read also checkpointed transactions after execution

kodemartin:
# Description of change

Fixes `mutation.executeTransactionBlock` which could fail while trying to construct the transaction response. The reason being that it was expecting that the transaction would be indexed through the optimistic path, which is not necessarily true.

### Release Notes

- [x] GraphQL: Fixes `mutation.executeTransactionBlock` which could fail if the transaction executed was indexed through the checkpoint path.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
@@ -610,7 +610,7 @@ impl TryFrom<StoredTransaction> for TransactionBlock {
 
     fn try_from(stored_tx: StoredTransaction) -> Result<Self, Error> {
         let checkpoint_viewed_at = stored_tx.checkpoint_sequence_number as u64;
-        let inner = TransactionBlockInner::try_from(stored_tx.clone())?;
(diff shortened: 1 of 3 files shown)


crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -605,6 +605,20 @@ impl TryFrom<StoredTransaction> for TransactionBlockInner {
     }
 }
 
+impl TryFrom<StoredTransaction> for TransactionBlock {
+    type Error = Error;
+
+    fn try_from(stored_tx: StoredTransaction) -> Result<Self, Error> {
+        let checkpoint_viewed_at = stored_tx.checkpoint_sequence_number as u64;
+        let inner = TransactionBlockInner::try_from(stored_tx.clone())?;
```
sergiupopescu199:
```suggestion
        let inner = TransactionBlockInner::try_from(stored_tx)?;
```

sergiupopescu199 (approved):
lgtm ✨

msarcev (approved):

Issue #8849, opened 2025-10-13
[AA][authority] Optimize authenticator inputs loading


miker83z:
In `handle_transaction()`, during `check_move_authenticator_inputs_for_signing()` the function `read_objects_for_signing()` is invoked twice. In both cases, it can happen that an object is read twice because it is both an input of the authenticator and of the TX. 

The inline TODO comment says 
```
// TODO: Since the authenticator input is loaded with a dedicated call, we can
// load the same object twice if it is presented in the transaction and
// authenticator inputs. 
```

In check_move_authenticator_inputs_for_signing we already uses the parameter tx_input_objects: &InputObjects 
- then, get the let auth_temp_inputs = &move_authenticator.object_to_authenticate().input_objects() and
- filter these by object id;
  - if the id is found in tx_input_objects then clone this InputObject
  - remove the input from auth_temp_inputs 
  - with the clone, start filling authenticator_input_objects
- This can be done before invoking  self.input_loader.read_objects_for_signing for the input objects, but a similar logic can be done also for the account object
- (additionally, the account object could be unified to the auth_temp_inputs for this process)

Something similar can be done for `read_objects_for_execution()` in `prepare_certificate()`.

PR #8936, opened 2025-10-17
fix(iota-core, transaction-checks): Avoid gas coins double check 

### LINKED ISSUE #8847: [AA][tx-checks] Avoid gas coins double check
Gas coins should not be double-checked if `GenericSignature` is of type `MoveAuthenticator`. This happens in `iota-core/authority` when `check_move_authenticator_input()` is firstly invoked and, after validation, `check_transaction_input()` or `check_certificate_input()` are invoked.

Modify `check_transaction_input()`, `check_transaction_input_with_given_gas()`, `check_certificate_input()` in `iota-transaction-checks` to support a bool flag to pass down to `check_gas()`. This would allow to execute the checking only when the flag is true.

Dkwcs:
[run-ci]

# Description of change

Avoid double gas checks during signing and execution stages in case of move authentication.

This PR introduces a boolean `is_gas_check_required` flag in the transaction-check flow to avoid double gas validation when a transaction has a `MoveAuthenticator`.

Previously, gas coins were validated twice during transaction signing:
- During `check_move_authenticator_input()` - where gas sufficiency is verified for both the authenticator and transaction execution.
- Again in `check_transaction_input()` or `check_certificate_input()` - performing the same checks a second time.

The main change is located in `iota-transaction-checks/lib.rs/check_gas`:
 
```
if is_gas_check_required {
    // Check the balance and coins consistency; Load all the gas coins.
    let objects: BTreeMap<_, _> = objects.iter().map(|o| (o.id(), o)).collect();
    let mut gas_objects = vec![];
    for obj_ref in gas {
        let obj = objects.get(&obj_ref.0);
        let obj = *obj.ok_or(UserInputError::ObjectNotFound {
            object_id: obj_ref.0,
            version: Some(obj_ref.1),
        })?;
        gas_objects.push(obj);
    }
    gas_status.check_gas_balance(&gas_objects, gas_budget_to_check)?;
}
```
        
Fixes #8847 .

Run tests for `iota-core`

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -931,7 +931,9 @@ impl AuthorityState {
                     tx_data,
                     is_gas_check_required,
                 )?;
-            //
+
+            // After checking the `MoveAuthenticator` inputs, gas check is not required
+            // in order to avoid gas double-checking.
             is_gas_check_required = false;
             let (kind, signer, _) = tx_data.execution_parts();
 
@@ -2183,7 +2185,7 @@ impl AuthorityState {
     ) -> IotaResult<SimulateTransactionResult> {
         // Cheap validity checks for a transaction, including input size limits.
         transaction.validity_check_no_gas_check(epoch_store.protocol_config())?;
-
+        let is_gas_check_required = true;
         let input_object_kinds = transaction.input_objects()?;
         let receiving_object_refs = transaction.receiving_objects();
 
@@ -2233,7 +2235,7 @@ impl AuthorityState {
                     gas_object,
                     &self.metrics.bytecode_verifier_metrics,
                     &self.config.verifier_signing_config,
-                    true,
+                    is_gas_check_required,
                 )?,
                 Some(gas_object_id),
             )
@@ -2247,7 +2249,7 @@ impl AuthorityState {
                     &receiving_objects,
                     &self.metrics.bytecode_verifier_metrics,
                     &self.config.verifier_signing_config,
-                    true,
+                    is_gas_check_required,
                 )?,
                 None,
             )
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -881,7 +881,7 @@ impl AuthorityState {
     ) -> IotaResult<VerifiedSignedTransaction> {
         // Ensure that validator cannot reconfigure while we are signing the tx
         let _execution_lock = self.execution_lock_for_signing();
-
+        let mut is_gas_check_required = true;
         let protocol_config = epoch_store.protocol_config();
         let reference_gas_price = epoch_store.reference_gas_price();
 
@@ -929,8 +929,10 @@ impl AuthorityState {
                     protocol_config,
                     reference_gas_price,
                     tx_data,
+                    is_gas_check_required,
                 )?;
-
+            //
+            is_gas_check_required = false;
             let (kind, signer, _) = tx_data.execution_parts();
 
             // Execute the Move authenticator.
@@ -960,8 +962,6 @@ impl AuthorityState {
             }
         }
 
-        // TODO: Gas coins should not be double-checked if `GenericSignature` is of type
-        // `MoveAuthenticator`.
         let (_gas_status, tx_checked_input_objects) =
             iota_transaction_checks::check_transaction_input(
                 protocol_config,
@@ -971,6 +971,7 @@ impl AuthorityState {
                 &tx_receiving_objects,
                 &self.metrics.bytecode_verifier_metrics,
                 &self.config.verifier_signing_config,
+                is_gas_check_required,
             )?;
 
         check_coin_deny_list_v1_during_signing(
@@ -1755,6 +1756,7 @@ impl AuthorityState {
     )> {
         let _scope = monitored_scope("Execution::prepare_certificate");
         let _metrics_guard = self.metrics.prepare_certificate_latency.start_timer();
+        let mut is_gas_check_required = true;
         let prepare_certificate_start_time = tokio::time::Instant::now();
 
         let protocol_config = epoch_store.protocol_config();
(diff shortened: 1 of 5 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -929,8 +929,10 @@ impl AuthorityState {
                     protocol_config,
                     reference_gas_price,
                     tx_data,
+                    is_gas_check_required,
                 )?;
-
+            //
```
miker83z: I guess you forgot to put the comment here

crates/iota-core/src/authority.rs
```diff
@@ -2226,6 +2233,7 @@ impl AuthorityState {
                     gas_object,
                     &self.metrics.bytecode_verifier_metrics,
                     &self.config.verifier_signing_config,
+                    true,
```
miker83z: I think it would be better to use the `is_gas_check_required` variable here, like in the other functions above.

crates/iota-core/src/authority.rs
```diff
@@ -2239,6 +2247,7 @@ impl AuthorityState {
                     &receiving_objects,
                     &self.metrics.bytecode_verifier_metrics,
                     &self.config.verifier_signing_config,
+                    true,
```
miker83z: I think it would be better to use the `is_gas_check_required` variable here, like in the other functions above.

miker83z (approved):
LGTM! 
I have just a couple of minor comments and then let's wait for the CI tests to become all green!

crates/iota-core/src/authority.rs
Dkwcs: Done

crates/iota-core/src/authority.rs
Dkwcs: Done

crates/iota-core/src/authority.rs
Dkwcs: Done

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -230,6 +238,7 @@ mod checked {
         gas_price: u64,
         authenticator_input_objects: InputObjects,
         tx_input_objects: &InputObjects,
+        is_gas_check_required: bool,
```
TheMrAI, line 241: Could this be removed? I checked the call paths and this seems to be always true.
Noticed it because of the name of the function.
`check_move_authenticator_input` suggests to me we already know this is move authenticator input, so we should know if special checks are required.

crates/iota-core/src/authority.rs
TheMrAI: Sorry for being late, but I have to disagree.
`is_gas_check_required` isn't necessary to appear as a local variable.
Especially not a mutable one. Here the variable serves no purpose but that of being a flag carrier, and for reusability it is now far from the point of impact.

At the callsite should one pass the desired value, and if it needs any explanation it should be explained right there.
You can simply shadow the value:
```
let is_gas_check_required = true;
...
let is_gas_check_required = false;
```
This is still nasty but better than being mutable.
Ideally we should just pass a true/false value and comment it if necessary.

crates/iota-core/src/authority.rs
```diff
@@ -5500,6 +5513,7 @@ impl AuthorityState {
         protocol_config: &ProtocolConfig,
         reference_gas_price: u64,
         transaction: &TransactionData,
+        is_gas_check_required: bool,
```
TheMrAI, line 5516: Maybe this isn't necessary either?

crates/iota-core/src/authority.rs
```diff
@@ -5614,6 +5629,7 @@ impl AuthorityState {
         transaction: &TransactionData,
         authenticator_input_objects: InputObjects,
         tx_input_objects: &InputObjects,
+        is_gas_check_required: bool,
```
TheMrAI, line 5632: Maybe this one?

TheMrAI (changes requested):
I can't quite comment on the logic implementation, but I would like to request to clean up the is_gas_check_required usage.
Bring the value closer to the callsite, remove mutability and remove uses that are potentially unnecessary.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -428,18 +441,20 @@ mod checked {
...
+                let obj = objects.get(&obj_ref.0);
+                let obj = *obj.ok_or(UserInputError::ObjectNotFound {
+                    object_id: obj_ref.0,
+                    version: Some(obj_ref.1),
+                })?;
+                gas_objects.push(obj);
+            }
+            gas_status.check_gas_balance(&gas_objects, gas_budget_to_check)?;
         }
-        gas_status.check_gas_balance(&gas_objects, gas_budget_to_check)?;
```
valeriyr, line 442: Currently, `gas_budget_to_check` is different depending on where this function is called:
```rust
authenticator:
let gas_budget_to_check = authenticator_gas_budget + transaction_gas_budget;

transaction:
let gas_budget_to_check = authenticator_computation_cost + transaction_gas_budget;
```
It is not enough just to add a flag and disable this check. At least we need to clean up the code that is not required.

valeriyr (changes requested):
@mirko @Dkwcs I am not sure why this PR is a draft, but overall, this solution doesn't look good to me.
I would prefer to consider a way to deal with this problem differently. We need to integrate the authenticator logic closer to the transaction one, because we want to produce the results from authentication([8927](https://github.com/iotaledger/iota/pull/8927)), we need to load(and probably check) authenticator inputs together with the related transaction ones([8849](https://github.com/iotaledger/iota/issues/8849)), the same is relevant to gas checking.
So, as I see with each change, the connection between the authenticator logic and the transaction one becomes closer and closer, which is different from the original implemented concept, and adding such flags doesn't help us to manage this normally.

crates/iota-transaction-checks/src/lib.rs
Dkwcs: Could you elaborate a little bit more?
Do you mean we have to provide not only a flag or even not a flag at all, but another mechanism of computation the gas_budget according to authenticator functionality?

TheMrAI (approved):

miker83z:
Closed in favor of #9100

PR #8941, opened 2025-10-19
fix(iota-core, transaction-checks): Harmonize check_transaction_input and check_transaction_input_with_given_gas 

### LINKED ISSUE #8848: [AA][tx-checks] Harmonize check_transaction_input and check_transaction_input_with_given_gas
In these functions there's a comment which says `MoveAuthenticator computation cost is not used here at the moment`. This because: 
- the IotaGasStatus returned by the two functions is not used 
- OR it is used, but in a DryRun/DevInspect tx 

Add parameter authenticator_computation_cost: u64 to both function and:
- set it to 0 in the callers that use no authenticate 
- use the validation_result in `iota-core::authority::handle_transaction_impl()`

Dkwcs:
[run-ci]

# Description of change

This PR introduces an explicit `authenticator_computation_cost` parameter in `check_transaction_input` and `check_transaction_input_with_given_gas` functions to represent the gas cost computation by the execution of a `MoveAuthenticator` validation.
Previously, comments in the code indicated that “MoveAuthenticator computation cost is not used here at the moment”. This PR resolves that limitation by properly propagating and using the authenticator gas computation cost value.

Fixes #8848 .

Run tests for iota-core

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -976,8 +976,8 @@ impl AuthorityState {
                 &tx_receiving_objects,
                 &self.metrics.bytecode_verifier_metrics,
                 &self.config.verifier_signing_config,
-                is_gas_check_required,
                 authenticator_computation_cost,
+                is_gas_check_required,
             )?;
 
         check_coin_deny_list_v1_during_signing(
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -884,6 +884,7 @@ impl AuthorityState {
         let mut is_gas_check_required = true;
         let protocol_config = epoch_store.protocol_config();
         let reference_gas_price = epoch_store.reference_gas_price();
+        let mut authenticator_computation_cost = 0;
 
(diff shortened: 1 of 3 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -2041,6 +2046,7 @@ impl AuthorityState {
                     gas_object,
                     &self.metrics.bytecode_verifier_metrics,
                     &self.config.verifier_signing_config,
+                    authenticator_computation_cost,
```
TheMrAI, line 2049: Can we put `authenticator_computation_cost` closer to its use in all cases?
Also in most of these functions it doesn't even need to exist as it will be a constant 0 value.

TheMrAI (changes requested):
Please remove `authenticator_cumputation_cost` variables and pass directly 0 when it is only zero, or if it absolutely necessary to have it as a named variable, bring it as close as possible to the callsite.

TheMrAI (approved):

miker83z:
Not necessary after #8927

PR #8951, opened 2025-10-20
feat(move-compiler): Add authenticator attribute

### LINKED ISSUE #8860: [AA] Add an `#[authenticator]` attribute to the language
We have a prototype of [adding a new attribute](https://github.com/iotaledger/iota/pull/7863) to the language and a verify_authenticate_func() [iota-verifier’s method](https://github.com/iotaledger/iota/pull/8477/files#diff-167101669fa8edb30627c6789e3947a447e9b59120e034e10762eef7a9e288d8) that checks if a function could be defined as authenticate().

A new attribute can be added such that the verifier checks the annotated function using verify_authenticate_func(). E.g.:

```move
#[authenticator]
public fun authenticate_ed25519(
    self: &IOTAccount,
    signature: vector<u8>,
    _: &AuthContext,
    ctx: &TxContext,
) { ... }
```

TheMrAI:
# Description of change

To support defining authenticator functions for the IOTA abstract account workflow the user must be able to
specify in Move code which functions should be considered "authenticators".
Moreover in the future there may be more than one version of authenticators so a version number must be supported as well.

This PR provides three formats to define a move authenticator with version.
As a Named attribute:
```
#[authenticator]
public fun authenticator_fun(...)
```
In this case the version number is inferred to be 1.

As an Assigned attribute:
```
#[authenticator = 1]
public fun authenticator_fun(...)
```
Where the version number is assigned explicitly.

And as a Parametrized attribute:
```
#[authenticator(version = 1)]
public fun authenticator_fun(...)
```
In this format the user must explicitly spell out the version number and assign a value to it. While the format is verbose, it may become a preferred variant as it could potentially be extended with additional features like:
```
#[authenticator(version = 1,  something_else = 53)]
```

# Implementation notes
Attributes do NOT reach the bytecode. Instead these are stored into the `FunctionInfo` and this passed upward. Later PRs will deal with handling the Authenticator attribute from within the `FunctionInfo` (in this PR, just `iota_types::move_package::FnInfo` was extended to support the authenticator version).

In commit [4fd1ed4](https://github.com/iotaledger/iota/pull/8951/commits/4fd1ed462d97212c81006428384aeb9285d57bec) the validation of the attribute was moved into the `iota_mode` typing. 

fixes #8860


external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -538,6 +540,7 @@ fn constant(
 
 fn functions(
     context: &mut Context,
+    reporter: &DiagnosticReporter,
```
miker83z: I guess we have 2 options here: 1. passing down `reporter` or 2. propagating up the error (changing the return of `functions` and `function` to `Result`). Have you thought about this? Is there a preferred way? 

Just asking, not suggesting, both solutions seems equivalent at first glance.

external-crates/move/crates/move-compiler/tests/move_2024/to_bytecode/authenticator_attribute/assigned/version_literals.move
miker83z, line 1: Not sure if `move-compiler/tests/move_2024` is the appropriate path because it is not really a feature of Move 2024, is there an alternative? 

Related to this, do we need to modify other feature gates to enable this new attribute safely?

external-crates/move/crates/move-ir-types/src/ast.rs
```diff
@@ -387,6 +387,7 @@ pub struct Function_ {
     pub visibility: FunctionVisibility,
     /// Is entry function
     pub is_entry: bool,
+    pub authenticator_version: Option<u8>,
```
miker83z:
```suggestion
    /// Is authenticator and, if so, which version uses
    pub authenticator_version: Option<u8>,
```

external-crates/move/crates/move-binary-format/src/deserializer.rs
```diff
@@ -1556,10 +1556,14 @@ fn load_function_def(cursor: &mut VersionedCursor) -> BinaryLoaderResult<Functio
         return Err(PartialVMError::new(StatusCode::INVALID_FLAG_BITS));
     }
 
+    // TODO fix this
+    let authenticator_version = None;
```
miker83z: Is it tricky or is it just a vector (empty if None, filled with 1 u8 otherwise)?

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
TheMrAI: Given that adding a diagnostic message triggers a build failure in the end, it seemed cleaner to pass that down, instead of propagating the error back up.
In the propagation case one also has to handle the error itself and how they want to represent it.

external-crates/move/crates/move-compiler/tests/move_2024/to_bytecode/authenticator_attribute/assigned/version_literals.move
TheMrAI: I am happy to move the tests to any position. Truth be told it was selected as an educated guess.
A similar path exists in reverse to confuse the enemy:  `external-crates/move/crates/move-compiler/tests/iota_mode/move_2024`, but the equivalent path will be removed from sui at some point.
We can make a separate folder. Move it to `external-crates/move/crates/move-compiler/tests/move_check/to_bytecode`, not sure why some tests are where they are.

external-crates/move/crates/move-binary-format/src/deserializer.rs
TheMrAI: The issue is not, how we store the value, but how that impacts the byte representation, how byte representations of previous version can be parsed/serialized.
For example the FunctionDef is serializer:
https://github.com/iotaledger/iota/blob/fd3deb51f256a6d3be4a8ad683ac4790a844ef46/external-crates/move/crates/move-binary-format/src/serializer.rs#L1658

external-crates/move/crates/move-compiler/tests/move_2024/to_bytecode/authenticator_attribute/assigned/version_literals.move
TheMrAI: | Related to this, do we need to modify other feature gates to enable this new attribute safely?
Probably, have to manage versioning for sure (if we put info into CompiledModule).

external-crates/move/crates/move-binary-format/src/file_format.rs
```diff
@@ -563,6 +563,11 @@ pub struct FunctionDefinition {
     pub visibility: Visibility,
     /// Marker if the function is intended as an entry function. That is
     pub is_entry: bool,
+    /// In case the function has been marked by "authenticator(version = x)"
+    /// attribute, this field will contain the specified version.
+    /// A value of None indicates that the function was not marked as an
+    /// authenticator.
+    pub authenticator_version: Option<u8>,
```
valeriyr: @TheMrAI @miker83z, I am leaving this comment just to discuss the idea.
What do you think if we created a struct to represent the authenticator info, for example:
```rust
pub struct AuthenticatorInfo {
    pub version: u8,
    ...
}
```
It will make it possible to conveniently extend the authenticator declaration with additional info:
```rust
// `description` can be shown in the explorer, for example.
#[authenticator(version = 2, description = b"User friendly description", ...)]
```
Probably you have ideas about what we can add at the moment?

external-crates/move/crates/move-compiler/tests/move_2024/to_bytecode/authenticator_attribute/assigned/version_literals.move
TheMrAI: Yepp, we can gate attributes by this mechanism: https://github.com/iotaledger/iota/blob/vm-lang/aa-auth/8860-add-authenticator-attribute/external-crates/move/crates/move-compiler/src/expansion/translate.rs#L1157

TheMrAI:
As discussed offline we abandoned for now trying to integrate attribute information into `CompiledModule` as that will carry the need to modify `move-binary-format` which can be dangerous.
Instead the attribute now travels through the `FunctionInfo`, `FnInfo` chains and reaches the verifier from there.

What we gain with this approach is that we can have minimal modifications to the compiler and thus avoid all potential danger. Don't have to design a method to be able to store attributes on chain, handle versions etc...

What we loose is that the solution is and will have its limitations as this additional will temporarily be stored (for now, later we mitigated with PackageMetadata).
If we trace [iota_verify_module_metered](https://github.com/iotaledger/iota/blob/vm-lang/aa-auth/8860-add-authenticator-attribute/iota-execution/latest/iota-verifier/src/verifier.rs#L17) usage, we will observer that on most paths the FnInfoMap (carrying the attribute) will not be set:
https://github.com/iotaledger/iota/blob/vm-lang/aa-auth/8860-add-authenticator-attribute/iota-execution/latest/iota-adapter/src/adapter.rs#L224
https://github.com/iotaledger/iota/blob/vm-lang/aa-auth/8860-add-authenticator-attribute/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs#L975

Only set in:
https://github.com/iotaledger/iota/blob/vm-lang/aa-auth/8860-add-authenticator-attribute/crates/iota-move-build/src/lib.rs#L363
This is a rather central call, but it is best to highlight the other cases in case it may be important.

Run quick test
`cargo run -p iota -- move build -p examples/move/time_locked`
(The quick hacks enabling this will be removed of course before the PR is finished.)

external-crates/move/crates/move-compiler/tests/move_2024/to_bytecode/authenticator_attribute/assigned/version_literals.move
TheMrAI: For the record:
So we can gate these attributes under editions.
This is the edition section from a Move.toml [package] section.
```
[package]
name = "time_locked"
edition = "2024"
```
In case we don't want to make a new edition (this is my assumption), then we can put the features under DEVELOPMENT edition in which case they will work.
```
[package]
name = "time_locked"
edition = "development"
```
Otherwise they will produce error like so ():
```
error[E13001]: feature is not supported in specified edition
   ┌─ ./sources/account.move:58:3
   │
58 │ #[authenticator]
   │   ^^^^^^^^^^^^^ 'authenticator' attributes are` under development and should not be used right now.
```
This way, users can't use them, unless they explicitly use development edition, and then we can later enable the attributes by moving them to 2024 edition, or completely removing the gate flag (probably the least intrusive option.)

But in the end we really don't need this at all. As this is a compiler side check, but since this implementation doesn't modify the bytecode of `CompiledModule` or has any other impact on the compiler there is no reason for this.
On the protocol config level we have to gate for the account abstraction feature so anyways. Otherwise the compiler parses any argument silently, whether it knows them or not, so the user facing behavior doesn't actually change even if the compiler properly supports the `authenticator` attribute.

external-crates/move/crates/move-binary-format/src/file_format.rs
TheMrAI: Yepp we can do that, but I would avoid actually implementing it right now. 
With the new approach that avoids the binary format completely we can modify this rather cheaply in the future.

miker83z:
Attributes do NOT reach the bytecode. Instead these are stored into the `FunctionInfo` and this passed upward. Later PRs will deal with handling the Authenticator attribute from within the `FunctionInfo` (in this PR, just `iota_types::move_package::FnInfo` was extended to support the authenticator version).

In commit [4fd1ed4](https://github.com/iotaledger/iota/pull/8951/commits/4fd1ed462d97212c81006428384aeb9285d57bec) the validation of the attribute was moved into the `iota_mode` typing.

miker83z:
@valeriyr @Dkwcs based on the talk we had related to this PR, we have 3 possibilities of implementation:
1. create custom `KnownAttribute`s and pass them dynamically to the `CompilationEnv`; this would be similar to Aptos implementation; I didn't go with this because it would be too much intrusive and would necessitate to change core structs.
2. have a `KnownAttribute` that lets it pass any attribute, that is, similar to `#exp(...)` but without the prefix; I didn't go with this approach because this would mean any attribute "would be known", so it won't really make sense to distinguish between unknown and known; everything will be accepted as attribute and it could be an issue for possible conflicts with future known attributes.
3. what I implemented in [fdc0d97](https://github.com/iotaledger/iota/pull/8951/commits/fdc0d97fb935b446cb34120e86f4f3f3f92a224b); I added a generic `FlavoredAttribute` which can then allow us to define new attributes within the `iota_mode` compiler part.

Dkwcs:
> @valeriyr @Dkwcs based on the talk we had related to this PR, we have 3 possibilities of implementation:
> 
> 1. create custom `KnownAttribute`s and pass them dynamically to the `CompilationEnv`; this would be similar to Aptos implementation; I didn't go with this because it would be too much intrusive and would necessitate to change core structs.
> 2. have a `KnownAttribute` that lets it pass any attribute, that is, similar to `#exp(...)` but without the prefix; I didn't go with this approach because this would mean any attribute "would be known", so it won't really make sense to distinguish between unknown and known; everything will be accepted as attribute and it could be an issue for possible conflicts with future known attributes.
> 3. what I implemented in [fdc0d97](https://github.com/iotaledger/iota/pull/8951/commits/fdc0d97fb935b446cb34120e86f4f3f3f92a224b); I added a generic `FlavoredAttribute` which can then allow us to define new attributes within the `iota_mode` compiler part.

As I see it, option 1 is the best choice for long-term support. However, as you described, it requires a significant amount of effort to implement.
Option 2 seems less suitable than the others - to be honest, I didn’t fully understand it.
Option 3 appears to be a good compromise between a full-fledged implementation (as in option 1) and our more adaptive use case.

external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
```diff
@@ -0,0 +1,159 @@
...
+///
+/// These variants not necessarily specify a single attribute
+/// , but a whole class of them like [KnownAttribute::Testing].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum KnownAttribute {
+    Authenticator(AuthenticatorAttribute),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AuthenticatorAttribute;
```
Dkwcs, line 21: I think it would be better to move `AuthenticatorAttribute` and other future attribute definitions into separate modules.
It could look something like this:
**known_attributes.rs**
**authenticator_attribute.rs**
**another_attribute.rs** 
etc

Or even create a dedicated module 
**known_attributes folder**
-    **mod.rs**
-    **authenticator_attribute.rs**
-    **another_attribute.rs**

external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
```diff
@@ -306,6 +306,11 @@ fn function(context: &mut Context, name: FunctionName, fdef: &T::Function) {
     if let Some(entry_loc) = entry {
         entry_signature(context, *entry_loc, name, signature);
     }
+    if let Some(sp!(authenticator_loc, authenticator_value)) =
+        attributes.get_(&iota_known_attributes::AuthenticatorAttribute.into())
+    {
+        authenticator_attribute(context, authenticator_loc, authenticator_value);
```
Dkwcs, line 312: Just some thoughts to consider - I assume we can (or should) validate both the uniqueness and, potentially, the signature of the authenticator function.
For example, it would be useful to enforce some basic checks such as:

- A function can have only one authenticator attribute.
- An authenticator function must include `auth_context`, “ctx” etc as input parameters.

I also recall that we discussed validating certain modifiers like **entry**, **view**, **test**, etc.

external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
miker83z: [80803e8](https://github.com/iotaledger/iota/pull/8951/commits/80803e86e6cc0e2ea448818f65674f7945f479c0)

miker83z:
Split in two so that's easier to cherry-pick commits needed by the view attribute: #9187 #9188

PR #8956, opened 2025-10-21
feat(wallet): Group ledger accounts by the device main public key

marc2332:
I was experimenting with this by creating yet another account source, but it doesnt make sense, so I redid it like this.


KeitoTadashi (approved):

msarcev (approved):
👌🏻

PR #8959, opened 2025-10-21
feat(examples): Add "Spending limit account" account abstraction example

### LINKED ISSUE #8651: Implement "Account With Spending Limit" using account abstraction
Example prototype: https://iotafoundation.atlassian.net/wiki/spaces/LSC/pages/2479620404/Account+With+Spending+Limit

theiari:
# Description of change

Implement "Account with Spending limit" example for abstract accounts.

fixes #8651


examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,62 @@
...
+    let spending_limit: &u64 = borrow(account_id);
+    assert!(amount <= *spending_limit, EOverspend);
+}
+
+public fun attach(account_id: &mut UID, amount: u64) {
+    assert!(!has(account_id), ESpendingLimitAlreadyAttached);
+    dynamic_field::add(account_id, SpendLimit {}, amount)
+}
+
+// === View Functions ===
```
TheMrAI, line 70: Now I am not entirely sure which direction we are going in, but if we want this to be in the utilities and accounts
direction, then don't forget to supply **detach** and **rotate** functions as well.

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,62 @@
...
+
+public fun has(account_id: &UID): bool {
+    dynamic_field::exists_(account_id, SpendLimit {})
+}
+
+public fun borrow(account_id: &UID): &u64 {
+    dynamic_field::borrow(account_id, SpendLimit {})
+}
+
+public fun borrow_mut(account_id: &mut UID): &mut u64 {
```
TheMrAI: I think this is a public function not a view one, but I get confused on this from time to time.

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,62 @@
...
+const ESpendingLimitAlreadyAttached: vector<u8> = b"Spending limit already attached.";
+
+#[error(code = 2)]
+const ESpendingLimitMissing: vector<u8> = b"Spending limit is missing.";
+
+// === Constants ===
+
+// === Structs ===
+
+public struct SpendLimit has copy, drop, store {}
```
TheMrAI: The pattern I tried to follow was to name the dynamic field and the containing module the same way.
spending_limit, SpendLimit, seems to be in minor inconsistency. Maybe it should be "SpendingLimit"?

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,93 @@
...
+        &mut id,
+        limit,
+    );
+    let spend_limit_account = SpendLimit { id };
+    iota::transfer::share_object(spend_limit_account);
+}
+
+public fun authenticate(
+    account: &SpendLimit,
+    amount: u64,
```
TheMrAI: Can you help me understand how this account type needs to work?
It looks somewhat odd that this amount is passed in as a value. The account can't really provide any protection in this regard right?
I have some vague memory that this is why we have the AuthContext, TxCommands fields, and vaguely remember you raising that there are some technical issues in calculating the gas I believe.

TheMrAI (commented):
Haven't checked the tests yet, just skimmed the rest.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,93 @@
...
+public fun public_key(account: &SpendLimit): &vector<u8> {
+    owner_public_key::borrow(&account.id)
+}
+
+// Get the authenticator info.
+public fun authenticator_info(account: &SpendLimit): &AuthenticatorInfoV1 {
+    account::borrow_auth_info_v1(&account.id)
+}
+
+// === Admin Functions ===
```
Dkwcs, line 198: I guess all of these empty comment sections should be cleaned up?
In account.move as well

examples/move/spending_limit/sources/account.move
Dkwcs: Good point from Davo. It’s not clear for me how it can be used for real authentication if we trust only the passed value? I assume we need some logic behind the arguments, like Davo mentioned, using Context etc., to work with a gas.

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,62 @@
...
+// === Public Functions ===
+
+public fun authenticate_with_amount(account_id: &UID, amount: u64) {
+    assert!(has(account_id), ESpendingLimitMissing);
+
+    let spending_limit: &u64 = borrow(account_id);
+    assert!(amount <= *spending_limit, EOverspend);
+}
+
+public fun attach(account_id: &mut UID, amount: u64) {
```
Dkwcs, line 44: In this case it’s a minor thing, but for the full-fledged example it would be nice to have an appropriate functions to change the limit(e.g. Update())

examples/move/spending_limit/sources/account.move
TheMrAI: Not necessarily. I started using that pattern to avoid issues in the future where we violate the [code conventions](https://docs.sui.io/concepts/sui-move-concepts/conventions#body).
Others follow different patterns, but this is probably the cleanest as the section names should always be the same and always in the same order.

examples/move/spending_limit/sources/account.move
theiari: Yeah, that's a good point. I guess a solution _might_ be to rely on `vector<Coin<T>>` type instead of u64.
I added a possible implementation  here : https://github.com/iotaledger/iota/pull/8959/commits/7ae37161f6921b49cd1db54be7b99b6d1b3aa51c .

Not sure if that's enough, maybe there must be some extra logic behind `AuthContext`

examples/move/spending_limit/sources/account.move
theiari: Just followed the latest flow 😄

examples/move/spending_limit/sources/spending_limit.move
theiari: Makes sense. I changed it to `SpendingLimit` in https://github.com/iotaledger/iota/pull/8959/commits/7ae37161f6921b49cd1db54be7b99b6d1b3aa51c.

examples/move/spending_limit/sources/spending_limit.move
theiari: Yep, you're right. Added both `detach` and `rotate` functions here: https://github.com/iotaledger/iota/pull/8959/commits/7ae37161f6921b49cd1db54be7b99b6d1b3aa51c.

examples/move/spending_limit/sources/spending_limit.move
theiari: Same as above, added both `detach` and `rotate` in https://github.com/iotaledger/iota/pull/8959/commits/7ae37161f6921b49cd1db54be7b99b6d1b3aa51c.

examples/move/spending_limit/sources/spending_limit.move
theiari: Moved to public function section here: https://github.com/iotaledger/iota/pull/8959/commits/7ae37161f6921b49cd1db54be7b99b6d1b3aa51c.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,112 @@
...
+    owner_public_key::borrow(&account.id)
+}
+
+// Get the authenticator info.
+public fun authenticator_info(account: &SpendLimit): &AuthenticatorInfoV1 {
+    account::borrow_auth_info_v1(&account.id)
+}
+
+// Calculate the sum of coin values.
+public fun calculate_coin_sum<T>(coins: &vector<Coin<T>>): u64 {
```
theiari: Not entirely sure this is ok and if it should be kept public as well

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,112 @@
...
+        &mut id,
+        limit,
+    );
+    let spend_limit_account = SpendLimit { id };
+    iota::transfer::share_object(spend_limit_account);
+}
+
+public fun authenticate<T>(
+    account: &SpendLimit,
+    coins: &vector<Coin<T>>,
```
lzpap: I don't get how this authenticate function prevents me from spending from the account. Some coins are passed here as auth inputs, but who to say that there are no more coins as PTB inputs that the authenticated transactions can spend?

On another note, spending limit without a time frame is rather useless, no?  If I can spend $100 at a time, but I can repeat it as many times as I can then it's pretty easy to overspend :trollface:

If the intention is to have an account where you can only spend $X/period, it's pretty tought to implement without write access in the auth function.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    iota::transfer::share_object(spend_limit_account);
+}
+
+public fun authenticate(
+    account: &SpendLimit,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+) {
+    iotaccount::ensure_tx_sender_is_account_id(&account.id, ctx);
```
miker83z: create a 
```move
fun ensure_tx_sender_is_account(self: &SpendLimit, ctx: &TxContext)
```
and use this instead of the iotaccount one.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    function_name,
+    module_name,
+    ProgrammableMoveCall,
+    argument_input,
+    is_object_data,
+    is_pure_data,
+    is_shared_object,
+    shared_object_data
+};
+use iotaccount::iotaccount;
```
miker83z: This example doesn't need to make use of the iotaccount, it can always use the `SpendLimit` account type in the functions.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+
+    assert!(has_right_package_id_and_withdraw_call(auth_ctx, ctx), EUnauthorizedWithdrawCall);
+
+    let actual_amount = extract_withdraw_amount(auth_ctx);
+
+    spending_limit::authenticate_with_amount(&account.id, actual_amount);
+}
+
+public fun withdraw_from_balance_reserve(
+    self: &mut iotaccount::IOTAccount,
```
miker83z:
```suggestion
    self: &mut SpendLimit,
```

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    amount: u64,
+    ctx: &mut TxContext,
+): Coin<iota::iota::IOTA> {
```
miker83z:
```move
use iota::iota::IOTA;
```
It is better looking to import it above and then only use `IOTA` here.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+
+    let reserve: &mut BalanceReserve<iota::iota::IOTA> = self.borrow_field_mut(
+        BalanceReserveKey {},
+        ctx,
+    );
+
+    assert!(balance::value(&reserve.balance) >= amount, EInsufficientBalanceReserve);
+
+    let withdrawn_balance = balance::split(&mut reserve.balance, amount);
+    coin::from_balance(withdrawn_balance, ctx)
```
miker83z, line 125: As discussed before, add a line before this one that reduces the spending limit by `withdrawn_balance`.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    assert!(is_pure_data(call_arg), EInvalidArgumentType);
+
+    let pure_bytes = call_arg.pure_data();
+    let mut bcs_reader = bcs::new(*pure_bytes);
+    return bcs::peel_u64(&mut bcs_reader)
+}
+
+public fun has_right_package_id_and_withdraw_call(
+    auth_ctx: &AuthContext,
+    ctx: &iota::tx_context::TxContext,
```
miker83z:
```suggestion
    ctx: &TxContext,
```
TxContext is always implicitly imported.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+): bool {
+    let commands = tx_commands(auth_ctx);
+    let hit = commands.find_index!(|cmd| {
+        if (command_to_int(cmd) != 0) {
```
miker83z: We don't need to call `withdraw_from_balance_reserve` necessarily as the first command, it can be in any position.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    let mut bcs_reader = bcs::new(*pure_bytes);
+    return bcs::peel_u64(&mut bcs_reader)
+}
+
+public fun has_right_package_id_and_withdraw_call(
+    auth_ctx: &AuthContext,
+    ctx: &iota::tx_context::TxContext,
+): bool {
+    let commands = tx_commands(auth_ctx);
+    let hit = commands.find_index!(|cmd| {
```
miker83z: I suggested `find_index`, but now I think the best would be to cycle over all the commands because there could be more than 1 occurrence of `withdraw_from_balance_reserve`. In that case, for each `withdraw_from_balance_reserve` command it must sum the input value `amount` into a `tot_amount`. Then, `tot_amount` must be compared to the spending limit.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+        if (command_to_int(cmd) != 0) {
+            return false
+        };
+
+        let call = move_call_data(cmd);
```
miker83z: This makes the TX abort. It should be:
```move
let call = match (command) {
    Command::MoveCall(c) => c,
    _ => continue,
}
```

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+    account: &SpendLimit,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+) {
+    iotaccount::ensure_tx_sender_is_account_id(&account.id, ctx);
+
+    owner_public_key::authenticate_ed25519(&account.id, signature, ctx.digest());
+
+    assert!(has_right_package_id_and_withdraw_call(auth_ctx, ctx), EUnauthorizedWithdrawCall);
```
miker83z: The correct package id can be taken here and passed as parameter:

```suggestion
    let package_id = iota::address::from_ascii_bytes(std::type_name::get<SpendLimit>()
        .get_address()
        .as_bytes());
    assert!(
        has_right_package_id_and_withdraw_call(package_id, auth_ctx, ctx),
        EUnauthorizedWithdrawCall,
    );
```

miker83z: I'm just not 100% sure about the ascii address conversion, you need to test this.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,280 @@
...
+
+    let actual_amount = extract_withdraw_amount(auth_ctx);
```
miker83z: Given this comment https://github.com/iotaledger/iota/pull/8959/files#r2481383617
`extract_withdraw_amount` shouldn't be a dedicated function, but the amount should be already extracted in `has_right_package_id_and_withdraw_call` (I'd also change the name of that function then)

examples/move/spending_limit/sources/account.move
theiari: Latest 00db9c6fe951381b980123face2b1130bf9b2e21 keeps on following the old approach, so I still need to change this.

examples/move/spending_limit/sources/account.move
theiari: Should be fixed in 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
theiari: I created a more advanced logic that checks and sum every occurrence of `withdraw_from_balance_reserve` call: 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
theiari: Fixed in 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
theiari: Fixed in 00db9c6fe951381b980123face2b1130bf9b2e21.

examples/move/spending_limit/sources/account.move
theiari: Fixed in 00db9c6fe951381b980123face2b1130bf9b2e21.

examples/move/spending_limit/sources/account.move
theiari: Changed in 00db9c6fe951381b980123face2b1130bf9b2e21.

But maybe I got it wrong, and I need to place in the `validate_and_calculate_withdrawals` function body.

examples/move/spending_limit/sources/account.move
theiari: I slightly followed a different approach, ehere the packageID is checked inside `validate_and_calculate_withdrawals` function: 00db9c6fe951381b980123face2b1130bf9b2e21.

I've also added a few tests.

examples/move/spending_limit/sources/account.move
theiari: Created here: 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
theiari: Removed iotaccount dependency, used SpedLimit account type instead : 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
miker83z: Is it? 😄

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+
+public fun authenticate(
+    account: &SpendLimit,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+): WithdrawProof {
+    owner_public_key::authenticate_ed25519(&account.id, signature, ctx.digest());
+
+    let total_amount = validate_and_calculate_withdrawals(auth_ctx, ctx);
```
miker83z: This check looks good now!

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+    assert!(total_amount > 0, EUnauthorizedWithdrawCall);
```
miker83z: There can be use cases in which you use the account but with no withdrawals? If yes then this line can be removed

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+
+    spending_limit::authenticate_with_amount(&account.id, total_amount);
+
+    WithdrawProof {
```
miker83z: `WithdrawProof` is not needed.

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+        account_id: object::id(account),
+        amount: total_amount,
+    }
+}
+
+public fun withdraw_from_balance_reserve(
+    self: &mut SpendLimit,
+    proof: WithdrawProof,
```
miker83z: Having seen the rest of the code, I think `WithdrawProof` is not needed. Just revert to amount u64 and it will be fine!

examples/move/spending_limit/sources/account.move
theiari: I think we can remove this line. 
Removed in 8190e836e99b787351e2ad50364e0dc131273aed

examples/move/spending_limit/sources/account.move
theiari: Removed in 8190e836e99b787351e2ad50364e0dc131273aed

examples/move/spending_limit/sources/account.move
theiari: Removed `WithdrawProof` and reverted to `u64` amount here: 8190e836e99b787351e2ad50364e0dc131273aed

examples/move/spending_limit/sources/account.move
theiari: Should be fixed in 8190e836e99b787351e2ad50364e0dc131273aed 😖

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+    };
+
+    false
+}
+
+public fun ensure_tx_sender_is_account(self: &SpendLimit, ctx: &TxContext) {
+    assert!(self.id.uid_to_address() == ctx.sender(), ETransactionSenderIsNotTheAccount);
+}
+
+public fun deposit_to_reserve(self: &mut SpendLimit, coin: Coin<IOTA>, ctx: &TxContext) {
```
theiari: I still have a small doubt on this function `deposit_to_reserve`.

Wouldn't be better to accept deposits from any address?

Right now with `borrow_field_mut`, we are carefully checking the sender address, but what if the spending limit account just receives tokens from anyone?

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,313 @@
...
+    let spend_limit_account = SpendLimit { id };
+    iota::transfer::share_object(spend_limit_account);
+}
+
+public fun authenticate(
+    account: &SpendLimit,
+    signature: vector<u8>,
+    auth_ctx: &AuthContext,
+    ctx: &TxContext,
+): u64 {
```
miker83z: authenticate functions must have no return type

examples/move/spending_limit/sources/account.move
miker83z: Yes, we can accept deposit from any address in this case. 
Just use `let reserve = dynamic_field::borrow_mut(&mut self.id, BalanceReserveKey {})` here.

theiari: Relaxed the deposit address check here: 00db9c6fe951381b980123face2b1130bf9b2e21

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+        &mut spend_limit_account.id,
+        limit,
+    );
+
+    // Finalize account creation.
+    account::create_account_v1(spend_limit_account, authenticator);
+}
+
+public fun authenticate(
+    account: &mut SpendLimit,
```
theiari: Not entirely sure here...

So for the update of the spendable amount we need this to be mutable right?

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Returns a mutable reference to the spending limit for the given account ID.
+public fun borrow_mut(account_id: &mut UID): &mut u64 {
+    dynamic_field::borrow_mut(account_id, SpendingLimit {})
+}
+
+// === View Functions ===
+
+public fun has(account_id: &UID): bool {
```
miker83z:
```suggestion
public fun exists(account_id: &UID): bool {
```

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,332 @@
...
+    // Attach spending limit.
+    spending_limit::attach(
+        &mut spend_limit_account.id,
+        limit,
+    );
+
+    // Finalize account creation.
+    account::create_account_v1(spend_limit_account, authenticator);
+}
+
```
miker83z, line 104:
```suggestion

#[authenticator]
```

examples/move/spending_limit/sources/account.move
miker83z: Right, this must be `account: &SpendLimit`. Check comment in `authenticate_with_amount`

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,87 @@
...
+public struct SpendingLimit has copy, drop, store {}
+
+// === Events ===
+
+// === Method Aliases ===
+
+// === Public Functions ===
+
+// Authenticates that the given amount is within the spending limit.
+public fun authenticate_with_amount(account_id: &mut UID, amount: u64) {
```
miker83z:
```suggestion
public fun authenticate_with_amount(account_id: &UID, amount: u64) {
```

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,87 @@
...
+    assert!(has(account_id), ESpendingLimitMissing);
+
+    let spending_limit = borrow_mut(account_id);
+    assert!(amount <= *spending_limit, EOverspend);
+
+    *spending_limit = *spending_limit - amount;
```
miker83z: move this subtraction into `withdraw_from_balance_reserve`

examples/move/spending_limit/sources/spending_limit.move
miker83z: then https://github.com/iotaledger/iota/pull/8959/changes#r2667806220

examples/move/spending_limit/sources/account.move
theiari: Added in 088a7bc32e110c4ecad3573b9cd320c18e35f214

examples/move/spending_limit/sources/account.move
theiari: Ok, switched back to `account: &SpendLimit` here: 088a7bc32e110c4ecad3573b9cd320c18e35f214

examples/move/spending_limit/sources/spending_limit.move
theiari: Should be fixed here: 088a7bc32e110c4ecad3573b9cd320c18e35f214

examples/move/spending_limit/sources/spending_limit.move
theiari: refactored here: 088a7bc32e110c4ecad3573b9cd320c18e35f214

examples/move/spending_limit/sources/spending_limit.move
theiari: Moved the subtraction operation to `withdraw_from_balance_reserve` with  088a7bc32e110c4ecad3573b9cd320c18e35f214

miker83z (approved):

examples/move/spending_limit/sources/account.move
```diff
@@ -0,0 +1,358 @@
...
+    while (i < len) {
+        let cmd = &commands[i];
+
+        let call_opt = as_move_call(cmd);
+
+        if (option::is_some(&call_opt)) {
+            let call = option::borrow(&call_opt);
+
+            if (is_valid_withdraw_call(call, auth_ctx, ctx)) {
```
Dkwcs: It’s a minor thing, but shouldn’t this call be aborted(something like `EInvalidWithdrawCall`) if withdraw call is invalid? In current case we’re just skipping it

examples/move/spending_limit/sources/spending_limit.move
```diff
@@ -0,0 +1,86 @@
...
+public fun rotate(account_id: &mut UID, amount: u64): u64 {
+    assert!(exists(account_id), ESpendingLimitMissing);
+    assert!(amount > 0, EInvalidLimit);
+    let prev_limit = dynamic_field::remove(account_id, SpendingLimit {});
+    dynamic_field::add(account_id, SpendingLimit {}, amount);
+    prev_limit
+}
+
+// Returns a mutable reference to the spending limit for the given account ID.
+public fun borrow_mut(account_id: &mut UID): &mut u64 {
```
Dkwcs: Attach and rotate reject 0 as amount, but we still can set it as 0 via `borrow_mut`
Probably we can avoid this by making `borrow_mut` public(package)?

Dkwcs (approved):
LGTM, left few comments

examples/move/spending_limit/sources/account.move
theiari: Maybe the _invalid_ name is misleading here.

The function `is_valid_withdraw_call` does only check if the i-th command is a withdrawal call.

examples/move/spending_limit/sources/spending_limit.move
theiari: Yes I think we can set `borrow_mut` to be public(package) instead

theiari: Changed here to `public(package)` fd9c7acc593280bf1ab326e8a6f9671043643352

examples/move/spending_limit/sources/account.move
theiari: Refactor _valid_ into _lookup_ here fd9c7acc593280bf1ab326e8a6f9671043643352

PR #8969, opened 2025-10-22
feat: support for move-view function calls (alpha version)

kodemartin:
# Description of change

Feature branch enabling support for move-view function calls (alpha version).

### Release Notes

- [x] GraphQL: Introducing a new `moveViewCall` query path to call move-view functions. :rotating_light:  The server now requires that the fullnode JSON-RPC url is specified in the configuration, to be able to run `mutation.executeTransactionBlock`, `query.{moveViewCall,dryRunTransactionBlock}`. :warning:  Usage of the `Query.moveViewCall` are only supported on `devnet`, as the feature is still in alpha.
- [x] JSON-RPC: Adds support for move view function calls through a new `iota_view` RPC method. :warning: Calls to the `iota_view` endpoint are only supported on  `devnet` , as the feature is still in alpha.
- [x] Rust SDK: Added move_view_call_tx_kind and single_move_view_call to the TransactionBuilder


sergiupopescu199 (approved):
Changes to files owned by the infra team lgtm! 🌸

marc2332 (approved):

miker83z (approved):

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -86,7 +88,6 @@ impl TransactionBuilder {
     async fn get_object_arg(
         &self,
         id: ObjectID,
-        objects: &mut BTreeMap<ObjectID, Object>,
```
DaughterOfMars, line 89: Was this just unused?

kodemartin: exactly!

DaughterOfMars (approved):

crates/iota-transaction-builder/src/utils.rs
DaughterOfMars: good find then :)

kodemartin: credit goes to @miker83z

PR #8926, opened 2025-10-16
feat(grpc): move proto files to `iota-grpc-types`

muXxer:
# Description of change

This PR moves the gRPC proto files to the `iota-grpc-types` crate.
This is cleaner than mixing the proto files with the server+client code.
It also adds versioning to the proto files.

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #8971, opened 2025-10-22
fix(core): remove dependencies on `iota-grpc-api`

muXxer:
# Description of change

The current implementation of the `iota-grpc-api` introduces a lot of additional dependencies in several crates, even if they are not even necessary. We tried to fix those in #8926 but this was not sufficient.

This PR removes those dependencies by splitting the client and server code, and by reversing a dependency between `iota-core` and `iota-grpc-api`, which was needed to implement a trait. Since `iota-grpc-api` is only used by `iota-node`, and that imports `iota-core` already, it makes more sense to make `iota-grpc-api` dependant on `iota-core` instead of the other way around.

`iota-grpc-api` is renamed to `iota-grpc-server` and the client code was moved to `iota-grpc-client`, so that users don't inherit all the dependencies of the server code.


bingyanglin (approved):
Thank you!

alexsporn (approved):

thibault-martinez (approved):

kodemartin (approved):

PR #8979, opened 2025-10-22
feat(iota-indexer-streaming): add prometheus metrics

### LINKED ISSUE #8711: [iota-indexer]: add prometheus metrics for the `IndexerStreamer`
Currently the `IndexerStreamer` does provide logs but for monitoring on production it would be convenient to provide also prometheus metrics for better monitoring of the streams performance.

possible metrics:
- total active subscribers per stream
- total lagging subscribers per stream
- current checkpoint sequence number sent to the stream, handy for catching if the `IndexerStreamer` is lagging behind the Indexer commit rate

sergiupopescu199:
# Description of change

This PR adds Prometheus metrics to the `iota-indexer-streaming`. It allows to monitor active subscribers, latency of internal process and to have a clear picture how it behaves under load.

fixes #8711 

- Used the `iota` cli, to quickly spin up a local network with a indexer & graphql server, since the latter uses the `iota-indexer-streming` we were able to see the metrics since it already exposes a Prometheus server.

> [!NOTE]
> This patch does not affect the indexer normal operations, thus no further tests were conducted. 

### Release Notes
- [x] Indexer: `iota-indexer-streaming` add Prometheus metrics.

### DIFF
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -28,6 +28,7 @@ use iota_indexer::{
     },
 };
 use iota_types::event::Event;
+use prometheus::IntGauge;
 use serde::Deserialize;
 use tokio::sync::broadcast;
 use tokio_postgres::{
@@ -486,10 +487,10 @@ impl InMemory {
 struct SubscriberStream<T> {
     /// The inner stream implementation we want to wrap.
     inner: BroadcastStream<T>,
-    /// The prometheus label we want to record metrics.
-    label: &'static str,
-    /// The collection of available prometheus metrics we want to record.
-    metrics: Arc<InMemoryStreamMetrics>,
+    /// Tracks if the subscriber is active.
+    active_subscriber_number: IntGauge,
+    /// Tracks if the subscriber is lagging.
+    lagging_subscribers: IntGauge,
     /// Tracks if this subscriber is currently lagged.
     is_lagging: bool,
 }
@@ -500,15 +501,15 @@ impl<T> SubscriberStream<T> {
         label: &'static str,
         metrics: Arc<InMemoryStreamMetrics>,
     ) -> Self {
-        metrics
-            .active_subscriber_number
-            .with_label_values(&[label])
-            .inc();
+        let active_subscriber_number = metrics.active_subscriber_number.with_label_values(&[label]);
+        let lagging_subscribers = metrics.lagging_subscribers.with_label_values(&[label]);
+
+        active_subscriber_number.inc();
 
         Self {
             inner,
-            label,
-            metrics,
+            active_subscriber_number,
+            lagging_subscribers,
             is_lagging: false,
         }
     }
@@ -516,10 +517,7 @@ impl<T> SubscriberStream<T> {
 
 impl<T> Drop for SubscriberStream<T> {
     fn drop(&mut self) {
-        self.metrics
-            .active_subscriber_number
-            .with_label_values(&[self.label])
-            .dec();
+        self.active_subscriber_number.dec();
     }
 }
 
@@ -529,27 +527,18 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let poll = Pin::new(&mut self.inner).poll_next(cx);
 
-        match poll {
-            Poll::Ready(Some(Ok(_))) => {
-                if self.is_lagging {
-                    self.metrics
-                        .lagging_subscribers
-                        .with_label_values(&[self.label])
-                        .dec();
-                    self.is_lagging = false;
-                }
+        if let Poll::Ready(Some(Ok(_))) = poll {
+            if self.is_lagging {
+                self.lagging_subscribers.dec();
+                self.is_lagging = false;
             }
-            Poll::Ready(Some(Err(BroadcastStreamRecvError::Lagged(_)))) => {
-                if !self.is_lagging {
-                    self.metrics
-                        .lagging_subscribers
-                        .with_label_values(&[self.label])
-                        .inc();
-                    self.is_lagging = true;
-                }
+        } else if let Poll::Ready(Some(Err(BroadcastStreamRecvError::Lagged(_)))) = poll {
+            if !self.is_lagging {
+                self.lagging_subscribers.inc();
+                self.is_lagging = true;
             }
-            _ => {}
         }
+
         poll
     }
 }
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -502,6 +502,10 @@ struct SubscriberStream<T> {
 }
 
 impl<T> SubscriberStream<T> {
+    /// Represents the duration of the lag state of the subscriber, mostly to
+    /// help Prometheus scrape the metric.
+    const LAG_STATE: Duration = Duration::from_secs(1);
+
     pub fn new(
         inner: BroadcastStream<T>,
         label: &'static str,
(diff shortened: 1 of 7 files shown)


crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -409,3 +473,83 @@ impl InMemory {
...
+pub struct SubscriberStream<T> {
+    /// The inner stream implementation we want to wrap.
+    inner: BroadcastStream<T>,
+    /// The prometheus label we want to record metrics.
+    label: &'static str,
+    /// The collection of available prometheus metrics we want to record.
+    metrics: Arc<InMemoryStreamMetrics>,
+    /// Tracks if this subscriber is currently lagged.
+    is_lagging: bool,
+}
```
kodemartin: Could you elaborate on the reasons we need this new wrapper? How is it supposed to be used by dependent libraries/apps?

sergiupopescu199: This needs to be a private struct.

The `SubscriberStream` struct is designed as a private wrapper around `BroadcastStream<T>` to encapsulate metrics collection and lag detection logic entirely within the `iota-indexer-streaming` crate. This ensures that dependent libraries and applications don't need to handle these concerns manually, reducing the risk of inconsistent or missing metrics.

Clients get a fully functional stream with built-in metrics, without needing to know about or access the `SubscriberStream` type directly. This abstraction keeps the API clean while ensuring robust monitoring.

fixed: https://github.com/iotaledger/iota/pull/8979/commits/a40d46c08b714bacc4e520adfb581bb79ceb3714

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -316,34 +342,54 @@ impl InMemory {
...
 
+        let elapsed = guard.stop_and_record();
+        debug!(
+            "broadcast data took: {:?}",
+            Duration::from_secs_f64(elapsed)
+        );
+
+        metrics
+            .broadcasted_to_subscribers_tx_seq_num_start_range
+            .set(start);
```
tomxey, line 387: I wonder, do we gain any info from having the range **start** logged in grafana? If we also log the **end** of the range then it tells us what is the current progress of the broadcast in general.
Or is it possible that the **end** is moving forward, but **start** is stuck?

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -316,34 +342,54 @@ impl InMemory {
...
         start: i64,
         end: i64,
         indexer_reader: &IndexerReader,
         event_tx: &broadcast::Sender<StoredEvent>,
         transaction_tx: &broadcast::Sender<StoredTransaction>,
     ) -> IndexerStreamingResult<()> {
-        let instant = Instant::now();
+        // auto-records duration on drop (function return).
+        let _timer = metrics.process_transaction_batch_latency.start_timer();
+        let guard = metrics.query_tx_from_indexer_db_latency.start_timer();
```
tomxey: Could we use some more distinct names for the timers? instead of timer and guard.

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -356,19 +402,38 @@ impl InMemory {
...
+
+        // we sacrifice per-event/transaction granularity to avoid degrading
+        // performance from frequent metric updates in a hot path.
+        metrics
+            .channel_queue_size
+            .with_label_values(&[METRICS_EVENT_LABEL])
+            .set(event_tx.len() as i64);
+
+        metrics
+            .channel_queue_size
```
tomxey: Is the amount of transactions that we are just sending equal to the queue size?
I'm thinking if the name maybe is not confusing here

crates/iota-indexer-streaming/src/metrics.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Latency of processing a batch of transactions, from query to broadcast
+    /// to subscribers.
+    pub process_transaction_batch_latency: Histogram,
+    /// Latency of processing a batch of database notifications. It includes the
+    /// tx range bounds resolution, the time taken to query the database, and
+    /// broadcast them to subscribers.
+    pub process_notification_batch_latency: Histogram,
+}
+
+impl InMemoryStreamMetrics {
```
tomxey, line 53: I have one suggestion for new metric.
We could maybe have something like "event/tx broadcast lag"?
Which would be computed as "current_timestamp - timestamp_of_event/tx" during broadcast. Or potentially at other places in the pipeline too. it would tell you how fast the whole pipeline is, and how fast data is delivered to subscribers.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: > Or is it possible that the end is moving forward, but start is stuck?

No, both metrics move together.

My primary goal is to also track the batch size, I'm not sure if in grafana we can derive it from the `end`. So I suppose having the start will not be harmful

crates/iota-indexer-streaming/src/metrics.rs
sergiupopescu199: the `notified_tx_seq_num_start_range` & `notified_tx_seq_num_end_range` shows the received transaction batch from indexer, and `broadcasted_to_subscribers_tx_seq_num_start_range` & `broadcasted_to_subscribers_tx_seq_num_end_range` shows the batch we're broadcasting to subscribers, if the last two metrics way less comapred to first ones we have a lag between received data from indexer & data sent to subscribers.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8979/commits/e6a084b6404e59dd4106bb8dae4cff125011d3aa

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/8979/commits/e6a084b6404e59dd4106bb8dae4cff125011d3aa

crates/iota-indexer-streaming/src/memory.rs
tomxey: is the batch size constant, or changing over time?

crates/iota-indexer-streaming/src/metrics.rs
tomxey: okay, that should be enough I guess

crates/iota-indexer-streaming/src/metrics.rs
```diff
@@ -0,0 +1,133 @@
...
+            .unwrap(),
+            broadcast_tx_and_ev_to_subscribers_latency: register_histogram_with_registry!(
+                "broadcast_tx_and_ev_to_subscribers_latency",
+                "Latency of broadcasting transactions and events to subscribers",
+                registry,
+            )
+            .unwrap(),
+            channel_pending_messages: register_int_gauge_vec_with_registry!(
+                "channel_queue_size",
+                "The number of messages not yet received by all active subscribers",
```
tomxey, line 114: Is the value of this metric actually equal to number of messages not yet received?
From looking at the code it looks like the number of messages that have been just sent, no matter if someone is receiving or not.
(e.g. will the value of this metric grow if receiver(s) stop receiving?)

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: Upon initialization you can change it after that is constant

crates/iota-indexer-streaming/src/metrics.rs
sergiupopescu199: Here's the [len](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Sender.html#method.len) documentation, from my understanding the metric does encapsulate its documentation.

crates/iota-indexer-streaming/src/memory.rs
tomxey: okay, let's keep both

crates/iota-indexer-streaming/src/metrics.rs
tomxey: oh, I misread the implementation.
now it makes sense, thanks

tomxey (approved):

kodemartin (approved):
lgtm

PR #8990, opened 2025-10-23
feat(graphql-rpc): Introduce Query.transactionBlocksByDigests

### LINKED ISSUE #8798: feat(graphql-rpc): Implement `Query.transactionBlocksByDigests` for optimistic transactions
In GraphQL after https://github.com/iotaledger/iota/pull/8645 we are then only missing functionality similar to `iota_multiGetTransactionBlocks`
We should investigate and come up with the optimal most ergonomic solution.
Naive approach: `Query.multiGetTransactionBlocks(digests: [Digest])`

tomxey:
# Description of change

Introduce `Query.transactionBlocksByDigests` that will fetch multiple transactions at once by digests

fixes https://github.com/iotaledger/iota/issues/8798

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL: New `Query.transactionBlocksByDigests` can now be used to fetch multiple transactions by digest. This will also include just-executed transactions that don't have a checkpoint assigned yet.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -388,6 +388,21 @@ impl Query {
...
+        &self,
+        ctx: &Context<'_>,
+        digests: Vec<Digest>,
+    ) -> Result<Vec<TransactionBlock>> {
+        let Watermark { checkpoint, .. } = *ctx.data()?;
+        let result = TransactionBlock::multi_query(ctx, digests, checkpoint)
+            .await
+            .extend()?;
+
+        // Return only found transactions, omit missing ones
```
sergiupopescu199: Is there a particular reason shy do we skip missing transactions? the same behavior for other muli get APIs in GraphQL?

I'm wondering if it's ok to omit missing tx, maybe we could instead return `Vec<Option<TransactionBlock>>`

The user will see the missing tx as `null` in this way if user provides 5 digests it can expect 5 elements in a vector and assert from it's index which digest is wrong based on returned `null` value.

crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -388,6 +388,21 @@ impl Query {
             .extend()
     }
 
+    /// Fetch multiple transaction blocks by their digests.
+    async fn multi_transaction_blocks(
+        &self,
+        ctx: &Context<'_>,
+        digests: Vec<Digest>,
+    ) -> Result<Vec<TransactionBlock>> {
+        let Watermark { checkpoint, .. } = *ctx.data()?;
```
kodemartin, line 407: Here we should a limit check similar to what `transactionBlocks` applies

https://github.com/iotaledger/iota/blob/f9da5b01fc8235ac6da35c8bbc748406c661e242/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs#L381-L388

crates/iota-graphql-rpc/src/types/query.rs
```diff
@@ -388,6 +388,21 @@ impl Query {
             .extend()
     }
 
+    /// Fetch multiple transaction blocks by their digests.
+    async fn multi_transaction_blocks(
```
kodemartin: I think that the selected name `multi_transaction_blocks` does not differ semantically from `transaction_blocks` and thus might cause confusion.

It is important to make it clear what the difference is. I believe that 

```rust
    async fn transaction_blocks_by_digests(
```
is as explicit as it gets. And moreover it complies with the convention herein that getter endpoints are named by the item(s) they get. 

The documentation then should further clarify that this returns transactions even if they are not yet indexed through a checkpoint.

kodemartin: A direct implication is that `transaction_blocks` exclusively refers to checkpointed transactions. This has to be documented now.

kodemartin (changes requested):
A few change requests.

Plus I think it still makes sense that we provide paginated response using transaction digests as the cursor. This can be delegated to a new issue.

crates/iota-graphql-rpc/src/types/query.rs
tomxey: We could do that. That's the approach upstream took: [ref](https://github.com/MystenLabs/sui/blob/cb968871471697b3b21edd95a4997a2870fa7497/crates/sui-indexer-alt-graphql/src/api/query.rs#L251)

tomxey: Done in fc7a6bb120d7650624873c7f3a8a4d968437669f

crates/iota-graphql-rpc/src/types/query.rs
tomxey: Done in ccda5098a8fb0f9fb447ac803abfc947b26c08cf

crates/iota-graphql-rpc/src/types/query.rs
tomxey: Done in ccda5098a8fb0f9fb447ac803abfc947b26c08cf

tomxey:
> A few change requests.
> 
> Plus I think it still makes sense that we provide paginated response using transaction digests as the cursor. This can be delegated to a new issue.

Created issue https://github.com/iotaledger/iota/issues/9156

sergiupopescu199 (approved):
lgtm! 🌸

kodemartin (approved):
lgtm

Don't forget to update the release notes

PR #8993, opened 2025-10-23
feat(iota-core): update neural network model server from congestion tracker module

salaheldinsoliman:
# Description of change

This PR adds update to the neural network third party server via a function `inform_model`. `inform_model` is called at the end of `process_checkpoint_effects`.
Any logic regarding the update itself is placed in `model_updater.rs`, that is not to clutter congestion_tracker with un-related logic

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/congestion_tracker.rs
+++ b/crates/iota-core/src/congestion_tracker.rs
@@ -163,8 +163,6 @@ impl CongestionTracker {
         let mut stats: HashMap<ObjectID, ObjectCheckpointStats> = HashMap::new();
         let mut raw_txs: Vec<RawTxItem> =
             Vec::with_capacity(congestion_txs_data.len() + clearing_txs_data.len());
-        let mut per_obj_min_clearing: HashMap<ObjectID, u64> = HashMap::new();
-        let mut per_obj_max_congestion: HashMap<ObjectID, u64> = HashMap::new();
 
         for (is_congested, tx) in congestion_txs_data
             .iter()
@@ -195,34 +193,14 @@ impl CongestionTracker {
                 touched_objects: tx.objects.clone(),
             });
 
-            // per-object min clearing for clearing txs
-            if !is_congested {
-                for oid in &tx.objects {
-                    per_obj_min_clearing
-                        .entry(*oid)
-                        .and_modify(|m| *m = (*m).min(tx.gas_price))
-                        .or_insert(tx.gas_price);
-                }
-            } else {
-                // Track per-object max congestion (using feedback if present, otherwise gas price)
-                let cong_req = tx.gas_price_feedback.unwrap_or(tx.gas_price);
-                for oid in &tx.objects {
-                    per_obj_max_congestion
-                        .entry(*oid)
-                        .and_modify(|m| *m = (*m).max(cong_req))
-                        .or_insert(cong_req);
-                }
-            }
         }
 
-        // Build per-object required price in checkpoint:
-        // - prefer lowest clearing if present; otherwise use highest congestion observed.
+        // Build per-object required price in checkpoint using
+        // get_suggested_gas_price_for_objects()
         let mut per_obj_required_in_cp: HashMap<ObjectID, u64> = HashMap::new();
         for oid in &touched {
-            if let Some(c) = per_obj_min_clearing.get(oid) {
-                per_obj_required_in_cp.insert(*oid, *c);
-            } else if let Some(max_cong) = per_obj_max_congestion.get(oid) {
-                per_obj_required_in_cp.insert(*oid, *max_cong);
+            if let Some(price) = self.get_suggested_gas_price_for_objects(std::iter::once(*oid)) {
+                per_obj_required_in_cp.insert(*oid, price);
             }
         }
 
--- a/crates/iota-core/src/congestion_tracker.rs
+++ b/crates/iota-core/src/congestion_tracker.rs
@@ -324,39 +324,34 @@ impl CongestionTracker {
                 continue;
             }
 
-            // Compute the set of mutated shared objects for this transaction once,
-            // and only use those for congestion/clearing accounting and model input.
-            let mutated_objects: Vec<ObjectID> = effects
-                .input_shared_objects()
-                .into_iter()
-                .filter_map(|object| match object {
-                    InputSharedObject::Mutate((id, _, _)) => Some(id),
-                    _ => None,
-                })
-                .collect();
-
-            if let Some(CongestedObjects(_congested_objects)) =
+            if let Some(CongestedObjects(congested_objects)) =
                 effects.status().get_congested_objects()
             {
-                // Only consider transactions that actually mutate shared objects.
-                if !mutated_objects.is_empty() {
-                    congestion_txs_data.push(TxData {
-                        checkpoint: checkpoint.sequence_number,
-                        digest: *effects.transaction_digest(),
-                        objects: mutated_objects.clone(),
-                        gas_price,
-                        gas_price_feedback: Some(
-                            effects
-                                .status()
-                                .get_feedback_suggested_gas_price()
-                                .unwrap_or(self.reference_gas_price),
-                        ),
-                        sui_prediction: sui_prediction,
-                        ogd_prediction: ogd_prediction,
-                        cleread: false,
-                    });
-                }
+                congestion_txs_data.push(TxData {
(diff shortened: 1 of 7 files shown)


crates/iota-core/src/congestion_tracker.rs
```diff
@@ -139,9 +143,109 @@ pub struct CongestionTracker {
...
+    /// Compose and send model update and training batches for the given checkpoint.
+    fn inform_model(
+        &self,
+        checkpoint: &VerifiedCheckpoint,
+        congestion_txs_data: &[TxData],
+        clearing_txs_data: &[TxData],
+    ) {
+        // 1) Build touched set
+        let mut touched: std::collections::HashSet<ObjectID> = std::collections::HashSet::new();
+        for tx in congestion_txs_data {
```
vekkiokonio: Here you don't differentiate between write shared objects, read shared objects and owned objects. That means that read-only or owned objects would affect the prediction, which shouldn't be the case

vekkiokonio: For instance, check the filter at line 327

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -139,9 +143,109 @@ pub struct CongestionTracker {
...
+        let mut snapshots: HashMap<ObjectID, ObjectSnapshot> = HashMap::new();
+        for oid in &touched {
+            if let Some(info) = self.get_congestion_info(*oid) {
+                snapshots.insert(
+                    *oid,
+                    ObjectSnapshot {
+                        latest_congestion_time: Some(info.latest_congestion_time),
+                        highest_congestion_gas_price: info.highest_congestion_gas_price,
+                        latest_clearing_time: info.latest_clearing_time,
+                        lowest_clearing_gas_price: info.lowest_clearing_gas_price.unwrap_or(0),
```
vekkiokonio:
```suggestion
                        lowest_clearing_gas_price: info.lowest_clearing_gas_price.unwrap_or(self.reference_gas_price),
```

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -139,9 +143,109 @@ pub struct CongestionTracker {
...
+                        hotness: info.hotness,
+                    },
+                );
+            }
+        }
+
+        // 3) Build per-checkpoint stats
```
vekkiokonio: Both 2) and 3) are "per-checkpoint stats". The difference is that 2) is retrieved directly from congestion_info, while 3) is computed in this function. Stats in 2) provide richer information as they also include gas prices. I'm wondering if stats in 3) are really needed and if they actually contributed to improve the NN prediction.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -139,9 +143,109 @@ pub struct CongestionTracker {
...
+            &stats,
+        );
+        self.model_updater.post_update(update_batch);
+
+        // 5) Build raw tx items and per-object min clearing for training
+        let mut raw_txs: Vec<RawTxItem> = Vec::with_capacity(
+            congestion_txs_data.len() + clearing_txs_data.len(),
+        );
+        let mut per_obj_min_clearing: HashMap<ObjectID, u64> = HashMap::new();
+        for tx in congestion_txs_data {
```
vekkiokonio: You introduce several for loops over the same data structures (`clearing_txs_data` and `congestion_txs_data`). I think you can combine 1), 3) and 5) into a single loop.

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -139,9 +143,109 @@ pub struct CongestionTracker {
...
+        }
+        for tx in clearing_txs_data {
+            raw_txs.push(RawTxItem {
+                tx_digest: tx.digest.to_string(),
+                is_congested: false,
+                gas_price: tx.gas_price,
+                gas_price_feedback: None,
+                touched_objects: tx.objects.clone(),
+            });
+            for oid in &tx.objects {
```
vekkiokonio: Your `per_obj_min_clearing` vector corresponds to `get_prediction_suggested_gas_price` computed for all objects in the checkpoint. However, your code doesn't handle the rare but possible case where no clearing data are present and all transactions are canceled for a given checkpoint

crates/iota-core/src/model_updater.rs
```diff
@@ -0,0 +1,480 @@
...
+    collections::HashMap,
+    fs::{File, OpenOptions},
+    io::Write,
+    path::PathBuf,
+    sync::{Arc, Mutex},
+};
+use iota_types::base_types::ObjectID;
+
+/// Default base URL for the predictor service.
+pub const DEFAULT_PREDICTOR_BASE_URL: &str = "http://predictor:9666";
```
vekkiokonio: Had to change this to `"http://0.0.0.0:9666"` to make it work

crates/iota-core/src/congestion_tracker.rs
salaheldinsoliman: Good point :+1: 
Step (2) is a post‑checkpoint snapshot from the cache (latest_* times, gas‑price extremes, hotness), 
while step (3) adds per‑checkpoint intensity: how many congested/clearing txs touched the object. The booleans in (3) are redundant (derivable from (2) by comparing latest_*_time to this checkpoint), but the counts are not and provide extra signal for the predictor. 
I think we should keep the counts for now, WDYT?

crates/iota-core/src/congestion_tracker.rs
salaheldinsoliman: Thanks, 
I replaced the separate loops for touched(only mutated), stats, and training items with one loop over `congestion_txs_data` + `clearing_txs_data`

crates/iota-core/src/congestion_tracker.rs
salaheldinsoliman: nice catch, what I did now is basically:
A second accumulator that tracks, per object, the highest congestion requirement seen in the checkpoint (using gas_price_feedback if present, otherwise the gas price,). 
After the single pass over transactions, I merge the two into a per_obj_required_in_cp map: it prefers the lowest clearing price when available, and otherwise falls back to that highest congestion value.

crates/iota-core/src/model_updater.rs
salaheldinsoliman: This has to run in the docker compose setup to work with this URL, I will add a comment for clarification

salaheldinsoliman:
Changes summary to address the PR comments:
- Consider only mutated objects, not all objects
- Refactored multiple for loops into a single pass
- Addressed the case where no clearing Txs are in a checkpoint

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: Sure, there is no harm to keep them

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: @salaheldinsoliman I think my comment was wrong here: you are using the same `congestion_txs_data` and `cleared_txs_data` from `process_checkpoint_effect()` so there is actually no need to include the mutate filter for congested objects -- in fact, we can only have congestion for write shared objects. I suggest to revert https://github.com/iotaledger/iota/pull/8993/commits/ddb1243d1d1071b25ca705766c0734eea8f54065

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: Here, I would suggest to reuse the existing congestion tracker logic. Specifically, I would suggest to directly use the `get_suggested_gas_price_for_objects()` function which already does what you want. I would remove `per_obj_min_clearing` and `per_obj_max_congestion` and use something like:

```
let price = self.get_suggested_gas_price_for_objects(std::iter::once(*oid));
```

to store data into `per_obj_required_in_cp`.

vekkiokonio (approved):
LGTM

Please address the two comments I added to the PR and then it can be merged.

salaheldinsoliman:
Hi @vekkiokonio, I addressed the remaining two comments :+1:

vekkiokonio (approved):

vekkiokonio:
Please flag this as draft since it won't be merged any time soon

PR #9045, opened 2025-10-28
Add move side of PackageMetadata implementation

TheMrAI:
# Description of change

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-framework/packages/iota-framework/sources/package_metadata.move
+++ b/crates/iota-framework/packages/iota-framework/sources/package_metadata.move
@@ -1,0 +1,86 @@
+module iota::package_metadata;
+
+use iota::function_handle::FunctionHandle;
+use std::ascii;
+
+// === Errors ===
+
+#[error(code = 0)]
+const EModuleNotFound: vector<u8> = b"The specified module is not part of the package.";
+#[error(code = 1)]
+const EFunctionNotFound: vector<u8> = b"The specified function is not part of the package.";
+
+// === Constants ===
+
+// === Structs ===
+
+public struct PackageMetadataV1 has key {
+    // PackageMetadata (derived) id
+    id: UID,
+    // Package id
+    package_id: ID,
+    // Package version
+    package_version: u64,
+    // Package name
+    package_name: ascii::String,
+    // Package modules
+    module_handles: vector<ascii::String>,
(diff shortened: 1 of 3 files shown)


crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,86 @@
...
+    module_handle: u16,
+    function_name: ascii::String,
+): &FunctionHandle {
+    let mut function_handle_index = self
+        .function_handles
+        .find_index!(|f| f.module_handle() == module_handle && f.name() == function_name);
+    assert!(function_handle_index.is_some(), EFunctionNotFound);
+    
+    self.function_handles.borrow(function_handle_index.extract())
+}
```
valeriyr, line 78: It looks semantically clearer when `find_*` functions return `Option`s instead of aborting.
Do we have a reason to abort here?

TheMrAI: All depends how we want to look at the implementation.
If you want to look at it as an immutable implementation that has to be in a valid state at all times
(given that immutability), then one can argue that these accessor functions don't need to return an Option. We can simply rely on the yet to be documented preconditions that as long as you put in a valid module_handle, function_name that you will get back the appropriate result. This way users don't have to manage error messages on the callsites either.

If you want to look at it as a general purpose data storage without any preconditions, it should return Options.

valeriyr: The name reflects the semantics, and almost always, a `find` function shouldn't crash if it can't find the result.
The function can be called `get_*`, or `borrow_*` in this case.
But aborting is a requirement for `AuthenticatorInfoV1` creation, not for these functions; these are public functions that can be called by anybody just to take the information, so the same logic should work as for a simple collection here.

valeriyr: Consider this as a proposal, the design is provided by @miker83z, so you can discuss this with him if you wish.

PR #9047, opened 2025-10-28
refactor: optimize authenticator inputs loading

### LINKED ISSUE #8849: [AA][authority] Optimize authenticator inputs loading

valeriyr:
# Description of change

The changes made in this PR are aimed at optimizing the loading of transaction inputs.

fixes #8849

### DIFF
--- a/crates/iota-types/src/transaction.rs
+++ b/crates/iota-types/src/transaction.rs
@@ -2512,6 +2512,63 @@ impl SenderSignedData {
                 Err(_) => false,
             })
     }
+
+    /// Returns all input objects including those from the sender
+    /// `MoveAuthenticator` if any.
+    pub fn collect_inputs(&self) -> IotaResult<Vec<InputObjectKind>> {
+        if let Some(move_authenticator) = self.sender_move_authenticator() {
+            let mut input_objects_set = self
+                .transaction_data()
+                .input_objects()?
+                .into_iter()
+                .collect::<HashSet<_>>();
+
+            input_objects_set.extend(move_authenticator.object_to_authenticate().input_objects());
+            input_objects_set.extend(move_authenticator.input_objects());
+
+            Ok(input_objects_set.clone().into_iter().collect::<Vec<_>>())
+        } else {
+            Ok(self.transaction_data().input_objects()?)
+        }
+    }
+
+    /// Splits the provided input objects into three categories:
+    /// 1. Input objects required by the transaction itself.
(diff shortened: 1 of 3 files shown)


miker83z (approved):
LGTM!

crates/iota-types/src/transaction.rs
```diff
@@ -2512,6 +2511,95 @@ impl SenderSignedData {
...
+                .into();
+
+            let account_objects = move_authenticator
+                .object_to_authenticate()
+                .input_objects()
+                .iter()
+                .map(|k| {
+                    input_objects_map
+                        .get(k)
+                        .cloned()
```
Dkwcs: Can we consider to avoid using double `.cloned()` and instead use something like:

```
  let account_objects = move_authenticator
      .object_to_authenticate()
      .input_objects()
      .iter()
      .map(|k| {
          // get() returns &&ObjectReadResult so clone exactly the ObjectReadResult once
          (*input_objects_map
              .get(k)
              .expect("Expected input object to be present”))
          .clone()
      })
      .collect::<Vec<_>>();
```

It’s also applicable to the code above

Dkwcs (approved):
LGTM, just left minor comment

crates/iota-types/src/transaction.rs
valeriyr: I don't know if it makes a lot of sense to get rid of double clone, but I did it in a bit different way:
f4041f2150

PR #9053, opened 2025-10-28
fix(wallet-dashboard): Allow viewing vesting schedule rewards after last payout is unlocked

### LINKED ISSUE #9048: Disclaim innacurate schedule if vesting schedule is over
For users that completed their whole schedule (2 years) the modal looks like:

Image

We should allow seeing it but add a warning info popup explaining that now that the vesting period has ended, the payouts in the schedule they are seeing might not be 100% accurate

msarcev:
Closes https://github.com/iotaledger/iota/issues/9048


VmMad (approved):

VmMad (approved):

KeitoTadashi (approved):

begonaalvarezd (changes requested):
I disagree with this PR, indeed if the last payout is unlocked we cant show the vesting schedule. If we do, we cant make sure is accurate unless we fetch the whole tx history

PR #9059, opened 2025-10-28
feat(wallet): Side Panel support

### LINKED ISSUE #9015: Side panel view for IOTA Wallet

marc2332:
Closes https://github.com/iotaledger/iota/issues/9015


apps/wallet/src/ui/app/components/menu/content/WalletSettingsMenuList.tsx
```diff
@@ -54,9 +58,10 @@ export function MenuList() {
     const networkConfig = network === Network.Custom ? getCustomNetwork() : getNetwork(network);
     const version = Browser.runtime.getManifest().version;
     const autoLockInterval = useAutoLockMinutes();
-    const isAppPopup = useAppSelector(
-        (state) => state.app.extensionViewType === ExtensionViewType.Popup,
-    );
+    const sidePanel = useSidePanel();
+    const sidePanelMutation = useSidePanelMutation();
+    const extensionType = useAppSelector((state) => state.app.extensionViewType);
+    console.log(sidePanelMutation.error, sidePanelMutation.data);
```
brancoder: leftover console log

marc2332: Done

apps/wallet/src/ui/app/pages/site-connect/index.tsx
```diff
@@ -47,6 +50,14 @@ export function SiteConnectPage() {
         return preselectedAccounts.concat(previouslyPermittedAccounts);
     });
 
+    function close() {
```
VmMad:
```suggestion
    function handleConnectionFinish() {
```

apps/wallet/src/polyfills/sidepanel.ts
```diff
@@ -0,0 +1,48 @@
...
+        return 'chrome' in window;
+    }
+
+    static async isEnabled() {
+        const options = await chrome.sidePanel.getOptions({});
+        return options.enabled || false;
+    }
+
+    static async enableAndGoTo(path: string) {
+        await chrome.sidePanel.setOptions({
```
VmMad, line 30: Instead of setting the options, it would be better to modify the manifest to be able to open the side panel from everywhere if possible
https://developer.chrome.com/docs/extensions/reference/api/sidePanel#every-site

apps/wallet/src/polyfills/sidepanel.ts
VmMad, line 1: This file is missing a copyright header

VmMad (changes requested):

apps/wallet/src/polyfills/sidepanel.ts
marc2332: No, as discussed in Slack

apps/wallet/src/ui/app/pages/site-connect/index.tsx
marc2332: I renamed it to `handleOnClose` to be more generic

apps/wallet/src/ui/app/pages/site-connect/index.tsx
```diff
@@ -47,6 +50,14 @@ export function SiteConnectPage() {
         return preselectedAccounts.concat(previouslyPermittedAccounts);
     });
 
+    function handleOnClose() {
```
VmMad: this function should not be called `close` because it doesn't always close, that's why i think `handleOnFinish` is a better name

VmMad (dismissed):
There's a couple of errors that i got when testing:
1. If I "Enable" the side panel and then close it, I'm not able to connect to the wallet dashboard for example (first connection, not autoconnect)
2. Transactions are not picked up correctly by the side panel, it looks like it reloads the homepage

apps/wallet/src/polyfills/sidepanel.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    static async open(path: string) {
+        const window = await chrome.windows.getCurrent({ populate: true });
+
+        if (!window.id) {
+            throw new Error('Failed to detect Window');
+        }
+
+        this.enableAndGoTo(path);
```
brancoder: should we add await here?
 `await this.enableAndGoTo(path);`

apps/wallet/src/background/permissions.ts
```diff
@@ -178,6 +179,13 @@ class Permissions {
             connection.pagelink,
             existingPermission,
         );
+
+        if (await SidePanel.isEnabled()) {
+            await SidePanel.enableAndGoTo(Permissions.getUiUrl(pRequest.id));
+            this.#_permissionWindows.set(pRequest.id, -3154731);
```
brancoder: we should extract the magic number -3154731 to a const

apps/wallet/src/background/transactions.ts
```diff
@@ -159,6 +160,38 @@ class Transactions {
 
         const txRequest = this.createTransactionRequest(request, origin, favIcon);
         await this.storeTransactionRequest(txRequest);
+
+        if (await SidePanel.isEnabled()) {
+            SidePanel.enableAndGoTo(
```
brancoder: are we missing await here?

brancoder (dismissed):
LG just couple of small comments

apps/wallet/src/background/permissions.ts
panteleymonchuk: agree, don't understand what is this magic number for

apps/wallet/src/background/transactions.ts
```diff
@@ -159,6 +160,38 @@ class Transactions {
...
+                                txRequest.txResultError = txResultError;
+                                txRequest.txSigned = txSigned;
+                                return txRequest;
+                            }
+                        }
+                        throw new Error('Rejected from user');
+                    }),
+                ),
+            );
+        }
```
panteleymonchuk, line 193: Seems this code duplicate code below. Could we unify it somehow?

apps/wallet/src/polyfills/sidepanel.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        await chrome.sidePanel.open({ windowId: window.id });
```
panteleymonchuk, line 54:
```suggestion
        return chrome.sidePanel.open({ windowId: window.id });
```

panteleymonchuk: I believe all of methods (or many of them) need to return promise, so we can use async/await properly and avoid race conditions

apps/wallet/src/ui/app/pages/approval-request/index.tsx
```diff
@@ -6,17 +6,37 @@ import {
...
+    const dispatch = useAppDispatch();
+
+    // Wallet fetches pending transaction requests on boot, but because sidepanel
+    // stays open we need to manually notify the wallet, to do this we check for a fetch query param
+    // If present and equal to true then dispatch a message to fetch the latest transaction requests.
+    // We need to dispatch this synchronously rather than async.
+    if (!transactionRequestsLoading.current && fetch) {
+        dispatch(clearTransactionRequests());
+        backgroundClient.sendGetTransactionRequests();
+        transactionRequestsLoading.current = true;
```
panteleymonchuk, line 37: Shoudn't it be wrapped by useEffect? Looks like a side effect

panteleymonchuk (changes requested):

apps/wallet/src/ui/app/pages/approval-request/index.tsx
marc2332: I did not wrap it in a useEffect so that it would run synchronously before this other useEffect ran https://github.com/iotaledger/iota/pull/9059/changes/BASE..90f69ca3a4e9fc9a62fbd63164510351d4058db6#diff-b28c15995c4142aa7f6af64e7d7210047d681fa33509f2bcd5043777a2256ad8R51 . I am not a fan of it, but I didnt see a better way at the moment

apps/wallet/src/polyfills/sidepanel.ts
marc2332: Not sure what you mean by that, the function already returns a promise, its `async` https://github.com/iotaledger/iota/pull/9059/changes/BASE..90f69ca3a4e9fc9a62fbd63164510351d4058db6#diff-f95560796f840557e4f3ce27ee0b77a699038a274ba21e7c717db8c1626a70f1R29

apps/wallet/src/background/transactions.ts
marc2332: I should have left a comment regarding `3154731`, I don't remember either.

apps/wallet/src/background/transactions.ts
marc2332: sure

apps/wallet/src/background/transactions.ts
marc2332: The code is not duplicated if you look carefully, there are small but key differences

apps/wallet/src/background/permissions.ts
marc2332: the number is an unique hardcoded ID just for the sidepanel, popup windows have their own random number, but sidepanel doesnt so I made one.

Now its better, check https://github.com/iotaledger/iota/pull/9059/commits/d274abb7588e45598b3ea1959b55d4860c26f6f5

apps/wallet/src/ui/app/pages/approval-request/index.tsx
panteleymonchuk: can we use useLayoutEffect then? It will allow you to dispatch actions after dom mutation but before browser painting.

apps/wallet/src/ui/app/components/menu/content/WalletSettingsMenuList.tsx
```diff
@@ -135,8 +143,17 @@ export function MenuList() {
             title: 'Expand View',
             icon: <Expand />,
             onClick: () => openInNewTab('/tokens'),
-            hidden: !isAppPopup,
+            hidden: extensionType != ExtensionViewType.Popup,
```
brancoder: should this be `!==`?

brancoder (commented):
LG just a comment about comparison in wallet settings menu

msarcev (changes requested):
Upon ledger importing, when modal mode we get redirected to fullscreen to connect the ledger (ledger select dialog). With sidepanel mode this doesn't seam to work, instead i get a "Ledger connection failed" error.

apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
```diff
@@ -113,7 +117,7 @@ export function AddAccountPage() {
         switch (actionType) {
             case AccountsFormType.ImportLedger:
                 ampli.openedConnectLedgerFlow({ sourceFlow });
-                if (isPopup) {
+                if (isPopupOrSidePanel) {
                     await openTabWithSearchParam('showLedger', 'true');
```
msarcev, line 121: Do you want to also do `await SidePanel.disable();` before going to tab mode ? Wdyt ? 
Otherwise we end up in fullscreen mode, but in the Setting panel the option still says "Disable Panel".

marc2332: disabing seems a bit too much no? I mean, you were in sidepanel for a reason to begin with

msarcev (approved):

panteleymonchuk (approved):
🎸

msarcev (approved):

brancoder (approved):
LGTM

VmMad (approved):

PR #9072, opened 2025-10-28
fix(iota): fix iota client ptb --upgrade command

### LINKED ISSUE #9056: iota client ptb --upgrade is broken
upgrade call fails with this error
```
Encountered error when building PTB:
  × Error when processing PTB
   ╭─[1:1]
 1 │ --upgrade packages/iota-names @0xd64205c4b10eff4b4adb00ab6f754cda8d8e7525985a31307f7e232481dfaf6e
   · ──────────────┬──────────────
   ·               ╰── Failed to publish the Move module(s), reason: Modules must all have 0x0 as their addresses. Violated by module ModuleId { address: b9d617f24c84826bf660a2f4031951678cc80c264aebc4413459fb2a95ada9ba, name: Identifier("name") }
```
even though in the Move.toml the address is set to 0x0

Thoralf-M:
# Description of change

Fix the ptb --upgrade command

skip_dependency_verification is set to true, because the dependencies are now fetched automatically based on the env and would then quickly not match anymore, it's the same in the client  upgrade command.
Added code to save the current dir and restore it after building a package, because otherwise it wouldn't work to upgrade multiple packages in a single PTB as the first one would set the base path to the initial package and then the referenced path wouldn't be valid anymore.

Fixes https://github.com/iotaledger/iota/issues/9056

Upgrading two packages at once as dry-run:
```
iota client ptb \
--upgrade packages/subnames @0x9111bcb07d64182407708651e0ba50a52b7be08027c39a3fb0db01ce1b87123b \
--upgrade packages/iota-names @0xd64205c4b10eff4b4adb00ab6f754cda8d8e7525985a31307f7e232481dfaf6e \
--sender @0x1ca3c38e888493f869ac35346a2041d6cf87b0b935ebba14b35a08811d8a76e4 \
--dry-run
````

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Fixed `iota client ptb --ugprade command`
- [ ] Rust SDK:
- [ ] REST API:

### CODE REVIEW

**Review by DaughterOfMars — APPROVED** [review]

**Comment by thibault-martinez** [comment]
> skip_dependency_verification is set to false

Did you mean `true`?

File: crates/iota/src/client_ptb/builder.rs [hunk 1]
```diff
@@ -1028,29 +1035,34 @@ impl<'a> PTBBuilder<'a> {
...
+                        build_config.install_dir.clone(),
+                        &chain_id,
+                        previous_id,
+                    )
+                    .map_err(|e| err!(path_loc, "{e}"))?;
+                }
+
+                // Restore the initial directory so subsequent commands are not affected
+                std::env::set_current_dir(initial_dir)
+                    .map_err(|e| err!(path_loc, "Failed to restore initial path: {e}"))?;
```
thibault-martinez: Is this something that should be done also if there was an error early return?

Thoralf-M:
> > skip_dependency_verification is set to false
> 
> Did you mean `true`?

Yes!

crates/iota/src/client_ptb/builder.rs
Thoralf-M: I don't think this matters as an error would just abort the transaction building and it then has no effect on subsequent commands

crates/iota/src/client_ptb/builder.rs
```diff
@@ -1014,12 +1014,19 @@ impl<'a> PTBBuilder<'a> {
...
-                let package_path = Path::new(&package_path);
-                let build_config = resolve_lock_file_path(build_config.clone(), Some(package_path))
-                    .map_err(|e| err!(path_loc, "{e}"))?;
+                let package_path = Path::new(&package_path)
+                    .canonicalize()
+                    .map_err(|e| err!(path_loc, "Failed to canonicalize package path: {e}"))?;
+
+                // Save the initial current directory
+                let initial_dir = std::env::current_dir()
+                    .map_err(|e| err!(path_loc, "Failed to get current path: {e}"))?;
```
thibault-martinez:
```suggestion
                    .map_err(|e| err!(path_loc, "Failed to get current directory: {e}"))?;
```

crates/iota/src/client_ptb/builder.rs
thibault-martinez:
```suggestion
                    .map_err(|e| err!(path_loc, "Failed to restore initial directory: {e}"))?;
```

thibault-martinez (approved):

PR #9086, opened 2025-10-28
feat(gRPC): add improved transaction filter

muXxer:
# Description of change

This PR adds a new transaction filter and a new event filter for gRPC (not used yet by the API).
It combines the two existing json-rpc filters into one, but also adds the possibility to filter for transactions that match a certain event filter. Also, the possibility to chain the filters was added (`Any`, `All`, `Or`, `And`) similar to the event filters.

### DIFF
--- a/crates/iota-core/src/event_filter.rs
+++ b/crates/iota-core/src/event_filter.rs
@@ -1,0 +1,129 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_json_rpc_types::{Filter, IotaEvent};
+use iota_metrics::monitored_scope;
+use iota_types::{
+    base_types::{IotaAddress, ObjectID, TransactionDigest},
+    error::IotaResult,
+};
+use move_core_types::{identifier::Identifier, language_storage::StructTag};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum GrpcEventFilter {
+    // Logical AND of several filters.
+    All(Vec<GrpcEventFilter>),
+    // Logical OR of several filters.
+    Any(Vec<GrpcEventFilter>),
+    // Logical NOT of a filter.
+    Not(Box<GrpcEventFilter>),
+
+    /// Return events emitted by the given transaction.
+    /// TODO: do we need that filter for streaming?
+    Transaction(
+        /// digest of the transaction, as base-64 encoded string
+        TransactionDigest,
+    ),
+
+    /// Query by sender address.
+    /// TODO: Is the same as the transaction's sender address. Do we need both?
+    Sender(IotaAddress),
+
+    /// Return events emitted in a specified Move package.
+    MovePackage(ObjectID),
+
+    /// Return events emitted in a specified Move module.
+    /// If the event is defined in Module A but emitted in a tx with Module B,
+    /// query `MoveModule` by module B returns the event.
+    /// Query `MoveEventModule` by module A returns the event too.
+    MoveModule {
+        /// the Move package ID
+        package: ObjectID,
+        /// the module name
+        module: Identifier,
+    },
+    /// Return events with the given Move module name where the event struct is
+    /// defined. If the event is defined in Module A but emitted in a tx
+    /// with Module B, query `MoveEventModule` by module A returns the
+    /// event. Query `MoveModule` by module B returns the event too.
+    MoveEventModule {
+        /// the Move package ID
+        package: ObjectID,
+        /// the module name
+        module: Identifier,
+    },
+    /// Return events with the given Move event struct name (struct tag).
+    /// For example, if the event is defined in `0xabcd::MyModule`, and named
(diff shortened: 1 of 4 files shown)


crates/iota-core/src/transaction_filter.rs
```diff
@@ -0,0 +1,152 @@
...
+    object::Owner,
+    transaction::{TransactionData, TransactionDataAPI},
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone)]
+pub struct TransactionDataWithEffectsAndEvents {
+    pub tx_data: TransactionData,
+    pub effects: IotaTransactionBlockEffects,
+    pub events: IotaTransactionBlockEvents,
```
bingyanglin, line 23: Should the effects and events type be `iota_types::effects::TransactionEffects` and `iota_types::effects::TransactionEvents`, respectively?
Other parts look good to me (Just a reminder the upstream removed `AND`/`OR` filters in `EventFilters` at https://github.com/MystenLabs/sui/commit/9c6029ba97c5aa9dfc67975eca373c5c7b92ecea).

piotrm50 (approved):

crates/iota-core/src/transaction_filter.rs
muXxer: I like the filters, but we should limit the recursion depth.

The types should be incrementally changed by other PR's to this feature branch to reflect the new gRPC types.

crates/iota-core/src/event_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+    // Logical AND of several filters.
+    All(Vec<GrpcEventFilter>),
+    // Logical OR of several filters.
+    Any(Vec<GrpcEventFilter>),
+    // Logical NOT of a filter.
+    Not(Box<GrpcEventFilter>),
+
+    /// Return events emitted by the given transaction.
+    /// TODO: do we need that filter for streaming?
+    Transaction(
```
alexsporn: This filter seems weird. Why would you filter for events of a specific transaction? I would remove it

crates/iota-core/src/event_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+        /// digest of the transaction, as base-64 encoded string
+        TransactionDigest,
+    ),
+
+    /// Query by sender address.
+    /// TODO: Is the same as the transaction's sender address. Do we need both?
+    Sender(IotaAddress),
```
alexsporn, line 25: Remove

crates/iota-core/src/event_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Return events whose JSON representation contains the given field path
+    /// with the specified value. The path should be a JSON pointer as
+    /// defined in RFC 6901.
+    MoveEventField {
+        path: String,
+        value: Value,
+    },
+
+    /// Return events emitted in [start_time, end_time] interval
+    TimeRange {
```
alexsporn: This makes no sense for the stream

crates/iota-core/src/transaction_filter.rs
```diff
@@ -0,0 +1,142 @@
...
+    // Logical OR of several filters.
+    Any(Vec<GrpcTransactionFilter>),
+    // Logical NOT of a filter.
+    Not(Box<GrpcTransactionFilter>),
+
+    /// Query transactions of any given kind in the input.
+    TransactionKind(Vec<IotaTransactionKind>),
+
+    /// Query by sender address.
+    FromAddress(IotaAddress),
```
alexsporn: Would maybe explicitly call it `Sender` here, since that is how it is called in all of the protocol

crates/iota-core/src/transaction_filter.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Query by recipient address.
+    ToAddress(IotaAddress),
```
alexsporn: How is a `recipient` defined here? An address that takes ownership of a mutated/created object?

crates/iota-core/src/event_filter.rs
muXxer: I will keep this one, because we decided to have the possibility to stream single events without the transactions (next to the tx streaming). According to @miker83z this is the main use case by dapp developers, and they only need the events mostly.

crates/iota-core/src/event_filter.rs
muXxer: Yeah had it removed locally already. Pushed a new commit! 👍

crates/iota-core/src/event_filter.rs
muXxer: Yeah had it removed locally already. Pushed a new commit! 👍

crates/iota-core/src/transaction_filter.rs
muXxer: Renamed to `Sender`.

crates/iota-core/src/transaction_filter.rs
muXxer: Renamed to `Receiver` and added a comment.

crates/iota-core/src/event_filter.rs
alexsporn: Ah, this is the `Event` sender, gotcha 👍

bingyanglin (approved):

PR #9096, opened 2025-10-29
upstream(indexer): pruner updates watermarks lower bound

### LINKED ISSUE #8731: upstream(indexer): pruner updates watermarks lower bound
indexer: pruner updates watermarks lower bound: https://github.com/MystenLabs/sui/pull/19650

tomxey:
# Description of change

Original description:

> With the committer writing upper bounds, the pruner can now read from watermarks and determine whether the lower bounds need to be updated. Pruner does this by spawning a separate task, without touching the extant pruning logic (so all things are as is.) It will ignore any entries from watermarks that do not correspond to a PrunableTable variant.

fixes https://github.com/iotaledger/iota/issues/8731

### DIFF
--- a/crates/iota-indexer/src/models/watermarks.rs
+++ b/crates/iota-indexer/src/models/watermarks.rs
@@ -23,17 +23,17 @@ pub struct StoredWatermark {
     /// this field. Pruner uses this to determine if pruning is necessary
     /// based on the retention policy.
     pub epoch_hi_inclusive: i64,
-    /// Inclusive lower epoch bound for this entity's data. Pruner updates this
-    /// field when the epoch range exceeds the retention policy.
-    pub epoch_lo: i64,
     /// Inclusive upper checkpoint bound for this entity's data. Committer
     /// updates this field. All data of this entity in the checkpoint must
     /// be persisted before advancing this watermark. The committer refers
     /// to this on disaster recovery to resume writing.
     pub checkpoint_hi_inclusive: i64,
-    /// Inclusive upper transaction sequence number bound for this entity's
+    /// Exclusive upper transaction sequence number bound for this entity's
     /// data. Committer updates this field.
-    pub tx_hi_inclusive: i64,
+    pub tx_hi: i64,
+    /// Inclusive lower epoch bound for this entity's data. Pruner updates this
+    /// field when the epoch range exceeds the retention policy.
+    pub epoch_lo: i64,
     /// Inclusive low watermark that the pruner advances. Corresponds to the
     /// epoch id, checkpoint sequence number, or tx sequence number
     /// depending on the entity. Data before this watermark is considered
@@ -45,18 +45,18 @@ pub struct StoredWatermark {
     /// determine whether to prune or wait long enough that all in-flight
     /// reads complete or timeout before it acts on an updated watermark.
     pub timestamp_ms: i64,
-    /// Column used by the pruner to track its true progress. Data at and below
-    /// this watermark can be immediately pruned.
-    pub pruner_hi_inclusive: Option<i64>,
+    /// Column used by the pruner to track its true progress. Data below this
+    /// watermark can be immediately pruned.
+    pub pruner_hi: i64,
 }
 
 impl StoredWatermark {
     pub fn from_upper_bound_update(entity: &str, watermark: CommitterWatermark) -> Self {
         StoredWatermark {
             entity: entity.to_string(),
-            epoch_hi_inclusive: watermark.epoch as i64,
-            checkpoint_hi_inclusive: watermark.cp as i64,
-            tx_hi_inclusive: watermark.tx as i64,
+            epoch_hi_inclusive: watermark.epoch_hi_inclusive as i64,
+            checkpoint_hi_inclusive: watermark.checkpoint_hi_inclusive as i64,
+            tx_hi: watermark.tx_hi as i64,
             ..StoredWatermark::default()
         }
     }
--- a/crates/iota-indexer/src/models/watermarks.rs
+++ b/crates/iota-indexer/src/models/watermarks.rs
@@ -2,15 +2,18 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 //
+use std::str::FromStr;
+
 use diesel::prelude::*;
 
 use crate::{
     ingestion::common::persist::CommitterWatermark,
+    pruning::pruner::PrunableTable,
     schema::watermarks::{self},
 };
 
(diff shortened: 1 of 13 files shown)


crates/iota-indexer/src/models/watermarks.rs
```diff
@@ -44,7 +47,7 @@ pub struct StoredWatermark {
     pub timestamp_ms: i64,
     /// Column used by the pruner to track its true progress. Data at and below
     /// this watermark can be immediately pruned.
-    pub pruner_lo: Option<i64>,
+    pub pruner_hi_inclusive: Option<i64>,
```
kodemartin: What purpose lies behind this change. I don't see this being used, and it seems incosistent with its documentation.

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -179,3 +222,58 @@ impl Pruner {
...
 }
+
+/// Task to periodically query the `watermarks` table and update the lower
+/// bounds for all watermarks if the entry exceeds epoch-level retention policy.
+async fn update_watermarks_lower_bounds_task(
+    store: PgIndexerStore,
+    retention_policies: HashMap<PrunableTable, u64>,
+    cancel: CancellationToken,
+) -> IndexerResult<()> {
+    let mut interval = tokio::time::interval(Duration::from_secs(5));
```
kodemartin: I would suggest a `const` instead of hardcoding the duration here.

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -179,3 +222,58 @@ impl Pruner {
...
+            continue;
+        };
+        let Some(epochs_to_keep) = retention_policies.get(&prunable_table) else {
+            error!("no retention policy found for prunable table {prunable_table}");
+            continue;
+        };
+        if let Some(new_epoch_lo) = watermark.new_epoch_lo(*epochs_to_keep) {
+            lower_bound_updates.push((prunable_table, new_epoch_lo));
+        };
+    }
```
kodemartin, line 276: I think it is plausible to let the loop finish gracefully updating all watermarks, making the update atomic.

Cancellation can be checked before the loop.

kodemartin (commented):
Looks good with some minor comments/suggestions.

crates/iota-indexer/src/models/watermarks.rs
tomxey: For pruner, this is the highest value that can be immediately deleted, and it's inclusive, so I would say the name makes sense for now. It's not used yet though so it still can change potentially.

crates/iota-indexer/src/pruning/pruner.rs
tomxey: The update is atomic, as it happens once after the loop. The loop is only used to collect the values that need to be updated, so I think it's fine if we return in the middle of it.

crates/iota-indexer/src/pruning/pruner.rs
tomxey: Done in 63fb502058b765876b2e841a8f6607e70da8b76c

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -73,6 +79,36 @@ pub enum PrunableTable {
...
+            PrunableTable::TxChangedObjects => tx,
+            PrunableTable::TxDigests => tx,
+            PrunableTable::TxInputObjects => tx,
+            PrunableTable::TxKinds => tx,
+            PrunableTable::TxRecipients => tx,
+            PrunableTable::TxSenders => tx,
+            PrunableTable::Checkpoints => cp,
+            PrunableTable::PrunerCpWatermark => cp,
+        }
+    }
```
sergiupopescu199, line 110: This version could improve a bit the readability.

```suggestion
pub fn select_reader_lo(&self, cp: u64, tx: u64) -> u64 {
    match self {
        PrunableTable::ObjectsHistory
        | PrunableTable::Checkpoints
        | PrunableTable::PrunerCpWatermark => cp,

        PrunableTable::Transactions
        | PrunableTable::Events
        | PrunableTable::EventEmitPackage
        | PrunableTable::EventEmitModule
        | PrunableTable::EventSenders
        | PrunableTable::EventStructInstantiation
        | PrunableTable::EventStructModule
        | PrunableTable::EventStructName
        | PrunableTable::EventStructPackage
        | PrunableTable::TxAffectedAddresses
        | PrunableTable::TxAffectedObjects
        | PrunableTable::TxCallsPkg
        | PrunableTable::TxCallsMod
        | PrunableTable::TxCallsFun
        | PrunableTable::TxChangedObjects
        | PrunableTable::TxDigests
        | PrunableTable::TxInputObjects
        | PrunableTable::TxKinds
        | PrunableTable::TxRecipients
        | PrunableTable::TxSenders => tx,
    }
}
```

crates/iota-indexer/src/pruning/pruner.rs
sergiupopescu199: I do agree with @kodemartin, 

Looking at the `CancellationToken` [implementation](https://docs.rs/tokio-util/latest/src/tokio_util/sync/cancellation_token.rs.html#206) calling frequently the `is_cancelled` method could add some overhead, inside it has a list of nodes, looking at its [implementation](https://docs.rs/tokio-util/latest/src/tokio_util/sync/cancellation_token/tree_node.rs.html#83). 

It does lock the Mutex every time we call this method, unless we have tens of thousands of records, the performance cost should be negligible, I think having the cancellation token on the `tokio::select` could be better, makes also the flow clearer, we either exit on cancellation token or we do update the lower bounds.

sergiupopescu199 (approved):
LGTM 🌸

crates/iota-indexer/src/pruning/pruner.rs
kodemartin: The db operation at the end is indeed atomic. My point is that once we start preparing the watermarks for update, I think graceful shutdown implies letting the preparation to finish and update the data.

crates/iota-indexer/src/pruning/pruner.rs
tomxey: Done in b5d97cdd563be00a94e8a449186821375b70fa97

crates/iota-indexer/src/pruning/pruner.rs
tomxey: Moved the check before the loop
469bb4131fa605df6d110d67e820f0921eb00bd6

kodemartin (approved):

PR #9100, opened 2025-10-30
refactor: avoid gas coins double check

### LINKED ISSUE #8847: [AA][tx-checks] Avoid gas coins double check

valeriyr:
# Description of change

This PR contains a refactoring to avoid gas coins double-check during transaction handling.

fixes #8847

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -901,27 +901,38 @@ impl AuthorityState {
             self.get_backing_package_store().as_ref(),
         )?;
 
-        let (tx_input_objects, tx_receiving_objects, auth_input_objects, move_account) =
+        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) =
             self.read_objects_for_signing(&transaction, epoch)?;
 
-        if let Some(move_authenticator) = transaction.sender_move_authenticator() {
+        let move_authenticator = transaction.sender_move_authenticator();
+
+        let (gas_status, tx_checked_input_objects, auth_checked_input_objects, authenticator_info) =
+            self.check_transaction_inputs_for_signing(
+                protocol_config,
+                reference_gas_price,
+                tx_data,
+                tx_input_objects,
+                &tx_receiving_objects,
+                move_authenticator,
+                auth_input_objects,
+                account_object,
+            )?;
+
+        check_coin_deny_list_v1_during_signing(
+            tx_data.sender(),
+            &tx_checked_input_objects,
+            &tx_receiving_objects,
+            &self.get_object_store(),
+        )?;
+
+        if let Some(move_authenticator) = move_authenticator {
             // It is supposed that `MoveAuthenticator` availability is checked in
             // `SenderSignedData::validity_check`.
 
-            let auth_input_objects =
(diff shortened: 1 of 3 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -901,27 +901,38 @@ impl AuthorityState {
             self.get_backing_package_store().as_ref(),
         )?;
 
-        let (tx_input_objects, tx_receiving_objects, auth_input_objects, move_account) =
+        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) =
```
miker83z, line 908:
```suggestion
        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object_opt) =
```

crates/iota-core/src/authority.rs
```diff
@@ -901,27 +901,38 @@ impl AuthorityState {
...
             self.read_objects_for_signing(&transaction, epoch)?;
 
-        if let Some(move_authenticator) = transaction.sender_move_authenticator() {
+        let move_authenticator = transaction.sender_move_authenticator();
+
+        let (gas_status, tx_checked_input_objects, auth_checked_input_objects, authenticator_info) =
```
miker83z, line 921:
```suggestion
        let (gas_status, tx_checked_input_objects, auth_checked_input_objects_opt, authenticator_info_opt) =
```

crates/iota-core/src/authority.rs
```diff
@@ -901,27 +901,38 @@ impl AuthorityState {
             self.get_backing_package_store().as_ref(),
         )?;
 
-        let (tx_input_objects, tx_receiving_objects, auth_input_objects, move_account) =
+        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) =
             self.read_objects_for_signing(&transaction, epoch)?;
 
-        if let Some(move_authenticator) = transaction.sender_move_authenticator() {
+        let move_authenticator = transaction.sender_move_authenticator();
```
miker83z, line 914:
```suggestion
        let move_authenticator_opt = transaction.sender_move_authenticator();
```

miker83z (approved):
LGTM! I left a suggestion to use `_opt` so that it is clearer that we don't always expect a Move authenticator there; however, if you prefer, instead of those suggestions you could write a long comment explaining what is happening

valeriyr:
> LGTM! I left a suggestion to use `_opt` so that it is clearer that we don't always expect a Move authenticator there; however, if you prefer, instead of those suggestions you could write a long comment explaining what is happening

I added some comments for clarification.
You can check them here: d7ca134c20

Dkwcs (approved):
LGTM

PR #9120, opened 2025-10-31
feat(consensus): update metrics

### LINKED ISSUE #9011: starfish: proposed block size metric is not correct

VorobyevIlya:
# Description of change

Pr updates and fixes metrics that may have been broken during sthe tarfish implementation.
It updates descriptions, naming, and removes some obsolete metrics.

fixes #9011 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -360,7 +360,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 self.context
                     .metrics
                     .node_metrics
-                    .invalid_shard_in_bundles
+                    .bundles_with_invalid_shards
                     .with_label_values(&[
                         peer_hostname.as_str(),
                         "handle_subscribed_block_bundle",
@@ -386,7 +386,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 self.context
                     .metrics
                     .node_metrics
-                    .invalid_shard_in_bundles
+                    .bundles_with_invalid_shards
                     .with_label_values(&[
                         peer_hostname.as_str(),
                         "handle_subscribed_block_bundle",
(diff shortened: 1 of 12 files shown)


crates/starfish/core/src/commit_syncer.rs
```diff
@@ -678,27 +678,7 @@ impl<C: NetworkClient> CommitSyncer<C> {
...
-            let forward_drift = block_header.timestamp_ms().saturating_sub(now_ms);
-            if forward_drift == 0 {
-                continue;
-            };
-            let peer_hostname = &inner.context.committee.authority(target_authority).hostname;
-            inner
-                .context
-                .metrics
-                .node_metrics
-                .block_timestamp_drift_ms
```
piotrm50, line 696: Why are we not tracking this anymore? I see that the metric itself is not removed - is it used anywhere else?
If we don't want to track this, then we should also not return ` _voting_block_headers` in the task above. Although I think it might be good to track the timestamp drift anyway.

crates/starfish/core/src/authority_service.rs
```diff
@@ -202,7 +202,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
             self.context
                 .metrics
                 .node_metrics
-                .invalid_block_headers
+                .bundles_with_invalid_headers
```
polinikita: I would use one metric with different type when processing bundles. For instance, no need to have this string "handle_subscribed_block_bundle" as it can be called only from here. Instead I would use e.g. bundles_with_invalid_parts and add a label "shard" or "header" when increasing it.

polinikita (approved):
I suggest to aggregate two metrics in one

crates/starfish/core/src/commit_syncer.rs
VorobyevIlya: We still track it in the authority service when we receive a bundle. I removed it from the commit syncer only. If we are trying to sync through commit syncer, then we are probably lagging far behind, and there shouldn't be any forward drift.

crates/starfish/core/src/authority_service.rs

crates/starfish/core/src/commit_syncer.rs
VorobyevIlya: I removed _voting_block_headers

piotrm50 (approved):

PR #9128, opened 2025-11-03
feat(gRPC): field masking and generation

bingyanglin:
# Description of change

- Ported field masking mechanism, including
  - The `field` module 
  - The `iota-proto-build` crate to generate fields, so we can use `FieldMaskUtil::validate` to verify the client-requested field paths.
- Change `EventID` to `EventId` to meet `prost`'s naming convention, so the generated Rust code from `protobuf` can be used without additional conversion (like change `ID` to `Id`).

Part of #8688


crates/iota-grpc-types/Makefile
thibault-martinez, line 1: Is this really necessary?

bingyanglin: Sui has the Makefile but more complicated. I am thinking it might be good to keep for ease of future extension but not sure.

crates/iota-proto-build/src/generate_fields.rs
```diff
@@ -0,0 +1,302 @@
...
+            }
+
+            pub use _field_impls::*;
+        };
+
+        let ast: syn::File = syn::parse2(code).expect("not a valid tokenstream");
+        let code = prettyplease::unparse(&ast);
+
+        // Add IOTA license header
+        buf.push_str("// Copyright (c) 2025 IOTA Stiftung\n");
```
muXxer: Same here.

muXxer (approved):

muXxer:
I fixed the copyrights, and replaced the makefile with a bash script. I also fixed the order of the imports, otherwise the output was random, and the git diff check failed.

thibault-martinez (approved):

Issue #7528, opened 2025-06-25
Rename Components and Functions/Fields


piotrm50:
Rename components and functions/fields to be more aligned with established naming conventions.

piotrm50:
We renamed some components that were standing out the most as part of different tasks.

PR #9129, opened 2025-11-03
refactor(consensus): speedup flushing and getting serialized transactions and headers in Dag State

### LINKED ISSUE #9089: starfish: create a test when a transaction is sequenced but the corresponding header is not traversed.
Right now it is possible to sequence transaction, but not commit the header. This might create certain issues (e.g. when reading the transction we assume that the header is in the storage and read the transaction commitment from it)

polinikita:
# Description of change

In this PR, we refactor multiple components:
- two new methods in `DagState `- `get_serialized_transactions()` and `get_serialized_block_headers()`. They are used when `HeaderSynchronizer` or `TransactionsSynchronizer` of a peer requests bytes of headers or transactions from the local node. This allows to not deserialize headers or transactions while the DagState is locked for the read operation.
- the fields of  in `DagState ` - `recent_shards` and `recent_transactions` are replaced in favour of `recent_shards_by_authority` and `recent_transactions_by_authority`. This is important to improve the time we spend for eviction (since it took a linear time before for every `flush()` call). Also, when doing commit syncing it allows to evict transactions faster
- Synchronizer is now called HeaderSynchronizer to avoid confusion
- Many fields called `blocks` are replaced by `headers`
- A new test `test_sequenced_transactions_no_headers` is added. It allows to check the case when some transactions are sequenced but the corresponding block headers are not.

Fixes #9089 and #9134 and partially #7528

### DIFF
--- a/crates/starfish/core/src/cordial_knowledge.rs
+++ b/crates/starfish/core/src/cordial_knowledge.rs
@@ -559,16 +559,18 @@ impl CordialKnowledge {
     /// Called when older rounds should be pruned globally.
     fn handle_evict_below(
         &mut self,
-        rounds: Vec<Round>,
+        evicted_rounds: Vec<Round>,
     ) -> Option<Vec<Vec<ConnectionKnowledgeMessage>>> {
         // Evict locally
         for (index, btree_map) in &mut self.cordial_knowledge.iter_mut().enumerate() {
-            let split_round = rounds[index];
+            // Increase by 1 for splitting as the evicted rounds are gone from memory
+            let split_round = evicted_rounds[index] + 1;
+            // Remove everything strictly below this round
             *btree_map = btree_map.split_off(&split_round);
         }
 
         // Prepare message for per-connection knowledge about eviction
-        self.prepare_evict_msgs(rounds)
+        self.prepare_evict_msgs(evicted_rounds)
     }
     #[inline]
     fn prepare_evict_msgs(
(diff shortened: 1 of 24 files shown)


filipdulic (approved):

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -848,15 +850,15 @@ impl ConnectionKnowledge {
     }
 
     /// Evict all connection knowledge below the given rounds (exclusive)
-    fn evict_below(&mut self, rounds_exclusive: Vec<Round>) {
+    fn evict_below(&mut self, rounds_to_evict_exclusive: Vec<Round>) {
```
VorobyevIlya: Why is it exclusive? The round in a vector is also evicted, as far as I can see

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -967,7 +969,7 @@ impl ConnectionKnowledge {
         //    be marked as known
         let own_index = self.context.own_index;
         let mut rounds = vec![Round::MIN; self.context.committee.size()];
-        rounds[own_index] = round_upper_bound_exclusive + 1; // We are supposed to send own block of this round in a bundle when calling this function with this parameter
+        rounds[own_index] = round_upper_bound_exclusive; // We are supposed to send own block of this round in a bundle when calling this function with this parameter
 
         self.evict_below(rounds);
```
VorobyevIlya, line 974: Here we also evict the rounds from the vector, so it is below or equal

VorobyevIlya (approved):

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Changed accordingly

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Renamed

PR #9132, opened 2025-11-04
feat: treat authenticator shared objects as TX input

### LINKED ISSUE #8937: [AA] Authenticator shared objects should be treated as TX input
Check any time `tx.shared_input_objects()` is invoked from the TransactionData and adapt this call to support MoveAuthenticator inputs when needed.

### LINKED ISSUE #9131: MoveAuthenticator transaction crashes the authority in localnet
I was playing around with executing move authenticator transactions in a localnet (branch https://github.com/iotaledger/iota/tree/vm-lang/aa-auth/8116-feature-branch ) with the [dynamic account demo](https://github.com/lzpap/dynamic-account-demo). As soon as I try to execute a transaction where authentication  is passing in move (with a shared object input `&Account`, the authority crashes with this panic:

```
thread 'k#99f25ef6..' panicked at crates/iota-core/src/authority/shared_object_version_manager.rs:213:77:
called `Option::unwrap()` on a `None` value
2025-11-04T09:13:17.185918Z ERROR node{name=k#99f25ef6..}: telemetry_subscribers: panicked at crates/iota-core/src/authority/shared_object_version_manager.rs:213:77:
called `Option::unwrap()` on a `None` value panic.file="crates/iota-core/src/authority/shared_object_version_manager.rs" panic.line=213 panic.column=77
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Notes:
 - when I tried to submit the tx knowing the auth function aborts, nothing crashed but I got an execution error back.
 - The `TransactionData` was not touching shared objects, only `authenticate()`

valeriyr:
# Description of change

Authenticator shared objects are treated as TX input.

fixes #8937
fixes #9131

### DIFF
--- a/crates/iota-analytics-indexer/src/handlers/mod.rs
+++ b/crates/iota-analytics-indexer/src/handlers/mod.rs
@@ -11,7 +11,7 @@ use iota_types::{
     base_types::ObjectID,
     effects::{TransactionEffects, TransactionEffectsAPI},
     object::{Object, Owner, bounded_visitor::BoundedVisitor},
-    transaction::{TransactionData, TransactionDataAPI},
+    transaction::{SenderSignedData, TransactionDataAPI},
 };
 use move_core_types::{
     annotated_value::{MoveStruct, MoveTypeLayout, MoveValue},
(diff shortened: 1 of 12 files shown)


miker83z:
We should discourage to use the TransactionDataAPI functions that are not considering the Move authenticator. Do you think we should add `#[deprecated]` so that these are not used anymore? Or, if deprecated doesn't work, for sure we should put some comments to indicate that the user might miss some inputs due to the Move authenticator.

valeriyr:
> We should discourage to use the TransactionDataAPI functions that are not considering the Move authenticator. Do you think we should add `#[deprecated]` so that these are not used anymore? Or, if deprecated doesn't work, for sure we should put some comments to indicate that the user might miss some inputs due to the Move authenticator.

Adding the `#[deprecated]` attribute leads to generating warnings where these functions are used, so I decided to add comments at the moment.
Renaming functions introduces a lot of changes to the project, so I've left the names as is for now.

miker83z (approved):
LGTM! 
(Please add also "IMPORTANT: " in the part of the comments starting with "This function does not..." )

valeriyr:
> LGTM! (Please add also "IMPORTANT: " in the part of the comments starting with "This function does not..." )

fixed in 19a10d36c4

Dkwcs (approved):
LGTM

PR #9137, opened 2025-11-04
feat(move-build, verifier): allow metadata in bytecode and fill metadata during package build

### LINKED ISSUE #9135: Store attributes in the CompiledModule metadata

miker83z:
# Description of change

This PR introduces a new protocol feature flag to allow the insertion of metadata in the serialized compiled modules (bytecode). This was operation was disabled before this PR, with the flag `no_extraneous_module_bytes`. 

This allows to fill this metadata field during the execution of `build_from_resolution_graph`. 

At the module deserializer level, the only performed check is that metadata have only 1 field and the metadata key is hardcoded. During the execution of the iota-verifier, the metadata value is checked for having well formed metadata struct (later authenticate and view functions checks will be performed here).

Fixes #9135 

### DIFF
--- a/external-crates/move/crates/move-binary-format/src/deserializer.rs
+++ b/external-crates/move/crates/move-binary-format/src/deserializer.rs
@@ -25,10 +25,7 @@ use crate::{
 impl CompiledModule {
     /// Deserialize a &[u8] slice into a `CompiledModule` instance.
     pub fn deserialize_with_defaults(binary: &[u8]) -> BinaryLoaderResult<Self> {
-        Self::deserialize_with_config(
-            binary,
-            &BinaryConfig::with_extraneous_bytes_check(false, false),
-        )
+        Self::deserialize_with_config(binary, &BinaryConfig::with_extraneous_bytes_check(false))
     }
 
     /// Deserialize a &[u8] slice into a `CompiledModule` instance with settings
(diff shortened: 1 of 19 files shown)


Dkwcs:
> This was operation was disabled before this PR, with the flag `no_extraneous_module_bytes`.

I’m a bit out of the context, but why this operation was disabled?

> This allows to fill this metadata field during the execution of build_from_resolution_graph.

Were we able to add any metadata during compilation before?

miker83z:
@Dkwcs 
> > This was operation was disabled before this PR, with the flag `no_extraneous_module_bytes`.
> 
> I’m a bit out of the context, but why this operation was disabled?

We inherited this from Sui, it was disabled because it is not used: https://github.com/mystenlabs/sui/commit/eaed49f03e22c680ee39094392f3147585144801

> > This allows to fill this metadata field during the execution of build_from_resolution_graph.
> 
> Were we able to add any metadata during compilation before?

Nope, still for the same reason of above.

Dkwcs (approved):

external-crates/move/crates/move-binary-format/src/deserializer.rs
```diff
@@ -572,16 +578,31 @@ fn build_common_tables(
                 check_table_size!(constant_pool, *constant_pool_max);
             }
             TableType::METADATA => {
-                if binary.check_no_extraneous_bytes() || binary.version() < VERSION_5 {
+                if !binary.allow_iota_metadata_bytes()
+                    && (binary.check_no_extraneous_bytes() || binary.version() < VERSION_5)
```
valeriyr: Is this check correct?
For example, if `allow_iota_metadata_bytes == ture` we don't check `binary.version()`.

valeriyr (approved):
LGTM

external-crates/move/crates/move-binary-format/src/deserializer.rs
miker83z: I was assuming that we reached this point coming from iota and then with the correct version (> 5). But it is safer without this assumption -> [00a1b46](https://github.com/iotaledger/iota/pull/9137/commits/00a1b4625734e374b535a1727ee67f98c1e96d27).

PR #9144, opened 2025-11-04
fix(docker): determine build script location to resolve relative paths correctly

roman1e2f5p8s:
# Description of change

The docker build scripts invoke another script using a relative path, so running them from some other directory, different from their location, will not work. For example, the instructions we have here https://github.com/iotaledger/iota/tree/develop/dev-tools/iota-private-network#1-build-docker-images do not work (__at least on Linux__) exactly because of this problem. This PR adds changes to determine docker build scripts' location so that relative paths used in the scripts can be resolved correctly.

Build a docker image, for example `iota-node`, from different working directories:

- From the root directory of the monorepo:

   ```sh
   cd $(git rev-parse --show-toplevel)
   docker/iota-node/build.sh -t iota-node --no-cache
   ```

   On `develop`, the build command fails with `docker/iota-node/build.sh: line 5: ./../utils/build-script.sh: No such file or directory`.

- From the `dev-tools/iota-private-network` directory as instructed [here](https://github.com/iotaledger/iota/tree/develop/dev-tools/iota-private-network#1-build-docker-images):

   ```sh
   cd "$(git rev-parse --show-toplevel)/dev-tools/iota-private-network"
   ../../docker/iota-node/build.sh -t iota-node --no-cache
   ```

   On `develop`, the build command fails with `../../docker/iota-node/build.sh: line 5: ./../utils/build-script.sh: No such file or directory`.

- From the `docker/iota-node` directory:

   ```sh
   cd "$(git rev-parse --show-toplevel)/docker/iota-node"
   ./build.sh -t iota-node --no-cache
   ```

   Only this build command seems to work on `develop`, meaning that we must first navigate to the build script directory and then run the script from there.

### DIFF
--- a/docker/iota-data-ingestion/build.sh
+++ b/docker/iota-data-ingestion/build.sh
@@ -4,6 +4,7 @@
 
 # Determine script's location to resolve the relative path correctly
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd -P)"
-cd "$SCRIPT_DIR"
+# The name of last directory in script's path is used for tag
+BASE_NAME="$(basename "$SCRIPT_DIR")"
 
-./../utils/build-script.sh --image-tag "iotaledger/iota-data-ingestion"
+"$SCRIPT_DIR/../utils/build-script.sh" --image-tag "iotaledger/$BASE_NAME"
--- a/docker/iota-data-ingestion/build.sh
+++ b/docker/iota-data-ingestion/build.sh
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Determine script's location to resolve the relative path correctly
-SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE}")" >/dev/null && pwd -P)
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd -P)"
 cd "$SCRIPT_DIR"
 
 ./../utils/build-script.sh --image-tag "iotaledger/iota-data-ingestion"
--- a/docker/iota-data-ingestion/build.sh
+++ b/docker/iota-data-ingestion/build.sh
@@ -1,4 +1,9 @@
 #!/bin/bash
 # Copyright (c) 2024 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
+
+# Determine script's location to resolve the relative path correctly
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE}")" >/dev/null && pwd -P)
+cd "$SCRIPT_DIR"
+
 ./../utils/build-script.sh --image-tag "iotaledger/iota-data-ingestion"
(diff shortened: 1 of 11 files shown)


docker/iota-data-ingestion/build.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
 # Copyright (c) 2024 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
+
+# Determine script's location to resolve the relative path correctly
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE}")" >/dev/null && pwd -P)
+cd "$SCRIPT_DIR"
```
alexsporn: doing a `cd` in a script is imho a bad design. If you are already calculating the `SCRIPT_DIR` then why not fix the invocation to use that instead?

alexsporn (changes requested):

docker/iota-data-ingestion/build.sh
alexsporn: You should also add `"` to the outer subshell in case there are spaces. Also this should use `BASH_SOURCE[0]` to get the script name only (like `$0`)

So `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd -P)"`

roman1e2f5p8s: > doing a `cd` in a script is imho a bad design. If you are already calculating the `SCRIPT_DIR` then why not fix the invocation to use that instead?

Right. I just have seen `cd` is used in multiple scripts, for example, [here](https://github.com/iotaledger/iota/blob/f6aaba41ac58c93f7fb8e6f2f381ea22d66804a4/scripts/simtest/install.sh#L6), so opted for that. But I will change it.

roman1e2f5p8s: > You should also add `"` to the outer subshell in case there are spaces. Also this should use `BASH_SOURCE[0]` to get the script name only (like `$0`)
> 
> So `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd -P)"`

Good points. Not sure how I missed `[0]` 😄

roman1e2f5p8s: > doing a `cd` in a script is imho a bad design. If you are already calculating the `SCRIPT_DIR` then why not fix the invocation to use that instead?

@alexsporn, we actually have to `cd` because `Dockerfile` will not be found otherwise:
https://github.com/iotaledger/iota/blob/f6aaba41ac58c93f7fb8e6f2f381ea22d66804a4/docker/utils/build-script.sh#L14-L17

roman1e2f5p8s: Alright, I made it working without using `cd`, but not entirely sure if the way I did it is a good approach.
The build scripts now look identical 😄

shufps (approved):

semenov-vladyslav (approved):
It would be nice to avoid copy-paste and unescaped double-quotes inside double-quotes, but if it works then it's fine by me.

kodemartin (approved):

roman1e2f5p8s:
> It would be nice to avoid copy-paste and unescaped double-quotes inside double-quotes, but if it works then it's fine by me.

Thanks for the suggestions, but can you please clarify where exactly the double quotes should be escaped?
Something like this `BASE_NAME="$(basename \"$SCRIPT_DIR\")"` does not work. We could do `BASE_NAME="$(basename $SCRIPT_DIR)"`, but these quotes there are not harmful.

And by "avoid copy-paste" you mean that all the build scripts look identical now? If yes, I think that is because we have a common script in `utils` which has the actual build logic.

muXxer (approved):

alexsporn (approved):

PR #9146, opened 2025-11-04
feat(consensus): fix consensus committed messages metric

### LINKED ISSUE #9017: Refactor BPS metric for Starfish
In Starfish the metric `consensus_committed_messages` has changed it's meaning and is not comparable with the same metric for Mysticeti. This is a bit confusing and problematic, because node operators' and our internal monitoring tool use this metric as one of the healthchecks. This should be changed to measure committed block headers and a new metric should be introduced to measure committed blocks with transactions.

VorobyevIlya:
# Description of change

PR fixes consensus_committed_messages that worked incorrectly for starfish and counted only blocks with non-empty vector of transactions.

fixes #9017 

### DIFF
--- a/crates/iota-core/src/consensus_handler.rs
+++ b/crates/iota-core/src/consensus_handler.rs
@@ -295,14 +295,19 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             .with_label_values(&[&leader_author.to_string()])
             .inc();
 
+        for (authority_index, number_of_committed_headers) in
+            consensus_output.number_of_headers_in_commit_by_authority()
+        {
+            self.last_consensus_stats
+                .stats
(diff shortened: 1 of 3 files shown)


filipdulic (approved):

crates/iota-core/src/consensus_handler.rs
```diff
@@ -295,14 +295,19 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             .with_label_values(&[&leader_author.to_string()])
             .inc();
 
+        for (authority_index, number_of_committed_headers) in
+            consensus_output.number_of_headers_in_commit_by_authority()
```
polinikita, line 299: The naming can be confusing for Mysticeti, but we can stick with this to not rename later.

polinikita (approved):

muXxer (approved):

PR #9152, opened 2025-11-05
upstream(graphql): add mutation payload size limit

### LINKED ISSUE #8726: upstream(graphql): Add a mutation payload size
https://github.com/MystenLabs/sui/pull/18017

kodemartin:
# Description of change

This patch introduces a new limit specific to transaction execution paths (`{mutation.execute,query.DryRun}TransactionBlock`. This allows transaction payloads to exceed the existing query limit size (5KB) and applies an upper that complies with the max size allowed by the protocol.

This fixes issues with large transactions (typically `publish`) not being able to execute.

`query.moveViewCall` is not handled here.

Fixes #8726 

### Release Notes

- [x] GraphQL: Adds a new `maxTxPayloadSize` configuration parameter under the `limits` section. This is the maximum size in bytes allowed for the `txBytes` and `signatures` payload in `mutation.executeTransactionBlock` and for `txBytes` in `query.dryRunTransactionBlock`. The default value is the max transaction size allowed by the protocol (128KB) plus the `Base64` encoding overhead (~1/3 of the size).


thibault-martinez (approved):
dev-tools change LGTM

crates/iota-graphql-rpc/src/extensions/query_limits_checker.rs
```diff
@@ -258,6 +405,17 @@ impl<'a> LimitsTraversal<'a> {
         Ok(())
     }
 
+    /// Error returned if transaction payloads exceed limit. Also sets the
+    /// transaction payload budget to zero to indicate it has been spent
+    /// (This is done to prevent future checks for smaller arguments from
+    /// succeeding even though a previous larger argument has already
+    /// failed).
+    fn tx_payload_size_error(&mut self) -> ServerError {
+        self.tx_payload_budget = 0;
```
tomxey: I think this is a bit unexpected effect for a function that generates an error message.
Could the caller handle it? It's already manipulating the value of `tx_payload_budget` so it wouldnt be strange if it would also assign `tx_payload_budget=0`

crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -1168,4 +1174,794 @@ pub mod tests {
         let resp = reqwest::get(&url_with_param).await.unwrap();
         assert_eq!(resp.status(), reqwest::StatusCode::GATEWAY_TIMEOUT);
     }
+
+    /// Execute a GraphQL request with `limits` in place, expecting an error to
+    /// be returned. Returns the list of errors returned.
```
tomxey: nit: the return value is a string, so I think we cannot call it a "list" here

crates/iota-graphql-rpc/tests/e2e_tests.rs
```diff
@@ -1004,12 +1033,192 @@ mod tests {
...
         cluster
             .wait_for_checkpoint_catchup(0, Duration::from_secs(10))
             .await;
         test_health_check_impl().await;
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_payload_total_exceeded() {
+        test_payload_total_exceeded_impl().await;
```
tomxey, line 1052: we have a lot of those payload size tests, what would you say about moving them to some dedicated file in `iota-graphql-rpc/tests/` ?

crates/iota-graphql-rpc/src/extensions/query_limits_checker.rs
```diff
@@ -258,6 +405,17 @@ impl<'a> LimitsTraversal<'a> {
...
+        self.reporter
+            .payload_size_error("Transaction payload too large")
```
tomxey, line 413: Could we also say what was the size of transaction payload? So the user knows how much he exceeded the value.

crates/iota-graphql-rpc/tests/e2e_tests.rs
```diff
@@ -1004,12 +1033,192 @@ mod tests {
...
 
         cluster
             .wait_for_checkpoint_catchup(0, Duration::from_secs(10))
             .await;
         test_health_check_impl().await;
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_payload_total_exceeded() {
```
tomxey, line 1051: Could we also test the edge case where user submits the maximum transaction size allowed by the protocol?

crates/iota-graphql-rpc/src/config.rs
```diff
@@ -277,7 +282,23 @@ impl ServiceConfig {
...
+    /// (including the signatures) allowed by the protocol, plus the Base64
+    /// overhead (roughly 1/3 of the original string).
+    async fn max_transaction_payload_size(&self) -> u32 {
+        self.limits.max_tx_payload_size
+    }
+
+    /// The maximum bytes allowed for the read part of GraphQL queries.
+    ///
+    /// In case of mutations or `dryRunTransactionBlocks` the `txBytes` and
+    /// signatures are not included in this limit.
```
tomxey: nit: `signatures`

crates/iota-graphql-rpc/tests/e2e_tests.rs
kodemartin: What would be the added value? This is a configuration parameter and we thoroughly test queries that exceed it.

crates/iota-graphql-rpc/src/extensions/query_limits_checker.rs
kodemartin: No we can't. This is lazily evaluated, so we just know it exceeded the limit set by the service. Can be more explitit about it here though.

crates/iota-graphql-rpc/tests/e2e_tests.rs
kodemartin: I think this warrants for a more general restructuring, distinguishing between different scopes of the `e2e_tests`.

For example the payload tests go together with the query tests.

Will create a new issue.

kodemartin: https://github.com/iotaledger/iota/issues/9196

crates/iota-graphql-rpc/src/extensions/query_limits_checker.rs
kodemartin: 9bafa6942b

crates/iota-graphql-rpc/src/server/builder.rs
kodemartin: 9bafa6942b

crates/iota-graphql-rpc/src/extensions/query_limits_checker.rs
kodemartin: 9bafa6942b

crates/iota-graphql-rpc/src/config.rs
kodemartin: 9bafa6942b

crates/iota-graphql-rpc/tests/e2e_tests.rs
kodemartin: The configuration parameter is by default set in a higher value than the expected max protocol size plus the Base64 overhead. This should give enough leeway to account for additional syntactic checks we employ here. These syntactic checks cause an intrinsic tolerance with respect to the exact max protocol value, but it is ok to be conservative here.

If it is found that a well-constructed transaction query that is exactly the max protocol size is cut off here we can revisit.

tomxey (approved):
lgtm

sergiupopescu199 (approved):
LGTM 🌸

- We need to update the `.snap` file.
- Tests error messages does not match making the assert fail
```rust
/// Error returned if transaction payloads exceed limit.
    fn tx_payload_size_error(&mut self) -> ServerError {
        self.reporter.payload_size_error(
            format!(
                "Transaction payload too large: exceeds {}",
                self.reporter.limits.max_tx_payload_size
            )
            .as_str(),
        )
    }
```
They all expest this `exceeds {}` but is not present"

```
 thread 'tests::test_payload_dry_run_exceeded' panicked at /actions-runner/_work/iota/iota/crates/iota-graphql-rpc/src/server/builder.rs:1264:9:
    assertion `left == right` failed
      left: "Transaction payload too large: exceeds 10. Requests are limited to 10 bytes or fewer on transaction payloads (all inputs to executeTransactionBlock or dryRunTransactionBlock) and the rest of the request (the query part) must be 400 bytes or fewer."
     right: "Transaction payload too large. Requests are limited to 10 bytes or fewer on transaction payloads (all inputs to executeTransactionBlock or dryRunTransactionBlock) and the rest of the request (the query part) must be 400 bytes or fewer."
```

PR #9175, opened 2025-11-06
feat(iota-core, iota-types): Introduce misbehavior reports

oliviasaa:
```mermaid
---
config:
  gitGraph:
    {parallelCommits: true,
    mainBranchName: 'Feature',
    rotateCommitLabel: true}
---
gitGraph
   commit id: "Initial commit"   
   checkout Feature
    branch misbehavior-reports
   commit id: "add MisbehaviorReports"
    checkout misbehavior-reports
    branch scoring-metrics
   commit id: "add ScoringMetrics"
    checkout scoring-metrics
   branch scorer
   commit id: "add Scorer"
    checkout scorer
   branch scoring-metrics-store
   commit id: "add ScoringMetricsStore"
branch report-validation
commit id: "add report validation" 
branch scoring-function
branch report-creation
branch advance-epoch
commit id: "add advance epoch v4" 
   checkout Feature
    merge misbehavior-reports tag: "THIS PR"
```

This PR introduces a new variant of `ConsensusTransactionKind` called `MisbehaviourReport`. It should be used to propagate a `VersionedMisbehaviorReport` (also introduced in this PR), which can contain both proofs of validator misbehavior and local views of performance metrics. 

### Release Notes

- [x] Protocol: Introduces a new variant of ConsensusTransactionKind to enable propagation of misbehavior reports
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -2673,6 +2673,19 @@ impl AuthorityPerEpochStore {
                 }
             }
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::MisbehaviorReport(authority, _, _),
+                ..
+            }) => {
+                if &transaction.sender_authority() != authority {
+                    warn!(
+                        "MisbehaviorReport authority {} does not match its author from consensus {}",
+                        authority, transaction.certificate_author_index
+                    );
+                    // Here we should update invalid report counts for the
+                    // authority
+                }
+            }
+            SequencedConsensusTransactionKind::External(ConsensusTransaction {
                 kind:
                     ConsensusTransactionKind::CapabilityNotificationV1(AuthorityCapabilitiesV1 {
                         authority,
@@ -3908,6 +3921,13 @@ impl AuthorityPerEpochStore {
             }) => {
                 // these are partitioned earlier
                 panic!("process_consensus_transaction called with end-of-publish transaction");
+            }
+            SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::MisbehaviorReport(_authority, _report, _),
+                ..
+            }) => {
+                // Validate report
+                Ok(ConsensusCertificateResult::ConsensusMessage)
             }
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
                 kind: ConsensusTransactionKind::CapabilityNotificationV1(capabilities),
--- a/crates/iota-types/src/messages_consensus.rs
+++ b/crates/iota-types/src/messages_consensus.rs
@@ -20,7 +20,8 @@ use shared_crypto::intent::IntentScope;
 
 use crate::{
     base_types::{
-        AuthorityName, ConciseableName, ObjectID, ObjectRef, SequenceNumber, TransactionDigest,
+        AuthorityName, CommitRound, ConciseableName, ObjectID, ObjectRef, SequenceNumber,
+        TransactionDigest,
     },
     crypto::{AuthoritySignature, DefaultHash},
     digests::{ConsensusCommitDigest, Digest},
@@ -90,6 +91,10 @@ pub enum ConsensusTransactionKey {
     NewJWKFetched(Box<(AuthorityName, JwkId, JWK)>),
     RandomnessDkgMessage(AuthorityName),
     RandomnessDkgConfirmation(AuthorityName),
+    // If a validator submits more than one report for the same round, we update the
+    // scoring metrics using the maximum reported metric, so CommitRound is sufficient to
+    // identify a report from a single AuthorityName.
+    MisbehaviorReport(AuthorityName, CommitRound),
     // New entries should be added at the end to preserve serialization compatibility. DO NOT
     // CHANGE THE ORDER OF EXISTING ENTRIES!
 }
@@ -102,6 +107,9 @@ impl Debug for ConsensusTransactionKey {
                 write!(f, "CheckpointSignature({:?}, {:?})", name.concise(), seq)
             }
             Self::EndOfPublish(name) => write!(f, "EndOfPublish({:?})", name.concise()),
+            Self::MisbehaviorReport(name, round) => {
+                write!(f, "MisbehaviorReport({:?},{:?})", name.concise(), round)
+            }
             Self::CapabilityNotification(name, generation) => write!(
                 f,
                 "CapabilityNotification({:?}, {:?})",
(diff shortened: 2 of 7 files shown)


crates/iota-types/src/messages_consensus.rs
```diff
@@ -90,6 +91,11 @@ pub enum ConsensusTransactionKey {
     NewJWKFetched(Box<(AuthorityName, JwkId, JWK)>),
     RandomnessDkgMessage(AuthorityName),
     RandomnessDkgConfirmation(AuthorityName),
+    // We current do not care about equivocations, so CommitRound is sufficient to identify a
+    // report from a single AuthorityName. If a validator equivocates and for some reason we end
+    // up with more than one report for the same round, we update the ScoringMetrics using the
+    // maximum reported metric.
```
cyberphysic4l: change terminology from equivocations. Just say submits more than one report per commit round.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -274,6 +284,34 @@ impl ConsensusTransactionKind {
...
+impl VersionedReport {
+    pub fn is_valid(&self) -> bool {
+        match self {
+            VersionedReport::V1(report) => report.verify(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ReportV1 {
```
cyberphysic4l: Need some comments here on the purpose of V1. My understanding is that is should be a basic report with counts only as we will not have proofs for provable misbehaviors.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -274,6 +284,34 @@ impl ConsensusTransactionKind {
...
+    pub faulty_blocks_provable: Vec<u64>,
+    pub faulty_blocks_unprovable: Vec<u64>,
+    pub equivocations: Vec<u64>,
+    pub missing_proposals: Vec<u64>,
+}
+
+impl ReportV1 {
+    pub fn verify(&self) -> bool {
```
cyberphysic4l: Should probably define this method for the enum VersionedReport and then match inside that method.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -464,6 +502,22 @@ impl ConsensusTransaction {
         }
     }
 
+    pub fn new_misbehaviour_report_message_v1(authority: AuthorityName, report: &ReportV1) -> Self {
```
cyberphysic4l:
```suggestion
    pub fn new_misbehaviour_report_v1(authority: AuthorityName, report: &ReportV1) -> Self {
```

crates/iota-types/src/messages_consensus.rs
cyberphysic4l:
```suggestion
    // If a validator submits more than one report for the same round, we update the
    // ScoringMetrics using the maximum reported metric, so CommitRound is sufficient to
    // identify a report from a single AuthorityName.
```

crates/iota-types/src/messages_consensus.rs
```diff
@@ -274,6 +284,34 @@ impl ConsensusTransactionKind {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum VersionedReport {
```
cyberphysic4l:
```suggestion
pub enum VersionedMisbehaviorReport {
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2672,6 +2672,18 @@ impl AuthorityPerEpochStore {
...
+            SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::MisbehaviourReport(authority, _, _),
+                ..
+            }) => {
+                if &transaction.sender_authority() != authority {
+                    warn!(
+                        "MisbehaviourReport authority {} does not match its author from consensus {}",
+                        authority, transaction.certificate_author_index
+                    );
+                    // Update invalid reports
```
cyberphysic4l: Should the verify function be called here too?

crates/iota-types/src/messages_consensus.rs
```diff
@@ -274,6 +284,34 @@ impl ConsensusTransactionKind {
...
+}
+
+impl VersionedReport {
+    pub fn is_valid(&self) -> bool {
+        match self {
+            VersionedReport::V1(report) => report.verify(),
+        }
+    }
+}
+
```
cyberphysic4l, line 298:
```suggestion

// MisbehaviorReportV1 contains lists of faulty blocks, equivocation and missing
// proposal counts for each authority. This first version only counts for the
// misbehaviors but no evidence is included.
```

cyberphysic4l (commented):
We should also add comments to the `MisbehaviorReport` struct and the block field to say that this form of misbehvior report will be deprecated now and avoid confusion. https://github.com/iotaledger/iota/blob/0ef0bb741e428fad90c4930209e945b545a8521a/consensus/core/src/block.rs#L630

crates/iota-core/src/authority/authority_per_epoch_store.rs
oliviasaa: yes, we verify before updating the invalid report count. I improved the comment a bit.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -274,6 +283,43 @@ impl ConsensusTransactionKind {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum VersionedMisbehaviorReport {
```
piotrm50, line 287: Do we need this at this moment? I think it's better to introduce the verification layer only when it's necessary.

crates/iota-types/src/messages_consensus.rs
piotrm50: If we do a match then this method is not even necessary - we can always assume that verification is correct without having additional method call that returns `true`. Although I think that the verification on the `VersionedMisbehaviorReport` is not needed yet either since we don't have a version that needs to be verified at this moment.

piotrm50 (changes requested):
Choose one and consistent spelling of `Misbehaviour` or `Misbehavior`.

crates/iota-types/src/messages_consensus.rs
oliviasaa: There will be a PR built on top of this branch with the full system (meaning that those reports are supposed to be sent and verified already in this feature branch). I can move the verification to this other PR, though.

piotrm50 (approved):
LGTM

crates/iota-types/src/messages_consensus.rs
```diff
@@ -464,6 +517,25 @@ impl ConsensusTransaction {
...
+            bcs::to_bytes(report).expect("report serialization should not fail");
+        let mut hasher = DefaultHasher::new();
+        serialized_report.hash(&mut hasher);
+        let tracking_id = hasher.finish().to_le_bytes();
+        Self {
+            tracking_id,
+            kind: ConsensusTransactionKind::MisbehaviorReport(
+                authority,
+                VersionedMisbehaviorReport::V1(report.clone()),
+                1,
```
piotrm50: CommitRound is hardcoded

oliviasaa: that's really odd, I'll fix it right now

jkrvivian (approved):

semenov-vladyslav (approved):
Would you please elaborate on the release notes in PR description. The current "Introduces misbehavior report type" might not be very user-friendly. Should it be a Protocol change rather that Nodes?

PR #9176, opened 2025-11-06
feat(iota-types): Add scoring metrics type

oliviasaa:
```mermaid
---
config:
  gitGraph:
    {parallelCommits: true,
    mainBranchName: 'Feature',
    rotateCommitLabel: true}
---
gitGraph
   commit id: "Initial commit"
   commit id: "add MisbehaviorReports" tag: "PR9175"
   checkout Feature
    branch scoring-metrics
   commit id: "add ScoringMetrics"
    checkout scoring-metrics
   branch scorer
   commit id: "add Scorer"
    checkout scorer
   branch scoring-metrics-store
   commit id: "add ScoringMetricsStore"
branch report-validation
commit id: "add report validation" 
branch scoring-function
branch report-creation
branch advance-epoch
commit id: "add advance epoch v4" 
 checkout Feature
    merge scoring-metrics tag: "THIS PR"

```
# Description of changes

This PR adds a new `VersionedScoringMetrics` type. `ScoringMetricsV1` is currently an atomic version of `MisbehaviorReportV1`, even though this is expected to change in their future versions. `VersionedScoringMetrics ` should be used to store misbehavior metrics about the rest of the committee and collected in the node, but also to store the information received in the misbehavior reports received from other authorities.

### Release Notes

- [x] Protocol: Introduces the VersionedScoringMetrics type, used to store misbehavior metrics about all authorities
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-types/src/scoring_metrics.rs
+++ b/crates/iota-types/src/scoring_metrics.rs
@@ -1,0 +1,291 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use iota_protocol_config::ProtocolConfig;
+
+use crate::messages_consensus::{MisbehaviorReportV1, VersionedMisbehaviorReport};
+
+// This struct represents the scoring metrics collected by all authorities. They
+// are stored locally by each authority and then converted to a misbehavior
+// report when they share their metrics with the network. When a report is
+// received, it is also used to update a variable of this type stored in the
+// Scorer. Any metric contained in this struct must be guaranteed to be
+// monotonically increasing, because of the way updates are applied from
+// reports.
+pub enum VersionedScoringMetrics {
+    V1(ScoringMetricsV1),
+}
+
+// Basic getters, setters and increments for the metrics.
+impl VersionedScoringMetrics {
+    pub fn new(committee_size: usize, protocol_config: &ProtocolConfig) -> Self {
+        match protocol_config.scorer_version_as_option() {
+            None | Some(1) => VersionedScoringMetrics::V1(ScoringMetricsV1::new(committee_size)),
+            _ => panic!("Unsupported scorer version"),
+        }
+    }
+
+    pub fn increment_faulty_blocks_provable(&self, authority_index: usize, increment: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.faulty_blocks_provable[authority_index]
+                    .fetch_add(increment, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn increment_faulty_blocks_unprovable(&self, authority_index: usize, increment: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.faulty_blocks_unprovable[authority_index]
+                    .fetch_add(increment, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn increment_equivocations(&self, authority_index: usize, increment: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.equivocations[authority_index].fetch_add(increment, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn increment_missing_proposals(&self, authority_index: usize, increment: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.missing_proposals[authority_index].fetch_add(increment, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn store_faulty_blocks_provable(&self, authority_index: usize, value: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.faulty_blocks_provable[authority_index].store(value, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn store_faulty_blocks_unprovable(&self, authority_index: usize, value: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.faulty_blocks_unprovable[authority_index].store(value, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn store_equivocations(&self, authority_index: usize, value: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.equivocations[authority_index].store(value, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn store_missing_proposals(&self, authority_index: usize, value: u64) {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => {
+                metrics.missing_proposals[authority_index].store(value, Ordering::Relaxed);
+            }
+        }
+    }
+
+    pub fn load_faulty_blocks_provable(&self) -> Vec<u64> {
+        match self {
+            VersionedScoringMetrics::V1(metrics) => metrics
+                .faulty_blocks_provable
+                .iter()
+                .map(|metric| metric.load(Ordering::Relaxed))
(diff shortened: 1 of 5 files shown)


crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,258 @@
...
+
+use iota_protocol_config::ProtocolConfig;
+
+use crate::messages_consensus::{ReportV1, VersionedReport};
+
+pub enum ScoringMetrics {
+    V1(ScoringMetricsV1),
+}
+
+pub struct ScoringMetricsV1 {
```
cyberphysic4l: Comment needed explaining what the intended usage of this is.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,258 @@
...
+                for (i, value) in report_v1.equivocations.iter().enumerate() {
+                    metrics.equivocations[i].store(*value, Ordering::Relaxed);
+                }
+                for (i, value) in report_v1.missing_proposals.iter().enumerate() {
+                    metrics.missing_proposals[i].store(*value, Ordering::Relaxed);
+                }
+                ScoringMetrics::V1(metrics)
+            }
+        }
+    }
```
cyberphysic4l: Add comment to these functions indicating that they are used to convert reports received from other authorities into `ScoringMetrics` type and then used to update the `Scorer` component.

crates/iota-types/src/scoring_metrics.rs
cyberphysic4l: If I understand correctly, this structure represent the scoring metrics reported by a single authority. They are stored locally by each authority and then converted to a misbehavior report when they want to share their metrics with the network. When a report is received, it should be converted to this type and stored in the `Scorer`.

cyberphysic4l: Perhaps a name to indicate that this represents a single authority viewpoint, such as `LocalScoringMetrics` would be helpful.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    pub fn missing_proposals_u64(&self) -> Vec<u64> {
+        match self {
+            ScoringMetrics::V1(metrics) => metrics
+                .missing_proposals
+                .iter()
+                .map(|metric| metric.load(Ordering::Relaxed))
+                .collect(),
+        }
+    }
```
cyberphysic4l: I suggest to rename these from `<metric>_u64` to `load_<metric>`.

crates/iota-types/src/scoring_metrics.rs
oliviasaa: I kept the name of the structure as it is, because it is not only used for local metrics counts, it is something more generic. But I agree that there is room for improvement in all these names, so I'm adding some aliases so anything involving those types are easier to understand.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,294 @@
...
+// reports.
+pub enum VersionedScoringMetrics {
+    V1(ScoringMetricsV1),
+}
+
+// Basic getters, setters and increments for the metrics.
+impl VersionedScoringMetrics {
+    pub fn new(committee_size: usize, protocol_config: &ProtocolConfig) -> Self {
+        match protocol_config.scorer_version_as_option() {
+            None | Some(1) => VersionedScoringMetrics::V1(ScoringMetricsV1::new(committee_size)),
```
piotrm50, line 25: Do we want to create a scorer when the config is set to None though? The comment in the protocol config says that they are not sent nor considered valid. Is it only for local testing purposes then? If so then please add a comment.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,294 @@
...
+    // increasing. Example: number of faulty blocks detected for a given authority
+    // is monotonically increasing by design; average faulty blocks per minute is
+    // not.
+    pub fn update_from_report(&self, report: &VersionedMisbehaviorReport) {
+        match (self, report) {
+            (VersionedScoringMetrics::V1(metrics), VersionedMisbehaviorReport::V1(report_v1)) => {
+                for (i, value) in report_v1.faulty_blocks_provable.iter().enumerate() {
+                    if *value > metrics.faulty_blocks_provable[i].load(Ordering::Relaxed) {
+                        metrics.faulty_blocks_provable[i].store(*value, Ordering::Relaxed);
+                    }
```
piotrm50: Loading and storing in separate method calls can lead to race conditions when some other task runs this method concurrently - say metric is set to 10, then task A wants to update it to 15, checks that 15 > 10, then task B on another thread wants to do the same, checks that 20 > 10 and then is the first to update the metric value to 20, then task A updates the value to 15, thus overwriting the higher value.

piotrm50 (changes requested):

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    // Given a VersionedMisbehaviorReport, create a VersionedScoringMetrics struct
+    // with the same values. Used when an authority receives a report from the
+    // network and needs to create a local copy of the metrics contained in it.
+    pub fn from_report(report: &VersionedMisbehaviorReport) -> Self {
+        match report {
+            VersionedMisbehaviorReport::V1(report_v1) => {
+                let committee_size = report_v1.faulty_blocks_provable.len();
```
piotrm50, line 214: is the report validated anywhere that it the number of elements matches the committee size?

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -0,0 +1,294 @@
...
+                    equivocations,
+                    missing_proposals,
+                })
+            }
+        }
+    }
+}
+
+pub struct ScoringMetricsV1 {
+    pub faulty_blocks_provable: Vec<AtomicU64>,
```
piotrm50: Provide description of each field as inferring the meaning from the name might be confusing and not obvious.

piotrm50: Another thing is that the fields are publicly accessible - is that necessary? `VersionedScoringMetrics` provides getters and setters of those fields for across different versions already.

crates/iota-types/src/scoring_metrics.rs
oliviasaa: oh, of course `MisbehaviorReportV1::verify` should not always return `true`. I'll add the verification.

oliviasaa: added to https://github.com/iotaledger/iota/pull/9175

crates/iota-types/src/scoring_metrics.rs
oliviasaa: It is for local testing, to be able to expose metrics relative to the scores themselves (whose logic is supposed to be moved from consensus to the iota-core crate, while consensus would only collect metrics), but also because the Scorer is also a field in the authority per epoch store, and I didn't want to create different versions for it (or use `Option<Scorer>`).

crates/iota-types/src/scoring_metrics.rs
oliviasaa: no, there is no need for that. I just changed it.

piotrm50 (approved):

piotrm50 (approved):

crates/iota-types/src/scoring_metrics.rs
oliviasaa: changed to fetch_max

semenov-vladyslav (approved):
Please elaborate on release notes in PR description. I think it should be Protocol rather than Nodes.

PR #9179, opened 2025-11-06
feat(docker): enable BuildKit and optimize build process with cache mounts

piotrm50:
# Description of change

This PR optimizes the Docker build process by enabling BuildKit and implementing cache mounts across all Dockerfiles in the repository.

The changes improve build performance by leveraging BuildKit's cache mount feature to persist:
- Cargo package registry (`/usr/local/cargo/registry`)
- Git dependencies (`/usr/local/cargo/git`)
- Compiled artifacts (`/iota/target` with locked sharing)

All Dockerfiles now declare BuildKit syntax support and use optimized RUN commands that leverage these cache mounts, significantly reducing rebuild times for iterative development.

N/A

### DIFF
--- a/docker/iota-data-ingestion/Dockerfile
+++ b/docker/iota-data-ingestion/Dockerfile
@@ -29,11 +29,13 @@ COPY external-crates external-crates
 COPY iota-execution iota-execution
 COPY Cargo.toml Cargo.lock ./
 
-# Build and copy the built binaries to the working directory
+# Build the binaries
 RUN cargo build --profile ${PROFILE} \
       --bin iota-data-ingestion \
-      --features ${CARGO_BUILD_FEATURES:=default} && \
-    cp $TARGET_FOLDER/iota-data-ingestion ./iota-data-ingestion
+      --features ${CARGO_BUILD_FEATURES:=default}
+
+# Copy the built binaries to the working directory
+RUN cp $TARGET_FOLDER/iota-data-ingestion ./iota-data-ingestion
 
 # Production image
 FROM debian:bookworm-slim AS runtime
--- a/docker/iota-data-ingestion/Dockerfile
+++ b/docker/iota-data-ingestion/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 # Build image (the specific rust version can also be passed, e.g. "1.82-bookworm")
 ARG RUST_IMAGE_VERSION=bookworm
 FROM rust:${RUST_IMAGE_VERSION} AS builder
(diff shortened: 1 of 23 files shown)


alexsporn:
I am not a friend of this tbh, I don't want the Dockerfiles we use for releasing the builds to have workaround for quicker rebuilds, etc. They should be as clean as possible.
Can we work around this in the build-scripts instead?

polinikita (approved):

muXxer:
@piotrm50 I'm with Alex here. Can we solve that differently?

muXxer:
@piotrm50 sorry, I reverted your changes in the Dockerfiles, and moved the logic to the `build-script.sh`. Does that work for you? Only the `DOCKER_BUILDKIT` needs to be set.

For example:
```bash
cd docker/iota-node/
DOCKER_BUILDKIT=1 ./build.sh
```

muXxer (approved):

muXxer:
@alexsporn can you have a look again?

piotrm50:
Good idea with changing that dynamically in a temp file :heavy_check_mark:

alexsporn (approved):

docker/iota-data-ingestion/Dockerfile
kodemartin, line 1: What was wrong with the previous version? Now in order to build the image in `debug` or any other non-release mode we have to override two build args.

docker/iota-data-ingestion/build.sh
```diff
@@ -7,4 +7,4 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd -P)"
 # The name of last directory in script's path is used for tag
 BASE_NAME="$(basename "$SCRIPT_DIR")"
 
-"$SCRIPT_DIR/../utils/build-script.sh" --image-tag "iotaledger/$BASE_NAME"
+DOCKER_BUILDKIT=1 "$SCRIPT_DIR/../utils/build-script.sh" --image-tag "iotaledger/$BASE_NAME"
```
kodemartin, line 10: Isn't BUILDKIT enabled by default? https://docs.docker.com/build/buildkit/

kodemartin (approved):
@muXxer filled me in on the motivation. lgtm

docker/iota-data-ingestion/build.sh
muXxer: For the context: We don't need it, we could give it another name, but this way it also works for users with older docker versions.

docker/iota-data-ingestion/Dockerfile
muXxer: For the context: Unfortunately rust builds always land in `target/{profile}` except for the "dev" profile, so we need a way to specify the folder.

PR #9180, opened 2025-11-06
feat(starfish): fetch and verify transactions during commit syncing

### LINKED ISSUE #8376: Decide on fetching strategy when fetched blocks have missing commited transactions
In `CommitSyncer` method `handle_fetch_result`.

### LINKED ISSUE #9163: Fetch transactions during commit syncing
During CommitSyncing we currently rely on TransactionsSynchronizer, which is tuned to be quickly fetch and react to recent missing committed transactions and not to handle large transaction fetches that need to be fetched when the node is lagging behind a lot or syncing from scratch.

piotrm50:
# Description of change

This PR adds proactive transaction fetching during commit syncing to improve synchronization for nodes that are far behind or syncing from scratch.

Previously, the consensus layer relied on `TransactionsSynchronizer` for fetching missing transactions, which is tuned for filling recent small gaps. When a node is far behind, it needs to fetch large amounts of transactions, which the current synchronizer is not optimized for.

During commit syncing, the `CommitSyncer` now:
1. Identifies committed transaction block references from commits being synced
2. Fetches transaction data directly from the peer authority alongside block headers
3. Verifies transactions against their block headers to ensure integrity
4. Adds verified transactions to the DAG before processing commits

Key changes:
- **commit.rs**: Extended `CertifiedCommit` to include verified transactions alongside block headers
- **commit_syncer.rs**: Added logic to identify and collect transaction block references, implemented concurrent fetching with staggered timing, and added transaction verification
- **core.rs**: Integrated transaction extraction from certified commits before processing
- **test_dag_builder.rs**: Updated test utilities to support transaction data

Resolves #9163

### DIFF
--- a/crates/starfish/core/src/commit_syncer.rs
+++ b/crates/starfish/core/src/commit_syncer.rs
@@ -68,7 +68,6 @@ use crate::{
     error::{ConsensusError, ConsensusResult},
     network::{NetworkClient, SerializedTransactions},
     stake_aggregator::{QuorumThreshold, StakeAggregator},
-    transactions_synchronizer::TransactionsSynchronizerHandle,
 };
 
 // Handle to stop the CommitSyncer loop.
@@ -119,7 +118,6 @@ impl<C: NetworkClient> CommitSyncer<C> {
         context: Arc<Context>,
         core_thread_dispatcher: Arc<dyn CoreThreadDispatcher>,
         commit_vote_monitor: Arc<CommitVoteMonitor>,
-        transactions_synchronizer: Arc<TransactionsSynchronizerHandle>,
         commit_consumer_monitor: Arc<CommitConsumerMonitor>,
         network_client: Arc<C>,
         block_verifier: Arc<dyn BlockVerifier>,
@@ -130,7 +128,6 @@ impl<C: NetworkClient> CommitSyncer<C> {
             core_thread_dispatcher,
             commit_vote_monitor,
             commit_consumer_monitor,
-            transactions_synchronizer,
             network_client,
             block_verifier,
             dag_state,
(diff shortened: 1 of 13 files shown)


crates/starfish/core/src/core.rs
```diff
@@ -483,6 +483,22 @@ impl Core {
...
+        // available when creating new commits.
+        let all_transactions: Vec<VerifiedTransactions> = certified_commits
+            .commits()
+            .iter()
+            .flat_map(|commit| commit.transactions())
+            .cloned()
+            .collect();
+
+        if !all_transactions.is_empty() {
+            self.add_transactions(all_transactions, TransactionSource::TransactionSynchronizer)?;
```
VorobyevIlya: Should source be commit-syncer?

crates/starfish/core/src/core.rs
```diff
@@ -483,6 +483,22 @@ impl Core {
         BTreeMap<BlockRef, BTreeSet<AuthorityIndex>>,
     )> {
         let _scope = monitored_scope("Core::add_certified_commits");
+
+        // First, collect and add all transactions from certified commits.
+        // Transactions must be added before processing commits to ensure they are
+        // available when creating new commits.
```
VorobyevIlya, line 489: Does it make sense to create blocks and use add_blocks function for those blocks for which we have both transactions and header here?

piotrm50: Yes, because those headers will not necessarily be accepted - in case when transaction is committed but the header is not. If we do it separately, then transactions are added regardless of whether we have the header in the DAGState or not. But it's actually a good point - perhaps we should skip adding non-committed headers to the Core, since we only needed them to verify that the transactions are correct. 

I'll turn the PR into a draft until I fix this and some other issues that I found in the meantime.

crates/starfish/core/src/core.rs
piotrm50: fixed

crates/starfish/core/src/commit_syncer.rs
```diff
@@ -671,15 +733,119 @@ impl<C: NetworkClient> CommitSyncer<C> {
...
+                        }
+                        Ok::<BTreeMap<BlockRef, Bytes>, ConsensusError>(result)
+                    }
+                })
+                .collect()
+        } else {
+            FuturesOrdered::new()
+        };
+
+        // 8. Process header and transaction requests concurrently
```
VorobyevIlya: wrong numbering, there were 8,9,10 above

VorobyevIlya (approved):

filipdulic (approved):

polinikita (approved):

PR #9181, opened 2025-11-06
feat: implement a template of authenticator info checking during attaching with a proof

### LINKED ISSUE #8903: [AA] Check account type during iota_framework::account::attach_auth_info_v1()
The function should check that the object to which the auth info is being attached is the same type as the type declared as first parameter of the authenticate function being attached.

valeriyr:
[run-ci]

# Description of change

This PR contains two parts:
1. `AuthenticatorInfo` was extended with a generic parameter `Account` which indicates the account type that can be unlocked with the related authenticator function.
2. Introduced `AuthenticatorInfoV1CompatibilityProof` that represents a proof that `AuthenticatorInfo` can be attached to an account. This can be considered as a temporal solution since we have a better idea how to proof `AuthenticatorInfo` and an account compatibility.
 
fixes #8903


iota-execution/latest/iota-verifier/src/account_auth_verifier.rs
```diff
@@ -38,6 +39,7 @@ use crate::verification_failure;
 pub fn verify_authenticate_func(
     module: &CompiledModule,
     function_identifier: Identifier,
+    account_type: &TypeTag,
```
miker83z, line 42: This information will be found in the package metadata after #9198. But it is fine to temporarily add it here to make it work.

miker83z (approved):
LGTM!

Dkwcs (approved):

PR #9183, opened 2025-11-06
fix(consensus): track failures to choose optimal peers and adjust parameters in the Starfish transaction synchronizer

### LINKED ISSUE #9162: Tune TransactionSynchronizer to not choke easily
Currently Transaction Synchronizer is saturated quite easily, as it runs only 1 live fetching task and 1 pending task. We need to tune parameters and remove some processing timeouts so that it runs more smoothly.

polinikita:
# Description of change

Transaction synchronizer fetches transactions either slowly or with some warning in many scenarios:
-  after commit syncer has synced a big number of certified commits, the transactions from the committed subDAGs are fetched quite slowly.
- when some connections are blocked, the transaction synchronizer produces too many failures with blocked peers.

In this PR, 
- we tune parameters of the transaction synchronizer to increase the number of concurrent high-level task and control the total number of spawned tasks.
- we introduce a simple memory tracking when the transaction fetch has failed for the last time for each given peer. This memory is then used to take proper peers for fetching
- we remove transaction verification from the transaction synchronizer (similarly to how it was done in the commit syncer) since the verification and validity of transactions was already performed by a quorum of the network.

Fixes #9162 

We have tested the change using the fuzz testing scenarios.

### DIFF
--- a/crates/starfish/core/src/transactions_synchronizer.rs
+++ b/crates/starfish/core/src/transactions_synchronizer.rs
@@ -92,12 +92,15 @@ impl SyncMethod {
 /// fetching from peers.
 struct LastFailureByPeer {
     inner: Mutex<Vec<Option<Instant>>>,
+    context: Context,
 }
 
 impl LastFailureByPeer {
-    fn new(committee_size: usize) -> Arc<Self> {
+    fn new(context: &Context) -> Arc<Self> {
+        let committee_size = context.committee.size();
         Arc::new(Self {
             inner: Mutex::new(vec![None; committee_size]),
+            context: context.clone(),
         })
     }
     fn update_with_new_instant(self: &Arc<Self>, peer: AuthorityIndex, new_instant: Instant) {
@@ -107,10 +110,7 @@ impl LastFailureByPeer {
 
     /// Determine which authorities are less reliable to fetch transactions.
     /// Returns less than f+1 authorities by stake.
-    fn get_excluded_authorities_by_stake(
-        self: &Arc<Self>,
-        context: &Context,
-    ) -> BTreeSet<AuthorityIndex> {
+    fn get_excluded_authorities_by_stake(self: &Arc<Self>) -> BTreeSet<AuthorityIndex> {
         let last_round_by_peer = { self.inner.lock().clone() };
 
         let mut indexed_rounds: Vec<(AuthorityIndex, Instant)> = last_round_by_peer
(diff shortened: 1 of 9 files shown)


crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -649,81 +740,77 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher>
                 // Stable order for tests
                 Box::new(blocks_by_authority.into_iter())
             } else {
-                let mut vec: Vec<_> = blocks_by_authority.into_iter().collect();
+                let excluded_authorities = last_failure_by_peer.get_excluded_authorities();
+                // Exclude authorities with latest recorded failures
+                let mut vec: Vec<_> = blocks_by_authority
+                    .into_iter()
+                    .filter(|(authority, _)| !excluded_authorities.contains(authority))
+                    .collect();
```
VorobyevIlya: Do I understand correctly that we exclude 2/3 authorities with latest failures, and then filter out all these authorities from blocks_by_authority? 
But in blocks_by_authority we don't have all authorities, is it possible that we will exclude too many? Maybe it is better to exclude 2/3 of authorities from blocks_by_authority?

piotrm50 (approved):

crates/starfish/core/src/transactions_synchronizer.rs
NaitsabesMue: Also if we filter out 2/3, and keep only 1/3, this 1/3 might just be those who have not the transactions. If we filter out only 1/3, we have at least 1/3 that have the transactions.

NaitsabesMue (changes requested):

crates/starfish/core/src/transactions_synchronizer.rs
NaitsabesMue: More general, if you exclude 2/3 you may end with only one honest. Since we do not weight by stake, the risk increases.

NaitsabesMue:
The name of the PR is not optimal as main contribution is about "memory tracking" when the transaction fetch has failed.

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: I tried some scenarios when 1/3 was no sufficient for good effect. But in general I agree with arguments. For safety reasons, I changed to return less than 1/3 by stake so the intersection is more than 1/3 by stake.

VorobyevIlya (approved):

NaitsabesMue (approved):

PR #9188, opened 2025-11-06
feat(compiler): add authenticator attribute

### LINKED ISSUE #8860: [AA] Add an `#[authenticator]` attribute to the language

miker83z:
# Description of change

To support defining authenticator functions for the IOTA abstract account workflow the user must be able to
specify in Move code which functions should be considered "authenticators".
Moreover in the future there may be more than one version of authenticators so a version number must be supported as well.

This PR provides three formats to define a move authenticator with version.
As a Named attribute:
```
#[authenticator]
public fun authenticator_fun(...)
```
In this case the version number is inferred to be 1.

As an Assigned attribute:
```
#[authenticator = 1]
public fun authenticator_fun(...)
```
Where the version number is assigned explicitly.

And as a Parametrized attribute:
```
#[authenticator(version = 1)]
public fun authenticator_fun(...)
```
In this format the user must explicitly spell out the version number and assign a value to it. While the format is verbose, it may become a preferred variant as it could potentially be extended with additional features like:
```
#[authenticator(version = 1,  something_else = 53)]
```

## Implementation notes
Attributes do NOT reach the bytecode. Instead these are stored into the `FunctionInfo` and this passed upward. Later PRs will deal with handling the Authenticator attribute from within the `FunctionInfo` (in this PR, just `iota_types::move_package::FnInfo` was extended to support the authenticator version).

Fixes #8860 

### DIFF
--- a/external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
+++ b/external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
@@ -1,0 +1,121 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! IOTA Authenticator Attribute
+
+use std::{collections::BTreeSet, fmt};
+
+use move_core_types::u256::U256;
+use move_ir_types::location::Loc;
+use move_symbol_pool::Symbol;
+use once_cell::sync::Lazy;
+
+use crate::{
+    expansion::ast::{Attribute_, AttributeValue, Attributes},
+    shared::known_attributes::{
+        AttributePosition, FlavoredAttribute, KnownAttribute as MoveKnownAttribute,
+    },
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AuthenticatorAttribute;
+
+impl AuthenticatorAttribute {
+    pub const AUTHENTICATOR: &'static str = "authenticator";
+    pub const VERSION: &'static str = "version";
+
+    pub const fn name(&self) -> &'static str {
+        Self::AUTHENTICATOR
+    }
+
+    pub fn expected_positions(&self) -> &'static BTreeSet<AttributePosition> {
+        static AUTHENTICATOR_POSITIONS: Lazy<BTreeSet<AttributePosition>> =
+            Lazy::new(|| BTreeSet::from([AttributePosition::Function]));
+        &AUTHENTICATOR_POSITIONS
+    }
+}
+
+//**************************************************************************************************
+// Attribute_ implementation
+//**************************************************************************************************
+
+impl Attribute_ {
+    /// Parses the authenticator attribute and returns the version number.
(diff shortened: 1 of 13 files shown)


external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
```diff
@@ -0,0 +1,121 @@
...
+        AttributePosition, FlavoredAttribute, KnownAttribute as MoveKnownAttribute,
+    },
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AuthenticatorAttribute;
+
+impl AuthenticatorAttribute {
+    pub const AUTHENTICATOR: &'static str = "authenticator";
+    pub const VERSION: &'static str = "version";
```
Dkwcs, line 25: 1. How do we track the version of the attribute in general? When this information well be useful, for instance if method has version 5 what is the outcome?
2. We expect uniqueness per function as I remember.
Just to clarify is it intended to be unique per authenticator function only?
Like is this completely okay in scope of the one module/package?

   ```
    // Allowed
    #[authenticator(0)]
    public fun auth_v0(&signer: &signer, ctx: &mut TxContext): bool {
        true
    }

    // allowed
    #[authenticator(0)]
    public fun another_auth_v0(&signer: &signer, ctx: &mut TxContext): bool {
        true
    }
```

valeriyr (approved):

external-crates/move/crates/move-compiler/src/iota_mode/known_attributes/authenticator.rs
miker83z: @Dkwcs 
> How do we track the version of the attribute in general? When this information well be useful, for instance if method has version 5 what is the outcome?

In the normal case anyone would just use `authenticator` with no version. This will make it so that devs are always using the latest authenticator version if we modify the logic behind verification/execution. There can be a case in which you want to set a specific version, e.g. `authenticator(1)` when we already reached version 3; this is the case where the dev is fine with the authenticator version 1 because the dev is upgrading a package for a bug and the authentication logic shouldn't change.

> We expect uniqueness per function as I remember.
Just to clarify is it intended to be unique per authenticator function only?
Like is this completely okay in scope of the one module/package?

What do you mean in detail? I didn't get your question. However the example you make is legit (just, the signer doesn't exist).

Dkwcs (approved):

PR #9207, opened 2025-11-10
fix(iota): `iota move build --dump-bytecode-as-base64 --ignore-chain`

thibault-martinez:
# Description of change

This PR fixes a bug in the `iota move build --dump-bytecode-as-base64 --ignore-chain` as the previous code did not correctly account for `--ignore-chain` flag. The main issue is that when `--ignore-chain` is used, the tree shaking algorithm will not work anymore.

Based on https://github.com/MystenLabs/sui/pull/21856

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Fixes `iota move build --dump-bytecode-as-base64 --ignore-chain`
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-move/src/build.rs
```diff
@@ -26,6 +26,7 @@ pub struct Build {
     /// Whether we are printing in base64.
     #[arg(long, global = true)]
     pub dump_bytecode_as_base64: bool,
+    /// [Mainly for testing, not recommended for production]
```
DaughterOfMars, line 29: I don't know about this comment tbh

crates/iota/src/iota_commands.rs
```diff
@@ -523,28 +523,30 @@ impl IotaCommand {
...
                         } else {
-                            read_api.get_chain_identifier().await.ok()
+                            let config = client_config
+                                .unwrap_or(iota_config_dir()?.join(IOTA_CLIENT_CONFIG));
+                            prompt_if_no_config(&config, false, true, true)?;
+                            let context = WalletContext::new(&config, None, None)?;
+
+                            let Ok(client) = context.get_client().await else {
+                                bail!(
+                                    "`iota move build --dump-bytecode-as-base64` requires a connection to the network. Current active network is {} but failed to connect to it.",
```
DaughterOfMars, line 558: we don't want to do any of this stuff if the flag is set? Wouldn't it be better to just check the flag and skip the tree-shaking?

miker83z:
Is there a test or can we create one to check the behavior of this combination of flags?

PR #9159, opened 2025-11-05
feat(iota-types, iota-execution): Introduction of new basic types (ChangeEpochTransactionV4)

oliviasaa:
```mermaid
---
config:
  gitGraph:
    {parallelCommits: true,
    mainBranchName: 'Feature',
    rotateCommitLabel: true}
---
gitGraph
   commit id: "Initial commit"
   commit id: "Old Scorer" tag: "PR8530"
   checkout Feature
      branch new-types
   commit id: "add ChangeEpochV4"
   checkout Feature
    branch misbehavior-reports
   commit id: "add MisbehaviorReports"
    checkout misbehavior-reports
    branch scoring-metrics
   commit id: "add ScoringMetrics"
    checkout scoring-metrics
   branch scorer
   commit id: "add Scorer"
    checkout scorer
   branch scoring-metrics-store
   commit id: "add ScoringMetricsStore"
branch report-validation
commit id: "add report validation" 
branch scoring-function
commit id: "add scoring function" 
branch report-creation
commit id: "add report creation" 
branch advance-epoch
commit id: "add advance epoch v4" 
 checkout Feature
    merge new-types tag: "THIS PR"
```
# Description of change

This PR introduces: 
- a new variant of `EndOfEpochTransactionKind` which includes the `scores` field. This transaction is used when score based rewards are enabled (i.e., whenever the newly added feature flag `score_based_rewards` is set to true). 
- a placeholder for the new `advance_epoch_v4` function, currently with the same logic as `advance_epoch_v3`
- related changes in `iota-graphql-rpc` crate, as the `ChangeEpochTransactionV4` type, and a new variant of `IotaEndOfEpochTransactionKind`.

Note: the changes in the PR are built on top of commit [c7eeee2](https://github.com/iotaledger/iota-rust-sdk/tree/feat/add-scores) of `iota-rust-sdk`, which is currently not even merged to `develop`. Thus, any changes present in this PR should be merged to `iota`'s `develop` branch together with the changes in `iota-rust-sdk` repo.

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Introduce type ChangeEpochTransactionV4
- [ ] Indexer:
[...]

PR #9232, opened 2025-11-11
refactor(graphql-rpc,json-rpc): extend existing types to accommodate protocol ChangeEpochV4

kodemartin:
# Description of change

This follows #9159 and refactors `graphql-` and `json-rpc` types, by adding the new `scores` field as an optional field to the API types. 

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
@@ -48,7 +48,6 @@ pub(crate) struct EndOfEpochTransaction {
 pub(crate) enum EndOfEpochTransactionKind {
     ChangeEpoch(ChangeEpochTransaction),
     ChangeEpochV2(ChangeEpochTransactionV2),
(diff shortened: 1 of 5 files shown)


crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
```diff
@@ -62,8 +61,6 @@ pub(crate) struct ChangeEpochTransaction {
 }
 
 // System transaction for advancing the epoch.
```
Thoralf-M: I think it would be good to comment that this type is also used for V3 and V4 since the fields are just optionally added. I was first confused why there is no V3 type and v4 getting removed as well, but makes sense since the type can be reused

cyberphysic4l (approved):

crates/iota-graphql-rpc/src/types/transaction_block_kind/end_of_epoch.rs
kodemartin: Good point! ff3804eb26

Thoralf-M (approved):

PR #9233, opened 2025-11-12
feat(gRPC): add `GetEpoch` implementation in gRPC

bingyanglin:
# Description of change

- Implemented`GetEpoch` in the gRPC API.
- Verified the `get_epoch` simtest by using `cargo simtest --package iota-e2e-tests --test grpc -- get_epoch`
  
Part of https://github.com/iotaledger/iota/issues/8688


muXxer (approved):

crates/iota-types/src/storage/read_store.rs
```diff
@@ -881,3 +883,20 @@ pub struct CoinInfo {
...
+pub struct EpochInfo {
+    pub epoch: u64,
+    pub protocol_version: Option<u64>,
+    pub start_timestamp_ms: Option<u64>,
+    pub end_timestamp_ms: Option<u64>,
+    pub start_checkpoint: Option<u64>,
+    pub end_checkpoint: Option<u64>,
+    pub reference_gas_price: Option<u64>,
+    /// System State as of the start of the epoch
+    pub system_state: Option<crate::iota_system_state::IotaSystemState>,
```
Thoralf-M: Why is this optional? If it's the state as of the start of the epoch it should always be known?

bingyanglin: IMHO it's more flexible to make it optional, so for the old system states which are already pruned/non-existing it will be still compatible, and this is consistent w/ Sui's pattern, see https://github.com/MystenLabs/sui/blob/main/crates/sui-types/src/storage/read_store.rs#L788-L797.

Thoralf-M: Isn't all the other info also pruned then? Don't have a strong opinion about this, just wondered what's the point, feel free to resolve the conversation

crates/iota-core/src/rest_index.rs
piotrm50, line 109: `CURRENT_DB_VERSION`  needs to be updated according to this comment.

crates/iota-core/src/rest_index.rs
```diff
@@ -394,6 +407,68 @@ impl IndexStoreTables {
...
+            return Ok(());
+        };
+        if epoch_info.epoch > 0 {
+            let prev_epoch = epoch_info.epoch - 1;
+            let mut current_epoch = self.epochs.get(&prev_epoch)?.unwrap_or_default();
+            current_epoch.epoch = prev_epoch; // set this incase there wasn't an entry
+            current_epoch.end_timestamp_ms = epoch_info.start_timestamp_ms;
+            current_epoch.end_checkpoint = epoch_info.start_checkpoint.map(|sq| sq - 1);
+            batch.insert_batch(&self.epochs, [(prev_epoch, current_epoch)])?;
+        }
```
piotrm50, line 371: This is a bit confusing - why are we reading previous epoch data and assigning it as a current epoch? Please add more comments.

piotrm50 (dismissed):

crates/iota-core/src/rest_index.rs
bingyanglin: Thanks for pointing this out. Updated it at [8d304f5](https://github.com/iotaledger/iota/pull/9233/commits/8d304f5f407da059d333feb5fbf70b8f3bfb9fb1). Note that the tables are not finalized yet (see https://github.com/iotaledger/iota/pull/9259), and we will update the tables and the version at one time.

crates/iota-core/src/rest_index.rs
bingyanglin: Agreed. Added more comments in [2acac6b](https://github.com/iotaledger/iota/pull/9233/commits/2acac6b4c43551f3e5bc6aa0da19768b05f8a122). Hope this is clearer.

crates/iota-types/src/storage/read_store.rs
muXxer: I am with you here thoralf, I changed it.
There is one edge case where the former epoch might not have existed yet, and we would only store the end epoch info for that epoch then, but I don't think this is useful at all. For the genesis epoch we might need to index it differently tho.

crates/iota-core/src/rest_index.rs
muXxer: I reverted the change regarding the `CURRENT_DB_VERSION`.
We stay at version 0 for now, because I also reverted the changes to all the existing structs in the existing tables. The new tables don't need to be indexed back to genesis for now, and we don't want to trigger re-indexation of the whole database with the next node upgrade.
We will reconsider this as soon as we ship the endpoints in the gRPC interface, that are using those new indexes.

muXxer: `CURRENT_DB_VERSION` was now already increased in another PR for this feature branch. 👍

crates/iota/src/iota_commands.rs
```diff
@@ -809,6 +809,7 @@ async fn start(
             }
 
             // Apply gRPC configuration if enabled
+            swarm_builder = swarm_builder.with_fullnode_enable_grpc_api(enable_grpc_api);
```
thibault-martinez, line 812: Maybe put it above the comment because otherwise it's weird if `enable_grpc_api` is false

thibault-martinez (approved):

kodemartin (approved):

PR #9252, opened 2025-11-12
upstream(graphql-rpc): deprecate sign address transaction filters

### LINKED ISSUE #9253: graphql: deprecate graphql tx signer filters
## Description

This is to make the proper rust changes that follow https://github.com/MystenLabs/sui/pull/19430

kodemartin:
# Description of change

This patch follows the respective upstream changes to deprecate tx signers filters on GraphQL.

### Release Notes

- [x] GraphQL: :warning:  Deprecates `TransactionBlockFilter.signAddress`, replacing it with `TransactionBlockFilter.sentAddress` which behaves identically. Similarly `AddressTransactionBlockRelationship.SIGN` is deprecated and replaced by `AddressTransactionBlockRelationship.SENT`. The deprecated filter will be removed after more than six months, with the `1.24.0` mainnet release.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
@@ -24,25 +24,32 @@ pub(crate) struct TransactionBlockFilter {
     /// An input filter selecting for either system or programmable
     /// transactions.
     pub kind: Option<TransactionBlockKindInput>,
-    /// Limit to transactions that occurred strictly after the given checkpoint.
+    /// Limit to transactions that occured strictly after the given checkpoint.
     pub after_checkpoint: Option<UInt53>,
     /// Limit to transactions in the given checkpoint.
     pub at_checkpoint: Option<UInt53>,
-    /// Limit to transaction that occurred strictly before the given checkpoint.
+    /// Limit to transaction that occured strictly before the given checkpoint.
     pub before_checkpoint: Option<UInt53>,
-
-    /// Limit to transactions that were signed by the given address.
+    /// Limit to transactions that were sent by the given address. NOTE: this
+    /// input filter has been deprecated in favor of `sentAddress` which has
+    /// clearer semantics. Both filters restrict transactions by their sender,
+    /// only, not signers in general.
+    ///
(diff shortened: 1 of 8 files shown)


crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
```diff
@@ -24,25 +24,32 @@ pub(crate) struct TransactionBlockFilter {
     /// An input filter selecting for either system or programmable
     /// transactions.
     pub kind: Option<TransactionBlockKindInput>,
-    /// Limit to transactions that occurred strictly after the given checkpoint.
+    /// Limit to transactions that occured strictly after the given checkpoint.
```
tomxey: nit: this is failing the typos check

crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
```diff
@@ -147,10 +158,16 @@ impl TransactionBlockFilter {
             )
             // If SystemTx, sender if specified must be 0x0. Conversely, if sender is 0x0, kind must be SystemTx.
             || matches!(
```
tomxey, line 160: Should we also keep the old check in addition to the new one?

```rust
|| matches!(
    (self.kind, self.sign_address),
    (Some(kind), Some(signer))
        if (kind == TransactionBlockKindInput::SystemTx)
            != (signer == IotaAddress::from(NativeIotaAddress::ZERO))
)
```

Both `sent_address` and `sign_address` filters are still effective so I think we should check both.

crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
kodemartin: 6656e0e51b

crates/iota-graphql-rpc/src/types/transaction_block/filter.rs
kodemartin: Good catch! 

49f790c7e2

tomxey (approved):
lgtm

sergiupopescu199 (approved):
LGTM 🌸

PR #9258, opened 2025-11-12
feat(consensus): optimize cordial knowledge logic to reduce header synchronization requests

### LINKED ISSUE #9214: starfish: debug why transactions are not all comming from the block stream and shard reconstructor

VorobyevIlya:
# Description of change

PR removes the dissemination worker and makes updates in cordial knowledge without additional channel messages.
We update cordial knowledge about the block ancestors when creating the bundle, so the ancestors from useful headers authors are guaranteed to be included in the bundle. It decreases the number of header synchronisation requests.

fixes #9214

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -17,7 +17,6 @@ use iota_macros::fail_point_async;
 use parking_lot::RwLock;
 use starfish_config::AuthorityIndex;
 use tokio::sync::{Mutex, broadcast, mpsc::Sender};
-use tokio::time::sleep;
 use tokio_util::sync::ReusableBoxFuture;
 use tracing::{debug, info, warn};
 
@@ -549,6 +548,12 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
             .add_block_headers(additional_block_headers.clone())
             .await
             .map_err(|_| ConsensusError::Shutdown)?;
+        self.context
+            .metrics
+            .node_metrics
+            .missing_ancestors_from_streaming
+            .with_label_values(&["headers"])
+            .observe(missing_ancestors.len() as f64);
 
         // 10. Add the block to dag, add its missing ancestors to the set
         let (missing_block_ancestors, missing_block_committed_transactions) = self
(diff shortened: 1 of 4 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -652,6 +672,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
                 let context = context.clone();
                 let connection_knowledge = connection_knowledge.clone();
                 async move {
+                    // tokio::time::sleep(std::time::Duration::from_millis(50)).await;
```
piotrm50: Debugging leftover?

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -435,15 +319,38 @@ impl CordialKnowledge {
...
-                    let buffer_size = self.cordial_knowledge_receiver.len() + 1;
+                    let mut batch = vec![msg];
+                    while let Ok(msg) = self.cordial_knowledge_receiver.try_recv() {
+                        batch.push(msg);
+                    }
+                    // Report the buffer size
                     self.context
                         .metrics
                         .node_metrics
                         .cordial_knowledge_buffer_size
```
piotrm50: Metric name seems to no longer correspond to the buffer size, but is rather a histogram of single batch sizes from the receiver.

crates/starfish/core/src/cordial_knowledge.rs
piotrm50: weird comment formatting

crates/starfish/core/src/authority_service.rs
polinikita: Remove this line

piotrm50 (approved):
I like this change - gets rid of a redundant layer (the worker). And we make sure that CordialKnowledge is aware of the Block's parents when creating a bundle.

crates/starfish/core/src/authority_service.rs
VorobyevIlya: yes. removed

crates/starfish/core/src/cordial_knowledge.rs
VorobyevIlya: renamed to cordial_knowledge_message_batch_size

crates/starfish/core/src/authority_service.rs

crates/starfish/core/src/cordial_knowledge.rs
VorobyevIlya: fixed the formatting

polinikita (approved):

PR #9259, opened 2025-11-13
upstream(core): rework rpc indexes

bingyanglin:
# Description of change

- Upstream range: [v1.46.3, 1.47.1)
- Port commit: 
  - https://github.com/MystenLabs/sui/commit/343906c76b15f93245b497c14f4489f90d8bad8d
  - https://github.com/MystenLabs/sui/commit/1a496a55ca8b02e936e984982f9ab56c121534e9
  - https://github.com/MystenLabs/sui/pull/21711/commits/aec90b4f713dca02d49c9ca9f823f5a7a0a503a7
  - https://github.com/MystenLabs/sui/pull/21711/commits/24f9ea25c0b27b55359be145ae352a8013766bc4
- Description:
  - Avoid std::thread::sleep() in safe_drop_db()
  - Add backoff to safe_drop_db
    - Mitigates the race between the destructor and metric threads. Reason: `rocksdb::DB::Destroy` returns an error immediately if any other reference is still alive
  - Reworks the 'rest-index' database to:
    - Add support for indexing information about epochs

None

### Release Notes

- [x] Nodes (Validators and Full nodes): Add support for indexing epochs, object type information for transactions, type information for dynamic fields. ⚠️⚠️⚠️ The migration may incur a downtime of several minutes upon restart of the full nodes, so operators are requested to stage it according to their SLA needs.⚠️⚠️⚠️


muXxer (approved):

crates/iota-types/src/coin.rs
```diff
@@ -70,18 +70,16 @@ impl Coin {
     /// balance Ok(Some(u64)). If it's not a Coin, return Ok(None).
     /// The cost is 2 comparisons if not a coin, and deserialization if its a
     /// Coin.
-    pub fn extract_balance_if_coin(object: &Object) -> Result<Option<u64>, bcs::Error> {
-        match &object.data {
-            Data::Move(move_obj) => {
-                if !move_obj.is_coin() {
-                    return Ok(None);
-                }
+    pub fn extract_balance_if_coin(object: &Object) -> Result<Option<(TypeTag, u64)>, bcs::Error> {
```
kodemartin: I think it is more intuitive to stick with the original signature and implementation. I would expect `extract_balance_if_coin` to extract the balance, not return an opaque tuple.

Callers that want to get the type information can always use the `&Object` passed and extract it explicitly.

kodemartin (commented):
lgtm for the infra part, except for the a signature change that I think we can avoid.

kodemartin (approved):

PR #9261, opened 2025-11-13
upstream(graphql): deduplicate configs and clap args, add flag to skip db compatibility

### LINKED ISSUE #9217: upstream(graphql): deduplicate configs and clap args
https://github.com/MystenLabs/sui/pull/19670
https://github.com/MystenLabs/sui/pull/19671
https://github.com/MystenLabs/sui/pull/19710

tomxey:
# Description of change

Original descriptions:

> Avoid duplicating fields for configs that are accepted as flags from the command-line and can also be read from TOML file. Use the same struct for both purposes but deocrate it with both clap and serde (or GraphQLConfig) annotations so that it can serve both purposes.

> Add a flag to optionally skip database migration compatibility checks.
This is helpful when trying to connect a local build up to a production
database to test a specific change, even if you are aware that other
queries may not be compatible.
To accommodate this change, the compatibility check was also moved into
`ServerBuilder` where the config is readily available. This also
slightly simplifies the `Server` itself, which no longer needs to hold
onto its own instance of the `Db`.

fixes https://github.com/iotaledger/iota/issues/9217

Comparing `cargo run --bin iota-graphql-rpc -- start-server --help` before and after the change.

- [x] Deployment of services using Docker. - Tested if graphql starts up successfully with `pg-services-local`

### DIFF
--- a/crates/iota-graphql-rpc/src/commands.rs
+++ b/crates/iota-graphql-rpc/src/commands.rs
@@ -6,6 +6,8 @@ use std::path::PathBuf;
 
 use clap::*;
 
+use crate::config::{ConnectionConfig, Ide, TxExecFullNodeConfig};
+
 #[derive(Parser)]
 #[command(name = "iota-graphql-rpc", about = "IOTA GraphQL RPC", author)]
 pub enum Command {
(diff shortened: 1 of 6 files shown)


crates/iota-graphql-rpc/src/commands.rs
```diff
@@ -21,34 +23,17 @@ pub enum Command {
...
-        /// Host to bind the server to
-        #[arg(long)]
-        host: Option<String>,
-        /// Port to bind the prom server to
-        #[arg(long)]
-        prom_port: Option<u16>,
-        /// Host to bind the prom server to
-        #[arg(long)]
-        prom_host: Option<String>,
+        #[clap(flatten)]
```
thibault-martinez: the `clap` attribute is deprecated, it is better to use the `arg` attribute

thibault-martinez: same for the other files

thibault-martinez (approved):
Approving dev-tools owned files

crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -450,6 +440,14 @@ impl ServerBuilder {
         )
         .map_err(|e| Error::ServerInit(format!("Failed to create pg connection pool: {e}")))?;
 
+        {
```
kodemartin: Isn't this the place where we should skip migration consistency check?

```suggestion
        if !config.connection.skip_migration_consistency_check {
```

kodemartin (commented):
Hey @tomxey! Looks good apart @thibault-martinez comments and a more specific suggestion that might warrant a change.

Also is the new config/cli tested against docker runs? Locally or on staging

crates/iota-graphql-rpc/src/commands.rs
tomxey: Done in e42a64ca29328df88940341a835293f6ba411bdc

crates/iota-graphql-rpc/src/server/builder.rs
tomxey: Very good catch, that was the very point of this change 😅 
Fixed in [d7b4d01](https://github.com/iotaledger/iota/pull/9261/commits/d7b4d01f9e6c2020fa97b319cf195f4ba0ca7d1c)

tomxey:
> Also is the new config/cli tested against docker runs? Locally or on staging

Tested with `pg-services-local` and graphql starts fine

kodemartin (approved):
lgtm, pending merge of https://github.com/iotaledger/iota/pull/9262 to remove the flakiness on the compatibility tests

PR #9264, opened 2025-11-13
upstream(iota-data-ingestion-core): introduce checkpoint upper limit

### LINKED ISSUE #9251: data ingestion: introduce checkpoint upper limit for ingestion job
https://github.com/MystenLabs/sui/pull/19225

sergiupopescu199:
# Description of change
This PR adds new APIs for the `IndexerExecutor` for managing the ingestion upper limit. Once the ingestion process reaches the upper limit the frameworks will initiate the graceful shutdown process.

This path diverges a bit from the upstream, mostly because of how graceful shutdown is handled and the different guarantees we aim to provide to the library users.

Difference from upstream behavior:

- Upstream: Reaching the limit immediately breaks the executor loop, dropping tasks, and it may exit before the limit checkpoint is actually processed.
- In our implementation, once the upper limit checkpoint is reached, no further checkpoints are sent to workers. The executor does not terminate immediately; it waits until workers advance the watermark past the limit. Only then is the graceful shutdown sequence started.

fixes #9251 

- Added a new test case for the checkpoint upper limit feature.
- Ran the tests 10 times 
    ```shell
    ```
- Modified the `examples/custom-indexer/rust/remote_reader.rs` to sync `testnet` data up to checkpoint using different checkpoint upper limits: 10, 100, 10k. All runs were successful, and the upper limit was respected. Logs from last run:
```log
2025-11-13T09:39:35.978788Z  INFO iota_data_ingestion_core::reader::fetch: cleaning processed files, watermark is 9999
2025-11-13T09:39:35.978810Z  INFO iota_data_ingestion_core::reader::fetch: cleaning processed files, watermark is 10000
2025-11-13T09:39:36.356335Z  INFO iota_data_ingestion_core::worker_pool: received checkpoint for processing 10000 for workflow remote_reader
Processing checkpoint: CheckpointSummary { epoch: 0, seq: 10000, content_digest: CdZuLf2JwogAM2dWzfUtaBecbcBrWZokmg59P2g585nQ,
            epoch_rolling_gas_cost_summary: GasCostSummary { computation_cost: 0, computation_cost_burned: 0, storage_cost: 0, storage_rebate: 0, non_refundable_storage_fee: 0 }}
2025-11-13T09:39:36.356392Z  INFO iota_data_ingestion_core::worker_pool: finished checkpoint processing 10000 for workflow remote_reader in 37.916µs
2025-11-13T09:39:36.356360Z  INFO iota_data_ingestion_core::executor: checkpoint upper limit reached: no further checkpoints will be sent, waiting for workers to process remaining checkpoints
2025-11-13T09:39:36.356439Z  INFO iota_data_ingestion_core::executor: checkpoint upper limit reached: last checkpoint was processed, shutdown process started
2025-11-13T09:39:36.356561Z  INFO iota_data_ingestion_core::worker_pool: Worker pool `remote_reader` terminated gracefully
2025-11-13T09:39:39.509208Z  INFO iota_data_ingestion_core::reader::v2: Read from remote. Current checkpoint number: 20000, pruning watermark: 10000
2025-11-13T09:39:39.509262Z  INFO iota_data_ingestion_core::reader::fetch: cleaning processed files, watermark is 10001
2025-11-13T09:39:39.509433Z  INFO iota_data_ingestion_core::history::reader: terminating the manifest sync loop
```

> [!NOTE]
> The patch does not affect the normal operation of the indexer, this optional APIs are not used by default by the indexer. Thus no further tests were conducted.

### Release Notes
- [x] Indexer: `iota-data-ingestion-core`: Introduced new APIs for managing the ingestion upper limit:
  - `IndexerExecutor::with_ingestion_limit`: offers high level upper limit policies through `IngestionLimit` enum. It provides explicit control to automatically start the graceful shutdown process upon reaching a defined upper limit (e.g., maximum checkpoint number or epoch boundary). 
  - `IndexerExecutor::shutdown_when`: offers the ability to provide a custom callback for defining more complex upper limit scenarios for the ingestion process.

### DIFF
--- a/crates/iota-data-ingestion-core/src/executor.rs
+++ b/crates/iota-data-ingestion-core/src/executor.rs
@@ -7,7 +7,7 @@ use std::{path::PathBuf, pin::Pin, sync::Arc};
 use futures::Future;
 use iota_metrics::spawn_monitored_task;
 use iota_rest_api::CheckpointData;
-use iota_types::messages_checkpoint::CheckpointSequenceNumber;
+use iota_types::{committee::EpochId, messages_checkpoint::CheckpointSequenceNumber};
 use prometheus::Registry;
 use tokio::{
     sync::{mpsc, oneshot},
@@ -20,7 +20,7 @@ use crate::{
     DataIngestionMetrics, IngestionError, IngestionResult, ReaderOptions, Worker,
     progress_store::{ExecutorProgress, ProgressStore, ProgressStoreWrapper, ShimProgressStore},
     reader::{
-        v1::{CheckpointReader as CheckpointReaderV1, IngestionLimit},
+        v1::CheckpointReader as CheckpointReaderV1,
         v2::{CheckpointReader as CheckpointReaderV2, CheckpointReaderConfig},
     },
     worker_pool::{WorkerPool, WorkerPoolStatus},
@@ -31,6 +31,37 @@ pub const MAX_CHECKPOINTS_IN_PROGRESS: usize = 10000;
 /// Callback function that when resolved to `true`, the executor will start the
 /// shutdown process.
 type ShutdownCallback = Box<dyn FnMut(&CheckpointData) -> bool + Send>;
+
+/// Common policies for upper limit checkpoint ingestion by the framework.
+///
+/// Once the limit is reached, the framework will start the graceful
+/// shutdown process.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum IngestionLimit {
+    /// Last checkpoint sequence number to process.
+    ///
+    /// After processing this checkpoint the framework will start the graceful
+    /// shutdown process.
+    MaxCheckpoint(CheckpointSequenceNumber),
+    /// Last checkpoint to process based on the given epoch.
+    ///
+    /// After processing this checkpoint the framework will start the graceful
+    /// shutdown process.
+    EndOfEpoch(EpochId),
+}
+
+impl IngestionLimit {
+    /// Returns `true` if the given checkpoint matches the limit.
+    fn matches(&self, checkpoint: &CheckpointData) -> bool {
+        match self {
+            IngestionLimit::MaxCheckpoint(max) => {
+                &checkpoint.checkpoint_summary.sequence_number > max
+            }
+            IngestionLimit::EndOfEpoch(max) => &checkpoint.checkpoint_summary.epoch > max,
+        }
+    }
+}
 
 /// Represents a common interface for checkpoint readers.
 ///
@@ -218,11 +249,38 @@ impl<P: ProgressStore> IndexerExecutor<P> {
     /// *further* new checkpoints to the worker pool and wait for all
     /// previously sent checkpoints to be processed by the workers
     /// before initiating the graceful shutdown sequence.
+    ///
+    /// Note:
+    ///
+    /// Calling this method after
+    /// [`with_ingestion_limit`](Self::with_ingestion_limit) replaces the
+    /// earlier predicate, and vice versa. They are not cumulative.
     pub fn shutdown_when<F>(&mut self, f: F)
     where
         F: FnMut(&CheckpointData) -> bool + Send + 'static,
     {
         self.shutdown_callback = Some(Box::new(f));
+    }
+
+    /// Adds an upper‑limit policy that determines when the ingestion
+    /// process should stop.
+    ///
+    /// This is a convenience method, it internally uses
+    /// [`shutdown_when`](Self::shutdown_when) by registering a predicate
+    /// derived from the provided [`IngestionLimit`].
+    ///
+    /// If the callback resolves to `true`, the **current checkpoint is
+    /// discarded**. The executor will then immediately stop sending any
+    /// *further* new checkpoints to the worker pool and wait for all
+    /// previously sent checkpoints to be processed by the workers
+    /// before initiating the graceful shutdown sequence.
+    ///
+    /// Note:
+    ///
+    /// Calling this method after [`shutdown_when`](Self::shutdown_when)
+    /// replaces the earlier predicate, and vice versa. They are not cumulative.
+    pub fn with_ingestion_limit(&mut self, limit: IngestionLimit) {
+        self.shutdown_when(move |checkpoint| limit.matches(checkpoint));
     }
 
     pub async fn update_watermark(
@@ -253,7 +311,6 @@ impl<P: ProgressStore> IndexerExecutor<P> {
         reader_options: ReaderOptions,
     ) -> IngestionResult<ExecutorProgress> {
         let reader_checkpoint_number = self.progress_store.min_watermark()?;
-        let ingestion_upper_limit = reader_options.ingestion_upper_limit;
         let (checkpoint_reader, checkpoint_recv, gc_sender, exit_sender) =
             CheckpointReaderV1::initialize(
                 path,
@@ -273,7 +330,6 @@ impl<P: ProgressStore> IndexerExecutor<P> {
                 exit_sender,
                 handle,
             },
-            ingestion_upper_limit,
         )
         .await
     }
@@ -291,14 +347,11 @@ impl<P: ProgressStore> IndexerExecutor<P> {
         config: CheckpointReaderConfig,
     ) -> IngestionResult<ExecutorProgress> {
         let reader_checkpoint_number = self.progress_store.min_watermark()?;
-        let ingestion_upper_limit = config.reader_options.ingestion_upper_limit;
-
         let checkpoint_reader = CheckpointReaderV2::new(reader_checkpoint_number, config).await?;
 
         self.run_executor_loop(
             reader_checkpoint_number,
             CheckpointReader::V2(checkpoint_reader),
-            ingestion_upper_limit,
         )
         .await
     }
@@ -308,7 +361,6 @@ impl<P: ProgressStore> IndexerExecutor<P> {
         &mut self,
         mut reader_checkpoint_number: u64,
         mut checkpoint_reader: CheckpointReader,
-        ingestion_upper_limit: Option<IngestionLimit>,
     ) -> IngestionResult<ExecutorProgress> {
         let worker_pools = std::mem::take(&mut self.pools)
             .into_iter()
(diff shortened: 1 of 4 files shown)


alexsporn:
I would try to generalize this approach and provide access to the CancellationToken from within the Worker (and also the Reducer) so that any generic stop condition can be implemented. If this is done on that level, then default implementations can be provided for stopping at an upper limit checkpoint or even at epoch boundaries (I did this in one of my projects).

sergiupopescu199:
@alexsporn @iotaledger/infrastructure

### Proposed Hybrid Approach for Ingestion Limits

After discussing the options, we've decided on a hybrid approach that provides both simplicity for common use cases and full flexibility for custom requirements.

#### 1. Rationale

The core motivation is to implement high-level cancellation policies **outside** of the `Worker` and `Reducer` traits. This approach:
* **Separates Concerns:** Keeps the checkpoint collection/filtering logic distinct from the processing logic (Workers/Reducer).
* **Simplifies User Logic:** Avoids requiring users to manually handle tokens inside every Worker/Reducer implementation for basic stopping needs (though they still can for advanced logic).

#### 2. The Hybrid Solution

We will implement two mechanisms that work in conjunction:

| Mechanism | Purpose | How it's Used |
| :--- | :--- | :--- |
| **IngestionLimit (Enum)** | For common stop guards (Max Checkpoint, Epoch). Defined in the **Reader Options**. | Simple configuration via the `ReaderOptions` passed to `Executor::run_with_config` or `Executor::run`. |
| **Shutdown Callback** | For arbitrary, complex custom logic (e.g., stop based on object ID, time, specific data). | Configured via the new `Executor::shutdown_when()` method. |

---

#### 3. Implementation Pseudocode

This logic will reside in the `IndexerExecutor`:

```rust
#[non_exhaustive]
pub enum IngestionLimit {
    MaxCheckpoint(u64),
    EndOfEpoch(u64), // stop after last checkpoint whose epoch <= given epoch
}

IndexerExecutor {
    // other fields
    shutdown_callback: Option<Box<dyn Fn(&CheckpointData) -> bool + Send>>,
}

impl IndexerExecutor {
    // Allows users to define custom stop logic
    // this is completly optional
    pub fn shutdown_when<F>(&mut self, f: F)
    where
        F: Fn(&CheckpointData) -> bool + Send + 'static,
    {
        self.shutdown_callback = Some(Box::new(f));
    }

    // Internally, the Executor will perform this check for every incoming checkpoint before
    // sending them to the workers. When it returns true, then it will start the graceful shutdown.
    fn should_shutdown(&self, checkpoint: &CheckpointData, ingestion_limit: Option<&IngestionLimit>,) -> bool {
        let is_reader_opt_upper_limit_reached = ingestion_limit.is_some_and(|upper_limit| /* Logic to match checkpoint & enum variants */);

        let is_callback_upper_limit_reached = self.shutdown_callback
            .as_ref()
            .is_some_and(|is_reached| is_reached(&checkpoint));

        is_reader_opt_upper_limit_reached || is_callback_upper_limit_reached
    }
}
```
A usage example

```rust
executor.register(SomeWorker);
executor.shutdown_when(|chk| {
    // Stop after epoch 42 finishes (i.e., when we see first checkpoint of epoch 43)
    chk.checkpoint_summary.epoch > 42
});
executor.run_with_config(SomeConfig).await
```

crates/iota-data-ingestion-core/src/reader/v1.rs
```diff
@@ -94,6 +126,13 @@ pub struct ReaderOptions {
     ///
     /// Default: 0.
     pub data_limit: usize,
+    /// Common policies for upper limit checkpoint ingestion by the framework.
+    ///
+    /// Once the limit is reached, the framework will start the graceful
+    /// shutdown process.
+    ///
+    /// Default: No ingestion limits are enforced.
+    pub ingestion_upper_limit: Option<IngestionLimit>,
```
kodemartin: This is a breaking change.

I would suggest keeping this field private and add a Builder method `with_ingestion_limit` to accommodate the new feature.

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -219,6 +253,7 @@ impl<P: ProgressStore> IndexerExecutor<P> {
         reader_options: ReaderOptions,
     ) -> IngestionResult<ExecutorProgress> {
         let reader_checkpoint_number = self.progress_store.min_watermark()?;
+        let ingestion_upper_limit = reader_options.ingestion_upper_limit;
```
kodemartin: Wouldn't it be simpler to set `self.shutdown_callback` based on the `ingestion_upper_limit` and use that instead down the call path?

kodemartin (changes requested):
looks good with a change request, plus a suggestion that might simplify a bit the implementation.

crates/iota-data-ingestion-core/src/executor.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9264/commits/f4ddb2dd418fd83ecc2f0250d80b69de40e5888a

crates/iota-data-ingestion-core/src/reader/v1.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9264/commits/f4ddb2dd418fd83ecc2f0250d80b69de40e5888a

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -326,6 +441,17 @@ impl<P: ProgressStore> IndexerExecutor<P> {
 
         Ok(self.progress_store.stats())
     }
+
+    /// Check if the current ingestion limit has been reached.
+    ///
+    /// Returns `true` if the ingestion limit has been reached.
+    /// If no ingestion limit is present or it has not been reached yet, the
+    /// function returns `false`.
+    fn is_ingestion_upper_limit_reached(&mut self, checkpoint: &CheckpointData) -> bool {
```
kodemartin: Nit: `should_shutdown` follows more intuitively the semantics of the feature.

```rust
    fn should_shutdown(&mut self, checkpoint: &CheckpointData) -> bool {
```

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -277,8 +368,25 @@ impl<P: ProgressStore> IndexerExecutor<P> {
...
 
         loop {
+            // the min watermark represents the lowest sequence number that
+            // has been processed by any worker pool. This guarantees that
+            // all worker pools have processed the checkpoint before the
+            // shutdown process starts.
+            if checkpoint_limit_reached.is_some_and(|ch_seq_num| {
+                self.progress_store
+                    .min_watermark()
+                    .map(|watermark| watermark > ch_seq_num)
```
kodemartin, line 400: If `watermark` is the lowest sequence number processed as described in the comment, shouldn't this be ?

```rust
                    .map(|watermark| watermark == ch_seq_num)
```
Also why do we need this check? Doesn't `tokio.cancel` trigger graceful shutdown for the worker pools allowing them to process pending checkpoints?

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -27,17 +28,59 @@ use crate::{
 
 pub const MAX_CHECKPOINTS_IN_PROGRESS: usize = 10000;
 
+/// Callback function that when resolved to `true`, the executor will start the
+/// shutdown process.
+type ShutdownCallback = Box<dyn FnMut(&CheckpointData) -> bool + Send>;
```
kodemartin: Using `FnMut` here puzzled me a bit. I would expect that we only need to read checkpoint data and assert if we must shutdown in the callback.

Then, after I saw the test scenarios (`fn basic_flow_with_custom_callback_inclusive() `), I realized that this is necessary to support `shutdown_after` semantics in the callback. Which requires users to inject state into the closure, and is in conflict with the `shutdown_when` builder method we expose which has actually `shutdown_before` semantics. That is we currently discard the `CheckpointData` for which the closure returns `true` 

So I have to ask: Why don't we apply `shutdown_after` semantics by including the `CheckpointData` for which the callback returns `true`? This covers the simple `IngestionLimit` cases plus any custom inclusive condition. 

Assuming that inclusive conditions on more refined checkpoint-data (like the included transaction in the test scenario) are more likely to be desired.

sergiupopescu199: I opted for an explicit handling inclusive and exclusive cases through an enum and removed the `FnMut`.

fixed: https://github.com/iotaledger/iota/pull/9264/commits/751dcfb6cee839fc707c6d394c5909f7f5034626

crates/iota-data-ingestion-core/src/executor.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9264/commits/751dcfb6cee839fc707c6d394c5909f7f5034626

crates/iota-data-ingestion-core/src/executor.rs
sergiupopescu199: Adjusted the comment. I think we still need this check. When the graceful shutdown is triggered only the checkpoints that the worker is currently processing will be processed and after that the worker will shut down. The remaining  checkpoints in worker pool's queue will be dropped.

This check makes sure that the worker pool sends all remaining checkpoints from its queue to workers to process. Every processed checkpoint from worker increments the watermark, so it will be always greater than the upper limit (watermark = lowest sequence number processed  + 1).

fixed: https://github.com/iotaledger/iota/pull/9264/commits/751dcfb6cee839fc707c6d394c5909f7f5034626

crates/iota-data-ingestion-core/src/executor.rs
kodemartin: nice!

kodemartin (approved):
lgtm 🏹

kodemartin (commented):
Please also update the release notes according to the latest changes.

crates/iota-data-ingestion-core/src/tests.rs
```diff
@@ -196,15 +209,253 @@ async fn basic_flow() {
...
+    fs::remove_dir_all(path).unwrap();
+}
+
+// Test: graceful shutdown via a custom callback.
+//
+// Scenario:
+// A transaction with a known digest is embedded only in checkpoint 10. The
+// callback `shutdown_when` inspects each processed checkpoint and returns
+// `true` if it contains the target transaction digest. Once the condition is
+// met, the Executor initiates graceful shutdown without handing that checkpoint
```
tomxey: nit: we are using "IncludeAndShutdown" in the test, so I would say that "the Executor hands over the checkpoint to workers, and then initiates graceful shutdown"

tomxey (approved):
lgtm, one small nit

crates/iota-data-ingestion-core/src/tests.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9264/commits/eb7547878cc89f2c006ed884e3bfcfb1dcac23c7

PR #9280, opened 2025-11-13
fix(wallet): fix accountsFinder start account index

Thoralf-M:
# Description of change

If there are accounts with index 0 and 2, the account finder would start with index 2 (length accounts). So if one deleted an account by accident or received funds to an account with lower account index than one still has in the wallet, then it's impossible to find this account using the balance finder. One would need to recreate the whole profile.
This PR fixes this by setting the account start index to the first account index that is missing. This can potentially check existing accounts again, but that should still be better and seems to be not create any duplicated entries/issues.


Thoralf-M:
Turns out the AccountsFinder isn't initialized with existing accounts, so it always starts at 0 and will also recheck existing accounts, so this is section relevant for subsequent runs and then there can't be any gaps

PR #9294, opened 2025-11-14
refactor(iota-graphql-rpc): change log level for Subscription responses

### LINKED ISSUE #9114: [iota-graphql-rpc]: Subscription log polution
For every Item sent to the subscriber the server generates logs, for a few subscribers i not an issue, for thousands we experience log pollution.

We shall downgrade from INFO to TRACE the emission of these logs only for the Subscription type of Responses.

``` log
025-10-31T11:58:18.288778Z  INFO iota_graphql_rpc::extensions::logger: [Response] {transactions: {__typename: "TransactionBlock", digest: "77DkmfwznPbA91hUx7shYR6i6Aszsv5FjUcRXRS4iVWp"}} query_id=3be1f6ea-c2e5-4643-886f-2c78c3687a93 session_id=127.0.0.1:64556
2025-10-31T11:58:18.288969Z  INFO iota_graphql_rpc::extensions::logger: [Response] {transactions: {__typename: "TransactionBlock", digest: "FL1jt1aEHyB4kQksLYZuFgwviq2PDZagSdM26UCvVZC1"}} query_id=3be1f6ea-c2e5-4643-886f-2c78c3687a93 session_id=127.0.0.1:64556
```

sergiupopescu199:
# Description of change

This PR modifies the log level from `INFO` to `DEBUG` for Subscription Responses in the Logger extension to avoid log pollution. 
Since every item received from the Subscription stream is treated as a GraphQL Response. With multiple subscribers at the same time, the logs become too polluted and hard to read.

fixes #9114 

- Ran the GraphQL server by using the iota cli
```shell
RUST_LOG=off,iota_graphql_rpc=info cargo r --features indexer -- start --force-regenesis   --with-indexer --with-graphql
```
When making Subscription requests, the logs were no longer present. If enabling `RUST_LOG=off,iota_graphql_rpc=debug`, the logs were visible.

```shell
2025-11-14T14:58:14.201495Z  INFO iota_graphql_rpc::extensions::logger: [Validation] query_id=9eef57e6-edf0-42cd-829b-6e4b469b8919 session_id=127.0.0.1:52594 complexity=3 depth=2
2025-11-14T14:58:14.456023Z DEBUG iota_graphql_rpc::extensions::logger: [Subscription] {transactions: {__typename: "TransactionBlock", digest: "4aQ17xRBjo9bCgQLcKUdqatUjm7Yck2kUtbJTfwDfj4c"}} query_id=9eef57e6-edf0-42cd-829b-6e4b469b8919 session_id=127.0.0.1:52594
```

> [!NOTE]
> This patch does not affect the normal operations of the indexer; thus, no tests were conducted.

### DIFF
--- a/crates/iota-graphql-rpc/src/extensions/logger.rs
+++ b/crates/iota-graphql-rpc/src/extensions/logger.rs
@@ -2,17 +2,25 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{fmt::Write, net::SocketAddr, sync::Arc};
+use std::{
+    fmt::Write,
+    net::SocketAddr,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+};
 
 use async_graphql::{
     PathSegment, Response, ServerError, ServerResult, ValidationResult, Variables,
     extensions::{
         Extension, ExtensionContext, ExtensionFactory, NextExecute, NextParseQuery, NextResolve,
-        NextValidation, ResolveInfo,
+        NextSubscribe, NextValidation, ResolveInfo,
     },
     parser::types::{ExecutableDocument, OperationType, Selection},
 };
 use async_graphql_value::ConstValue;
+use futures::stream::BoxStream;
 use tracing::{debug, error, info, warn};
 use uuid::Uuid;
 
@@ -44,16 +52,31 @@ impl ExtensionFactory for Logger {
     fn create(&self) -> Arc<dyn Extension> {
         Arc::new(LoggerExtension {
             config: self.config.clone(),
+            is_subscription: AtomicBool::default(),
         })
     }
 }
 
 struct LoggerExtension {
     config: LoggerConfig,
+    /// Marks if the current request is a subscription.
+    is_subscription: AtomicBool,
 }
 
 #[async_trait::async_trait]
 impl Extension for LoggerExtension {
+    fn subscribe<'s>(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        stream: BoxStream<'s, Response>,
+        next: NextSubscribe<'_>,
+    ) -> BoxStream<'s, Response> {
+        // flag this operation as a subscription so later hooks (execute) can
+        // adjust logging.
+        self.is_subscription.store(true, Ordering::Relaxed);
+        next.run(ctx, stream)
+    }
+
     // This hook is used to get the top level node name for recording in the metrics
     // which top level nodes are being called.
     async fn resolve(
@@ -199,10 +222,19 @@ impl Extension for LoggerExtension {
                         "[Schema] {}", resp.data
                     );
                 }
-                _ => info!(
+                _ if self.is_subscription.load(Ordering::Relaxed) => {
+                    // a subscription can emit many payloads; to avoid flooding normal response
+                    // logs we log subscription payloads at debug level.
+                    debug!(
                         %query_id,
                         %session_id,
-                        "[Response] {}", resp.data
+                        "[Subscription] {}", resp.data
+                    );
+                }
+                _ => info!(
+                    %query_id,
+                    %session_id,
+                    "[Response] {}", resp.data
                 ),
             }
         }


crates/iota-graphql-rpc/src/extensions/logger.rs
```diff
@@ -44,16 +52,31 @@ impl ExtensionFactory for Logger {
...
 impl Extension for LoggerExtension {
+    fn subscribe<'s>(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        stream: BoxStream<'s, Response>,
+        next: NextSubscribe<'_>,
+    ) -> BoxStream<'s, Response> {
+        // flag this operation as a subscription so later hooks (execute) can
+        // adjust logging.
+        self.is_subscription.store(true, Ordering::Relaxed);
```
kodemartin, line 76: Doesn't this imply that the `is_subscription` flag will be set to true for every subsequent request, regardless of the underlying operation? So for `query, mutation` as well?

Is `LoggerExtension` instantiated once, or for every request?

crates/iota-graphql-rpc/src/extensions/logger.rs
```diff
@@ -199,10 +222,19 @@ impl Extension for LoggerExtension {
                         "[Schema] {}", resp.data
                     );
                 }
-                _ => info!(
+                _ if self.is_subscription.load(Ordering::Relaxed) => {
```
kodemartin, line 225: Can't you take advantage of  the `operation_name`  here to resolve if the request is a subscription?

crates/iota-graphql-rpc/src/extensions/logger.rs
sergiupopescu199: The `is_subscription` flag is scoped to a single GraphQL execution request. An extension instance is created per request via the `ExtensionFactory::create` call. That means:

1. For a `query` or `mutation` request:
   - A fresh `LoggerExtension` instance is created.
   - `is_subscription` will be always `false`.
   - No invocation of the [`Extension::subscription`](https://docs.rs/async-graphql/latest/async_graphql/extensions/trait.Extension.html#method.subscribe) trait method, instead the [`Extension::request`](https://docs.rs/async-graphql/latest/async_graphql/extensions/trait.Extension.html#method.request)trait method (and any other non-subscription hooks) runs for the lifecycle of that request.

2. For a `subscription` request:
   - A fresh `LoggerExtension` instance is created for the initial subscription operation.
   - `is_subscription` is set to `true` for that instance only.
   - That same instance then lives for the lifetime of the subscription stream, but it does not affect any other requests.

crates/iota-graphql-rpc/src/extensions/logger.rs
sergiupopescu199: Using `operation_name` isn’t a reliable way to decide whether a request is a subscription.

Reasons:
1. It’s optional at the transport level: if the document has a single operation, the client can omit `operationName`.
2. Anonymous operations (allowed only when the document contains exactly one operation) have no name at all.
3. A name does not encode operation type; a subscription can be named like a query (e.g. `Transactions`, `Events`, etc.).

Concrete examples:

Named subscription (extension receives `operation_name = Some("Transactions")`):
```graphql
subscription Transactions {
    transactions {
        __typename
        ... on TransactionBlock {
            digest
        }
        ... on Lagged {
            count
        }
    }
}
```

Anonymous subscription (extension receives `operation_name = None`):
```graphql
subscription {
    transactions {
        __typename
        ... on TransactionBlock {
            digest
        }
        ... on Lagged {
            count
        }
    }
}
```

Both execute as subscriptions; the only difference is the presence of a client-chosen name.

After reviewing the `Extension` trait, the reliable discriminator is the framework’s own lifecycle separation:
- For subscription operations, `Extension::subscription` is invoked.
- For query or mutation operations, only `Extension::request` is invoked.

Per the async-graphql [documentation](https://async-graphql.github.io/async-graphql/en/extensions_inner_working.html#request), this dispatch is guaranteed; This implies that we can safely set an internal `is_subscription` flag based on which hook fires rather than inspecting `operation_name`.

crates/iota-graphql-rpc/src/extensions/logger.rs
kodemartin: Thank you for the double check and clarifications!

crates/iota-graphql-rpc/src/extensions/logger.rs
kodemartin: Thank you for the clarifications!

kodemartin (approved):
lgtm

tomxey (approved):

PR #9296, opened 2025-11-15
feat(execution): create a package metadata object at publish time

### LINKED ISSUE #8861: [AA] Create `PackageMetadata` type
After the validation of a package, if some functions were found using the #[authenticator] attribute, an immutable object can be automatically created. This object must be of type PackageMetadata: 

```move
public struct PackageMetadataV1 has key {
    // PackageMetadata (derived) id
    id: UID,
    // Package id
    package_id: ID,
    // Package version
    package_version: u64, //TODO check max num of versions and change u64 accordingly
    // Package name
    package_name: ascii::String,
    // Package modules
    module_handles: Vec<ascii::String>,
    /// Handles to external and internal functions.
    function_handles: Vec<FunctionHandle>,    
}

public struct FunctionHandle has copy, drop, store {
    /// The module that defines the function.
    module_handle: u16, // index of the module_handles vector  
    /// The name of the function.
    name: ascii::String,
    /// Flag for authenticator functions.
    is_authenticator: bool,
}
```

During the publishing/updating of a package including at least one #[authenticator] attribute,  an immutable PackageMetadataV1 object is automatically created. 

An AuthenticatorInfoV1 can be created using this function from PackageMetadataV1:
```move
    self: &PackageMetadataV1,
    module_name: ascii::String,
    function_name: ascii::String,
): AuthenticatorInfoV1 {
    // Find module
    let module_handle = self.module_handles.find_index!(|m| m == module_name);
    assert!(module_handle.is_some(), EModuleNotFound);

    // Find function
    let function_handle = self.function_handles.find_index!(
      |f| 
      f.module_handle == *module_handle.borrow() 
      && f.name == function_name
      && f.is_authenticator
    );
    assert!(function_handle.is_some(), EFunctionNotFoundOrNotAuthenticator);    
    
    AuthenticatorInfoV1 { package: self.package_id(), module_name, function_name }
}
```

### LINKED ISSUE #8862: [AA] Adapt the iota-execution publish and upgrade flows to create a PackageMetadata immutable object
Similar to the automatic creation of an UpgradeCap, when a package is published or upgraded, a new PackageMetadata immutable object must be created. Draft prototype in [#8736](https://github.com/iotaledger/iota/pull/8736).

Check the possibility and implement the object id derivation of a PackageMetadata as in https://docs.sui.io/concepts/sui-move-concepts/derived-objects.

miker83z:
# Description of change

This PR removes the previous method for verifying the authenticate functions which worked at runtime. Then, it introduces the creation of a Package Metadata immutable object during a package publishing/upgrade. Finally, it allows to use such metadata to create `AuthenticatorInfoV1` instances for accounts.

Fixes #8861 
Fixes #8862 


crates/iota-protocol-config/src/lib.rs
```diff
@@ -2251,6 +2259,7 @@ impl ProtocolConfig {
                         // max auth gas budget is in NANOS and an absolute value 1IOTA
                         cfg.max_auth_gas = Some(1_000_000_000);
                         cfg.feature_flags.move_auth = true;
+                        cfg.feature_flags.publish_package_metadata_v1 = true;
                         // === Native Function Costs ===
                         // `account` module
                         cfg.check_auth_info_v1_cost_base = Some(1000);
```
Dkwcs: As I see this field(`check_auth_info_v1_cost_base`) was used in `check_auth_info_v1` that has been removed.
Where is it using now? Is it usable?

Dkwcs:
Correct me if I'm wrong, are there any possible exploits present in the runtime ?
We verify metadata during the publishing, but can we trust these data on 100% ?

iota-execution/latest/iota-verifier/src/authenticator_verifier.rs
```diff
@@ -138,6 +139,82 @@ pub fn verify_authenticate_func_v1(
...
 }
 
+/// Verify that the first parameter type of the authenticate function matches
+/// the account type.
+fn verify_authenticate_account_type(param: &SignatureToken) -> Result<(), String> {
+    use SignatureToken::*;
+
+    match param {
+        Reference(ref_param) => match &**ref_param {
+            Datatype(_) => Ok(()),
```
Dkwcs: Is it sufficient to check only the `Datatype`?
Previously, we also validated the `account_type` using
`verify_account_type(module, &function_signature.0[0], account_type).`
Perhaps, since the verification logic was moved to the publishing stage, we no longer need to perform this check here? If yes, then the comment a bit confusing for me (`/// Verify that the first parameter type of the authenticate function matches the account type.`)

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,81 @@
...
+public struct PackageMetadataV1 has key {
+    id: UID,
+    /// Storage ID of the package represented by this metadata
+    /// The object id of the runtime package metadata object is derived from
+    /// this value.
+    storage_id: ID,
+    /// Runtime ID of the package represented by this metadata
+    runtime_id: ID,
+    /// Version of the package represented by this metadata
+    package_version: u64,
```
Dkwcs, line 36: Where do we update `package_version` when the package is updated?
Maybe I missed this point

miker83z: When the package is upgraded, a new PackageMetadata object will be created and this `package_version` will be filled with the upgraded version number.

iota-execution/latest/iota-verifier/src/authenticator_verifier.rs
miker83z: Previously we were running this check at runtime, meaning that we had the exact information about the Account object we were trying to attach this authenticate function to. In that case, had a way to check the account type directly against the authenticate first parameter. 

At build time we have no info about the Account type, we just have the info about the type declared as the first parameter of the authenticate function. So, now, we just check that the first parameter is a reference to a Datatype and not other types.

You are right that the comment must be changed. Also, we could make the check more constrained and only allow datatypes that have at least the `key` ability.

Cargo.lock
```diff
@@ -4306,11 +4306,11 @@ checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
 [[package]]
 name = "futures-timer"
 version = "3.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
+source = "git+https://github.com/iotaledger/iota-sim.git?branch=tokio-1.46.1#20adc0b089bf9072268eea0f059fb8ea2461702f"
```
valeriyr: Do we expect this change, or should it be reverted?
I see several links updated to `iota-sim`.

crates/iota-e2e-tests/tests/abstract_account/abstract_account/sources/abstract_account.move
```diff
@@ -45,7 +47,9 @@ public struct AbstractAccount has key {
 ///
 /// The `AuthenticatorInfo` will be attached to the account being built.
 public fun builder(
-    authenticator: AuthenticatorInfoV1<AbstractAccount>,
```
valeriyr, line 50: We discussed an option of creating an account with a framework function:
```rust
public fun create_account_v1<T: key>(obj: T, info: AuthenticatorInfoV1<T>) {
    create_account_impl(obj, info)
}
```
So, maybe it would be better to leave the `AuthenticatorInfoV1` generic parameter and not change the interface of the examples?
We don't really need to use `PackageMetadataV1` here; it makes the account implementation and testing much more complex. AuthenticatorInfoV1<Account> makes sense because we can easily limit the authenticator type that can be attached to an account.
Even if we leave the proof for a while, it will be cleaner to leave the generic parameter.

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -88,23 +84,19 @@ public fun attach_auth_info_v1<Account: key>(
 /// Rotate the account-related authenticator.
 /// The `authenticator` instance will replace the account dynamic field specified by the `AuthenticatorInfoV1Key` name;
 /// the previous value will be returned.
-public fun rotate_auth_info_v1<Account: key>(
-    account_id: &mut UID,
-    proof: AuthenticatorInfoV1CompatibilityProof<Account>,
-): AuthenticatorInfoV1<Account> {
+public fun rotate_auth_info_v1(account_id: &mut UID, proof: AuthenticatorInfoV1CompatibilityProof) {
```
valeriyr: What is the reason for removing the return value?

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,151 @@
...
+/// Returns the `AuthenticatorMetadataV1` associated with the specified
+/// `module_name` and `function_name`, if any.
+public fun authenticator_metadata_v1(
+    self: &ModuleMetadataV1,
+    function_name: ascii::String,
+): Option<AuthenticatorMetadataV1> {
+    self.authenticator_metadata.find_index!(|m| m.function_name == function_name).and!(|index| {
+        option::some(self.authenticator_metadata[index])
+    })
+}
```
valeriyr: It would be nice to have functions with borrow semantics as well, not to create extra copies where it is not necessary

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,151 @@
...
+
+/// Return the account type of the authenticator represented by this metadata
+public fun account_type(self: &AuthenticatorMetadataV1): TypeName {
+    self.account_type
+}
+
+public fun try_get_authenticator_metadata_v1(
```
valeriyr: Here as well;
Do we need this function? It looks pretty straightforward to get AuthenticatorMetadataV1 from PackageMetadataV1 without it.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -357,6 +355,10 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     move_auth: bool,
 
+    // If true, enables publishing package metadata v1 along with the package.
+    #[serde(skip_serializing_if = "is_false")]
+    publish_package_metadata_v1: bool,
```
valeriyr: I would consider using a parameter like the following instead of bool:
```
package_metadata_version: Option<u64>
```

crates/iota-types/src/move_package.rs
```diff
@@ -986,3 +995,157 @@ impl RuntimeModuleMetadataV1 {
...
+    ) -> Self {
+        let mut modules_metadata = VecMap { contents: vec![] };
+
+        for (module_name, fns_metadata) in modules_info.into_iter() {
+            let mut value = ModuleMetadataV1 {
+                authenticator_metadata: vec![],
+            };
+
+            for (fn_name, account_type) in fns_metadata {
+                value.authenticator_metadata.push(AuthenticatorMetadataV1 {
```
valeriyr: Vec can't be used here as is; we need to add validation that it doesn't contain duplicates.
There is no guarantee that it doesn't; in this struct we don't know from where this information is taken.

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -856,6 +886,134 @@ mod checked {
...
     }
 
+    /// Creates package metadata for a Move package by extracting module
+    /// metadata and wrapping it in a `PackageMetadata`. The function iterates
+    /// through the provided modules, collecting metadata associated with
+    /// the IOTA_METADATA_KEY key. It then constructs the package metadata
+    /// wrapper using the collected module metadata, storage ID, runtime ID,
+    /// and package version and finally freezes it. If no relevant metadata
+    /// is found, the function exits without creating any package metadata.
+    fn create_and_freeze_package_metadata_if_present(
```
valeriyr, line 896:
```suggestion
    fn create_and_freeze_package_metadata_v1_if_present(
```

crates/iota-types/src/move_package.rs
```diff
@@ -986,3 +995,157 @@ impl RuntimeModuleMetadataV1 {
...
+/// content, i.e., PackageMetadataV1 is the type used on-chain.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum PackageMetadata {
+    V1(PackageMetadataV1),
+}
+
+impl PackageMetadata {
+    /// Create a `PackageMetadata` for the newly
+    /// published/upgraded package at `package_id`
+    pub fn new(
```
valeriyr:
```suggestion
    pub fn new_v1(
```

valeriyr (changes requested):

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -15,70 +17,64 @@ const EAuthenticatorInfoV1NotAttached: vector<u8> =
...
 
 /// Dynamic field key, where the system will look for a potential
 /// authenticate function.
 public struct AuthenticatorInfoV1Key has copy, drop, store {}
 
 /// Represents a validated authenticate function.
-/// `Account` is a phantom type representing the account type which can be authenticated.
 #[allow(unused_field)]
-public struct AuthenticatorInfoV1<phantom Account: key> has copy, drop, store {
+public struct AuthenticatorInfoV1 has copy, drop, store {
```
Dkwcs: What did we decide about renaming this to more suitable name?
Like `AuthenticatorDescriptor` or similar?

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -15,70 +17,64 @@ const EAuthenticatorInfoV1NotAttached: vector<u8> =
...
-        authenticator,
+        authenticator: AuthenticatorInfoV1 {
+            package: package_metadata.storage_id(),
+            module_name,
+            function_name,
+        },
     }
 }
 
 /// Attach the `authenticator` instance to the account.
```
Dkwcs: I guess comments should correspond the new signature. In this case it's a proof not authenticator instance itself.
We should recheck all the comments.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,151 @@
...
+/// Represents the metadata of a Move package. This includes information
+/// such as the storage ID, runtime ID, version, and metadata for the
+/// functions contained within the package.
+public struct PackageMetadataV1 has key {
+    id: UID,
+    /// Storage ID of the package represented by this metadata
+    /// The object id of the runtime package metadata object is derived from
+    /// this value.
+    storage_id: ID,
+    /// Runtime ID of the package represented by this metadata
```
Dkwcs: Is this comment useful in this context? 
Could we elaborate the meaning or remove it?

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,151 @@
...
+/// An IOTA package can have associated metadata that provides,
+/// on-chain, additional information about the package.
+module iota::package_metadata;
+
+use iota::vec_map::VecMap;
+use std::ascii;
+use std::type_name::TypeName;
+
+/// Key type for deriving the package metadata object address
+public struct PackageMetadataKey has copy, drop, store {}
```
Dkwcs, line 21: Is it should be V1 as well?

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) 2024 IOTA Stiftung
```
Dkwcs:
```suggestion
// Copyright (c) 2025 IOTA Stiftung
```

crates/iota-protocol-config/src/lib.rs
miker83z: Yes, indeed. Removed [1bfa278](https://github.com/iotaledger/iota/pull/9296/commits/1bfa278b57c8615136037c07da5cf712655dd328)

Cargo.lock
miker83z: I pushed the changes after running a sim test and without re-running cargo check or similar. Fixed [9fc51cc](https://github.com/iotaledger/iota/pull/9296/commits/9fc51cc2a5fbe8fbf54a7d1a9f995b143492fcdb)

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: The return was introduced because drop was not used as ability for `AuthenticatorInfoV1 `. So, `rotate_auth_info_v1 ` was never modified after adding `drop` to the abilities.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: [458c47a](https://github.com/iotaledger/iota/pull/9296/commits/458c47ab97f2ddb02f199452dbda2dd6e21dd8a9)

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: [458c47a](https://github.com/iotaledger/iota/pull/9296/commits/458c47ab97f2ddb02f199452dbda2dd6e21dd8a9)

crates/iota-protocol-config/src/lib.rs
miker83z: [6ec1eb0](https://github.com/iotaledger/iota/pull/9296/commits/6ec1eb0371629f439ac5439164a10f2d22e29c36)

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
miker83z: Why? The inner logic can be easily modified to support a v2 in the future, I don't see necessary the creation of  `create_and_freeze_package_metadata_v2_if_present` in that case.

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: Should be done in a different PR!

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: Nope, this can be used without versioning as we won't have the need to have 2 different keys for 2 different versions in the future.

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: The comment is still valid, because that instance is attached. I added "It uses a AuthenticatorInfoV1CompatibilityProof to obtain that instance".

crates/iota-types/src/move_package.rs
miker83z: Moved the creation up, where it is sure that the verifier was run. Then I added a check for duplicates in the verifier [6878cb6](https://github.com/iotaledger/iota/pull/9296/commits/6878cb68455603c2f7bfce53da5a158a5209c8ce)

crates/iota-e2e-tests/tests/abstract_account/abstract_account/sources/abstract_account.move
miker83z: To talk about

crates/iota-framework/packages/iota-framework/sources/account.move
valeriyr: Not really, `drop` was there from the very beginning.
I added the return value to reflect the standard `remove` notation that we have in the framework when the previous value is returned. It makes it possible to use it somehow if it is necessary, for events, for example, or it will be just dropped automatically anyway if it is not used.

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
valeriyr: ok, should this function have a version or protocol config as a parameter in this case?
Without this, it is not clear from the interface which version is created.
I agree that this parameter can be added later.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: Updated here [6878cb6](https://github.com/iotaledger/iota/pull/9296/commits/6878cb68455603c2f7bfce53da5a158a5209c8ce)

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
miker83z: Yes it can have the protocol config as parameter. I would not add this now, but later.

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: We should add the events in any case within the `account.move` functions. The idea here (following the other changes as well) was also to hide `AuthenticatorInfoV1` from the users, they don't really need to handle it.

miker83z: [25c2bcf](https://github.com/iotaledger/iota/pull/9296/commits/25c2bcfe896ff4788e4bdc96df60d518b20f985e)

crates/iota-e2e-tests/tests/abstract_account/abstract_account/sources/abstract_account.move
miker83z: [25c2bcf](https://github.com/iotaledger/iota/pull/9296/commits/25c2bcfe896ff4788e4bdc96df60d518b20f985e)

crates/iota-types/src/move_package.rs
miker83z: Refactored [c01c0c0](https://github.com/iotaledger/iota/pull/9296/commits/c01c0c08ad394d6a481d828ad83d9208e1082cf8)

crates/iota-framework/packages/iota-framework/sources/account.move
```diff
@@ -49,13 +53,22 @@ public struct AuthenticatorInfoV1CompatibilityProof<phantom Account: key> has dr
 ///
 /// This function cannot be used in `move unit tests` as there is no mechanism to refer to the package being tested.
 public fun create_auth_info_v1<Account: key>(
-    package: address,
+    package_metadata: &PackageMetadataV1,
     module_name: ascii::String,
     function_name: ascii::String,
 ): AuthenticatorInfoV1<Account> {
```
valeriyr, line 59: As an option, we can consider moving this function to the `PackageMetadataV1` module.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,171 @@
...
+public fun try_get_modules_metadata_v1(
+    self: &PackageMetadataV1,
+    module_name: &ascii::String,
+): Option<ModuleMetadataV1> {
+    self.modules_metadata.try_get(module_name)
+}
+
+/// Borrow the module metadata list of the package represented by this metadata.
+/// Aborts if the module is not found.
+public fun modules_metadata_v1(
```
valeriyr, line 80: Should be renamed to `borrow_modules_metadata_v1` for consistency?

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,171 @@
...
+): Option<AuthenticatorMetadataV1> {
+    self.authenticator_metadata.find_index!(|m| m.function_name == *function_name).and!(|index| {
+        option::some(self.authenticator_metadata[index])
+    })
+}
+
+/// Borrow the `AuthenticatorMetadataV1` associated with the specified
+/// `function_name`.
+/// Aborts if the authenticator metadata is not found for that function.
+public fun authenticator_metadata_v1(
```
valeriyr, line 102: Should be renamed to `borrow_ authenticator_metadata_v1` for consistency?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -357,6 +355,10 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     move_auth: bool,
 
+    // If true, enables publishing package metadata v1 along with the package.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    publish_package_metadata: Option<u64>,
```
valeriyr:
```suggestion
    // Package metadata version, defines the package metadata version published with the package.
    #[serde(skip_serializing_if = "Option::is_none")]
    publish_package_metadata_version: Option<u64>,
```
`FeatureFlags` contains only boolean flags, so it probably makes sense to move this property to the `ProtocolConfig`

crates/iota-types/src/derived_object.rs
```diff
@@ -0,0 +1,134 @@
...
+where
+    T: Into<IotaAddress>,
+{
+    let parent_address = parent.into();
+
+    // Wrap `T` into `DerivedObjectKey<T>` type (to preserve on-chain namespacing)
+    let wrapper_type_tag = TypeTag::Struct(Box::new(StructTag {
+        address: IOTA_FRAMEWORK_ADDRESS,
+        module: Identifier::new("derived_object").unwrap(),
+        name: Identifier::new("DerivedObjectKey").unwrap(),
```
valeriyr: Should we create constants for this for consistency?

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -856,6 +886,136 @@ mod checked {
...
+            }
+        }
+
+        // Only publish package metadata if there is at least one module with
+        // relevant metadata
+        if !modules_metadata_map.is_empty() {
+            // Create the package metadata "special" object UID
+            let metadata_uid = context.package_derived_metadata_id(storage_id)?;
+            // Create the package metadata object content
+            let metadata_wrapper = PackageMetadata::new_v1(
```
valeriyr:
```suggestion
            let metadata = PackageMetadata::new_v1(
```
Not so important, of course, you can ignore it if you wish. We have a different metadata wrapper struct, so it would probably be a bit clearer.

valeriyr (approved):
LGTM!

crates/iota-framework/packages/iota-framework/sources/account.move
miker83z: `AuthenticatorInfoV1` can't be easily instantiated in a different module. I would need an additional type to be created in package_metadata.move and then pass it here.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: I am following the approach used in iota_system and other framework modules where the function with the field name returns a read-only reference to that field. Yes, I was also thinking that `borrow_` is consistent with the rest of the functions in this module, but I think it won't be consistent with the rest of the iota framework modules.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: https://github.com/iotaledger/iota/pull/9296#discussion_r2564833553

crates/iota-protocol-config/src/lib.rs
miker83z: So I would suggest to revert to a `publish_package_metadata: bool` because it is indeed a flag to enable a feature. In the future, we'll have a bool for v2, if necessary, like  `native_charging_v2` and `consensus_linearize_subdag_v2`. From the protocol config perspective, the version is not really important (at least for now). We could handle it internally like `Transaction` or `TransactionEffects`.

valeriyr: I would prefer to have a version in the config, but I don't insist; you can decide what option to leave

crates/iota-types/src/move_package.rs
```diff
@@ -986,3 +995,164 @@ impl RuntimeModuleMetadataV1 {
...
+        StructTag {
+            address: IOTA_FRAMEWORK_ADDRESS,
+            module: PACKAGE_METADATA_MODULE_NAME.to_owned(),
+            name: PACKAGE_METADATA_V1_STRUCT_NAME.to_owned(),
+            type_params: vec![],
+        }
+    }
+
+    pub fn to_bcs_bytes(&self) -> Vec<u8> {
+        bcs::to_bytes(&self).unwrap()
```
Dkwcs, line 1141: Is it safe here? If yes, would you leave some comment as you did before?

Dkwcs (approved):

crates/iota-types/src/move_package.rs
miker83z: [40b2281](https://github.com/iotaledger/iota/pull/9296/commits/40b22810f79420530d390d12c9a21d251a774535)

crates/iota-protocol-config/src/lib.rs
miker83z: [40b2281](https://github.com/iotaledger/iota/pull/9296/commits/40b22810f79420530d390d12c9a21d251a774535)

crates/iota-types/src/derived_object.rs
miker83z: [40b2281](https://github.com/iotaledger/iota/pull/9296/commits/40b22810f79420530d390d12c9a21d251a774535)

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
miker83z: [40b2281](https://github.com/iotaledger/iota/pull/9296/commits/40b22810f79420530d390d12c9a21d251a774535)

PR #9297, opened 2025-11-15
feat: add retriers on create instance

### LINKED ISSUE #9274: retry deployement requsts

filipdulic:
# Description of change

The `run_instances` aws command fails to return the minimum number of instances for both spot and on-demand. It needs to be retried a few times to collect the appropriate amount.

- requests spot instances in a loop for 5 minutes attempting to collect as many available ones as needed.
- requests on-demand instances in a loop until the full quantity of required instances is not collected.
- displays a status of the time left in the spot instance loop as well as the number of instances collected.
- added the `assets/run_command.sh` script to execute a command on every Node instance. 
- adds `RUSTFLAGS=-Ctarget-cpu=native` build flag to optimize for CPU in test available instructions.

fixes #9274.

### DIFF
--- a/crates/iota-aws-orchestrator/assets/run_command.sh
+++ b/crates/iota-aws-orchestrator/assets/run_command.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+# Script to execute commands on all testbed Node machines
+# Usage: ./run_command.sh <command> [args...]
+# example: run_command.sh \
+# 'ADDR=$(grep -m 1 admin node.log | sed -E "s/.*address=([^ ]+).*/\1/"); \
+# curl -X POST "http://$ADDR/spammer/start?tps=20&mean_size=30000&std_dev_size=3000"'
+# collects the admin address from the node.log and executes the curl admin command
+
+set -e
+
+# Check if at least one argument is provided
+if [ $# -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


crates/iota-aws-orchestrator/assets/run_command.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Script to execute commands on all testbed Node machines
+# Usage: ./orchestrate-command.sh <command> [args...]
```
polinikita: Correct Usage  to run_command.sh

crates/iota-aws-orchestrator/assets/run_command.sh
```diff
@@ -0,0 +1,85 @@
...
+# example: run_command.sh \
```
polinikita, line 5: Is this the only command that can be executed (it will stay on a feature branch) or there are other use cases?

filipdulic: any command can be executed, this is just the command I used.

piotrm50 (approved):

piotrm50 (approved):

NaitsabesMue (approved):

polinikita (approved):

PR #9304, opened 2025-11-17
feat: use `SenderSignedData::input_objects` instead of `TransactionData::input_objects`

### LINKED ISSUE #9278: [AA] Check all codebase places where the `input_objects` (shared + owned objs) are used and consider usage of authenticator

valeriyr:
# Description of change

Replace the function usage in several places, but there are some that still need to be checked:
1. Indexing.
2. Replays.

fixes #9278

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -889,17 +889,14 @@ impl AuthorityState {
 
         let tx_data = transaction.data().transaction_data();
 
-        let tx_and_auth_input_objects = transaction.input_objects()?.collect::<Vec<_>>();
-        let tx_receiving_objects = tx_data.receiving_objects();
-
         // Note: the deny checks may do redundant package loads but:
         // - they only load packages when there is an active package deny map
         // - the loads are cached anyway
         iota_transaction_checks::deny::check_transaction_for_signing(
             tx_data,
             transaction.tx_signatures(),
-            &tx_and_auth_input_objects,
-            &tx_receiving_objects,
+            &tx_data.input_objects()?,
+            &tx_data.receiving_objects(),
             &self.config.transaction_deny_config,
             self.get_backing_package_store().as_ref(),
         )?;
@@ -908,13 +905,8 @@ impl AuthorityState {
         // Authenticator input objects and the account object are loaded in the same
         // call if there is a sender `MoveAuthenticator` signature present in the
         // transaction.
-        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) = self
-            .read_objects_for_signing(
-                &transaction,
-                &tx_and_auth_input_objects,
-                &tx_receiving_objects,
-                epoch,
-            )?;
+        let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) =
+            self.read_objects_for_signing(&transaction, epoch)?;
 
         // Get the sender `MoveAuthenticator`, if any.
         // Only one `MoveAuthenticator` signature is possible, since it is not
@@ -1321,7 +1313,7 @@ impl AuthorityState {
             .execution_load_input_objects_latency
             .start_timer();
 
-        let input_objects = certificate.input_objects()?.collect::<Vec<_>>();
+        let input_objects = certificate.collect_all_inputs_for_reading()?;
 
         let input_objects = self.input_loader.read_objects_for_execution(
             epoch_store,
@@ -1331,7 +1323,7 @@ impl AuthorityState {
             epoch_store.epoch(),
         )?;
 
-        certificate.split_inputs_into_groups(input_objects)
+        certificate.split_inputs_into_groups_for_reading(input_objects)
     }
 
     /// Test only wrapper for `try_execute_immediately()` above, useful for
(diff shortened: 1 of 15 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -5444,6 +5455,8 @@ impl AuthorityState {
     fn read_objects_for_signing(
```
miker83z, line 5447: In this file, in the function `fn index_tx` there's:
```rust
cert.data()
    .intent_message()
    .value
    .input_objects()?
```

miker83z (approved):
Please also write a comment in the indexer and replay crates where it would be needed to include the MoveAuthenticator inputs. This will be done later.

crates/iota-core/src/authority.rs
valeriyr: I haven't touched the places related to indexing in this PR because, in terms of the indexer, transaction inputs are always associated with TransactionData, so if this PR works, it is probably better to refactor in a dedicated PR.

valeriyr:
> Please also write a comment in the indexer and replay crates where it would be needed to include the MoveAuthenticator inputs. This will be done later.

It is easy to find the places where `TransactionDataAPI::input_objects` is used. The problem is that it is not always possible to just replace `TransactionData::input_objects` with `SenderSignedData::input_objects` because this data will not be interpreted correctly. So the best option will be to check this with the responsible team and create related issues, I think.

crates/iota-types/src/transaction.rs
```diff
@@ -2499,122 +2516,171 @@ impl SenderSignedData {
...
-                .into_iter()
-                .collect::<HashSet<_>>();
+    /// Returns all unique input objects including those from the sender
+    /// `MoveAuthenticator` if any for reading.
+    ///
+    /// Although some shared objects(with a different mutability flag, for
+    /// example) can be duplicated in the transaction and authenticator
+    /// object lists, we load them independently to make it possible to
+    /// analyze the inputs in the transaction checkers.
+    pub fn collect_all_inputs_for_reading(&self) -> IotaResult<Vec<InputObjectKind>> {
```
miker83z:
```suggestion
    pub fn collect_all_input_object_kind_for_reading(&self) -> IotaResult<Vec<InputObjectKind>> {
```

crates/iota-types/src/transaction.rs
```diff
@@ -2499,122 +2516,171 @@ impl SenderSignedData {
...
 
     /// Splits the provided input objects into three groups:
-    /// 1. Input objects required by the transaction itself.
+    /// 1. Input objects required by the transaction itself; may contain
+    ///    duplicates if an IOTA coin is used both as an input and a gas coin.
     /// 2. Input objects required by the sender `MoveAuthenticator`, including
     ///    the object to authenticate.
     /// 3. The object to authenticate from the sender `MoveAuthenticator`.
-    pub fn split_inputs_into_groups(
+    pub fn split_inputs_into_groups_for_reading(
```
miker83z:
```suggestion
    pub fn split_input_objects_into_groups(
```

crates/iota-types/src/transaction.rs
```diff
@@ -1290,6 +1290,23 @@ impl SharedInputObject {
     pub fn into_id_and_version(self) -> (ObjectID, SequenceNumber) {
         (self.id, self.initial_shared_version)
     }
+
+    /// Merges another SharedInputObject into self.
+    /// If there is a conflict in mutability, the resulting object will be
+    /// mutable. Errors if the id or initial_shared_version do not match.
+    pub fn union(&mut self, other: &SharedInputObject) -> UserInputResult<()> {
```
miker83z:
```suggestion
    pub fn left_union(&mut self, other: &SharedInputObject) -> UserInputResult<()> {
```

crates/iota-types/src/transaction.rs
```diff
@@ -3047,6 +3113,44 @@ impl InputObjectKind {
             Self::SharedMoveObject { mutable, .. } => *mutable,
         }
     }
+
+    /// Merges another InputObjectKind into self.
+    /// For shared objects, if either is mutable, the result is mutable.
+    /// Fails if either is not a shared object, or if the IDs or initial
+    /// versions do not match.
+    pub fn union(&mut self, other: &InputObjectKind) -> UserInputResult<()> {
```
miker83z:
```suggestion
    pub fn left_union_with_checks(&mut self, other: &InputObjectKind) -> UserInputResult<()> {
```

miker83z (approved):
Just some nit renamings

kodemartin (approved):
lgtm for the changes in `iota-analytics-indexer`

crates/iota-types/src/transaction.rs
Dkwcs: As an option we can consider `merge` for the function name.

Dkwcs (approved):

crates/iota-types/src/transaction.rs
valeriyr: renamed in a51f8f4b59

crates/iota-types/src/transaction.rs
valeriyr: Renamed in a51f8f4b59
But I left the `_for_reading` postfix at least for now. Since we have several input objects set, this postfix shows where the inputs should be used.

crates/iota-types/src/transaction.rs
valeriyr: Renamed in a51f8f4b59

crates/iota-types/src/transaction.rs
valeriyr: Renamed in a51f8f4b59

PR #9306, opened 2025-11-17
feat: add `MoveAuthenticator::validity_check`

### LINKED ISSUE #9105: [AA] Create a `MoveAuthenticator` `validity_check`
This validity check should be similar to `ProgrammableTransaction::validity_check` however, should be split in 2 phases:
- [ ] the inputs check can happen somewhere around [here `SenderSignedData::validity_check`](https://github.com/iotaledger/iota/blob/10abe7c258a6c99979ee0dd0a2dfaffedae42de6/crates/iota-types/src/transaction.rs#L2409).
- [ ] the command check can only happen when the authenticate function is marked as "valid" by the iota-verifier; this because otherwise the user could be left out from their account if a check fails during execution.

valeriyr:
[run-ci]

# Description of change

`MoveAuthenticator::validity_check` implementation.

fixes #9105

### DIFF
--- a/crates/iota-types/src/move_authenticator.rs
+++ b/crates/iota-types/src/move_authenticator.rs
@@ -153,8 +153,10 @@ impl MoveAuthenticator {
 
         // `ProtocolConfig::max_arguments` is used to check the call arguments because
         // MoveAuthenticator is considered as a simple programmable Move call.
+        //
+        // The limit includes the object to authenticate, so we subtract 1 here.
         fp_ensure!(
-            self.call_args().len() < (config.max_arguments() as usize),
+            self.call_args().len() < (config.max_arguments() - 1) as usize,
             UserInputError::SizeLimitExceeded {
                 limit: "maximum arguments in MoveAuthenticator".to_string(),
                 value: config.max_arguments().to_string()
--- a/crates/iota-types/src/move_authenticator.rs
+++ b/crates/iota-types/src/move_authenticator.rs
@@ -106,8 +106,7 @@ impl MoveAuthenticator {
                         return Err(UserInputError::Unsupported(
                             "MoveAuthenticator cannot authenticate mutable shared objects"
                                 .to_string(),
-                        )
-                        .into());
+                        ));
                     }
 
                     (*id, None, None)
--- a/crates/iota-types/src/move_authenticator.rs
+++ b/crates/iota-types/src/move_authenticator.rs
@@ -8,7 +8,6 @@ use std::{
 
 use fastcrypto::{error::FastCryptoError, traits::ToFromBytes};
 use iota_protocol_config::ProtocolConfig;
-use move_core_types::language_storage::TypeTag;
 use once_cell::sync::OnceCell;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -23,6 +22,7 @@ use crate::{
     signature::{AuthenticatorTrait, VerifyParams},
     signature_verification::VerifiedDigestCache,
     transaction::{CallArg, InputObjectKind, ObjectArg, SharedInputObject},
+    type_input::TypeInput,
 };
 
 /// MoveAuthenticator is a GenericSignature variant that enables a new
@@ -35,7 +35,7 @@ pub struct MoveAuthenticator {
     call_args: Vec<CallArg>,
     /// Type arguments for the Move authenticate function
     #[schemars(with = "String")]
-    type_arguments: Vec<TypeTag>, // TypeInput???
+    type_arguments: Vec<TypeInput>,
     /// The object that is authenticated. Represents the account being the
     /// sender of the transaction.
     object_to_authenticate: CallArg,
@@ -56,7 +56,7 @@ impl Hash for MoveAuthenticator {
 impl MoveAuthenticator {
     pub fn new_for_testing(
         call_args: Vec<CallArg>,
-        type_arguments: Vec<TypeTag>,
+        type_arguments: Vec<TypeInput>,
         object_to_authenticate: CallArg,
     ) -> Self {
         Self {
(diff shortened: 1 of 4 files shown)


crates/iota-types/src/transaction.rs
```diff
@@ -2653,6 +2624,53 @@ impl SenderSignedData {
...
+                        .to_string(),
+                )
+                .into()
+            );
+
+            // TODO(https://github.com/iotaledger/iota/issues/8966): The following
+            // restrictions are temporary added until we implement `MoveAuthenticator`
+            // support for sponsors.
+
+            fp_ensure!(
```
Dkwcs, line 2654: It's extremely minor, however I would put this check above all others, I think semantically it looks better
Because at start we have `if authenticators_num > 0` and then we're narrowing this check to  authenticators_num == 1.
Up to you

Dkwcs (approved):

crates/iota-types/src/transaction.rs
```diff
@@ -2501,6 +2470,8 @@ impl SenderSignedData {
             .validity_check(config)
             .map_err(Into::<IotaError>::into)?;
 
+        self.move_authenticators_validity_check(tx_data, config)?;
```
Dkwcs, line 2473: Also, would it be better to check for the existence of authenticators beforehand?

Like if auth_num > 0 -> move_authenticators_validity_check

crates/iota-types/src/transaction.rs
valeriyr: As mentioned in the previous comment, these are conditions that were temporarily added until we implement MoveAuthenticator support for sponsors.
These conditions are checked in the `if authenticators_num > 0 { ... }` block because they make sense only if we have such signatures. So I think we should leave it as is and will clean it when the sponsor's support is implemented.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -142,6 +140,55 @@ impl MoveAuthenticator {
...
+
+        // Inputs validity check.
+        //
+        // `validity_check` is not called for `object_to_authenticate` because it is
+        // already validated with a dedicated function.
+
+        // `ProtocolConfig::max_arguments` is used to check the call arguments because
+        // MoveAuthenticator is considered as a simple programmable Move call.
+        fp_ensure!(
+            self.call_args().len() < (config.max_arguments() as usize),
```
miker83z: Should be `config.max_arguments() - 1` considering that the account object is not part of `call_args` but it is used as input as well.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -142,6 +140,55 @@ impl MoveAuthenticator {
...
+            UserInputError::SizeLimitExceeded {
+                limit: "maximum arguments in MoveAuthenticator".to_string(),
+                value: config.max_arguments().to_string()
+            }
+        );
+
+        // TODO: should we handle duplicate inputs somehow?
```
miker83z: I don't see it in the PTB validity_check. Probably this check is happening in iota-transaction-checks also for PTBs, could you check?

crates/iota-types/src/transaction.rs
valeriyr: It is checked inside; there is no reason to duplicate the checked

crates/iota-types/src/move_authenticator.rs
valeriyr: fixed in da847ea9c7

crates/iota-types/src/move_authenticator.rs
valeriyr: We can find several places where inputs are checked for duplication:
https://github.com/iotaledger/iota/blob/da847ea9c76a56128b0b509bc27bf54c1a07109a/crates/iota-types/src/transaction.rs#L1456
https://github.com/iotaledger/iota/blob/da847ea9c76a56128b0b509bc27bf54c1a07109a/crates/iota-types/src/transaction.rs#L1057
As I understand, this check makes sense because `ProgrammableTransaction` holds all the inputs as `CallArg`s while `ProgrammableMoveCall` commands refer to these inputs using `Argument`, so if you want to use the same argument twice, you don't need to have it twice in the list.

valeriyr: We also have a PTB inputs limit check:
https://github.com/iotaledger/iota/blob/da847ea9c76a56128b0b509bc27bf54c1a07109a/crates/iota-types/src/transaction.rs#L1089-L1096
https://github.com/iotaledger/iota/blob/da847ea9c76a56128b0b509bc27bf54c1a07109a/crates/iota-protocol-config/src/lib.rs#L478-L480
Where `MoveAuthenticator::input_objects` are not included.

crates/iota-types/src/transaction.rs
Dkwcs: I was suggesting to move it out, but feel free to keep it as you think is best.

crates/iota-types/src/move_authenticator.rs
miker83z: Related to your first comment It makes sense to add the duplication check within `self.call_args()` here then. 
The second comment is something broader to deal with in beta -> #9334

miker83z (commented):
Could you also update the branch and check the state of the newly added AA tests in `iota-adapter-tests`?

crates/iota-types/src/move_authenticator.rs
valeriyr: I can't say that it is a common case, but adding a duplication check means that we can't use in an auhenticator call the same object twice:
```rust
public fun authenticate(o1: &Object, o2: &Object, _: &AuthContext, _: &TxContext) {}

authenticate(0x1, 0x1, _, _)
```
Do we want to introduce this limitation?

valeriyr:
> Could you also update the branch and check the state of the newly added AA tests in `iota-adapter-tests`?

The snapshot is updated: 3634be4161

crates/iota-types/src/move_authenticator.rs
valeriyr: Implemented here: 80a33f232c

miker83z (approved):

PR #9310, opened 2025-11-17
feat(consensus): add traversed headers tracker

### LINKED ISSUE #9302: Change logic for committing transactions
Currently, it is possible that we commit transactions from block B, but don't traverse the header and don't include the header in any commit. We want to change the logic for committing transactions so that can't happen.

VorobyevIlya:
# Description of change

PR adds traversed headers tracker in Linearizer. Now, when we are trying to commit transactions, we are checking that the corresponding header was actually traversed. It guarantees us that if for some block transactions are committed, then the corresponding header is also committed.

A protocol parameter was added for this change. It is enabled since version 16 in all networks (but it affects only Starfish which is run only on Devnet now).

fixes #9302 

### Release Notes

- [x] Protocol: logic for committing transactions in Starfish was changed
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/starfish/core/src/commit_observer.rs
+++ b/crates/starfish/core/src/commit_observer.rs
@@ -183,7 +183,7 @@ impl CommitObserver {
                 .leader
                 .round;
             self.linearizer
-                .evict_old_acknowledgments(max_solid_commit_leader_round);
+                .evict_old_acknowledgments_and_headers(max_solid_commit_leader_round);
         }
         tracing::trace!("Committed & sent {sent_sub_dags:#?}");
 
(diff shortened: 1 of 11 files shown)


crates/starfish/core/src/commit_observer.rs
```diff
@@ -183,7 +183,7 @@ impl CommitObserver {
                 .leader
                 .round;
             self.linearizer
-                .evict_old_acknowledgments(max_solid_commit_leader_round);
+                .evict_old_acknowledgments_and_headers(max_solid_commit_leader_round);
```
polinikita: Maybe it would be better to call it evict_linearizer() or evict_state()

muXxer (approved):

polinikita (approved):

Thoralf-M (approved):
Approving the changes for the files owned by dev-tools

PR #9320, opened 2025-11-18
fix: RUSTSEC-2025-0119 advisory

thibault-martinez:
# Description of change

Fixes https://rustsec.org/advisories/RUSTSEC-2025-0119.html


Thoralf-M (approved):

marc2332 (approved):
~doesn't count but I give it anyway~

Actually it does?

semenov-vladyslav (approved):

PR #9324, opened 2025-11-18
feat(authority): make simulate_transaction work with both dry_run and dev_inspect

### LINKED ISSUE #9323: [core,execution] Add Authority's simulate_transaction support for both dry_run and dev_inspect
Based on some upstream changes: https://github.com/MystenLabs/sui/pull/21918

miker83z:
# Description of change

Changed the implementation of `simulate_transaction` in the authority. Now it supports the execution of both DRY RUN (when passing `VmChecks::ENABLED`) and DEV INSPECT (when passing `VmChecks::DISABLED`).

The REST API was not changed and a dry run was hardcoded for the `simulate_transaction` API (as it was working before). Not sure if this is needed to be changed as well. If yes, `TransactionSimulationResponse` should be changed to include `execution_result` and `SimulateTransactionQueryParameters` should be changed to include the `VmChecks` param.

The current tests should check that the `authority::simulate_transaction` dry run path works correctly. No new tests were added for the dev inspect path. If we don't change the REST API now, these new tests could be written once the gRPC is ready.

Fixes #9323 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -110,7 +110,7 @@ use iota_types::{
         ProtocolConfig, SupportedProtocolVersions, SupportedProtocolVersionsWithHashes,
     },
     transaction::*,
-    transaction_executor::SimulateTransactionResult,
+    transaction_executor::{SimulateTransactionResult, VmChecks},
 };
 use itertools::Itertools;
 use move_binary_format::{CompiledModule, binary_config::BinaryConfig};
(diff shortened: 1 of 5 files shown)


muXxer (approved):
Looks good to me. We don't need to do changes to the existing REST API.

I only wonder why our gas value is 1000x bigger than on upstream.

muXxer (approved):

bingyanglin (approved):

valeriyr (approved):

PR #9325, opened 2025-11-18
feat(consensus): introduce network latencies in orchestrator

### LINKED ISSUE #9125: Try to introduce network latencies when running AWS network in a single region
We've introduced artificial latencies to mimic geodistribution of nodes in a local network. Try to do the same in AWS network.

filipdulic:
# Description of change

## Overview
Implements simulated latencies between nodes in the same private network. The latency matrix can be manipulated via multiple options on the `benchmark` command.

## `benchmark` command options
- adds option `--max_latencies` which defines the maximum round trip value between two nodes in the private network, defaults to 400ms.
- adds option `--number_of_clusters` which groups multiple nodes into the defined number of clusters, defaults to None
- adds option `--latency-perturbation-spec` which for now only has one option `broken-triangle`, a number of broken triangles is introduced into the latency matrix.
- adds option `--number_of_triangles` defines the number of broken triangles to introduce if the `broken-triangle` `--latency-perturbation-spec` option is selected, default is 5.
- adds option `--added-latency` defines by how many milliseconds should the broken triangle latency be delayed, default is 20ms.
- adds option `protocol_switch_each_epoch` which specifies that the node should switch between `mysticeti` and `starfish` protocols on each epoch.

fixes #9125.

- [] Patch-specific tests (correctness, functionality coverage)

### DIFF
--- a/crates/iota-aws-orchestrator/src/net_latency/mod.rs
+++ b/crates/iota-aws-orchestrator/src/net_latency/mod.rs
@@ -1,5 +1,15 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{client::Instance, net_latency::latency_matrix_builder::LatencyMatrixBuilder};
+
 pub mod latency_matrix_builder;
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub enum TopologyLayout {
     /// All Nodes are distributed with their own latencies, no clusters
     Geographical,
@@ -7,6 +17,7 @@ pub enum TopologyLayout {
     Clustered { number_of_clusters: usize },
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub enum PerturbationSpec {
     /// No Perturbation introduced
     None,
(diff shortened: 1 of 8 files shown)


VorobyevIlya (approved):

crates/iota-aws-orchestrator/src/net_latency/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    cmd.push_str(&format!(
+        "sudo tc class add dev {iface} parent 1: classid 1:1 htb rate 1gbit && "
+    ));
+
+    // Add one netem band per IP
+    for (i, (instance, latency)) in latency_vector.iter().enumerate() {
+        let ip = instance.private_ip;
+        let handle = i + 10; // avoid conflict with default bands
+        cmd.push_str(&format!(
+            "sudo tc class add dev {iface} parent 1:1 classid 1:{handle} htb rate 1gbit && "
```
polinikita, line 57: what means 1gbit here? The aws network is usually provided with 10gbit/sec

crates/iota-aws-orchestrator/src/net_latency/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+        ));
+        cmd.push_str(&format!(
+            "sudo tc qdisc add dev {iface} parent 1:{handle} handle {handle}: netem delay {latency}ms && "
```
polinikita, line 60: the code here without any jitter, right?

polinikita (approved):

crates/iota-aws-orchestrator/src/net_latency/mod.rs
filipdulic: minimum guarantied rate in that class, has nothing to do with the NIC speed.

PR #9328, opened 2025-11-19
feat(gRPC): add `GetObjects` implementation in gRPC

### LINKED ISSUE #9390: gRPC: Add `GetObjects` endpoint
Implement the `GetObjects` endpoint in the iota-grpc-server crate.

bingyanglin:
# Description of change

- Implemented `GetObjects` in the gRPC API.
- Verified the `get_objects` simtest by using `cargo simtest --package iota-e2e-tests --test grpc -- get_objects`
  
Closes #9390 

### DIFF
--- a/crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
+++ b/crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
@@ -391,3 +391,67 @@ async fn get_objects_empty_request() {
         "has_next should be false for empty request"
     );
 }
+
+#[sim_test]
+async fn get_objects_nonexistent() {
+    let test_cluster = TestClusterBuilder::new()
+        .with_fullnode_enable_grpc_api(true)
+        .build()
+        .await;
+
+    let mut grpc_client = LedgerServiceClient::connect(test_cluster.grpc_url())
+        .await
+        .unwrap();
+
+    // Request objects that don't exist
+    let responses = assert_get_objects_request(
+        &mut grpc_client,
+        vec![
+            ObjectRequest {
+                object_ref: Some(ObjectReference {
+                    object_id: Some("0xdead".to_string()),
+                    version: None,
+                    digest: None,
+                }),
+            },
+            ObjectRequest {
+                object_ref: Some(ObjectReference {
+                    object_id: Some("0xbeef".to_string()),
+                    version: None,
+                    digest: None,
+                }),
+            },
+        ],
+        None,
+        None,
+        &[], // Skip field mask validation for error responses
+        "non-existent objects",
+    )
+    .await;
+
+    // Verify all results contain errors (not objects)
+    let mut error_count = 0;
+    for response in &responses {
+        for obj_result in &response.objects {
+            assert!(
+                obj_result.error_opt().is_some(),
+                "Expected error for non-existent object"
+            );
+            assert!(
+                obj_result.object_opt().is_none(),
+                "Expected no object for non-existent object"
+            );
+
+            let error = obj_result.error_opt().unwrap();
+            // Verify error has a non-zero code (indicating an actual error)
+            assert!(
+                error.code != 0,
+                "Error should have non-zero code, got: {}",
+                error.code
+            );
+            error_count += 1;
+        }
+    }
+
+    assert_eq!(error_count, 2, "Should receive 2 errors");
+}
--- a/crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
+++ b/crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
@@ -19,11 +19,11 @@ use iota_types::base_types::ObjectID;
 use prost_types::FieldMask;
 use test_cluster::TestClusterBuilder;
 
-use crate::{impl_field_presence_checker, utils::assert_field_presence};
+use crate::{impl_field_presence_checker, utils::assert_nested_field_masks};
 
 // Generate the FieldPresenceChecker implementation for Object
 impl_field_presence_checker!(Object, {
-    "reference" => reference,
+    "reference" => reference [nested],
     "bcs" => bcs,
 });
 
@@ -34,33 +34,12 @@ impl_field_presence_checker!(ObjectReference, {
     "digest" => digest,
 });
 
-/// Helper function to assert object fields based on expected fields
-fn assert_object_fields(
-    object: &Object,
-    expected_top_level: &[&str],
-    expected_reference: &[&str],
-    scenario: &str,
-) {
-    // Check top-level fields
-    assert_field_presence(object, expected_top_level, scenario);
-
-    // If reference is expected, check nested fields
-    if expected_top_level.contains(&"reference") {
-        if let Some(ref reference) = object.reference {
-            assert_field_presence(reference, expected_reference, scenario);
-        } else {
-            panic!("Expected reference field in {scenario}, but it was None");
-        }
-    }
-}
-
 async fn assert_get_objects_request(
     client: &mut LedgerServiceClient<tonic::transport::Channel>,
     requests: Vec<ObjectRequest>,
     read_mask: Option<FieldMask>,
     max_message_size_bytes: Option<u32>,
-    expected_top_level: &[&str],
-    expected_reference: &[&str],
+    expected_field_mask_paths: &[&str],
     scenario: &str,
 ) -> Vec<GetObjectsResponse> {
     let request = GetObjectsRequest {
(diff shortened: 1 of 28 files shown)


crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
```diff
@@ -0,0 +1,436 @@
...
+) -> GetObjectsResponse {
+    let request = GetObjectsRequest {
+        requests: Some(ObjectRequests { requests }),
+        read_mask,
+        max_message_size_bytes,
+    };
+
+    let mut stream = client.get_objects(request).await.unwrap().into_inner();
+
+    let response = stream.next().await.unwrap().unwrap();
```
muXxer: Wouldn't we need to loop until there is no `has_next` in the response?

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
```diff
@@ -0,0 +1,436 @@
...
+        .build()
+        .await;
+
+    let mut grpc_client = LedgerServiceClient::connect(test_cluster.grpc_url())
+        .await
+        .unwrap();
+
+    let object_id = ObjectID::from_hex_literal("0x5").unwrap().to_string();
+
+    // Table-driven tests for single-object readmask scenarios
```
muXxer: Table-driven tests sound a bit like an AI comment.
```suggestion
    // Tests for single-object readmask scenarios
```

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
muXxer: Right now we always pass "false". Maybe we should remove that field, and return a vector of all `GetObjectResponse`?

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
```diff
@@ -0,0 +1,436 @@
...
+    "object_id" => object_id,
+    "version" => version,
+    "digest" => digest,
+});
+
+/// Helper function to assert object fields based on expected fields
+fn assert_object_fields(
+    object: &Object,
+    expected_top_level: &[&str],
+    expected_reference: &[&str],
```
muXxer: This function could also accept the field masks for top level and reference in one, and do the same logic as the field masks themselves (split by the . in the string). That should be more generic and easier to use, because we pass the same values to the field mask and the assert.

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
```diff
@@ -0,0 +1,436 @@
...
+                successful_objects += 1;
+                assert!(object.reference.is_some(), "object should have reference");
+                assert!(object.bcs.is_some(), "object should have bcs");
+                let reference = object.reference.as_ref().unwrap();
+                assert!(
+                    reference.object_id.is_some(),
+                    "object should have object_id"
+                );
+                assert!(reference.version.is_some(), "object should have version");
+                assert!(reference.digest.is_some(), "object should have digest");
```
muXxer: isn't that the field mask check?
we can use the assert function here.

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
```diff
@@ -0,0 +1,436 @@
...
+        !response.has_next,
+        "has_next should be false for empty request"
+    );
+
+    // Stream should be exhausted
+    assert!(
+        stream.next().await.is_none(),
+        "Stream should be exhausted after empty response"
+    );
+}
```
muXxer, line 438: we don't have a test for objects that don't exist.

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -86,12 +87,12 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
     async fn get_objects(
         &self,
-        _request: tonic::Request<grpc_ledger_service::GetObjectsRequest>,
+        request: tonic::Request<grpc_ledger_service::GetObjectsRequest>,
     ) -> std::result::Result<tonic::Response<Self::GetObjectsStream>, tonic::Status> {
-        // not implemented - return empty stream
-        let stream = futures::stream::empty();
-        let stream: Self::GetObjectsStream = Box::pin(stream);
-        Ok(Response::new(stream))
+        // Get the lazy stream
```
muXxer:
```suggestion
```

crates/iota-grpc-types/src/proto/iota/grpc/v0/object.rs
```diff
@@ -5,3 +5,47 @@
...
+
+use crate::{
+    field::FieldMaskTree,
+    merge::Merge,
+    v0::{bcs::BcsData, types::ObjectReference},
+};
+
+impl Merge<&iota_sdk2::types::object::Object> for Object {
+    fn merge(&mut self, source: &iota_sdk2::types::object::Object, mask: &FieldMaskTree) {
+        if mask.contains("bcs") {
```
muXxer, line 18: The only problem here is, we need to wrap objects into a type with a version IMO. We can do that as a last PR to the feature branch for all endpoints, but we should add a TODO here that we don't forget it.

crates/iota-grpc-server/src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+
+            loop {
+                // Try to get the next item
+                match requests_iter.next() {
+                    Some($item_pattern) => {
+                        // Process the item using the provided block
+                        let (result_item, item_size) = $process_block;
+
+                        // Check if adding this item would exceed the limit
+                        if current_size + item_size > $max_message_size && !current_batch.is_empty() {
```
muXxer, line 57: I think we should return an error if the `current_batch` was empty, but the item still didn't fit.
Otherwise we might exceed the message size in some cases.

muXxer (changes requested):

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
bingyanglin: Thanks for pointing this out, I enhanced  it at [f1b43fc](https://github.com/iotaledger/iota/pull/9328/commits/f1b43fc3711f834d1693910884a59261550acc0b).

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
bingyanglin: Agreed, I added a more generic helper function in [a64380f](https://github.com/iotaledger/iota/pull/9328/commits/a64380f6e24628f606bcbed471910d0ccd49e653) and modified the code accordingly.

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
bingyanglin: Thanks for pointing this out. I added a more generic helper function in [a64380f](https://github.com/iotaledger/iota/pull/9328/commits/a64380f6e24628f606bcbed471910d0ccd49e653) and modified the code accordingly.

crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_objects.rs
bingyanglin: Thanks for identifying this, I added the test in [d2f32ae](https://github.com/iotaledger/iota/pull/9328/commits/d2f32ae623c12bf98ca5d856fa6881b6421f1504).

crates/iota-grpc-types/src/proto/iota/grpc/v0/object.rs
bingyanglin: Agreed, I added the TODO comment in [85797e3](https://github.com/iotaledger/iota/pull/9328/commits/85797e3f4c84f41f7c1f43b1deda2abe62343b1d).

crates/iota-grpc-server/src/utils.rs
bingyanglin: Thanks for identifying this bug! I fixed it at [85f4747](https://github.com/iotaledger/iota/pull/9328/commits/85f4747c1bf4b7cb663adca2fd36086f5b34f7ae).

crates/iota-grpc-server/src/utils.rs
```diff
@@ -44,6 +44,15 @@ macro_rules! create_batching_stream {
...
                         let (result_item, item_size) = $process_block;
 
+                        // Check if a single item exceeds the message size limit
+                        if item_size > $max_message_size {
+                            Err($crate::error::RpcError::new(
+                                tonic::Code::InvalidArgument,
+                                format!("Single item size ({} bytes) exceeds max message size ({} bytes)",
+                                    item_size, $max_message_size)
+                            ))?;
+                        }
```
muXxer: I think we can move that into the check below, by moving the `!current_batch.is_empty()` inside the if block.

```rust
// Check if adding this item would exceed the limit
if current_size + item_size > $max_message_size {
    if current_batch.is_empty() {
        // Single item exceeds the message size limit
```

muXxer (approved):

PR #9361, opened 2025-11-20
feat(iota-core, iota-types): Add scoring function

oliviasaa:
```mermaid
---
config:
  gitGraph:
    {parallelCommits: true,
    mainBranchName: 'Feature',
    rotateCommitLabel: true}
---
gitGraph
   commit id: "Initial commit"
    commit id: "add MisbehaviorReports" tag: "PR9175"
   commit id: "add ScoringMetrics" tag: "PR9176"
   commit id: "add Scorer" tag: "PR9177"
   commit id: "add ScoringMetricsStore" tag: "PR9178"
commit id: "add report validation" tag: "PR9362"
   checkout Feature
branch scoring-function
branch report-creation
checkout Feature
merge  scoring-function tag: "THIS PR"
```
# Description of change

We Introduce the scoring formulas used for the validator report mechanism, which where only placeholder before this PR.

### Release Notes

- [x] Protocol: Introduce new scoring formulas for the validator score mechanism
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/authority/scorer.rs
+++ b/crates/iota-core/src/authority/scorer.rs
@@ -12,6 +12,9 @@ use iota_types::{
     scoring_metrics::VersionedScoringMetrics,
 };
 
+const MAX_SCORE: u64 = 2_u64.pow(16); // Note: must be consistent with MAX_SCORE in validator_set.move in iota-framework.
+const SCALE_FACTOR: u64 = 2_u64.pow(16);
+
 /// Holds all information related to scoring of authorities in the committee.
 pub struct Scorer {
     // The current metrics counts collected by the authority, i.e., the local view of the node
@@ -49,8 +52,6 @@ impl Scorer {
                     committee_size,
                     protocol_config,
                 ));
-
-                let max_score = 2_u64.pow(16);
                 let (received_metrics, has_not_sent_report, current_scores, invalid_reports_count) =
                     (0..committee_size)
                         .map(|_| {
@@ -60,7 +61,7 @@ impl Scorer {
                                 // Initially, none of the authorities had sent any valid report.
                                 AtomicBool::new(true),
                                 // Current scores initialized to max score.
-                                AtomicU64::new(max_score),
+                                AtomicU64::new(MAX_SCORE),
                                 // Invalid reports count initialized to zero.
                                 AtomicU64::new(0),
                             )
@@ -68,26 +69,15 @@ impl Scorer {
                         .collect();
                 let scale_factor = 2_u64.pow(16);
                 let parameters = ParametersV1 {
-                    max_score,
-                    scale_factor,
-                    allowances: MisbehaviorsV1 {
-                        faulty_blocks_provable: 2,
-                        faulty_blocks_unprovable: 1,
-                        missing_proposals: 1000,
-                        equivocations: 0,
-                    },
-                    maximums: MisbehaviorsV1 {
-                        faulty_blocks_provable: 10,
-                        faulty_blocks_unprovable: 5,
-                        missing_proposals: 5000,
-                        equivocations: 1,
-                    },
-                    weights: MisbehaviorsV1 {
-                        faulty_blocks_provable: scale_factor * 30 / 100,
-                        faulty_blocks_unprovable: scale_factor * 10 / 100,
-                        missing_proposals: scale_factor * 35 / 100,
-                        equivocations: 1,
-                    },
+                    max_score: MAX_SCORE,
+                    scale_factor: SCALE_FACTOR,
+                    allowances: vec![1, 2, 1000, 0],
+                    maximums: vec![5, 10, 5000, 1],
+                    weights: vec![
+                        SCALE_FACTOR * 30 / 100,
+                        SCALE_FACTOR * 10 / 100,
+                        SCALE_FACTOR * 35 / 100,
+                    ],
                 };
                 // Assert that the allowance for major misbehaviors is 0,
                 // maximum is 1 and weight is 1. This is because major misbehaviors should
@@ -433,7 +423,7 @@ mod tests {
     use iota_types::messages_consensus::{MisbehaviorsV1, VersionedMisbehaviorReport};
 
     use crate::authority::authority_per_epoch_store::scorer::{
-        ParametersV1, Scorer, calculate_median_report, calculate_scores_v1,
+        MAX_SCORE, ParametersV1, SCALE_FACTOR, Scorer, calculate_median_report, calculate_scores_v1,
     };
 
     fn mock_protocol_config(consensus_choice: ConsensusChoice) -> ProtocolConfig {
@@ -681,26 +671,15 @@ mod tests {
     #[test]
     fn test_calculate_scores_v1() {
         let parameters = ParametersV1 {
-            max_score: 2_u64.pow(16),
-            scale_factor: 2_u64.pow(16),
-            allowances: MisbehaviorsV1 {
-                faulty_blocks_provable: 1,
-                faulty_blocks_unprovable: 2,
-                missing_proposals: 1000,
-                equivocations: 0,
-            },
-            maximums: MisbehaviorsV1 {
-                faulty_blocks_provable: 5,
-                faulty_blocks_unprovable: 10,
-                missing_proposals: 5000,
-                equivocations: 1,
-            },
-            weights: MisbehaviorsV1 {
-                faulty_blocks_provable: 2_u64.pow(16) * 30 / 100,
-                faulty_blocks_unprovable: 2_u64.pow(16) * 10 / 100,
-                missing_proposals: 2_u64.pow(16) * 35 / 100,
-                equivocations: 1,
-            },
+            max_score: MAX_SCORE,
+            scale_factor: SCALE_FACTOR,
+            allowances: vec![1, 2, 1000, 0],
+            maximums: vec![5, 10, 5000, 1],
+            weights: vec![
+                SCALE_FACTOR * 30 / 100,
+                SCALE_FACTOR * 10 / 100,
+                SCALE_FACTOR * 35 / 100,
+            ],
         };
 
         let median_reports = MisbehaviorsV1 {
--- a/crates/iota-core/src/authority/scorer.rs
+++ b/crates/iota-core/src/authority/scorer.rs
@@ -91,7 +91,7 @@ impl Scorer {
                 };
                 // Assert that the allowance for major misbehaviors is 0,
                 // maximum is 1 and weight is 1. This is because major misbehaviors should
-                // reduce the score to 0 is there are any occurences.
+                // reduce the score to 0 is there are any occurrences.
                 // Only equivocation is considered a major misbehavior in this version.
                 assert!(
                     parameters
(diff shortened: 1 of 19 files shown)


crates/iota-core/src/authority/scorer.rs
```diff
@@ -11,17 +11,30 @@ use iota_types::{
...
-
-#[expect(dead_code)]
+/// Holds all information related to scoring of authorities in the committee.
 pub struct Scorer {
+    // The current metrics counts collected by the authority, i.e., the local view of the node
+    // about the behaviour of the rest of the committee, according to the blocks received.
     pub(crate) current_local_metrics_count: Arc<VersionedScoringMetrics>,
+    // The metrics counts received from other authorities, i.e., the information contained in the
+    // MisbehaviourReports received by the authority. If an authority has not sent a report, its
+    // received_metrics will be all zeroed.
```
cyberphysic4l:
```suggestion
    // entry in this vector will be all zeroed.
```

crates/iota-core/src/authority/scorer.rs
```diff
@@ -30,45 +43,67 @@ impl Scorer {
...
-                                AtomicU64::new(0),
+                                // Current scores initialized to max score.
+                                AtomicU64::new(max_score),
+                                // Invalid reports count initialized to zero.
                                 AtomicU64::new(0),
                             )
                         })
                         .collect();
+                let parameters = ParametersV1 {
+                    max_score,
```
cyberphysic4l: Can we just use `max_score` here in place of `2_u64.pow(16)` in all these params? If we changed `max_score` would these all change too? Or is it a separate value?

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
     }
+}
 
+// Methods for ScorerVersion::V1
+impl Scorer {
     fn update_scores_v1(&self) {
-        // Placeholder
+        // Vector with the highest received reports from each authority and their voting
+        // power. Authorities that did not send reports are filtered out.
+        let highest_received_reports_from_authority = self
```
cyberphysic4l, line 157: What does `highest` refer to? Should it be `latest` instead, indicating that we use the most recent received report from each authority? `highest` could imply we take the max value or something.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+            .iter()
+            .zip(self.voting_power.iter())
+            .zip(self.has_not_sent_report.iter())
+            .filter(|((_, _), is_missing)| !is_missing.load(Ordering::Relaxed))
+            .map(|((metrics, voting_power), _)| (metrics.to_report(), *voting_power))
+            .collect::<Vec<(VersionedMisbehaviorReport, VotingPower)>>();
+        // Ensure that we have at least one report to calculate the scores, otherwise we
+        // do nothing.
+        if highest_received_reports_from_authority.is_empty() {
+        } else {
```
cyberphysic4l, line 168:
```suggestion
        if !highest_received_reports_from_authority.is_empty() {
```

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+// (allowed misbehaviors without any punishment) and a maximum (number of
+// misbehaviors that lead to zero score). Based on those values, we calculate a
+// score per metric, and then combine them into a final score. Each individual
+// score is also  an integer between 0 and max_score, and the weights used for
+// the combination are such that sum(weights) + baseline_score = scale_factor.
+// Thus, we need max_score*scale_factor < 2^64 to avoid overflows.
+fn calculate_scores_v1(median_reports: Vec<MedianMetricVec>, parameters: ParametersV1) -> Vec<u64> {
+    // Ensure we have 4 median reports: faulty blocks provable, faulty blocks
+    // unprovable, missing blocks, equivocations. This is naturally guaranteed when
+    // these data come from MisbehaviorReports, since they would been considered
```
cyberphysic4l:
```suggestion
    // these data come from MisbehaviorReports, since they would have been considered
```

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+/// This method returns a vector of MedianMetricVec, one per metric in
+/// VersionedMisbehaviorReport
+fn calculate_median_report(
+    reports_and_voting_power: &[(VersionedMisbehaviorReport, VotingPower)],
+) -> Vec<MedianMetricVec> {
+    // Calls to this method should ensure that we have at least one report to
+    // process.
+    assert!(!reports_and_voting_power.is_empty());
+
+    let number_of_metrics = reports_and_voting_power[0].0.iterate_over_metrics().len();
```
cyberphysic4l, line 212: define a constant `number_of_metrics_v1` because you use the magic number 4 below to check this as well.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+// score per metric, and then combine them into a final score. Each individual
+// score is also  an integer between 0 and max_score, and the weights used for
+// the combination are such that sum(weights) + baseline_score = scale_factor.
+// Thus, we need max_score*scale_factor < 2^64 to avoid overflows.
+fn calculate_scores_v1(median_reports: Vec<MedianMetricVec>, parameters: ParametersV1) -> Vec<u64> {
+    // Ensure we have 4 median reports: faulty blocks provable, faulty blocks
+    // unprovable, missing blocks, equivocations. This is naturally guaranteed when
+    // these data come from MisbehaviorReports, since they would been considered
+    // invalid otherwise.
+    assert_eq!(median_reports.len(), 4);
```
cyberphysic4l: A suggested above, define a constant for the number of metrics in v1 instead.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -30,45 +43,67 @@ impl Scorer {
...
                         .collect();
+                let parameters = ParametersV1 {
+                    max_score,
+                    scale_factor: 2_u64.pow(16),
+                    allowances: vec![1, 2, 1000, 0],
+                    maximums: vec![5, 10, 5000, 1],
+                    weights: vec![
+                        2_u64.pow(16) * 30 / 100,
+                        2_u64.pow(16) * 10 / 100,
+                        2_u64.pow(16) * 35 / 100,
```
cyberphysic4l: Will these remain hardcoded here or be defined elsewhere in another PR? If they remain here it would be worth commenting here which values correspond to which metrics and how they should be interpreted.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -30,45 +43,67 @@ impl Scorer {
...
                             )
                         })
                         .collect();
+                let parameters = ParametersV1 {
+                    max_score,
+                    scale_factor: 2_u64.pow(16),
+                    allowances: vec![1, 2, 1000, 0],
+                    maximums: vec![5, 10, 5000, 1],
+                    weights: vec![
+                        2_u64.pow(16) * 30 / 100,
```
cyberphysic4l: These are `scale_factor`, right? Should define it above and use it here.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+
+    metrics_scores_to_final_scores(
+        &non_equivocation_metrics_scores,
+        parameters.weights,
+        equivocation_score,
+        baseline_score,
+        parameters.scale_factor,
+        parameters.max_score,
+    )
+}
```
cyberphysic4l: I actually think it might be better not to group all the allowances, maximums and weights into a vector, it just end up being more confusing. It will take more lines of code, but I think it would be better to just explicitly list them all, `missing_allowance`, etc...

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +112,301 @@ impl Scorer {
...
+            let to_add = metric_score
+                .iter()
+                .map(|x| x * weight)
+                .collect::<Vec<u64>>();
+            final_scores
+                .iter_mut()
+                .zip(to_add.iter())
+                .for_each(|(a, b)| *a += *b);
+        });
+    // Combine equivocation scores into final scores, and scale down. The final
```
cyberphysic4l: Mention that equivocation score is between zero and max_score too, which is why we are dividing by max_score.

cyberphysic4l (changes requested):
My main comment is just about the scoring function and the iterating over the metrics to calculate the separate scores. My feeling is that it might be better to simply enumerate all the metrics and the're parameters in the scoring function by name for the sake of clarity, even if this means more lines of code. I feel we don't get much benefit from having them in a vector. We can discuss this further.

cyberphysic4l (approved):

crates/iota-core/src/authority/scorer.rs
```diff
@@ -143,24 +487,232 @@ mod tests {
...
+                }),
+                10_u64,
+            ),
+            (
+                VersionedMisbehaviorReport::V1(MisbehaviorsV1 {
+                    faulty_blocks_provable: vec![7, 8, 9],
+                    faulty_blocks_unprovable: vec![10, 11, 12],
+                    missing_proposals: vec![4, 5, 6],
+                    equivocations: vec![1, 2, 0],
+                }),
```
bingyanglin, line 647: This setting exist multiple times, maybe it's cleaner to define something like `default_test_parameters() -> ParametersV1` so as to reduce the lines of code, and easier to maintain?

crates/iota-core/src/authority/scorer.rs
```diff
@@ -77,37 +143,315 @@ impl Scorer {
...
+        .collect()
+}
+
+fn metric_to_score(value: u64, allowance: u64, max: u64, max_score: u64) -> u64 {
+    if value <= allowance {
+        max_score
+    } else if value >= max {
+        0
+    } else {
+        // Add overflow checks
```
bingyanglin:
```suggestion
        // TODO: add overflow checks
```
Also it might be good to define a corresponding issue for the TODO if you haven't PRed and/or planned.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -30,45 +44,97 @@ impl Scorer {
...
         match protocol_config.scorer_version_as_option() {
             None | Some(1) => {
+                // Local metrics count are always initialized as zero.
                 let current_local_metrics_count = Arc::new(VersionedScoringMetrics::new(
                     committee_size,
                     protocol_config,
                 ));
 
-                let (received_metrics, metrics_missing_from, current_scores, invalid_reports_count) =
+                let max_score = 2_u64.pow(16);
```
bingyanglin: Would it be better to define a const like 
`const MAX_SCORE: u64 = 2_u64.pow(16);` in the front of the code? Same for the scale_factor.

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,6 +123,20 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
                 .set(checkpoint_seq as i64);
         }
 
+        // We also send misbehavior reports to consensus at this point. Misbehavior
+        // reports containing proofs of misbehaviour can be send whenever the
+        // misbehavior is detected, but we choose to send the ones that include only
+        // unprovable counts at this point, due to periodicity reasons and to ensure a
+        // (approximate) synchronization with the score updates.
+        let misbehavior_report = epoch_store.scorer.current_local_metrics_count.to_report();
```
piotrm50: So basically all validators send the report after each checkpoint they create? Do I understand correctly?

crates/iota-core/src/authority/scorer.rs
cyberphysic4l: Done, issue created https://github.com/iotaledger/iota-private/issues/242

PR #9368, opened 2025-11-20
feat(verifier): make the authenticate function force the entry modifier

### LINKED ISSUE #9364: [AA] Make the authenticate function an entry function

miker83z:
# Description of change

- Changed the entry functions verifier check to support `AuthContext` as the second to last parameter. This might be a reason for an execution CUT.
- Changed the check of authenticate functions to only support (public and private) entry functions. Now it only need to check that params are not receiving and are either primitive types or immutable references.

Fixes #9364 


Dkwcs:
Should this PR also include `iota-adapter-transcational-tests` update?

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/option.move
```diff
@@ -24,185 +28,196 @@ public struct NonObject has copy, drop, store {}
...
 // Option
 
-// PASS
-public fun primitive_immutable_reference(
-    _account: &Account,
-    _arg: &Option<u8>,
-    _auth_ctx: &AuthContext,
-    _ctx: &TxContext,
-) {}
+// FAIL Invalid 'entry' parameter type
```
Dkwcs, line 31: Why does it fail? Because of Option arg?

iota-execution/latest/iota-verifier/src/authenticator_verifier.rs
```diff
@@ -143,190 +136,84 @@ pub fn verify_authenticate_func_v1(
...
-        }
-        DatatypeInstantiation(datatype_instance) => {
-            let (idx, type_args) = &**datatype_instance;
-            let resolved_struct = resolve_struct(module, *idx);
-            if resolved_struct == RESOLVED_STD_OPTION && type_args.len() == 1 {
-                verify_pure_input_type(module, function_type_args, &type_args[0])
-            } else {
-                Err(format!(
-                    "Invalid pure type. A datatype instantiation must be an option of pure types, offending argument: {param:?}"
+                    "Invalid parameter type for authenticate function: {}. Valid types are immutable references to anything but receiving objects, or primitive types.",
```
Dkwcs: Extremely minor notice - do we need comma before `but`.
Alternately I would rephrase it to `Invalid parameter type for authenticate function: {}. Valid types are immutable references to anything, except receiving objects, or primitive types`

The same notice for the error message above

iota-execution/latest/iota-verifier/src/entry_points_verifier.rs
```diff
@@ -180,7 +181,17 @@ fn verify_entry_function_impl(
 
     let all_non_ctx_params = match params.0.last() {
```
Dkwcs, line 182: I would suggest simplified version of these checks.
1. Removing inner match
2. Less manual indexing
3. A bit more structured 

```suggestion
    let all_non_ctx_params = match params.0.last() {
        Some(last_param) if TxContext::kind(view, last_param) != TxContextKind::None => {
            let mut non_ctx_params = &params.0[..params.0.len() - 1];

            if let Some((second_last, rest)) = non_ctx_params.split_last() {
                if AuthContext::kind(view, second_last) != AuthContextKind::None {
                    non_ctx_params = rest;
                }
            }

            non_ctx_params
        }
        _ => &params.0[..],
    };
```

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
```diff
@@ -3,9 +3,13 @@
 
 module authenticate::vector;
 
-use authenticate::account::Account;
 use iota::auth_context::AuthContext;
 
+// Test account struct
+public struct Account has key {
+    id: UID,
+}
```
valeriyr, line 11: What is the reason to duplicate this type everywhere?
It is not really important for these tests; can be declared only once.

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/object.move
```diff
@@ -3,33 +3,45 @@
...
 // FAIL
-public fun by_mutable_ref(
+//#[authenticator]
+public entry fun by_mutable_ref(
     _account: &Account,
     _object: &mut Object,
-    _auth_ctx: &AuthContext,
+    _actx: &AuthContext,
     _ctx: &TxContext,
 ) {}
```
valeriyr, line 47: I am not sure that I see the reason for having these tests anymore.
If we just comment some use cases, comment the `#[authenticator]` attribute. We check only positive ones that don't fail.
Are we going to create a test for each failed case later?

external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
```diff
@@ -608,7 +608,16 @@ fn entry_signature(
     } = signature;
     let all_non_ctx_parameters = match parameters.last() {
         Some((_, _, last_param_ty)) if tx_context_kind(last_param_ty) != TxContextKind::None => {
-            &parameters[0..parameters.len() - 1]
+            let params_without_tx_ctx = &parameters[0..parameters.len() - 1];
```
valeriyr, line 611: Do we need to check `AuthContext` on the compiler level?
Can it be considered as a simple object parameter by the compiler?

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/option.move
miker83z: Because it's a reference to an option (not allowed for entry functions)

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/object.move
miker83z: All these failing cases will be checked with a dedicated test in `iota-verifier-transactional-tests`. Right now, these cannot be implemented because we need a change in the `move-transactional-tests` (i.e., carry over the annotation informations, which, currently, are dropped).

external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
miker83z: That's the main change. We need the `entry` function definition to change and support AuthContext. Because this one is not an Object (no key ability) and not even a primitive value.

crates/iota-e2e-tests/tests/abstract_account/authenticate/sources/vector.move
miker83z: For this https://github.com/iotaledger/iota/pull/9368#discussion_r2559741845

iota-execution/latest/iota-verifier/src/entry_points_verifier.rs
miker83z: [ddecdf4](https://github.com/iotaledger/iota/pull/9368/commits/ddecdf4a2aed9dab398084eb81e49edee5f833ed)

iota-execution/latest/iota-verifier/src/authenticator_verifier.rs
miker83z: [ddecdf4](https://github.com/iotaledger/iota/pull/9368/commits/ddecdf4a2aed9dab398084eb81e49edee5f833ed)

Dkwcs (approved):
LGTM

PR #9372, opened 2025-11-21
feat(framework-snapshot): added v16 snapshots

alexsporn:
# Description of change

Added procotol v16 snapshots.


valeriyr (approved):

alexsporn:
Merging with failing CI because the issue was added before to develop and should be fixed in a separate PR.

PR #9381, opened 2025-11-21
feat(starfish): Limit synchronizer parallelism and add cache for verified blocks

### LINKED ISSUE #9357: Debug and fix high CPU usage preventing the node from catching up when running Synchronizer
We notice high CPU usage when Synchronizer is fetching a lot of blocks to catch up after a hiccup. Sometimes this spirals out of control, because CPU is choking and the whole node slows down, preventing the node from catching up with the quorum, resulting in node not contributing to consensus at all for long periods of time.

piotrm50:
# Description of change

This is a PR to test a potential fix for an issue with the header synchronizer. It limits parallelism of fetching blocks in the HeaderSynchronizer and implements an additional filter to avoid processing redundantly fetched blocks which should reduce CPU usage.

Resolves #9357 

### DIFF
--- a/crates/starfish/core/src/header_synchronizer.rs
+++ b/crates/starfish/core/src/header_synchronizer.rs
@@ -156,12 +156,12 @@ impl InflightBlockHeadersMap {
         for block_ref in missing_block_refs {
             let authorities = inner.entry(block_ref).or_default();
 
-            // Check if this peer is already fetching this block
+            // Check if this peer is already fetching this header
             if authorities.contains(&peer) {
                 continue;
             }
 
-            // Count total authorities currently fetching this block
+            // Count total authorities currently fetching this header
             let total_count = authorities.len();
 
             // Determine the limit based on the sync method
--- a/crates/starfish/core/src/header_synchronizer.rs
+++ b/crates/starfish/core/src/header_synchronizer.rs
@@ -101,6 +101,15 @@ impl std::fmt::Display for SyncMethod {
     }
 }
 
+impl SyncMethod {
+    fn max_authorities_to_fetch_per_block_header(&self) -> usize {
+        match self {
+            SyncMethod::Live => MAX_AUTHORITIES_TO_LIVE_FETCH_PER_BLOCK_HEADER,
+            SyncMethod::Periodic => MAX_AUTHORITIES_TO_FETCH_PER_BLOCK_HEADER,
+        }
+    }
+}
+
 struct BlocksGuard {
     map: Arc<InflightBlockHeadersMap>,
     block_refs: BTreeSet<BlockRef>,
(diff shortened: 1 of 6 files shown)


crates/starfish/core/src/header_synchronizer.rs
```diff
@@ -109,22 +138,40 @@ impl InflightBlockHeadersMap {
...
         for block_ref in missing_block_refs {
-            // check that the number of authorities that are already instructed to fetch the
-            // block is not higher than the allowed and the `peer_index` has not
-            // already been instructed to do that.
             let authorities = inner.entry(block_ref).or_default();
-            if authorities.len() < MAX_AUTHORITIES_TO_FETCH_PER_BLOCK_HEADER
-                && authorities.get(&peer).is_none()
-            {
+
+            // Check if this peer is already fetching this block
```
polinikita:
```suggestion
            // Check if this peer is already fetching this header
```

crates/starfish/core/src/header_synchronizer.rs
```diff
@@ -109,22 +138,40 @@ impl InflightBlockHeadersMap {
...
+            if authorities.contains(&peer) {
+                continue;
+            }
+
+            // Count total authorities currently fetching this block
```
polinikita:
```suggestion
            // Count total authorities currently fetching this header
```

polinikita (approved):
For starfish, this seems to be less important improvement as the most heavy operation is the signature verification which takes 50-100microseconds.

crates/starfish/core/src/header_synchronizer.rs
```diff
@@ -73,10 +81,31 @@ const MAX_PERIODIC_SYNC_PEERS: usize = 4;
 /// based on their knowledge of the DAG.
 const MAX_PERIODIC_SYNC_RANDOM_PEERS: usize = 2;
 
+/// The maximum number of verified block header references to cache for
+/// deduplication.
+const VERIFIED_HEADERS_CACHE_CAP: usize = 200_000;
+
+/// Represents the different methods used for synchronization
```
filipdulic, line 88: Nit:
you can
```rust
impl SyncMethod {
  fn max_authorities_to_fetch_per_block_header(&self) -> usize {
    match self {
      SyncMethod::Live => 2,
      SyncMethod::Periodic => 1,
    }
  }
}
```
using this enum method instead of 
```rust
let max_limit = match method {
                SyncMethod::Live => MAX_AUTHORITIES_TO_LIVE_FETCH_PER_BLOCK_HEADER,
                SyncMethod::Periodic => MAX_AUTHORITIES_TO_FETCH_PER_BLOCK_HEADER,
            };
```

filipdulic (approved):

Issue #9338, opened 2025-11-19
Implement CommitV2 and PendingSubDag structures with feature gate


piotrm50:
Implement new commit structures to support fast commit synchronization:

### Changes Required

1. **New Structures:**
   ```rust
   pub struct TransactionRef {
       pub round: Round,
       pub authority: AuthorityIndex,
       pub transactions_commitment: TransactionsCommitment
   }
   
   pub(crate) struct CommitV2 {
       index: CommitIndex,
       previous_digest: CommitDigest,
       timestamp_ms: BlockTimestampMs,
       leader: BlockRef,
       blocks: Vec<BlockRef>,
       committed_transactions: Vec<TransactionRef>,
       reputation_scores: Vec<(AuthorityIndex, u64)>,
   }
   
   pub struct PendingSubDag {
       pub base: SubDagBase,
       pub committed_transaction_refs: Vec<TransactionRef>,
   }
   ```

2. **Linearizer Updates:**
   - Modify linearizer to read `dag_state` when getting `committed_transactions`
   - Get corresponding `TransactionsCommitment` for each committed transaction
   - Populate `CommitV2` with `reputation_scores` (obtained after DAG processing)

3. **DagState Updates:**
   - Store transactions and shards with `TransactionRef` identifier

4. **Feature Gate:**
   - Add feature gate to allow gradual rollout
   - Ensure backward compatibility with existing commit structures

5. **CommitSyncer Compatibility:**
   - Update current `CommitSyncer` logic to work with new structs
   - Maintain existing functionality while supporting new format

### Acceptance Criteria
- [x] New structures implemented and compile without errors
- [x] Feature gate properly controls new vs old behavior
- [ ] Linearizer correctly populates `TransactionRef` and `TransactionsCommitment`
- [ ] DagState stores transactions with new identifier format
- [ ] Existing CommitSyncer continues to work with new structures
- [ ] Unit tests for new structures and conversions
- [ ] Integration tests verify backward compatibility

PR #9383, opened 2025-11-21
feat(consensus): introduce transaction ref

### LINKED ISSUE #9359: Introducing TransactionRef
As a first step in implementing the fast commit syncer, we introduce a TransactionRef

VorobyevIlya:
# Description of change

This PR introduces TransactionRef that differs from BlockRef and additionally contains TransactionCommitment. We are going to use TransactionRef to work with transactions, in particular, to fetch transactions. It will allow us to fetch and verify transactions without fetching corresponding headers, which is useful in commit_syncer.

Usage of TransactionRef is controlled by the protocol flag consensus_transaction_ref. If it is set to true, we use TransactionRef; otherwise, we use BlockRef to work with transactions. To ensure smooth migration, we introduce GenericTransactionRef, which has TransactionRef and BlockRef as its variants.

We use GenericTransactionRef in our internal functions. We create the second versions of Commit, SerializedTransaction, and ShardsWithProof, and use the second version if the flag is set to true. We create new fields in RocksDB and MemStore that use TransactionRef as keys, and write to these new fields if the new feature is enabled. We store recent transactions and shards in dag_state with GenericTransactionRef as key.

We modify the verification of the fetched transactions; in the new version, it doesn't require a header.

For now, TransactionRef contains a block header digest for simplicity; it will be removed in the next PRs.

Fixes #9359 and #9338 

Lots of tests were modified to work with a new feature enabled and disabled. Split cluster was also tested to ensure that the new feature doesn't break anything. Also the code was tested in the local network.


crates/starfish/core/src/dag_state.rs
```diff
@@ -518,30 +534,33 @@ impl DagState {
...
-                    transactions[index] = Some(transaction);
+        for (index, gen_transactions_ref) in gen_transactions_refs.iter().enumerate() {
+            if gen_transactions_ref.round() == GENESIS_ROUND {
+                // Extract the BlockRef from GenericTransactionRef
+                let genesis_key = match gen_transactions_ref {
+                    GenericTransactionRef::BlockRef(br) => *br,
+                    GenericTransactionRef::TransactionRef(tr) => {
+                        BlockRef::new(tr.round, tr.author, tr.block_digest)
+                    }
+                };
```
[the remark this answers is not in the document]
VorobyevIlya: where is it duplicated? @copilot

crates/starfish/core/src/block_header.rs
```diff
@@ -366,6 +376,135 @@ impl Hash for BlockRef {
     }
 }
 
+#[derive(Clone, Copy, Serialize, Deserialize, Default, PartialEq, Eq, PartialOrd, Ord)]
```
polinikita: i think that one can create a new rust-file to cover all the description and logic around transaction_ref? What do you think?

crates/starfish/core/src/commit.rs
```diff
@@ -148,8 +183,70 @@ impl CommitAPI for CommitV1 {
...
+    /// Digest of the previous commit.
+    /// Set to CommitDigest::MIN for the first commit after genesis.
+    previous_digest: CommitDigest,
+    /// Timestamp of the commit, max of the timestamp of the leader block and
+    /// previous Commit timestamp.
+    timestamp_ms: BlockTimestampMs,
+    /// A reference to the commit leader.
+    leader: BlockRef,
+    /// Refs to committed blocks, in the commit order.
+    blocks: Vec<BlockRef>,
```
polinikita: can we rename this to headers or block_headers? Also change the comment?

VorobyevIlya: done

crates/starfish/core/src/block_header.rs
VorobyevIlya: I created a new file and moved this logic there

thibault-martinez (approved):

muXxer (approved):

muXxer (commented):
Approving `crates/iota-protocol-config/src/lib.rs`

crates/starfish/core/src/block_header.rs
```diff
@@ -431,20 +433,128 @@ impl AsRef<[u8]> for BlockHeaderDigest {
 }
 
 #[derive(Clone, Copy, Serialize, Deserialize, Default, PartialEq, Eq, PartialOrd, Ord)]
-pub struct TransactionsCommitment([u8; starfish_config::DIGEST_LENGTH]);
+pub struct TransactionsCommitment(pub(crate) [u8; starfish_config::DIGEST_LENGTH]);
```
polinikita, line 436: do we need pub(crate) inside?

crates/starfish/core/src/commit.rs
```diff
@@ -148,8 +190,74 @@ impl CommitAPI for CommitV1 {
...
+
+    fn timestamp_ms(&self) -> BlockTimestampMs {
+        self.timestamp_ms
+    }
+
+    fn leader(&self) -> BlockRef {
+        self.leader
+    }
+
+    fn blocks(&self) -> &[BlockRef] {
```
polinikita: Would it be hard to change the name of this function `blocks()` to `headers()` or `block_headers()`?

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -108,9 +109,14 @@ pub(crate) struct CordialKnowledge {
 #[derive(Debug)]
 pub enum CordialKnowledgeMessage {
     /// A new verified block header to integrate into cordial knowledge.
-    NewHeader(VerifiedBlockHeader),
+    /// Includes transaction commitments of all blocks acknowledged by this
+    /// header.
+    NewHeader {
+        header: VerifiedBlockHeader,
+        ack_transactions_commitments: Vec<Option<TransactionsCommitment>>,
```
polinikita, line 116: It is not clear why it is an Option here.

crates/starfish/core/src/storage/rocksdb_store.rs
```diff
@@ -19,40 +19,60 @@ use super::{CommitInfo, Store, WriteBatch};
...
     /// Stores SignedBlockHeader by refs.
     block_headers: DBMap<(Round, AuthorityIndex, BlockHeaderDigest), Bytes>,
-    /// Stores Transactions by refs
+    /// Stores Transactions by block refs
     transactions: DBMap<(Round, AuthorityIndex, BlockHeaderDigest), Bytes>,
+    /// Stores Transactions by transaction refs
+    transactions_by_tr_refs: DBMap<(Round, AuthorityIndex, TransactionsCommitment), Bytes>,
     /// A secondary index that orders refs first by authors.
     digests_by_authorities: DBMap<(AuthorityIndex, Round, BlockHeaderDigest), ()>,
+    /// A secondary index that orders refs first by authors.
```
polinikita: the comment is the same as in above field

crates/starfish/core/src/storage/rocksdb_store.rs
```diff
@@ -19,40 +19,60 @@ use super::{CommitInfo, Store, WriteBatch};
...
     /// Stores info related to Commit that helps recovery.
     commit_info: DBMap<(CommitIndex, CommitDigest), CommitInfo>,
+    /// Context to access protocol configuration
+    #[cfg_attr(not(test), allow(dead_code))]
+    context: Arc<Context>,
 }
 
 impl RocksDBStore {
     const TRANSACTIONS_CF: &'static str = "transactions";
+    const TRANSACTIONS_BY_TR_REF_CF: &'static str = "transactions_by_tr_refs";
```
polinikita: better to have a final name for this struct (BY_TR_REF is not well readable) as it will not be easy to change it without protocol parameter

polinikita (approved):
Approved. Suggested a few renames. Also, it is advisable to add a TODO with an issue what to change and clean in the codebase once we enable the flag on mainnet (many functions like `check_ref_consistency()` should be cleaned at that point)

crates/starfish/core/src/block_header.rs
VorobyevIlya: Yes. Or I can remove it and instead create a function that returns the digest

crates/starfish/core/src/commit.rs
VorobyevIlya: No, it is not hard. Renamed

crates/starfish/core/src/storage/rocksdb_store.rs
VorobyevIlya: comment was rewritten

crates/starfish/core/src/storage/rocksdb_store.rs
VorobyevIlya: What do you think will be a good name?

VorobyevIlya:
> Approved. Suggested a few renames. Also, it is advisable to add a TODO with an issue what to change and clean in the codebase once we enable the flag on mainnet (many functions like `check_ref_consistency()` should be cleaned at that point)

I think we will remove GenericTransactionRef at that point, right? Then we will need to remove all such functions.

crates/starfish/core/src/storage/mem_store.rs
```diff
@@ -14,47 +15,62 @@ use starfish_config::AuthorityIndex;
...
     commit_info: BTreeMap<(CommitIndex, CommitDigest), CommitInfo>,
 }
 
 impl MemStore {
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new(context: Arc<Context>) -> Self {
         MemStore {
             inner: RwLock::new(Inner {
                 transactions: BTreeMap::new(),
+                transactions_by_tr_refs: BTreeMap::new(),
```
piotrm50: I'd abbreviate it to `tx_ref` instead of `tr_ref` here and in other places.

crates/starfish/core/src/transaction_ref.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub author: AuthorityIndex,
+    pub transactions_commitment: TransactionsCommitment,
+    pub block_digest: BlockHeaderDigest,
+}
+
+impl fmt::Display for TransactionRef {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(
+            f,
+            "Tr{}({},{})",
```
piotrm50: I think `Tx` would be more understandable.

crates/starfish/core/src/storage/rocksdb_store.rs
piotrm50: I think we should not use `tr` as a short for `transaction` but rather `tx`, thus I'd suggest `tx_ref` instead.

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -279,19 +280,19 @@ impl InflightTransactionsMap {
     /// has been called, it will panic.
     fn unlock_transactions(
         self: &Arc<Self>,
-        block_refs: &BTreeSet<BlockRef>,
+        block_refs: &BTreeSet<GenericTransactionRef>,
```
piotrm50: block_refs doesn't fit anymore. Also, gen_tr_ref is a bit weird, I'd call it tx_refs and tx_ref correspondingly.

crates/starfish/core/src/linearizer.rs
```diff
@@ -145,7 +177,8 @@ impl Linearizer {
                 .iter()
                 .map(|block| block.reference())
                 .collect::<Vec<BlockRef>>(),
-            committed_transactions,
+            generic_committed_transactions,
```
piotrm50: it makes sense that the type is called Generic, but I think it adds unnecessary confusion when we call variable like that.

crates/starfish/core/src/dag_state.rs
```diff
@@ -518,30 +547,28 @@ impl DagState {
     /// found.
     pub(crate) fn get_verified_transactions(
         &self,
-        block_refs: &[BlockRef],
+        gen_transactions_refs: &[GenericTransactionRef],
```
piotrm50: I think here it's also not necessary to prefix the varname with `gen_`

piotrm50 (commented):
Suggested some variable renames. I agree that we should eventually remove GenericTransactionRef.

crates/starfish/core/src/storage/mem_store.rs
VorobyevIlya: done

crates/starfish/core/src/transaction_ref.rs
VorobyevIlya: done

crates/starfish/core/src/transactions_synchronizer.rs
VorobyevIlya: I renamed it to TRANSACTIONS_BY_TX_REF_CF
and did renaming in unlock_transactions

crates/starfish/core/src/dag_state.rs
VorobyevIlya: I removed the prefix gen wherever there is only one type of refs, and generic refs can't be confused with ordinary transaction refs

PR #9401, opened 2025-11-24
refactor(iota-transactional-test-runner): Handle the automatic generation of gas payment for the AbstractTransaction command

### LINKED ISSUE #9276: [AA][tx-test-runner] Handle the automatic generation of gas payment for the `AbstractTransaction` command
This means making it so that the flag `--gas-payment` is not mandatory anymore to run the command.

Dkwcs:
# Description of change

Refactoring of `iota-transactional-test-runner`: lightweight gas funding for abstract accounts

A lightweight implementation of gas funding for abstract accounts has been introduced during the abstract subcommand execution.
The default test account now funds the `abstract account` within the inner transaction, eliminating the need to manually split and transfer coins in the `.move` test files.
This approach avoids complex modifications to the `test-runner` while enabling seamless inner funding for abstract accounts.

`.move` test files content have been updated and cleared. 

Fixes #9276 .

Run `iota-adapter-transactional-tests` tests
With `cargo simtest`


crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1250,6 +1251,60 @@ fn merge_output(left: Option<String>, right: Option<String>) -> Option<String> {
 }
 
 impl IotaTestAdapter {
+    async fn fund_address_for_testing(
```
miker83z: NIT: could you move the function down next to where it is invoked?

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1250,6 +1251,60 @@ fn merge_output(left: Option<String>, right: Option<String>) -> Option<String> {
...
+
+        let tx = to_sender_signed_transaction(
+            data,
+            sender
+                .key_pair
+                .as_ref()
+                .expect("Default account must have keypair"),
+        );
+
+        let (effects, _) = self.executor.execute_txn(tx).await?;
```
miker83z: In my opinion, it would be better to reuse the methods that are used also in other commands:
`self.sign_txn(Some(sender), data);` and `self.execute_txn(transaction).await?`. These should fit here if you change the implementation slightly.

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1337,10 +1392,13 @@ impl IotaTestAdapter {
         };
         let aa_sender_addr = aa_arg.id().into();
 
+        let gas_ref = self
+            .fund_address_for_testing(aa_sender_addr, 100_000_000_000_000)
```
miker83z:
```suggestion
            .fund_address_for_testing(aa_sender_addr, GAS_FOR_TESTING)
```
There is already a constant defined for this scope. Furthermore, in any case, constants need to be declared in the top section of the file.

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: I avoided using adapter's methods on purpose, because of this limitations 
```
      if let Some(prev) = self.digest_enumeration.insert(task, *digest) {
            panic!(
                "Task {task} executed two transactions (expected at most one): {prev}, {digest}"
            );
        }
```

It doesn't allow to execute 2 transactions and it's okay, because our transaction is kinda auxiliary

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: GAS_FOR_TESTING is not enough we will encounter - `Insufficient coin balance for operation`
I'll create another const for AA gas case.

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Fixed cba43a003caf5a796bbccaeff46d3215d3dc84f5

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Fixed cba43a003caf5a796bbccaeff46d3215d3dc84f5

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1348,10 +1354,13 @@ impl IotaTestAdapter {
         };
         let aa_sender_addr = aa_arg.id().into();
 
+        let gas_ref = self
+            .fund_address_for_testing(aa_sender_addr, GAS_FOR_ABSTRACT_ACCOUNT)
+            .await?;
```
valeriyr, line 1355: Does it mean that we fund the account address each time we call the `abstract` command?

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -2150,6 +2164,48 @@ impl IotaTestAdapter {
...
+            let gas_price: u64 = self.gas_price;
+            let new_coin_arg =
+                builder.command(Command::SplitCoins(Argument::GasCoin, vec![amount_arg]));
+            builder.command(Command::TransferObjects(vec![new_coin_arg], recipient_arg));
+
+            let pt = builder.finish();
+
+            TransactionData::new_programmable(sender, gas, pt, gas_budget, gas_price)
+        };
+        let transaction = self.sign_txn(None, data, true);
```
valeriyr: Can we remove the `is_default_acc` flag because if the first parameter is None the `default_account` will be chosen anyway?

Dkwcs: Fixed 8d4e370cdb0d07295b20020e5682fa91deefcefb

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Yes

valeriyr (approved):
I would say that fund the account address each time when the `abstract` command is called is not the best option, looks like a workaround, since the command is created to execute a transaction. But if you think that it is the best choice at the moment I am approving.

Dkwcs:
> I would say that fund the account address each time when the `abstract` command is called is not the best option, looks like a workaround, since the command is created to execute a transaction. But if you think that it is the best choice at the moment I am approving.

It's a workaround, so.  Account have been created completely from Move code.
Also we've removed `init-abstract` command in case to have better flexibility.

miker83z (approved):

PR #9423, opened 2025-11-24
feat(aws): add `iota-build-cache-server` crate

muXxer:
# Description of change

By using the `aws-orchestrator`, we spawn a lot of different instances which first need to compile the different binaries used in the following tests. This wastes around 20 minutes per instance. For bigger tests with a lot of instances it makes sense to pre-compile the used binaries, before actually starting all instances.

This PR introduces an `iota-build-cache-server`, which can be run on a dedicated server, outside of the `aws-orchestrator` setup. The server can pre-build the binaries using the additional bash scripts. After the binaries are ready, the `aws-orchestrator` can be started and run tests right away.

To use the `build-cache-server` in `aws-orchestrator`, this section needs to be added to the orchestrator config:
```json
    "build_cache": {
        "enabled": true,
        "servers": {
            "x86_server": {
                "targets": ["x86-64", "x86-64-v2", "x86-64-v3"],
                "url": "http://localhost:8080",
                "username": "",
                "password": ""
            },
            "arm_server": {
                "targets": ["aarch64"],
                "url": "http://localhost:8081",
                "username": "",
                "password": ""
            }
        }
    },
```


crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -320,14 +322,221 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
...
+        ];
+
+        // Check if build cache is enabled
+        if let Some(build_config) = &self.settings.build_cache_server {
+            if build_config.enabled {
+                display::action("Using build cache for binary distribution");
+                self.update_with_build_cache(build_config, commit, &binaries)
+                    .await?;
+            } else {
+                self.update_with_local_build(&binaries).await?;
```
piotrm50:
```suggestion
                display::action("Updating from build cache failed - building locally");
                self.update_with_local_build(&binaries).await?;
```

muXxer: Why failed? It's just not enabled, and therefore we do the normal flow.

crates/iota-build-cache-server/src/server.rs
```diff
@@ -0,0 +1,388 @@
...
+                    .unwrap())
+            }
+        },
+        Err(e) => {
+            warn!("Invalid build request: {e}");
+            Ok(Response::builder()
+                .status(StatusCode::BAD_REQUEST)
+                .header(CONTENT_TYPE, "text/plain")
+                .body(Full::new(Bytes::from(format!("Invalid request: {e}"))))
+                .unwrap())
```
Thoralf-M: There's a lot of repetition for these responses, might want to refactor with fns like these
```Rust
fn bad_request(msg: impl Into<String>) -> Response {
    (StatusCode::BAD_REQUEST, [(CONTENT_TYPE, "text/plain")], msg.into()).into_response()
}

fn internal_error(msg: impl Into<String>) -> Response {
    let msg = msg.into();
    error!("{msg}");
    (StatusCode::INTERNAL_SERVER_ERROR, [(CONTENT_TYPE, "text/plain")], msg).into_response()
}
```

crates/iota-build-cache-server/scripts/config.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Default build parameters
+export COMMIT="develop"     # Git commit/branch/tag to build
+export CPU_TARGET="native"  # CPU target (native, skylake, etc.)
+
+# Default binaries to build/check/download
+export BINARIES="iota,iota-node,stress"
+
+# Timeout settings (in seconds)
+export BUILD_TIMEOUT=2700  # 45 minutes
```
Thoralf-M, line 26: Is 45 minutes high enough? Who knows on which toaster this will run :trollface:

muXxer: it's a config file, that is not my problem :trollface: 
and the sever I want to run it on, should be fine with 20mins

crates/iota-build-cache-server/src/server.rs
muXxer: Fixed, thanks for the suggestion! Makes sense :D

jkrvivian (approved):

crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -320,14 +322,230 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
...
+    /// target
+    async fn detect_cpu_targets_for_instances(
+        &self,
+    ) -> TestbedResult<HashMap<String, Vec<Instance>>> {
+        let instances_needing_binaries = self
+            .instances()
+            .iter()
+            .filter(|i| i.role != InstanceRole::Metrics)
+            .cloned()
+            .collect::<Vec<_>>();
```
jkrvivian: There is a function `instances_without_metrics()` that can be used.

jkrvivian (changes requested):

crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -320,14 +322,230 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
...
+
+        if instances_needing_binaries.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        display::action("Detecting CPU targets for all instances");
+
+        let context = CommandContext::new();
+        let command = "rustc --print target-cpus".to_string();
```
jkrvivian:
```suggestion
        let command = "source \"$HOME/.cargo/env\" && rustc --print target-cpus".to_string();
```

I'm testing this feature, and this command was stuck because rustc wasn't in the `PATH` for non-interactive SSH sessions. So we need to source the envs.

crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -320,14 +322,230 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
...
+                // Extract the CPU name from "(currently cpu-name)"
+                if let Some(start) = line.find("(currently ") {
+                    let cpu_part = &line[start + 11..]; // Skip "(currently "
+                    if let Some(end) = cpu_part.find(')') {
+                        let cpu_name = cpu_part[..end].trim();
+                        return Some(cpu_name.to_string());
+                    }
+                }
+            }
+        }
```
jkrvivian: We should find the line, for example: `x86-64                  - This is the default target CPU for the current build target (currently x86_64-unknown-linux-gnu)` 
instead of 
`native - Select the CPU of the current host (currently apple-m4).`

And we take the beginning of the line `x86-64`, this is what Rust uses by default. The very specific CPU target extracted from `native` line would possibly not be a supported option for the cache server to build.

I was testing on AWS instances, and the CPU target is `skylake-avx512`, which is not working on the cache server.

jkrvivian (dismissed):

crates/iota-build-cache-server/src/client.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Create a new build cache client with basic authentication.
+    fn with_auth(
+        server_address: &str,
+        credentials: Option<(String, String)>,
+    ) -> Result<Self, std::net::AddrParseError> {
+        let socket_addr: SocketAddr = server_address.parse()?;
+
+        let build_instance_ip = socket_addr.ip().to_string();
+        let port = socket_addr.port();
+
```
jkrvivian: This does not support addresses like `http://` or `https://`. 😢

crates/iota-aws-orchestrator/src/orchestrator.rs
piotrm50: righty, nevermind

crates/iota-aws-orchestrator/src/orchestrator.rs
muXxer: Actually we need to make sure that the server runs on "x86_64", because that's what we should build for.
The [reed solomon lib](https://github.com/AndersTrier/reed-solomon-simd) will then be able to dynamically use avx2 instructions. Unfortunately we can't easily build for aarch64 on our build cache server.

alexsporn: > Unfortunately we can't easily build for aarch64 on our build cache server

We have a dedicated aarch64 server that we use for Docker builds that could build these if required.

muXxer: I would say we can add that later if needed. I made the supported targets configurable in the build-cache-server.

But x86_64 is hardcoded for now in the aws-orchestrator. We need to add different build-cache-server URLs per architecture if we want to add that feature.

muXxer: I added support for aarch64 now as well

crates/iota-aws-orchestrator/src/build_cache.rs
```diff
@@ -0,0 +1,401 @@
...
+            ));
+
+            let cache_client = BuildCacheClient::with_credentials(
+                cache_server.url.as_str(),
+                cache_server.username.clone(),
+                cache_server.password.clone(),
+            )
+            .map_err(|e| {
+                BuildCacheError::Cache(format!("Invalid server URL for {cpu_target}: {e}"))
+            })?;
```
jkrvivian, line 100: Maybe it's better to pre-build the `cache_client`s, so we don't need to re-create it every iteration.

muXxer: That's on purpose here, because there might be different build cache servers per cpu target 👍

filipdulic (approved):

alexsporn (approved):

PR #9424, opened 2025-11-25
feat(iota-core, iota-types): Add advance epoch v4

oliviasaa:
# Description of change

Enables the creation of a new version of advance epoch transaction and new advance epoch function in iota-framework to modify rewards based on scores. The existing validator reporting system remains whereby rewards can be slashed entirely when a validator is reported. Now we additionally modulate the rewards with the score provided as an argument to the new version of the advance epoch function.

### Release Notes

- [x] Protocol: Enables the creation of a new version of advance epoch transaction and new advance epoch function in iota-framework to modify rewards based on scores.
- [x] Indexer: Enables a new `ChangeEpochV4` system transaction which may appear in checkpoints for networks where `score_based_rewards` protocol config is enabled. Consequently **a dependency bump is necessary for `iota_data_ingestion_core` library consumers (custom indexers) to prevent outages**.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-core/src/authority/scorer.rs
+++ b/crates/iota-core/src/authority/scorer.rs
@@ -389,7 +389,7 @@ mod tests {
     use iota_types::messages_consensus::{MisbehaviorReportV1, VersionedMisbehaviorReport};
 
     use crate::authority::authority_per_epoch_store::scorer::{
-        ParametersV1, Scorer, calculate_median_report, calculate_scores_v1,
+        MAX_SCORE, ParametersV1, SCALE_FACTOR, Scorer, calculate_median_report, calculate_scores_v1,
     };
 
     fn mock_protocol_config(consensus_choice: ConsensusChoice) -> ProtocolConfig {
--- a/crates/iota-core/src/authority/scorer.rs
+++ b/crates/iota-core/src/authority/scorer.rs
@@ -8,7 +8,8 @@ use std::sync::{
 
 use iota_protocol_config::ProtocolConfig;
 use iota_types::{
-    messages_consensus::VersionedMisbehaviorReport, scoring_metrics::VersionedScoringMetrics,
+    messages_consensus::{MisbehaviorsV1, VersionedMisbehaviorReport},
+    scoring_metrics::VersionedScoringMetrics,
 };
 
 const MAX_SCORE: u64 = 2_u64.pow(16); // Note: must be consistent with MAX_SCORE in validator_set.move in iota-framework.
@@ -21,7 +22,7 @@ pub struct Scorer {
     pub(crate) current_local_metrics_count: Arc<VersionedScoringMetrics>,
     // The metrics counts received from other authorities, i.e., the information contained in the
     // MisbehaviourReports received by the authority. If an authority has not sent a report, its
-    // received_metrics will be all zeroed.
+    // entry in this vector will be all zeroed.
     received_metrics: Vec<VersionedScoringMetrics>,
     // Indicates whether an authority did not send any misbehavior reports in the epoch. We use
     // this to differentiate an authority that did not send a report from another one who sent
@@ -66,6 +67,7 @@ impl Scorer {
                             )
                         })
                         .collect();
+                let scale_factor = 2_u64.pow(16);
                 let parameters = ParametersV1 {
                     max_score: MAX_SCORE,
                     scale_factor: SCALE_FACTOR,
@@ -77,6 +79,24 @@ impl Scorer {
                         SCALE_FACTOR * 35 / 100,
                     ],
                 };
+                // Assert that the allowance for major misbehaviors is 0,
+                // maximum is 1 and weight is 1. This is because major misbehaviors should
+                // reduce the score to 0 is there are any occurrences.
+                // Only equivocation is considered a major misbehavior in this version.
+                assert!(
+                    parameters
+                        .allowances
+                        .iter_major_misbehaviors()
+                        .all(|&a| a == 0)
+                        && parameters
+                            .maximums
+                            .iter_major_misbehaviors()
+                            .all(|&m| m == 1)
+                        && parameters
+                            .weights
+                            .iter_major_misbehaviors()
+                            .all(|&w| w == 1)
+                );
 
                 Self {
                     current_local_metrics_count,
@@ -175,7 +195,7 @@ impl Scorer {
 /// VersionedMisbehaviorReport
 fn calculate_median_report(
     reports_and_voting_power: &[(VersionedMisbehaviorReport, VotingPower)],
-) -> Vec<MedianMetricVec> {
+) -> MisbehaviorsV1<MedianMetricVec> {
     // Calls to this method should ensure that we have at least one report to
     // process.
     assert!(!reports_and_voting_power.is_empty());
@@ -200,7 +220,7 @@ fn calculate_median_report(
     let median_report = reports_and_voting_power_per_metric
         .iter_mut()
         .map(|vec| calculate_weighted_median(vec.as_mut_slice()))
-        .collect::<Vec<MetricVec>>();
+        .collect::<MisbehaviorsV1<MedianMetricVec>>();
     median_report
 }
 
@@ -252,11 +272,12 @@ struct ParametersV1 {
     max_score: u64,
     scale_factor: u64,
     // Allowed misbehaviors without any punishment
-    allowances: Vec<u64>,
+    allowances: MisbehaviorsV1<u64>,
     // Number of misbehaviors that lead to zero score
-    maximums: Vec<u64>,
-    // Weights for each non-equivocation metric. Their sum + baseline_score = scale_factor
-    weights: Vec<u64>,
+    maximums: MisbehaviorsV1<u64>,
+    // Weights for each metric. The sum of minor misbehavior weights + baseline_score =
+    // scale_factor. Major misbehavior weights are either 0 or 1.
+    weights: MisbehaviorsV1<u64>,
 }
 
 // Aliases for better readability.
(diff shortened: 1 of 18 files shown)


crates/iota-protocol-config/src/lib.rs
```diff
@@ -1448,7 +1448,12 @@ impl ProtocolConfig {
     }
 
     pub fn score_based_rewards(&self) -> bool {
-        self.feature_flags.score_based_rewards
+        let res = self.feature_flags.score_based_rewards;
+        assert!(
+            res && self.scorer_version.is_none(),
```
cyberphysic4l:
```suggestion
            res || self.scorer_version.is_none(),
```

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1448,7 +1448,12 @@ impl ProtocolConfig {
     }
 
     pub fn score_based_rewards(&self) -> bool {
-        self.feature_flags.score_based_rewards
+        let res = self.feature_flags.score_based_rewards;
```
cyberphysic4l: Why the variable name `res`?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1448,7 +1448,12 @@ impl ProtocolConfig {
...
+        assert!(
+            res && self.scorer_version.is_none(),
+            "score_based_rewards requires scorer_version to be set"
+        );
+        res && self.select_committee_from_eligible_validators()
```
cyberphysic4l: Why do we require `select_committee_from_eligible_validators`?

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -948,6 +948,31 @@ mod checked {
...
+            CallArg::IOTA_SYSTEM_MUT, // wrapper: &mut IotaSystemState
+            CallArg::Pure(bcs::to_bytes(&params.epoch).unwrap()), // new_epoch: u64
+            CallArg::Pure(bcs::to_bytes(&params.next_protocol_version.as_u64()).unwrap()), /* next_protocol_version: u64 */
+            CallArg::Pure(bcs::to_bytes(&params.storage_rebate).unwrap()), // storage_rebate: u64
+            CallArg::Pure(bcs::to_bytes(&params.non_refundable_storage_fee).unwrap()), /* non_refundable_storage_fee: u64 */
+            CallArg::Pure(bcs::to_bytes(&params.reward_slashing_rate).unwrap()), /* reward_slashing_rate: u64 */
+            CallArg::Pure(bcs::to_bytes(&params.epoch_start_timestamp_ms).unwrap()), /* epoch_start_timestamp_ms: u64 */
+            CallArg::Pure(bcs::to_bytes(&params.max_committee_members_count).unwrap()), /* max_committee_members_count: u64 */
+            CallArg::Pure(bcs::to_bytes(&params.eligible_active_validators).unwrap()), /* eligible_active_validators: Vec<u64> */
+                                                                                       /*  CallArg::Pure(bcs::to_bytes(&params.scores).unwrap()), // scores: Vec<u64> */
```
cyberphysic4l: TODO: uncomment when the move code is added.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -1087,9 +1113,10 @@ mod checked {
             reward_slashing_rate: protocol_config.reward_slashing_rate(),
             epoch_start_timestamp_ms: change_epoch_v2.epoch_start_timestamp_ms,
             max_committee_members_count: protocol_config.max_committee_members_count(),
-            // AdvanceEpochV2 does not use this field, but keeping them to avoid creating a
+            // AdvanceEpochV2 does not use those fields, but keeping them to avoid creating a
```
cyberphysic4l:
```suggestion
            // AdvanceEpochV2 does not use these fields, but keeping them to avoid creating a
```

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -1177,8 +1207,9 @@ mod checked {
             epoch_start_timestamp_ms: change_epoch_v4.epoch_start_timestamp_ms,
             max_committee_members_count: protocol_config.max_committee_members_count(),
             eligible_active_validators: change_epoch_v4.eligible_active_validators,
+            scores: change_epoch_v4.scores,
```
cyberphysic4l, line 1209: Remove the TODO comment above to pass scores.

cyberphysic4l (changes requested):

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -1274,29 +1257,6 @@ fun adjust_next_epoch_commission_rate(validators: &mut vector<ValidatorV1>) {
     }
 }
 
-/// Process the validator report records of the epoch and return the addresses of the
-/// non-performant committee validators according to the input threshold.
-fun compute_slashed_validators(
```
alexsporn, line 1279: why is this being removed here?

cyberphysic4l: I wondered the same thing. This commit was WIP from Olivia and seemed to be removing the existing reporting system. I added it back in because I wasn't aware this was a planned change. 
Now the slashing from reporting is applied after the adjustment of rewards based on score.

cyberphysic4l: I also kept the ValidatorEpochInfoEventV1 alone and use the existing field for the tallying rule score to place the overall score (validator score * slashing). That field was previously either 0 (slashed) or 1 (not slashed). Now it's between 0 (zero score and/or slashed) and MAX_SCORE (full score and not slashed).

cyberphysic4l (approved):

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -1326,31 +1331,38 @@ fun compute_adjusted_reward_distribution(
...
         // Check if the validator is slashed
         let adjusted_staking_reward_amount = if (
             slashed_validator_indices_set.contains(&committee_members[i])
         ) {
             // Use the slashing rate to compute the amount of staking rewards slashed from this punished validator.
             // Use u128 to avoid multiplication overflow.
             let staking_reward_adjustment_u128 =
-                ((unadjusted_staking_reward_amount as u128) * (reward_slashing_rate as u128)) / BASIS_POINT_DENOMINATOR;
-            unadjusted_staking_reward_amount - (staking_reward_adjustment_u128 as u64)
+                ((score_adjusted_staking_reward_amount as u128) * (reward_slashing_rate as u128)) / BASIS_POINT_DENOMINATOR;
```
alexsporn: I think this `u128` cast is not needed anymore since the value should already be an `u128`

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -1133,6 +1160,9 @@ mod checked {
             epoch_start_timestamp_ms: change_epoch_v3.epoch_start_timestamp_ms,
             max_committee_members_count: protocol_config.max_committee_members_count(),
             eligible_active_validators: change_epoch_v3.eligible_active_validators,
+            // AdvanceEpochV3 does not use this field, but keeping them to avoid creating a
```
alexsporn: Maybe already put it in plural form like above?

alexsporn (approved):
Left small nitpicks, but rest lgtm

lzpap:
Updated the release notes with a warning for the `iota_data_ingestion_core` consumers. Please keep in mind, that any change in checkpoint contents makes custom indexer deployments break without a dependency update, causing downtime.
For our own indexer it is usually not a problem since we release the new sw together with the node.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -49,7 +49,7 @@ impl Scorer {
                     protocol_config,
                 ));
 
-                let max_score = 2_u64.pow(16);
+                let max_score = 2_u64.pow(16); // Note: must be consistent with MAX_SCORE in validator_set.move in iota-framework.
```
valeriyr: I would propose to create a constant `MAX_SCORE` instead of using a local variable.

crates/iota-framework/packages/iota-system/sources/validator_set.move
```diff
@@ -161,6 +162,7 @@ const EValidatorSetEmpty: u64 = 13;
 const ENotACommitteeValidator: u64 = 14;
 const EInvalidStakeAmount: u64 = 15;
 const EInvalidEligibleValidatorIndex: u64 = 16;
+const EInvalidRewardAdjustmentData:u64 = 19;
```
valeriyr:
```suggestion
const EInvalidRewardAdjustmentData: u64 = 19;
```
It is not a big issue, but if you use VSCode, I suggest installing the `Move Formatter` extension and, optionally, setting up autoformatting on save. Unfortunately, we don't have a formatting check on CI.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1448,7 +1448,12 @@ impl ProtocolConfig {
     }
 
     pub fn score_based_rewards(&self) -> bool {
-        self.feature_flags.score_based_rewards
+        let score_based_rewards = self.feature_flags.score_based_rewards;
+        assert!(
+            score_based_rewards || self.scorer_version.is_none(),
```
valeriyr:
```suggestion
            !score_based_rewards || self.scorer_version.is_some(),
```
Just to double-check the condition, from the related message, it looks like we need to check that `scorer_version` is set when `score_based_rewards` is enabled.

valeriyr (approved):
Looks good to me!
Just in case, some simulation tests are broken, but probably it is ok for this PR, since I don't see the snapshots update included that is required for ChangeEpochV4 execution.

crates/iota-protocol-config/src/lib.rs
cyberphysic4l: You're right, thanks for catching that.

Issue #8846, opened 2025-10-13
Node: Investigate if `iota-aws-orchestrator` fits for our performance tests


muXxer:
We need to be able to run a globally distributed network to run some tests.

Following things need to be configurable: 
- validator count
- hardware specs
- spam base load (TPS, type of messages, needs to be reproducible)
- test with and without indexer
- duration of the test

In case it is a low-hanging fruit, it would be nice if we could also use the iota-mainnet-genesis to have some bigger ledger state for the test.

We also need to check how this crate produces reports / allows tracing.

- [x] We would like to get the end-to-end latency metrics (support multiple workloads)
- [ ] Can we export the data that is collected in grafana, or should we collect the prometheus metrics for all nodes as well and then use it as a data source later?
- [ ]  Add indexer to the setup, and add a flag to enable the clients to query from the indexer
     - [ ] Figure out how `stress` queries indexer / fullnode

jkrvivian:
## Notes for running iota-aws-orchestrator

1. deploy: `cargo run --bin iota-aws-orchestrator -- testbed deploy --instances 4 --dedicated-clients 2`
2. run benchmark: `cargo run --bin iota-aws-orchestrator -- benchmark --committee 4 fixed-load --loads 100 --duration 180 --dedicated-clients 2 --benchmark-type 50`
    - `--use-internal-ip-addresses`  : use private ips
    - `--skip-testbed-update` : to skip rebuilding the project
    - `--skip-testbed-configuration` : to skip running configuration
    - `--log-processing` : to download logs from nodes and clients
3. stop / destroy instances: `cargo run --bin iota-aws-orchestrator -- testbed destroy`    
    - `--keep-monitoring: stop / destroy all instances, and keep monitoring server running`

## The `stress` binary in iota-benchmark

The spam load is generated by the `stress` binary in `crates/iota-benchmark/src/bin/stress.rs`. To further customize transaction types for spamming, modify the `fn client_command()` in `crates/iota-aws-orchestrator/src/protocol/iota.rs`.

The stress test supports the following transaction types:
- Transfer object: sending the same object between sender and receiver
- Shared counter: Increase the counter in a shared object
- Batch payment:  send 1 IOTA to `batch_payment_size` recipients
- Delegation: stake the validator: `1. Split 1 IOTA → 2. Delegate to validator -> 1. -> 2.`
- Shared object deletion: Create/ Read/ Inc/ Delete shared objects, operation get called randomly
[...]

PR #9429, opened 2025-11-25
fix(core): Fix fullnode setup in iota-aws-orchestrator

jkrvivian:
# Description of change

This PR fixes the fullnode IP configuration, updates the health check to use the latest-checkpoint API, and includes several minor improvements.

Part of #8846 

### DIFF
--- a/crates/iota-aws-orchestrator/src/orchestrator.rs
+++ b/crates/iota-aws-orchestrator/src/orchestrator.rs
@@ -430,11 +430,12 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
 
             // Wait until all fullnodes are reachable (otherwise clients might fail when a
             // fullnode is not listening yet).
+            display::action("Await fullnode ready...");
             let commands = self
                 .client_instances
                 .iter()
                 .cloned()
-                .map(|i| (i, "curl http://127.0.0.1:9000".to_owned())); // fullnodes default json RPC address
+                .map(|i| (i, "curl http://127.0.0.1:9000 -H 'Content-Type: application/json' -d '{\"jsonrpc\":\"2.0\",\"method\":\"iota_getLatestCheckpointSequenceNumber\",\"params\":[],\"id\":1}'".to_owned())); // fullnodes default json RPC address
             self.ssh_manager.wait_for_success(commands).await;
 
             display::done();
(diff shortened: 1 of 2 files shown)


crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -477,18 +478,18 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
...
-        metrics_commands.append(
-            &mut self
-                .protocol_commands
-                .nodes_metrics_command(self.node_instances.clone(), parameters),
-        );
+        // metrics_commands.append(
+        //     &mut self
+        //         .protocol_commands
+        //         .nodes_metrics_command(self.node_instances.clone(), parameters),
+        // );
```
muXxer: So can we remove the whole block then? Is this now available from the consensus nodes?

crates/iota-aws-orchestrator/src/orchestrator.rs
```diff
@@ -430,11 +430,12 @@ impl<P: ProtocolCommands<T> + ProtocolMetrics, T: BenchmarkType> Orchestrator<P,
 
             // Wait until all fullnodes are reachable (otherwise clients might fail when a
             // fullnode is not listening yet).
+            display::action("Await fullnode ready...");
             let commands = self
                 .client_instances
                 .iter()
                 .cloned()
-                .map(|i| (i, "curl http://127.0.0.1:9000".to_owned())); // fullnodes default json RPC address
+                .map(|i| (i, "curl http://127.0.0.1:9000 -H 'Content-Type: application/json' -d '{\"jsonrpc\":\"2.0\",\"method\":\"iota_getLatestCheckpointSequenceNumber\",\"params\":[],\"id\":1}'".to_owned())); // fullnodes default json RPC address
```
muXxer: Can you add a comment that we query the latest checkpoint to be sure the nodes are fully started?

crates/iota-aws-orchestrator/src/protocol/iota.rs
```diff
@@ -195,6 +195,18 @@ impl ProtocolCommands<IotaBenchmarkType> for IotaProtocol {
...
+                    false => &instance.main_ip,
+                };
+
+                // Update fullnode.yaml to use the correct IP for P2P config using a safer approach
+                // Escape quotes for proper handling inside tmux wrapper
+                let update_p2p_config = format!(
+                    "sed -i 's|listen-address: \\\"127.0.0.1:|listen-address: \\\"0.0.0.0:|' {0} && sed -i 's|external-address: /ip4/127.0.0.1/|external-address: /ip4/{1}/|' {0}",
+                    config_path.display(),
+                    fullnode_ip
+                );
```
muXxer: Instead of "using a safer approach", can you write in the comment what it actually does?
I guess replacing localhost listen address with 0.0.0.0 and replacing external address with the actual fullnode ip?

muXxer (changes requested):

crates/iota-aws-orchestrator/src/orchestrator.rs
jkrvivian: Yes, just checked, they are available now.

muXxer (approved):

piotrm50 (approved):

PR #9435, opened 2025-11-26
feat(iota): allow // comments in `iota client ptb` commands

### LINKED ISSUE #9432: `iota client ptb` fails if there is a # comment
## Bug description

Running a script with a PTB command that includes a comment like `# some comment` fails with `command not found: --move-call`
```shell
iota client ptb \
# some comment
--move-call iota::tx_context::sender \
--assign sender \
--split-coins gas "[1]" \
--assign coin \
--transfer-objects "[coin]" sender \
--dry-run
```

There is even a file that includes comments https://github.com/iotaledger/iota/blob/develop/crates/iota/tests/ptb_files/comments/ptb_with_comments.ptb, but this doesn't seem to be sufficient to work with bash/zsh

When fixing this issue, only run `cargo` commands in the `crates/iota/` dir as it will otherwise take much longer to compile. The ptb logic is in `crates/iota/src/client_ptb/`

## Expected behaviour

Commented lines should be ignored

## Actual behaviour

ptb parser stops

Thoralf-M:
# Description of change

Allow `//` comments in `iota client ptb` commands, the comments must be finished with `\` to continue with the commands. This is a limitation of bash/zsh and also the reason why `#` can't be used, as the parts afterwards wouldn't be visible to the `iota client ptb` command.

This allows
```zsh
iota client ptb \
// Get the sender address \
--move-call iota::tx_context::sender \
--assign sender \
// Split 1 NANOS from gas coin \
--split-coins gas "[1]" \
--assign coin \
// Transfer the new coin to the sender \
--transfer-objects "[coin]" sender \
--dry-run

# normal comment
iota client ptb \
// another test \
--dry-run

# commented PTB call
iota client ptb \
// another test \
// --split-coins gas "[1]" works as comment \
--dry-run

# commented PTB call, but with text before, still runs
iota client ptb \
// will still run: --split-coins gas "[1]" \
--dry-run

# empty comment, next PTB call will be ignored
iota client ptb \
// \
// --split-coins gas "[1]" \
--dry-run
```

Fixes https://github.com/iotaledger/iota/issues/9432

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Add `//` comment support in `iota client ptb` commands
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota/src/client_ptb/lexer.rs
+++ b/crates/iota/src/client_ptb/lexer.rs
@@ -61,21 +61,50 @@ impl<'l, I: Iterator<Item = &'l str>> Lexer<'l, I> {
         self.buf = "";
     }
 
-    /// Skip comment tokens until we hit a command (starts with --) or EOF.
-    /// This handles multi-word comments like "// this is a comment".
+    /// Skip tokens until we find the next real command or EOF.
+    ///
+    /// Since shell line continuations (`\`) convert newlines to spaces, we
+    /// cannot detect line boundaries. Instead, we use `//` and `--` as
+    /// delimiters:
+    ///
+    /// - `//` marks the start of a comment
+    /// - `--` marks the start of a command
+    ///
+    /// To comment out a `--command`, place `//` directly
+    /// before it (as the immediately preceding token). Any `--` that follows
+    /// non-`//` tokens is treated as a real command.
+    ///
+    /// # Examples
+    ///
+    /// ```text
+    /// // comment text --cmd       → "--cmd" runs (text before it)
+    /// // --cmd                    → "--cmd" is commented (// directly before)
+    /// // --cmd1 --cmd2            → "--cmd1" commented, "--cmd2" runs
+    /// // // --cmd                 → "--cmd" is commented (// directly before)
+    /// // text // --cmd            → "--cmd" is commented (// directly before)
+    /// ```
     fn skip_comment(&mut self) {
         self.skip_rest_of_token();
-        // Skip subsequent tokens until we hit a command or EOF
+        let mut after_comment_marker = true;
+
         for next in self.tokens.by_ref() {
-            self.offset += next.len() + 1; // +1 for space between tokens
-            if next.starts_with("--") || next.starts_with("//") {
-                // Found next command or another comment, set buffer to this token
-                self.buf = next;
-                return;
+            self.offset += next.len() + 1;
+
+            if next.starts_with("//") {
+                after_comment_marker = true;
+            } else if next.starts_with("--") {
+                if after_comment_marker {
+                    // "--" directly after "//" → commented out, keep skipping
+                    after_comment_marker = false;
+                } else {
+                    // "--" after other tokens → real command
+                    self.buf = next;
+                    return;
+                }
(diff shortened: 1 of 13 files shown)


docs/content/developer/references/cli/ptb.mdx
```diff
@@ -15,7 +15,7 @@ The `client ptb` command allows you to specify the transactions for execution in
 The following list itemizes all the available args for the `iota client ptb` command. Use the `--help` for a long help version that includes some examples on how to use this command.
 
 ```
-Build, preview, and execute programmable transaction blocks. Depending on your shell, you might have to use quotes around arrays or other passed values. Use --help to see examples for how to use the core functionality of this command.
+Build, preview, and execute programmable transaction blocks. Depending on your shell, you might have to use quotes around arrays or other passed values. Use // for comments (with trailing \). Use --help to see examples for how to use the core functionality of this command.
```
thibault-martinez:
```suggestion
Build, preview, and execute programmable transaction blocks. Depending on your shell, you might have to use quotes around arrays or other passed values. Use `//` for comments (with trailing `\`). Use --help to see examples for how to use the core functionality of this command.
```

crates/iota/src/client_ptb/lexer.rs
```diff
@@ -54,6 +54,31 @@ impl<'l, I: Iterator<Item = &'l str>> Lexer<'l, I> {
...
     }
 
+    /// Skip the rest of the current shell token (used for comments).
+    fn skip_rest_of_token(&mut self) {
+        let len = self.buf.len();
+        self.offset += len;
+        self.buf = "";
+    }
+
+    /// Skip comment tokens until we hit a command (starts with --) or EOF.
```
thibault-martinez: Maybe I misunderstand but shouldn't we skip until an EOL? Otherwise we can run a commented command?

DaughterOfMars (commented):
tbh I'm not sure I see the utility here

Thoralf-M:
> tbh I'm not sure I see the utility here

Explaining what happens in a PTB could be useful and also commenting out part of a PTB for testing different combinations

DaughterOfMars:
> > tbh I'm not sure I see the utility here
> 
> Explaining what happens in a PTB could be useful and also commenting out part of a PTB for testing different combinations

The problem I see is that this doesn't allow you to write a shell file and then copy paste it into the command line because of the comments having a different format. So this allows you to write commands instead in, what, a `.md` file? I suppose this helps for our examples in docs... I just don't know.

crates/iota/src/client_ptb/lexer.rs
Thoralf-M: Hm good point, the problem is we can't detect end of lines in the parser since the shell removes this information before
What we could do is to ignore `--` if it comes directly after `//`, but if we have an empty line, so just `// \` and then a command again, it would also ignore this command

Thoralf-M:
Closing since we can't detect newlines in the ptb parser and then the behavior can be confusing

PR #9443, opened 2025-11-26
feat(cli): add AA support

### LINKED ISSUE #9408: AA CLI support

thibault-martinez:
# Description of change

Add Account Abstraction support to the CLI.

Fixes https://github.com/iotaledger/iota/issues/9408

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [X] CLI: Abstract Account support
        -  `KeyStore` can store account addresses (`StoredKey::Account`)
        - `iota keytool` can handle account addresses
        - `iota client add-account` and `iota client sign` commands
        - `iota client` `--auth-call-args` and `--auth-type-args` options
        - `iota client ptb` `--auth-call-args` and `--auth-type-args` options
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -804,6 +809,9 @@ impl From<EndOfEpochTransactionKind> for crate::transaction::EndOfEpochTransacti
                     eligible_active_validators: change_epoch_v3.eligible_active_validators,
                 })
             }
+            EndOfEpochTransactionKind::ChangeEpochV4(_change_epoch_v4) => {
+                todo!()
```
DaughterOfMars, line 813: which lib is this missing from?

crates/iota-keys/src/keystore.rs
```diff
@@ -209,6 +213,7 @@ pub struct Alias {
 pub enum StoredKey {
     #[serde(with = "serde_iota_keypair")]
     KeyPair(IotaKeyPair),
+    Account(IotaAddress),
```
DaughterOfMars, line 216: Could be nice to also have a `default_auth_call_args` and `default_auth_type_args` here

Thoralf-M: What would default_auth_call_args be? Don't see anything default useful here as if something default would be sufficient to unlock the account, anyone could just copy it and also do transactions from this account

DaughterOfMars: well isn't that the point of the keystore? That it stores enough info to sign your transactions?

crates/iota-rest-api/src/transactions/execution.rs
```diff
@@ -314,6 +314,7 @@ fn coins(objects: &[Object]) -> impl Iterator<Item = (&Address, Coin)> + '_ {
             Owner::Address(address) => address,
             Owner::Object(object_id) => object_id.as_address(),
             Owner::Shared { .. } | Owner::Immutable => return None,
+            _ => unreachable!("a new enum variant was added and needs to be handled"),
```
Thoralf-M, line 317: Should we mention the enum type? Same below with `Transaction`
```suggestion
            _ => unreachable!("a new Owner enum variant was added and needs to be handled"),
```

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -93,14 +93,68 @@ impl std::fmt::Display for PTBCommandResult {
     }
 }
 
+/// Result of extracting auth arguments from the initial PTB args.
+struct ExtractedAuthArgs {
+    /// The remaining args after auth arguments are removed.
+    remaining_args: Vec<String>,
```
Alex6323: nit: would read a bit better if that would actually be the last field in the struct

crates/iota-rest-api/src/transactions/execution.rs
thibault-martinez: I didn't in the SDK, my reasoning was that we would forget if/when we rename the enum, I can do it anyway if you want. It's only meant to us so I'd say it's fine tbh.

crates/iota-keys/src/keystore.rs
Thoralf-M: For key based accounts yes, for accounts with the auth it very much depends on the auth function. We added the accounts there now more so you can switch to it and can use commands like `iota client gas` and prepare transactions without manually needing to set `--sender <account address>` but it's not always a simple sign operation that is needed for the authentication
One could potentially let the user add some script for the authentication, which gets executed to get the call args, but that's more complex and we said this can be added later if see a need for it

crates/iota-types/src/iota_sdk_types_conversions.rs
thibault-martinez: It's been added to the external SDK but the PR to add it in the monorepo hasn't been merged yet

muXxer (approved):

crates/iota/src/client_ptb/ptb.rs
```diff
@@ -93,14 +93,68 @@ impl std::fmt::Display for PTBCommandResult {
...
+                let mut values = Vec::new();
+                while let Some(next) = iter.peek() {
+                    if next.starts_with("--") {
+                        break;
+                    }
+                    values.push(iter.next().unwrap().clone());
+                }
+
+                if arg == "--auth-call-args" {
+                    auth_call_args = Some(values);
```
Alex6323, line 128: Could it happen that `values` remains empty despite `--auth-call-args` or `--auth-type-args` being found? Then it also should print the help dialog, but it currently wouldn't, right? You could add a small method to this struct, that you call in the `PTB::execute` that checks it's valid?

crates/iota-rest-api/src/transactions/execution.rs
Thoralf-M: Ok fine

Thoralf-M (approved):

crates/iota-keys/src/keystore.rs
DaughterOfMars: sure but you could easily have an auth function where you just provide static data

crates/iota-types/src/iota_sdk_types_conversions.rs
DaughterOfMars: cool, okay

crates/iota/src/client_ptb/ptb.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/9443/commits/e8cfb3756f3cd781f8223f4ab07280fb78082b92

crates/iota-keys/src/keystore.rs
thibault-martinez: After discussing we Thoralf, we don't think such an auth function is going to make sense outside of testing purposes so having a default is not necessary

DaughterOfMars: Alrighty

DaughterOfMars (approved):

crates/iota/tests/cli_tests.rs
```diff
@@ -6521,3 +6521,223 @@ async fn test_ptb_gas_coins_smashing() -> Result<(), anyhow::Error> {
...
+        display: HashSet::new(),
+    })
+    .execute(context)
+    .await?;
+    assert!(matches!(
+        ptb_resp,
+        IotaClientCommandResult::TransactionBlock(ref tx) if tx.effects.as_ref().unwrap().status().is_ok()
+    ));
+
+    // Perform auth transaction with client sign command
```
thibault-martinez, line 6692: Comment is a bit off?

Thoralf-M: Wouldn't say so, here it starts to prepare the tx that gets then signed

miker83z (approved):

piotrm50 (approved):
approving changes in `iota-types`

Alex6323 (approved):

PR #9452, opened 2025-11-27
feat(rest-kv): Add support for multiget requests

### LINKED ISSUE #9413: feat(rest-kv): Add support for bulk operations
Some requests to the indexer that may require fallback use of the REST KV may need to read data of many transactions/objects at once. (e.g. to compute object and balance changes of a transaction that touches 100 objects, we will need to fetch those 100 objects from the KV, along with the transaction).

To not flood the server with thousands of requests in such cases, we can introduce multi-get requests to the REST KV server, so that requests for multiple keys can be handled in scope of single request.

tomxey:
# Description of change

This change introduces new `POST /{item_type}` endpoint to iota-rest-kv.

The endpoint has similar function to already existing `GET /{item_type}/{key}` endpoint, but it allows many keys to be fetched at once, not only one.

Examples:

- Single item GET for `0x5` obj at version `1844674`: `curl -X GET http://127.0.0.1:3555/ob/AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAXCJRwAAAAAAA`
- Single item GET for `0x6` obj at version `1844675`: `curl -X GET http://127.0.0.1:3555/ob/AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAbDJRwAAAAAAA`
- Multiget to get both those objects: 
```
curl -X POST http://127.0.0.1:3555/ob \
  -H "Content-Type: application/json" \
  -d '{
  "keys": [
    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAXCJRwAAAAAAA",
    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAbDJRwAAAAAAA"
  ]
}'
```

This returns a BCS-serialized `Vec<Option<Bytes>>` with the same length and order as the `keys` array in the input payload. Found values will be `Some`, not found values will be `None`.

The base64-url encoding of keys on the input was kept in the multiget version, since that is the format `iota-storage` expects when constructing the `Key`s.

fixes https://github.com/iotaledger/iota/issues/9413

The server has been started locally, connected to the devnet bigtable, and queried with custom script to test different multi and single object get scenarios.

### DIFF
--- a/crates/iota-rest-kv/src/main.rs
+++ b/crates/iota-rest-kv/src/main.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{fs, path::PathBuf};
+use std::{fs, num::NonZeroUsize, path::PathBuf};
 
 use anyhow::Result;
 use clap::Parser;
@@ -45,6 +45,12 @@ pub struct RestApiConfig {
     #[serde(flatten)]
     pub kv_store_config: KvStoreConfig,
     pub server_address: std::net::SocketAddr,
+    #[serde(default = "default_multiget_max_items")]
+    pub multiget_max_items: NonZeroUsize,
+}
+
+fn default_multiget_max_items() -> NonZeroUsize {
+    NonZeroUsize::new(100).unwrap()
 }
 
 #[tokio::main]
--- a/crates/iota-rest-kv/src/main.rs
+++ b/crates/iota-rest-kv/src/main.rs
@@ -50,7 +50,7 @@ pub struct RestApiConfig {
 }
 
 fn default_multiget_max_items() -> NonZeroUsize {
-    NonZeroUsize::new(100).unwrap()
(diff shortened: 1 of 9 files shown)


crates/iota-rest-kv/src/main.rs
```diff
@@ -45,6 +45,12 @@ pub struct RestApiConfig {
     #[serde(flatten)]
     pub kv_store_config: KvStoreConfig,
     pub server_address: std::net::SocketAddr,
+    #[serde(default = "default_multiget_max_items")]
+    pub multiget_max_items: NonZeroUsize,
+}
+
+fn default_multiget_max_items() -> NonZeroUsize {
+    NonZeroUsize::new(100).unwrap()
```
sergiupopescu199:
```suggestion
    NonZeroUsize::new(100).expect("value should be greater than 0")
```

sergiupopescu199 (approved):
lgtm ✨

kodemartin (approved):
lgtm

PR #9453, opened 2025-11-27
feat(move): Add custom attributes

miker83z:
# Description of change
This PR cherry picks 2 commits in common to enable authenticator and move view attributes at the move level.
Changes:
1. Added a generic FlavoredAttribute which can then allow us to define new attributes within the iota_mode compiler part.
2. Introduces a new protocol feature flag (iota_metadata_module_bytes) to allow the insertion of metadata in the serialized compiled modules (bytecode). This kind of operation was disabled before this PR, with the flag no_extraneous_module_bytes. Instead, iota_metadata_module_bytes  allows to fill this metadata field during the execution of build_from_resolution_graph. At the module deserializer level, the only performed check is that metadata have only 1 field and the metadata key is hardcoded. During the execution of the iota-verifier, the metadata value is checked for having well formed metadata struct (later authenticate and view functions checks will be performed here).

### Release Notes

- [x] Protocol: Added metadata_in_module_bytes flag to allow the insertion of IOTA specific metadata in the serialized compiled modules (bytecode).
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -372,6 +372,11 @@ struct FeatureFlags {
     // If true, then transactions are committed only for traversed headers
     #[serde(skip_serializing_if = "is_false")]
     consensus_commit_transactions_only_for_traversed_headers: bool,
+
+    // If true, it allows metadata bytes indexed with the iota key in a compiled module
+    // This flag is used to provide the correct MoveVM configuration for clients.
+    #[serde(skip_serializing_if = "is_false")]
+    iota_metadata_module_bytes: bool,
 }
 
 fn is_true(b: &bool) -> bool {
@@ -1261,6 +1266,10 @@ impl ProtocolConfig {
 
     pub fn passkey_auth(&self) -> bool {
         self.feature_flags.passkey_auth
+    }
+
+    pub fn iota_metadata_module_bytes(&self) -> bool {
+        self.feature_flags.iota_metadata_module_bytes
     }
 
     pub fn max_transaction_size_bytes(&self) -> u64 {
@@ -2314,6 +2323,9 @@ impl ProtocolConfig {
                     // Enable committing transactions only for traversed headers in Starfish
                     cfg.feature_flags
                         .consensus_commit_transactions_only_for_traversed_headers = true;
+                    if chain != Chain::Mainnet && chain != Chain::Testnet {
+                        cfg.feature_flags.iota_metadata_module_bytes = true;
+                    }
                 }
                 // Use this template when making changes:
                 //
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -19,7 +19,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-pub const MAX_PROTOCOL_VERSION: u64 = 16;
+pub const MAX_PROTOCOL_VERSION: u64 = 17;
 
 // Record history of protocol version allocations here:
 //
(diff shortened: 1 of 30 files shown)


thibault-martinez (approved):

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2314,6 +2323,9 @@ impl ProtocolConfig {
                     // Enable committing transactions only for traversed headers in Starfish
                     cfg.feature_flags
                         .consensus_commit_transactions_only_for_traversed_headers = true;
+                    if chain != Chain::Mainnet && chain != Chain::Testnet {
+                        cfg.feature_flags.iota_metadata_module_bytes = true;
```
lzpap: @miker83z do we want to enable this already on devnet/alphanet?

miker83z: Yes, just alphanet for now. We can cherrypick the squashed commit of this PR into develop and include it in the next release. This would allow the move view attribute to be be implemented more easily, independently from anything else.

Then, I'd wait the enabling of either the view attribute or authenticator attribute before enabling iota metadata also on devnet.

bingyanglin (approved):
Approved the files owned by the node team.

crates/iota-protocol-config/src/lib.rs
alexsporn: There is no differentiation between Alphanet and Devnet, both are "unknown" Chains

crates/iota-protocol-config/src/lib.rs
```diff
@@ -372,6 +372,11 @@ struct FeatureFlags {
     // If true, then transactions are committed only for traversed headers
     #[serde(skip_serializing_if = "is_false")]
     consensus_commit_transactions_only_for_traversed_headers: bool,
+
+    // If true, it allows metadata bytes indexed with the iota key in a compiled module
+    // This flag is used to provide the correct MoveVM configuration for clients.
+    #[serde(skip_serializing_if = "is_false")]
+    iota_metadata_module_bytes: bool,
```
alexsporn: I understand that it is because it is metadata on the `iota` flavor of Move, but why do we need to write iota here in the config?

crates/iota-types/src/move_package.rs
```diff
@@ -791,3 +791,111 @@ fn build_upgraded_type_origin_table(
...
+                ),
+            }),
+        }
+    }
+}
+
+/// The list of iota attribute types recognized by the compiler.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum IotaAttribute {
+    // Attribute(Attribute),
```
Thoralf-M, line 876: Why is this an empty enum? Is it even needed then? Or if you want to populate it in the future, then maybe it should have `#[non_exhaustive]`?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1263,6 +1268,10 @@ impl ProtocolConfig {
         self.feature_flags.passkey_auth
     }
 
+    pub fn iota_metadata_module_bytes(&self) -> bool {
```
muXxer: Can we keep the same order for the getters, as for the definition of the protocol parameters?
IMO we should move this getter after `pub fn consensus_commit_transactions_only_for_traversed_headers(&self) -> bool {`

miker83z: Yes, right [2ba7b55](https://github.com/iotaledger/iota/pull/9453/commits/2ba7b5564c16ebd15c21d9e35579767332f7f7bf)

crates/iota-protocol-config/src/lib.rs
miker83z: [2ba7b55](https://github.com/iotaledger/iota/pull/9453/commits/2ba7b5564c16ebd15c21d9e35579767332f7f7bf)

crates/iota-types/src/move_package.rs
miker83z: It is populated in the next PR #9461

valeriyr (approved):

muXxer (approved):

PR #9461, opened 2025-11-27
feat(execution): create a package metadata object at publish time

### LINKED ISSUE #8467: Interface considerations for programmable_transaction.move
In PR: https://github.com/iotaledger/iota/pull/8146
We had a discussion about the possibility of returning Options in most functions:
https://github.com/iotaledger/iota/pull/8146/files#r2293068271
We aren't sure if references may be returned in Options, though.

Similarly it could be an option to remove much of the interface, which was broached in this comment:
https://github.com/iotaledger/iota/pull/8146#pullrequestreview-3143599150

This ticket isn't concerned with describing the solution only logging the possibilities.
The above have been discussed to a certain extent, but more investigation is warranted as this will be a public interface.

The exact path isn't clear. For once many of these functions could be considered `view` functions, but we could also claim that users could write their own code as well, to handle all options.

miker83z:
# Description of change

This PR introduces the creation of a Package Metadata immutable object during a package publishing/upgrade. This allows to use such metadata to create `AuthenticatorInfoV1` instances for accounts.

[Internal AA architecture document](https://iotafoundation.atlassian.net/wiki/x/EoBOng).

Description:
- every time a function uses the `#[authenticator]` annotation, the node’s publish/upgrade execution tries to create a `PackageMetadata` immutable object.
- a new set of functions in the `iota-verifier` crate implement the checking of the requirements for a package metadata to be valid; the information for such validation is extracted from the package bytecode (see #9453)
- `iota-verifier::verify_authenticate_func_v1` checks the correctness of the declared authenticate() function (that was previously annotated with #[authenticator])
- once the PackageMetadata immutable object is created, then the framework method `create_auth_info_v1` can be used as the only way to create an `AuthenticatorInfoV1` instance.
- the `AuthenticatorInfoV1` struct includes the fields to uniquely identify a function on-chain: package id, module name, function name.
- An Account object is an object with a dynamic field being a `AuthenticatorInfoV1`.
- Then, to any user-created object, an `AuthenticatorInfoV1` can be “attached“. The only requirement is that the AuthenticatorInfoV1<T> is compatible with T, where T is exactly the type of the object set to become an Account.

### Release Notes

- [x] Protocol: Add the publish_package_metadata. This enables the creation of an immutable object at publish/upgrade time when some known attributes (e.g., the authenticator function attribute) are found in the source code.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-framework-snapshot/manifest.json
```diff
@@ -415,7 +415,7 @@
     ]
   },
   "16": {
-    "git_revision": "6bfe08416465aec56850c60be65c6b827b3c10ee",
```
alexsporn, line 418: This needs to be reverted, we already shipped a v16

miker83z: Moved to 17

thibault-martinez (approved):

kodemartin (approved):

bingyanglin (approved):

valeriyr (approved):

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -0,0 +1,171 @@
...
+    assert!(modules.length() == type_names.length());
+    let addr = iota::derived_object::derive_address_for_testing(
+        storage_id,
+        PackageMetadataKey {},
+    );
+    let id = object::new_uid_from_hash(addr);
+    let mut modules_metadata = iota::vec_map::empty<ascii::String, ModuleMetadataV1>();
+    let mut i = 0;
+    while (i < modules.length()) {
+        let module_name = modules[i];
```
Dkwcs, line 147: It's marked as "test_only", however, in this case, if we have at least two module_name entries, the second one won’t be added:

```
public fun insert<K: copy, V>(self: &mut VecMap<K, V>, key: K, value: V) {
    assert!(!self.contains(&key), EKeyAlreadyExists);
    self.contents.push_back(Entry { key, value })
}
```

So the loop with the condition `while (i < modules.length())` doesn't have much sense here, as only the first module gets inserted.

Example:
We have module1 with authenticator1, authenticator2 and it should be transformed into module1 - authenticator1, module1 - authenticator2

In case of `create_package_metadata_v1_for_testing_one_authenticator` it works, but if we want to test at least 2 authenticators in the same module it won't work.

Dkwcs (approved):

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z: Fixed here [d43cb09](https://github.com/iotaledger/iota/pull/9461/commits/d43cb099306cf7385db6bf02bed3f5afcb388930)

crates/iota-framework/packages/iota-framework/sources/programmable_transaction.move
```diff
@@ -0,0 +1,335 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+module iota::programmable_transaction;
+
+use std::ascii::String;
+use std::type_name::TypeName;
+
+// === Constants ===
+
+const EInvalidEnumVariant: u64 = 0;
```
alexsporn: Why not use the newer error syntax?

crates/iota-framework/packages/iota-framework/sources/programmable_transaction.move
```diff
@@ -0,0 +1,335 @@
...
+}
+
+public enum Command has copy, drop {
+    MoveCall(ProgrammableMoveCall),
+    TransferObjects(vector<Argument>, Argument),
+    SplitCoins(Argument, vector<Argument>),
+    MergeCoins(Argument, vector<Argument>),
+    Publish(vector<vector<u8>>, vector<ID>),
+    MakeMoveVec(Option<TypeName>, vector<Argument>),
+    Upgrade(vector<vector<u8>>, vector<ID>, ID, Argument),
```
alexsporn: Maybe add comments here to explain the fields

crates/iota-types/src/auth_context.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        if is_auth_context(module_addr, module_name, struct_name) {
+            kind
+        } else {
+            AuthContextKind::None
+        }
+    }
+}
+
+// TODO: add a deserializer that can handle the Command::MoveCall and
```
alexsporn, line 110: what about this TODO?

alexsporn (approved):
lgtm 🚀

crates/iota-framework/packages/iota-framework/sources/programmable_transaction.move
miker83z: Removed in [ac957d3](https://github.com/iotaledger/iota/pull/9461/commits/ac957d3fdfd71f505727b31847939219278a96a4). Now public functions return options.

crates/iota-framework/packages/iota-framework/sources/programmable_transaction.move
miker83z: In [ac957d3](https://github.com/iotaledger/iota/pull/9461/commits/ac957d3fdfd71f505727b31847939219278a96a4) these now have dedicated structs.

crates/iota-types/src/auth_context.rs
miker83z: Currently we have no need of a deserializer (we have no usage from Move->Rust but only from Rust->Move). In any case we plan dedicated structs for a new AuthContext version (see #9505) that will probably not need dedicated de/serializers. 
Do you think I should I remove the todo?

PR #9466, opened 2025-11-28
fix(move-transactional-test-runner): Make MaybeNamedCompiledModule keep the annotations

### LINKED ISSUE #9369: [move][tx-test-runner] Make MaybeNamedCompiledModule keep the annotations
Currently, the annotations written in the `*.move` files in transactional-tests are lost. This means that we can't use `#[authenticator]` for instance and, after #9296, all tests related to authenticators are failing because of this. 

The reason is that, when `MaybeNamedCompiledModule` is [created](https://github.com/iotaledger/iota/blob/610276483bf646b2cef5f5db45ed18fdc1f41863/external-crates/move/crates/move-transactional-test-runner/src/framework.rs#L592) in move-transactional-test-runner, the `AnnotatedCompiledModule`'s annotations are lost. These should be carried over. Consider carrying over the function infos as well.

Check #9188 iota-move-build for following the flow of the authenticator attribute from an AnnotatedCompiledModule to a FnInfo. Should be similar also for transactional-test, where `MaybeNamedCompiledModule` can carry this information. 
#9187 is also related.

Dkwcs:
# Description of change

This PR fixes the loss of function annotations in Move transactional tests and ensures that authenticator information such (#[authenticator] attributes) is correctly propagated into `CompiledModule` metadata though `MaybeNamedCompiledModule`.
`MaybeNamedCompiledModule` now carries `function_infos` in order to fill metadata during the package publishing

Iota-adapter-transactional abstract account tests have been fixed.

Added tests:
several_authenticators.move
without_attribute.move

Fixes #9369 .

Run `iota-adapter-transactional-tests` tests
With `cargo simtest`


crates/iota-types/Cargo.toml
```diff
@@ -60,6 +60,7 @@ tracing.workspace = true
 
 # internal dependencies
 consensus-config.workspace = true
+move-compiler.workspace = true
```
muXxer: Can we avoid this? `iota-types` is a dependency for so many crates, but `move-compiler` is not. And `move-compiler` itself also has a lot of dependencies itself. IMO `iota-types` should be kept lean, that's why we for example also introduce a crate `grpc-api-types` to not inherit all dependencies from `grpc-api-server`.

crates/iota-types/src/move_package.rs
```diff
@@ -98,6 +104,56 @@ pub struct FnInfoKey {
 /// A map from function info keys to function info
 pub type FnInfoMap = BTreeMap<FnInfoKey, FnInfo>;
 
+/// Create a function info map from function infos
+/// functions infos contain information about functions and their attributes in
+/// a module
```
muXxer:
```suggestion
/// Create a function info map from function infos.
/// Functions infos contain information about functions and their attributes in
/// a module.
```

crates/iota-types/src/move_package.rs
```diff
@@ -98,6 +104,56 @@ pub struct FnInfoKey {
...
+        if let Some(version) = get_authenticator_version_from_fun(fn_name, module, fn_info_map) {
+            runtime_metadata.add_function_attribute(
+                fn_name.to_string(),
+                IotaAttribute::authenticator_attribute(version),
+            );
+        }
+    }
+
+    if !runtime_metadata.is_empty() {
+        module.metadata.push(Metadata {
```
muXxer: `module.metadata` is a vector. We wouldn't really realize here that `IOTA_METADATA_KEY` might already exist in that `metadata`, right?. Would it make sense to add a function that loops over all entries in `metadata` first, before appending to the end? Or are we sure that can never ever happen? If this function here for example is called twice, we have two metadata with the same key, right?

muXxer (changes requested):

crates/iota-move-build/src/lib.rs
```diff
@@ -159,18 +155,7 @@ impl BuildConfig {
...
-                let is_test = mod_is_test || info.attributes.is_test_or_test_only();
-                let authenticator_version = info.attributes.get_authenticator();
-                fn_info_map.insert(
-                    FnInfoKey { fn_name, mod_addr },
-                    FnInfo {
-                        is_test,
-                        authenticator_version,
-                    },
-                );
-            }
```
miker83z, line 173: @Dkwcs I would keep these lines here and copy them into `iota-transactional-test-runner/src/test_adapter.rs`. We would have duplication, but I think it would be ok because the two publishing behaviors can change in the future; for instance, in tx-test-runner we don't need the test attribute right now. 

This would solve the comment https://github.com/iotaledger/iota/pull/9466/files#r2572910944

crates/iota-types/src/move_package.rs
miker83z: The deserialization of packages impedes to have more than 1 entry and that entry must be an IOTA key -> https://github.com/iotaledger/iota/pull/9453/files#diff-ac7b3c85b608636baa496a4eee63e52689d7d7a3d3aa29293470a8166a53c13eR590

We should normally reach this `fill_module_with_metadata` point after a deserialization. If not, and 2 entries are added, later the verifier will check again size and key, both before publishing and before any execution (involving that package): https://github.com/iotaledger/iota/pull/9453/files#diff-34b124a43c79dee4d9f50a90fbc2064ef598e00c3c153ca48adfeecf61a01b0bR32

crates/iota-move-build/src/lib.rs
Dkwcs: Done

crates/iota-types/Cargo.toml
Dkwcs: Done

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -2665,3 +2681,51 @@ impl ReadStore for IotaTestAdapter {
         self.executor.try_get_full_checkpoint_contents(digest)
     }
 }
+
+/// Create a function info map from the provided compiled modules
+fn fn_info_map(modules: &[MaybeNamedCompiledModule]) -> FnInfoMap {
+    let mut fn_info_map: FnInfoMap = BTreeMap::new();
+    for m in modules {
```
miker83z: Since these are now duplicated, you could merge the for cycles of `fn_info_map` and `fill_metadata`

crates/iota-types/src/move_package.rs
Dkwcs: Not relevant

miker83z (approved):
LGTM!

crates/iota-transactional-test-runner/src/test_adapter.rs
Dkwcs: Fixed 9e07c9a19753c9d0a07d263c4bb88c78bda9c227

muXxer (approved):

valeriyr (approved):

PR #9476, opened 2025-12-01
feat(iota-indexer): add HistoricalFallbackReader client impl

### LINKED ISSUE #9411: feat(indexer): Create HistoricalFallbackReader
Use ExtendedHttpKvStore client, and convert the results to native indexer format (Stored* types mostly).
To be used for fallback by IndexerReader.
In the future it is also meant to be used by graphql.

sergiupopescu199:
# Description of change

This PR adds a `HistoricalFallbackReader` client, as a high level interface to fetch more structured data from the historical fallback storage needed for the indexer JSRON RPC API.

fixes #9411 

> [!NOTE]
> This change does not affect the normal operation of the indexer, thus no tests were conducted.

### Release Notes
- [x] Indexer: introduce the `HistoricalFallbackReader` client, as a high level interface to fetch more structured data from the historical fallback storage needed for the indexer JSRON RPC API.

### DIFF
--- a/crates/iota-indexer/src/historical_fallback/reader.rs
+++ b/crates/iota-indexer/src/historical_fallback/reader.rs
@@ -1,0 +1,523 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Module containing the historical fallback reader implementation.
+//!
+//! This module provides a high-level client to interact with the historical
+//! fallback storage. It enables integration with the
+//! [`IndexerReader`](crate::read::IndexerReader) for fallback capabilities when
+//! the indexer is unable to fetch data from the database, which is especially
+//! useful when pruning is enabled.
+
+#![expect(dead_code)]
+
+use std::collections::HashMap;
+
+use futures::future;
+use iota_json_rpc_types::{CheckpointId, IotaEvent};
+use iota_rest_api::CheckpointTransaction;
+use iota_types::{
+    base_types::{ObjectID, SequenceNumber},
+    digests::TransactionDigest,
+    effects::{TransactionEffects, TransactionEffectsAPI},
+    event::EventID,
+    messages_checkpoint::{CertifiedCheckpointSummary, CheckpointSequenceNumber},
+    object::Object,
+};
+use itertools::{Itertools, izip};
+
+use crate::{
+    errors::{IndexerError, IndexerResult},
+    historical_fallback::{
+        client::{HttpRestKVClient, KeyValueStoreClient},
+        convert::{HistoricalFallbackEvents, HistoricalFallbackTransaction},
+    },
+    models::{
+        checkpoints::StoredCheckpoint, objects::StoredObject, transactions::StoredTransaction,
+    },
+    read::PackageResolver,
+};
+
+/// Represents the Input objects of a transaction.
+pub type InputObjects = Vec<Object>;
+/// Represents the Output objects of a transaction.
+pub type OutputObjects = Vec<Object>;
+
+/// A high-level client to interact with the historical fallback storage.
+///
+/// Provides convenient methods to fetch data from the historical fallback
+/// storage, with automatic conversions into types the Indexer uses when
+/// fetching from the database. This enables integration with the
+/// [`IndexerReader`](crate::read::IndexerReader) for fallback capabilities
+/// when the indexer is unable to fetch data from the database, which is
+/// especially useful when pruning is enabled.
+pub(crate) struct HistoricalFallbackReader {
+    /// Client responsible for fetching data from the historical fallback
+    /// storage through REST API interface.
+    client: HttpRestKVClient,
+    // TODO: The package resolver should support fetching an object by ID from fallback storage
+    // in case the objects table is pruned in the indexer.
+    //
+    // Current approach: We use the package resolver to fetch packages by ID from Postgres.
+    //
+    // Future requirements: Once the objects table is pruned, the package resolver will need
+    // to fetch objects by ID from the historical fallback reader. This depends on the range
+    // scan feature being implemented on the KV REST API side.
+    package_resolver: PackageResolver,
+}
+
+impl HistoricalFallbackReader {
+    pub fn new(rest_kv_url: &str, package_resolver: PackageResolver) -> IndexerResult<Self> {
+        let client = HttpRestKVClient::new(rest_kv_url)?;
+        Ok(Self {
+            client,
+            package_resolver,
+        })
+    }
+
+    /// Resolves the input and output objects from a given transaction effects.
+    pub(crate) async fn resolve_transaction_input_output_objects(
+        &self,
+        transaction_effects: &TransactionEffects,
+    ) -> IndexerResult<(InputObjects, OutputObjects)> {
+        let input_object_keys = transaction_effects.modified_at_versions();
+
+        let output_object_keys = transaction_effects
+            .all_changed_objects()
+            .into_iter()
+            .map(|((object_id, version, _object_digest), _owner, _kind)| (object_id, version))
+            .collect::<Vec<(ObjectID, SequenceNumber)>>();
+
+        let (raw_input_objects, raw_output_objects) = tokio::try_join!(
+            self.client.multi_get_objects(&input_object_keys),
+            self.client.multi_get_objects(&output_object_keys),
+        )?;
+
+        let input_objects = raw_input_objects
+            .into_iter()
+            .zip(&input_object_keys)
+            .map(|(object, (object_id, version))| {
+                object.ok_or_else(|| IndexerError::HistoricalFallbackObjectNotFound {
+                    object_id: *object_id,
+                    version: *version,
+                })
+            })
+            .collect::<IndexerResult<Vec<Object>>>()?;
+
+        let output_objects = raw_output_objects
+            .into_iter()
+            .zip(&output_object_keys)
+            .map(|(object, (object_id, version))| {
+                object.ok_or_else(|| IndexerError::HistoricalFallbackObjectNotFound {
+                    object_id: *object_id,
+                    version: *version,
+                })
+            })
+            .collect::<IndexerResult<Vec<Object>>>()?;
+
+        Ok((input_objects, output_objects))
+    }
+
+    /// Resolves the checkpoint summaries by the provided transaction digests
+    pub(crate) async fn resolve_checkpoint_summaries_by_tx_digests(
+        &self,
+        tx_digests: &[TransactionDigest],
+    ) -> IndexerResult<HashMap<TransactionDigest, CertifiedCheckpointSummary>> {
+        let tx_to_checkpoint_seq_num = self
+            .client
+            .multi_get_transactions_perpetual_checkpoints(tx_digests)
+            .await?;
+
+        // deduplicate checkpoint sequence numbers to avoid fetching the same summary
+        // multiple times.
+        let unique_seq_nums = tx_to_checkpoint_seq_num
+            .iter()
+            .flatten()
+            .unique()
+            .copied()
+            .collect::<Vec<CheckpointSequenceNumber>>();
+
+        let summaries_by_seq = self
+            .client
+            .multi_get_checkpoints_summaries_by_sequence_numbers(&unique_seq_nums)
+            .await?
+            .into_iter()
+            .flatten()
+            .map(|summary| (summary.sequence_number, summary))
+            .collect::<HashMap<CheckpointSequenceNumber, CertifiedCheckpointSummary>>();
+
+        // map each tx digest to its checkpoint summary
+        let summaries = tx_digests
+            .iter()
+            .zip(tx_to_checkpoint_seq_num)
+            .filter_map(|(digest, seq_num)| {
+                let seq = seq_num?;
+                summaries_by_seq
+                    .get(&seq)
+                    .cloned()
+                    .map(|summary| (*digest, summary))
+            })
+            .collect();
+
+        Ok(summaries)
+    }
+
+    /// Fetches a checkpoint by either a [`CheckpointSequenceNumber`] or
+    /// [`CheckpointDigest`](iota_types::digests::CheckpointDigest).
+    pub(crate) async fn checkpoint(
+        &self,
+        id: CheckpointId,
+    ) -> IndexerResult<Option<StoredCheckpoint>> {
+        let (summaries, contents) = match id {
+            CheckpointId::SequenceNumber(sequence_number) => {
+                let seq_nums = [sequence_number];
+                tokio::try_join!(
+                    self.client
+                        .multi_get_checkpoints_summaries_by_sequence_numbers(&seq_nums),
+                    self.client.multi_get_checkpoints_contents(&seq_nums)
+                )?
+            }
+            CheckpointId::Digest(digest) => {
+                let summaries = self
+                    .client
+                    .multi_get_checkpoints_summaries_by_digests(&[digest])
+                    .await?;
+
+                let Some(seq_num) = summaries
+                    .first()
+                    .and_then(|summary| summary.as_ref())
+                    .map(|summary| summary.sequence_number)
+                else {
+                    return Ok(None);
+                };
+
+                let contents = self
+                    .client
+                    .multi_get_checkpoints_contents(&[seq_num])
+                    .await?;
+
+                (summaries, contents)
+            }
+        };
+
+        let checkpoint = summaries
+            .into_iter()
+            .zip(contents)
+            .next()
+            .and_then(|(s, c)| Some(StoredCheckpoint::from((s?, c?))));
+
+        Ok(checkpoint)
+    }
+
+    /// Fetches multiple checkpoints from the historical fallback storage.
+    ///
+    /// Returns checkpoints in paginated form, supporting both ascending and
+    /// descending order.
+    ///
+    /// # Pagination Behavior
+    ///
+    /// | cursor | descending | Result |
+    /// |--------|------------|--------|
+    /// | `None` | `false` | Starts from checkpoint 0 |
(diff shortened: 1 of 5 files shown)


crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+/// storage, with automatic conversions into types the Indexer uses when
+/// fetching from the database. This enables integration with the
+/// [`IndexerReader`](crate::read::IndexerReader) for fallback capabilities
+/// when the indexer is unable to fetch data from the database, which is
+/// especially useful when pruning is enabled.
+pub(crate) struct HistoricalFallbackReader {
+    /// Client responsible for fetching data from the historical fallback
+    /// storage through REST API interface.
+    client: HttpRestKVClient,
+    // TODO: The package resolver should support fetching an object by ID from fallback storage
```
tomxey: I think for now we can assume that `objects` will not be pruned.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+                    .first()
+                    .and_then(|summary| summary.as_ref())
+                    .map(|summary| summary.sequence_number)
+                else {
+                    return Ok(None);
+                };
+
+                let contents = self
+                    .client
+                    .multi_get_checkpoints_contents(&[seq_num])
```
tomxey, line 198: Maybe call to "multi_get_checkpoints_contents" could already be common for both "SequenceNumber" and "Digest" path? Just an idea, to be considered. Would reduce some code duplication.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+    ///
+    /// | cursor | descending | Result |
+    /// |--------|------------|--------|
+    /// | `None` | `false` | Starts from checkpoint 0 |
+    /// | `None` | `true` | Starts from latest checkpoint |
+    /// | `Some(n)` | `false` | Starts from checkpoint n+1 |
+    /// | `Some(n)` | `true` | Starts from checkpoint n-1 |
+    ///
+    /// # NOTE
+    /// `StoredCheckpoint.sequence_number` is hardcoded to 0, due to missing
```
tomxey: We are fetching checkpoints by sequence numbers here, so we know it already. Could we fill it then?

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+        )?;
+
+        let Some(historical_event) = events.into_iter().next().and_then(|events| {
+            let summary = checkpoint_summaries.get(&tx_digest).cloned()?;
+            Some(HistoricalFallbackEvents::new(events?, summary))
+        }) else {
+            return Ok(vec![]);
+        };
+
+        historical_event
```
tomxey: nit: `historical_events` would me more appropriate I think, since it's not just one

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+    /// Fetches transactions belonging to a specific checkpoint.
+    ///
+    /// Returns transactions in paginated form, supporting both ascending and
+    /// descending order within the checkpoint.
+    ///
+    /// # Pagination Behavior
+    ///
+    /// | cursor | descending | Result |
+    /// |--------|------------|--------|
+    /// | `None` | `false` | Starts from first transaction in checkpoint |
```
tomxey: nit: could this table also be formatted nicely in code? (the same goes for other tables in this file)

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+            .flatten()
+        else {
+            return Ok(vec![]);
+        };
+
+        let tx_digests: Vec<TransactionDigest> = contents.iter().map(|b| b.transaction).collect();
+
+        // apply ordering
+        let tx_digests: Vec<TransactionDigest> = if is_descending {
+            tx_digests.into_iter().rev().collect()
```
tomxey: I'm not sure if rust type system would allow this, but could we avoid the collect here and keep the type as iterator?

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+        // apply cursor: skip transactions until after the cursor.
+        //
+        // This relies on transactions being ordered within checkpoint contents,
+        // so we can skip until we find the cursor, then skip the cursor itself.
+        let tx_digests: Vec<TransactionDigest> = if let Some(cursor) = cursor {
+            tx_digests
+                .into_iter()
+                .skip_while(|digest| *digest != cursor)
+                .skip(1) // skip the cursor itself
+                .collect()
```
tomxey: as above

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+        } else {
+            tx_digests
+        };
+
+        // apply limit
+        let tx_digests = tx_digests
+            .into_iter()
+            .take(limit)
+            .collect::<Vec<TransactionDigest>>();
```
tomxey, line 442: could it be the only collect in this function?

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+        )?;
+
+        let futures =
+            izip!(transactions, effects, events).map(|(transaction, effects, events)| async {
+                let (Some(transaction), Some(effects)) = (transaction, effects) else {
+                    return Ok(None);
+                };
+
+                let Some(summary) = checkpoint_summaries.get(transaction.digest()).cloned() else {
+                    return Ok(None);
```
tomxey: If we return error for missing in/out objs, I think we could return error also for missing cp data, since it means that KV data is inconsistent. Or do you have a different view on this?

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,523 @@
...
+            .into_iota_events(self.package_resolver.clone(), tx_digest)
+            .await
+    }
+
+    /// Fetches transactions from the provided transaction digests.
+    ///
+    /// # NOTE
+    /// The `StoredTransaction.tx_sequence_number` is hardcoded to 0, due to
+    /// missing data in the historical fallback. It can be derived but the
+    /// operations could be expensive. Can be added in future iterations.
```
tomxey: Would it be enough to fetch cp contents? Or such operation already is too heavy?

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: I've provided comments for each branch in the implementation itself. The doc table serves as a high-level overview for API consumers. I'm not sure duplicating it in the code would add benefit—what specific improvement did you have in mind?

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: That's a good observation, I didn't want to change the trait method for now, but I suppose we can rethink it once we finish the first iteration of the fallback feature. Since we'll need to iterate over it multiple times it could be a good opportunity to improve the trait, client & reader in one go.

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: it should've been `StoredCheckpoint.successful_tx_num`, this value is not used when converting into JSON RPC Checkpoint, to compute this value we should for every transaction fetch its effects and count all successful transactions so it adds a bit of overhead but since is not used we just skip it.

fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: Good observation, it makes perfect sense.

fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: I also included the contents to calculate the tx sequence number just in case. the operation is not heavy I will say but we'll see in the future.

fixed: https://github.com/iotaledger/iota/pull/9476/commits/4154485f7b251f64ed96bbfcfcafd89e325c4d30

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// could be expensive. Can be added in future iterations.
+    pub(crate) async fn checkpoints(
+        &self,
+        // TODO: The `latest_checkpoint` parameter is a temporary workaround. The KV store
+        // doesn't currently have a way to determine the latest available checkpoint.
+        // This can be removed once either:
+        // - Range scan is implemented on the KV REST API, or
+        // - The `get_latest_checkpoint()` method is re-added (removed during upstream pull)
+        latest_available_checkpoint: CheckpointSequenceNumber,
+        cursor: Option<CheckpointSequenceNumber>,
```
kodemartin: I think that you can get the same functionality by requiring that `cursor: CheckpointSequenceNumber`. So `latest_available_checkpoint` might not be necessary here.

If the caller has access to `latest_available_checkpoint` there is still the flexibility to define the page by setting properly the `limit`.

This would be simpler.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// |--------|------------|--------|
+    /// | `None` | `false` | Starts from first transaction in checkpoint |
+    /// | `None` | `true` | Starts from last transaction in checkpoint |
+    /// | `Some(tx)` | `false` | Starts after `tx`, ascending |
+    /// | `Some(tx)` | `true` | Starts after `tx`, descending |
+    pub(crate) async fn checkpoint_transactions(
+        &self,
+        cursor: Option<TransactionDigest>,
+        checkpoint_sequence_number: CheckpointSequenceNumber,
+        limit: usize,
```
kodemartin, line 397: This can be an `Option<usize>` so as to allow callers to fetch all transactions at once.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// Events are indexed by their position in the transaction (event_seq = 0,
+    /// 1, 2, ...).
+    ///
+    /// | cursor | descending | Result |
+    /// |--------|------------|--------|
+    /// | `None` | `false` | Starts from event_seq 0 |
+    /// | `None` | `true` | Starts from last event |
+    /// | `Some(seq)` | `false` | Starts after event_seq |
+    /// | `Some(seq)` | `true` | Starts before event_seq |
+    pub(crate) async fn transaction_events(
```
kodemartin: What is the purpose of providing pagination? You seem to fetch all events of the transaction on each call, which is hardly optimal. 

This question can be generalized for other paginated requests here. Pages are probably best handled on the application after we fetch everything from BigTable in the least round trips possible and the cache the result in memory. 

I am not suggesting that this implemented now. But if pagination is not absolutely necessary for events I say we skip it.

kodemartin: Considering the name of the method, its distinction from `events` is not clear.

crates/iota-indexer/src/historical_fallback/reader.rs
tomxey: I meant sth like: (so that the table is aligned, content is the same)

```
// | cursor     | descending | Result                                      |
// |------------|------------|---------------------------------------------|
// | `None`     | `false`    | Starts from first transaction in checkpoint |
// | `None`     | `true`     | Starts from last transaction in checkpoint  |
// | `Some(tx)` | `false`    | Starts after `tx`, ascending                |
// | `Some(tx)` | `true`     | Starts after `tx`, descending               |
```

By "in code" I meant here (as opposed to how it's rendered after rust doc is built).

crates/iota-indexer/src/historical_fallback/reader.rs
tomxey: I didn't meant to change the trait, the scope of comment was only this function.

on high level, current state:

```rust
        let (summaries, contents) = match id {
            CheckpointId::SequenceNumber(sequence_number) => {
                tokio::try_join!(
                    self.client
                        .multi_get_checkpoints_summaries_by_sequence_numbers(&seq_nums),
                    self.client.multi_get_checkpoints_contents(&seq_nums)
                )?
            }
            CheckpointId::Digest(digest) => {
                self
                    .client
                    .multi_get_checkpoints_summaries_by_digests(&[digest])
                    .await?;
                self
                    .client
                    .multi_get_checkpoints_contents(&[seq_num])
                    .await?;
            }
        };
 ```


suggested state:

```rust
        let summaries = match id {
            CheckpointId::SequenceNumber(sequence_number) => {
                self.client
                    .multi_get_checkpoints_summaries_by_sequence_numbers(&seq_nums),
            }
            CheckpointId::Digest(digest) => {
                self
                    .client
                    .multi_get_checkpoints_summaries_by_digests(&[digest])
                    .await?;
            }
        };
        let contents = self
            .client
            .multi_get_checkpoints_contents(&[seq_num])
            .await?;
```

This is still just a suggestion, to be considered, if you don't like the idea you can drop the comment.

crates/iota-indexer/src/historical_fallback/reader.rs
tomxey: The API of the HistoricFallbackReader was tailored to the needs of fallback in IndexerReader, to handle fallback for specific endpoints. The goal was not to support general case.
This particular method is used as fallback for `query_transaction_blocks` ([PR](https://github.com/iotaledger/iota/pull/9493) with usage, for reference).
As `limit` is a required argument there, having it as option would not give us any benefits (without further refactors and changing the API behaviour).

This is just to provide reasoning for the current design. (the same reasoning would apply to https://github.com/iotaledger/iota/pull/9476#discussion_r2584316150 I think)

kodemartin: Ok fair enough. We can refine later on caching solutions.

crates/iota-indexer/src/historical_fallback/reader.rs
kodemartin: We can refine later on caching solutions.

kodemartin (approved):
lgtm

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: The `SequenceNumber` path currently fetches summaries and contents concurrently using `try_join!`, which would be lost if we extracted the common call, I would like to keep this meaningful optimization. The duplication is minimal (one method call), so I'd prefer keeping the performance benefit.

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/bc8baeaad6a05addb1651126b75b15e5cc540b08

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/bc8baeaad6a05addb1651126b75b15e5cc540b08

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9476/commits/bc8baeaad6a05addb1651126b75b15e5cc540b08

tomxey (approved):

PR #9481, opened 2025-12-01
fix(CLI+REST): fix `estimate_gas_budget_from_gas_cost` calculation

muXxer:
# Description of change

The function `estimate_gas_budget_from_gas_cost` can be found twice in the codebase. I moved the implementation to `iota-types` in the first commit, and changed all call sites. 

The second commit fixes the budget estimation with a backport from [upstream](https://github.com/MystenLabs/sui/pull/24354). I had to find equivalent parameters in the call sites, because upstream forgot to fix both implementations, so I couldn't port the change 1:1. 

In the `Display` function of `DryRunTransactionBlockResponse` there was no access to the `ProtocolConfig`, but I used `ProtocolConfig::get_for_max_version_UNSAFE()`, which is a good enough replacement in this case I guess.

I also had to change the call to dry_run, to use the ephemeral gas coin in the CLI and in the REST API by not providing a gas object.

# Description from upstream:

The budget estimation logic in the simulate gRPC API was underestimating the required gas budget, causing transactions to fail with INSUFFICIENT_GAS errors.

## The Bug
When estimating the gas budget for a transaction:

1. If the request provided payment coins, we used those coins on the estimation simulation AND used their balance as the budget. This caused the estimation to fail with
`InsufficientCoinBalance` because the entire balance was reserved for gas, leaving nothing available for the actual transaction operations.

2. We didn't check the status of the budget estimation effects - we just grabbed the used gas and computed our estimate. This meant the estimate was wrong because it didn't include the
parts of the PTB that didn't execute after the failure.

In mainnet with lower gas prices, this caused estimates that were too low, resulting in `INSUFFICIENT_GAS` failures.

## The Fix
1. Fixed estimation simulation setup:

- Set payment coins to vec![] for the budget simulation (to defer to the ephemeral gas coin)
- Set budget to max_tx_gas() instead of coin balance
- Check the status of the budget estimation transaction to ensure it succeeded before using its gas costs

2. Account for gas loading costs: Since we now run estimation with `payment = vec![]`, we need to manually calculate the cost to load gas payment objects

- Cost = `(num_payment_objects - 1) * 40 bytes * obj_access_cost_read_per_byte * reference_gas_price`
- Subtract 1 because the simulation already loaded one ephemeral gas coin
- Default to 255 payment objects if none specified (protocol maximum)
- Verified empirically that a `GasCoin` is exactly 40 bytes in BCS encoding

3. Improved estimation calculation:

- Base cost from simulation (computation + storage)
- Add gas loading cost
- Round up to gas rounding step (1000 MIST)
- Add safe overhead buffer (1000 * reference_gas_price)
- Clamp to max_tx_gas

4. Increased ephemeral gas coin: Bumped `DEV_INSPECT_GAS_COIN_VALUE` from 1M SUI to 1B SUI to prevent balance issues in simulations

## Test plan
Reproduced the bug locally on a mainnet fullnode using the exact failing transaction from the bug report. Confirmed the fix:

Before: `"success": false, "error": "INSUFFICIENT_GAS",` budget estimate: 1,104,880 MIST
After: `"success": true,` budget estimate: 3,071,000 MIST (correctly accounts for all costs)

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Fixed budget estimation bug that caused transactions to fail with "INSUFFICIENT_GAS" error by properly simulating budget estimation and accounting for gas payment object
loading costs
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-types/src/gas.rs
+++ b/crates/iota-types/src/gas.rs
@@ -5,6 +5,29 @@
 
 pub use checked::*;
 
+// An amount of gas (in gas units) that is added to transactions as an overhead
+// to ensure transactions do not fail.
+pub const GAS_SAFE_OVERHEAD: u64 = 1000;
+
+/// Estimate the gas budget using the gas_cost_summary from a previous DryRun
+///
+/// The estimated gas budget is computed as following:
+/// * the maximum between A and B, where: A = computation cost +
+///   GAS_SAFE_OVERHEAD * reference gas price B = computation cost + storage
+///   cost - storage rebate + GAS_SAFE_OVERHEAD * reference gas price overhead
+///
+/// This gas estimate is computed similarly as in the TypeScript SDK
+pub fn estimate_gas_budget_from_gas_cost(
+    gas_cost_summary: &GasCostSummary,
+    reference_gas_price: u64,
+) -> u64 {
+    let safe_overhead = GAS_SAFE_OVERHEAD * reference_gas_price;
+    let computation_cost_with_overhead = gas_cost_summary.computation_cost + safe_overhead;
+
+    let gas_usage = gas_cost_summary.net_gas_usage() + safe_overhead as i64;
+    computation_cost_with_overhead.max(if gas_usage < 0 { 0 } else { gas_usage as u64 })
+}
+
 #[iota_macros::with_checked_arithmetic]
 pub mod checked {
 
--- a/crates/iota-types/src/gas.rs
+++ b/crates/iota-types/src/gas.rs
@@ -9,23 +9,69 @@ pub use checked::*;
 // to ensure transactions do not fail.
 pub const GAS_SAFE_OVERHEAD: u64 = 1000;
 
-/// Estimate the gas budget using the gas_cost_summary from a previous DryRun
-///
-/// The estimated gas budget is computed as following:
-/// * the maximum between A and B, where: A = computation cost +
-///   GAS_SAFE_OVERHEAD * reference gas price B = computation cost + storage
-///   cost - storage rebate + GAS_SAFE_OVERHEAD * reference gas price overhead
-///
-/// This gas estimate is computed similarly as in the TypeScript SDK
-pub fn estimate_gas_budget_from_gas_cost(
-    gas_cost_summary: &GasCostSummary,
-    reference_gas_price: u64,
-) -> u64 {
-    let safe_overhead = GAS_SAFE_OVERHEAD * reference_gas_price;
-    let computation_cost_with_overhead = gas_cost_summary.computation_cost + safe_overhead;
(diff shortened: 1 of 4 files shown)


crates/iota-types/src/gas.rs
```diff
@@ -5,6 +5,78 @@
...
+///
+/// The estimation includes:
+/// 1. Base cost from gas_cost_summary (computation + storage costs)
+/// 2. Cost of loading gas payment objects (which weren't loaded during
+///    simulation)
+/// 3. Rounding up to the protocol gas rounding step (typically 1000 NANOS)
+/// 4. Adding safe overhead buffer (1000 * reference_gas_price)
+/// 5. Clamping to max_tx_gas protocol limit
+pub fn estimate_gas_budget_from_gas_cost(
+    gas_cost_summary: &crate::gas::GasCostSummary,
```
bingyanglin:
```suggestion
    gas_cost_summary: &GasCostSummary,
```

crates/iota-types/src/gas.rs
```diff
@@ -5,6 +5,78 @@
...
+    reference_gas_price: u64,
+    num_payment_objects_on_request: usize,
+    protocol_config: &iota_protocol_config::ProtocolConfig,
+) -> u64 {
+    const GAS_SAFE_OVERHEAD: u64 = 1000;
```
bingyanglin: Defined twice?

crates/iota-types/src/gas.rs
```diff
@@ -5,6 +5,78 @@
...
+        .saturating_mul(GAS_COIN_SIZE_BYTES)
+        .saturating_mul(protocol_config.obj_access_cost_read_per_byte());
+
+    // Round up to the nearest gas rounding step (in gas units)
+    let rounded_gas_loading_cost_units =
+        if let Some(step) = protocol_config.gas_rounding_step_as_option() {
+            match gas_loading_cost_units.checked_next_multiple_of(step) {
+                Some(rounded) => rounded,
+                None => u64::MAX,
+            }
```
bingyanglin, line 58: In the upstream there is a function `round_up_to_nearest` and used here, not sure why we don't have them.

muXxer: I don't want to define a function for something that exists in the standard libs.
`checked_next_multiple_of` does in principle what they try to do IMO.

bingyanglin: Yes and ours is better, thanks for the explanation.

bingyanglin (approved):

muXxer:
Wrong approach, see #9528

PR #9485, opened 2025-12-02
refactor: create dedicated native functions for the AA object creation

### LINKED ISSUE #9459: [AA] Create dedicated native functions for the AA object creation
Instead of having a AuthenticatorInfoV1CompatibilityProof, all operations related to the AA object creation (and rotation) should be handled by native function(s)

valeriyr:
[run-ci]

# Description of change

The account creation Move API was simplified by removing `AuthenticatorInfoV1CompatibilityProof`.

fixes #9459


external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
```diff
@@ -1050,7 +1054,7 @@ fn check_private_transfer(context: &mut Context, loc: Loc, mcall: &ModuleCall) {
     let current_module = context.current_module();
     if current_module
         .value
-        .is(&IOTA_ADDR_VALUE, TRANSFER_FUNCTION_NAME)
+        .is(&IOTA_ADDR_VALUE, TRANSFER_MODULE_NAME)
```
miker83z, line 1057: was this a "bug"?

external-crates/move/crates/move-symbol-pool/src/lib.rs
```diff
@@ -96,6 +96,10 @@ static_symbols!(
     "migration",
     "beta",
     "development",
+    "account",
```
miker83z, line 99: We should think about the usage of simply "account" for a module name, might be limiting in the future.

miker83z (approved):
LGTM!

external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
valeriyr: It was a typo, here the module name is checked, it worked because:
`TRANSFER_FUNCTION_NAME` == `TRANSFER_MODULE_NAME` == `transfer`

iota-execution/latest/iota-verifier/src/private_generics.rs
```diff
@@ -39,6 +40,17 @@ pub const TRANSFER_IMPL_FUNCTIONS: &[&IdentStr] = &[
     ident_str!("receive_impl"),
 ];
 
+pub const PUBLIC_ACCOUNT_FUNCTIONS: &[&IdentStr] = &[
+    ident_str!("create_auth_info_v1"),
+    ident_str!("borrow_auth_info_v1"),
+    ident_str!("has_auth_info_v1"),
+];
+pub const PRIVATE_ACCOUNT_FUNCTIONS: &[&IdentStr] = &[
```
Dkwcs, line 48: As for consideration we can rename private to internal to not confuse these private functions with visibility modifiers at least.

iota-execution/latest/iota-verifier/src/private_generics.rs
```diff
@@ -140,6 +154,46 @@ fn verify_private_transfer(
     Ok(())
 }
 
+fn verify_private_account(
```
Dkwcs: An adition to my previous comment above. 
Is this function verifies some "private" account or "private" functions that account contains?
I would suggest to add a comment that explains what actually happens here despite of possible renaming.

It's about all similar methods.

Dkwcs (approved):
LGTM, left comments for consideration

iota-execution/latest/iota-verifier/src/private_generics.rs
valeriyr: The term 'private' is used here to describe the functions that have specific verifier rules. The same name is used for all the other such functions, even the file is called `private_generics.rs`. So, for consistency it makes sense to leave it as is.
If we want to rename this array or provide a new term, we need to do this in a dedicated PR.

iota-execution/latest/iota-verifier/src/private_generics.rs
valeriyr: `account` in this function name defines a module name. This can be checked in a place where it is called.
It is consistent with current codebase. About renaming I left a comment [here](https://github.com/iotaledger/iota/pull/9485#discussion_r2597583554).

valeriyr: The function names I changed here: 97c57fa748

miker83z (approved):

Dkwcs (approved):

PR #9499, opened 2025-12-03
fix(iota-core): Fix differences between StarfishConsensusHandler and MysticetiConsensusHandler

### LINKED ISSUE #9484: Check that Starfish and Mysticeti adapter/handler have the same functionality
Make sure that iota-core components that are creating and handling consensus instances handle them in the same manner. Both consensus have the same interfaces so they should be handled in the same way.

piotrm50:
# Description of change

Fix differences between Starfish and Mysticeti in handling of previously processed commits at startup by introducing a replay mechanism. Adjust `StarfishConsensusHandler` to manage commit ranges effectively and log replay progress.

Fixes #9484 

### DIFF
--- a/crates/iota-core/src/unit_tests/mysticeti_manager_tests.rs
+++ b/crates/iota-core/src/unit_tests/mysticeti_manager_tests.rs
@@ -4,20 +4,26 @@
 
 use std::{sync::Arc, time::Duration};
 
+use arc_swap::ArcSwap;
+use consensus_core::{CommitDigest, CommitRef, CommittedSubDag, TestBlock, VerifiedBlock};
 use fastcrypto::traits::KeyPair;
 use futures::FutureExt;
-use iota_metrics::RegistryService;
+use iota_metrics::{RegistryService, monitored_mpsc::unbounded_channel};
 use iota_swarm_config::network_config_builder::ConfigBuilder;
-use iota_types::messages_checkpoint::{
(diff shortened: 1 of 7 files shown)


crates/iota-core/src/unit_tests/mysticeti_manager_tests.rs
```diff
@@ -137,3 +143,91 @@ async fn test_mysticeti_manager() {
...
+
+    // THEN verify that the highest handled commit is only updated for newer commits
+    // (8-10) Commits 1-7 should have been processed as prior commits and
+    // not update the monitor
+    let highest_handled = commit_consumer_monitor.highest_handled_commit();
+    assert_eq!(
+        highest_handled, 10,
+        "Expected highest handled commit to be 10, got {}",
+        highest_handled
+    );
```
muXxer, line 232: I'm sorry, but I don't understand that part, neither in mysticeti nor in starfish.
All 10 commits are equally created, so where is the difference at all? Only because we set the `last_processed_commit_at_startup` to 7?
Is this test really checking what we want? At least the last assert comment is also weird, because the assert only checks for highest handled = 10.
Do we even check they are correctly replayed?

piotrm50: I implemented those tests with wrong assumptions in my mind. They are in reality not needed, so I removed them.

muXxer (approved):

polinikita (approved):

PR #9526, opened 2025-12-04
fix(iota-types): use digest for ConsensusTransactionKey of MisbehaviorReport

cyberphysic4l:
# Description of change

The `ConsensusTransactionKey` for `MisbehaviorReport` uses authority name and the checkpoint sequence number, but this may cause ambiguity if an authority submits more than one report in one checkpoint. It is better to use the digest of the misbehavior report. 

This PR adds a digest to the misbehavior reports and uses this for the consensus transactions key.

### DIFF
--- a/crates/iota-types/src/messages_consensus.rs
+++ b/crates/iota-types/src/messages_consensus.rs
@@ -288,7 +288,8 @@ impl ConsensusTransactionKind {
 pub enum VersionedMisbehaviorReport {
     V1(
         MisbehaviorsV1<Vec<u64>>,
-        #[serde(skip)] OnceCell<MisbehaviorReportDigest>,
+        #[serde(skip)]
+        OnceCell<MisbehaviorReportDigest>,
     ),
 }
 
--- a/crates/iota-types/src/messages_consensus.rs
+++ b/crates/iota-types/src/messages_consensus.rs
@@ -15,6 +15,7 @@ use fastcrypto::{error::FastCryptoResult, groups::bls12381, hash::HashFunction};
 use fastcrypto_tbls::dkg_v1;
 use fastcrypto_zkp::bn254::zk_login::{JWK, JwkId};
 use iota_sdk_types::crypto::IntentScope;
+use once_cell::sync::OnceCell;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -22,8 +23,8 @@ use crate::{
     base_types::{
         AuthorityName, ConciseableName, ObjectID, ObjectRef, SequenceNumber, TransactionDigest,
     },
-    crypto::{AuthoritySignature, DefaultHash},
-    digests::{ConsensusCommitDigest, Digest},
+    crypto::{AuthoritySignature, DefaultHash, default_hash},
+    digests::{ConsensusCommitDigest, Digest, MisbehaviorReportDigest},
     message_envelope::{Envelope, Message, VerifiedEnvelope},
     messages_checkpoint::{CheckpointSequenceNumber, CheckpointSignatureMessage},
     supported_protocol_versions::{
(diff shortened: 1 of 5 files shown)


crates/iota-types/src/messages_consensus.rs
```diff
@@ -293,20 +286,35 @@ impl ConsensusTransactionKind {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum VersionedMisbehaviorReport {
-    V1(MisbehaviorsV1<Vec<u64>>),
+    V1(
+        MisbehaviorsV1<Vec<u64>>,
+        #[serde(skip)] OnceCell<MisbehaviorReportDigest>,
```
muXxer, line 291: Does the `#[serde(skip)]` need to be in the same line? Looks weird :D

muXxer (approved):
LGTM, please rebase and merge :)

crates/iota-types/src/messages_consensus.rs
cyberphysic4l: My linter keeps changing it to this when I save but I don't see other examples, so I will change it to be on the line above instead.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -293,20 +286,36 @@ impl ConsensusTransactionKind {
...
-    /// Returns an iterator over references to some of the fields in the report.
+    // Returns an iterator over references to some of the fields in the report.
     pub fn iterate_over_metrics(&self) -> std::vec::IntoIter<&Vec<u64>> {
         match self {
-            VersionedMisbehaviorReport::V1(report) => report.iter(),
+            VersionedMisbehaviorReport::V1(report, _) => report.iter(),
+        }
+    }
+    // Returns the digest of the misbehavior report, caching it if it has not
+    // been computed yet.
```
roman1e2f5p8s:
```suggestion

    /// Returns the digest of the misbehavior report, caching it if it has not
    /// been computed yet.
```

crates/iota-types/src/messages_consensus.rs
```diff
@@ -293,20 +286,36 @@ impl ConsensusTransactionKind {
...
+
     pub fn verify(&self, committee_size: usize) -> bool {
         match self {
-            VersionedMisbehaviorReport::V1(report) => report.verify(committee_size),
+            VersionedMisbehaviorReport::V1(report, _) => report.verify(committee_size),
         }
     }
-
-    /// Returns an iterator over references to some of the fields in the report.
+    // Returns an iterator over references to some of the fields in the report.
```
roman1e2f5p8s:
```suggestion

    /// Returns an iterator over references to some of the fields in the report.
```

crates/iota-types/src/messages_consensus.rs
```diff
@@ -293,20 +286,36 @@ impl ConsensusTransactionKind {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum VersionedMisbehaviorReport {
-    V1(MisbehaviorsV1<Vec<u64>>),
+    V1(
+        MisbehaviorsV1<Vec<u64>>,
+        #[serde(skip)]
+        OnceCell<MisbehaviorReportDigest>,
```
roman1e2f5p8s:
```suggestion
        #[serde(skip)]
        // to prevent cargo fmt putting the skip attribute in the same line
        OnceCell<MisbehaviorReportDigest>,
```

crates/iota-types/src/messages_consensus.rs
roman1e2f5p8s: You might want to turn other comments over structs like `MisbehaviorsV1` and functions into doc strings.

roman1e2f5p8s (approved):
Left minor formatting and styling suggestions.

oliviasaa (approved):

PR #9530, opened 2025-12-05
feat(iota-indexer): add historical fallback support in Read Api

### LINKED ISSUE #9473: feat(indexer): Use HistoricalFallbackReader for ReadApi

sergiupopescu199:
# Description of change
This PR adds provides historical fallback support for the following Read Api methods:

- `get_events`
- `get_checkpoint`
- `get_checkpoints`
- `get_transaction_block`
- `multi_get_transaction_blocks`
- `try_get_past_object`
- `try_get_object_before_version`
- `try_multi_get_past_objects`

fixes #9473 

> [!NOTE]
> This patch only provides the initial functionality, thus no thorough test were made since we have a dedicated issue.

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -21,6 +21,7 @@ use diesel::{
     sql_types::{BigInt, Bool},
 };
 use fastcrypto::encoding::{Encoding, Hex};
+use futures::FutureExt;
 use iota_json_rpc_types::{
     AddressMetrics, Balance, CheckpointId, Coin as IotaCoin, DisplayFieldsResponse, EpochInfo,
     EventFilter, IotaCoinMetadata, IotaEvent, IotaMoveValue, IotaObjectDataFilter,
@@ -274,50 +275,28 @@ impl IndexerReader {
     /// Searches the requested object version and checkpoint sequence number
     /// in `objects_version` and fetches the requested object from
     /// `objects_history`.
+    ///
+    /// Returns [`IndexerError::DataPruned`] if the object version exists but
+    /// history was pruned
     pub(crate) async fn get_past_object_read(
         &self,
         object_id: ObjectID,
         object_version: SequenceNumber,
         before_version: bool,
-    ) -> Result<PastObjectRead, IndexerError> {
+    ) -> IndexerResult<PastObjectRead> {
         let object_version_num = object_version.value() as i64;
 
         // Query objects_version to find the requested version and relevant
         // checkpoint sequence number considering the `before_version` flag.
-        let pool = self.get_pool();
-        let object_id_bytes = object_id.to_vec();
-        let object_version_info: Option<StoredObjectVersion> =
-            run_query_async!(&pool, move |conn| {
-                let mut query = objects_version::dsl::objects_version
-                    .filter(objects_version::object_id.eq(&object_id_bytes))
-                    .into_boxed();
-
-                if before_version {
-                    query = query.filter(objects_version::object_version.lt(object_version_num));
-                } else {
-                    query = query.filter(objects_version::object_version.eq(object_version_num));
-                }
-
-                query
-                    .order_by(objects_version::object_version.desc())
-                    .limit(1)
-                    .first::<StoredObjectVersion>(conn)
-                    .optional()
-            })?;
+        let object_version_info = self
+            .db()
+            .get_object_version(object_id, object_version, before_version)
+            .await?;
 
         let Some(object_version_info) = object_version_info else {
             // Check if the object ever existed.
-            let pool = self.get_pool();
-            let object_id_bytes = object_id.to_vec();
-            let latest_existing_version: Option<i64> = run_query_async!(&pool, move |conn| {
-                objects_version::dsl::objects_version
-                    .filter(objects_version::object_id.eq(&object_id_bytes))
-                    .order_by(objects_version::object_version.desc())
-                    .select(objects_version::object_version)
-                    .limit(1)
-                    .first::<i64>(conn)
-                    .optional()
-            })?;
+            let latest_existing_version =
+                self.db().latest_existing_object_version(object_id).await?;
 
             return match latest_existing_version {
                 Some(latest) if object_version_num > latest => Ok(PastObjectRead::VersionTooHigh {
@@ -326,45 +305,13 @@ impl IndexerReader {
                     latest_version: SequenceNumber::from(latest as u64),
                 }),
                 Some(_) => Ok(PastObjectRead::VersionNotFound(object_id, object_version)),
-                None => {
-                    // fallback to historical storage
-                    let Some(fallback) = self.fallback_reader() else {
-                        return Ok(PastObjectRead::ObjectNotExists(object_id));
-                    };
-
-                    let Some(obj) = fallback
-                        .objects(&[(object_id, object_version)], before_version)
-                        .await?
-                        .pop()
-                        .flatten()
-                    else {
-                        return Ok(PastObjectRead::VersionNotFound(object_id, object_version));
-                    };
-
-                    // Note: We use `StoredObject.try_into_object_read` here instead of
-                    // `StoredHistoryObject.try_into_past_object_read` because the fallback
-                    // storage returns `StoredObject`. Both methods share the same logic for
-                    // resolving the MoveStructLayout via `package_resolver.type_layout()`.
-                    // The key difference is that `try_into_past_object_read` also handles
-                    // the `WrappedOrDeleted` status, which for this iteration, we handle explicitly
-                    // as a `VersionNotFound`.
-                    match obj
-                        .try_into_object_read(self.package_resolver.clone())
-                        .await?
-                    {
-                        ObjectRead::NotExists(_) | ObjectRead::Deleted(_) => {
-                            Ok(PastObjectRead::VersionNotFound(object_id, object_version))
-                        }
-                        ObjectRead::Exists(obj_ref, object, layout) => {
-                            Ok(PastObjectRead::VersionFound(obj_ref, object, layout))
-                        }
-                    }
-                }
+                None => Ok(PastObjectRead::ObjectNotExists(object_id)),
             };
         };
 
-        // Query objects_history for the object with the requested version.
+        // query objects_history for the object with the requested version.
         let history_object = self
+            .db()
             .get_stored_history_object(
                 object_id,
                 object_version_info.object_version,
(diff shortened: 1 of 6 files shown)


crates/iota-indexer/src/read.rs
```diff
@@ -295,7 +312,40 @@ impl IndexerReader {
...
+                        .try_into_object_read(self.package_resolver.clone())
+                        .await?
+                    {
+                        ObjectRead::NotExists(_) | ObjectRead::Deleted(_) => {
+                            Ok(PastObjectRead::VersionNotFound(object_id, object_version))
+                        }
+                        ObjectRead::Exists(obj_ref, object, layout) => {
+                            Ok(PastObjectRead::VersionFound(obj_ref, object, layout))
+                        }
+                    }
```
tomxey: if this is KV specific, maybe all this code could be moved to HistoricalFallbackReader?

crates/iota-indexer/src/read.rs
```diff
@@ -584,28 +645,123 @@ impl IndexerReader {
...
+            }
+
+            // ascending from cursor+1: expect [c+1, ..., c+limit]
+            (Some(c), false) => {
+                // empty result means cursor is at or beyond the latest checkpoint.
+                if checkpoints.is_empty() {
+                    return Ok(vec![]);
+                };
+                // cap range to actual results length in case we're near the end.
+                (c + 1..=c.saturating_add(limit.min(checkpoints.len()) as u64)).collect()
```
tomxey: I think parts of this logic could be simplified actually. E.g. when paginating upward with cursor, if the first checkpoint that you receive from DB is equal to "c+1" then it means the result is correct (if c+1 is not pruned, then also bigger checkpoints are not pruned).

This computation of "expected_range" is actually incorrect I think.

Let's say cps 0..=10 are pruned, and latest cp in indexer is 15.

User queries with cursor=8, ascending, limit=7.
Due to pruning you receive from DB: [11, 12, 13, 14, 15], len is 5
User should receive: [9, 10, 11, 12, 13, 14, 15]

expected range computed by "(c + 1..=c.saturating_add(limit.min(checkpoints.len()) as u64)).collect()" will be: `9..=13`, which is incorrect.

tomxey (changes requested):

crates/iota-indexer/src/read.rs
```diff
@@ -584,28 +645,123 @@ impl IndexerReader {
...
+                    .map(iota_json_rpc_types::Checkpoint::try_from)
+                    .collect::<Result<Vec<_>, IndexerError>>()
+            })
+            .await?;
+
+        // resolve the expected range of checkpoint sequence numbers
+        let expected_range: Vec<CheckpointSequenceNumber> = match (cursor, descending_order) {
+            // ascending from 0: expect [0, 1, ..., limit-1].
+            (None, false) => {
+                // if genesis is present, data is not pruned, use actual results length which
```
tomxey: as below, if genesis is present then that's it, the data is complete, we will not have situation where genesis is there but sth else is pruned

the expected range for fallback is always "(0..limit as u64).collect()", no matter how many checkpoints are returned from DB

crates/iota-indexer/src/read.rs
```diff
@@ -783,40 +932,83 @@ impl IndexerReader {
...
-            .iter()
-            .map(|tx| &tx.transaction_digest)
-            .collect::<HashSet<&Vec<u8>>>();
-        let missing_digests = digests
+        let missing_digests = check_for_missing_tx_digests(&digests, &checkpointed_txs);
+        let optimistic_txs = self
+            .spawn_blocking(move |this| this.get_optimistic_transactions(&missing_digests))
+            .await?;
+
+        let checkpointed_txs = checkpointed_txs
```
tomxey: the variable name is incorrect at this point I think, those are not only checkpointed txs

crates/iota-indexer/src/read.rs
```diff
@@ -783,40 +932,83 @@ impl IndexerReader {
         Ok(result)
     }
 
-    fn multi_get_transactions(
+    async fn multi_get_transactions_in_blocking_task(
         &self,
         digests: &[TransactionDigest],
     ) -> Result<Vec<StoredTransaction>, IndexerError> {
+        let check_for_missing_tx_digests =
+            |digests: &[Vec<u8>], checkpointed_txs: &[StoredTransaction]| -> Vec<Vec<u8>> {
```
tomxey: the arg name `checkpointed_txs` is not 100% accurate as optimistic txs will also be passed there, maybe use arg names like: `requested_digests` and`fetched_txs`?

crates/iota-indexer/src/read.rs
```diff
@@ -1453,16 +1655,22 @@ impl IndexerReader {
...
+        {
+            (Some(timestamp_ms as u64), events)
+        } else if let Some(events) = self.get_optimistic_transaction_events(digest).await? {
+            (None, events)
+        } else if let Some(fallback) = self.fallback_reader() {
+            // fallback to historical storage.
+            return fallback.all_events(digest).await;
+        } else {
+            // is not guaranteed that all transactions have events.
+            (None, vec![])
```
tomxey: Does this change behavior of the endpoint?

Seems that previously it failed if tx with "digest" was not found, but now it will return []

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/2575f692a95468ca3098c278afabbcf6828bc328

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/2575f692a95468ca3098c278afabbcf6828bc328

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/2575f692a95468ca3098c278afabbcf6828bc328

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/2575f692a95468ca3098c278afabbcf6828bc328

crates/iota-indexer/src/read.rs
sergiupopescu199: reworked a bit the overall logic making it simpler.

fixed: https://github.com/iotaledger/iota/pull/9530/commits/2575f692a95468ca3098c278afabbcf6828bc328

crates/iota-indexer/src/read.rs
```diff
@@ -1453,22 +1653,29 @@ impl IndexerReader {
...
+        // try checkpointed data, then optimistic, then fallback if not found
+        let (timestamp_ms, serialized_events) = if let Some((timestamp_ms, events)) =
+            self.try_get_checkpointed_transaction_events(digest).await?
+        {
+            (Some(timestamp_ms as u64), events)
+        } else if let Some(events) = self.get_optimistic_transaction_events(digest).await? {
+            (None, events)
+        } else if let Some(fallback) = self.fallback_reader() {
+            // fallback to historical storage.
+            return fallback.all_events(digest).await;
```
tomxey: If I understand correctly this will return Ok both for existing and not existing transactions, but we want to return Err in case such transaction digest does not exist.

In this case I think `all_events` could return error in case it was unable to find cp summary for given tx. Since that's what will happen if user supplies invalid tx digest.

crates/iota-indexer/src/read.rs
```diff
@@ -783,40 +930,83 @@ impl IndexerReader {
...
     }
 
-    fn multi_get_transactions(
+    async fn multi_get_transactions_in_blocking_task(
         &self,
         digests: &[TransactionDigest],
     ) -> Result<Vec<StoredTransaction>, IndexerError> {
+        let check_for_missing_tx_digests =
+            |requested_digests: &[Vec<u8>], fetched_txs: &[StoredTransaction]| -> Vec<Vec<u8>> {
+                let checkpointed_tx_digests_set = fetched_txs
```
tomxey: nit: `fetched_txs_digests_set` or sth similar

crates/iota-indexer/src/read.rs
```diff
@@ -601,11 +753,13 @@ impl IndexerReader {
...
     /// It uses both the `tx_global_order` table
     /// and the `checkpoints` table to cover old transactions.
-    fn deep_check_all_transactions_are_indexed(
+    pub(crate) async fn deep_check_all_transactions_are_indexed_in_blocking_task(
         &self,
         digests: &[TransactionDigest],
     ) -> Result<bool, IndexerError> {
-        let stored_transactions = self.multi_get_transactions(digests)?;
+        let stored_transactions = self
+            .multi_get_transactions_in_blocking_task(digests)
```
tomxey: going for fallback to do dependency check bothers me a bit, but maybe we can give it a second though in a separate issue

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+        descending_order: bool,
+        limit: usize,
+        db_response: &[iota_json_rpc_types::Checkpoint],
+    ) -> bool {
+        match (cursor, descending_order) {
+            // pruning always removes from the lowest checkpoint upwards, so no gaps.
+            // If genesis (checkpoint 0) is present, data is intact.
+            (None, false) => db_response
+                .first()
+                .is_none_or(|chk| chk.sequence_number == 0),
```
tomxey: shouldn't the condition be reversed? we should fetch from fallback if != 0 ?

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+
+            // for descending, cursor just sets an upper bound. DB returns the highest checkpoint
+            // available. If we got fewer than limit, some data was pruned.
+            (None, true) | (Some(_), true) => db_response.len() != limit,
```
tomxey: there is one more edge case, if you are paginating down, and you reach cp 0 then you cannot get more checkpoints, so you can get less cps than `limit`

maybe the following conditon:

`response_len != limit && last_cp != 0` ?

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+
+            // if first checkpoint matches cursor + 1, data is contiguous from that point.
+            (Some(c), false) => db_response
+                .first()
+                .is_none_or(|chk| chk.sequence_number == c + 1),
```
tomxey: as above, shouldnt the condition be reversed?

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+        }
+
+        // resolve the expected range of checkpoint sequence numbers
+        let checkpoints_keys: Vec<CheckpointSequenceNumber> = match (cursor, descending_order) {
+            // ascending from 0: expect [0, 1, ..., limit-1].
+            (None, false) => (0..limit as u64).collect(),
+
+            // ascending from cursor+1: expect [c+1, ..., c+limit]
+            (Some(c), false) => {
+                // cap range to actual results length in case we're near the end.
```
tomxey: is this comment still valid?

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+            (None, true) => {
+                let Some(latest_checkpoint) = checkpoints.first() else {
+                    // empty table with no cursor to anchor the range.
+                    //
+                    // unlike other cases where we can derive the expected range from
+                    // the cursor, here we rely on the DB to provide the starting point.
+                    // Without any checkpoints, we have no reference for the fallback
+                    // to know which sequence numbers to fetch.
+                    return Err(IndexerError::DataPruned(
+                        "requested checkpoint range not available".into(),
```
tomxey: if there are 0 checkpoints in the DB then it's unlikely that it's a pruning issue, I think we could pick some other error type here. we will never prune all cps

crates/iota-indexer/src/read.rs
```diff
@@ -584,14 +645,105 @@ impl IndexerReader {
...
+            return Err(IndexerError::DataPruned(
+                "requested checkpoint range not available".into(),
+            ));
+        };
+
+        fallback
+            .checkpoints(checkpoints_keys)
+            .await?
             .into_iter()
+            .flatten()
```
tomxey, line 696: have you though about performing some consistency check here? e.g. that there are no gaps in checkpoints returned, and that checkpoints are allowed to be missing only at the end of the range (because they can simply not exist yet). Or do you think that it's not a good idea?

crates/iota-indexer/src/read.rs
sergiupopescu199: It's not perfect because I think in future iterations we should decide the role that the fallback should play for the indexer. For example, should pruned data found in the fallback be treated as data existing on the indexer, or should we treat it as pruned/unavailable? This affects the semantics of dependency checks and other queries.

We could open an issue to discuss the fallback's contract more formally - things like:
- Should fallback data be transparent to API consumers?
- Should certain operations (like dependency checks) prefer local data over fallback?
- How do we handle consistency between local and fallback data?"

In this first iteration I've opted to treat fallback data as indexer data.

crates/iota-indexer/src/read.rs
sergiupopescu199: I was thinking about this, but for the ascending cases:

```rust
  // ascending from 0: expect [0, 1, ..., limit-1].
  (None, false) => (0..limit as u64).collect(),

  // ascending from cursor+1: expect [c+1, ..., c+limit]
  (Some(c), false) => (c + 1..=c.saturating_add(limit as u64)).collect(),

```
While paginating, the `limit` could be greater than the latest available checkpoint, so the consistency check would always fail.

However, if we guarantee that the DB always retains at least `limit` checkpoints (i.e., pruning never goes below that threshold), then this branch would never trigger a fallback fetch - we'd always have enough local data to satisfy the request. In that scenario, we could safely add the consistency check since any fallback response with gaps would indicate a real problem.

For now I'll leave it without the consistency check, but we could revisit this if we enforce a minimum retention policy on the pruning feature."

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9530/commits/34c3050e92f96d0fb217cacc42f5bb490f4e53ee

crates/iota-indexer/src/read.rs
sergiupopescu199: I don't think this is KV-specific. it's more business/API logic. The `HistoricalFallbackReader` is responsible for fetching raw data (`StoredObject`) from the fallback storage, while `IndexerReader` handles the conversion to API types like `PastObjectRead`.

crates/iota-indexer/src/read.rs
tomxey: okay, let's leave this for now

crates/iota-indexer/src/read.rs
tomxey: Here I meant it bothers me because of performance. This is part of transaction execution logic, users trying to execute transactions that depend on old pruned transactions would have to wait for fallback service (given that it's configured). 

We should think about this when working on pruning later.

tomxey (approved):

crates/iota-indexer/src/read.rs
sergiupopescu199: Removed the fallback usage.

kodemartin (approved):
lgtm

PR #9538, opened 2025-12-05
fix(consensus): fix the logic for advancing threshold clock for a supermajority validator

### LINKED ISSUE #9537: consensus: threshold clock is not updated correctly for a committee with a supermajority validator
we replay the dag with own block header, but advance the threshold clock incorrectly
/// Add the block reference that have been accepted and advance the round
    /// accordingly.
    pub(crate) fn add_block(&mut self, block: BlockRef) {
        match block.round.cmp(&self.round) {
            // Blocks with round less than what we currently build are irrelevant here
            Ordering::Less => {}
            // If we processed block for round r, we also have stored 2f+1 blocks from r-1
            Ordering::Greater => {
                self.aggregator.clear();
                self.aggregator.add(block.author, &self.context.committee);
                self.round = block.round;
            }
            Ordering::Equal => {
                if self.aggregator.add(block.author, &self.context.committee) {
                    self.aggregator.clear();
                    // We have seen 2f+1 blocks for current round, advance
                    self.round = block.round + 1;

                    // now record the time of receipt from last quorum
                    let now = Instant::now();
                    self.context
                        .metrics
                        .node_metrics
                        .quorum_receive_latency
                        .observe(now.duration_since(self.quorum_ts).as_secs_f64());
                    self.quorum_ts = now;
                }
            }
        }
    }
it is very specific for a single validator committee. The threshold clock should advance for one round further when we are on the arm
Ordering::Greater => {
                self.aggregator.clear();
                self.aggregator.add(block.author, &self.context.committee);
                self.round = block.round;
            }
In the logs we see a message

2025-12-05T15:03:29.163941Z DEBUG node{name=k#99f25ef6..}: starfish_core::core: Skip proposing for round 1121 as last known proposed round is 1121
from this function. When recovering, we should have advanced one round further, but the above arm didn't update the threshold clock in a right way
/// Whether the core should propose new blocks.
    pub(crate) fn should_propose(&self) -> bool {
        let clock_round = self.dag_state.read().threshold_clock_round();
        let core_skipped_proposals = &self.context.metrics.node_metrics.core_skipped_proposals;

        if !self.quorum_subscribers_exists {
            debug!("Skip proposing for round {clock_round}, don't have a quorum of subscribers.");
            core_skipped_proposals
                .with_label_values(&["no_quorum_subscriber"])
                .inc();
            return false;
        }

        let Some(last_known_proposed_round) = self.last_known_proposed_round else {
            debug!(
                "Skip proposing for round {clock_round}, last known proposed round has not been synced yet."
            );
            core_skipped_proposals
                .with_label_values(&["no_last_known_proposed_round"])
                .inc();
            return false;
        };
        if clock_round <= last_known_proposed_round {
            debug!(
                "Skip proposing for round {clock_round} as last known proposed round is {last_known_proposed_round}"
            );
            core_skipped_proposals
                .with_label_values(&["higher_last_known_proposed_round"])
                .inc();
            return false;
        }

        true
    }

The correct arm should be
Ordering::Greater => {
                self.aggregator.clear();
                if self.aggregator.add(block.author, &self.context.committee) {
               self.aggregator.clear();
                    // We have seen 2f+1 blocks for current round, advance
                    self.round = block.round + 1;

                    // now record the time of receipt from last quorum
                    let now = Instant::now();
                    self.context
                        .metrics
                        .node_metrics
                        .quorum_receive_latency
                        .observe(now.duration_since(self.quorum_ts).as_secs_f64());
                    self.quorum_ts = now;
} else {
                self.round = block.round;
}
            }

This is a very specific issue of a single validator committee when one block/block header is enough to advance the clock round. The function was not designed to cover this case

polinikita:
# Description of change

Fix threshold clock round advancement logic when processing blocks from higher rounds.

  Previously, when the threshold clock jumped to a higher round (receiving a block with `round > current_round`), the first block's author stake was recorded but the quorum check wasn't performed. This caused issues for a validator with super-majority stake (>2/3), as they should immediately advance the round upon receiving their own block from a higher round, but the quorum check only ran in the Equal branch. 

This  change is especially important when running the network containing only one validator as a simple restart would not let the validator to properly restart the consensus logic.

  The fix refactors the logic by:
  1. Extracting quorum checking into a new `try_advance_round()` helper method
  2. Calling `try_advance_round()` after every block addition, regardless of whether the block triggered a round jump or was added to the current round's aggregator.
  3. This ensures super-majority validators properly advance rounds when jumping to higher rounds.

  The same fix is applied to both consensus/core (Mysticeti) and crates/starfish/core (Starfish) threshold clock implementations.

Fixes #9537 

- [x] I have added tests that prove my fix is effective or that my feature works:
    - test_threshold_clock_round_skip_then_quorum: Verifies that after jumping to a higher round, subsequent blocks can form quorum
    - test_threshold_clock_super_majority_round_skip: Verifies that a super-majority authority (>2/3 stake) immediately advances the round when jumping to a
  higher round

### DIFF
--- a/crates/starfish/core/src/transactions_synchronizer.rs
+++ b/crates/starfish/core/src/transactions_synchronizer.rs
@@ -577,7 +577,7 @@ impl<C: NetworkClient, D: CoreThreadDispatcher> TransactionsSynchronizer<C, D> {
 
     /// Starts a task to fetch missing transactions from other authorities.
     async fn start_fetch_missing_transactions_task(&mut self) -> ConsensusResult<()> {
-        info!("Kick in periodic synchronizer to fetch missing transactions");
+        debug!("Kick in periodic synchronizer to fetch missing transactions");
         // Get missing transactions from the core
         let missing_transactions = self
             .core_dispatcher
--- a/crates/starfish/core/src/transactions_synchronizer.rs
+++ b/crates/starfish/core/src/transactions_synchronizer.rs
@@ -577,7 +577,6 @@ impl<C: NetworkClient, D: CoreThreadDispatcher> TransactionsSynchronizer<C, D> {
 
     /// Starts a task to fetch missing transactions from other authorities.
     async fn start_fetch_missing_transactions_task(&mut self) -> ConsensusResult<()> {
-        debug!("Kick in periodic synchronizer to fetch missing transactions");
         // Get missing transactions from the core
         let missing_transactions = self
             .core_dispatcher
--- a/consensus/core/src/threshold_clock.rs
+++ b/consensus/core/src/threshold_clock.rs
@@ -29,9 +29,8 @@ impl ThresholdClock {
         }
     }
 
-    /// Add a block's author to the aggregator. If quorum (2f+1) is reached,
-    /// advance to the next round and record latency metrics.
-    /// Returns true if quorum was reached.
+    /// If quorum (2f+1) is reached, advance to the next round and record
+    /// latency metrics. Returns true if quorum was reached.
     fn try_advance_round(&mut self, new_round: Round) -> bool {
         if self.aggregator.reached_threshold(&self.context.committee) {
             self.aggregator.clear();
--- a/consensus/core/src/threshold_clock.rs
+++ b/consensus/core/src/threshold_clock.rs
@@ -29,35 +29,49 @@ impl ThresholdClock {
         }
     }
 
-    /// Add the block reference that have been accepted and advance the round
-    /// accordingly.
+    /// Add a block's author to the aggregator. If quorum (2f+1) is reached,
+    /// advance to the next round and record latency metrics.
+    /// Returns true if quorum was reached.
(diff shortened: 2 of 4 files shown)


kmk142789 (approved):

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -577,7 +577,7 @@ impl<C: NetworkClient, D: CoreThreadDispatcher> TransactionsSynchronizer<C, D> {
 
     /// Starts a task to fetch missing transactions from other authorities.
     async fn start_fetch_missing_transactions_task(&mut self) -> ConsensusResult<()> {
-        info!("Kick in periodic synchronizer to fetch missing transactions");
+        debug!("Kick in periodic synchronizer to fetch missing transactions");
```
piotrm50: Already removed this log completely in a separate PR.

piotrm50 (approved):

VorobyevIlya (approved):

PR #9546, opened 2025-12-08
feat(iota-core, iota-framework): Add adjust_rewards_by_score feature flag

cyberphysic4l:
# Description of change

Adds a new feature flag `adjust_rewards_by_score` to control whether the calculated scores should be used to adjust rewards during the advance epoch function. This new flag is passed to `ChangeEpochV4` to change the behaviour of the advance epoch function where rewards are adjusted at the end of each epoch.  
`score_based_rewards` is also renamed to `calculate_validator_scores` and is used to indicate whether the scores are calculated and passed to `ChangeEpochV4`, but does not indicate whether those scores affect rewards.

### Release Notes

- [x] Protocol: Introduces feature flags for protocol v18, `calculate_validator_scores` and `adjust_rewards_by_score` indicated if validator scores are calculated and whether they are used to adjust rewards, respectively.
- [x] Indexer: Enables a new `ChangeEpochV4` system transaction which may appear in checkpoints for networks where `calculate_validator_scores` protocol config is enabled. Consequently a dependency bump is necessary for `iota_data_ingestion_core` library consumers (custom indexers) to prevent outages.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


Cargo.toml
```diff
@@ -457,7 +457,7 @@ iota-rosetta = { path = "crates/iota-rosetta" }
 iota-rpc-loadgen = { path = "crates/iota-rpc-loadgen" }
 iota-sdk = { path = "crates/iota-sdk" }
 # core-types with json format for REST API
-iota-sdk2 = { package = "iota-rust-sdk", git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "c7eeee26ef4b4c0e0cf7b17fcec14b7e1e79910c", features = ["hash", "serde", "schemars"] }
+iota-sdk2 = { package = "iota-rust-sdk", git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "014d144cc3153e9fba58237670a197c84ad3a040", features = ["hash", "serde", "schemars"] }
```
thibault-martinez: You may want to update `protocol-research/feat/score-integration` because this has changed a bit

kodemartin (approved):

Cargo.toml
cyberphysic4l: I think I have resolved this correctly now, please take a look.

Cargo.toml
```diff
@@ -298,7 +298,7 @@ indexmap = { version = "2.11.0", features = ["serde"] }
 indicatif = "0.18.3"
 insta = { version = "1.21.1", features = ["redactions", "yaml", "json"] }
 integer-encoding = "3.0.1"
-iota-sdk-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "c7eeee26ef4b4c0e0cf7b17fcec14b7e1e79910c", features = ["hash", "serde", "schemars"] }
+iota-sdk-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "2b19703f2082e4bd4af9b73832eb93c08c851f0a", features = ["hash", "serde", "schemars"] }
```
thibault-martinez: Will need to update the revision when your SDK PR gets squashed

thibault-martinez (approved):

Cargo.toml
thibault-martinez: It's been merged, you can use the develop revision now

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1459,13 +1464,22 @@ impl ProtocolConfig {
...
+        );
+        calculate_validator_scores
+    }
+
+    pub fn adjust_rewards_by_score(&self) -> bool {
+        let adjust = self.feature_flags.adjust_rewards_by_score;
         assert!(
-            !score_based_rewards || self.scorer_version.is_some(),
-            "score_based_rewards requires scorer_version to be set"
+            !adjust || self.scorer_version.is_some(),
```
muXxer: Should there also be a check that `calculate_validator_scores` is enabled when `adjust_rewards_by_score` is set?

crates/iota-types/src/transaction.rs
```diff
@@ -620,12 +624,11 @@ impl EndOfEpochTransactionKind {
                         "selecting committee only among validators supporting the protocol version not supported".to_string(),
                     ));
                 }
-                // This has to be uncommented before merging to develop
-                // if config.score_based_rewards() {
-                //     return Err(UserInputError::Unsupported(
-                //         "score based rewards enabled and
-                // required".to_string(),     ));
-                // }
+                if config.calculate_validator_scores() {
```
muXxer, line 627: Do we also need to check that `adjust_rewards_by_score` is not enabled?

crates/iota-protocol-config/src/lib.rs
cyberphysic4l: No harm to have this check, I will add it.

crates/iota-types/src/transaction.rs
cyberphysic4l: Sure, I'm not sure it's necessary to check, but I will add the check anyway. Note: it is not required to be enabled for V4 as in other cases, so no check being added there.

muXxer (approved):

valeriyr (approved):

PR #9548, opened 2025-12-08
fix(indexer): Mark indexer store tasks as "CALLED_FROM_BLOCKING_POOL"

### LINKED ISSUE #9464: bug(indexer): Flaky `test_optimistic_tables_pruning`
Sometimes on CI the test fails with following error:

```
failures:

---- write_api::test_optimistic_tables_pruning stdout ----

thread 'tokio-runtime-worker' panicked at crates/iota-indexer/src/store/pg_indexer_store.rs:1720:9:
you are calling a blocking DB operation directly on an async thread. Please use IndexerReader::spawn_blocking instead to move the operation to a blocking thread
stack backtrace:
```

Which traces to the call to:
```rust
fn get_watermarks(&self) -> Result<(Vec<StoredWatermark>, i64), IndexerError> {
        // read_only transaction, otherwise this will block and get blocked by write
        // transactions to the same table.
        run_query_with_retry!(
            &self.blocking_cp,
            |conn| {
                ...
            },
            PG_DB_COMMIT_SLEEP_DURATION
        )
    }
```

But the function seems to be wrapped in a blocking task:

```rust
    async fn get_watermarks(&self) -> Result<(Vec<StoredWatermark>, i64), IndexerError> {
        self.execute_in_blocking_worker(move |this| this.get_watermarks())
            .await
    }
```

To be investigated what is the exact cause.

tomxey:
# Description of change

Some of the blocking DB queries require CALLED_FROM_BLOCKING_POOL to be set to true, otherwise they panic.

PgIndexerStore should set this flag to true similarly to how IndexerReader is doing it, so that blocking DB queries work properly.

fixes https://github.com/iotaledger/iota/issues/9464

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -79,7 +79,10 @@ use crate::{
         objects, objects_history, objects_snapshot, objects_version, optimistic_transactions,
         packages, pruner_cp_watermark, transactions, tx_digests, tx_global_order,
     },
-    store::{diesel_macro::*, package_resolver::IndexerStorePackageResolver},
+    store::{
+        diesel_macro::{mark_in_blocking_pool, *},
+        package_resolver::IndexerStorePackageResolver,
+    },
     types::{IndexerResult, OwnerType},
 };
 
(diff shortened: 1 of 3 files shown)


sergiupopescu199 (approved):
lgtm

kodemartin (approved):

PR #9558, opened 2025-12-09
fix(script): fix `iota-sdk-types` special case in `cargo_sort.py`

muXxer:
# Description of change

This PR fixes the special case for the external rust sdk in `cargo_sort.py`


thibault-martinez (commented):
I don't know tbh, for me, and as suggested by the Cargo.toml file, `internal dependencies` are only workspace members, the SDK is not...

thibault-martinez (approved):

scripts/cargo_sort/cargo_sort.py
```diff
@@ -347,7 +347,8 @@ def run_dprint_fmt(directory):
     internal_crates_dict = get_package_names_from_cargo_tomls(args.target)
     
     # add special cases
-    internal_crates_dict["iota-rust-sdk"] = None
+    internal_crates_dict["iota-sdk-types"] = None
```
DaughterOfMars, line 350: may want to also handle iota-sdk since we'll probably end up using that instead

thibault-martinez: I told Max to only add the 2 crates we are presently using in the monorepo and that we would add more ourselves when/if needed. I don't mind them being preventively added if you prefer that.

DaughterOfMars: ok

PR #9563, opened 2025-12-09
feat(graphql-rpc): Optimize reading objects_history in consistent views

### LINKED ISSUE #7296: GraphQL: Requested data is outside the available range
## Bug description

Requesting some paginated data with many pages fails at some point with
```
[
  {
    "message": "Requested data is outside the available range",
    "locations": [
      {
        "line": 1,
        "column": 23
      }
    ],
    "path": [
      "objects"
    ],
    "extensions": {
      "code": "BAD_USER_INPUT"
    }
  }
]
```
query is always the same, just cursor updated to the one from the pageInfo

Related to this, what is even the definition of this `available range`? Shouldn't everything be available?

## Version

- Version number, commit, or branch: 1.2.1-rc

## Steps To reproduce the bug

Fails in mainnet and testnet, devnet works fine (I guess because there aren't that many objects)
bash or TS script:

```bash
#!/bin/bash

ENDPOINT="https://graphql.mainnet.iota.cafe"
TYPE="0x3::staking_pool::StakedIota"
CURSOR="null"
declare -A UNIQUE_ADDRESSES=()  # Associative array acts as a hash set
STAKE_OBJECTS=0

while :; do
  if [[ "$CURSOR" == "null" ]]; then
    AFTER_ARGS=""
  else
    ESCAPED_CURSOR=$(echo -n "$CURSOR" | tr -d '\n')
    AFTER_ARGS=", after: \\\"$ESCAPED_CURSOR\\\""
  fi

  QUERY="{\"query\":\"query getStakedIota { objects(filter: {type: \\\"$TYPE\\\"}$AFTER_ARGS) { nodes { address owner { ... on AddressOwner { owner { address } } } } pageInfo { hasNextPage endCursor } } }\"}"

  RESPONSE=$(curl -s -X POST "$ENDPOINT" \
    --data "$QUERY")

  # Check for errors in the response
  ERRORS=$(echo "$RESPONSE" | jq -r '.errors // empty')
  if [[ -n "$ERRORS" ]]; then
  echo "Error in GraphQL response:"
  echo "$RESPONSE" | jq '.errors'
  break
  fi

  # Extract nodes and pageInfo
  NODES=$(echo "$RESPONSE" | jq -c '.data.objects.nodes[]?')
  for NODE in $NODES; do
    STAKE_OBJECTS=$((STAKE_OBJECTS+1))
    OWNER_ADDR=$(echo "$NODE" | jq -r '.owner.owner.address // empty')
    if [[ -n "$OWNER_ADDR" ]]; then
      UNIQUE_ADDRESSES["$OWNER_ADDR"]=1  # Add to hash set
    else
      echo "No owner found for node: $NODE" >&2
    fi
  done

  HAS_NEXT=$(echo "$RESPONSE" | jq -r '.data.objects.pageInfo.hasNextPage')
  CURSOR=$(echo "$RESPONSE" | jq -r '.data.objects.pageInfo.endCursor')

  echo "${#UNIQUE_ADDRESSES[@]} unique addresses, $STAKE_OBJECTS stake objects"
  if [[ "$HAS_NEXT" != "true" ]]; then
    break
  fi
done

echo "Total unique addresses: ${#UNIQUE_ADDRESSES[@]}"
echo "Total stake objects: $STAKE_OBJECTS"
# echo "Unique addresses:"
# printf "%s\n" "${!UNIQUE_ADDRESSES[@]}"
```
```TS
import { GraphQLQueryOptions, GraphQLQueryResult, IotaGraphQLClient } from '@iota/iota-sdk/graphql';
import { graphql } from '@iota/iota-sdk/graphql/schemas/2025.2';

const STAKED_IOTA_TYPE = "0x3::staking_pool::StakedIota";
// const TIMELOCKED_STAKED_IOTA_TYPE = "0x3::timelocked_staking::TimelockedStakedIota";

(async () => {
    const gqlClient = new IotaGraphQLClient({
        url: 'https://graphql.testnet.iota.cafe',
    });

    let res = { stakeObjects: 0, addresses: new Set<string>() };

    let cursorSection;
    while (true) {
        let query = `query getStakedIota($type: String, $cursor: String) {
            objects(filter: {type: $type}, after: $cursor) {
                nodes {
                    address
                    owner {
                        ... on AddressOwner {
                            owner {
                                address
                            }
                        }
                    }
                }
                pageInfo {
                hasNextPage
                endCursor
                }
            }
        }`;

        let result: GraphQLQueryResult = await queryGraphQl(
            gqlClient,
            query,
            {
                type: STAKED_IOTA_TYPE,
                cursor: cursorSection
            },
        );

        if (result.errors) {
            throw new Error("GraphQL query failed: " + JSON.stringify(result.errors));
            break;
        }

        // @ts-ignore
        for (let node of result.data.objects.nodes) {
            res.stakeObjects++;
            if (node.owner.owner) {
                res.addresses.add(node.owner.owner.address)
            } else {
                console.warn("No owner found for node: ", node);
            }
        }

        console.log(res.addresses.size, " addresses, ", res.stakeObjects, " stake objects");
        // @ts-ignore
        if (result.data.objects.pageInfo.hasNextPage) {
            // @ts-ignore
            cursorSection = result.data.objects.pageInfo.endCursor;
            console.log("next cursor:" + cursorSection);
        } else {
            break;
        }
    }
    console.log(res);
})();

async function queryGraphQl(
    gqlClient: IotaGraphQLClient,
    query: string,
    variables: Record<string, any>,
): Promise<GraphQLQueryResult> {
    const options: GraphQLQueryOptions = {
        query: graphql(query),
        variables,
    };
    return gqlClient.query(options);
}

```

## Expected behaviour

It should get the data from all pages and never error

## Actual behaviour

At some point it fails with an error `[{"message":"Requested data is outside the available range","locations":[{"line":2,"column":3}],"path":["objects"],"extensions":{"code":"BAD_USER_INPUT"}}]`

tomxey:
# Description of change

Optimize reading objects_history in consistent views.

The original approach was reading all matching (that passed `filter_fn` and the `range` check) rows from `objects_history` and then filtered them to leave only those that for given object_id DON'T have any other row in `objects_history` that satisfies `newer_criteria`, `range` check and that have higher version than selected row.

The filter we use when getting objects is different than the filter we use when trying to determine if it is the most recent version,  and this actually is relevant.

Imagine we have object A which at version 3 belongs to Bob, and at Version 5 belongs to Alice.
Now if we want to find all objects belonging to Bob, and in our available range we see A both in versions 3 and 5, we shouldn't return A:3 because that's the most recent version that satisfies the criteria but rather we should not return the object at all, because at the most recent version of it in our available range the object doesn't belong to Bob.

So the behaviour of the new query is like: "fetch all objects that match the criteria, but eventually return only those that happen to be the most recent version of given object in given checkpoint range", which should give the same results as the original approach. (the assumption for this is that `newer_criteria` will NOT filter out object version that should be returned, and it is the case in current codebase).

fixes https://github.com/iotaledger/iota/issues/7296

### DIFF
--- a/crates/iota-graphql-rpc/src/consistency.rs
+++ b/crates/iota-graphql-rpc/src/consistency.rs
@@ -119,21 +119,20 @@ pub(crate) fn build_objects_query(
     filter_fn: impl Fn(RawQuery) -> RawQuery,
     newer_criteria: impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
-    // Subquery to be used in `LEFT JOIN` against the inner queries for more recent
-    // object versions
-    let newer = newer_criteria(filter!(
-        query!("SELECT object_id, object_version FROM objects_history"),
-        format!(
-            r#"checkpoint_sequence_number BETWEEN {} AND {}"#,
-            range.first, range.last
-        )
-    ));
-
     let mut snapshot_objs_inner = query!("SELECT * FROM objects_snapshot");
     snapshot_objs_inner = filter_fn(snapshot_objs_inner);
 
     let mut snapshot_objs = match view {
         View::Consistent => {
+            // Subquery to be used in `LEFT JOIN` for more recent object versions
+            let newer = newer_criteria(filter!(
+                query!("SELECT object_id, object_version FROM objects_history"),
+                format!(
+                    r#"checkpoint_sequence_number BETWEEN {} AND {}"#,
+                    range.first, range.last
+                )
+            ));
+
             // The `LEFT JOIN` serves as a filter to remove objects that have a more recent
             // version
             let mut snapshot_objs = query!(
@@ -163,36 +162,44 @@ pub(crate) fn build_objects_query(
 
     // Similar to the snapshot query, construct the filtered inner query for the
     // history table.
-    let mut history_objs_inner = query!("SELECT * FROM objects_history");
-    history_objs_inner = filter_fn(history_objs_inner);
+    let mut history_window = query!("SELECT * FROM objects_history");
+    history_window = filter_fn(history_window);
 
     let mut history_objs = match view {
         View::Consistent => {
             // Additionally bound the inner `objects_history` query by the checkpoint range
-            history_objs_inner = filter!(
-                history_objs_inner,
+            history_window = filter!(
+                history_window,
                 format!(
                     r#"checkpoint_sequence_number BETWEEN {} AND {}"#,
                     range.first, range.last
                 )
             );
 
-            let mut history_objs = query!(
-                r#"SELECT candidates.* FROM ({}) candidates
-                    LEFT JOIN ({}) newer
-                    ON (candidates.object_id = newer.object_id AND candidates.object_version < newer.object_version)"#,
-                history_objs_inner,
-                newer
+            let newest = newer_criteria(filter!(
+                query!("SELECT object_id, MAX(object_version) AS max_version FROM objects_history"),
+                format!(
+                    r#"checkpoint_sequence_number BETWEEN {} AND {}"#,
+                    range.first, range.last
+                )
+            ))
+            .group_by("object_id");
+
+            let history_objs = query!(
+                r#"WITH history_window AS ({}),
+                    newest AS ({})
+                    SELECT candidates.* FROM history_window candidates
+                    JOIN newest
+                    ON candidates.object_id = newest.object_id
+                    AND candidates.object_version = newest.max_version"#,
+                history_window,
+                newest
             );
-            history_objs = filter!(history_objs, "newer.object_version IS NULL");
             history_objs
         }
         View::Historical => {
             // The cursor pagination logic refers to the table with the `candidates` alias
-            query!(
-                "SELECT candidates.* FROM ({}) candidates",
-                history_objs_inner
-            )
+            query!("SELECT candidates.* FROM ({}) candidates", history_window)
         }
     };
 


crates/iota-graphql-rpc/src/consistency.rs
```diff
@@ -163,36 +162,44 @@ pub(crate) fn build_objects_query(
...
-                history_objs_inner,
-                newer
+            let newest = newer_criteria(filter!(
+                query!("SELECT object_id, MAX(object_version) AS max_version FROM objects_history"),
+                format!(
+                    r#"checkpoint_sequence_number BETWEEN {} AND {}"#,
+                    range.first, range.last
+                )
+            ))
+            .group_by("object_id");
```
kodemartin, line 186: I think we can benefit if we apply `filter_fn` on `newest` as well.

tomxey: I think it would be incorrect, let me bring back the section from the description:


> Imagine we have object A which at version 3 belongs to Bob, and at Version 5 belongs to Alice.
Now if we want to find all objects belonging to Bob, and in our available range we see A both in versions 3 and 5, we shouldn't return A:3 because that's the most recent version that satisfies the criteria but rather we should not return the object at all, because at the most recent version of it in our available range the object doesn't belong to Bob.

So let's say range is (100, 300), which means that cp_viewed_at is 300.
So we want to find objects owned by Bob at checkpoint 300.

let's say: `filter_fn = |obj| obj.owner = "Bob`, which for our range returns object A at version 3
so `history_window = [A:3]`

for this, `newer_criteria` will not filter out anything, so it's a noop
for the situation described above, `newest = [A:5, ...]` (three dots because it can also contain objs other than A)

Finally, we return only those rows from `history_window` that also appear in `newest`, so in this case we don't return anything, and that's correct because at checkpoint 300 Bob did not own any objects. The object that was owned by him earlier is owned by Alice at checkpoint 300.

------

Now if we also apply `filter_fn` when computing `newest`, we will get the most recent versions of objects in our range that satisfy the `filter_fn`.
In our example we would get: `newest = [A:3]`, which would match what we have in `history_window` so we will return [A:3] as a final result. Which I think is incorrect, and it's also not how the original query behaved.

kodemartin: You are absolutely right!

kodemartin (approved):
lgtm

sergiupopescu199 (approved):

PR #9564, opened 2025-12-09
feat(consensus): introduce fast commit syncer for benchmarking

VorobyevIlya:
# Description of change

This PR introduces a fast commit syncer.
Fast commit syncer requests commits and transactions from these commits, but doesn't need headers.
Transactions are verified with transaction commitment from transaction refs.
Then CommittedSubdags are created and added to the core.
This is a first version for benchmarking, so the addition of committed subdags is not really done properly at the moment.
For now, we just want to measure how much we can win by not downloading the headers.

#9565 #9566 #9567

- [x] Basic tests (linting, compilation, formatting, unit/integration tests).
- [x] I have checked that new and existing unit tests pass locally with my changes: we benchmarked the performance of a new fast commit syncer and it works very fast, comparable to checkpoint synchronizer.

### DIFF
--- a/crates/starfish/core/src/linearizer.rs
+++ b/crates/starfish/core/src/linearizer.rs
@@ -114,6 +114,11 @@ impl Linearizer {
             }
         }
 
+        for block_header in &to_commit {
+            self.traversed_headers_tracker
+                .insert(block_header.reference());
+        }
+
         // Collect all block references for transactions that reached quorum after
         // adding acknowledgments
         let committed_transactions = to_commit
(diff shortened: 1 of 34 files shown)


crates/starfish/core/src/linearizer.rs
```diff
@@ -114,6 +114,11 @@ impl Linearizer {
             }
         }
 
+        for block_header in &to_commit {
```
filipdulic: this looks redundant considering the code above.

muXxer (approved):
Approving the change to the grafana dashboard.

polinikita (approved):

PR #9492, opened 2025-12-02
feat: Add the MoveAuthenticator generic signature variant and enable AA transactions

miker83z:
# Description of change

This part involves the development of a new iota type to represent a new `GenericSignature` variant, that is the `MoveAuthenticator` type, and a execution path to authenticate an Account being the sender of a TX.

[Internal AA architecture document](https://iotafoundation.atlassian.net/wiki/x/EoBOng).

- The `MoveAuthenticator` type has a unique field which is `call_args`; this is a vector of `CallArg` working similarly to the `ProgrammableTransactionBlock` inputs.
- The `MoveAuthenticator` is used to pass the inputs to the move call indicated by the `AuthenticatorInfoV1` found attached to the Account object.
- This authentication job is performed by the Validator, which:
  - checks that the `MoveAuthenticator` is correctly formed
  - loads `MoveAuthenticator` input objects at a specific version
  - loads the `AuthenticatorInfoV1` from the account object dynamic field
  - finally, executes the `authenticate()` function (that has no effects to store)

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [x] Protocol: Add the flag move_auth and the protocol parameter max_auth_gas, which allow to enable the authentication of Abstract Accounts and define the maximum gas budget allowed for the execution of the authentication.
- [x] Nodes (Validators and Full nodes): Enable the authentication of Abstract Accounts on Devnet.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Add a new GenericSignature variant named MoveAuthenticator, which allows to authenticate Abstract Accounts. 
- [ ] REST API:

PR #9586, opened 2025-12-11
feat: Account Abstraction beta

### LINKED ISSUE #8805: [AA] Account Abstraction Beta
Tracks the implementation of the core features of Account Abstraction after the alpha version (see #7901) and before the beta release.

miker83z:
# Description of change

Enables Account Abstraction core features at the protocol level.

[Internal AA architecture document](https://iotafoundation.atlassian.net/wiki/x/EoBOng).

Main protocol changes:

- feat(move): Add custom attributes #9453
  - added a generic FlavoredAttribute which can then allow us to define new attributes within the iota_mode Move compiler part;
  - allow the insertion of specific metadata in the serialized compiled modules, i.e., bytecode (protocol feature flag `metadata_in_module_bytes`). 

- feat(execution): create a package metadata object at publish time #9461
  - every time a function uses the `#[authenticator]` annotation, the node’s publish/upgrade execution tries to create a `PackageMetadata` immutable object;
  - a new set of functions in the `iota-verifier` crate implement the checking of the requirements for a package metadata to be valid;
  - once the PackageMetadata immutable object is created, then the framework method `create_auth_function_ref_v1` can be used as the only way to create an `AuthenticatorFunctionRefV1` instance.
  - the `AuthenticatorFunctionRefV1` struct includes the fields to uniquely identify a function on-chain: package id, module name, function name.
  - An Account object is an object with a dynamic field being a `AuthenticatorFunctionRefV1`.

- feat: Add the MoveAuthenticator generic signature variant and enable AA transactions #9492
  - introduced a new iota type to represent a new `GenericSignature` variant, that is the `MoveAuthenticator` type, and a execution path to authenticate an Account being the sender of a TX.
  - The `MoveAuthenticator` type has a unique field which is `call_args`; this is a vector of `CallArg` working similarly to the `ProgrammableTransactionBlock` inputs.
  - The `MoveAuthenticator` is used to pass the inputs to the move call indicated by the `AuthenticatorFunctionRefV1` found attached to the Account object.
  - This authentication job is performed by the Validator, which: (1) loads `MoveAuthenticator` input objects at a specific version; (2) then, executes the `authenticate()` function referenced by `AuthenticatorFunctionRefV1`, having no effects to store to the ledger.


begonaalvarezd (approved):

polinikita (approved):

kodemartin (approved):

thibault-martinez (approved):

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -47,25 +47,35 @@ mod checked {
...
+                reference_gas_price,
+                gas,
+                gas_budget,
+                transaction.gas_price(),
+            )
+        }
     }
 
-    #[instrument(level = "trace", skip_all, fields(tx_digest = ?transaction.digest()))]
+    #[instrument(level = "trace", skip_all)]
```
muXxer: Why was the digest removed here?

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -215,13 +287,13 @@ mod checked {
...
             reference_gas_price,
             transaction,
+            authentication_gas_budget,
         )?;
         check_objects(transaction, input_objects)?;
 
         Ok(gas_status)
     }
 
-    #[instrument(level = "trace", skip_all)]
```
muXxer, line 224: Any why was that removed here?

muXxer (approved):
I skipped iota-e2e-test files, but the rest LGTM!

Approving files owned by node/core-protocol.

miker83z:
Currently, the merging is blocked by this issue: #9862

crates/iota-transaction-checks/src/lib.rs
miker83z: It was an error, fixed now!

examples/move/account/.gitignore
Dr-Electron, line 1: Why is this the only gitignore? If we keep it shouldn't it be located somewhere else? More "root" to catch all build folder?

examples/move/account/README.md
```diff
@@ -0,0 +1,72 @@
+# Account
+
+## Install iota binary with AA feature
+
+(only needed until AA is stable or replace commands to use `./target/debug/iota` instead)
```
Dr-Electron: Weird sentence but as we won't keep you can also keep it like this if you want 😆 

```suggestion
(only needed until AA is stable Alternatively, you can build the iota binary and use the one in  `./target/debug/iota` instead)
```

Or maybe I am misunderstanding the sentence

examples/move/spending_limit/.gitignore
Dr-Electron, line 1: Same here again. Do we want to gitignore all build folder?

Dr-Electron (approved):
Approving the examples. Just a comment about gitignore usage and once sentence

examples/move/account/README.md
miker83z: Updated

valeriyr (approved):

roman1e2f5p8s (approved):

PR #9593, opened 2025-12-11
feat(examples): cherry pick AA examples from alpha

miker83z:
# Description of change

All the examples in this PR are fixed and ready to be merged.

There are a couple of approaches that were used when the examples were created:
1. `Stand-alone accounts`. A package contains the account struct itself as well as the related authentication logic.
2. `IOTAccount`. An example where an account is implemented to be extended with data and the related authentication logic from 3rd-party packages. This example is not finished yet; it requires security rules to be implemented to control access to the internal state.
3. `Extension packages`. Packages where data(a public key, etc.) and the related authenticator logic are implemented in a dedicated module; It makes it possible to reuse such utilities when a new account type is implemented.

We need to define an approach(es) that we want to demonstrate to users and categorize the examples in a separate pull request.


examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
```diff
@@ -0,0 +1,264 @@
...
+
+// -------------------------------------- Creation --------------------------------------
+
+/// Creates a new `DynamicMultisigAccount` instance as a shared object with the given
+/// members, threshold and authenticator.
+public fun create(
+    members_addresses: vector<address>,
+    members_weights: vector<u64>,
+    threshold: u64,
+    authenticator: AuthenticatorInfoV1<DynamicMultisigAccount>,
```
Dkwcs, line 50: Do we have to pass the packageMetadata instead of authenticator directly?
(Also related to example below)

Dkwcs (approved):
LGTM, left one comment.

examples/move/dynamic_multisig_account/sources/dynamic_multisig_account.move
valeriyr: It shouldn't be a part of this PR; technically, it is not necessary because it is possible to create an `AuthenticatorInfoV1` instance in a dedicated MoveCall in a PTB.
We can consider adding this in the following PR related to the examples. We need to discuss this and determine what the examples should look like before making any changes to them(not to waste time and not to create more mess than we already have).

theiari: Yeah, I think it'd be better to keep it "simpler", but yes we could consider adding it later on.

theiari (approved):
Looks good, replied to that question on `dynamic_multisig_account::create` function

Dr-Electron (approved):

PR #9154, opened 2025-11-05
feat(explorer, evm-bridge, dashboard): update amplitude to v2

panteleymonchuk:
Closes #9127

PR #9606, opened 2025-12-12
feat(evm-bridge, explorer, dashboard): use cookie-policy page from iota/core

### LINKED ISSUE #9318: Fix cookie policy page
- Improve page spacings
- Ensure page mantains the site layout, with footer and navbar
- List amplitude cookies too

Do it for all projects where the cookie manager is installed: explorer, dashboard & bridge

panteleymonchuk:
Closes: #9318
~~Depends on #9154~~


begonaalvarezd (commented):
After discussing in a call, here is a summary of the necessary changes
- We have to list all cookies in the cookie policy page at all times, not just the cookies that are present at the time. The cookie policy page is informational
- Necessary cookies are not amplitude cookies, the only necessary cookie we have is the one we set for storing user's preferences
- We have to update the purpose description of the necessary cookie to "Stores the user´s Amplitude cookies consent state for the current domain"
- Amplitude cookies are not necessary cookies but additional cookies
- We have to list the ampltiude cookies in a generic way and not with the project id, so they should be listed as `AMP_` & `AMP_MKTG_`. Take the descriptions and expirations from the documentation https://amplitude.com/docs/sdks/analytics/browser/cookies-and-consent-management
- After user rejects amplide analytics we are missing a `reset` https://amplitude.com/docs/sdks/analytics/browser/cookies-and-consent-management#remove-amplitude-cookies
- Also, lets make sure the cookie policy is accesible in all apps at all times: dashboard in welcome page and in settings next to terms, bridge & explorer in footer

Juligs (approved):

msarcev (approved):

PR #9612, opened 2025-12-12
feat(bridge, explorer, dashboard): add setNetworkGroup

### LINKED ISSUE #9238: Add missing network amplitude group
Similar to how we have it in the wallet https://github.com/iotaledger/iota/blob/16c6fe67a7a7d55777a2949c3f95a758190decfd/apps/wallet/src/shared/analytics/amplitude.ts#L63 we should also add this grouping to the dashboard

### LINKED ISSUE #9241: Add missing network amplitude group
Similar to how we have it in the wallet https://github.com/iotaledger/iota/blob/16c6fe67a7a7d55777a2949c3f95a758190decfd/apps/wallet/src/shared/analytics/amplitude.ts#L63 we should also add this grouping to the explorer

Check if the user property needs to be added too

panteleymonchuk:
Closes #9238


msarcev (changes requested):
Issues building dashboard

brancoder (changes requested):
conflicts 🙏

PR #9616, opened 2025-12-15
feat(graphql-rpc): add `Event.transactionBlock` field

### LINKED ISSUE #9402: upstream(graphql): Event.transactionBlock
https://github.com/MystenLabs/sui/pull/19669

kodemartin:
# Description of change

This patch exposes the parent transaction of an event through the available event query paths.

As this introduces the possibility of recursive queries (e.g. `events -> tranasctionBlock -> events -> ...`, new complexity limits are introduced on the validation level exploiting [https://async-graphql.github.io/async-graphql/en/depth_and_complexity.html](async-graphql tooling) as opposed to the ad-hoc validation alredy present.

With the new limits, queries that try to recurse unnecessarily through the available query paths are rejected as "Too complex".

Resolves #9402 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL: :bulb:  A new field `Event.transactionBlock` is introduced that exposes the parent transaction of an event.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-graphql-rpc/src/server/builder.rs
+++ b/crates/iota-graphql-rpc/src/server/builder.rs
@@ -90,6 +90,8 @@ use crate::{
 /// The default allowed maximum lag between the current timestamp and the
 /// checkpoint timestamp.
 const DEFAULT_MAX_CHECKPOINT_LAG: Duration = Duration::from_secs(300);
+/// The maximum complexity allowed for native [`async_graphql`] checks.
+const MAX_COMPLEXITY: usize = 3000;
 
 pub(crate) struct Server {
     router: Router,
@@ -537,7 +539,7 @@ impl ServerBuilder {
 
 fn schema_builder() -> SchemaBuilder<Query, Mutation, Subscription> {
     async_graphql::Schema::build(Query, Mutation, Subscription)
-        .limit_complexity(6000)
+        .limit_complexity(MAX_COMPLEXITY)
         .register_output_type::<IMoveObject>()
         .register_output_type::<IObject>()
         .register_output_type::<IOwner>()
--- a/crates/iota-graphql-rpc/src/server/builder.rs
+++ b/crates/iota-graphql-rpc/src/server/builder.rs
@@ -91,7 +91,7 @@ use crate::{
 /// checkpoint timestamp.
 const DEFAULT_MAX_CHECKPOINT_LAG: Duration = Duration::from_secs(300);
 /// The maximum complexity allowed for native [`async_graphql`] checks.
-const MAX_COMPLEXITY: usize = 3000;
+const MAX_COMPLEXITY: usize = 6000;
 
 pub(crate) struct Server {
     router: Router,
--- a/crates/iota-graphql-rpc/src/server/builder.rs
+++ b/crates/iota-graphql-rpc/src/server/builder.rs
@@ -537,6 +537,7 @@ impl ServerBuilder {
 
 fn schema_builder() -> SchemaBuilder<Query, Mutation, Subscription> {
     async_graphql::Schema::build(Query, Mutation, Subscription)
+        .limit_complexity(6000)
         .register_output_type::<IMoveObject>()
         .register_output_type::<IObject>()
         .register_output_type::<IOwner>()
--- a/crates/iota-graphql-rpc/src/types/checkpoint.rs
+++ b/crates/iota-graphql-rpc/src/types/checkpoint.rs
@@ -152,7 +152,6 @@ impl Checkpoint {
     }
 
     /// The epoch this checkpoint is part of.
-    #[graphql(complexity = 0)]
     async fn epoch(&self, ctx: &Context<'_>) -> Result<Option<Epoch>> {
         Epoch::query(
             ctx,
--- a/crates/iota-graphql-rpc/src/types/checkpoint.rs
+++ b/crates/iota-graphql-rpc/src/types/checkpoint.rs
@@ -16,6 +16,7 @@ use iota_types::messages_checkpoint::CheckpointDigest;
 use serde::{Deserialize, Serialize};
 
 use crate::{
+    config::DEFAULT_PAGE_SIZE,
     connection::ScanConnection,
     consistency::Checkpointed,
     data::{self, Conn, DataLoader, Db, DbConnection, QueryExecutor},
@@ -92,12 +93,14 @@ impl Checkpoint {
     /// in Base58. This hash can be used to verify checkpoint contents by
     /// checking signatures against the committee, Hashing contents to match
     /// digest, and checking that the previous checkpoint digest matches.
+    #[graphql(complexity = 0)]
     async fn digest(&self) -> Result<String> {
         Ok(self.digest_impl().extend()?.base58_encode())
     }
 
     /// This checkpoint's position in the total order of finalized checkpoints,
     /// agreed upon by consensus.
+    #[graphql(complexity = 0)]
     async fn sequence_number(&self) -> UInt53 {
         self.sequence_number_impl().into()
     }
@@ -105,17 +108,20 @@ impl Checkpoint {
     /// The timestamp at which the checkpoint is agreed to have happened
     /// according to consensus. Transactions that access time in this
     /// checkpoint will observe this timestamp.
+    #[graphql(complexity = 0)]
     async fn timestamp(&self) -> Result<DateTime> {
         DateTime::from_ms(self.stored.timestamp_ms).extend()
     }
 
     /// This is an aggregation of signatures from a quorum of validators for the
     /// checkpoint proposal.
+    #[graphql(complexity = 0)]
     async fn validator_signatures(&self) -> Base64 {
         Base64::from(&self.stored.validator_signature)
     }
 
     /// The digest of the checkpoint at the previous sequence number.
+    #[graphql(complexity = 0)]
     async fn previous_checkpoint_digest(&self) -> Option<String> {
         self.stored
             .previous_checkpoint_digest
@@ -125,6 +131,7 @@ impl Checkpoint {
 
     /// The total number of transaction blocks in the network by the end of this
     /// checkpoint.
+    #[graphql(complexity = 0)]
     async fn network_total_transactions(&self) -> Option<UInt53> {
         Some(self.network_total_transactions_impl().into())
     }
@@ -133,6 +140,7 @@ impl Checkpoint {
     /// storage fee accumulated during this epoch, up to and including this
     /// checkpoint. These values increase monotonically across checkpoints
     /// in the same epoch, and reset on epoch boundaries.
+    #[graphql(complexity = 0)]
     async fn rolling_gas_summary(&self) -> Option<GasCostSummary> {
         Some(GasCostSummary {
             computation_cost: self.stored.computation_cost as u64,
@@ -144,6 +152,7 @@ impl Checkpoint {
     }
 
     /// The epoch this checkpoint is part of.
+    #[graphql(complexity = 0)]
     async fn epoch(&self, ctx: &Context<'_>) -> Result<Option<Epoch>> {
         Epoch::query(
             ctx,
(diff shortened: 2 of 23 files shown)


crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -599,6 +649,56 @@ impl Loader<DigestKey> for Db {
...
+            results.insert(
+                *key,
+                TransactionBlock {
+                    inner: TransactionBlockInner::try_from(stored.clone())?,
+                    checkpoint_viewed_at,
+                },
+            );
+        }
+
+        Ok(results)
```
sergiupopescu199: nit: we could use iterators.

```suggestion
    keys.into_iter()
        .filter_map(|key| {
            let lookup_key = key.tx_sequence_number as i64;
            seq_num_to_tx.get(&lookup_key).map(|stored| (key, stored))
        })
        .map(|(key, stored)| {
            let checkpoint_viewed_at =
            if key.checkpoint_viewed_at < stored.checkpoint_sequence_number as u64 {
                // Disable usage as a cursor, as the checkpoint is not in the available range
                UNAVAILABLE_CHECKPOINT_SEQUENCE_NUMBER
            } else {
                key.checkpoint_viewed_at
            };

            let inner = TransactionBlockInner::try_from(stored.clone())?;

            Ok((
               *key,
               TransactionBlock {
                  inner,
                  checkpoint_viewed_at,
               },
              ))
        })
       .collect::<Result<HashMap<SeqKey, TransactionBlock>, _>>()
```

sergiupopescu199 (approved):
lgtm ✨

Left an optional nit.

crates/iota-graphql-rpc/src/server/builder.rs
```diff
@@ -537,6 +537,7 @@ impl ServerBuilder {
 
 fn schema_builder() -> SchemaBuilder<Query, Mutation, Subscription> {
     async_graphql::Schema::build(Query, Mutation, Subscription)
+        .limit_complexity(6000)
```
tomxey: why 6000? maybe this should be a const? or configurable parameter?

crates/iota-graphql-rpc/src/types/event/mod.rs
```diff
@@ -64,11 +84,28 @@ type Query<ST, GB> = data::Query<ST, events::table, GB>;
...
+
+        TransactionBlock::query(ctx, key.into()).await.extend()
+    }
+
     /// The Move module containing some function that when called by
     /// a programmable transaction block (PTB) emitted this event.
     /// For example, if a PTB invokes A::m1::foo, which internally
     /// calls A::m2::emit_event to emit an event,
     /// the sending module would be A::m1.
+    #[graphql(complexity = 0)]
```
tomxey: why 0, and not 1, or the default? (here and in other places)

this one is doing some DB load after all, so it's not free.

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
tomxey: nit: it can be looked up by digest or seq num

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
kodemartin: We could indeed. But this is a deliberate choice as I tend to weigh readability more when it comes to complex map operations such as this one. You can immediately see that by refactoring we end up with more lines, additional levels of nesting, and implicit code through closures.

To counterbalance for the performance implications I reserved the space in advance to avoid reallocations: https://github.com/iotaledger/iota/pull/9616/commits/b96f289f2a17e957743aadfda26347951497f604

crates/iota-graphql-rpc/src/server/builder.rs
kodemartin: This should be a const. Refactored with https://github.com/iotaledger/iota/pull/9616/commits/a5e25b452bbf24b350e3d03c530673202f364b95

6000 was just an arbitrary number that was more than `50^2` (the complexity threshold in the path we seek to limit assuming max page size) and `20^3` (the recursive query path assuming default page size that we want to abort).

Reduced to 3000.

crates/iota-graphql-rpc/src/types/event/mod.rs
kodemartin: Good question. The default is 1, but this makes it hard to tune the complexity limit as calculation relies on `page_size * child_complexity`. If we don't eliminate non-critical paths `child_complexity` might be increased also on non-critical paths. Consider this example
```
{
a  {  # (page_size * child_complexity)
    b {  # (page_size * child_complexity) 
        c # 1
        d # 1
        ...
        i   # 1
    } 
   }
```
Assuming the critical path is `a -> b > d`, all other selected fields weigh equally to the `child_complexity` of `b` .

So setting non-critical paths to 0 essentially highlights the critical path, simplifying higher-level tuning.

And it can be considered as a fallback to the previous configuration that had no complexity limit.

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
kodemartin: update: https://github.com/iotaledger/iota/pull/9616/commits/75a69df019d153de1c720198f83eb45a90097e2c

crates/iota-graphql-rpc/src/types/event/mod.rs
tomxey: Hmm, ok. Do you now if `complexity = 0` means that accessing the field is free, but you still pay the complexity of the returned type, or it means that the field is completely free, no matter what data it returns and how deep the query is on that path?

The [documentation](https://async-graphql.github.io/async-graphql/en/depth_and_complexity.html) is quite vague on this.

If it means that the field and all it's children are totally free the setting complexity to 0 for some fields may open up possibilities for exploitation.

e.g. (knowing that `dependencies` is a free field, you can get lots of txs for a transaction that has a lot of dependencies, and the from then go to events, or to deps again, etc..)

```graphql
{
  epoch(id: 4){
    epochId
    transactionBlocks (first:2) {
      edges {
        node {
          digest
          effects {
            dependencies {
              edges {
                node {
                  digest
                  effects {
                    dependencies {
                      edges {
                        node {
                          digest
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

kodemartin: On that level of checks it is free, as it was before. When I say on that level I mean no the library-native checks. Complexity is calculated on the query validation level.

But it seems that I simply missed to identify the `dependencies` path as critical.

kodemartin: fixed with 108ff27bcd

crates/iota-graphql-rpc/src/types/transaction_block_effects.rs
```diff
@@ -198,6 +203,7 @@ impl TransactionBlockEffects {
     }
 
     /// Transactions whose outputs this transaction depends upon.
+    #[graphql(complexity = "child_complexity")]
```
tomxey, line 206: Since this returns many paginated transactions in general, have you considered using the formula:

```rust
first.or(last).unwrap_or(DEFAULT_PAGE_SIZE as u64) as usize * child_complexity
```

?

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -141,23 +141,69 @@ pub(crate) struct TransactionBlockCursor {
...
     /// chain.
+    #[graphql(complexity = 0)]
     async fn digest(&self) -> Option<String> {
         self.native_signed_data()
             .map(|s| Base58::encode(s.digest()))
     }
 
     /// The address corresponding to the public key that signed this
     /// transaction. System transactions do not have senders.
+    #[graphql(complexity = 0)]
```
tomxey: I think any path that can eventually reach event or tx can be a critical path,
e.g.

```graphql
{
  transactionBlocks {
    edges {
      node {
        digest
        sender {
          transactionBlocks {
            edges {
              node {
                digest
              }
            }
          }
        }
      }
    }
  }
}
```

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -173,6 +219,7 @@ impl TransactionBlock {
     ///
     /// If the owner of the gas object(s) is not the same as the sender, the
     /// transaction block is a sponsored transaction block.
+    #[graphql(complexity = 0)]
     async fn gas_input(&self, ctx: &Context<'_>) -> Option<GasInput> {
```
tomxey, line 223: This contains sponsor address, and from the address user can query transactions blocks again.

```graphql
{
  transactionBlocks {
    edges {
      node {
        digest
        gasInput {
          gasSponsor {
            transactionBlocks {
              edges {
                node {
                  digest
                }
              }
            }
          }
        }
      }
    }
  }
}
```

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -196,6 +243,7 @@ impl TransactionBlock {
 
     /// The type of this transaction as well as the commands and/or parameters
     /// comprising the transaction of this kind.
+    #[graphql(complexity = 0)]
```
tomxey: e.g. if kind is genesis tx, then user can query events of this tx, and from there query txs, and so on

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -224,6 +274,7 @@ impl TransactionBlock {
     /// reference that sets a deadline after which validators will no longer
     /// consider the transaction valid. By default, there is no deadline for
     /// when a transaction must execute.
+    #[graphql(complexity = 0)]
```
tomxey: this returns epoch, and from epoch user can query transaction blocks

crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -144,6 +152,7 @@ impl Checkpoint {
     }
 
     /// The epoch this checkpoint is part of.
+    #[graphql(complexity = 0)]
```
tomxey: as above, from epoch user can query txs

crates/iota-graphql-rpc/src/types/transaction_block_effects.rs
```diff
@@ -262,12 +268,14 @@ impl TransactionBlockEffects {
     }
 
     /// Effects to the gas object.
+    #[graphql(complexity = 0)]
```
tomxey: GasEffects contains object, which then links to a tx, and so on

crates/iota-graphql-rpc/src/types/transaction_block_effects.rs
kodemartin: Yes, relying only `child_complexity` removes the constants but still maintains the possibility of exponential growth. This makes it a bit easier to tune the complexity limit.

kodemartin (commented):
@tomxey  it becomes apparent that tuning the complexity limit becomes tedious this way.

`complexity = 0` was an easy way to direct the algorithm on critical paths, but with so many recursive paths, it is very hard to maintain, and set a proper complexity limit that will ban the recursive paths only.

So, apart from the highest-level critical paths that are currently configured in a refined way (i.e. dropping non-critical branches), I will customize the complexity for all the `transaction_block` paths and let the rest weigh in with default complexity. This will lead to stricter limits for users for the moment.

The robust way to handle this would be to identify recursive paths while parsing and validating the query. 

At any case, the patch is still more robust than the upstream version, and we do have a fallback check on complexity and output_nodes through the limit-checker extension.

tomxey (approved):

PR #9657, opened 2025-12-17
upstream(iota): specific error message rather than OS message when file not found in client ptb

thibault-martinez:
# Description of change

This PR fixes the `iota client ptb upgrade/publish` to check if the package path exists and throw an error rather than an IO error.

Upstream PR: https://github.com/MystenLabs/sui/pull/24345


crates/iota/src/client_ptb/builder.rs
```diff
@@ -997,6 +1005,20 @@ impl<'a> PTBBuilder<'a> {
...
+                if !package_path.exists() {
+                    error!(
+                        path_loc,
+                        "Package path '{}' does not exist",
+                        package_path.display()
+                    );
+                }
+
+                let package_path = package_path
+                    .canonicalize()
```
thibault-martinez, line 1019: Canonicalize also checks for existence but it's maybe still better to check independently because it could fail for another reason

Thoralf-M: Wouldn't the error from canonicalize say the different reasons?

thibault-martinez: I think it would differentiate both, but the inexistence would still say "No such file or directory" which is what we are trying to remove here. I will try it and report.

Alex6323 (approved):

crates/iota/src/client_ptb/builder.rs
thibault-martinez: Before
```sh
thibault@Mac ~/i/iota (chore/better-non-existent-package-error) [1]> ./target/debug/iota client ptb --upgrade hello 0xcc7df6a2cd2a0307217062742d003090df7cfe1f93db4403f8538fee0e193a09
Encountered error when building PTB:
  × Error when processing PTB
   ╭────
 1 │ --upgrade hello 0xcc7df6a2cd2a0307217062742d003090df7cfe1f93db4403f8538fee0e193a09 
   · ───────┬───────
   ·        ╰── Failed to canonicalize package path: No such file or directory (os error 2)
   ╰────

Could not build PTB due to previous error
```

After
```sh
thibault@Mac ~/i/iota (chore/better-non-existent-package-error)> ./target/debug/iota client ptb --upgrade hello 0xcc7df6a2cd2a0307217062742d003090df7cfe1f93db4403f8538fee0e193a09
Encountered error when building PTB:
  × Error when processing PTB
   ╭────
 1 │ --upgrade hello 0xcc7df6a2cd2a0307217062742d003090df7cfe1f93db4403f8538fee0e193a09 
   · ───────┬───────
   ·        ╰── Package path 'hello' does not exist
   ╰────

Could not build PTB due to previous error
```

Thoralf-M: Makes sense, with the package path in the error message it's much better :+1:

Thoralf-M (approved):

Issue #8840, opened 2025-10-13
[indexer]: refine indexer metrics


kodemartin:
## Description

`iota-indexer` has a single metrics set that includes measures across various scopes (e.g. analytics, ingestion, optimistic indexing, database commit etc.).

We would like to redefine the metrics collected according to the logical scope, so that downstream libraries/apps that we maintain have more flexibility in using selected metrics only.

PR #9658, opened 2025-12-17
feat(iota-indexer): add cache in `HttpRestKVClient`

### LINKED ISSUE #9501: feat(indexer): Cache critical Rest KV queries
## Description

This is motivated by the redundancy while paging events requested from bigtable. In particular we fetch all events and then create the page

https://github.com/iotaledger/iota/blob/2cc658c693879b8d19c66f50935bf2d0d5d50c40/crates/iota-indexer/src/historical_fallback/reader.rs#L487

We would like to minimize requests to the BigTable data.

A simplified architecture is as follows
```
bigtable <-> kv-rest <-> indexer-fallback
```

### Tentative Requirements 

* Identify the critical paths
   * Is every request worthwhile caching?
   * An hierarchy of entities and incurring costs would thus be helpful to define.
* Cache on the `HttpKVClient`  on the indexer
* Cache raw data, and let transformations repeat
* Cap memory usage for the cache (LRU, or expiration to be decided during implementation)

sergiupopescu199:
# Description of change

The following PR adds a caching layer to the `HttpRestKVClient`. The implementation is very similar to the one already implemented for the `HttpKVStore`.

fixes #9501 

Started the indexer reader and checked if the requests would be cached.

> [!NOTE]
> This patch does not affect the normal operation of the indexer. By default pruning is not enabled and also the fallback feature is enabled only if an additional CLI argument is provided during the startup. No further tests were conducted.

### DIFF
--- a/crates/iota-indexer/src/historical_fallback/client.rs
+++ b/crates/iota-indexer/src/historical_fallback/client.rs
@@ -3,8 +3,13 @@
 
 //! Module containing the client for interacting with the REST API KV server.
 
+use std::time::Duration;
+
 use bytes::Bytes;
-use futures::stream::{self, StreamExt};
+use futures::{
+    TryStreamExt, future,
+    stream::{self, StreamExt},
+};
 use iota_storage::http_key_value_store::{ItemType, Key};
 use iota_types::{
     base_types::{ObjectID, SequenceNumber},
@@ -16,6 +21,7 @@ use iota_types::{
     object::Object,
     transaction::Transaction,
 };
+use moka::sync::{Cache as MokaCache, CacheBuilder as MokaCacheBuilder};
 use reqwest::{
     Client, Url,
     header::{CONTENT_LENGTH, HeaderValue},
@@ -24,8 +30,12 @@ use serde::{Deserialize, Serialize};
 use tap::TapFallible;
 use tracing::{error, info, instrument, trace, warn};
 
-use crate::{IndexerError, errors::IndexerResult};
+use crate::{
+    IndexerError, errors::IndexerResult,
+    historical_fallback::metrics::HistoricalFallbackClientMetrics,
+};
 
+const CACHE_TIME_TO_IDLE: Duration = Duration::from_secs(600);
 /// Request payload for multi_get containing list of keys.
 #[derive(Serialize, Debug)]
 struct MultiGetRequest {
@@ -124,13 +134,17 @@ pub(crate) struct HttpRestKVClient {
     batch_size: usize,
     /// Maximum number of concurrent batch requests
     max_concurrent_batches: usize,
+    cache: MokaCache<Url, Bytes>,
+    metrics: HistoricalFallbackClientMetrics,
 }
 
 impl HttpRestKVClient {
     pub fn new(
         base_url: &str,
+        cache_size: u64,
         batch_size: usize,
         max_concurrent_batches: usize,
+        metrics: HistoricalFallbackClientMetrics,
     ) -> IndexerResult<Self> {
         info!(
             "creating HttpRestKVClient with base_url: {base_url}, batch_size: {batch_size}, max_concurrent_batches: {max_concurrent_batches}",
@@ -146,11 +160,17 @@ impl HttpRestKVClient {
 
         let base_url = Url::parse(&base_url)?;
 
+        let cache = MokaCacheBuilder::new(cache_size)
+            .time_to_idle(CACHE_TIME_TO_IDLE)
+            .build();
+
         Ok(Self {
             base_url,
             client,
             batch_size,
             max_concurrent_batches,
+            metrics,
+            cache,
         })
     }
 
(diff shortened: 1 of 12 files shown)


crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -148,45 +163,86 @@ impl HttpRestKVClient {
...
+            })?;
+
+            // process fetched results: for each missing key that was successfully fetched,
+            // update the cache with the new data and populate the corresponding slot in
+            // results at original index position.
+            for (fetch_result, (key, index)) in fetched_results.into_iter().zip(missing.into_iter())
+            {
+                if let Some(bytes) = fetch_result {
+                    let cache_key = self.get_url(&key)?;
+                    self.cache.insert(cache_key, bytes.clone());
```
tomxey: should we also check for empty bytes here, like we do in `fetch`?

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -148,45 +163,86 @@ impl HttpRestKVClient {
...
-            .collect();
-        let payload = MultiGetRequest { keys: encoded_keys };
-
-        trace!(
-            "fetching batch of {} keys from url: {url}",
-            payload.keys.len()
-        );
-
-        let resp = self.client.post(url.clone()).json(&payload).send().await?;
+        // pre-allocate results with None. Cache hits will replace None with
```
tomxey, line 179: do you think we could move the whole caching logic one level higher, to the `multi_fetch` function?
then we could create batches out of `missing` rows.
with current approach it's possible that we will make requests that are not fully utilizing the batch size

would also be nice to separate caching logic from batching logic so I would suggest the current `multi_fetch` function to be renamed to e.g. `multi_fetch_from_kv`, then the new `multi_fetch` function could look like:

```rust
let results, missing = ...;
let fetched_results = multi_fetch_from_kv(missing);
<fill_cache_and_results>
return results;
```

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -108,11 +117,17 @@ impl HttpRestKVClient {
 
         let base_url = Url::parse(&base_url)?;
 
+        let cache = MokaCacheBuilder::new(cache_size)
+            .time_to_idle(CACHE_TIME_TO_IDLE)
```
tomxey, line 161: have you considered using LRU approach for removing cache entries only when the cache is full?

At any case, I think it would be good to set some [max_capacity](https://docs.rs/moka/latest/moka/sync/struct.CacheBuilder.html#method.max_capacity) regardless of whether we will use the `time_to_idle` or not.

sergiupopescu199: 1. In the moka docs by default it uses an [TinyLFU](https://docs.rs/moka/latest/moka/index.html#tinylfu) which is a combination of LRU & LFU policy when we reach the max capacity if time to idle is not expired. We can explicitly opt for a pure LRU policy. This default policy though seems quite good and I'm keen to leave it for now.

1. `MokaCacheBuilder::new(cache_size)` is equivalent to `max_capacity`.

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9658/commits/d5de4a64f1c43e7974a344709a7d3ed8524860c0

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9658/commits/d5de4a64f1c43e7974a344709a7d3ed8524860c0

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -188,53 +170,29 @@ impl HttpRestKVClient {
...
-                .iter()
-                .map(|(key, _)| {
-                    let (_, encoded) = key.to_path_elements();
-                    encoded
+            let chunks = missing
+                .chunks(self.batch_size)
+                .map(|chunk| chunk.iter().map(|(key, _)| *key).collect())
+                .collect::<Vec<Vec<Key>>>();
+
+            let chunks_len = chunks.iter().map(|chunk| chunk.len()).sum();
```
tomxey: is it the same as `missing.len()`?

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9658/commits/d8d2e2580500f8f608e14e36c0e0800b6ad28b0d

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -64,14 +65,18 @@ pub(crate) struct HistoricalFallbackReader {
 impl HistoricalFallbackReader {
     pub fn new(
         rest_kv_url: &str,
+        cache_size: u64,
         package_resolver: PackageResolver,
         fallback_kv_multi_fetch_batch_size: usize,
         fallback_kv_concurrent_fetches: usize,
+        metrics: HistoricalFallbackClientMetrics,
```
kodemartin: I think it's more flexible to accept the `Registry` and construct the scope-specific metrics. This way this wlll be constructed lazily and will not affect higher level components.

Also considering #8840

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -127,21 +144,60 @@ impl HttpRestKVClient {
...
+        // Some(value), and cache misses will remain None until we fetch them
+        // from the REST API.
+        let mut results = vec![None; keys.len()];
+        // track keys that missed the cache, preserving their original positions.
+        // Each entry is a tuple of (key, original_index) so we can merge fetched data
+        // back into the correct position in the results vector.
+        let mut missing = Vec::with_capacity(keys.len());
+
+        for (index, key) in keys.iter().enumerate() {
+            let cache_key = self.get_url(key)?;
```
kodemartin: Why do we need the entire url for the cache key, and not the `key` itself, or a representation of?

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -127,21 +144,60 @@ impl HttpRestKVClient {
...
+                results[index] = Some(bytes);
+            } else {
+                self.metrics.record_cache_miss(key.item_type());
+                missing.push((*key, index));
+            }
+        }
 
-        let mut flattened = Vec::new();
-        while let Some(batch_result) = results.next().await {
-            flattened.extend(batch_result?);
```
kodemartin, line 173: Nit: I personally consider the previous version more readable, as `try_fold` adds an extra nesting layer for the side effects. Plus the error handling is also more transparent in the previous case.

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -127,21 +144,60 @@ impl HttpRestKVClient {
...
+        if !missing.is_empty() {
```
kodemartin: Nit:


```suggestion
        if missing.is_empty() {
            return Ok(results)
        }
        <logic for cache misses>
```
would spare a nesting level

kodemartin (commented):
Looks good with minor suggestions and nits

tomxey (approved):
lgtm
also note that single key fetch method was removed on the base branch

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9658/commits/c7ed7b94e29862c95ec612fbaaea30ebf66add96

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: Oh interesting, I though that it would've been more polished by reducing some lines of code. But the nesting layer is a good thing to keep in mind.

fixed: https://github.com/iotaledger/iota/pull/9658/commits/c7ed7b94e29862c95ec612fbaaea30ebf66add96

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: This decision was inherited form the node team implementation. I've extended the `Key`'s derive macro to also include the `Hash` and other most common used traits. Now the cache uses `Key` directly.

fixed: https://github.com/iotaledger/iota/pull/9658/commits/c7ed7b94e29862c95ec612fbaaea30ebf66add96

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9658/commits/c7ed7b94e29862c95ec612fbaaea30ebf66add96

muXxer (approved):

kodemartin (approved):
lgtm

PR #9716, opened 2026-01-07
feat(consensus): reinitialize consensus components after fast commit sync work

### LINKED ISSUE #9715: Reinitialize the consensus compoenents after FastCommitSyncer fetched data

polinikita:
# Description of change

PR focuses on making fast sync safe to hand back to normal consensus by reinitializing the state and all core components; it also refactors commit syncer scheduling/verification and adds a restart divergence test.

Breakdown of changes:

- Adds close‑to‑quorum fast sync flow that fetches a header window and reinitializes consensus components once catch‑up completes.
 - Persists and consumes reputation scores/commit info to rebuild leader schedule and scoring state after fast sync, plus DagState cache reinit logic.
- Refactors commit syncers with shared scheduling helpers and header verification to tighten fetch correctness.
- Moves GC/flush ordering in commit observer and adds reinitialize path that rebuilds linearizer state during recovery.
- Expands restart node tests to ensure no divergence happens after fast sync and two crashes.

Important assumption that should be addressed in the following PRs:
- The node can operate mostly normally if the fast sync is completed. It will fetch enough headers to reinitialize the state after a crash.
- If the node db is corrupted, while the consensus db is not, then it will not be possible to replay the recovery logic since it will attempt to reconstruct all PendingSubDAG which require existence of all headers. All headers are not fetched during fast sync. In such a case, one needs to remove both node and consensus dbs.
- Core thread ignores most commands while fast sync is ongoing; normal proposal/processing is effectively paused until reinitialization happens
 - Leader schedule update from commit scores assumes scores correspond to the window ending at commit_index - 1; any change to how scores are embedded must preserve this.

By benchmarking on the local deployment we observe a significant improvement in the speed of syncing. Specifically, during 1 second 60-120 seconds of committed subDAGs could be replayed.

Fixes #9715 
 
### DIFF
--- a/crates/starfish/core/src/commit_syncer/fast.rs
+++ b/crates/starfish/core/src/commit_syncer/fast.rs
@@ -22,6 +22,7 @@ use crate::{
     commit::{CommitAPI as _, CommitRange, CommittedSubDag, GENESIS_COMMIT_INDEX, TrustedCommit},
     commit_syncer::{
         CommitSyncType, CommitSyncerHandle, Inner, fetch_loop as shared_fetch_loop,
+        handle_fetch_join_error, requeue_partial_range, schedule_commit_ranges,
         try_start_fetches as shared_try_start_fetches, verify_fetched_headers,
         verify_transactions_with_transactions_refs,
     },
@@ -124,14 +125,15 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
                 }
                 // Handles results from fetch tasks.
                 Some(result) = self.inflight_fetches.join_next(), if !self.inflight_fetches.is_empty() => {
-                    if let Err(e) = result {
+                    if let Err(ref e) = result {
                         if e.is_panic() {
-                            std::panic::resume_unwind(e.into_panic());
+                            std::panic::resume_unwind(result.unwrap_err().into_panic());
                         }
-                        warn!("[{}] Fetch cancelled. FastCommitSyncer shutting down: {}", self.inner.sync_type.as_str(), e);
-                        // If any fetch is cancelled or panicked, try to shutdown and exit the loop.
-                        self.inflight_fetches.shutdown().await;
-                        return;
+                        if handle_fetch_join_error(e, &self.inner.sync_type) {
+                            // If any fetch is cancelled or panicked, try to shutdown and exit the loop.
+                            self.inflight_fetches.shutdown().await;
+                            return;
+                        }
                     }
                     let (target_end, commits, committed_subdags) = result.unwrap();
                     self.handle_fetch_result(target_end, commits, committed_subdags).await;
@@ -208,7 +210,7 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
         let local_commit_index = self.inner.dag_state.read().last_commit_index();
         let highest_handled_index = self.inner.commit_consumer_monitor.highest_handled_commit();
         let highest_scheduled_index = self.highest_scheduled_index.unwrap_or(0);
-        let unhandled_commits_threshold = self.unhandled_commits_threshold();
+        let unhandled_commits_threshold = self.inner.unhandled_commits_threshold();
         let step = self
             .inner
             .sync_type
@@ -238,11 +240,9 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
             let fetch_after_index = self
                 .synced_commit_index
                 .max(self.highest_scheduled_index.unwrap_or(0));
-            // When the node is falling behind, schedule pending fetches which will be
-            // executed on later.
 
             debug!(
-                "[{}] Checking to schedule fetches: synced_commit_index={}, highest_handled_index={}, highest_scheduled_index={}, quorum_commit_index={}, unhandled_commits_threshold={}, fetch_after_index={}, step={}",
+                "[{}] Checking to schedule fetches: synced_commit_index={}, highest_handled_index={}, highest_scheduled_index={}, quorum_commit_index={}, unhandled_commits_threshold={}, fetch_after_index={}",
                 self.inner.sync_type.as_str(),
                 self.synced_commit_index,
                 highest_handled_index,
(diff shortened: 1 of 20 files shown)


crates/starfish/core/src/leader_schedule.rs
```diff
@@ -57,9 +60,15 @@ impl LeaderSchedule {
         let leader_swap_table = dag_state.read().recover_last_commit_info().map_or(
             LeaderSwapTable::default(),
             |(last_commit_ref, last_commit_info)| {
+                let range_end = last_commit_info.reputation_scores.commit_range.end();
+                let seed_index = if range_end == GENESIS_COMMIT_INDEX {
+                    last_commit_ref.index
+                } else {
+                    range_end
+                };
```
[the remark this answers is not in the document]
polinikita: addressed

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -373,6 +610,204 @@ impl<C: NetworkClient> CommitSyncer<C> for FastCommitSyncer<C> {
...
+
+        // Get block refs from recent commits stored during fast sync
+        // TODO: The commits might not yet stored, but only fetched and pending
+        // processing.
+        let block_refs = inner
+            .dag_state
+            .read()
+            .get_block_refs_for_recent_commits(num_commits);
+
+        // TODO: we anyway need to reinitialize even if there are no block refs,
```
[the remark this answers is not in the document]
polinikita: changed

crates/starfish/core/src/dag_state.rs
```diff
@@ -316,6 +289,123 @@ impl DagState {
...
+        let commit_recovery_start_index = self
+            .recover_last_commit_info()
+            .map(|(commit_ref, commit_info)| {
+                let range_end = commit_info.reputation_scores.commit_range.end();
+                if range_end == GENESIS_COMMIT_INDEX {
+                    commit_ref.index.saturating_add(1)
+                } else {
+                    range_end.saturating_add(1)
+                }
+            })
```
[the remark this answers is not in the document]
polinikita: consolidated

crates/starfish/core/src/leader_schedule.rs
```diff
@@ -787,8 +930,8 @@ mod tests {
         let unscored_subdags = vec![
             CommittedSubDag::new(
                 leader_ref,
-                blocks,
-                last_commit.block_headers().to_vec(),
+                blocks.clone(),
+                blocks.iter().map(|b| b.reference()).collect::<Vec<_>>(),
```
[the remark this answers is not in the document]
polinikita: This is fine here

crates/starfish/core/src/leader_schedule.rs
```diff
@@ -177,6 +233,84 @@ impl LeaderSchedule {
...
+
+        // Determine the commit range for these scores.
+        // Reputation scores are attached to the *first* commit after a schedule
+        // update, so the scores correspond to the previous window ending at
+        // commit_index - 1.
+        let range_end = commit_index.saturating_sub(1);
+        if range_end == GENESIS_COMMIT_INDEX {
+            return;
+        }
+        let range_start = range_end.saturating_sub(self.num_commits_per_schedule as u32 - 1);
```
VorobyevIlya, line 261: Why do we need -1? Then the range size is self.num_commits_per_schedule  - 1

polinikita: the interval [range_start, range_end] contains exactly self.num_commits_per_schedule indices.

VorobyevIlya (approved):

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -83,58 +99,240 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
...
+        // Skip scheduling depending on sync type and gap threshold.
+        let gap = quorum_commit_index.saturating_sub(dag_state_commit_index);
+        let should_schedule = self.inner.sync_type.should_schedule(
+            gap,
+            self.inner.context.parameters.commit_sync_gap_threshold,
+            self.inner
+                .context
+                .protocol_config
+                .consensus_transaction_ref(),
+        ) || self.has_fetched_data;
```
piotrm50: I would reverse this condition so that `self.has_fetched_data` evaluates first so that we skip the `should_schedule` function call if the first condition is true.

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -52,6 +61,13 @@ pub(crate) struct FastCommitSyncer<C: NetworkClient> {
     // The commit index that is the max of highest local commit index and commit index inflight to
     // Core. Used to determine if fetched blocks can be sent to Core without gaps.
     synced_commit_index: CommitIndex,
+    // Whether the syncer is in "close to quorum" mode, meaning remaining gap < batch size.
+    // When this is true, the syncer will fetch block headers and transactions for cached rounds
+    // before completing fast sync.
+    close_to_quorum_mode: bool,
+    // Whether the fast syncer has actually fetched any data. Close-to-quorum mode only
+    // activates after this is true. Reset to false after reinitialization completes.
+    has_fetched_data: bool,
```
piotrm50, line 70: Do I understand correctly that this is needed to avoid unnecessary reinitialization in the case when no data has been fetched?

crates/starfish/core/src/commit_syncer/fast.rs
piotrm50: Also, why `self.has_fetched_data==true` is enough to schedule ranges? Is it because the fast commit syncer is doing something so we might as well schedule some more ranges?

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -83,58 +99,240 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
...
+                let range_start = current_fetch_after + 1;
+                let range_end = quorum_commit_index;
+                debug!(
+                    "[{}] Scheduling final partial fetch for commit range {}..={} (remaining_gap={})",
+                    self.inner.sync_type.as_str(),
+                    range_start,
+                    range_end,
+                    remaining_gap
+                );
+                self.pending_fetches
```
piotrm50, line 309: Aren't we adding the last range twice? First we get it from `schedule_commit_ranges` and then here the second time?

crates/starfish/core/src/leader_schedule.rs
```diff
@@ -787,8 +924,8 @@ mod tests {
         let unscored_subdags = vec![
             CommittedSubDag::new(
                 leader_ref,
-                blocks,
-                last_commit.block_headers().to_vec(),
+                blocks.clone(),
+                blocks.iter().map(|b| b.reference()).collect::<Vec<_>>(),
```
piotrm50: Is this change necessary?

crates/starfish/core/src/core.rs
```diff
@@ -510,6 +514,7 @@ impl Core {
         self.add_block_headers(block_headers)
     }
 
+    #[allow(dead_code)]
```
piotrm50: is it still dead code? perhaps we should change to `#[expect(dead_code)]` to remember to remove it when it's used.

piotrm50 (commented):
left some small questions

crates/starfish/core/src/commit_syncer/fast.rs
polinikita: yes, exactly.
The order will be changed in the next commit

crates/starfish/core/src/commit_syncer/fast.rs
polinikita: Correct. Only when we fetched at least one full range of 1000 commits

crates/starfish/core/src/commit_syncer/fast.rs
polinikita: No, here we add only a small gap range, which is less than one step (1000 commits). in function schedule_commit_ranges we include only full ranges of 1000 commits. We also take into account what we has scheduled (see above).

crates/starfish/core/src/core.rs
polinikita: Good note! I remove it in the next commit

crates/starfish/core/src/leader_schedule.rs
polinikita: This is not needed. Changed along with other changes when had a problem with a wrong leader scheduler

piotrm50 (approved):

PR #9719, opened 2026-01-07
feat(indexer): Make `persist_sequentially` more resilient to hanging

### LINKED ISSUE #5873: [indexer]: Object snapshot processing gets stuck after buffer gets full
# Problem

In `CommonHandler`, we have a check to ensure that the buffer will not grow indefinitely: https://github.com/iotaledger/iota/blob/3a7daeb263e78bdc1af9520cd0fd7ae68ff1902f/crates/iota-indexer/src/handlers/mod.rs#L93

This can lead to cases where we stop accepting new items from the stream while not having the `next_cp_to_process` in the buffer: https://github.com/iotaledger/iota/blob/3a7daeb263e78bdc1af9520cd0fd7ae68ff1902f/crates/iota-indexer/src/handlers/mod.rs#L116

Which freezes the processing.

# Solution

The check `if unprocessed.len() >= UNPROCESSED_CHECKPOINT_SIZE_LIMIT {` should be relaxed, to see if we have the next needed checkpoint in buffer, if not we should keep reading until it's there.

Also, in case we decide to skip reading from stream, it would be nice to `tokio::time::sleep` for a short while (0.5s?), otherwise the loop starts to output a lot of `"Unprocessed checkpoint size reached limit {UNPROCESSED_CHECKPOINT_SIZE_LIMIT}, skip reading from stream..."`.

Also, the log `Unprocessed checkpoint size reached limit...` could be enhanced with more information, like min/max checkpoint number in buffer, and `next_cp_to_process`. To ease future debugging if similar issue would occur.

Also, the `UNPROCESSED_CHECKPOINT_SIZE_LIMIT` could be made configurable using env var, with default value of 1000. To have a chance of adjusting this value in case the buffer would happen to be too small.

tomxey:
# Description of change

Fixes `persist_sequentially` function that could potentially hang if buffer of checkpoints would be full, but we would be missing the next checkpoint to be processed.

fixes https://github.com/iotaledger/iota/issues/5873

### DIFF
--- a/crates/iota-indexer/src/ingestion/common/persist.rs
+++ b/crates/iota-indexer/src/ingestion/common/persist.rs
@@ -4,7 +4,7 @@
 //! Types and associated logic to use while persisting
 //! data to the database.
 
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, time::Duration};
 
 use async_trait::async_trait;
 use futures::{FutureExt, StreamExt};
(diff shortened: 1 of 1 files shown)


sergiupopescu199 (approved):
lgtm ✨

kodemartin (approved):
lgtm

Issue #9620, opened 2025-12-15
feature(gRPC): add client functions


muXxer:
We need to add client convenience functions. The clients shouldn't need to bother with the gRPC proto types.
For example in "get_transactions", the client should only pass the `iota-sdk-types::Digest`, and the read mask. Everything else should be hidden in the client.

We should also add functions like "get_transaction", that is a wrapper around "get_transactions" but only with a single digest passed.

All the functions should return the `iota-sdk-types`.

While implementing, we should check if all BCS serialized fields in the gRPC protocol have an enumeration with a version wrapped around the actual type.

muXxer:
Added in #9738

PR #9738, opened 2026-01-12
feat(gRPC): add high-level client API calls

bingyanglin:
# Description of change

Add high-level gRPC API calls, which makes the types of input parameters and responses be all from `iota-sdk-types` for ease of usage.

Fixes https://github.com/iotaledger/iota/issues/9620

### DIFF
--- a/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
+++ b/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
@@ -18,8 +18,8 @@ impl Client {
     /// Get the latest checkpoint.
     ///
     /// Note: If you only need the latest checkpoint sequence number (not the
-    /// full checkpoint data), use [`ResponseExt::checkpoint_height()`] on any
-    /// gRPC response instead.
+    /// full checkpoint data), use [`crate::ResponseExt::checkpoint_height()`]
+    /// on any gRPC response instead.
     ///
     /// # Example
     ///
--- a/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
+++ b/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
@@ -1,0 +1,133 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! High-level API for checkpoint queries.
+
+use iota_grpc_types::{
+    field::FieldMask,
+    v0::ledger_service::{GetCheckpointDataRequest, get_checkpoint_data_request},
+};
+use iota_sdk_types::{CheckpointSequenceNumber, Digest, SignedCheckpointSummary};
+
+use crate::{
+    Client,
+    api::{Result, TryFromProtoError},
+};
+
+impl Client {
+    /// Get the latest checkpoint.
+    ///
+    /// Note: If you only need the latest checkpoint sequence number (not the
+    /// full checkpoint data), use [`ResponseExt::checkpoint_height()`] on any
+    /// gRPC response instead.
+    ///
+    /// [`ResponseExt::checkpoint_height()`]: crate::ResponseExt::checkpoint_height
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// # use iota_grpc_client::Client;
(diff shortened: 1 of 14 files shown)


semenov-vladyslav (approved):

crates/iota-grpc-client/src/api/ledger/transactions.rs
```diff
@@ -0,0 +1,163 @@
...
+                .map(|d| TransactionRequest {
+                    digest: Some((*d).into()),
+                })
+                .collect(),
+        };
+
+        let request = GetTransactionsRequest {
+            requests: Some(requests),
+            read_mask: Some(field_mask_with_default(read_mask, TRANSACTIONS_READ_MASK)),
+            max_message_size_bytes: None,
```
muXxer: Shouldn't that point to the `max_decoding_message_size` of the client?

crates/iota-grpc-client/src/api/ledger/transactions.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// let txs = client
+    ///     .get_transactions(
+    ///         &[digest],
+    ///         Some("transaction.bcs,signatures.bcs,effects.bcs"),
+    ///     )
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn get_transactions(
```
muXxer, line 67: Maybe we could also directly implement streaming methods for such functions (e.g. `get_transactions_stream` and yield the individual results back to the caller, and this function here could just utilize the streaming method directly and collect all results and return them.

crates/iota-grpc-client/src/api/ledger/objects.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// let objs = client.get_objects(&[(object_id, None)], None).await?;
+    ///
+    /// // With reference info (if you also want object metadata)
+    /// let objs = client
+    ///     .get_objects(&[(object_id, None)], Some("bcs,reference"))
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn get_objects(
```
muXxer, line 54: Also add a streaming method here.

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -0,0 +1,131 @@
...
+        let request = GetCheckpointDataRequest {
+            checkpoint_id: Some(checkpoint_id),
+            checkpoint_read_mask: Some(FieldMask {
+                paths: vec!["summary.bcs".to_string()],
+            }),
+            transactions_filter: None,
+            transaction_read_mask: None,
+            events_filter: None,
+            event_read_mask: None,
+            max_message_size_bytes: None,
```
muXxer: Same here, we need the max bytes from the client.

muXxer (changes requested):

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
bingyanglin: Thanks for pointing this out. Addressed this at [eefc435](https://github.com/iotaledger/iota/pull/9738/commits/eefc435b604dae3240583dd3063c4cb82a2719b6).

crates/iota-grpc-client/src/api/ledger/transactions.rs
bingyanglin: Thanks for pointing this out. Addressed this at [eefc435](https://github.com/iotaledger/iota/pull/9738/commits/eefc435b604dae3240583dd3063c4cb82a2719b6).

muXxer (approved):

crates/iota-grpc-client/src/api/ledger/objects.rs
bingyanglin: Good suggestion, as we will have streaming in our server side. Created #9795 to trace it.

crates/iota-grpc-client/src/api/ledger/transactions.rs
bingyanglin: Good suggestion, as we will have streaming in our server side. Created #9795 to trace it.

PR #9775, opened 2026-01-14
refactor(iota-kvstore, iota-rest-kv): optimize bigtable query

### LINKED ISSUE #9510: feat(indexer): optimize fallback support in the indexer
The scope of this Epic is to improve the basic implementation introduced in scope of https://github.com/iotaledger/iota/issues/9002
 
The improvements can be of both categories:
 - functional (e.g. supporting fallback for more queries, e.g. full support for past objects queries)
 - performance (e.g. multiget, caching)

### LINKED ISSUE #9542: feat(rest-kv): Optimize amount of data fetched from bigtable
In iota-rest-kv, when fetching part of transaction, e.g. effects, first the [get_transactions](https://github.com/iotaledger/iota/blob/cf3688143fab9a09ced7ddf7fbfb9e716b4d5b16/crates/iota-rest-kv/src/bigtable.rs#L83) function is called, and then effects are extracted out of the result.

But the `get_transaction` function is getting more data than just the effects (e.g. [events](https://github.com/iotaledger/iota/blob/cf3688143fab9a09ced7ddf7fbfb9e716b4d5b16/crates/iota-kvstore/src/bigtable/client.rs#L157) are fetched), since it is constructing the full Transaction. Most of this data is then later dropped.

We could optimize the way we fetch data from BigTable to only fetch cells with the data we need, to avoid cases where we fetch data, and then not use it and drop it.

sergiupopescu199:
# Description of change

Optimize BigTable queries in `iota-rest-kv` to avoid over-fetching. Previously, when requesting specific data (e.g., a transaction), the query would fetch all columns for that row key (including effects, events, etc.), even though only one column qualifier was needed. This patch filters queries to fetch only the requested column, reducing bandwidth and improving latency.

fixes #9542, #9510 

The tests were done manually by starting the `iota-rest-kv` on devnet, and used the `HttpRestKVClient` to fetch data making sure all worked correctly.

> [!NOTE]

### DIFF
--- a/crates/iota-rest-kv/src/bigtable.rs
+++ b/crates/iota-rest-kv/src/bigtable.rs
@@ -112,7 +112,7 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let keys = digests.iter().map(|tx| tx.inner().to_vec()).collect();
+                let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
                 multi_get_cell(
                     &mut client,
@@ -128,7 +128,7 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let keys = digests.iter().map(|tx| tx.inner().to_vec()).collect();
+                let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
                 multi_get_cell(
                     &mut client,
@@ -146,7 +146,7 @@ impl KvStoreClient {
 
                 let keys = seq_nums
                     .iter()
-                    .map(|sq| sq.to_be_bytes().to_vec())
+                    .map(|sq| Some(sq.to_be_bytes().to_vec()))
                     .collect();
 
                 multi_get_cell(
@@ -165,7 +165,7 @@ impl KvStoreClient {
 
                 let keys = seq_nums
                     .iter()
-                    .map(|sq| sq.to_be_bytes().to_vec())
+                    .map(|sq| Some(sq.to_be_bytes().to_vec()))
                     .collect();
 
                 multi_get_cell(
@@ -184,8 +184,8 @@ impl KvStoreClient {
 
                 let digest_keys = checkpoint_digests
                     .iter()
-                    .map(|digest| digest.inner().to_vec())
-                    .collect::<Vec<Vec<u8>>>();
+                    .map(|digest| Some(digest.inner().to_vec()))
+                    .collect::<Vec<Option<Vec<u8>>>>();
 
                 fetch_checkpoint_summary_by_digests(&mut client, digest_keys).await
             }
@@ -195,7 +195,7 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let keys = digests.iter().map(|tx| tx.inner().to_vec()).collect();
+                let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
                 multi_get_cell(
                     &mut client,
@@ -213,7 +213,10 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let keys = object_keys.iter().map(raw_object_key).collect();
+                let keys = object_keys
+                    .iter()
+                    .map(|key| Some(raw_object_key(key)))
+                    .collect();
 
                 multi_get_cell(&mut client, OBJECTS_TABLE, keys, DEFAULT_COLUMN_QUALIFIER).await
             }
@@ -223,7 +226,7 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let keys = digests.iter().map(|tx| tx.inner().to_vec()).collect();
+                let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
                 multi_get_cell(
                     &mut client,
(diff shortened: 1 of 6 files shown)


crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -101,81 +112,165 @@ impl KvStoreClient {
...
+                        let cell_name = std::str::from_utf8(&name).map_err(anyhow::Error::from)?;
+                        if cell_name == CHECKPOINT_SUMMARY_COLUMN_QUALIFIER {
+                            // map from sequence number back to original digest index
+                            if let Some(&digest_index) = seq_num_to_digest_index.get(&row.key) {
+                                results[digest_index] = Some(Bytes::from(value));
+                            }
+                        } else {
+                            error!("unexpected column {cell_name:?} in checkpoints table")
+                        }
+                    }
```
tomxey: I think it would be nice here to somehow reuse the logic from "multi_get_cell", since it's very similar.

Did you consider calling "multi_get_cell" here first, and then sorting it's results?
Or adapting "multi_get_cell" so that it accepts a list like: [None, Some(seq_num_1), Some(seq_num_2), None, ..] and returns a result that can be returned directly here.

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -101,81 +112,165 @@ impl KvStoreClient {
...
             Key::CheckpointSummaryByDigest(_) => {
                 let checkpoint_digests = Self::extract_keys(&keys, |k| match k {
                     Key::CheckpointSummaryByDigest(checkpoint_digest) => Some(*checkpoint_digest),
                     _ => None,
                 })?;
 
-                let checkpoints = client
-                    .get_checkpoints_by_digest(&checkpoint_digests)
-                    .await?;
+                // pre-allocate results with None. Matching cells will replace None with
```
tomxey: With the implementation in this match arm being so long. I would suggest to extract (at least some parts of it) to separate function.

Or, even better, try to refactor it so that it's not that long.

I feel that with some modifications to "multi_get_cell" to make it more generic we could have the implementation here looking as follows:
```
let seq_num_keys = multi_get_cell(
    &mut client,
    CHECKPOINTS_BY_DIGEST_TABLE,
    keys,
    None,
)
.await;

return multi_get_cell(
    &mut client,
    CHECKPOINTS_TABLE,
    seq_num_keys,
    CHECKPOINT_SUMMARY_COLUMN_QUALIFIER,
)
.await;
```

Or am I missing some aspect?

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -101,81 +112,165 @@ impl KvStoreClient {
                     _ => None,
                 })?;
 
-                let transactions = client.get_transactions(&digests).await?;
```
tomxey, line 104: Should functions like "client.get_transactions" be removed after this refactor?

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -101,81 +112,165 @@ impl KvStoreClient {
...
-                    *tx_data.transaction.digest()
-                });
-                Self::extract_values_and_serialize(ordered, |tx_data| Some(&tx_data.effects))
+                multi_get_cell(
+                    &mut client,
+                    TRANSACTIONS_TABLE,
+                    keys,
+                    EFFECTS_COLUMN_QUALIFIER,
+                )
+                .await
```
tomxey: Did you consider having this logic e.g. in `client.get_transaction_effects` ? For this and other match arms.

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: These function is part of the public `KeyValueStoreReader` trait, so removing it would be a breaking change. Additionally, it's actively used in `iota-kvstore/main.rs`, so it should be retained for backward compatibility.

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9775/commits/254d1e20ef14980e558004bbe55843358975a9e8

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9775/commits/c30c354fe451c79e0d6155f13f693e13f8e8eb49

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: The `BigTableClient` is an external struct from the `iota-bigtable` crate. Previously, high-level methods were added through the public `KeyValueStoreReader` and `KeyValueStoreWriter` traits. To avoid breaking changes to the public API, I've kept the optimization logic private within `iota-rest-kv` rather than extending those traits.

tomxey: got it

tomxey (approved):
lgtm 👌

kodemartin (approved):
lgtm!

PR #9776, opened 2026-01-14
feat(move-core-types): Replace basic types with `iota-sdk` versions

DaughterOfMars:
## Description 

Replaces the following types with the `iota-sdk` versions:

- `StructTag`
- `TypeTag`
- `AccountAddress`
- `Identifier`
- `IdentStr`

These types were replaced with re-exports to reduce the amount of refactoring needed.

This PR uses the https://github.com/iotaledger/iota-rust-sdk/tree/feat/more-core-fns branch with some added functionality back-ported from the core repo. Some functions were skipped and re-implemented using different methods locally (ex. `AccountAddress::random()` is now `AccountAddress::new(rand::random())`).


muXxer (approved):
Changes to workspace `Cargo.toml` look fine.

external-crates/move/Cargo.toml
```diff
@@ -54,6 +54,7 @@ indexmap = "2.11.0"
 inline_colorization = "0.1.6"
 insta = "1.42.0"
 internment = { version = "0.5.0", features = ["arc"] }
+iota-sdk = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "c3b4004802bbc291480419436be22a5b6699e182", features = ["hash", "serde", "schemars"] }
```
Thoralf-M: Should probably only import the types crate and not the whole sdk here

DaughterOfMars: true

miker83z: What do you think about a dedicated move-types crate in the iota-rust-sdk? This would help us separating the contexts and make modifications/upstream pulls easier in the future. I already see that `Address` tries to merge `AccountAddress` and `IotaAddress`, but I think it would be better to keep them separated, at least at the first stage. 
@DaughterOfMars

DaughterOfMars: Why keep them separate when the types are identical?

I don't mind having another crate. I think that makes sense.

miker83z: AccountAddress is the representation of an address in the VM and the lang. Conceptually, it is different from the IotaAddress, which represent the address at the state level. Potentially, the IotaAddress could be implemented in a different way, while keeping the AccountAddress as it is to preserve the VM logics. 

Currently, both are implemented equally; but, I see the IotaAddress as a superset of AccountAddress instead of the same entity. Maybe we could continue with the superset approach instead of replacing them with 1 entity or instead of keeping them as they are. 

In any case, though, I think this PR covers already a big scope. So I would recommend to keep AccountAddress and IotaAccount separated for now, and re-assess a possible change later (possibly in the PR where the IotaAddress type will be replaced in the other crates).

miker83z: Then, also, let's go for the iota-sdk-move-core-types crate 👍🏻

PR #9780, opened 2026-01-15
feat(gRPC): add e2e test cases

### LINKED ISSUE #9621: feature(gRPC): add tests
Currently the tests mostly check the "read masks".

We should go through all endpoints and add test cases for example for "missing data" in the backend, or every other possible edge case, to test the full functionality, not only in the "happy-case".

These tests should be e2e tests. While we implement the tests, we should make sure that we are using the client functions, instead of the gRPC proto types directly, to test the full path.

bingyanglin:
# Description of change

- Add e2e tests for gRPC clients.
  - We cover edge cases but try to make the test set minimal, like if they are covered by Rust's type system, or similar error conditions (e.g., gas_budget=1 and gas_budget=0), we don't include them because they are redundant.
  - Run `cargo simtest -p iota-e2e-tests -- client::` for all the e2e tests.
- Fixes the signatures field mask (changing signatures.bcs to signatures) in the gRPC client
- Implements the previously unimplemented `get_checkpoint_data` endpoint in the server for ease our e2e tests
  - Note that we might need change after checkpoint streaming is completed
- Corrects signature deserialization to use BCS (found by our e2e tests)

Fixes #9621

### DIFF
--- a/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
+++ b/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
@@ -1,0 +1,77 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_macros::sim_test;
+
+use super::common::{assert_not_found_error, setup_grpc_test};
+
+#[sim_test]
+async fn get_latest_checkpoint() {
+    let (_test_cluster, client) = setup_grpc_test(1).await;
+
+    let checkpoint = client
+        .get_latest_checkpoint()
+        .await
+        .expect("Failed to get latest checkpoint");
+
+    assert!(
+        checkpoint.checkpoint.sequence_number >= 1,
+        "Latest checkpoint sequence number should be at least 1"
+    );
+}
+
+#[sim_test]
+async fn get_checkpoint_by_sequence() {
+    let (_test_cluster, client) = setup_grpc_test(2).await;
+
+    let checkpoint = client
+        .get_checkpoint(1)
+        .await
+        .expect("Failed to get checkpoint by sequence number");
+
+    assert_eq!(
+        checkpoint.checkpoint.sequence_number, 1,
+        "Checkpoint sequence number should match requested"
+    );
+}
+
+#[sim_test]
+async fn get_checkpoint_sequence_zero() {
+    let (_test_cluster, client) = setup_grpc_test(1).await;
+
+    let checkpoint = client
+        .get_checkpoint(0)
+        .await
+        .expect("Failed to get genesis checkpoint");
+
+    assert_eq!(
+        checkpoint.checkpoint.sequence_number, 0,
+        "Genesis checkpoint should have sequence number 0"
+    );
+}
+
+#[sim_test]
+async fn get_checkpoint_nonexistent_sequence() {
+    let (_test_cluster, client) = setup_grpc_test(1).await;
+
+    let result = client.get_checkpoint(999_999_999).await;
+    assert_not_found_error(result);
+}
+
+#[sim_test]
+async fn get_checkpoint_future_sequence() {
+    let (_test_cluster, client) = setup_grpc_test(1).await;
+
+    let latest = client
+        .get_latest_checkpoint()
+        .await
+        .expect("Failed to get latest checkpoint");
+
+    let future_sequence = latest.checkpoint.sequence_number + 100;
+    let result = client.get_checkpoint(future_sequence).await;
+
+    assert!(
+        result.is_err(),
+        "Fetching future checkpoint should return an error"
+    );
+}
--- a/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
+++ b/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
@@ -6,67 +6,44 @@ use iota_macros::sim_test;
 use super::common::{assert_not_found_error, setup_grpc_test};
 
 #[sim_test]
-async fn get_latest_checkpoint() {
-    let (_test_cluster, client) = setup_grpc_test(1).await;
-
-    let checkpoint = client
-        .get_latest_checkpoint()
-        .await
-        .expect("Failed to get latest checkpoint");
-
-    assert!(
-        checkpoint.checkpoint.sequence_number >= 1,
-        "Latest checkpoint sequence number should be at least 1"
-    );
-}
-
-#[sim_test]
-async fn get_checkpoint_by_sequence() {
+async fn get_checkpoint_scenarios() {
     let (_test_cluster, client) = setup_grpc_test(2).await;
 
-    let checkpoint = client
-        .get_checkpoint(1)
-        .await
-        .expect("Failed to get checkpoint by sequence number");
-
-    assert_eq!(
-        checkpoint.checkpoint.sequence_number, 1,
-        "Checkpoint sequence number should match requested"
-    );
-}
-
-#[sim_test]
-async fn get_checkpoint_sequence_zero() {
-    let (_test_cluster, client) = setup_grpc_test(1).await;
-
-    let checkpoint = client
-        .get_checkpoint(0)
-        .await
-        .expect("Failed to get genesis checkpoint");
-
-    assert_eq!(
-        checkpoint.checkpoint.sequence_number, 0,
-        "Genesis checkpoint should have sequence number 0"
-    );
-}
-
-#[sim_test]
-async fn get_checkpoint_nonexistent_sequence() {
-    let (_test_cluster, client) = setup_grpc_test(1).await;
-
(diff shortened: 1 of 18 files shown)


crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_macros::sim_test;
+
+use super::common::{assert_not_found_error, setup_grpc_test};
```
semenov-vladyslav: Does it make sense to aggregate all tests in this module into one? `setup_grpc_test` spawns a `TestCluster` which is a pretty slow thing. And creating it over and over for the sake of each unit test may be too much.

crates/iota-e2e-tests/tests/grpc/client/common.rs
```diff
@@ -0,0 +1,145 @@
...
+            let sig_bytes = bcs::to_bytes(sig).expect("BCS serialization failed");
+            bcs::from_bytes(&sig_bytes).expect("Signature deserialization failed")
+        })
+        .collect();
+
+    SignedTransaction {
+        transaction,
+        signatures,
+    }
+}
```
semenov-vladyslav: Why not use `iota_sdk_types_conversions` from `iota-types`?

crates/iota-e2e-tests/tests/grpc/client/execute.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    let result = client.execute_transaction(signed_tx, None).await;
+
+    // Transaction with invalid signature should be rejected
+    assert!(
+        result.is_err(),
+        "Transaction with invalid signature should fail"
+    );
+
+    match result {
```
semenov-vladyslav: There is `matches!` macro.

crates/iota-e2e-tests/tests/grpc/client/execute.rs
```diff
@@ -0,0 +1,172 @@
...
+        is_success(result1.effects.status()),
+        "First execution should succeed"
+    );
+
+    let result2 = client.execute_transaction(signed_tx, None).await;
+
+    // Re-submitting the same transaction may either:
+    // - Return the cached successful result
+    // - Return an error indicating the transaction was already executed
+    // Both behaviors are acceptable for idempotency
```
semenov-vladyslav: Pretty weird behaviour. Purely from rust PoV (essentially) the same result may be Ok or Err (depending on whether tx is cached or not). Clients will have to remember to check for AlreadyExists error code and not fail in that case.

semenov-vladyslav (approved):
Looks good, just a few minor comments.

crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
bingyanglin: Agreed, I also aggregated other modules too, except for `execute` tests. 
The execute tests cannot be aggregated because they use the wallet's gas selection to create transactions but execute via the gRPC client, which doesn't update the wallet's local object cache. Note that this is expected since the gRPC client and wallet are independent components with no shared state. Subsequent transactions would reference stale gas object versions, causing ObjectVersionUnavailableForConsumption errors. And I don't want to add more complexity here (IMHO it's unnecessary). Fixed at [3ae0149](https://github.com/iotaledger/iota/pull/9780/commits/3ae0149b4a6dc8f264b65a7dc89b3a4ebd775f4a).

crates/iota-e2e-tests/tests/grpc/client/common.rs
bingyanglin: Fixed at [7f717c2](https://github.com/iotaledger/iota/pull/9780/commits/7f717c2bccc7f8951f4672cc416c7c35e32cc951).

crates/iota-e2e-tests/tests/grpc/client/execute.rs
bingyanglin: Fixed at [1856704](https://github.com/iotaledger/iota/pull/9780/commits/1856704ad9a1fd40affa572a01f2f477acb0dca9).

crates/iota-e2e-tests/tests/grpc/client/execute.rs
bingyanglin: Thanks for pointing this out. For our test cases it should always return cached results, fixed at [2101e91](https://github.com/iotaledger/iota/pull/9780/commits/2101e9146b59b5cbd99de53901e98921205cd2bc).

crates/iota-e2e-tests/tests/grpc/client/execute.rs
```diff
@@ -0,0 +1,146 @@
...
+    // Verify gas was charged
+    let gas_summary = result.effects.gas_summary();
+    assert!(
+        gas_summary.computation_cost > 0 || gas_summary.storage_cost > 0,
+        "Some gas should have been charged"
+    );
+}
+
+#[sim_test]
+async fn execute_transaction_response_fields() {
```
muXxer: this could already be checked in `execute_transaction_transfer`

crates/iota-e2e-tests/tests/grpc/client/simulate.rs
```diff
@@ -0,0 +1,104 @@
...
+        Ok(response) => {
+            assert!(
+                !is_success(response.effects.status()),
+                "Simulation with insufficient gas should fail in effects"
+            );
+        }
+        Err(_) => {
+            // Server rejected the transaction due to invalid gas budget
+        }
+    }
```
muXxer: It should always be the same expected behaviour, why are there two possible outcomes allowed?

crates/iota-e2e-tests/tests/grpc/client/simulate.rs
```diff
@@ -0,0 +1,104 @@
...
+        Ok(response) => {
+            assert!(
+                !is_success(response.effects.status()),
+                "Simulation with insufficient balance should fail in effects"
+            );
+        }
+        Err(_) => {
+            // Server rejected the transaction due to insufficient balance
+        }
+    }
```
muXxer: same here.

crates/iota-e2e-tests/tests/grpc/client/common.rs
```diff
@@ -0,0 +1,114 @@
...
+    Digest::new(transaction_digest.into_inner())
+}
+
+/// Assert that a result is a "not found" style error.
+///
+/// The server may return different error types for not-found conditions:
+/// - `NotFound` or `InvalidArgument` gRPC status
+/// - `ProtoConversion` error when server returns empty data
+/// - `Server` error with a message
+pub fn assert_not_found_error<T: std::fmt::Debug>(result: Result<T, Error>) {
```
muXxer: I'm not sure if that is a good idea to combine all the different failure scenarios. We should expect a certain defined behaviour in the tests. So better add helper functions like `is_proto_conversion_err`, `is_grpc_error(code)` etc. What do you think?

muXxer (changes requested):

crates/iota-e2e-tests/tests/grpc/client/execute.rs
bingyanglin: Merged at [ad1038a](https://github.com/iotaledger/iota/pull/9780/commits/ad1038aba8b76e3e9dbb356160abe4f2e362328d).

crates/iota-e2e-tests/tests/grpc/client/simulate.rs
bingyanglin: It should go for the Err one only, fixed at [8702dd6](https://github.com/iotaledger/iota/pull/9780/commits/8702dd6a7543a2a01ca5c8c88e9603e1cfda7214).

crates/iota-e2e-tests/tests/grpc/client/simulate.rs
bingyanglin: It should go for the Ok branch only, fixed at [8702dd6](https://github.com/iotaledger/iota/pull/9780/commits/8702dd6a7543a2a01ca5c8c88e9603e1cfda7214).

crates/iota-e2e-tests/tests/grpc/client/common.rs
bingyanglin: I agree it's easier to trace if we have explicit errors, fixed at [7075543](https://github.com/iotaledger/iota/pull/9780/commits/707554398a1c9b4881b6ae6b0c97474b32debe23) and [4540246](https://github.com/iotaledger/iota/pull/9780/commits/4540246b086611a5131cf0aa18b85f0e21498154).

muXxer (approved):

PR #9782, opened 2026-01-15
feat(consensus): save voting blocks

### LINKED ISSUE #9766: fast sync: save voting blocks when fast syncing
to improve the overall reliability, it makes sense to store voting blocks for latest commits when fast syncing. This would not increase the bandwidth consumption, but make a small increase in the computational resources

VorobyevIlya:
# Description of change

This PR optimizes the search for voting blocks for the last commit in range by looking at the latest index in the range that has at least some votes instead of iterating over all indexes sequentially, which can take a long time in fast sync when we don't have votes for many indexes.

We also start to save voting headers that the validator fetches during fast sync and store them in a separate structure. Validator uses these headers when other validators request voting blocks from them. Validator doesn't use these blocks for header synchronization.

PR includes new tests and metrics.

Fixes #9766 

### DIFF
--- a/crates/starfish/core/src/authority_node.rs
+++ b/crates/starfish/core/src/authority_node.rs
@@ -218,6 +218,7 @@ impl ConsensusAuthority {
             network_client.clone(),
             block_verifier.clone(),
             dag_state.clone(),
+            store.clone(),
         )
         .start();
 
@@ -229,6 +230,7 @@ impl ConsensusAuthority {
             network_client.clone(),
             block_verifier.clone(),
             dag_state.clone(),
+            store.clone(),
         )
         .start();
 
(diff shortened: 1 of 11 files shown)


crates/starfish/core/src/authority_node.rs
```diff
@@ -1451,4 +1453,358 @@ mod tests {
...
+                }
+            }
+            sleep(Duration::from_millis(50)).await;
+        }
+
+        // Phase 4: Stop validator A
+        let last_processed_a = consumer_monitors[validator_a_index].highest_handled_commit();
+        authorities.remove(validator_a_index).stop().await;
+
+        // Phase 5: Let the remaining authorities (C, D, E) continue committing
```
polinikita: There 7 authorities, not 5

crates/starfish/core/src/authority_service.rs
```diff
@@ -916,14 +951,60 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        // Merge results: use voting headers where available, fallback headers otherwise
+        let mut fallback_iter = fallback_headers.into_iter();
+        let certifier_block_headers: Vec<VerifiedBlockHeader> = voting_headers
             .into_iter()
-            .flatten()
+            .zip(certifier_block_refs.iter())
+            .map(|(h, block_ref)| {
+                h.or_else(|| fallback_iter.next().flatten())
+                    .unwrap_or_else(|| {
+                        panic!("Block header {:?} should exist in storage", block_ref)
```
polinikita: Add more explanation to this panic that stored CommitVotes are not consistent with BlockHaders.

polinikita: This is a new panic in the flow. Previously, we didn't panic here with flatten and those inconsistencies were on the catching up validator. In the worst case, one validator does not catch up. Now, we can crash all of the, if something is broken

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -504,6 +508,14 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
             .await
             .expect("Spawn blocking should not fail")?;
 
+        // Store the voting block headers for later use when serving fetch_commits
```
polinikita, line 512: I would add a comment that this additional storage will not give a lot of computational/storage overhead in the process

crates/starfish/core/src/authority_node.rs
VorobyevIlya: fixed

crates/starfish/core/src/authority_service.rs
VorobyevIlya: I changed it into an error

crates/starfish/core/src/commit_syncer/fast.rs
VorobyevIlya: added comment

polinikita (approved):

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -653,6 +666,7 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
         );
 
         // Shuffle target authorities for load balancing
+        #[allow(unused_mut)]
```
piotrm50:
```suggestion
        #[cfg_attr(test, expect(unused_mut))]
```

crates/starfish/core/src/commit_syncer/fast.rs
```diff
@@ -505,6 +509,15 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
             .await
             .expect("Spawn blocking should not fail")?;
 
+        // Store the voting block headers for later use when serving fetch_commits
+        // requests. The overhead is small since we store votes only for the last commit
+        // in a batch.
+        if !voting_block_headers.is_empty() {
```
piotrm50: Can it ever be empty?

crates/starfish/core/src/commit_syncer/mod.rs
```diff
@@ -439,6 +450,7 @@ where
     loop {
         // Attempt to fetch commits and blocks through min(committee size,
         // MAX_NUM_TARGETS) peers.
+        #[allow(unused_mut)]
```
piotrm50:
```suggestion
        #[cfg_attr(test, expect(unused_mut))]
```

crates/starfish/core/src/metrics.rs
```diff
@@ -964,6 +966,16 @@ impl NodeMetrics {
                 &["authority", "source"],
                 registry
             ).unwrap(),
+            commit_sync_voting_block_headers_hits: register_int_counter_with_registry!(
+                "commit_sync_voting_block_headers_hits",
+                "Number of voting block headers served from voting storage during commit sync.",
```
piotrm50, line 971: meaning hits to in-memory cache, as opposed to storage fallback below?

piotrm50 (approved):
Left a few small nit comments

crates/starfish/core/src/commit_syncer/fast.rs
VorobyevIlya: Actually, no. I removed if

crates/starfish/core/src/commit_syncer/fast.rs
VorobyevIlya: fixed

crates/starfish/core/src/commit_syncer/mod.rs
VorobyevIlya: fixed

crates/starfish/core/src/metrics.rs
VorobyevIlya: No, this is also from the store, but from the separate structure for voting headers, where we save headers received during fast sync for certification of the commit range

PR #9792, opened 2026-01-15
fix(docs): move legacy migration tool outside of wallet section

Dr-Electron:
# Description of change

A suggestion from my side to move the legacy tool to its own place as it is wallet agnostic. Not sure if it should get such a prominent position but it makes the most sense to me. We could also think about not showing it in the sidebar at all and just have a link or being able to find it through search.

Related to #9617.

### DIFF
--- a/docs/content/sidebars/users.js
+++ b/docs/content/sidebars/users.js
@@ -26,7 +26,6 @@ const users = [
                 type: 'category',
                 label: 'Import Method',
                 items: [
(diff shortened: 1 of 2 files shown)


begonaalvarezd (approved):

29988122 (approved):
LGTM as it makes sense to put the page in the outmost layer to potentially make more dormant/old time users see it.

PR #9793, opened 2026-01-15
feat(starfish): add sequential restart simtests with strict transacti…

### LINKED ISSUE #9764: fast sync: add a simtest mimicking fast sync scenarios
Due to integration of fast syncing which might leave the state (store) in unrecoverable state in some cases. The simtest should test a complex scenario with several restarts of different nodes.

Ensure that transaction_ref is enabled in protocol parameters.

polinikita:
# Description of change

This PR removes the only simtest from the starfish crate, which mimicks the restart of one node. However, it adds several comprehensive sequential restart simulation tests to validate Starfish consensus resilience during node restarts, along with strict transaction and commit digest consistency verification across authorities. In these tests, nodes restarts in cycle with different time parameters for running and stopping and with/without cleaning consensus dbs.

In the following (after we fix the recovery logic in case of aborted fast syncing), we can enhance the simtests. Specifically, uncomment the last prepared simtest, in which we have persisted dbs, short stable run, long stop time.

Fixes #9764

### DIFF
--- a/crates/starfish/simtests/src/tests/simtests.rs
+++ b/crates/starfish/simtests/src/tests/simtests.rs
@@ -1,9 +1,12 @@
 // Copyright (c) Mysten Labs, Inc.
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
+
+#![cfg_attr(not(test), expect(unused))]
+
 #[cfg(msim)]
 mod test {
-    use std::{sync::Arc, time::Duration};
+    use std::{collections::HashSet, sync::Arc, time::Duration};
 
     use iota_config::local_ip_utils;
     use iota_macros::sim_test;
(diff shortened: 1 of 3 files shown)


piotrm50 (approved):

crates/starfish/simtests/src/tests/simtests.rs
```diff
@@ -41,23 +44,181 @@ mod test {
...
+            .collect();
+
+        if commit_indices.is_empty() {
+            return;
+        }
+
+        let min_commit: u32 = *commit_indices.iter().min().unwrap();
+        let max_commit = *commit_indices.iter().max().unwrap();
+
+        for commit_idx in min_commit..=max_commit {
```
VorobyevIlya, line 99: Why are we verifying that the digests are the same for such a segment? We start from the minimal among the highest handled commits. Why do we start here?

crates/starfish/simtests/src/tests/simtests.rs
```diff
@@ -66,88 +227,259 @@ mod test {
...
+        // Verify consistency after baseline progress
+        verify_commit_consistency(&authorities, "after initial progress").await;
+
+        // Sequential restart cycles for each authority
+        // Only restart authorities 0..NUM_OF_AUTHORITIES-1 to have one always alive
+        // authority with clean state to serve as sync source
+        // TODO: once fast sync logic can handle early abortion, we can include all
+        // authorities in the restart cycles
+        for authority_idx in 0..(NUM_OF_AUTHORITIES - 1) {
+            for cycle in 0..restart_config.cycles_per_authority {
```
VorobyevIlya, line 307: What do we do here? Do we restart the same node multiple times? What for?

VorobyevIlya (commented):
left two questions

crates/starfish/simtests/src/tests/simtests.rs
polinikita: Commits are chained by hashes, so it suffices to check the consistency of the latest ones. This was the simplest way for me to check that

crates/starfish/simtests/src/tests/simtests.rs
polinikita: We restart each node twice (cycles_per_authority = 2). When it is restarted for the first time (and no cleaned db), we expect normal boot (since db is consistent) and after the boot, the fast sync is activated, Fetching transactions and trusted commits is fine, but no headers are fetched except (probably) the tail. When it is restarted for the second time, the situation in storage is different. With long stopping time and short run time, it could simply not fetch all the headers for the tail and lead to the problem of booting the node.

VorobyevIlya (approved):

PR #9794, opened 2026-01-16
feat!: Replace `SequenceNumber` with SDK `Version`

### LINKED ISSUE #9835: Replace `SequenceNumber` with SDK version

DaughterOfMars:
# Description of change

Fixes #9835
Replaces the `SequenceNumber` in the `iota-types` crate with the `Version` from the SDK.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Some methods have been removed from SequenceNumber due to crate externalization (iota-rust-sdk::Version is replacing iota-types::SequenceNumber).
- [ ] REST API:


crates/iota-rest-api/src/transactions/execution.rs
```diff
@@ -413,7 +414,9 @@ pub(super) fn simulate_transaction_impl(
     parameters: &SimulateTransactionQueryParameters,
     transaction: Transaction,
 ) -> Result<TransactionSimulationResponse> {
-    let Transaction::V1(transaction) = transaction;
+    let Transaction::V1(transaction) = transaction else {
+        unimplemented!("a new enum variant was added and needs to be handled");
+    };
```
Thoralf-M, line 419: Why these changes?

thibault-martinez: Because we made the enums non-exhaustive. I also did this in the CLI PR.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -296,7 +296,7 @@ impl Scenario {
             let mut object_ref = object.compute_object_reference();
             self.outputs.live_object_markers_to_delete.push(object_ref);
             // in the authority this would be set to the lamport version of the tx
-            object_ref.1.increment();
+            object_ref.1 += 1;
```
Alex6323: can't we add the `increment()` API to our type? I prefer it over manually incrementing, and it would remove a lot of changed lines.

crates/iota-indexer/tests/rpc-tests/read_api.rs
```diff
@@ -1565,7 +1565,7 @@ fn try_get_past_object_version_not_found() {
 
         wait_for_objects_history(tx_digest, store, client).await;
 
-        let missing_version = gas_ref.1.one_before().expect("version should be > 0");
+        let missing_version = gas_ref.1 - 1;
```
Alex6323: This would change the behavior in release (not panicking anymore), right?

crates/iota-rest-api/src/transactions/execution.rs
Thoralf-M: For the transaction sure, but this PR just replaces the SequenceNumber?

thibault-martinez: Doesn't matter, since the revision is updated, they all need to be handled

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
DaughterOfMars: We could, but it seems a bit weird tbh

crates/iota-indexer/tests/rpc-tests/read_api.rs
DaughterOfMars: true, but shouldn't this not panic anyway?

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
Alex6323: Do you think it's weird In general or because it's just a type alias atm? 

I would prefer it to be a new-type tbh, because then we can enforce certain invariants and restrict it to be monotonically increasing and so on,. It should also never be possible to do `+= 2` etc. Even if it's unlikely that we make such a mistakes, it would be nicer if it's already enforced by the type.

crates/iota-indexer/tests/rpc-tests/read_api.rs
Alex6323: I dunno tbh. It was an `expect` before, so I guess that was on purpose ("should panic instead of wrap around").

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
DaughterOfMars: But why wouldn't it be possible? It already is with the existing type anyway

crates/iota-indexer/tests/rpc-tests/read_api.rs
DaughterOfMars: sure but I expect this is one of those "this will never actually panic, or otherwise we've reached the year 22560" type panics

bingyanglin (commented):
IMHO it might be good to have `increment`/`decrement` methods (like Alex said) instead of manual adding, so we can have overflow protection and make it consistent with `CheckpointSequenceNumber` and `RandomnessRound` where we use `checked_add` for them.

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -214,7 +214,7 @@ impl AuthorityStorePruner {
             for ObjectKey(object_id, seq_number) in object_tombstones_to_prune {
                 for result in perpetual_db.objects.safe_iter_with_bounds(
                     Some(ObjectKey(object_id, VersionNumber::MIN_VALID_INCL)),
-                    Some(ObjectKey(object_id, seq_number.next())),
+                    Some(ObjectKey(object_id, seq_number.next().unwrap())),
```
Thoralf-M, line 217: How about .expect("exceeding max sequence number") instead of unwrap()?

DaughterOfMars: the message is now defined in the SDK package as 
```
#[derive(thiserror::Error, Debug)]
pub enum VersionError {
    #[error("cannot increment Version: maximum valid Version has already been reached")]
    InvalidIncrement,
    #[error("cannot decrement Version: minimum valid Version has already been reached")]
    InvalidDecrement,
}
```
and these fns return a Result

Thoralf-M: Ah, that's fine then

Thoralf-M (approved):

thibault-martinez:
Don't merge it yet please

bingyanglin (approved):

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
miker83z: I also think it is cleaner to have `increment()` and we are also used to it

miker83z (approved):
LGTM, I just left a comment for possibly reintroducing `increment()`

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
DaughterOfMars: already done :)

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -642,10 +645,10 @@ impl TryFrom<TransactionKind> for crate::transaction::TransactionKind {
                     ),
                     random_bytes: randomness_state_update.random_bytes,
                     randomness_obj_initial_shared_version: randomness_state_update
-                        .randomness_obj_initial_shared_version
-                        .into(),
+                        .randomness_obj_initial_shared_version,
                 })
             }
+            _ => unimplemented!("a new enum variant was added and needs to be handled"),
```
thibault-martinez, line 650: rebase to get rid of this and other changes

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1038,19 +1038,14 @@ impl TryFrom<TransactionEffects> for IotaTransactionBlockEffects {
 fn owned_objref_string(obj: &OwnedObjectRef) -> String {
     format!(
         " ┌──\n │ ID: {} \n │ Owner: {} \n │ Version: {} \n │ Digest: {}\n └──",
-        obj.reference.object_id,
-        obj.owner,
-        u64::from(obj.reference.version),
-        obj.reference.digest
+        obj.reference.object_id, obj.owner, obj.reference.version, obj.reference.digest
```
kodemartin, line 1041: shouldn't this be 

```suggestion
        obj.reference.object_id, obj.owner, obj.reference.version.0, obj.reference.digest
```

crates/iota-types/src/iota_sdk_types_conversions.rs
DaughterOfMars: After merging, I think I prefer `unimplemented` over `unreachable`. Do you want to create another separate PR or just leave this small rename here?

crates/iota-json-rpc-types/src/iota_transaction.rs
DaughterOfMars: Doesn't matter, the display impl is identical

crates/iota-types/src/iota_sdk_types_conversions.rs
thibault-martinez: Another, and maybe same in the SDK for consistency

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: Still, is there a reason to rely on the identity of the implementations, and not the type itself? These blocks here explicitly evaluate the `u64` and it would be thus more future proof to take the inner value that is of the same type. 

What will happen if some future maintainer decides to change the display impl for `SequenceNumber`? The existing impl on `SequenceNumber` is certainly different for example

```rust
impl fmt::Display for SequenceNumber {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{:#x}", self.0)
    }
}
```

DaughterOfMars: The same could be said about any implementation of any type

consensus/core/src/synchronizer.rs
```diff
@@ -2194,6 +2194,139 @@ mod tests {
         assert_eq!(inflight_blocks_map.num_of_locked_blocks(), 0);
     }
 
+    #[tokio::test]
+    async fn test_process_fetched_blocks_duplicates() {
+        // GIVEN
```
Alex6323, line 2199: I am wondering what the relation is of this added test to the `SequenceNumber` replacement?

DaughterOfMars: I don't think I did this? Maybe you are looking at a merge commit?

DaughterOfMars: Ah, the rebase got messed up because I accidentally pulled develop 🙈

roman1e2f5p8s (commented):
In general, the changes look good to me. I have only a few concerns/questions:

- Why is the inner `u64` in `pub struct Version(pub u64);` public now? IMO, the previous implementation provides better encapsulation. Furthermore, keeping the `value()` method would introduce fewer changes in our codebase and fewer breaking changes in any external APIs.
- Maybe we could keep the old methods (optionally annotating them as deprecated) like `new()`, `from_u64`, etc. to introduce fewer breaking changes?
- For the multiple `unwrap()` uses, perhaps you can add messages about why you expect each of them should not fail.

DaughterOfMars:
>     * Why is the inner `u64` in `pub struct Version(pub u64);` public now? IMO, the previous implementation provides better encapsulation. Furthermore, keeping the `value()` method would introduce fewer changes in our codebase and fewer breaking changes in any external APIs.

Either way we are exposing the implementation with `fn value()`. But in general I don't really see the point in encapsulating this particular type, as it could easily be a u64, the only reason for the struct is implementation details like the consts.

>     * Maybe we could keep the old methods (optionally annotating them as deprecated) like `new()`, `from_u64`, etc. to introduce fewer breaking changes?

I think the reality is the changes we are going to make will be grossly breaking all over and there's really no point in trying to avoid that.

>     * For the multiple `unwrap()` uses, perhaps you can add messages about why you expect each of them should not fail.

I would, but I don't actually know why it's expected that they should not fail 😆

roman1e2f5p8s:
> Either way we are exposing the implementation with `fn value()`.

Right, but that exposes only read/immutable access. Making the inner `u64` public would allow manipulating the version manually, rather than forcing to use `nex()`, `increment()`, etc.

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: Ok, will not insist on this.

kodemartin (approved):
lgtm

DaughterOfMars:
> > Either way we are exposing the implementation with `fn value()`.
> 
> Right, but that exposes only read/immutable access. Making the inner `u64` public would allow manipulating the version manually, rather than forcing to use `nex()`, `increment()`, etc.

How do you feel about impls on the type that allow calling `Version(x) + 5` for instance?
Or the ability to do this, which is possible with the current API? ```*version = Version::new_u64(some_value)```

roman1e2f5p8s (approved):
As I mentioned, the inner `u64` in `pub struct Version(pub u64);` could be private to enhance encapsulation and safer manipulations.

muXxer (approved):
Approving files owned by node/core. 👍

crates/iota-rest-api/src/accounts.rs
```diff
@@ -102,13 +102,10 @@ impl ListAccountOwnedObjectsQueryParameters {
     }
 }
 
-#[serde_with::serde_as]
 #[derive(Debug, serde::Serialize, serde::Deserialize, schemars::JsonSchema)]
 pub struct AccountOwnedObjectInfo {
     pub owner: Address,
     pub object_id: ObjectId,
-    #[serde_as(as = "iota_types::iota_serde::BigInt<u64>")]
```
muXxer, line 110: Could it be that this change causes the failing replay test?
```
thread 'client_commands::profiler_tests::test_profiler' (283032) panicked at crates/iota/src/unit_tests/profiler_tests.rs:67:20:
    Failed to execute replay command. HINT: if the test is flaky, update the tx_digest to a more recent one by running "update_profiler_tx.sh".: IotaRpcError: "Rpc(ParseError(Error(\"invalid type: integer `0`, expected a string\", line: 1, column: 722)))"
    ```

File: crates/iota-json-rpc-tests/tests/read_api.rs [hunk 9]
```diff
@@ -1462,7 +1462,7 @@ async fn try_get_past_object_version_too_high() {
             .unwrap();
 
         assert!(
-            matches!(rpc_past_obj, IotaPastObjectResponse::VersionTooHigh{object_id: obj_id, asked_version, latest_version} if obj_id == object_id && asked_version == seq_num && latest_version == SequenceNumber::from_u64(1))
+            matches!(rpc_past_obj, IotaPastObjectResponse::VersionTooHigh{object_id: obj_id, asked_version, latest_version} if obj_id == object_id && asked_version == seq_num && latest_version.as_u64() == 1)
```
thibault-martinez: Could add Eq with an u64

crates/iota-types/Cargo.toml
```diff
@@ -65,7 +65,7 @@ iota-enum-compat-util.workspace = true
 iota-macros.workspace = true
 iota-network-stack.workspace = true
 iota-protocol-config.workspace = true
-iota-sdk-types.workspace = true
+iota-sdk-types = { workspace = true, features = ["serde", "schemars", "proptest"] }
```
thibault-martinez: Shouldn't `proptest` be brought only in dev-deps?

crates/iota-rest-api/src/accounts.rs
DaughterOfMars: It shouldn't be, since the implementation is the same in the SDK

crates/iota-types/src/iota_sdk_types_conversions.rs
DaughterOfMars: They are already using `unimplemented` in the SDK

crates/iota-core/src/authority/shared_object_version_manager.rs
```diff
@@ -422,7 +423,7 @@ mod tests {
                 .unwrap(),
             randomness_obj_version
         );
-        let next_randomness_obj_version = randomness_obj_version.next();
+        let next_randomness_obj_version = randomness_obj_version + 1;
```
thibault-martinez: revert pls

thibault-martinez (approved):

thibault-martinez:
Can you be a bit more explicit about the fast we replaced this type by the SDK one in the release notes please

piotrm50 (approved):

PR #9806, opened 2026-01-16
fix(dashboard): meta image not working in production deploys

### LINKED ISSUE #6987: fix(dashboard): Fix image in link unfurling for prod
Vercel for some reason doesn't populate the meta image link properly in prod wallet-dashboard.

 `<meta name="og:image" content="http://localhost:3000/metadata-image.png"/>`

But it does in a preview instance

` <meta property="og:image" content="https://wallet-dashboard-aowg2qzb4-iota1.vercel.app/metadata-image.png"/>`

Related discus https://iotafoundation.slack.com/archives/C057SC8EW15/p1748007961683739

⚠ Make the necessary changes in the PR so we test on the `staging` environment, and only when confirmed that is works, we pass the solution to the `production` env

VmMad:
# Description of change

Fixes #6987 


apps/wallet-dashboard/lib/constants/metadata.constants.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import type { Metadata } from 'next';
+
+const PRODUCTION_BASE_URL = 'https://wallet-dashboard.iota.org';
```
marc2332: this got me thinking, can we maybe use https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_URL ? Would it make sense?

marc2332 (changes requested):

apps/wallet-dashboard/lib/constants/metadata.constants.ts
VmMad: in theory it uses it as default when metadataBase it's `undefined`, that's why i only defined it in production (the only place it fails): https://nextjs.org/docs/14/app/api-reference/functions/generate-metadata#default-value

VmMad: after more research it seems like the problem comes from the workflow: we are building locally in the machine, but `VERCEL_URL` is only set when the deploy is done in vercel, **not** with `vercel build && vercel deploy --prebuilt` bu rather with `vercel deploy ...` and we cannot do just change it to not build it in the CI because the deploy gets assigned to the last commit author, which would throw errors if for example the commit author is not in the vercel team

marc2332 (approved):

begonaalvarezd (approved):

PR #9809, opened 2026-01-16
upstream(core): Checkpoints that have been locally built by validators do not need tx-by-tx re-execution in CheckpointExecutor

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/8e938bb4802483a3f6a4adedf7ccbb4abb916b7f
- Description:

When validators have built checkpoints locally, they have already done
the work of ensuring that transactions are executed. Comparing the local
vs certified checkpoint digest is sufficient to ensure that the
execution results agree.

Part of #9768

### DIFF
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -84,25 +84,24 @@ pub(crate) struct CheckpointExecutionData {
     pub checkpoint_contents: CheckpointContents,
     pub tx_digests: Vec<TransactionDigest>,
     pub fx_digests: Vec<TransactionEffectsDigest>,
+}
+
+pub(crate) struct CheckpointTransactionData {
     pub transactions: Vec<VerifiedExecutableTransaction>,
     pub effects: Vec<TransactionEffects>,
+    pub executed_fx_digests: Vec<Option<TransactionEffectsDigest>>,
 }
 
 pub(crate) struct CheckpointExecutionState {
     pub data: CheckpointExecutionData,
-    executed_fx_digests: Vec<Option<TransactionEffectsDigest>>,
     accumulator: Option<Accumulator>,
     full_data: Option<CheckpointData>,
 }
 
 impl CheckpointExecutionState {
-    pub fn new(
-        data: CheckpointExecutionData,
-        executed_fx_digests: Vec<Option<TransactionEffectsDigest>>,
-    ) -> Self {
+    pub fn new(data: CheckpointExecutionData) -> Self {
         Self {
             data,
-            executed_fx_digests,
             accumulator: None,
             full_data: None,
         }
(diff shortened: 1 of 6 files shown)


polinikita (approved):

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -396,14 +395,88 @@ impl CheckpointExecutor {
...
+        let (tx_digests, fx_digests): (Vec<_>, Vec<_>) = checkpoint_contents
+            .iter()
+            .map(|digests| (digests.transaction, digests.effects))
+            .unzip();
+
+        self.epoch_store
+            .insert_finalized_transactions(&tx_digests, sequence_number)
+            .expect("failed to insert finalized transactions");
+
+        CheckpointExecutionState::new(CheckpointExecutionData {
```
bingyanglin: Here in default the accumulator is set to be `None`, and later in our `commit_checkpoint` it passes this `None` to `accumulate_running_root `,  and then it will panic because it tries to read from disk and accumulator does not exist yet!(`state_accumulator.rs`, Line 338) So this is bug in the validator path. However in our fullnode path it explicitly sets the accumulator which is correct. See
```rust
            // The early versions of the accumulator (prior to effectsv2) rely on db
            // state, so we must wait until all transactions have been executed
            // before accumulating the checkpoint.
            ckpt_state.accumulator = Some(
                self.accumulator
                    .accumulate_checkpoint(&tx_data.effects, sequence_number, &self.epoch_store)
                    .expect("epoch cannot have ended"),
            );
```
in the `execute_checkpoint_fullnode()`.

This is fixe at https://github.com/MystenLabs/sui/pull/22103 though, where `notify_read_checkpoint_state_hasher().await` blocks until the background `CheckpointStateHasher` task has finished accumulating and called `notify()`. 

MystenLabs/sui#22103  should be ported with this PR when we merge to develop branch.

bingyanglin (approved):
Approved because my comment will be addressed in the future PR.

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
jkrvivian: This is a change from the commit I’m trying to port here.

It will no longer panic on line 338 in `state_accumulator.rs`. If acc is None, it now falls back to the default function via `unwrap_or_else` and retrieves the checkpoint from the same tables that were previously used in `notify_read_checkpoint_state_accumulator`.

I skipped porting this part because `notify_read_checkpoint_state_accumulator` was removed in commit https://github.com/iotaledger/iota/commit/e9fe49532a#diff-d37f044885a5d8c94e295f27980979d6c249bf267ee3ce7c4748a8f72f9cd683L1370

bingyanglin: But is it possible we panic because data does not exist in `state_hash_by_checkpoint`? E.g., the executor's `commit_checkpoint` can be called before the background `CheckpointAccumulator` finishes. This happens because:

- Builder sends to channel (async, non-blocking)
- Builder writes `locally_computed_checkpoints` to DB
- Executor sees `locally_computed_checkpoints` exists -> takes validator path
- Executor calls `accumulate_running_root` with None
- Fallback reads `state_hash_by_checkpoint` -> NOT THERE YET -> panic

The old `notify_read_checkpoint_state_accumulator` would wait for the data to exist. The current `unwrap_or_else` fallback just reads, so if the data isn't there, it panics instead of waiting.

Or do I miss anything?

jkrvivian: Alright, there could be edge cases. Added in https://github.com/iotaledger/iota/pull/9809/commits/383fff7b0963072d693d32c1c183ce327c61042d

bingyanglin (approved):

roman1e2f5p8s (approved):
Approving CODEOWNERS file(s).

muXxer (approved):

PR #9814, opened 2026-01-19
upstream(core): Test infra improvements

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/3042cddd67b86d622ec9bb9df1f19e7c9dfc30be
- Description:
- randomize cache capacities to exercise cache miss/eviction paths

Part of #9768

### DIFF
--- a/crates/iota-common/src/random.rs
+++ b/crates/iota-common/src/random.rs
@@ -1,0 +1,49 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// Get a random number generator.
+#[inline(always)]
+pub fn get_rng() -> impl rand::Rng {
+    rand::thread_rng()
+}
+
+// Need our own Either implementation so we can impl rand::RngCore for it.
+pub enum Either<L, R> {
+    Left(L),
+    Right(R),
+}
+
+impl<L, R> rand::RngCore for Either<L, R>
+where
+    L: rand::RngCore,
+    R: rand::RngCore,
+{
+    fn next_u32(&mut self) -> u32 {
+        match self {
+            Either::Left(l) => l.next_u32(),
+            Either::Right(r) => r.next_u32(),
+        }
+    }
+
+    fn next_u64(&mut self) -> u64 {
+        match self {
+            Either::Left(l) => l.next_u64(),
+            Either::Right(r) => r.next_u64(),
+        }
+    }
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        match self {
+            Either::Left(l) => l.fill_bytes(dest),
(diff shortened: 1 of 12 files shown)


crates/iota-core/src/checkpoints/mod.rs
bingyanglin, line 1889: In the upstream we added
```
assert_reachable!("skip checkpoint certification");
```
, any reason we don't add it?

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -797,21 +815,31 @@ impl WritebackCache {
...
+                            id,
+                            LatestObjectCacheEntry::Object(key.1, obj.clone().into()),
+                            ticket,
+                        );
+                        match obj {
+                            ObjectOrTombstone::Object(object) => Ok(Some(object)),
+                            ObjectOrTombstone::Tombstone(_) => Ok(None),
+                        }
+                    }
+                    _ => {
```
muXxer: This should be `Ok(None)`, and we should have an error branch so we can propagate potential errors IMO.

bingyanglin (commented):
The `antithesis_sdk` and the related code are not ported, do we ignore them by purpose?

crates/iota-common/src/util.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
```
muXxer:
```suggestion
// Modifications Copyright (c) 2026 IOTA Stiftung
```

crates/iota-common/src/random.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
```
muXxer:
```suggestion
// Modifications Copyright (c) 2026 IOTA Stiftung
```

crates/iota-common/src/random.rs
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/// Get a random number generator.
+#[inline(always)]
+pub fn get_rng() -> impl rand::Rng {
+    rand::thread_rng()
+}
+
+// Need our own Either implementation so we can impl rand::RngCore for it.
+pub enum Either<L, R> {
```
muXxer: I think we don't need this Either at all, and also not the implementation. We never return "EitherLeft" but directly the "thread_rng".

muXxer (changes requested):

crates/iota-core/src/checkpoints/mod.rs
jkrvivian: We will skip porting antithesis related changes.

muXxer (approved):

PR #9854, opened 2026-01-19
feat(iota-e2e-tests): Test the receiving of a gas coin in two separate TXs

### LINKED ISSUE #9371: [AA][e2e] Test the receiving of a gas coin in two separate TXs
Scenario:
- AA account object owns 1 Coin IOTA
- TX 1: sender being AA, using the coin as gas (so owned object)
- TX 2: sender being a random `Bob` address, `Bob` calls the “receive“ function from the AA shared object (a function which does not check `ensure_tx_sender_is_account`) with the same Coin IOTA as input

The test should check what happens if the gas coin is locked by validators in TX1 and then TX2 is sequenced by consensus before TX1.

Dkwcs:
# Description of change

TODO

Fixes #9371 .

`cargo simtest run test_receiving_gas_in_two_separate_txs`


miker83z:
Closed in favor of #9869

PR #9856, opened 2026-01-19
chore(docs): refactor developer section structure

### LINKED ISSUE #5524: IOTA Docs Proposal: Developer Menu
**Summary:**
- Add a toggle or switch in the docs to be able to switch entirely between docs for **L1 IOTA Mainnet** or **L2 IOTA EVM**.
- These are completely different networks, VMs, and tech stacks and the only thing connecting the two is the **IOTA Token Bridge**.
- The **IOTA Token Bridge** and other products that consider the bridging mechanism between both networks, can be part of both Docs, IOTA Mainnet & IOTA EVM.
- Anything else, only compatible with either of the two networks, should therefore be only documented in the respective Docs of the particular network.
- The Docs for IOTA EVM will be much shorter than the ones for IOTA Mainnet, because we develop/offer much more tooling, products, etc. for the latter. However, as described in [IOTA Docs Proposal: Top-Level Tabs](https://github.com/iotaledger/iota/issues/5522), we also here can and should reference compatible and popular 3rd party tooling, e.g. **Metamask**, **Remix**, **Truffle**, etc.

The attachments provide an example of such a switch, leading to a clear separation of content for the both networks:

Image

Image

Dr-Electron:
# Description of change

This PR tries to come up with a better structure for the developer section

Fixes #5524


vivekjain23 (approved):

docs/site/src/css/custom.css
```diff
@@ -483,7 +483,8 @@ h4 {
 .navbar__item.navbar-icon-developer,
 .navbar__item.navbar-icon-operator,
 .navbar__item.navbar-icon-references,
-.navbar__item.navbar-icon-users {
+.navbar__item.navbar-icon-users,
+.navbar__item.navbar-icon-workshops {
```
29988122, line 487: Image

I don't know, workshop icon aligned on the right side seems strange to me - but I think it's ok.

Let's push it first and change it later (if ever needed)

29988122 (approved):

docs/site/src/css/custom.css
Dr-Electron: Yeah I would be glad to get some help on styling. This is also not a responsive layout as on mobile already the button did break 😓

PR #9857, opened 2026-01-20
feat: Improve `select_gas` performance

### LINKED ISSUE #9801: tx builder: select_gas() should not prefetch all gas coins
## Description

select_gas() currently fetches all gas coins, just to then iterate over the ids and fetch the objects with their data again. This is causing issues and is not efficient, it should just fetch objects page per page and check then for one with sufficient funds instead.

## Motivation

Efficiency

## Requirements

1. Don't always fetch all gas coins, just request them page by page with their data and break early if one is found with sufficient amount

DaughterOfMars:
## Description 

Alternative to #9846. I felt that the implementation was still not effective, and instead slightly modified the signature of `DataReader` to allow pagination. There are some additional changes to reduce cloning that were incidental.

Now `select_gas` only fetches the pages and retrieves all the data it needs without additional queries.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: Don't set a next_cursor if !has_next_page
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Modified `get_owned_objects()` to allow more efficient `select_gas()` implementation that doesn't fetch all pages if not needed

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -34,7 +34,7 @@ use iota_transaction_builder::DataReader;
 use iota_types::{
     TypeTag,
     balance::Supply,
-    base_types::{IotaAddress, ObjectID, ObjectInfo, SequenceNumber, VersionNumber},
+    base_types::{IotaAddress, ObjectID, SequenceNumber, VersionNumber},
     coin::{CoinMetadata, TreasuryCap},
     coin_manager::CoinManager,
     committee::EpochId,
@@ -249,9 +249,7 @@ impl IndexerReader {
             .await?;
 
         if let Some(object) = stored_object {
-            object
-                .try_into_object_read(self.package_resolver.clone())
-                .await
+            object.try_into_object_read(&self.package_resolver).await
         } else {
             Ok(ObjectRead::NotExists(object_id))
         }
@@ -321,10 +319,7 @@ impl IndexerReader {
             .await?;
 
         match history_object {
-            Some(obj) => {
-                obj.try_into_past_object_read(self.package_resolver.clone())
-                    .await
-            }
+            Some(obj) => obj.try_into_past_object_read(&self.package_resolver).await,
             None => Err(IndexerError::DataPruned(format!(
                 "Object version {} not found in objects_history for object {object_id}",
                 object_version_info.object_version
@@ -379,10 +374,7 @@ impl IndexerReader {
         // The key difference is that `try_into_past_object_read` also handles
         // the `WrappedOrDeleted` status, which for this iteration, we handle explicitly
         // as a `VersionNotFound`.
-        match obj
-            .try_into_object_read(self.package_resolver.clone())
-            .await?
-        {
+        match obj.try_into_object_read(&self.package_resolver).await? {
             ObjectRead::NotExists(_) | ObjectRead::Deleted(_) => {
                 Ok(PastObjectRead::VersionNotFound(object_id, object_version))
             }
@@ -876,14 +868,13 @@ impl IndexerReader {
     ) -> IndexerResult<Vec<IotaTransactionBlockResponse>> {
         let mut tx_block_responses_futures = vec![];
         for stored_tx in stored_txes {
-            let package_resolver_clone = self.package_resolver();
             let options_clone = options.clone();
-            tx_block_responses_futures.push(tokio::task::spawn(
-                stored_tx.try_into_iota_transaction_block_response(
-                    options_clone,
-                    package_resolver_clone,
-                ),
-            ));
+            let package_resolver = self.package_resolver.clone();
+            tx_block_responses_futures.push(tokio::task::spawn(async move {
+                stored_tx
+                    .try_into_iota_transaction_block_response(options_clone, &package_resolver)
+                    .await
+            }));
         }
 
         let tx_blocks = futures::future::join_all(tx_block_responses_futures)
@@ -2315,8 +2306,8 @@ impl IndexerReader {
         ))
     }
 
-    pub fn package_resolver(&self) -> PackageResolver {
-        self.package_resolver.clone()
+    pub fn package_resolver(&self) -> &PackageResolver {
+        &self.package_resolver
     }
 
     pub async fn pending_active_validators(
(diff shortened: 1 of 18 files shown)


crates/iota-transaction-builder/src/utils.rs
```diff
@@ -47,26 +47,39 @@ impl TransactionBuilder {
...
+                            .ok_or_else(|| anyhow!("bcs field is unexpectedly empty"))?
+                            .try_as_move()
+                            .ok_or_else(|| anyhow!("Cannot parse move object to gas object"))?
+                            .bcs_bytes,
+                    )?;
+                    if !input_objects.contains(&obj.object_id) && gas.value() >= gas_budget {
+                        return Ok(obj.object_ref());
+                    }
                 }
+                if page.next_cursor.is_none() {
```
Thoralf-M: This is never none
```Rust
#[path = "../utils.rs"]
mod utils;
use std::str::FromStr;

use iota_sdk::IotaClientBuilder;
use iota_types::base_types::IotaAddress;

#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
    let client = IotaClientBuilder::default().build_testnet().await?;

    let sender = IotaAddress::from_str(
        "0x111111111504e9350e635d65cd38ccd2c029434c6a3a480d8947a9ba6a15b215",
    )?;

    let coins = client
        .coin_read_api()
        .get_coins(sender, None, None, None)
        .await?;
    let gas_coin = coins.data.into_iter().next().unwrap();

    let gas_budget = 50_000_000;

    let _tx_data = client
        .transaction_builder()
        .split_coin_equal(sender, gas_coin.coin_object_id, 10, None, gas_budget)
        .await?;

    Ok(())
}
```
loops forever, I added a log at the start inside of the loop and logged each coin amount:
```
loop
gas value: 1
gas value: 1
gas value: 1
gas value: 1
gas value: 40000000
gas value: 1
gas value: 40000000
gas value: 1
gas value: 40000
gas value: 40000
gas value: 1
gas value: 1
gas value: 1
gas value: 1
gas value: 40000000
gas value: 1
gas value: 40000000
gas value: 40000
gas value: 40000
gas value: 40000000
gas value: 40000000
gas value: 40000
gas value: 40000
gas value: 40000
gas value: 40000
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 40000000
gas value: 40000
gas value: 40000
gas value: 1
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 40000
gas value: 1
gas value: 1
gas value: 1
gas value: 40000
gas value: 1
gas value: 1
gas value: 1
gas value: 40000000
gas value: 40000
gas value: 40000
gas value: 40000
gas value: 40000
gas value: 1
gas value: 40000000
loop
gas value: 40000000
gas value: 40000000
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 1
gas value: 1
gas value: 1
gas value: 1
gas value: 40000
gas value: 40000
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 1
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 40000000
gas value: 40000000
gas value: 40000
gas value: 1
gas value: 40000000
gas value: 40000
gas value: 40000000
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 40000
gas value: 40000
gas value: 40000
gas value: 1
gas value: 1
gas value: 1
gas value: 1
gas value: 40000
gas value: 40000000
gas value: 40000000
loop
loop
loop
loop
loop
loop
loop
loop
loop
loop
loop
loop
...
```

kodemartin (approved):
lgtm

crates/iota-indexer/src/read.rs
```diff
@@ -2367,25 +2358,47 @@ impl DataReader for IndexerReader {
...
-                let object = Object::try_from(object)?;
-                let object_ref = object.compute_object_reference();
-                let info = ObjectInfo::new(&object_ref, &object);
-                Ok(info)
-            })
-            .collect::<Result<Vec<_>, _>>()
+        let mut res = Vec::new();
+
+        let mut next_cursor = None;
+        if stored_objects.len() > limit {
```
bingyanglin: Do we also need `&& limit > 0` here?

bingyanglin (commented):
Not sure if the condition was missed, others LGTM.

sdk/kiosk/test/e2e/e2e.test.ts
```diff
@@ -78,18 +78,6 @@ describe('Testing Kiosk SDK transaction building & querying e2e', () => {
...
-            address: toolbox.address(),
-            pagination: {
-                limit: 1,
-                cursor: page.nextCursor!,
-            },
-        });
-        expect(emptyPage.hasNextPage).toBe(false);
-        expect(emptyPage.nextCursor).toBe(page.nextCursor);
-        expect(emptyPage.kioskIds).toHaveLength(0);
-        expect(emptyPage.kioskOwnerCaps).toHaveLength(0);
```
Thoralf-M, line 92: Shouldn't delete this, just  modify the nextCursor check?
Same below

crates/iota-indexer/src/read.rs
DaughterOfMars: yes, thank you for reminding me <3

sdk/kiosk/test/e2e/e2e.test.ts
DaughterOfMars: No, the problem is the test is wrong. It can't use the previous page's cursor to get an empty page because there is no cursor returned now. If it uses it it will get the first page again instead

Thoralf-M: Ah that was a second request already, all good then

Thoralf-M (approved):

crates/iota-json-rpc/src/transaction_builder_api.rs
```diff
@@ -52,15 +52,38 @@ impl DataReader for AuthorityStateDataReader {
...
+                IotaObjectResponse::try_from_object_read_and_options(read, &options)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let next_cursor = if result.len() > limit {
+            // Here the cursor is the first object id of the next page
+            result.pop().unwrap().object_id().ok()
+        } else {
+            None
+        };
```
bingyanglin, line 80: I am not pretty sure about this, why this logic is different from `crates/iota-indexer/src/read.rs` [Lines 2377-2386?](https://github.com/iotaledger/iota/pull/9857/changes#diff-e71241584ad87fed880fdf426a1ea5f3515a8dc9773c16929f7c4c12dbe39adaR2378-R2386)?

DaughterOfMars: This API takes the first cursor of the next page rather than the last cursor of the current page :/

bingyanglin (approved):

brancoder (approved):

PR #9869, opened 2026-01-20
fix(core): reject receiving objects owned by AA accounts

### LINKED ISSUE #9371: [AA][e2e] Test the receiving of a gas coin in two separate TXs

### LINKED ISSUE #9862: [AA] Remove Receive ability for AA Account shared objects
- [x] Validate receiving object owner exists: 
  - When processing receiving object arguments, check if the owner object exists in the database. If the owner does not exist, validators must refuse to sign the certificate.
- [x] Block receiving objects owned by Account objects: 
  - For receiving object arguments, reject transactions where the owner is an Account object. Validators must refuse to sign such transactions.
- [ ] Enforce fresh UID for Account creation: 
  - When creating an Account object, validate that the UID (ObjectID) is freshly created within the same transaction. Reject Account creation with pre-existing UIDs.

miker83z:
# Description of change

Based on this scenario -> #9371 Receiving objects could be used to make AA TX certificates invalid, even after these were signed by the majority of validators. 

This PR makes so that any TX trying to receive an object owned by an AA account (i.e., an object having a  `AuthenticatorFunctionRef` dynamic field) would abort execution (a validator could still sign the certificate, but the execution makes sure that the received objects will not be received and won't bump the sequence number).

This PR also fixes a small issue related to duplicate function names in a package. 

Closes #9371
Fixes #9862 


muXxer (approved):

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -291,4 +295,65 @@ impl TransactionInputLoader {
...
+        }
+
+        // Build a lookup: input object ID -> version
+        let input_objects_map: HashMap<ObjectID, SequenceNumber> = input_results
+            .iter_objects()
+            .map(|obj| (obj.id(), obj.version()))
+            .collect();
+
+        // for each receiving object, check if its owner is an account object
+        for receiving in receiving_results.iter_objects() {
```
Dkwcs: Minor: would it be better to rename it to `receiving_obj` or similar.

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -291,4 +295,65 @@ impl TransactionInputLoader {
...
+            let Owner::AddressOwner(owner_address) = receiving.owner() else {
+                // if not an address owner, it will fail in later checks
```
Dkwcs: Does it make sense to conduct this check here?

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -291,4 +295,65 @@ impl TransactionInputLoader {
...
+                &AuthenticatorFunctionRefV1Key::default().to_bcs_bytes(),
+            )
+            .expect("derive_dynamic_field_id should not fail");
+
+            // Try to fetch the authenticator field at or before the shared version
+            if let Some(_auth_field_obj) = self
+                .cache
+                .try_find_object_lt_or_eq_version(auth_field_id, version)
+                .ok()
+                .flatten()
```
Dkwcs: If `try_find_object_lt_or_eq_version` fails due to some issues( cache read error), we would skip the restriction and allow a receiving object owned by account object to pass “validation” -  is this expected behavior? 

For instance: 
- `owner_id` is an account object and does have `AuthenticatorFunctionRefV1` dynamic field
- `auth_field_id` exists at <= version
Cache returns some error, then .`ok().flatten()` turns that into `None`, so we do not throw `ReceivingObjectForAccountObject`

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -109,14 +112,15 @@ impl TransactionInputLoader {
...
-                .into(),
-            receiving_results,
-        ))
+        let input_results: InputObjects = input_results
+            .into_iter()
+            .map(Option::unwrap)
+            .collect::<Vec<_>>()
+            .into();
+
+        self.validate_receiving_objects_not_in_account_objects(&input_results, &receiving_results)?;
```
alexsporn: This new check should be feature-gated to the AA features.

piotrm50 (approved):

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -352,67 +454,438 @@ async fn test_receiving_gas_in_two_separate_txs() -> Result<(), anyhow::Error> {
...
-        .craft_aa_receive_gas_ptb(conflict_coin_ref, AA_RECEIVE_OBJECT_FN_NAME_NO_SENDER_CHECK)?;
+    let pt2 = test_env.craft_aa_receive_gas_ptb(
+        conflict_coin_ref,
+        AA_MODULE_NAME,
+        AA_RECEIVE_OBJECT_FN_NAME_NO_SENDER_CHECK,
+    )?;
     let tx2_data = test_env.craft_tx_from_pt(pt2, bob_gas, bob, None).await?;
+    // Create the TX envelope and send it for validators signing
     let tx2 = test_env.test_cluster.wallet.sign_transaction(&tx2_data);
     // This must fail during signing because of the conflict Coin
```
lzpap: Wrong comment

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -352,67 +454,438 @@ async fn test_receiving_gas_in_two_separate_txs() -> Result<(), anyhow::Error> {
...
+            Some(client_ip),
+        )
+        .await
+        .unwrap();
     assert!(
-        tx2_resp.confirmed_local_execution.unwrap_or(false) && !tx2_resp.errors.is_empty(),
-        "TX2 must succeed to reproduce the intended race. Resp={:#?}",
-        tx2_resp
+        effects_cert.summary_for_debug().status.is_err(),
+        "Expected the TX execution to fail due to conflict on receiving object"
```
lzpap: it shall fail because receiving is not supported on the AA object, no?

lzpap (commented):
@miker83z please modify `crates/iota-adapter-transactional-tests/tests/abstract_account/account/receive_object.move` as the tests are failing due to that test case expecting a receive to go through on an account object

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: Fixed in [dab65e1](https://github.com/iotaledger/iota/pull/9869/commits/dab65e1bfca858bc33d654e6ba8c4aeaccfc1611)

crates/iota-e2e-tests/tests/abstract_account_tests.rs
miker83z: Fixed in [dab65e1](https://github.com/iotaledger/iota/pull/9869/commits/dab65e1bfca858bc33d654e6ba8c4aeaccfc1611)

alexsporn (approved):

lzpap (approved):
lgtm!

crates/iota-adapter-transactional-tests/tests/abstract_account/account/receive_object.snap
```diff
@@ -41,13 +41,13 @@ gas summary: computation_cost: 1000000, computation_cost_burned: 1000000, storag
 
 task 6, lines 66-67:
 //# abstract --account immshared(2,2) --ptb-inputs object(2,2) receiving(5,0)
-mutated: object(2,0), object(2,2), object(5,0)
-gas summary: computation_cost: 1000000, computation_cost_burned: 1000000, storage_cost: 3351600,  storage_rebate: 3351600, non_refundable_storage_fee: 0
+Error: Transaction Effects Status: Move Runtime Abort. Location: iota::transfer::receive_impl (function index 12) at offset 0, Abort Code: 5
+Debug of error: MoveAbort(MoveLocation { module: ModuleId { address: iota, name: Identifier("transfer") }, function: 12, instruction: 0, function_name: Some("receive_impl") }, 5) at command Some(0)
```
Dkwcs, line 45: Does this error explain the receiving case enough?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2371,6 +2372,9 @@ impl ProtocolConfig {
                         cfg.feature_flags.enable_move_authentication = true;
                         // Max auth gas budget is in NANOS and an absolute value 0.25 IOTA
                         cfg.max_auth_gas = Some(250_000_000);
+                        // Increase the base cost for transfer receive object in devnet, since the
+                        // implementation now does check if parent is not an account.
+                        cfg.transfer_receive_object_cost_base = Some(100);
```
valeriyr, line 2377: Just to clarify, how was the new value calculated?
The current cost is `52`, `dynamic_field_has_child_object_cost_base` equals `100`.
Should the new value be `52` + `100` = `152`?

crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_19.snap
```diff
@@ -165,7 +165,7 @@ object_record_new_uid_cost_base: 52
 transfer_transfer_internal_cost_base: 52
 transfer_freeze_object_cost_base: 52
 transfer_share_object_cost_base: 52
-transfer_receive_object_cost_base: 52
+transfer_receive_object_cost_base: 100
```
Dkwcs, line 168: Just would like to know - what is the mechanism for changing those values? (Why 100 not other value, like 152? )

iota-execution/latest/iota-move-natives/src/transfer.rs
```diff
@@ -31,6 +33,9 @@ const E_RECEIVING_OBJECT_TYPE_MISMATCH: u64 = 2;
 // Represents both the case where the object does not exist and the case where
 // the object is not able to be accessed through the parent that is passed-in.
 const E_UNABLE_TO_RECEIVE_OBJECT: u64 = 3;
+// Represents the case where it is trying to receive an object owned by an
+// account.
+const E_ACCOUNT_CANNOT_RECEIVE_OBJECT: u64 = 5;
```
Dkwcs, line 38: Should it be 4 instead of 5? In order to maintain logical error numbering

crates/iota-move-build/src/lib.rs
```diff
@@ -158,13 +158,18 @@ impl BuildConfig {
...
             let mod_is_test = u.attributes.is_test_or_test_only();
             for (_, s, info) in &u.function_infos {
                 let fn_name = s.as_str().to_string();
                 let is_test = mod_is_test || info.attributes.is_test_or_test_only();
                 let authenticator_version = info.attributes.get_authenticator();
                 fn_info_map.insert(
-                    FnInfoKey { fn_name, mod_addr },
+                    FnInfoKey {
+                        fn_name,
+                        mod_name: mod_name.clone(),
```
valeriyr, line 170: Looks like `clone` can be removed.

Dkwcs (approved):
LGTM, left minor comments

crates/iota-types/src/move_package.rs
```diff
@@ -92,6 +92,7 @@ pub struct FnInfo {
 /// Uniquely identifies a function in a module
 pub struct FnInfoKey {
     pub fn_name: String,
+    pub mod_name: String,
```
valeriyr, line 95: Why do we need a module name here? Does it mean that `FnInfoKey` currently doesn't identify a function uniquely? 
Do we have a test that shows the issue?

crates/iota-protocol-config/src/lib.rs
lzpap: the current value of `52` covers the cost of:
 1. querying the database against an objects id, loading up the child,
 2. comparing the type of the loaded object to the one expected.

With the receive parent check, we only do essentially `1`, as we know that under that key there can only be stored a dynamic field with a certain type, so no need to check. Therefore, the new check only adds an additional database query which makes us think that roughly double the cost is appropriate, hence `100`

lzpap: @valeriyr we could also do `104` if you deem the heavy lifting is the db query.

crates/iota-adapter-transactional-tests/tests/abstract_account/account/receive_object.snap
miker83z: No, but that's the only one we could obtain through a native function I guess.

iota-execution/latest/iota-move-natives/src/transfer.rs
miker83z: Number 4 only appears on the Move side, we should take into consideration that also on the rust side

crates/iota-protocol-config/src/lib.rs
valeriyr: I am asking because, as far as I see, we added the same check as can be found here(`child_object_exists` is the only function that is called here):
https://github.com/iotaledger/iota/blob/4a711d20f4814842a06eea8909815b9ed6acc720/iota-execution/latest/iota-move-natives/src/dynamic_field.rs#L432
It costs `100`.

crates/iota-types/src/move_package.rs
miker83z: > Does it mean that FnInfoKey currently doesn't identify a function uniquely?

Exactly, this is like this also upstream (but it only affects `test` attributes in that case)

crates/iota-move-build/src/lib.rs
miker83z: It's in a loop!

PR #9876, opened 2026-01-21
feat(wallet,explorer): update app backends deprecated methods

### LINKED ISSUE #9630: Update apps backend deprecated methods
updated deprecated methods ex. `loadFeatures`

brancoder:
# Description of change

resolves #9630


evavirseda (approved):

apps/wallet/src/ui/app/index.tsx
```diff
@@ -97,6 +97,11 @@ export function App() {
         [accounts],
     );
     const backgroundClient = useBackgroundClient();
+
+    useEffect(() => {
+        backgroundClient.refreshFeatures();
+    }, [backgroundClient]);
```
marc2332: I don't see why we need this at all, when the background client is initialized it already refreshes the features https://github.com/iotaledger/iota/blob/59276535742aade9e14f4581ac8fb82d2ad92b27/apps/wallet/src/ui/app/background-client/index.ts#L69-L70.
```suggestion
```

apps/wallet/src/ui/app/background-client/index.ts
```diff
@@ -614,6 +614,10 @@ export class BackgroundClient {
         );
     }
 
+    public refreshFeatures() {
+        return this.loadFeatures();
+    }
```
marc2332:
```suggestion
```

marc2332 (changes requested):
Might as well also update other deprecated features no? E.g `setFeatures` is now `setPayload` 

Image

apps/wallet/src/ui/app/index.tsx
brancoder: we need it to refresh the features when user closes the wallet popup and reopens it. This can be tested by running the apps backend locally and changing a feature. Reopening the wallet should have the new updated features.

marc2332: I dont understand @brancoder, you added a useEffect so that when the backgroundClient changes you refresh the features, but when a background client is created it already refreshes the features, so the useEffect is not needed. What am I missing?

brancoder: When a user reopens the wallet popup background client doesn't get recreated, as its a singleton, so the init() method is not called again. The useEffect ensures features are refreshed on each wallet open. That said,  I think we could remove the `backgroundClient` from the dependencies array of the useEffect

marc2332 (approved):
🤝

PR #9879, opened 2026-01-21
fix(starfish): separate recovery of commit observer into the linearizer initialization and sending commits to the consensus handler

### LINKED ISSUE #9873: starfish: don't load PendingSubDags for all prior commits
Currently, we load PendingSubDags for all commits starting with last processed commit in the consensus handler. This is not accurate and require existence of all headers in the consensus db. With fast sync, this is no longer a requirement for proper booting the node. We can use trusted commits to get CommittedSubDAGs without headers. This will a) allow for rebooting after fast syncing, b) faster rebooting process.

polinikita:
# Description of change

Fix panic during consensus restart when block headers could be missing for some commits in the prefix, e.g. after fast sync finished the job, but the node db for the epoch is cleaned or last processed commit in the handler is lower that the last commit produced by the consensus.                                   
                                                             
  Changes:                                                   
  - Split commit observer recovery into two phases:         
       1. linearizer and solidifier state rebuild (from recent commits only)
       2. unprocessed commit resending (with empty headers)          
  - Add RestartMode enum to simtests with new ResetLastProcessed mode that validates the fix   

Fixes #9873 

- [x] Patch-specific tests (correctness, functionality coverage): Added one more simtest
- [x] I have added tests that prove my fix is effective or that my feature works: Added one more test that was failing before the fix

### DIFF
--- a/crates/starfish/core/src/commit_observer.rs
+++ b/crates/starfish/core/src/commit_observer.rs
@@ -275,72 +275,130 @@ impl CommitObserver {
             return;
         }
 
-        self.commit_solidifier
-            .set_last_committed_index(last_commit_index);
+        // Phase 1: Resend all possible committed sub-dags that haven't been processed
+        let sent_commits =
+            self.resend_unprocessed_commits(last_processed_commit_index, last_commit_index);
 
-        // Phase 1: Recover linearizer state (only gc_depth*2 commits needed)
-        self.recover_linearizer_state(last_commit_index);
-
-        // Phase 2: Resend unprocessed commits
-        self.resend_unprocessed_commits(last_processed_commit_index, last_commit_index);
+        // Phase 2: Recover linearizer and solidifier state
+        self.recover_linearizer_and_solidifier_state(last_commit_index);
 
         info!(
-            "Commit observer recovery completed, took {:?}",
+            "Commit observer recovery completed, resent {} commits, took {:?}",
+            sent_commits.len(),
             now.elapsed()
         );
     }
 
-    /// Recovers linearizer trackers from recent commits.
-    /// Only processes commits in the gc_depth*2 window.
-    fn recover_linearizer_state(&mut self, last_commit_index: CommitIndex) {
+    /// Recovers linearizer trackers from recent commits and seeds the
+    /// commit solidifier with any unprocessed commits.
+    fn recover_linearizer_and_solidifier_state(&mut self, last_commit_index: CommitIndex) {
         let linearizer_recovery_start = last_commit_index
             .saturating_sub(self.context.protocol_config.gc_depth() * 2)
             .max(1);
+        let solidifier_recovery_start = self.last_sent_commit_index.saturating_add(1).max(1);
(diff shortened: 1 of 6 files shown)


crates/starfish/core/src/commit_observer.rs
```diff
@@ -326,70 +432,47 @@ impl CommitObserver {
...
-            // traversed-headers gate when committing transactions.
-            self.linearizer
-                .record_traversed_headers(pending_sub_dag.headers.iter());
-
-            // Recover transaction acknowledgments tracker state by adding transaction
-            // acknowledgments from all pending sub-dags that still might
-            // correctly acknowledge transactions.
-            for ((round, authority_idx), transaction_acknowledgments) in
-                pending_sub_dag.transaction_acknowledgments().into_iter()
+            // Create CommittedSubDag with empty headers (like fast sync)
```
VorobyevIlya: misplaced comment

VorobyevIlya (approved):

crates/starfish/core/src/commit_observer.rs
```diff
@@ -326,70 +428,46 @@ impl CommitObserver {
...
-                    debug!(
-                        "Not sending solid commit as commit index {} <= \
-                    {last_processed_commit_index} last processed index",
-                        solid_sub_dag.commit_ref.index
-                    );
-                }
-            }
+            let transactions = transaction_results
+                .into_iter()
+                .map(|tx| tx.expect("Missing committed transaction during recovery"))
```
piotrm50: The expect message is a bit confusing - are we expect missing  transaction during recovery? 

```suggestion
                .map(|tx| tx.expect("all transactions should be available during recovery"))
```

crates/starfish/core/src/commit_observer.rs
```diff
@@ -326,70 +428,46 @@ impl CommitObserver {
...
+            committed_subdags.push(committed_subdag);
         }
 
-        info!(
-            "Commit observer recovery completed, took {:?}",
-            now.elapsed()
-        );
+        let sent_indices = self
+            .send_sub_dags(&committed_subdags, source)
+            .expect("Failed to send commits during recovery");
```
piotrm50:
```suggestion
            .expect("should send commits during recovery");
```

crates/starfish/core/src/commit_observer.rs
```diff
@@ -218,105 +219,206 @@ impl CommitObserver {
...
+                pending_for_solidifier.push(pending_sub_dag);
+            }
+        }
+
+        if !pending_for_solidifier.is_empty() {
+            let (solid_sub_dags, _missing) = self
+                .commit_solidifier
+                .try_get_solid_sub_dags(&pending_for_solidifier);
+            self.send_sub_dags(&solid_sub_dags, source)
+                .expect("Failed to send solid commits during recovery");
```
piotrm50:
```suggestion
                .expect("should  successfuly send solid commits during recovery");
```

piotrm50 (approved):
small comments regarding the expect messages

filipdulic (approved):

PR #9883, opened 2026-01-21
feat(indexer): Use watermarks table in pruner

### LINKED ISSUE #7554: [indexer]: revisit pruning logic for non-partitioned tables
## Description

Pruning is a data-modification operation that is non critical in terms of performance. Thus adding indexes like the set that #7534 tried to introduces does not seem warranted, when one considers the implications on maintaining the indexes to the write load of the server.

The existing pruning logic for non-partitioned tables iterates over checkpoints and deletes rows matching that transaction range included in the checkpoint. Considering, that typically checkpoints include up to 4,5 transactions, this is rather suboptimal, requiring multiple queries to span the entire range of transactions in an epoch.

Instead we should make it so that pruning process the transaction range on the basis of an epoch and deletes rows associated with batches of transactions (e.g. 10000)  so as to make the sequential scanning forced without indexing an acceptable candidate.

### LINKED ISSUE #9829: [indexer]: Refactor pruning logic to use watermarks
We should adapt the [Pruner::start](https://github.com/iotaledger/iota/blob/8fe4d49fd573c45965cc69ee43541fa3dd7db056/crates/iota-indexer/src/pruning/pruner.rs#L146) job to read watermarks and prune tables based on it.

tomxey:
# Description of change

Use watermarks table in pruner.
So that every table is pruned independently, and we wait till in-flight reads timeout before we remove data.

Some changes are still necessary for reads to actually check watermarks table before reading actual data, to be done in separate PRs.

fixes https://github.com/iotaledger/iota/issues/9829
fixes #7554 

Standard clippy and existing integration tests

### DIFF
--- a/crates/iota-indexer/src/pruning/pruner.rs
+++ b/crates/iota-indexer/src/pruning/pruner.rs
@@ -5,6 +5,7 @@
 use std::{collections::HashMap, time::Duration};
 
 use serde::{Deserialize, Serialize};
+use strum::IntoEnumIterator;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info};
 
@@ -19,12 +20,23 @@ use crate::{
 };
 
 const UPDATE_WATERMARKS_LOWER_BOUNDS_TASK_INTERVAL: Duration = Duration::from_secs(5);
+/// Delay in milliseconds before pruning data after watermark timestamp.
+/// This delay allows in-flight reads that may be accessing data scheduled for
+/// pruning to complete or timeout, ensuring safe pruning without affecting
+/// active queries.
+#[cfg(any(test, feature = "pg_integration", feature = "shared_test_runtime"))]
+const PRUNING_DELAY_MS: u64 = 1000; // 1 second for tests
+
+#[cfg(not(any(test, feature = "pg_integration", feature = "shared_test_runtime")))]
+const PRUNING_DELAY_MS: u64 = 60000; // 60 seconds for production
+
+/// Maximum number of transactions to prune in a single batch for ByTransaction
+/// strategy
+const MAX_TRANSACTIONS_PER_PRUNE_BATCH: u64 = 100;
 
 pub struct Pruner {
     pub store: PgIndexerStore,
     pub partition_manager: PgPartitionManager,
-    // TODO: we can remove this when pruner logic is updated to use `retention_policies`.
-    pub epochs_to_keep: u64,
     pub retention_policies: HashMap<PrunableTable, u64>,
     pub metrics: IndexerMetrics,
 }
@@ -46,6 +58,7 @@ pub struct Pruner {
     Serialize,
     Deserialize,
     Clone,
+    Copy,
 )]
 #[strum(serialize_all = "snake_case")]
 #[serde(rename_all = "snake_case")]
@@ -63,8 +76,6 @@ pub enum PrunableTable {
     EventStructName,
     EventStructPackage,
 
-    TxAffectedAddresses,
-    TxAffectedObjects,
     TxCallsPkg,
     TxCallsMod,
     TxCallsFun,
@@ -74,9 +85,44 @@ pub enum PrunableTable {
     TxKinds,
     TxRecipients,
     TxSenders,
+    TxWrappedOrDeletedObjects,
 
     Checkpoints,
     PrunerCpWatermark,
+}
+
+/// Represents how a table is pruned
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PruningStrategy {
+    /// Drop partition by epoch number (for epoch-partitioned tables)
+    ByEpochPartition,
+    /// Delete rows by checkpoint number
+    ByCheckpoint,
+    /// Delete rows by transaction sequence number
+    ByTransaction,
+}
+
+/// Represents a specific chunk of data to be pruned
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PruningChunk {
+    /// Prune an entire epoch partition
+    Epoch(u64),
+    /// Prune a specific checkpoint
+    Checkpoint(u64),
+    /// Prune a range of transactions [start..=end] inclusive
+    TransactionRange(u64, u64),
+}
+
+impl PruningChunk {
+    /// Returns the start of the next chunk to prune (used for updating
+    /// pruner_hi watermark)
+    fn next_chunk_start(&self) -> u64 {
+        match self {
+            PruningChunk::Epoch(epoch) => epoch + 1,
+            PruningChunk::Checkpoint(checkpoint) => checkpoint + 1,
+            PruningChunk::TransactionRange(_, end) => end + 1,
+        }
+    }
 }
 
 impl PrunableTable {
@@ -95,8 +141,6 @@ impl PrunableTable {
             | PrunableTable::EventStructModule
             | PrunableTable::EventStructName
             | PrunableTable::EventStructPackage
-            | PrunableTable::TxAffectedAddresses
-            | PrunableTable::TxAffectedObjects
             | PrunableTable::TxCallsPkg
             | PrunableTable::TxCallsMod
             | PrunableTable::TxCallsFun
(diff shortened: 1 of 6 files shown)


crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -1400,134 +1354,185 @@ impl PgIndexerStore {
...
-                    min_tx,
-                    max_tx,
-                    "Failed to prune event_struct_package table"
-                ];
-                Ok::<(), IndexerError>(())
-            },
-            PG_DB_COMMIT_SLEEP_DURATION
-        )
-    }
+    /// Prune a single transaction or event index table by transaction range
```
sergiupopescu199:
```suggestion
    /// Prunes a single transaction or event index table by transaction range
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+                        .map(move |start| {
+                            let end = (start + MAX_TRANSACTIONS_PER_PRUNE_BATCH).min(range_end);
+                            PruningChunk::TransactionRange(start, end - 1)
+                        }),
+                )
+            }
         }
     }
+
+    /// Wait for the pruning delay to ensure in-flight reads complete or timeout
```
sergiupopescu199:
```suggestion
    /// Waits for the pruning delay to ensure in-flight reads complete or timeout
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+                );
+            }
+        }
+
+        self.metrics.last_pruned_epoch.set(watermark.epoch_lo - 1);
+
+        Ok(())
+    }
+
+    /// Create an iterator of pruning chunks based on the watermark and pruning
```
sergiupopescu199:
```suggestion
    /// Creates an iterator of pruning chunks based on the watermark and pruning
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+    }
+
+    /// Run the persistent pruning task for this executor's table
+    pub async fn run_pruning_task(self) -> IndexerResult<()> {
+        info!(
+            "Starting persistent pruning task for table {}",
+            self.table.as_ref()
+        );
+
+        let mut interval = tokio::time::interval(Duration::from_secs(5));
```
sergiupopescu199: could we add the `Duration` as a constant in order to have a better familiarity for what's used?

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+                _ = interval.tick() => {
+                    if let Err(e) = self.prune_once_maybe().await {
+                        error!("error pruning table {}: {e}", self.table.as_ref());
+                    }
+                }
+            }
+        }
+    }
+
+    /// Perform one pruning iteration for this executor's table
```
sergiupopescu199:
```suggestion
    /// Performs one pruning iteration for this executor's table
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+        Self {
+            table,
+            store,
+            partition_manager,
+            metrics,
+            cancel,
+        }
+    }
+
+    /// Run the persistent pruning task for this executor's table
```
sergiupopescu199:
```suggestion
    /// Runs the persistent pruning task for this executor's table
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+            let wait_duration = Duration::from_millis(pruning_allowed_timestamp_ms - now_ms);
+            info!(
+                "waiting {}ms for in-flight reads to timeout before pruning table {} (watermark timestamp: {}, delay: {}ms)",
+                wait_duration.as_millis(),
+                self.table.as_ref(),
+                watermark_timestamp_ms,
+                PRUNING_DELAY_MS
+            );
+
+            tokio::select! {
```
sergiupopescu199: This [method](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.run_until_cancelled) seems kinda interesting. 

```rust
self.cancel
    .run_until_cancelled(tokio::time::sleep(wait_duration))
    .await
    .ok_or_else(|| {
        info!("pruning task for table {} cancelled during delay", self.table.as_ref());
        IndexerError::Generic("Pruning task cancelled".to_string())
    })?;
```

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+                    self.table.as_ref()
+                );
+                Box::new(
+                    (pruner_hi..range_end)
+                        .step_by(MAX_TRANSACTIONS_PER_PRUNE_BATCH as usize)
+                        .map(move |start| {
+                            let end = (start + MAX_TRANSACTIONS_PER_PRUNE_BATCH).min(range_end);
+                            PruningChunk::TransactionRange(start, end - 1)
+                        }),
+                )
```
sergiupopescu199, line 358: By checking the requirements of issue #7554, the proposed code seems to already solve it, I'm wondering is the only change we should make is to modify the `MAX_TRANSACTIONS_PER_PRUNE_BATCH=10_000` to achieve the desired result the issue is describing.

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -105,9 +149,294 @@ impl PrunableTable {
...
+                    if let Err(e) = self.prune_once_maybe().await {
+                        error!("error pruning table {}: {e}", self.table.as_ref());
+                    }
+                }
+            }
+        }
+    }
+
+    /// Perform one pruning iteration for this executor's table
+    async fn prune_once_maybe(&self) -> IndexerResult<()> {
```
sergiupopescu199: Could we adjust the name of the function`prune_once_maybe`? first time I read it I thought that it will return an `Option`. We could rename it `prune` or `check_and_prune`.

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: done in 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: Looks neat, done in 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: done in 5ea7971031

crates/iota-indexer/src/pruning/pruner.rs
tomxey: I've raised the batch size to 1000 and also introduced a small delay between chunks to relieve the DB a bit, in 5ea7971031 .
Do you think it will be enough?
(I'm thinking that we could also prune checkpoint-based tables in batches, but I wanted to limit the scope of this PR so it's not here)

kodemartin (approved):
lgtm with the suggestion of using batches also when we prune by checkpoints. This can be delegated into a subsequent issue though.

sergiupopescu199 (approved):
lgtm ✨ 
Awesome implementation!!

PR #9884, opened 2026-01-21
chore(workflows): simplify release workflow by reusing version and ref

Dr-Electron:
# Description of change

This should make things a little bit simpler I think. Not sure if we want or need it 😆

### DIFF
--- a/scripts/homebrew/update_formula.sh
+++ b/scripts/homebrew/update_formula.sh
@@ -3,8 +3,9 @@
 ROOT=$(git rev-parse --show-toplevel || realpath "$(dirname "$0")/../..")
 
 tag=$1
-org=${2:-"iotaledger"}
-repository=${3:-"iota"}
+version=$2
+org=${3:-"iotaledger"}
+repository=${4:-"iota"}
 
 if [ -z "$GH_TOKEN" ]; then
     echo "Environment variable GH_TOKEN must be set!"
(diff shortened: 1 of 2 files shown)


bingyanglin (approved):

DaughterOfMars:
It's not clear to me that this is simpler 😆

Dr-Electron:
> It's not clear to me that this is simpler 😆

Haha fair. But it reuses the same stuff instead of recalculating it 👍

scripts/homebrew/update_formula.sh
```diff
@@ -3,16 +3,16 @@
...
+org=${3:-"iotaledger"}
+repository=${4:-"iota"}
 
 if [ -z "$GH_TOKEN" ]; then
     echo "Environment variable GH_TOKEN must be set!"
     exit 1
 fi
 
 # Remove leading `v` from tag
-version=$(echo "${tag}" | sed -En "s|v?(.+)|\1|p")
```
Alex6323, line 15: need to remove the commment as well then ;)

scripts/homebrew/update_formula.sh
```diff
@@ -3,16 +3,16 @@
...
+version=$2
+org=${3:-"iotaledger"}
+repository=${4:-"iota"}
 
 if [ -z "$GH_TOKEN" ]; then
     echo "Environment variable GH_TOKEN must be set!"
     exit 1
 fi
 
 # Remove leading `v` from tag
```
Dr-Electron:
```suggestion
```

thibault-martinez:
Not going to do this in the end

PR #9885, opened 2026-01-22
feat(starfish): adjust commit range during request

### LINKED ISSUE #9822: Auto-extend commit range for proof fetching in FastCommitSyncing
Automatically extend the commit range during FastCommitSyncing to ensure proof is fetched successfully.

VorobyevIlya:
# Description of change

This PR modifies the commit fetching mechanism during the fast sync by allowing the extension of the commit range. It may be useful when for the last commit in the initial range we don't have enough voting headers, but a quorum exists for one of the later commits. The commit range shouldn't be extended to more than twice the initial length.

Fix #9822 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -1017,6 +1017,27 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         else {
             return Ok((vec![], vec![]));
         };
+        let commit_range_length = new_commit_end - commit_range.start() + 1;
+        match commit_sync_type {
+            CommitSyncType::Regular => {
+                if commit_range_length > batch_size as CommitIndex {
+                    return Err(ConsensusError::TooManyCommitsReceived {
+                        count: commit_range_length,
+                        limit: batch_size as CommitIndex,
+                        sync_type: "regular",
+                    });
+                }
+            }
+            CommitSyncType::Fast => {
+                if commit_range_length > 2 * batch_size as CommitIndex {
+                    return Err(ConsensusError::TooManyCommitsReceived {
+                        count: commit_range_length,
+                        limit: 2 * batch_size as CommitIndex,
+                        sync_type: "fast",
+                    });
+                }
+            }
+        }
 
         // Then scan commits up to the certifiable index
         let commits = self
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -984,25 +984,28 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
 
         // Bound the range based on sync type.
         let batch_size = commit_sync_type.commit_sync_batch_size(&self.context);
-        let inclusive_end = commit_range
+        let inclusive_bound = commit_range
             .end()
             .min(commit_range.start() + batch_size as CommitIndex - 1);
-        // First, find the highest certifiable commit index
-        let Some((highest_index, certifier_block_refs)) = self
-            .find_highest_certifiable_commit_in_range(
-                &commit_range,
-                inclusive_end,
-                &commit_sync_type,
-            )?
+
+        // Find certifiable commit based on sync type
+        let find_certifiable_commit = |commit_sync_type: &CommitSyncType| -> ConsensusResult<Option<(CommitIndex, Vec<BlockRef>)>> {
+            match commit_sync_type {
+                CommitSyncType::Regular => self.find_highest_certifiable_commit_in_range(&commit_range, inclusive_bound, commit_sync_type),
+                CommitSyncType::Fast => self.find_lowest_certifiable_commit_from(inclusive_bound, commit_sync_type),
+            }
+        };
+
+        let Some((new_commit_end, certifier_block_refs)) =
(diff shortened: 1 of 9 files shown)


filipdulic (approved):

crates/starfish/core/src/authority_service.rs
```diff
@@ -452,6 +452,117 @@ impl<C: CoreThreadDispatcher> AuthorityService<C> {
...
+                }
+            }
+        }
+    }
+
+    /// Finds the lowest commit index from search_from that can be certified
+    /// with available votes. Returns the lowest certifiable commit index and
+    /// the block refs (votes) that certify it, or None if no certifiable
+    /// commit is found.
+    fn find_lowest_certifiable_commit_from(
```
polinikita, line 521: should we stop at the point which is not larger than the batch size in this function? This will save computational costs for us and will not require handling a larger interval with potential error output

crates/starfish/core/src/authority_service.rs
```diff
@@ -452,6 +452,117 @@ impl<C: CoreThreadDispatcher> AuthorityService<C> {
             .with_label_values(&[peer_hostname, "handle_subscribed_block_bundle"])
             .inc_by(digests_to_exclude.len() as u64);
     }
+
+    /// Finds the highest commit index in the commit range up to search_up_to
+    /// that can be certified with available votes. Returns the highest
+    /// certifiable commit index and the block refs (votes) that certify it,
+    /// or None if no certifiable commit is found.
+    fn find_highest_certifiable_commit_in_range(
```
polinikita, line 460: In place where we use this, this could imply that the range for regular commit syncer is increased (up to batch_size), while it is written in comments
`   // The range for returned trusted commits starts at the same index, but the end
    // can be different, bigger for fast sync and smaller for regular.`

crates/starfish/core/src/authority_service.rs
```diff
@@ -885,80 +1000,49 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
-                stake_aggregator.add(v.author, &self.context.committee);
+        let Some((new_commit_end, certifier_block_refs)) =
+            find_certifiable_commit(&commit_sync_type)?
+        else {
+            return Ok((vec![], vec![]));
+        };
+        let commit_range_length = new_commit_end - commit_range.start() + 1;
+        match commit_sync_type {
+            CommitSyncType::Regular => {
+                if commit_range_length > batch_size as CommitIndex {
```
polinikita, line 1031: How is this possible given we bounded above as follows:
```
let inclusive_bound = commit_range
            .end()
            .min(commit_range.start() + batch_size as CommitIndex - 1);
```

crates/starfish/core/src/authority_service.rs
```diff
@@ -885,80 +1000,49 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
-                    .inc();
-                commits.pop();
-                // Continue searching from index - 1
-                search_up_to = index.saturating_sub(1);
-                if search_up_to < commit_range.start() {
-                    commits.clear();
-                    break 'commit;
+            CommitSyncType::Fast => {
+                if commit_range_length > 2 * batch_size as CommitIndex {
+                    return Err(ConsensusError::TooManyCommitsReceived {
```
polinikita: Not sure that the error is descriptive in this case. We might artificially increase the range by scanning, while the commit range in the request was correct

polinikita: Based on transaction handling, we gracefully truncate the ranges if a large one is requested by applying the local parameters

polinikita (changes requested):
Left comments. Also, it seems that we don't control how many commits we receive (e.g. 2 * fetch size). If it is too many, then we might need to reject the response we receive

crates/starfish/core/src/authority_service.rs
VorobyevIlya: When we define search_up_to we take the minimum with the includive end of the commit_range

crates/starfish/core/src/authority_service.rs
VorobyevIlya: we can do this. I added the additional check

crates/starfish/core/src/authority_service.rs
VorobyevIlya: Do you mean that it is redundant? I agree, it is just a sanity check, and cannot be really reached.

crates/starfish/core/src/authority_service.rs
VorobyevIlya: We shouldn't increase to more than 2*batch_size anyway, if the node is in the valid state and our assumptions are correct. So it should also be unreachable. I don't see how we can truncate here, because we need the certificates for the last commit. How can we truncate without proofs?

VorobyevIlya:
> Left comments. Also, it seems that we don't control how many commits we receive (e.g. 2 * fetch size). If it is too many, then we might need to reject the response we receive

I changed the error description and added the check on the receiving end

crates/starfish/core/src/authority_service.rs
polinikita: We can truncate the original range only based on our local parameters.

polinikita (approved):

PR #9909, opened 2026-01-23
feat(indexer): Implement backgorund task for reading watermarks table

### LINKED ISSUE #9899: [indexer]: Implement backgorund task for reading `watermarks` table
To not hurt performance reading of `watermarks` table during user queries could be achieved by running a background task that fetches from watermarks task periodically and stores the result in memory, so that all reads can access it right away

tomxey:
# Description of change

To not hurt performance reading of watermarks table during user queries could be achieved by running a background task that fetches from watermarks task periodically and stores the result in memory, so that all reads can access it right away

fixes https://github.com/iotaledger/iota/issues/9899

### DIFF
--- a/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
+++ b/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
@@ -5,8 +5,12 @@
 use std::time::Duration;
 
 use iota_indexer::{
-    apis::GovernanceReadApi, db::ConnectionPoolConfig, metrics::IndexerMetrics,
-    pruning::watermark_task::WatermarkTask, read::IndexerReader, store::PgIndexerStore,
+    apis::GovernanceReadApi,
+    db::ConnectionPoolConfig,
+    metrics::IndexerMetrics,
+    pruning::watermark_task::{WatermarkCache, WatermarkTask},
+    read::IndexerReader,
+    store::PgIndexerStore,
 };
 use iota_json_rpc_types::Stake as RpcStakedIota;
 use iota_types::{
@@ -43,12 +47,13 @@ impl PgManager {
         let connection_pool = iota_indexer::db::new_connection_pool(&db_url.into(), &config)
             .map_err(|e| Error::Internal(format!("Failed to create connection pool: {e}")))?;
 
-        // Create store and watermark task for pruning support
+        // Create store and watermark cache for pruning support
         let store = PgIndexerStore::new(connection_pool.clone(), indexer_metrics);
-        let watermark_task = WatermarkTask::new(store);
+        let watermark_cache = WatermarkCache::new();
 
         // Start watermark task with cancellation token
-        let watermark_cache = watermark_task.start(cancellation_token);
+        let watermark_task = WatermarkTask::new(store, watermark_cache.clone());
+        watermark_task.start(cancellation_token);
 
         // Create reader with watermark cache
         Ok(IndexerReader::new(connection_pool, watermark_cache))
--- a/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
+++ b/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
@@ -4,12 +4,16 @@
 
 use std::time::Duration;
 
-use iota_indexer::{apis::GovernanceReadApi, db::ConnectionPoolConfig, read::IndexerReader};
+use iota_indexer::{
+    apis::GovernanceReadApi, db::ConnectionPoolConfig, metrics::IndexerMetrics,
+    pruning::watermark_task::WatermarkTask, read::IndexerReader, store::PgIndexerStore,
+};
 use iota_json_rpc_types::Stake as RpcStakedIota;
 use iota_types::{
     governance::StakedIota as NativeStakedIota,
     iota_system_state::iota_system_state_summary::IotaSystemStateSummary as NativeIotaSystemStateSummary,
 };
+use tokio_util::sync::CancellationToken;
 
 use crate::error::Error;
 
@@ -28,12 +32,27 @@ impl PgManager {
         db_url: impl Into<String>,
         pool_size: u32,
         timeout_ms: u64,
+        indexer_metrics: IndexerMetrics,
+        cancellation_token: CancellationToken,
     ) -> Result<IndexerReader, Error> {
         let mut config = ConnectionPoolConfig::default();
         config.set_pool_size(pool_size);
         config.set_statement_timeout(Duration::from_millis(timeout_ms));
-        IndexerReader::new_with_config(db_url, config)
-            .map_err(|e| Error::Internal(format!("Failed to create reader: {e}")))
+
+        // Create connection pool
+        let connection_pool = iota_indexer::db::new_connection_pool(&db_url.into(), &config)
+            .map_err(|e| Error::Internal(format!("Failed to create connection pool: {e}")))?;
+
+        // Create store and watermark task for pruning support
+        let store = PgIndexerStore::new(connection_pool.clone(), indexer_metrics);
+        let watermark_task = WatermarkTask::new(store);
+        let watermark_cache = watermark_task.cache();
+
+        // Start watermark task with cancellation token
+        watermark_task.start(cancellation_token);
+
+        // Create reader with watermark cache
+        Ok(IndexerReader::new(connection_pool, watermark_cache))
     }
 }
 
--- a/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
+++ b/crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
@@ -46,10 +46,9 @@ impl PgManager {
         // Create store and watermark task for pruning support
         let store = PgIndexerStore::new(connection_pool.clone(), indexer_metrics);
         let watermark_task = WatermarkTask::new(store);
-        let watermark_cache = watermark_task.cache();
 
         // Start watermark task with cancellation token
-        watermark_task.start(cancellation_token);
+        let watermark_cache = watermark_task.start(cancellation_token);
 
         // Create reader with watermark cache
         Ok(IndexerReader::new(connection_pool, watermark_cache))
--- a/crates/iota-indexer/src/indexer.rs
+++ b/crates/iota-indexer/src/indexer.rs
@@ -24,7 +24,7 @@ use crate::{
     },
     metrics::IndexerMetrics,
     processors::processor_orchestrator::ProcessorOrchestrator,
-    pruning::{optimistic_pruner::OptimisticPruner, pruner::Pruner},
+    pruning::{optimistic_pruner::OptimisticPruner, pruner::Pruner, watermark_task::WatermarkTask},
     read::IndexerReader,
     store::{IndexerAnalyticalStore, IndexerStore, PgIndexerStore},
 };
@@ -172,7 +172,11 @@ impl Indexer {
             env!("CARGO_PKG_VERSION")
         );
 
-        let mut read = IndexerReader::new(connection_pool.clone());
+        // Create and start the watermark task that will track pruning state
+        let watermark_task = WatermarkTask::new(store.clone());
+        let watermark_cache = watermark_task.cache();
+
+        let mut read = IndexerReader::new(connection_pool.clone(), watermark_cache);
 
         if let HistoricFallbackOptions {
             fallback_kv_url: Some(ref url),
@@ -195,7 +199,7 @@ impl Indexer {
             info!("No config for HistoricalFallbackReader provided, skipping...");
         }
 
-        let handle = build_json_rpc_server(store, registry, read, config, metrics)
+        let handle = build_json_rpc_server(store, registry, read, config, metrics, watermark_task)
             .await
             .expect("json rpc server should not run into errors upon start.");
         tokio::spawn(async move { handle.stopped().await })
@@ -204,6 +208,7 @@ impl Indexer {
 
         Ok(())
     }
+
     pub async fn start_analytical_worker<
         S: IndexerAnalyticalStore + Clone + Send + Sync + 'static,
     >(
--- a/crates/iota-indexer/src/indexer.rs
+++ b/crates/iota-indexer/src/indexer.rs
@@ -25,10 +25,13 @@ use crate::{
     metrics::IndexerMetrics,
     processors::processor_orchestrator::ProcessorOrchestrator,
     pruning::{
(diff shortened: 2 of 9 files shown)


crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
```diff
@@ -28,12 +32,27 @@ impl PgManager {
...
+        let connection_pool = iota_indexer::db::new_connection_pool(&db_url.into(), &config)
+            .map_err(|e| Error::Internal(format!("Failed to create connection pool: {e}")))?;
+
+        // Create store and watermark task for pruning support
+        let store = PgIndexerStore::new(connection_pool.clone(), indexer_metrics);
+        let watermark_task = WatermarkTask::new(store);
+        let watermark_cache = watermark_task.cache();
+
+        // Start watermark task with cancellation token
+        watermark_task.start(cancellation_token);
```
sergiupopescu199: Seems like `watermark_task.start` already returns the cache handler.

```suggestion
        // Start watermark task with cancellation token
        let watermark_cache = watermark_task.start(cancellation_token);
```

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2025 IOTA Stiftung
```
sergiupopescu199:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl WatermarkCache {
+    /// Creates a new empty watermark cache
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Gets the watermark for a specific entity
+    /// Returns None if the entity has no watermark
```
sergiupopescu199:
```suggestion
    /// Returns `None` if the entity has no watermark
```

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
...
+    ingestion::common::persist::CommitterTables,
+    models::watermarks::StoredWatermark,
+    store::{IndexerStore, PgIndexerStore},
+};
+
+/// How often to refresh watermarks from the database
+const WATERMARK_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
+
+/// In-memory cache of pruning watermarks
+/// Uses ArcSwap for lock-free reads to avoid write starvation under high read
```
sergiupopescu199: In this case we could use rust doc link to the struct

Something like this: 

```rust
/// Uses [`ArcSwap`](ArcSwap) for lock-free reads to avoid write starvation under high read 
```

Maybe also we could enrich the docs based on this comment: https://github.com/iotaledger/iota/pull/9909/changes#r2721727632 in order to quickly understand the purpose of the struct.

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Gets the watermark for a specific entity
+    /// Returns None if the entity has no watermark
+    pub fn get(&self, entity: CommitterTables) -> Option<StoredWatermark> {
+        let cache = self.inner.load();
+        cache.watermarks.get(entity.as_ref()).cloned()
+    }
+
+    /// Updates the cache with fresh watermarks from the database
+    /// Uses ArcSwap for lock-free reads - writes create new Arc instead of
```
sergiupopescu199: similar to this comment: https://github.com/iotaledger/iota/pull/9909/changes#r2721696759

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
...
+            store,
+            cache: WatermarkCache::new(),
+            update_interval: WATERMARK_UPDATE_INTERVAL,
+        }
+    }
+
+    /// Gets a handle to the watermark cache for reading
+    pub fn cache(&self) -> WatermarkCache {
+        self.cache.clone()
+    }
```
sergiupopescu199: The current rustdoc doesn't clearly explain that cloned WatermarkCache instances share the same underlying data and will reflect updates from the background task. Could we please elaborate more on that?

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Fetches watermarks from the database and updates the cache
+    async fn update_watermarks(&self) -> Result<(), IndexerError> {
+        let (watermarks, _timestamp_ms) = self.store.get_watermarks().await?;
+
+        if !watermarks.is_empty() {
+            info!("Updated {} watermarks from database", watermarks.len());
+        }
+
+        self.cache.update(watermarks);
```
sergiupopescu199, line 129: Mostly a general question: The watermark table will be updated so frequently that essentially every time we query it it's essentially very likely that the data will be different compared to the ones we have in cache? So no need for a comparison between the stored data & fetched one to avoid unnecessary arc swaps?

sergiupopescu199 (commented):
Left some nits but the overall implementation looks really nice! Very clever approach using the `ArcSwap`.

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: Ingestion pipeline will update it every checkpoint so it's almost certain that it will be always different in a comparison here.
I think I prefer to keep it simple and always swap, once few seconds it shouldn't be a big cost.

crates/iota-graphql-rpc/src/context_data/db_data_provider.rs
tomxey: b92c5f0a0c

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: b92c5f0a0c

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: Done in b92c5f0a0c
Please check

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: b92c5f0a0c

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: b92c5f0a0c

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: Added more explanation in b92c5f0a0c

crates/iota-indexer/src/indexer.rs
```diff
@@ -172,7 +172,11 @@ impl Indexer {
             env!("CARGO_PKG_VERSION")
         );
 
-        let mut read = IndexerReader::new(connection_pool.clone());
+        // Create and start the watermark task that will track pruning state
+        let watermark_task = WatermarkTask::new(store.clone());
+        let watermark_cache = watermark_task.cache();
```
kodemartin: You do create it but don't actually start it here.

Instead you delegate it inside `build_json_rpc_server`. I don't understand why, though?

Would it be more transparent if you kept scopes simple and separate and instead

* Create the cache (which is what you want to read eitherway) explicitly here through `WatermarkCache::new`.
* Pass the cache to the `Task::new`. This makes sense as it is a requirement for the task to update it. This indirection of creating it as a side-effect inside the task and then exposing a method to return it seems unnecessary.
* Start the task independently of the `json_rpc_server`.

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,181 @@
...
+#[derive(Clone)]
+pub struct WatermarkCache {
+    inner: Arc<ArcSwap<WatermarkCacheInner>>,
+}
+
+#[derive(Default)]
+struct WatermarkCacheInner {
+    /// Map from entity name to its watermark data
+    watermarks: HashMap<String, StoredWatermark>,
+}
```
kodemartin: Why not have a type alias instead? `WatarmarkCacheInner = HashMap<String, StoredWatermark>`

sergiupopescu199 (approved):
lgtm ✨

crates/iota-indexer/src/pruning/watermark_task.rs
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Background task that periodically fetches pruning watermarks from the
+//! database and maintains them in memory for RPC reads to check data
+//! availability.
+
+use std::{collections::HashMap, sync::Arc, time::Duration};
+
+use arc_swap::ArcSwap;
```
kodemartin, line 10: Just to make sure: Have you taken into account the [caveats](https://docs.rs/arc-swap/latest/arc_swap/index.html#before-using) of using `ArcSwap`?

As mentioned this is a niece structure, and furthermore introduces a new API that comes with less common pitfalls than the widely used lock primitives (e.g. `RwLock`). Pitfalls elude the compiler and thus are left to the implementation for avoiding them making maintenance slightly more complex.

In this case, especially the limitations of [load](https://docs.rs/arc-swap/latest/arc_swap/docs/limitations/index.html#too-many-guards) should be kept in mind while using it afterwards.

kodemartin (commented):
Looks good with a few minor suggestions.

Plus a more structural general one. The watermarks table is now used for synchronizing between several processes

1. The writer part, which updates `checkpoint_inclusive_hi`
2. The task that updates `reader_lo, epoch_lo` upon conditional change of (1) in the context of pruning
3. The pruning tasks that deletes data based on `reader_lo` updated by (2), and now 
4. This cache that stores in memory `reader_lo` upon changes of (2) for any reading process.

All these happen through polling, which might be simple in the implementation but it makes it implicitly hard to fine tune (consider the different polling intervals) synchronization between the different processes , and to eventually guarantee certain actions on a happens-before or after conditions. 

It seems thus plausible that at a subsequent iteration we switch to an event-based synchronization pattern, where different processes use the `LISTEN/NOTIFY` to get updates, and poll only upon restart to ensure reliability.

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: Actually I didn't. Thanks for pointing that out.

But after taking a read of the docs, it looks rather acceptable to me:

From [docs](https://docs.rs/arc-swap/latest/arc_swap/docs/performance/index.html#choosing-the-right-reading-operation)
```
The number of borrows each thread can have at each time (across all [Guard](https://docs.rs/arc-swap/latest/arc_swap/struct.Guard.html)s) is limited. If this limit is exceeded, an owned instance is created instead.

Therefore, if you intend to hold onto the loaded value for extended time span, you may prefer [load_full](https://docs.rs/arc-swap/latest/arc_swap/struct.ArcSwapAny.html#method.load_full). It loads the pointer instance ([Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html)) without borrowing, which is slower (because of the possible contention on the reference count), but doesn’t consume one of the borrow slots, which will make it more likely for following [load](https://docs.rs/arc-swap/latest/arc_swap/struct.ArcSwapAny.html#method.load)s to have a slot available.
```

If the "slow" path is to clone the Arc but still not do any locking then I think this is fine.
We are doing quick reads anyway, just get one value from the hashmap and exit immediately, I think 8 simultaneous reads per thread should be sufficient in most cases.
(can we actually have 8 reads from the same thread if this is async code that is doing the read, and there are no yield points during the read?, not sure if the RPC server is spawning many threads or how is it handling the requests)

crates/iota-indexer/src/indexer.rs
tomxey: Applied the refactor in 45bd6a3729
The task is still started in `build_json_rpc_server` since we are also starting the system package task there and it seemed consistent to me. We also have the cancellation token created there so I didn't have to move it's creation here.
Let me know what you think.

crates/iota-indexer/src/pruning/watermark_task.rs
tomxey: done in 45bd6a3729

tomxey:
@kodemartin 
As for using the `LISTEN/NOTIFY`, it would make the whole process faster, but given that this is a pruning process, we don't really have to hurry about it.
I think implementation using "LISTEN/NOTIFY" would complicate the code so I'm not sure it's worth it.

> and to eventually guarantee certain actions on a happens-before or after conditions

do you have some specific action on your mind?

crates/iota-indexer/src/pruning/watermark_task.rs
kodemartin: Cloning the `Arc` through `load_full` is actually what saves you from the 8 slot cap per thread. Not that these are CPU threads not tokio threads.

So this is definitely something to consider while implementing the reads.

kodemartin:
> @kodemartin As for using the `LISTEN/NOTIFY`, it would make the whole process faster, but given that this is a pruning process, we don't really have to hurry about it. I think implementation using "LISTEN/NOTIFY" would complicate the code so I'm not sure it's worth it.
> 
> > and to eventually guarantee certain actions on a happens-before or after conditions
> 
> do you have some specific action on your mind?

The main concern I tried to raise is not about efficiency. Although it does seem like an overkill to poll every few seconds for events that occur in the scale of a day.

My concern is about complexity in tuning and actually synchronizing between the processes. In general having to align between different polling intervals when reasoning about order of events between process seems harder than guarantees that the event-based pattern would provide. For example, notifications on a change in `reader_lo` would be sent to all interested processes effectively at the same time.

As I pointed out originally, I agree that implementation is more complex. It is maintaining and extending synchronization logic on the base pattern that is made simpler with events.

crates/iota-indexer/src/indexer.rs
kodemartin: I think that it would make sense to take the system-package task out as well for the same reasons I laid out above. Having all tasks created upfront makes it more transparent for the reader.

tomxey: Moved in c346c65cef

kodemartin (approved):
lgtm 🌸

tomxey:
Created an issue for the listen/notify: https://github.com/iotaledger/iota/issues/9943

PR #9912, opened 2026-01-23
fix(starfish): add source for block headers and don't update cordial knowledge when commit syncing

### LINKED ISSUE #9911: starfish: fix unbounded cordial_knowledge channel buffering during syncing
The validator node could crash during commit syncing due to buffering in the unbounded cordial_knowledge channel.

We need either avoid using unbounded channel here or make regular pulling

Image

polinikita:
# Description of change

This PR fixes a problem of potentially growing RAM in the unbounded cordial knowledge channel, when performing commit syncing. The fix is two-fold:
 1. We add a source for block headers (for full blocks in streaming, block headers in bundles in streaming, header synchronizer and commit syncer). When the source is commit syncer, we don't update the cordial knowledge with such messages. Such messages are anyway outdated and will not improve cordial dissemination.
 2. We change the unbounded channel to bounded (with capacity 10_000) and use optimistic try_send. This is fine since the cordial knowledge can tolerate skips of block headers and will not panic. This in turn forces us to add a watcher channel for eviction for which we can't so simply skip messages (o/w this could lead to the unlimited grow of connection knowledge structures). 

As a benefit of this small refactor, we can now observe in metrics the number of headers contributed through each of the potential sources.

Fixes #9911 

- [x] I have added tests that prove my fix is effective or that my feature works: The fix was validated on the local deployment.

### DIFF
--- a/crates/starfish/core/src/block_manager/mod.rs
+++ b/crates/starfish/core/src/block_manager/mod.rs
@@ -29,7 +29,7 @@ use crate::{
     },
     block_manager::block_suspender::BlockSuspender,
     context::Context,
-    dag_state::{DagState, TransactionSource},
+    dag_state::{BlockHeaderSource, DagState, TransactionSource},
 };
 
 /// Block manager suspends incoming blocks until they are connected to the
@@ -84,6 +84,7 @@ impl BlockManager {
         self.write_block_headers_and_transactions_to_dag_state(
             block_headers_to_accept.clone(),
             accepted_transactions,
+            BlockHeaderSource::BlockStreaming,
         );
 
         (block_headers_to_accept, missing_block_headers)
@@ -99,6 +100,7 @@ impl BlockManager {
     pub(crate) fn try_accept_block_headers(
         &mut self,
         block_headers: Vec<VerifiedBlockHeader>,
+        source: BlockHeaderSource,
     ) -> (Vec<VerifiedBlockHeader>, BTreeSet<BlockRef>) {
         let _s = monitored_scope("BlockManager::try_accept_block_headers");
         // Headers are added through synchronizer, commit syncer and cordial
@@ -110,6 +112,7 @@ impl BlockManager {
         self.write_block_headers_and_transactions_to_dag_state(
             block_headers_to_accept.clone(),
             accepted_transactions,
+            source,
         );
         (block_headers_to_accept, ancestors_to_fetch)
     }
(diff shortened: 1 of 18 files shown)


crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -28,6 +31,16 @@ use crate::{
 /// relevant. 40 rounds correspond to at least 2 second due to the minimum block
 /// delay
 const MAX_ROUND_GAP_FOR_USEFUL_PARTS: Round = 40;
+/// Capacity of the cordial knowledge channel. For normal operation with
+/// 100 authorities, this allows buffering up to 5 seconds of headers at 200ms
```
VorobyevIlya: do you mean 20 blocks/second?

filipdulic (approved):

VorobyevIlya (approved):

crates/starfish/core/src/cordial_knowledge.rs
polinikita:
```suggestion
/// 100 authorities, this allows buffering up to 5 seconds of headers at 20
```

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -84,6 +84,7 @@ impl BlockManager {
         self.write_block_headers_and_transactions_to_dag_state(
             block_headers_to_accept.clone(),
             accepted_transactions,
+            BlockHeaderSource::BlockStreaming,
```
piotrm50, line 87: This method is always called with block streaming? I think it's better to have consistency of function signatures and add source as a parameter here as well, instead of having hard-coded source in the function body to avoid problems in the future in case something changes.

crates/starfish/core/src/cordial_knowledge.rs
```diff
@@ -357,6 +390,24 @@ impl CordialKnowledge {
...
 
+    fn append_eviction_msgs_if_changed(
+        &mut self,
+        vec_connection_knowledge_msgs_batch: &mut [Vec<ConnectionKnowledgeMessage>],
+    ) {
+        if !self.eviction_rounds_receiver.has_changed().unwrap_or(false) {
+            return;
+        }
+        let evicted_rounds = self.eviction_rounds_receiver.borrow_and_update().clone();
+        if evicted_rounds.len() != self.context.committee.size() {
```
piotrm50, line 401: Isn't that a problem worth a `warn` log at least?

piotrm50 (approved):
Left 2 small comments.

crates/starfish/core/src/block_manager/mod.rs
polinikita: Ok, Agree

polinikita: I will this update in a separate PR. There are a few more thing that we can make more jointly for transaction source.

crates/starfish/core/src/cordial_knowledge.rs
polinikita: Added in https://github.com/iotaledger/iota/pull/9930

crates/starfish/core/src/block_manager/mod.rs
polinikita: https://github.com/iotaledger/iota/pull/9930

PR #9914, opened 2026-01-26
feat(consensus, sc-platform): Score integration types refactoring

oliviasaa:
# Description of change

Three types used in the validator scoring mechanism were defined separately with duplicated field lists, despite representing the same information in different storage formats. This PR consolidates them into a  single generic struct (`MisbehaviorsV1<T>`) with type aliases for each use case, so that field definitions and ordering are maintained in one place.
 
Summary of changes:
  - Extract `MisbehaviorsV1<T>` from `messages_consensus.rs` into its own module (`misbehavior_counts`)                 
  - Replace the old `ScoringMetricsV1` struct with a type alias `MisbehaviorsV1<Vec<AtomicU64>>`, eliminating the duplicated field definitions      
  - Reuse `VersionedScoringMetrics` for per-authority persistent storage.
  - Storage model changed from per-authority entries keyed by `AuthorityIndex` to a single blob containing all authorities' metrics.
  - Add versioned update methods to `VersionedScoringMetrics` and `MysticetiScoringMetricsStore`                                                                                                                     
  - Add iterator tests for `MisbehaviorsV1`                                                                                                                                                                    


consensus/core/src/storage/store_tests.rs
```diff
@@ -304,22 +304,22 @@ async fn read_and_scan_commits(
...
-            faulty_blocks_provable: 0,
-            faulty_blocks_unprovable: 0,
-            equivocations: 0,
-            missing_proposals: 0,
-        },
+        VersionedStorageScoringMetrics::new_v1_for_test(
+            1, // faulty_blocks_provable
+            2, // faulty_blocks_unprovable
+            4, // missing_proposals
+            3, // equivocations
```
bingyanglin: any reason we have 4 before 3?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -84,7 +84,7 @@ impl RocksDBStore {
             Self::COMMITS_CF;<(CommitIndex, CommitDigest), Bytes>,
             Self::COMMIT_VOTES_CF;<(CommitIndex, CommitDigest, BlockRef), ()>,
             Self::COMMIT_INFO_CF;<(CommitIndex, CommitDigest), CommitInfo>,
-            Self::SCORING_METRICS_CF;<AuthorityIndex, StorageScoringMetrics>
+            Self::SCORING_METRICS_CF;<AuthorityIndex, VersionedStorageScoringMetrics>
```
bingyanglin: Node will crash directly when reading persisted scoring metrics after an update? Instead of directly changing the columns of existing `scoring_metrics` table, I guess we need to deprecate the old table and enforce the new data access the new table in next release version first, and then in the next next release version we can directly remove the old table.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -193,33 +193,33 @@ impl Scorer {
...
+//     - Metric M1: [1, 1, 1, 1] (values for authorities A_1, A_2, A_3, A_4)
+//     - Metric M2: [2, 2, 2, 2] (values for authorities A_1, A_2, A_3, A_4)
+// - Report R_3 from authority A_3 with voting power VP_3 = 1:
+//     - Metric M1: [2, 2, 2, 2] (values for authorities A_1, A_2, A_3, A_4)
+//     - Metric M2: [1, 1, 1, 1] (values for authorities A_1, A_2, A_3, A_4)
+//
+// For Metric M1, we have that the median metric vector is [1, 1, 1, 1].
+//
+// This method returns a vector of MedianMetricVec, one per metric in
+// VersionedMisbehaviorReport
```
bingyanglin, line 220: Any reason to change the comment style to be invisible for `cargo doc`?

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    pub fn misbehaviors(&self) -> (&T, &T, &T, &T) {
+        (
+            &self.faulty_blocks_provable,
+            &self.faulty_blocks_unprovable,
+            &self.missing_proposals,
+            &self.equivocations,
+        )
+    }
```
bingyanglin: Dead code?

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -239,57 +241,68 @@ impl VersionedScoringMetrics {
...
+    }
+
+    // Returns an iterator over references to the metric values.
+    pub fn iterate_over_metrics(&self) -> std::vec::IntoIter<&u64> {
+        match self {
+            VersionedStorageScoringMetrics::V1(inner) => inner.iter(),
+        }
+    }
+
+    pub fn new_v1_for_test(
```
bingyanglin: Do we need #[cfg(test)] for it?

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -4,100 +4,120 @@
...
             VersionedScoringMetrics::V1(metrics) => {
-                metrics.equivocations[authority_index].store(value, Ordering::Relaxed);
+                metrics.equivocations()[authority_index].store(value, Ordering::Relaxed);
             }
         }
     }
 
+    // Validity checks are done at a higher level to ensure authority_index is
+    // valid.
     pub fn store_missing_proposals(&self, authority_index: usize, value: u64) {
```
Thoralf-M: Many of the fns don't even mutate the authority_index, why have the validity check comment for them?

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: Could you explain exactly which case are you concerned about? When the epoch changes and the scoring metrics are initialized, there is nothing to read from the storage (unless something changed _very_ recently).

crates/iota-core/src/authority/scorer.rs
oliviasaa: This is a really long explanation that I'm not sure it belongs to those docs

crates/iota-types/src/scoring_metrics.rs
oliviasaa: Maybe I didn't phrase the comment really well; I just meant them as a (quite obvious) warning that you should make sure `authority_index` is a valid index outside of those functions, because nothing is being checked here, so they will just panic if called with a large number. Do you have a suggestion for this?

crates/iota-types/src/scoring_metrics.rs
oliviasaa: I don't completely understand why, but `#[cfg(test)]` doesn't work with the tests I'm using this function for. They are cross crate tests

consensus/core/src/storage/store_tests.rs
oliviasaa: no, the fields were switched at some point and the numbers remained the same. Fixed in [be13ea7](https://github.com/iotaledger/iota/pull/9914/commits/be13ea7e60e40eefc93328042aae9223e9ca1fb7)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: There are several PRs being built on top of this one, but I think I can remove this function for now. Fixed in [be13ea7](https://github.com/iotaledger/iota/pull/9914/commits/be13ea7e60e40eefc93328042aae9223e9ca1fb7)

crates/iota-types/src/scoring_metrics.rs
Thoralf-M: I get that there are no checks for the `increment_` fns, there the comment makes sense, but the `store_` fns don't mutate the number, so why have the comment for them?

oliviasaa: I don't understand; both fns (the `increment_` and `store_`) would panic if `authority_index` is out of bounds... what's the difference here?

oliviasaa: I can remove the comments, they are too repetitive anyway. I'm just trying to understand if there is something I'm not seeing here

crates/iota-types/src/scoring_metrics.rs
bingyanglin: The best practice should be that, in the `crates/iota-types/Cargo.toml` we add a `test-utils` feature (or other good naming), like
```toml
[features]
...
test-utils = [] 
```
, and here before the `new_v1_for_test` we gate the function by the `#[cfg(any(test, feature = "test-utils"))]`, and in `crates/iota-common/Cargo.toml` we forward this feature by using
```toml
[features]
...
test-utils = ["iota-types/test-utils"]  
```
, then in the `consensus/core/Cargo.toml`, we need
```toml
[dev-dependencies]
...
iota-common = { workspace = true, features = ["test-utils"] } 
```

So the feature can be propagate cross creates accordingly.

Like the `Context::new_for_test()`, `Slot::new_for_test()` we have this pattern, but unfortunately yours needs to cross crates, thus the feature flag is needed.

However, I checked the codebase in other  places, not every test method follow this pattern... So I guess this is good to have but not critical right now.

consensus/core/src/storage/rocksdb_store.rs
bingyanglin: Our nodes will bump version within an epoch. Let's say we are in epoch 100, and the node runs the old version in 00:00 - 12:00, using the old schema. And then we shut down the node and bump the version, so at 12:10 it starts to use the new schema but still in the epoch 100. Then when the node starts to read the metrics in epcoh 100, it tries to deserialize old format, then it will crash. Or do I miss something?

muXxer: I agree, if nodes restart with the new version and scoring was already enabled on these nodes, what happens if old data is read from the same table?
I guess you need to add a `scoring_metrics_v2` instead and mark the old one as `#[deprecated]`. We can remove the old table and the old type in an upcoming version then.

crates/iota-types/src/scoring_metrics.rs
Thoralf-M: Nvm, I thought the value stored there would get out of bounds, so it's fine as it is

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: ok, I think I got it... I didn't know that nodes could bump versions in the middle of an epoch. For some reason I thought this was not supposed to happen

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -188,24 +212,15 @@ impl VersionedScoringMetrics {
...
-                }
-                for (i, value) in report_v1.equivocations.iter().enumerate() {
-                    metrics.equivocations[i].fetch_max(*value, Ordering::Relaxed);
-                }
-                for (i, value) in report_v1.missing_proposals.iter().enumerate() {
-                    metrics.missing_proposals[i].fetch_max(*value, Ordering::Relaxed);
-                }
-            }
+        if !self.has_compatible_version(report) {
+            panic!("Incompatible scorer version");
```
Thoralf-M: Fine to panic here now?

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: I made some changes in [7279129](https://github.com/iotaledger/iota/pull/9914/commits/727912922d9aa4309d277cba4855788113aca1fb). Let me know if this solves the issue completely.

Just a comment about the changes: all those metrics are only added to storage when there is a change in the counts. So suppose that a node recovers the metrics from the "old" storage in epoch E. If a validator does not misbehave in this same epoch, the metrics for this validator will not be added to the "new" storage. For this reason, I'm forcing the metrics to be written in the new CF when we get them from the old one the first time. Just let me know if the way I'm doing this is way too ugly. I see why it would be, but I assume it is reasonable given that it's a one time fix and the old CF will be removed soon.

crates/iota-types/src/scoring_metrics.rs
oliviasaa: No, it's not fine at all, because this version is not being checked anywhere. I'll make some changes.

oliviasaa: made changes here [4a43777](https://github.com/iotaledger/iota/pull/9914/commits/4a43777b5dd1721331ef34e5d73db879ac968265)

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -239,57 +244,68 @@ impl VersionedScoringMetrics {
...
+        }
+    }
 }
 
-pub struct ScoringMetricsV1 {
-    faulty_blocks_provable: Vec<AtomicU64>,
-    faulty_blocks_unprovable: Vec<AtomicU64>,
-    missing_proposals: Vec<AtomicU64>,
-    equivocations: Vec<AtomicU64>,
+// Misbehavior counts using u64, used for storage. Given an authrity, each field
```
Thoralf-M:
```suggestion
// Misbehavior counts using u64, used for storage. Given an authority, each field
```

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -239,57 +244,68 @@ impl VersionedScoringMetrics {
...
+// of this type is a u64 with the metric value for that specific authority.
+type StorageScoringMetricsV1 = MisbehaviorsV1<u64>;
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub enum VersionedStorageScoringMetrics {
```
Thoralf-M: Why have this enum and don't just store the VersionedScoringMetrics?

oliviasaa: We needed to convert the VersionedScoringMetrics to something else while storing anyway, because of the atomic types (I don't remember if the problem was related to cloning or serializing it, tbh... maybe both?).
It was also just natural to store those metrics separately per validator because of how the dag state was already implemented.

Thoralf-M: I'm not convinced, this can be solved with custom serde/clone impls if necessary
Looking at the code it seems like the metrics are always written/read together for all authorities, therefore storing them separated doesn't seem to bring any benefit

Suggested simplification, unless you already have more changes planned for the future where it would benefit from having them separated:

- Add serde support for VersionedScoringMetrics by serializing atomics as u64 values
- Store the entire VersionedScoringMetrics as a single blob instead of per-authority
- Remove VersionedStorageScoringMetrics entirely

This would simplify:

- No new_from() extracting single authority
- No loop filling in missing authorities with zeros
- Single read/write operation instead of N operations
- Less code duplication between the two types

oliviasaa: Actually, the changes planned for the future involve adding this metric to Starfish storage (we don't currently count those metrics in Starfish; they are always zeroed). So what it would benefit the most would be getting this thing right since the start and never changing it again 😅 
I also don't think we need these changes for the next cut, so I'm going to take a look at this. Thanks for the suggestions!

bingyanglin (approved):

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl MisbehaviorsV1<u64> {
+    pub fn new_zeroed() -> Self {
+        Self::new(0, 0, 0, 0)
+    }
+}
+
+impl MisbehaviorsV1<Vec<u64>> {
+    // Verifies that all fields have the expected committee size.
+    pub fn verify(&self, committee_size: usize) -> bool {
```
semenov-vladyslav: I'd call it `validate`. `verify` suggests the values themselves are somehow verified. While `validate` does some trivial validation checks.

crates/iota-types/src/scoring_metrics.rs
oliviasaa: fixed in [b81a4d3](https://github.com/iotaledger/iota/pull/9914/commits/b81a4d3a5e36b2db9e1040c246c8e6f53022b07b)

consensus/core/src/storage/mem_store.rs
```diff
@@ -134,14 +136,12 @@ impl Store for MemStore {
...
-            .iter()
-            .map(|(&authority_index, metrics)| (authority_index, metrics.clone()))
-            .collect::<Vec<_>>();
-        Ok(metrics_by_author)
+        _committee: &Committee,
+    ) -> ConsensusResult<Option<VersionedScoringMetrics>> {
+        match &self.inner.read().scoring_metrics {
+            Some(metrics) => Ok(Some(metrics.snapshot())),
+            None => Ok(None),
+        }
```
semenov-vladyslav:
```suggestion
        Ok(self.inner.read().scoring_metrics.map(VersionedScoringMetrics::snapshot))
```

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -78,13 +85,22 @@ impl RocksDBStore {
...
-        let (blocks, digests_by_authorities, commits, commit_votes, commit_info, scoring_metrics) = reopen!(&rocksdb,
+        let (
+            blocks,
+            digests_by_authorities,
+            commits,
+            commit_votes,
+            commit_info,
+            scoring_metrics,
+            scoring_metrics_v2,
+        ) = reopen!(&rocksdb,
```
semenov-vladyslav, line 88: We should probably switch to using `DBMapUtils::open_tables_read_write`, otherwise removing the deprecated table would require some manual work. Not saying it must be done in this PR though.

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -325,3 +367,167 @@ impl Store for RocksDBStore {
         Ok(Some((CommitRef::new(key.0, key.1), commit_info)))
     }
 }
+
+// TODO: delete this function after the migration is done and the legacy
```
semenov-vladyslav: You can also create a github issue and reference it here so that it's not forgotten accidentally.

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,210 @@
...
+        &self.missing_proposals
+    }
+
+    // Returns a reference to the equivocations field.
+    pub fn equivocations(&self) -> &T {
+        &self.equivocations
+    }
+
+    // Returns an iterator over references to all misbehavior fields.
+    pub fn iter(&self) -> std::vec::IntoIter<&T> {
```
semenov-vladyslav:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = &T> {
```

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,210 @@
...
+        vec![
+            &self.faulty_blocks_provable,
+            &self.faulty_blocks_unprovable,
+            &self.missing_proposals,
+        ]
+        .into_iter()
+    }
+}
+
+impl<T> FromIterator<T> for MisbehaviorsV1<T> {
```
semenov-vladyslav: Are you sure you need this impl. It seems artificial and weird way to construct MisbehaviorsV1.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -4,100 +4,153 @@
...
-// This struct represents the scoring metrics collected by all authorities. They
-// are stored locally by each authority and then converted to a misbehavior
-// report when they share their metrics with the network. When a report is
-// received, it is also used to update a variable of this type stored in the
-// Scorer. Any metric contained in this struct must be guaranteed to be
-// monotonically increasing, because of the way updates are applied from
-// reports.
+// Misbehavior counts using atomic counters, for in-memory concurrent updates.
+// Each field is a `Vec<AtomicU64>` with one entry per authority.
+type ScoringMetricsV1 = MisbehaviorsV1<Vec<AtomicU64>>;
```
semenov-vladyslav: Why `MisbehaviorsV1<Vec<AtomicU64>>` and not `Vec<MisbehaviorsV1<AtomicU64>>`. The latter variant makes more sense to me: it's a single vector instead of four with the invariant that all four vectors must have the same size.

semenov-vladyslav: Scoring metric and misbehavior count are so different terms. It would be nice either to unify them and use one of them, or at least explain that currently the only validator scoring metric we have is the misbehavior counts.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -4,100 +4,153 @@
...
             VersionedScoringMetrics::V1(metrics) => {
-                metrics.missing_proposals[authority_index].fetch_add(increment, Ordering::Relaxed);
+                metrics.missing_proposals()[authority_index]
+                    .fetch_add(increment, Ordering::Relaxed);
             }
         }
     }
 
+    // Validity checks are done at a higher level to ensure authority_index is
+    // valid.
```
semenov-vladyslav:
```suggestion
```
I don't think you need to explain this.

consensus/core/src/scoring_metrics_store.rs
```diff
@@ -158,13 +160,33 @@ impl MysticetiScoringMetricsStore {
             return;
         }
```
semenov-vladyslav, line 164: No, this should be a panic because it indicates internal logic error that may lead to incorrect behavior if silenced.

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,210 @@
...
+        Self::new(
+            vec![0; committee_size],
+            vec![0; committee_size],
+            vec![0; committee_size],
+            vec![0; committee_size],
+        )
+    }
+
+    // Verifies that all fields have the expected committee size.
+    pub fn verify(&self, committee_size: usize) -> bool {
```
semenov-vladyslav: Is there a way to ensure that after the committee changes (in size or the order of validators) the corresponding scoring metrics are updated accordingly. Having this verify function will not means that this does not (or may not) happen.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -193,41 +193,41 @@ impl Scorer {
...
 fn calculate_median_report(
     reports_and_voting_power: &[(VersionedMisbehaviorReport, VotingPower)],
 ) -> MisbehaviorsV1<MedianMetricVec> {
     // Calls to this method should ensure that we have at least one report to
     // process.
     assert!(!reports_and_voting_power.is_empty());
 
-    let number_of_metrics = reports_and_voting_power[0].0.iterate_over_metrics().len();
+    let number_of_metrics = reports_and_voting_power[0].0.iter().len();
```
semenov-vladyslav: That's a very specific way to get the number of metrics? Perhaps there should be a method to do that?

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -188,24 +232,13 @@ impl VersionedScoringMetrics {
...
-                    metrics.missing_proposals[i].fetch_max(*value, Ordering::Relaxed);
-                }
-            }
+        if !self.has_compatible_version(report) {
+            warn!(
+                "Metrics counts being updated according to a report with incompatible version, but report versions were already checked before this point!"
+            );
+        }
+        for (current_value, new_value) in self.iter().flatten().zip(report.iter().flatten()) {
+            current_value.fetch_max(*new_value, Ordering::Relaxed);
```
semenov-vladyslav: This should probably be a method of Misbehaviors. At least you could check that the same version is used and the same number of metrics is used.

semenov-vladyslav (changes requested):

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: fixed in [799ebcb](https://github.com/iotaledger/iota/pull/9914/commits/799ebcbb4606067f0e8bf3721afadc3778759af0)

consensus/core/src/storage/mem_store.rs
oliviasaa: fixed in [799ebcb](https://github.com/iotaledger/iota/pull/9914/commits/799ebcbb4606067f0e8bf3721afadc3778759af0)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: fixed in [799ebcb](https://github.com/iotaledger/iota/pull/9914/commits/799ebcbb4606067f0e8bf3721afadc3778759af0)

crates/iota-types/src/scoring_metrics.rs
oliviasaa: fixed in [799ebcb](https://github.com/iotaledger/iota/pull/9914/commits/799ebcbb4606067f0e8bf3721afadc3778759af0)
I didn't completely remove the comment, but now it's written just once at the beginning

consensus/core/src/scoring_metrics_store.rs
oliviasaa: fixed in [799ebcb](https://github.com/iotaledger/iota/pull/9914/commits/799ebcbb4606067f0e8bf3721afadc3778759af0)

crates/iota-types/src/scoring_metrics.rs
cyberphysic4l: They are all misbehavior counts at the moment but we expect to have misbehavior proofs in the future as well. I think we have only plans to track things that are misbehaviors though, so perhaps this could be changed from ScoringMetrics to MisbehaviorMetrics.

cyberphysic4l (approved):

crates/iota-types/src/scoring_metrics.rs
oliviasaa: I'm actually more annoyed with `MisbehaviorsV1` than `ScoringMetrics`.
I'd change the naming from `MisbehaviorsV1` to `MisbehaviorsCountsV1`, because no matter what we track in the future, what is supposed to be in this struct will always be counts. Then in the future we can add `MisbehaviorsProofs` and then the `VersionedMisbehaviorReport::V2` can include a `MisbehaviorsCountsV2` and a `MisbehaviorsProofsV1` (or `Option` of them).

oliviasaa: Then `ScoringMetricsV2` would include the `MisbehaviorsCountsV2` and counts for the `MisbehaviorsProofs`. 
This is the plan, and I'm totally open to suggestions about the names

oliviasaa: one option is:

- renaming `Misbehaviors` to `UnprovableMisbehaviorsCounts` / `ReportedMisbehaviorsCounts`
- calling whatever holds the proofs to be sent in reports `MisbehaviorsProofs`
- calling the counts derived from the proofs `ProvableMisbehaviorsCounts` / `MisbehaviorsProofCounts`
- renaming `ScoringMetrics` to `MisbehaviorsCounts`. 

This is what makes more sense to me.

However, note that calling provable/unprovable here might a bit confusing right now because we would have a `provable_faulty_blocks` field inside `UnprovableMisbehaviorsCounts` (i.e., the current `Misbehaviors`). But it won't be confusing anymore after we introduce proofs, because then we'll never have any field called "provable" inside `UnprovableMisbehaviorsCounts`

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: I guess we would have to index everything according to `AuthorityName`, maybe? or at least having a copy of the `AuthorityName`s somewhere in the `Scorer`... 
Now, why would the committee change? nothing of this survives epoch changes

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: yeah, I would rather not touch this since I'm not the "owner" of the storage...

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: I added it as a "todo before merging" in the PR description

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: I think it's cleaner to use `collect` when doing type conversions. But "need" is a strong word...

crates/iota-types/src/scoring_metrics.rs
oliviasaa: I'm gonna be honest about this: if I had a strong reason for this, I forgot which it was 😅 
That being said, I would rather not change this now, because then we need to change the version of the reports as well

crates/iota-core/src/authority/scorer.rs
oliviasaa: I added a comment about this in [c99955d](https://github.com/iotaledger/iota/pull/9914/commits/c99955d45d9d98c9c583e09f4a6805cd23dce388). This part of the code needs a bit of love in general, but this can be done in a different PR when we update the scoring formulas

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) 2025 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-types/src/scoring_metrics.rs
oliviasaa: done in [5b9eb80](https://github.com/iotaledger/iota/pull/9914/commits/5b9eb80c2acaea0bad6025604ead1f33fab4b0ff)

crates/iota-types/src/misbehavior_counts.rs
semenov-vladyslav: Why would you need a `verify` function to check against the committee size if it stays constant during the epoch?

crates/iota-types/src/misbehavior_counts.rs
semenov-vladyslav: It's cleaner for an unbounded homogenous collection. The elements in MisbehaviorsV1 are heterogenous, they represent the counts of different things. What would happen if iterator contains more than 4 items? What if the items are reordered (by mistake, of course)? It would be hard to debug such mistakes.

crates/iota-types/src/misbehavior_counts.rs
semenov-vladyslav: Ok, got it. To validate values received from other validators.

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,232 @@
...
+            &self.missing_proposals,
+            &self.equivocations,
+        ]
+        .into_iter()
+    }
+
+    // Returns an iterator over references to major misbehavior fields.
+    // Major misbehaviors carry a higher penalty in the scoring system.
+    pub fn iter_major_misbehaviors(&self) -> impl Iterator<Item = &T> {
+        vec![&self.equivocations].into_iter()
```
semenov-vladyslav:
```suggestion
        std::iter::once(&self.equivocations)
```

crates/iota-types/src/scoring_metrics.rs
semenov-vladyslav: Maybe not strong enough?

semenov-vladyslav (approved):

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,232 @@
...
+// scoring mechanism, with a value for each metric. The metrics (misbeheaviors)
+// include, faulty blocks, equivocation and missing proposal counts for each
+// authority. This first version does not include any type of proof. Any metric
+// contained in this struct must be guaranteed to be monotonically increasing,
+// because of the way updates are applied from reports.
+//
+// The type parameter `T` determines the storage type for the metrics:
+// - `T = Vec<u64>` for reports (one value per authority)
+// - `T = Vec<AtomicU64>` for atomic metrics collected and stored locally (one
+//   atomic per authority)
```
Thoralf-M, line 18: Could create types like this
```Rust
type MisbehaviorsV1Reports = MisbehaviorsV1<Vec<u64>>;
type MisbehaviorsV1Metrics = MisbehaviorsV1<Vec<AtomicU64>>;
```
and then also have the impl blocks `impl MisbehaviorsV1Reports{...}` to make this clearer

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: changed in [edfb2d4](https://github.com/iotaledger/iota/pull/9914/commits/edfb2d41b84d943ac6ed71c20127708aeff363a3)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: changed in [edfb2d4](https://github.com/iotaledger/iota/pull/9914/commits/edfb2d41b84d943ac6ed71c20127708aeff363a3)

consensus/core/src/dag_state.rs
```diff
@@ -248,7 +248,10 @@ impl DagState {
 
         // Initialize scoring metrics according to the metrics in store and the blocks
         // that were loaded to cache.
-        let recovered_scoring_metrics = state.store.scan_scoring_metrics().expect("Database error");
+        let recovered_scoring_metrics = state
+            .store
+            .scan_scoring_metrics(&state.context.committee)
+            .expect("Database error");
```
piotrm50: do we expect database error?

```suggestion
            .expect("database read success");
```

crates/iota-core/src/authority/scorer.rs
```diff
@@ -193,41 +193,43 @@ impl Scorer {
...
+// VersionedMisbehaviorReport
 fn calculate_median_report(
     reports_and_voting_power: &[(VersionedMisbehaviorReport, VotingPower)],
 ) -> MisbehaviorsV1<MedianMetricVec> {
     // Calls to this method should ensure that we have at least one report to
     // process.
     assert!(!reports_and_voting_power.is_empty());
 
-    let number_of_metrics = reports_and_voting_power[0].0.iterate_over_metrics().len();
+    // TODO: create a new function to calculate this, or rework a bit on this part
```
piotrm50, line 228: Leftover todo? Should this be resolved in this PR or in a next one? Is this tracked in an issue?

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,235 @@
...
+        .into_iter()
+    }
+}
+
+impl<T> FromIterator<T> for MisbehaviorsV1<T> {
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut iterator = iter.into_iter();
+        Self {
+            faulty_blocks_provable: iterator.next().expect(
+                "Not enough
```
piotrm50: weird newline in expect. Also - do we "expect" not enough elements in iterator? Same for the expect messages below.

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,235 @@
...
+elements in iterator",
+            ),
+            faulty_blocks_unprovable: iterator.next().expect("Not enough elements in iterator"),
+            missing_proposals: iterator.next().expect("Not enough elements in iterator"),
+            equivocations: iterator.next().expect(
+                "Not enough elements in
+iterator",
```
piotrm50: weird newline again

crates/iota-types/src/misbehavior_counts.rs
piotrm50: agree that it's a stretch of a usecase for FromIterator

crates/iota-types/src/misbehavior_counts.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    // Returns a reference to the equivocations field.
+    pub fn equivocations(&self) -> &T {
+        &self.equivocations
+    }
+
+    // Returns an iterator over references to all misbehavior fields.
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        vec![
```
piotrm50: I think we can avoid creating a new vector on each call if we use an array or chain of std::iter::once like below. We can apply similar change in `iter_minor_misbehaviors`

```suggestion
        [
```

Thoralf-M (approved):

crates/iota-core/src/authority/scorer.rs
oliviasaa: This will be in a different PR, and it's tracked in the Core Protocol project Board.

consensus/core/src/dag_state.rs
oliviasaa: fixed in [389644d](https://github.com/iotaledger/iota/pull/9914/commits/389644de2e5c037181e3a074357015e25decb829)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: fixed in [389644d](https://github.com/iotaledger/iota/pull/9914/commits/389644de2e5c037181e3a074357015e25decb829)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: fixed in [389644d](https://github.com/iotaledger/iota/pull/9914/commits/389644de2e5c037181e3a074357015e25decb829)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: fixed in [872a04f](https://github.com/iotaledger/iota/pull/9914/commits/872a04fcf80372c9e751eda6ea7f406c737bf5e2)

crates/iota-types/src/misbehavior_counts.rs
oliviasaa: I removed the FromIterator in [f647b0d](https://github.com/iotaledger/iota/pull/9914/commits/f647b0d43822150ed9730b8a0ba3937485cb1faf).
There is still a fn that does the same thing, but at least now it is a private function and not this impl.

If it's ok for you guys, I'd stop touching `MisbehaviorsV1` in this PR, and work on any other requested change when we introduce `MisbehaviorsV2`, which won't take long. All inputs are welcome for that. 

The most important part in this PR is to make sure the non-internal types (reports) and storage types are well defined and that, even we change one thing or another to improve the codebase, those types won't have a different serialization (so we don't have to change versions for them unless we actually add new metrics)

piotrm50 (approved):

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -147,11 +164,12 @@ impl Store for RocksDBStore {
                 )
                 .map_err(ConsensusError::RocksDBFailure)?;
         }
-        for (authority, metrics) in write_batch.scoring_metrics {
+        if let Some(metrics) = &write_batch.scoring_metrics {
             batch
-                .insert_batch(&self.scoring_metrics, [(authority, metrics)])
+                .insert_batch(&self.scoring_metrics_v2, [(&(), metrics)])
                 .map_err(ConsensusError::RocksDBFailure)?;
```
semenov-vladyslav:
```suggestion
                .insert_batch(&self.scoring_metrics_v2, [(&(), metrics)])?;
```
Here and in other places, no need for `map_err`, the `?` operator will find the `impl From<TypedStoreError> for ConsensusError`.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -193,41 +193,43 @@ impl Scorer {
...
+//     - Metric M1: [2, 2, 2, 2] (values for authorities A_1, A_2, A_3, A_4)
+//     - Metric M2: [1, 1, 1, 1] (values for authorities A_1, A_2, A_3, A_4)
+//
+// For Metric M1, we have that the median metric vector is [1, 1, 1, 1].
+//
+// This method returns a vector of MedianMetricVec, one per metric in
+// VersionedMisbehaviorReport
 fn calculate_median_report(
     reports_and_voting_power: &[(VersionedMisbehaviorReport, VotingPower)],
 ) -> MisbehaviorsV1<MedianMetricVec> {
```
semenov-vladyslav, line 223: Should the return type be `VersionedMisbehaviorReport`?

crates/iota-core/src/authority/scorer.rs
```diff
@@ -238,17 +240,25 @@ fn calculate_median_report(
...
-    // Calculate and return the weighted median for each metric
-    let median_report = reports_and_voting_power_per_metric
+    // Calculate the weighted median for each metric and construct the result
+    let mut medians = reports_and_voting_power_per_metric
         .iter_mut()
-        .map(|vec| calculate_weighted_median(vec.as_mut_slice()))
-        .collect::<MisbehaviorsV1<MedianMetricVec>>();
-    median_report
+        .map(|vec| calculate_weighted_median(vec.as_mut_slice()));
+    MisbehaviorsV1::new(
```
semenov-vladyslav:
```suggestion
    MisbehaviorsV1::collect_from(
```

semenov-vladyslav (approved):

crates/iota-core/src/authority/scorer.rs
oliviasaa: We don't need the atomics (like in `VersionedMisbehaviorReport`).
By the way, I added to one of my tasks in the project board to improve the code in this file. If that's ok for you, I'll reach out to you _while I'm doing it_, instead of only after the review, to make sure this file is more readable and less of a mess

crates/iota-core/src/authority/scorer.rs
oliviasaa: done in [d948437](https://github.com/iotaledger/iota/pull/9914/commits/d948437a670aae462c5d1c5c6542d8f74813c224)

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: done in [d948437](https://github.com/iotaledger/iota/pull/9914/commits/d948437a670aae462c5d1c5c6542d8f74813c224)

oliviasaa:
Given recent changes in the implementation plans, this PR won't be merged and another one will be open with the new modifications

PR #9915, opened 2026-01-26
chore(docs): Update Layerzero EVM bridging docs 

29988122:
Update ERC20 bridging tutorials with updated Move example instead of EVM example.

### DIFF
--- a/docs/content/developer/iota-evm/how-tos/send-ERC20-across-chains.md
+++ b/docs/content/developer/iota-evm/how-tos/send-ERC20-across-chains.md
@@ -9,15 +9,18 @@ teams:
   - iotaledger/l2-smart-contract
 ---
 
-# Send ERC20 Tokens Across Chains
+# Send ERC20 Tokens Between IOTA EVM and IOTA L1
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


Dr-Electron:
Looks good to me. Don't forget to change the PR title to follow the guidelines. Otherwise we can't merge.
Also you might want to follow the guidelines for commits.

Back on topic: Do you think we should split this to have one guide which we add to the EVM section and another to the Move section? Or should we show the same page on both? 🤔 

I think I prefer the split option to have a clear focus on Move. But not sure. This also already works well and could be used as a guide to migrate stuff from EVM as I mentioned 🤔

29988122:
> hink I prefer the split option to have a clear focus on Move. But not sure. This also already works well and could be used as a guide to migrate stuff from EVM as I mentioned

Split option 100%. That's the info I got && what I planned in the first place.
Let me modify the PR, change the commit style/PR title, and include the IOTA Move <> SUI Move bridging example page in the corresponding section of the doc site.

vivekjain23 (commented):
LGTM,
But IMO split option will be much better: EVM section and another to the Move section and also for better context

Dr-Electron (approved):
Let's merge this for now until we split it

vivekjain23 (approved):

PR #9924, opened 2026-01-26
fix(sdk, dashboard): update chain in wallet provider

### LINKED ISSUE #9923: WalletProvider chain doesn't update when network changes
## Bug description

When switching networks in the wallet dashboard, the chain value in WalletProvider doesn't update, causing `useCurrentChain()` to return stale values.

## Steps To reproduce the bug

1. Open wallet dashboard
2. Switch to a different network
3. Try to sign a transaction
4. The transaction uses the old chain value instead of the new network's chain

## Expected behaviour

Network change should update `persistedNetworkId` in localStorage
- WalletProvider's chain should update reactively
- `useCurrentChain()` should return the current network's chain

brancoder:
# Description of change

resolves #9923


sdk/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -76,6 +76,11 @@ export function WalletProvider({
         }),
     );
 
+    // Update chain in the store when it changes
+    useEffect(() => {
+        storeRef.current.getState().setChain(chain);
+    }, [chain]);
```
marc2332: No need for the useEffect, you can replace it with an if:

```suggestion
    // Update chain in the store when it changes
    if(storeRef.current.getState().chain != chain) {
        storeRef.current.getState().setChain(chain);
    }
```

marc2332 (changes requested):

evavirseda (approved):
works!

marc2332 (approved):

PR #9926, opened 2026-01-26
feat(indexer): Reorganize "watermarks" table columns for clarity

### LINKED ISSUE #9902: [indexer]: Rethink column names in the watermarks table
Currently we are using the non-intuitive _hi _lo columns that sometimes are inclusive, sometimes exclusive.

We should switch to something that describes the meaning of columns more clearly.

tomxey:
# Description of change

Reorganize "watermarks" table columns for clarity

Column Renames:
- `checkpoint_hi_inclusive` → `max_committed_cp`
- `tx_hi` → `max_committed_tx` (exclusive → inclusive, with adjustment)
- `epoch_hi_inclusive` → `current_epoch`
- `epoch_lo` → `min_available_epoch`
- `pruner_hi` → `lowest_unpruned_key`
- `timestamp_ms` → `min_bounds_updated_at_timestamp_ms`
- `reader_lo` → split into `min_available_cp` and `min_available_tx`

fixes https://github.com/iotaledger/iota/issues/9902

### DIFF
--- a/crates/iota-indexer/migrations/pg/2026-01-26-095900_rename_watermarks_columns/up.sql
+++ b/crates/iota-indexer/migrations/pg/2026-01-26-095900_rename_watermarks_columns/up.sql
@@ -1,0 +1,60 @@
+-- Add min_available_tx and min_available_cp columns to watermarks table
+ALTER TABLE watermarks
+ADD COLUMN min_available_tx BIGINT,
+ADD COLUMN min_available_cp BIGINT;
+
+-- Populate initial values for min_available_tx and min_available_cp based on epoch_lo
+-- For each watermark entry, find the corresponding epoch and use its first_checkpoint_id
+-- and first_tx_sequence_number
+UPDATE watermarks w
+SET
+    min_available_cp = COALESCE(e.first_checkpoint_id, 0),
+    min_available_tx = COALESCE(e.first_tx_sequence_number, 0)
+FROM epochs e
+WHERE e.epoch = w.epoch_lo;
+
+-- For watermarks where epoch_lo doesn't exist in epochs table (shouldn't happen in normal operation),
+-- set to 0 as a safe default
+UPDATE watermarks
+SET
+    min_available_cp = COALESCE(min_available_cp, 0),
+    min_available_tx = COALESCE(min_available_tx, 0)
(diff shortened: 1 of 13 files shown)


crates/iota-indexer/migrations/pg/2026-01-26-095900_rename_watermarks_columns/up.sql
```diff
@@ -0,0 +1,60 @@
...
+ALTER TABLE watermarks
+RENAME COLUMN epoch_lo TO min_available_epoch;
+
+-- Rename pruner_hi to lowest_unpruned_key for clarity
+ALTER TABLE watermarks
+RENAME COLUMN pruner_hi TO lowest_unpruned_key;
+
+-- Rename timestamp_ms to min_bounds_updated_at_timestamp_ms for clarity
+ALTER TABLE watermarks
+RENAME COLUMN timestamp_ms TO min_bounds_updated_at_timestamp_ms;
```
kodemartin, line 60: I think you could be more concrete, and by implication a little bit less verbose

```suggestion
RENAME COLUMN timestamp_ms TO min_available_epoch_timestamp_ms;
```
It is when updating `min_available_epoch` that we update the timestamp. The other lower bounds are derived from the epoch eitherway.

kodemartin (approved):
Looks good with a rename suggestion.

crates/iota-indexer/migrations/pg/2026-01-26-095900_rename_watermarks_columns/up.sql
tomxey: I was considering naming of this sort, but I feel that it kindof suggests that it's a timestamp of the epoch, not the timestamp at which the epoch was updated.
Also, from the perspective of the users of the table, all min_* bounds are updated at the same time and this timestamp works for all of them. Epoch is used by the updating algorithm to determine if update is needed, but it's implementation detail.

With that, I'm inclined to keep the `min_bounds_updated_at_timestamp_ms` name.
What do you think?

kodemartin: It's fine, ok!

sergiupopescu199 (approved):
lgtm ✨

PR #9930, opened 2026-01-26
refactor(starfish): unify source for transactions and headers

### LINKED ISSUE #9931: doc: Unify sources for transaction and header in the consensus layer
Currently, we use several enums for transactions and headers. Sometimes the source is the same and we use different enums. For the ease of maintenance, we can modify those to use one enum.

polinikita:
# Description of change

This PR consolidates BlockHeaderSource and TransactionSource enums into a single DataSource enum for consistent tracking of data origins across the consensus Starfish layer.  

We also unify the function update_transaction_metadata to behave similarly to the one for headers.

Fixes #9931 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -33,7 +33,7 @@ use crate::{
     context::Context,
     cordial_knowledge::CordialKnowledgeHandle,
     core_thread::CoreThreadDispatcher,
-    dag_state::{BlockHeaderSource, DagState},
+    dag_state::{DagState, DataSource},
     encoder::ShardEncoder,
     error::{ConsensusError, ConsensusResult},
     header_synchronizer::HeaderSynchronizerHandle,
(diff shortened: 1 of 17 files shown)


VorobyevIlya (approved):

crates/starfish/core/src/authority_service.rs
```diff
@@ -576,7 +576,7 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         // 10. Add the block to dag, add its missing ancestors to the set
         let (missing_block_ancestors, missing_block_committed_transactions) = self
             .core_dispatcher
-            .add_blocks(vec![verified_block])
+            .add_blocks(vec![verified_block], DataSource::BlockStreaming)
```
filipdulic, line 579: different from the source state in the `add_block_headers` call in line 565 above, is this intentional?

filipdulic (approved):

crates/starfish/core/src/authority_service.rs
polinikita: Yes, this serves for both transactions and headers (same source). This is a path needed for any consensus protocol  For additional headers, we use cordial dissemination and want to evaluate how useful it is for us

PR #9936, opened 2026-01-26
fix(docs): remove breaking brackets from docs

Dr-Electron:
# Description of change

This is currently breaking the docs in some builds. Not sure if this is any special syntax but I think it should be safe to remove it

### DIFF
--- a/crates/iota-types/src/iota_system_state/iota_system_state_summary.rs
+++ b/crates/iota-types/src/iota_system_state/iota_system_state_summary.rs
@@ -33,7 +33,7 @@ pub enum IotaSystemStateSummary {
 }
 
 /// This is the JSON-RPC type for the
-/// [`IotaSystemStateV1`](super::iota_system_state_inner_v1::IotaSystemStateV1)
+/// `IotaSystemStateV1`(super::iota_system_state_inner_v1::IotaSystemStateV1)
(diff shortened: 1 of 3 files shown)


bingyanglin (approved):

crates/iota-types/src/iota_system_state/iota_system_state_summary.rs
```diff
@@ -33,7 +33,7 @@ pub enum IotaSystemStateSummary {
 }
 
 /// This is the JSON-RPC type for the
-/// [`IotaSystemStateV1`](super::iota_system_state_inner_v1::IotaSystemStateV1)
+/// `IotaSystemStateV1`(super::iota_system_state_inner_v1::IotaSystemStateV1)
```
Thoralf-M, line 36: Without the [] this is not a clickable link anymore

marc2332: exactly, its like markdown. [text](link).

Dr-Electron: Yeah but super: is not a valid url. Our library tries to find a port in that url ;)
But I found out that this is a known issue with node 25 which is not officially supported by docusaurus anyway. So I wait for them to fix it and use node 24 in the meantime

PR #9939, opened 2026-01-27
feat(wallet): attach user global properties to every amplitude event

### LINKED ISSUE #9240: Attach user global properties to every amplitude event
Attach these properties to every amplitude event: 

- `walletAppMode` : fullscreen / extension / side panel
- `network`: mainnet, testnet, devnet, custom
- `appVersion`: eg 1.12.0

VmMad:
# Description of change

I realized that in amplitude, when a user is grouped with the `(none)` network user property is because the network property is not being sent. This PR tries to fix that from the start.

Closes #9240 


evavirseda (dismissed):
looks good! i haven't been able to reproduce the issue locally, but i understand that explicitly setting the network property right after initialization closes the gap where the value was missing, so the logic looks sound

apps/wallet/src/shared/analytics/amplitude.ts
```diff
@@ -33,6 +36,8 @@ export async function initAmplitude() {
         },
     });
 
+    setNetworkGroup(network, customRpc);
```
panteleymonchuk: currently we are not sure we will use groups. Probably we'll change a way how we provide info about "network" context

VmMad: this funcion also sets a `user property` in the amplitude dashboard, that's why I'm using it. it allows to filter some data even though it's called `group`

apps/wallet/src/shared/analytics/amplitude.ts
```diff
@@ -67,14 +72,39 @@ export function getNetworkName(network: Network, customRpc?: string | null): str
...
 }
 
+type AmplitudeIdentityOptions = {
+    network?: Network;
+    customRpc?: string | null;
+    appType?: AppType;
+};
+
 /**
  * Update the user's network group in Amplitude.
```
begonaalvarezd: lets pls update these comments 🙏🏼 pls review the whole PR for outdated comments

begonaalvarezd (approved):

apps/wallet/src/shared/analytics/amplitude.ts
```diff
@@ -81,14 +86,39 @@ export function getNetworkName(network: Network, customRpc?: string | null): str
...
+
 /**
- * Update the user's network group in Amplitude.
- * This allows filtering events by network in Amplitude analytics.
+ * Set the Amplitude user identity with current app context.
+ * Updates user properties: network, walletAppMode, and walletVersion.
+ * This allows filtering and segmenting analytics events by these dimensions.
  */
-export function setNetworkGroup(network: Network, customRpc?: string | null): void {
+export function setAmplitudeIdentity(options?: AmplitudeIdentityOptions): void {
```
brancoder: can we remove the `options` prop since we don't use it in wallet

apps/wallet/src/ui/app/redux/store/amplitudeMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { RootState } from '_src/ui/app/redux/rootReducer';
+import type { Middleware } from '@reduxjs/toolkit';
+
+/**
+ * Redux middleware that keeps the Amplitude user identity in sync with Redux state.
+ * Automatically updates identity when network, customRpc, or appType changes,
+ * ensuring all subsequent events include the latest user context.
+ */
+export const amplitudeMiddleware: Middleware<{}, RootState> = (storeAPI) => (next) => (action) => {
+    const { network, customRpc, appType } = storeAPI.getState().app;
```
brancoder: the appType is undefined, should this be `extensionViewType`

apps/wallet/src/ui/app/redux/store/amplitudeMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const result = next(action);
+
+    const newApp = storeAPI.getState().app;
+    if (
+        newApp.network !== network ||
+        newApp.customRpc !== customRpc ||
+        newApp.appType !== appType
+    ) {
+        setAmplitudeIdentity();
```
brancoder: to fix circular dependency maybe we can remove store from `amplitude.ts` ad pass the options props
```
 setAmplitudeIdentity({
            network: newApp.network,
            customRpc: newApp.customRpc,
            extensionViewType: newApp.extensionViewType,
        });
```

apps/wallet/src/shared/analytics/amplitude.ts
```diff
@@ -39,6 +42,8 @@ export async function initAmplitude() {
         },
     });
 
+    setAmplitudeIdentity();
```
brancoder: to fix circular dep can we remove `setAmplitudeIdentity` from initAmplitude and maybe call it after initAmplitude ex. 

Image

brancoder (changes requested):
can we check the importing of store in the aplitude.ts as it introduces circular dependency.

`store/index.ts → amplitudeMiddleware.ts → amplitude.ts → store/index.ts`

apps/wallet/src/ui/app/redux/store/amplitudeMiddleware.ts
brancoder: the `appType` is still mentioned in this comment

apps/wallet/src/shared/analytics/amplitude.ts
```diff
@@ -86,14 +88,31 @@ export function getNetworkName(network: Network, customRpc?: string | null): str
...
+export function setAmplitudeIdentity(options: AmplitudeIdentityOptions): void {
     if (!ampli.isLoaded) {
         return;
     }
-    const networkName = getNetworkName(network, customRpc);
-    ampli.client.setGroup('network', networkName);
+
+    const networkName = getNetworkName(options.network, options.customRpc);
+
+    const walletAppMode = options.extensionViewType ?? getAppViewType();
```
brancoder: now that the options are mandatory we could remove this fallback to `?? getAppViewType()`, right?

brancoder (changes requested):
LG just minor  changes

brancoder (approved):
LGTM 🥇

PR #9940, opened 2026-01-27
chore: Use Display impls in error implementations instead of Debug

DaughterOfMars:
## Description

Uses the error Display impls instead of Debug in error wrappers as it provides more context.


crates/iota-types/src/error.rs
```diff
@@ -357,10 +357,10 @@ pub enum IotaObjectResponseError {
     Eq, PartialEq, Clone, Debug, Serialize, Deserialize, Error, Hash, AsRefStr, IntoStaticStr,
 )]
 pub enum IotaError {
-    #[error("Error checking transaction input objects: {:?}", error)]
+    #[error("Error checking transaction input objects: {}", error)]
```
thibault-martinez:
```suggestion
    #[error("Error checking transaction input objects: {error}")]
```

crates/iota-types/src/error.rs
```diff
@@ -357,10 +357,10 @@ pub enum IotaObjectResponseError {
...
     UserInput { error: UserInputError },
 
-    #[error("Error checking transaction object: {:?}", error)]
+    #[error("Error checking transaction object: {}", error)]
```
thibault-martinez:
```suggestion
    #[error("Error checking transaction object: {error}")]
```

crates/iota-adapter-transactional-tests/tests/children/child_of_shared_object.snap
```diff
@@ -62,4 +62,4 @@ Contents: t2::o2::Obj2 {
 
 task 8, line 84:
 //# run t1::o1::use_o2_o3 --args object(5,1) object(4,0) --sender A
-Error: Error checking transaction input objects: InvalidChildObjectArgument { child_id: object(4,0), parent_id: object(5,0) }
+Error: Error checking transaction input objects: Object object(4,0) is owned by object object(5,0). Objects owned by other objects cannot be used as input arguments
```
thibault-martinez, line 65: Maybe we should do something about `Error: Error`

DaughterOfMars: Yeah it's not the only place where this is done. I'm happy to change that too

thibault-martinez: We probably should but let's do another PR, let's get this merged so we rebase the feature branch.

thibault-martinez (approved):

valeriyr (approved):

muXxer (approved):

miker83z (approved):

PR #9954, opened 2026-01-27
feat(aa): the field object for AuthenticatorFunctionRef is considered as loaded child

### LINKED ISSUE #9279: [AA] `AuthenticatorInfoV1` should be considered as loaded child
Check: 
- the feasibility of the change
- if it produces noticeable effects
- provide a solution idea (that could be a code change or not to change)

miker83z:
# Description of change

This PR makes it so that the `TemporaryStore` is aware of the fact that the field object containing the `AuthenticatorFunctionRef` was loaded from the store as a child. An entry is created in the `TemporaryStore` field `loaded_runtime_objects`. This makes so that:
- `InnerTemporaryStore::loaded_runtime_objects` is aware of the `AuthenticatorFunctionRef` field object; however, [transaction outputs](https://github.com/iotaledger/iota/blob/5869933bce1786eebd66c2da9f67c6256063573e/crates/iota-core/src/transaction_outputs.rs#L45) do not make use of this field.
- `TemporaryStore::check_execution_results_consistency()` is aware of the `AuthenticatorFunctionRef` field object; however, the consistency here is for mutated objects (which is not the case of `AuthenticatorFunctionRefV1`).
- `TemporaryStore::get_object_modified_at` is aware of the `AuthenticatorFunctionRef` field object; however, this is used for mutated objects (which is, again, not the case).

Additionally: 
- `loaded_runtime_objects` is normally filled at the [end of the execution of a PTB command](https://github.com/iotaledger/iota/blob/5869933bce1786eebd66c2da9f67c6256063573e/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs#L123), using the [ChildObjectStore's `cached_objects` map](https://github.com/iotaledger/iota/blob/5869933bce1786eebd66c2da9f67c6256063573e/iota-execution/latest/iota-move-natives/src/object_runtime/mod.rs#L487). 
- Object are cached there when [invoking `get_or_fetch_object_from_store`](https://github.com/iotaledger/iota/blob/5869933bce1786eebd66c2da9f67c6256063573e/iota-execution/latest/iota-move-natives/src/object_runtime/object_store.rs#L251). 
- In this PR, this cache is not used. It could be pre-filled when initializing the execution context (e.g., [here](https://github.com/iotaledger/iota/blob/5869933bce1786eebd66c2da9f67c6256063573e/iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs#L102)); but, it seems this would only increase the performances (caching) and not have further effects.

This means that, overall, the changes of this PR should not produce noticeable effects during or post execution; but, the change seems to be more conceptually aligned with the dynamic field operation of reading.

This change would enable having the `AuthenticatorFunctionRef` field object considered for the feature brought by the following upstream commit https://github.com/MystenLabs/sui/commit/89b03a818d9bf19707e06e9df8b4992fcf36cec0

A versioning for the rust `AuthenticatorFunctionRef` type was also introduced.

#9279 

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -56,7 +56,11 @@ use iota_types::committee::CommitteeTrait;
 use iota_types::{
     IOTA_SYSTEM_ADDRESS, TypeTag,
     account_abstraction::{
-        account::AuthenticatorFunctionRefV1Key, authenticator_function::AuthenticatorFunctionRefV1,
+        account::AuthenticatorFunctionRefV1Key,
+        authenticator_function::{
+            AuthenticatorFunctionRef, AuthenticatorFunctionRefForExecution,
+            AuthenticatorFunctionRefV1,
+        },
     },
     authenticator_state::get_authenticator_state,
     base_types::*,
@@ -75,6 +79,7 @@ use iota_types::{
     error::{ExecutionError, IotaError, IotaResult, UserInputError},
     event::{Event, EventID, SystemEpochInfoEvent},
     executable_transaction::VerifiedExecutableTransaction,
+    execution::DynamicallyLoadedObjectMetadata,
     execution_config_utils::to_binary_config,
     execution_status::ExecutionStatus,
     fp_ensure,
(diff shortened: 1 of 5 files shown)


bingyanglin (approved):
Approved `iota-core` and `iota-types`.

valeriyr (approved):

PR #9955, opened 2026-01-27
fix(consensus): fix for amnesia test

VorobyevIlya:
# Description of change

PR fixes error in amnesia_recovery_test

### DIFF
--- a/crates/starfish/core/src/authority_node.rs
+++ b/crates/starfish/core/src/authority_node.rs
@@ -862,7 +862,6 @@ mod tests {
         // authority
         let received_from_authority_1 = timeout(Duration::from_secs(10), async {
             'outer: while let Some(result) = output_receivers[index_1].recv().await {
-                info!("{:?}", result);
                 for header in &result.headers {
                     if header.round() > GENESIS_ROUND && header.author() == index_1 {
                         break 'outer;
--- a/crates/starfish/core/src/authority_node.rs
+++ b/crates/starfish/core/src/authority_node.rs
@@ -792,6 +792,7 @@ mod tests {
         //   which will make authority 1 retry during amnesia recovery until it has
         //   finally managed to successfully get back 2f+1 responses, once authority 2
         //   is up and running again.
+        sleep(Duration::from_secs(1)).await;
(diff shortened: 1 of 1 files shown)


nikitapolyanskii (approved):

polinikita (approved):

crates/starfish/core/src/authority_node.rs
```diff
@@ -858,7 +861,8 @@ mod tests {
         // We wait until we see at least one committed block authored from this
         // authority
         let received_from_authority_1 = timeout(Duration::from_secs(10), async {
-            'outer: while let Some(result) = receiver_1.recv().await {
+            'outer: while let Some(result) = output_receivers[index_1].recv().await {
+                info!("{:?}", result);
```
Thoralf-M: Should this stay here?

crates/starfish/core/src/authority_node.rs
```diff
@@ -792,6 +792,7 @@ mod tests {
         //   which will make authority 1 retry during amnesia recovery until it has
         //   finally managed to successfully get back 2f+1 responses, once authority 2
         //   is up and running again.
+        sleep(Duration::from_secs(1)).await;
```
Thoralf-M, line 795: Is this sleep really needed or is `output_receivers[index_1] = receiver_1;` fixing it alone? If something takes more time, it would be better to wait for it in a different way by checking that it actually reaches the wanted state than adding a sleep with a "random" sleep duration

VorobyevIlya: No, it is not really needed; it was needed to reliably replicate the test failure. 
It should work with or without this sleep; only with sleep, we test the case when more commits were made before the stop of the authority. I would prefer to leave it

crates/starfish/core/src/authority_node.rs
VorobyevIlya: Do you mean logging? We can remove it

VorobyevIlya: removed

Thoralf-M (approved):

PR #9971, opened 2026-01-28
feat(gRPC): add checkpoint endpoints

### LINKED ISSUE #9388: gRPC: Add `GetCheckpointData` endpoint
Implement the `GetCheckpointData` endpoint in the `iota-grpc-server` crate.

### LINKED ISSUE #9389: gRPC: Add `StreamCheckpointData` endpoint
Implement the `StreamCheckpointData` endpoint in the `iota-grpc-server` crate.

muXxer:
# Description of change

This PR adds the `GetCheckpointData` and `StreamCheckpointData` endpoints to the gRPC interface.

fixes #9389
fixes #9388


crates/iota-grpc-server/src/transaction_filter.rs
```diff
@@ -0,0 +1,467 @@
...
+            }
+            iota_types::transaction::TransactionKind::EndOfEpochTransaction(_) => {
+                TransactionKind::EndOfEpochTransaction
+            }
+            iota_types::transaction::TransactionKind::RandomnessStateUpdate(_) => {
+                TransactionKind::RandomnessStateUpdate
+            }
+        }
+    }
+}
```
bingyanglin, line 57: Do we also need to filter the `TransactionKind::SystemTransaction`

crates/iota-grpc-server/src/types.rs
```diff
@@ -434,17 +356,265 @@ impl GrpcReader {
...
+                // Send final batch if any remaining
+                if !events_batch.is_empty() {
+                    let events_message = grpc_ledger_service::CheckpointData {
+                        payload: Some(Payload::Events(grpc_event::Events {
+                            events: events_batch,
+                        })),
+                    };
+                    yield Ok(events_message);
+                }
+            }
```
bingyanglin: By this approach we need to collect everything into `all_event` and batch and yield here. Can we move `events_batch`/`events_batch_size` up into the tx loop and do size check and yield? Maybe we allow interleaving event and transaction messages in the stream or it's a bad idea?

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -69,63 +127,226 @@ impl Client {
...
+                        yield response;
+                    }
+
+                    None => {
+                        // Empty payload - skip
+                        continue;
+                    }
+                }
+            }
+        }
```
bingyanglin, line 357: I am thinking an edge case here: if the gRPC stream ends while a checkpoint header was received but no `EndMarker` followed, the partial checkpoint is silently discarded. 
We also have lots of silent data absence in current design for other endpoints without propagating errors though, like for the BCS serialization failures for all of our endpoints we just return `None`, maybe we can define an issue for that? Or we be consistent to allow silent error?

crates/iota-grpc-server/src/ledger_service/get_checkpoint.rs
```diff
@@ -0,0 +1,210 @@
...
+            seq,
+        )) => seq,
+        Some(grpc_ledger_service::get_checkpoint_data_request::CheckpointId::Digest(_digest)) => {
+            // TODO: do we have a lookup table for that?
+            return Err(Status::unimplemented(
+                "checkpoint lookup by digest is not yet implemented",
+            )
+            .into());
+        }
+        Some(grpc_ledger_service::get_checkpoint_data_request::CheckpointId::Latest(_)) => service
```
bingyanglin, line 144: Maybe we change the proto field of `Latest` to a unit-like type? The boolean value is either ignored here or rejected at `fn get_checkpoint_data() in `crates/iota-grpc-server/src/ledger_service/get_checkpoint_data.rs`, so it looks not very consistent.

crates/simulacrum/src/state_reader.rs
bingyanglin: Do we assume the system sate will not be changed in our simulacrum use cases? Because in this way for historical epoch we still use the current system state.

crates/iota-grpc-server/src/ledger_service/get_checkpoint.rs
bingyanglin: Also, the `crates/iota-grpc-server/src/ledger_service/get_checkpoint_data.rs` is not in our module tree, it should be removed at all?

crates/simulacrum/src/state_reader.rs
bingyanglin: I am thinking do we need extra check here to assert when the input epoch is 0?

crates/iota-grpc-server/src/types.rs
```diff
@@ -434,17 +356,265 @@ impl GrpcReader {
...
+                            // Build transaction only if transactions_mask is requested
+                            if transactions_mask.is_some() {
+                                // Apply transaction filter if present
+                                if let Some(ref tx_filter) = transaction_filter {
+                                    if !tx_filter.matches_transaction(state_reader.clone(), &checkpoint_transaction) {
+                                        continue; // Skip non-matching transactions
+                                    }
+                                }
+
+                                let executed_tx = grpc_transaction::ExecutedTransaction::merge_from(
```
bingyanglin: Might be good to have comment here (or in the `Merge` method) that the `signatures`, `checkpoint`, and `timestamp` will be always `None` here?

bingyanglin:
Overall looks good to me, just the above remarks. Also might be good to address the following so easier to merge the `grpc-implementation` branch quicker in the future
- In [simulacrum/src/lib.rs#L717](https://github.com/iotaledger/iota/blob/feat/grpc-implementation/crates/simulacrum/src/lib.rs#L717), the `self.` should be `store.` otherwise the `self.` will go through the `self.with_store` and acquire the same RwLock again (deadlock).
- Same for [simulacrum/src/lib.rs#L684](https://github.com/iotaledger/iota/blob/feat/grpc-implementation/crates/simulacrum/src/lib.rs#L684) we can pass `store` instead.

crates/iota-types/src/storage/read_store.rs
```diff
@@ -333,54 +332,130 @@ pub trait ReadStore: ObjectStore {
...
+                .try_multi_get_transactions_with_events_and_effects(transaction_digests)?;
+
+            for tx_with_events_and_effects in txs_with_events_and_effects {
+                yield self.get_checkpoint_transaction(tx_with_events_and_effects);
+            }
+        })
+    }
+
+    // Fetch all checkpoint data
+    // TODO fix return type to not be anyhow
```
Thoralf-M, line 436: TODO in this PR?

crates/iota-grpc-server/src/types.rs
muXxer: That was actually wrong, I added the missing fields by adding a context.

crates/iota-types/src/storage/read_store.rs
muXxer: This comment was there before, it is in the actual codebase. I just split these functions sir :D

crates/iota-grpc-server/src/transaction_filter.rs
muXxer: Not here, because there is no such thing as a SystemTransaction in the transaction kinds. But in the filter we will need to match against several types if that filter kind is passed. I will fix it.

crates/simulacrum/src/state_reader.rs
muXxer: I added support for historic state.

piotrm50 (approved):
approving `iota-types` and `iota-core` changes

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
muXxer: > We also have lots of silent data absence in current design for other endpoints without propagating errors though, like for the BCS serialization failures for all of our endpoints we just return None, maybe we can define an issue for that? Or we be consistent to allow silent error?
I aded trowing and error if the stream ends unexpectedly.

Can you open an issue for this part of your message an specify what you mean?

bingyanglin: Yes, I created https://github.com/iotaledger/iota/issues/9996 and assigned it to me.

bingyanglin (approved):

PR #9974, opened 2026-01-28
feat(starfish): add metrics for skipped block headers

### LINKED ISSUE #9972: starfish: investigate discrepancy between postfiltered headers in cordial dissemination and accepted in dag_state
We observed in simulation on distributed network for 70 validators that there is a discrepancy in accepted block headers in the dag state (many additional headers are going through the filter but are not accepted). We can add more metric to identify the reason why that happens.

polinikita:
# Description of change

Add new metrics to track block header filtering in Starfish consensus:
- `additional_headers_round_gap`: Histogram measuring round difference between additional headers in bundles and the main block
- `core_skipped_headers`: Unified counter with `authority` and `source` labels tracking headers filtered during core execution (replaces separate `block_manager_filtered_processed_headers_by_authority` and `dag_state_skipped_block_headers`)

Increase the filter size from 10_000 to 100_000 in authority service to avoid inconsitencies in processed/filtered headers in core and authority service. This does not change the performance in local tests.

Also includes Grafana dashboard improvements:
- Add panels for the new metrics
- Add descriptions to all 61 panels that were missing them
- Align panel widths to standard (w=12 for 2 panels, w=8 for 3)

Fixes #9972 

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -49,7 +49,7 @@ use crate::{
 
 pub(crate) const COMMIT_LAG_MULTIPLIER: u32 = 5;
 
-const MAX_FILTER_SIZE: u32 = 10000;
+const MAX_FILTER_SIZE: u32 = 100000;
 
 struct FilterForHeaders {
     header_digests: DashSet<BlockHeaderDigest>,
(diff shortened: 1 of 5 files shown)


bingyanglin (approved):
Approved `startish-overview.json`

VorobyevIlya (approved):

PR #9975, opened 2026-01-29
feat(indexer): Do dependency checking in optimistic indexing using input objects

### LINKED ISSUE #9891: [indexer]: Refactor dependency check during optimistic execution to check existence of input objects
## Description

We currently rely on `tx_global_order` and the `transactions` table to check dependencies during optimistic execution, but this is rather an overkill. What we want to ensure with that check is that all input objects of the transaction are already indexed before indexing its output objects. And while checking the transactions solves this, certainly we shouldn’t need to maintain the entire history to check if near live objects exist in the index. I think therefore that it is plausible that we revisit the approach by relying on an object check instead.

tomxey:
# Description of change

Do dependency checking of transactions based on input objects instead of previous transactions.
To make the dependency checking process resilient to pruning.

fixes https://github.com/iotaledger/iota/issues/9891

### DIFF
--- a/crates/iota-indexer/src/optimistic_indexing.rs
+++ b/crates/iota-indexer/src/optimistic_indexing.rs
@@ -136,7 +136,7 @@ impl OptimisticTransactionExecutor {
             .iter()
             .map(|ob| (ob.id(), ob.version()))
             .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


kodemartin (approved):
lgtm 🌸

sergiupopescu199 (approved):
lgtm ✨

PR #9984, opened 2026-01-29
fix(indexer): enable partition pruning when fetching StoredTransactions

### LINKED ISSUE #9983: [iota-indexer]: `iota_getTransactionBlock` & `iota_multiGetTransactionBlocks` db queries triggers a scan of all partitions.
On `alphanet` we noticed that when requesting one or more transaction blocks it would take several seconds to return the response. On other networks it would take several milliseconds.

We noticed that the db query was not performing partition pruning, thus it triggered the scan of all partitions.

sergiupopescu199:
# Description of change

This patch refactors the database query in order to use partition pruning when requesting one or multiple `StoredTransaction`.

The postgres query planner was not able to execute partition pruning when using `JOIN` or `IN` clauses.
This [response](https://www.postgresql.org/message-id/CAKJS1f9nWfAYDMF8-ysG%3DTQZ5r9vB%3Dpkk7-y7zQKdZExOraJ3w%40mail.gmail.com) does add some more context.

fixes #9983

- Connected the indexer reader to `alphanet` database and executed the following JSON RPC query
```shell
curl --location 'localhost:9000' \
--data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "iota_multiGetTransactionBlocks",
    "params": [
        [
            "AUKbNtEvQzL8JJBTeLAwGV6z76C8829gtjgqVGDiRJP2",
            "Dgs1TReXNrM8fPpxbG56hM69Mmucvmm6b1xoHoc6NMq8",
            "AiTcsZFcxzBMCcKJWKnaHDoPq8Y5bVnHjGMheXFTiwXM"
        ],
        {
            "showInput": true,
            "showRawInput": false,
            "showEffects": true,
            "showEvents": true,
            "showObjectChanges": false,
            "showBalanceChanges": false,
            "showRawEffects": false
        }
    ]
}'
```
it takes around 120ms compared to 10s-13s.

> [!NOTE]
> This patch does not impact the ingestion pipeline, only the JSON RPC part which was tested.

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -2770,14 +2770,18 @@ impl<'a> DBReader<'a> {
     ) -> IndexerResult<Vec<StoredTransaction>> {
         let pool = self.main_reader.get_pool();
         run_query_async!(&pool, |conn| {
+            // using two-step query to allow partition pruning during execution.
+            let tx_sequence_numbers = tx_digests::table
+                .filter(tx_digests::tx_digest.eq_any(&digests))
+                .select(tx_digests::tx_sequence_number)
+                .load::<i64>(conn)?;
+
+            if tx_sequence_numbers.is_empty() {
+                return Ok(vec![]);
+            }
+
             transactions::table
-                .inner_join(
-                    tx_digests::table
-                        .on(transactions::tx_sequence_number.eq(tx_digests::tx_sequence_number)),
-                )
-                // we filter the tx_digests table because it is indexed by digest,
-                // transactions table is not
-                .filter(tx_digests::tx_digest.eq_any(digests))
+                .filter(transactions::tx_sequence_number.eq_any(tx_sequence_numbers))
                 .select(StoredTransaction::as_select())
                 .load::<StoredTransaction>(conn)
         })


tomxey (approved):

crates/iota-indexer/src/read.rs
```diff
@@ -2769,15 +2769,22 @@ impl<'a> DBReader<'a> {
         digests: Vec<Vec<u8>>,
     ) -> IndexerResult<Vec<StoredTransaction>> {
         let pool = self.main_reader.get_pool();
+
         run_query_async!(&pool, |conn| {
+            // we use a two step query to enable partition pruning.
+            // using JOIN or IN clauses do not support partition pruning and triggers a scan
```
kodemartin: This is not entirely true. Both JOIN and IN allow partition pruning when conditions contain literal values. E.g. `IN (1, 2, 3)` or `ON <condition> AND <partition-key> IN (literals)`

The second query below actually uses IN.

It is when using expressions whose values are not known during planing or execution that this does not work. You can also add the link to the Postgres mail thread.

kodemartin (approved):
lgtm with a small suggestion

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/9984/commits/63334041de11c4866f85943a12a18b5dd8e4eb6c

PR #10014, opened 2026-02-02
chore: Replace `Digest` with SDK version

### LINKED ISSUE #9838: Replace `Digest` with SDK version

DaughterOfMars:
## Description

Replace the `Digest` types with the SDK version.


crates/iota-rest-api/src/reader.rs
```diff
@@ -80,7 +80,7 @@ impl StateReader {
 
         use super::transactions::TransactionNotFoundError;
 
-        let transaction_digest = digest.into();
+        let transaction_digest = digest;
```
Thoralf-M: no need for the new var assignment anymore

DaughterOfMars: removed

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -281,13 +281,14 @@ impl AuthorityStorePruner {
...
 
             if let Some(event_digest) = effects.events_digest() {
-                if let Some(next_digest) = event_digest.next_lexicographical() {
-                    perpetual_batch.schedule_delete_range(
-                        &perpetual_db.events,
-                        &(*event_digest, 0),
-                        &(next_digest, 0),
-                    )?;
-                }
+                // TODO: Maybe we need a next_lexicographical which returns an option? How
```
muXxer: If there would be the situation that all bytes are already 255, in the old implementation it would return None, indication that there is no next value. Now it would wrap around and return all zeros. That might be fine for rocksdb, because it might silently delete nothing, but it depends on the database implementation itself and might change with different versions of rocksdb. We can't be sure, so I personally would prefer to keep the old behaviour that it returns None. Also it's cleaner, if anyone ever uses this function somewhere else, to not cause similar issues.

crates/iota-rest-api/openapi/openapi.json
thibault-martinez, line 1: @muXxer Can you have a look at this please?

crates/iota-core/src/authority/authority_store_pruner.rs
thibault-martinez: Let's bring back the Option

DaughterOfMars: done

crates/iota-types/src/unit_tests/base_types_tests.rs
```diff
@@ -486,16 +486,16 @@ fn next_lexicographical_digest() {
     output[31] = 1;
     assert_eq!(
         TransactionDigest::ZERO.next_lexicographical(),
```
thibault-martinez, line 488: May want to bring back the opt version in these tests

DaughterOfMars: I don't really see the point tbh. We can really get rid of these tests since they are testing things we already test in the SDK

thibault-martinez: Fine, we'll do it when the owners ask us to

Thoralf-M (approved):

thibault-martinez (approved):

thibault-martinez:
CI failing

PR #10020, opened 2026-02-02
feat(indexer): Use generic pruner to prune optimistic transactions

### LINKED ISSUE #9887: [indexer]: Convert optimistic pruning to also use the generic `watermarks` table pruning approach
We could get rid of the custom pruning logic for optimistic transactions now that we have flexibility of configuring pruning policies per table.

tomxey:
# Description of change

Optimistic tables will now be pruned by the generic mechanism, the same that is used for all other tables.

1. It required some adjustments, since to prune by precalculated ranges we had to start pruning by optimistic_sequence_number which was not supported by the generic pruner.
A workaround was applied to use optimistic_sequence_number values instead of tx_seq in watermarks table for the "optimistic_transactions" row.
 
2. Lowest available tx (which also is the pruning threshold) is only estimated and may be off by few txs, but it should not matter for pruning purposes. (the pruning will still work correctly, just the pruning boundary may not be precisely at the beginning of the epoch)

fixes https://github.com/iotaledger/iota/issues/9887


kodemartin (approved):
lgtm

sergiupopescu199 (approved):
lgtm 🎄

PR #10027, opened 2026-02-02
feat(gRPC): resolve dependencies of `grpc-types` and `grpc-client` on internal crates

muXxer:
# Description of change

This PR removes the dependencies of `iota-grpc-types` and `iota-grpc-client` on the internal `iota-sdk` and `move-core-types`. This enables us to move these two crates to the external `iota-rust-sdk`.


crates/iota-grpc-types/src/proto/iota/grpc/v0/command.rs
```diff
@@ -5,55 +5,30 @@
...
+                    nested_result_index: None,
+                })
+            }
+            iota_sdk_types::transaction::Argument::NestedResult(idx, nested_idx) => {
                 argument::Kind::Result(argument::Result {
                     index: Some(idx as u32),
                     nested_result_index: Some(nested_idx as u32),
                 })
             }
+            _ => panic!("Unsupported argument type for gRPC conversion"),
```
bingyanglin: Maybe like the following instead of panic?
```rust
_ => return Err(format!("Unsupported argument type for gRPC conversion: {:?}", arg).into()),
```
Otherwise if the SDK introduces a new argument type before this conversion is update we panic here.

Just in this way we might need to use `TryFrom` instead of `From`. I guess fallible conversion is better for gRPC server?

Thoralf-M (approved):

semenov-vladyslav (approved):

bingyanglin (approved):

PR #10072, opened 2026-02-04
fix(ts-sdk): fix graphql transport error

### LINKED ISSUE #9143: Graphql Transport Error: `Invalid typed array length: -822796575`
When doing this query on testnet, we get an error in the graphql transport library:  `RangeError: Invalid typed array length: -822796575` note that this only happens when passing `options.showInput: true`, but without it, it works

```ts
                const {
                    data: currentData,
                    nextCursor,
                    hasNextPage,
                } = await client.queryTransactionBlocks({
                    options: {
                        showInput: true,
                    },
                    filter: {
                        FromAddress: '0xe1bdd4f74c2724029da8e038ba45e267926599583982ca3938ead60281b9da4b',
                        MoveFunction: {
                            package: '0x0000000000000000000000000000000000000000000000000000000000000002',
                            module: 'timelock',
                            function: 'unlock_with_clock',
                        },
                    },
                    cursor,
                    limit: 50,
                    order: 'descending',
                });

```

KeitoTadashi:
# Description of change

Return raw bytes as number[] using Array.from() instead of parsing as string.

fixes #9143 


sdk/graphql-transport/src/mappers/transaction-block.ts
```diff
@@ -222,7 +222,7 @@ function mapTransactionInput(input: typeof bcs.CallArg.$inferType): IotaCallArg
     if (input.Pure) {
         return {
             type: 'pure',
-            value: bcs.string().parse(fromBase64(input.Pure.bytes)),
+            value: Array.from(fromBase64(input.Pure.bytes)),
```
marc2332: Did you try this?
```suggestion
            value: bcs.vector(bcs.u8()).parse(fromBase64(input.Pure.bytes)),
```
Seems to match more the how https://github.com/iotaledger/iota/blob/a82d92ce410fdd74e0f1121681fdd4e607040488/sdk/typescript/src/bcs/bcs.ts#L88 does it

marc2332 (changes requested):
Changeset missing too

sdk/graphql-transport/src/mappers/transaction-block.ts
KeitoTadashi: good point

brancoder: while testing `bcs.vector(bcs.u8()).parse(fromBase64(input.Pure.bytes)),` i got the same error as reported in the task.
using this approach fixed the issue : `Array.from(fromBase64(input.Pure.bytes)),`

brancoder (changes requested):

brancoder (approved):
LGTM

sdk/graphql-transport/src/mappers/transaction-block.ts
marc2332: for the sake of being more accurate, lgtm in any case
```suggestion
            value: Uint8Array.from(fromBase64(input.Pure.bytes)),
```

marc2332 (approved):

PR #10087, opened 2026-02-05
feat(json-rpc): resolve clever errors

### LINKED ISSUE #6256: Improve move errors in JSON RPC
## Description

If a PTB failed due to a move error that was defined as vector<u8> like this
https://github.com/iotaledger/iota/blob/b79cea7a6cc701d0dad33978d5a3bf80af97284e/examples/move/errors/sources/errors.move#L8-L9
the RPC currently returns the error like this:
```
MoveAbort(MoveLocation { module: ModuleId { address: 974d61b4e495aa6b464759cbb9e1acea321246151d5db9afd3aacd1d8cda03dc, name: Identifier("errors") }, function: 3, instruction: 1, function_name: Some("vector_error") }, 9223372165703925763) in command 0
```
where `9223372165703925763` is the tagged abort code which maps to vector<u8> error message (`This is an error message.`.
But for users this number is not really helpful, it would be better to get the error message that was defined in the code as it happens in the CLI.
In the CLI the error is resolved and replaced here https://github.com/iotaledger/iota/blob/1ad39f9527e03375c7d644eee57108d4c9e93956/crates/iota/src/clever_error_rendering.rs#L39

Example tx in the testnet: `7iSLnPw8krZbeRkXnoVkyM2Edh2T7RBushPT1FPSrMmr`
```
iota client tx-block 7iSLnPw8krZbeRkXnoVkyM2Edh2T7RBushPT1FPSrMmr
Status: Failure { error: "1st command aborted within function '0x974d61b4e495aa6b464759cbb9e1acea321246151d5db9afd3aacd1d8cda03dc::errors::vector_error' at line 30. Aborted with 'EVectorError' -- 'This is an error message.'" }
```

More information about the tagged abort code:
```
## Compilation scheme

Non-u64 constants used in an abort context, and that have been annotated
with `#[error(code = <n>)]` will turn into a tagged abort code (a `u64`)
with the following layout
``
    // |<tagbit>|<reserved>|<line number>|<identifier index>|<constant index>|
    //   1-bit    15-bits      16-bits        16-bits          16-bits
`` 

i.e., we tag these special abort codes, and then bit-pack source line
number, then the index in the constant pool that holds the string name
of the constant that was used, and then the index in the constant pool
of the actual constant value. Note, that in some cases the index of the
constant name into the constant pool and the index into the constant
pool for the constants value may be the same, e.g., if you have
something like
``move
#[error]
const MyError: vector<u8> = b"MyError";
``

In the case of "naked assertions" with no abort code, we simply set the
tag bit, and line number parts of the bitfield, and set the error code,
identifier, and constant value indices to their max value (`0xFFFF`). If
the line number is greater than `u16::MAX` then the value is clamped at
that value.
```

## Open questions (optional)

Should the tagged abort code be replaced by the message as in the CLI or should the message be added separated?

kodemartin:
# Description of change

This patch implements the necessary conversions that would allow the json-rpc services to resolve clever errors upon execution failure.

GraphQL already handled this cases, and here we reuse the same functionality using `iota-json-rpc-types`.

Fixes #6256

Manual testing for Indexer JSON-RPC

Before the patch
```
$ curl -s -X POST https://api.testnet.iota.cafe   -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","id":1,"method":"iota_getTransactionBlock","params":["7iSLnPw8krZbeRkXnoVkyM2Edh2T7RBushPT1FPSrMmr", {"showEffects": true}]}' | jq .result.effects.status
{
  "status": "failure",
  "error": "MoveAbort(MoveLocation { module: ModuleId { address: 974d61b4e495aa6b464759cbb9e1acea321246151d5db9afd3aacd1d8cda03dc, name: Identifier(\"errors\") }, function: 3, instruction: 1, function_name: Some(\"vector_error\") }, 9223372165703925763) in command 0"
}
```
After the patch
```
$ curl -s -X POST http://localhost:9000   -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","id":1,"method":"iota_getTransactionBlock","params":["7iSLnPw8krZbeRkXnoVkyM2Edh2T7RBushPT1FPSrMmr", {"showEffects": true}]}' | jq .result.effects.status
{
  "status": "failure",
  "error": "Error in 1st command, from '0x974d61b4e495aa6b464759cbb9e1acea321246151d5db9afd3aacd1d8cda03dc::errors::vector_error' (line 30), abort 'EVectorError': This is an error message."
}
```

### Release Notes

- [x] JSON-RPC: Move Abort Errors are now resolved into a more human-readable form.


crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1494,6 +1599,7 @@ impl From<ExecutionStatus> for IotaExecutionStatus {
                 error,
                 command: Some(idx),
             } => Self::Failure {
+                // TODO: this is where we should intervene
```
tomxey: what does it mean?

crates/iota-json-rpc/src/read_api.rs
```diff
@@ -1385,12 +1398,12 @@ fn convert_to_response(
...
     if opts.show_effects && cache.effects.is_some() {
-        let effects = cache.effects.unwrap().try_into().map_err(|e| {
-            anyhow!(
-                // TODO: is this a client or server error?
-                "Failed to convert transaction block effects with error: {e}"
-            )
-        })?;
+        let native_effects = cache
+            .effects
+            .expect("show_effects should have populated effects");
```
tomxey, line 1403: are we 100% sure this will not happen?
returning a Result seemed safer

crates/iota-json-rpc-types/src/iota_transaction.rs
kodemartin: it means that I should remove it 🤣 commend added during my first scan to act as a placeholder. Forgot to clean up

crates/iota-json-rpc/src/read_api.rs
kodemartin: it was `unwrap` before

tomxey: indeed, missed that 👌

tomxey (approved):

thibault-martinez (approved):

miker83z (approved):

crates/iota-indexer/tests/data/clever_errors/sources/clever_errors.move
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2024 IOTA Stiftung
```
Dkwcs: Should the date be different ? Like 2026

crates/iota-indexer/tests/data/clever_errors/sources/clever_errors.move
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+module clever_errors::clever_errors {
+    #[error]
+    const ENotFound: vector<u8> = b"Element not found in vector 💥 🚀 🌠";
```
Dkwcs, line 7: I’m wondering why the error looks like this.
I see that Sui follows the same approach, but is there something behind it - perhaps a specific reason?

Dkwcs (approved):
LGTM, left minor comments

crates/iota-indexer/tests/data/clever_errors/sources/clever_errors.move
kodemartin: What do you mean? THese are utf-8 characters.

sergiupopescu199 (approved):
lgtm 🚀

muXxer (approved):

crates/iota-indexer/tests/data/clever_errors/sources/clever_errors.move
kodemartin: https://github.com/iotaledger/iota/pull/10087/commits/27ea64f05bf972951baadd8303ac404f9e3d3755

Thoralf-M (approved):

crates/iota-indexer/tests/data/clever_errors/sources/clever_errors.move
Dkwcs: I meant this💥 🚀 🌠
image

PR #10093, opened 2026-02-05
feat(iota-data-ingestion-core): enable usage of gRPC checkpoint stream

### LINKED ISSUE #7923: [data-ingestion]: support gRPC streams as checkpoint source

sergiupopescu199:
# Description of change

This patch refactors the `CheckpointReader` V2 with the ability to use the IOTA Node's gRPC checkpoint stream, as an alternative to REST API.

fixes #7923

All the tests were done locally by using the `iota` cli to create a private network:

- `cargo r --features indexer -- start --network.config ./network`
- Stoped the network
- Added these changes to enable gRPC in the `fullnode.yaml` config file 
```yaml
enable-grpc-api: true
grpc-api-config:
  address: "0.0.0.0:50051"
```
- Started the network again with gRPC enabled this time
- Used the `remote_reader.rs` to start a custom indexer by using the `RemoteUrl::Fullnode("http://localhost:50051".into())`
- inspected that checkpoint were received with either gRPC and REST API.
- genesis checkpoint is also downloaded correctly, used a gRPC connection to alphanet node.

> [!NOTE]
> This patch does not affect the normal operation of the indexer since it uses by default the `CheckpointReader` V1

### Release Notes
- [x] Indexer: add gRPC checkpoint stream support for the `CheckpointReader` V2.


crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -14,8 +14,8 @@ use iota_config::{
     node::ArchiveReaderConfig,
     object_storage_config::{ObjectStoreConfig, ObjectStoreType},
 };
-use iota_grpc_api::{CheckpointClient, CheckpointContent, NodeClient};
-use iota_grpc_types::CheckpointData as GrpcCheckpointData;
+use iota_grpc_client::Client as GrpcClient;
+use iota_grpc_server::constants::MAX_MESSAGE_SIZE_BYTES;
```
muXxer: In my opinion you should not have `iota_grpc_server` as a dependency at all. And it's only for this constant here right?

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -311,85 +323,22 @@ impl CheckpointReaderActor {
...
-        }
-        let show_full_checkpoint = true;
+    async fn relay_from_fullnode(&mut self, client: &mut GrpcClient) -> IngestionResult<()> {
         let mut checkpoints_stream = client
             .stream_checkpoints(
                 Some(self.current_checkpoint_number),
-                None, // internally it resolves to u64::MAX
-                show_full_checkpoint,
+                None,
+                Some("checkpoint,transactions"),
```
muXxer, line 343: Do you really need all info? For example the transaction effects/events etc? Or is a more granular read mask maybe better?

muXxer (changes requested):

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: @muXxer 

In this case the framework exposes to crate users the [`CheckpointData`](https://iotaledger.github.io/iota/iota_types/full_checkpoint_content/struct.CheckpointData.html) struct, it does require all these info, like events, effects, input & output objects ect.

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10093/commits/6512abaa1010e9d948c1fa7a7b6bc5919dbab62e

muXxer (approved):

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -671,30 +647,53 @@ impl CheckpointReader {
...
+            transaction: iota_sdk_types::SignedTransaction {
+                transaction: value.transaction().map_err(Error::ProtoConversion)?,
+                signatures: value.signatures().map_err(Error::ProtoConversion)?,
+            },
+            effects: value.effects().map_err(Error::ProtoConversion)?,
+            events: value.events().map_err(Error::ProtoConversion)?,
+            input_objects: value
+                .input_objects()
+                .map_err(Error::ProtoConversion)?
+                .unwrap_or_default(),
```
tomxey: is it still a correct response if this field is missing?

bingyanglin (approved):

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: In gRPC & proto all fields are optional, so the server implementation must ensure that data is correctly transmitted.

Inspecting the server side code the input & output objects are always returned when providing the `transactions` readmask, which we do.

tomxey: so maybe we should return error if they are missing, and not return the default?
I guess it's better to not hide the error in case it would appear in the future

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10093/commits/02cb0415f4812a969f7219d32099d2099d60367c

tomxey (approved):

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -101,11 +108,32 @@ impl RemoteStore {
...
+            RemoteUrl::Fullnode(ref url) => {
+                let grpc_client = GrpcClient::connect(url)
+                    .and_then(|grpc_client| async {
+                        // check if we can make gRPC request to client
+                        match grpc_client.get_service_info(None).await {
+                            Ok(_) => Ok(grpc_client.with_max_decoding_message_size(
+                                GRPC_MAX_DECODING_MESSAGE_SIZE_BYTES,
+                            )),
+                            Err(e) => Err(e),
+                        }
```
kodemartin: Nit: You can simplify to 
```rust
                        grpc_client.get_service_info(None).await?;
                        Ok(grpc_client
                            .with_max_decoding_message_size(GRPC_MAX_DECODING_MESSAGE_SIZE_BYTES))
```

crates/iota-data-ingestion-core/src/reader/v2.rs
```diff
@@ -101,11 +108,32 @@ impl RemoteStore {
...
+                        RemoteStore::Fullnode(grpc_client)
+                    }
+                    // fallback to REST API if gRPC fails
+                    Err(e) => {
+                        debug!("unable to establish a gRPC connection to fullnode: {e}");
+                        info!("using REST API as checkpoint stream");
+                        RemoteStore::RestApiFullnode(iota_rest_api::Client::new(url))
+                    }
+                }
+            }
```
kodemartin: Nit: matching directly on L112 would be more compact.

kodemartin (approved):
lgtm 🎉  with a few nits

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10093/commits/8be6996d18298429726a821bbeca6358ead85619

crates/iota-data-ingestion-core/src/reader/v2.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10093/commits/8be6996d18298429726a821bbeca6358ead85619

PR #10100, opened 2026-02-05
feat(docker): install heaptrack instead of jemalloc in all docker images

muXxer:
# Description of change

This PR introduces the `heaptrack` binary in all docker images, so that memory profiling becomes easier.

To create a profile, you need to build the docker image with rustflags set:
```bash
export RUSTFLAGS="-C debuginfo=2"
# Then build your Docker image
docker build --build-arg RUSTFLAGS="-C debuginfo=2" -t iotaledger/iota-node .
```

Or use the provided debug build script:

```bash
./docker/iota-node/build_debuginfo.sh
```

To profile memory usage with Docker containers, you can run the container with heaptrack enabled:

```bash
docker run -it --rm \
  -v $(pwd)/heaptrack-data:/tmp/heaptrack-data \
  iotaledger/iota-node \
  heaptrack -o /tmp/heaptrack-data/iota-profile iota-node [your-node-arguments]
```

You can also attach heaptrack to an already running process inside a Docker container:
```bash
docker exec -it <container-name> heaptrack -p <pid> -o /tmp/profile-name
```

This will:

1. Run the IOTA node under heaptrack profiling
2. Save the profiling data to a volume-mounted directory for analysis
3. Generate a profile file named `iota-profile.heaptrack` in the mounted directory

To analyze the generated profiles:

1. Install heaptrack on your host system. On Debian/Ubuntu: `apt-get install heaptrack-gui heaptrack`.
2. Open the profile file with the heaptrack GUI: `heaptrack_gui /path/to/iota-profile.heaptrack`.
3. Or generate text reports: `heaptrack_print /path/to/iota-profile.heaptrack`.

> **tip**
>
> For production environments, consider the performance impact of memory profiling. Heaptrack can introduce overhead, so use it primarily for debugging memory issues or performance analysis in non-production environments.

### DIFF
--- a/crates/telemetry-subscribers/observability_guides.md
+++ b/crates/telemetry-subscribers/observability_guides.md
@@ -197,29 +197,49 @@ To exit the process on panic, set the `CRASH_ON_PANIC` environment variable when
 
 ### Memory Profiling
 
-IOTA uses the [jemalloc memory allocator](https://jemalloc.net/) by default on most platforms, and there is code that enables automatic memory profiling using jemalloc's sampling profiler, which is very lightweight and designed for production use. The profiling code spits out profiles, at most, every 5 minutes, and only when total memory has increased by a default 20%. Profiling files are named `jeprof.<TIMESTAMP>.<memorysize>MB.prof` so that it is easy to
-correlate to metrics and incidents, for ease of debugging.
+IOTA Docker images include [heaptrack](https://github.com/KDE/heaptrack), a heap memory profiler that tracks all memory allocations and can help identify memory leaks and excessive memory usage patterns. Heaptrack provides detailed information about where memory is allocated and can generate comprehensive reports for analysis.
 
-For the memory profiling to work, you need to set the environment variable `_RJEM_MALLOC_CONF=prof:true`. If you use the [Docker image](https://hub.docker.com/r/iotaledger/iota-node) they are set automatically.
+For optimal profiling results with readable stack traces and function names, build Docker images with debug symbols by setting RUSTFLAGS before building:
 
-Running some allocator-based heap profilers such as [Bytehound](https://github.com/koute/bytehound) will essentially disable automatic jemalloc profiling, because they interfere with or don't implement `jemalloc_ctl` stats APIs.
+```bash
+export RUSTFLAGS="-C debuginfo=2"
+# Then build your Docker image
+docker build --build-arg RUSTFLAGS="-C debuginfo=2" -t iotaledger/iota-node .
+```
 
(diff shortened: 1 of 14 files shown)


crates/telemetry-subscribers/observability_guides.md
```diff
@@ -197,29 +197,49 @@ To exit the process on panic, set the `CRASH_ON_PANIC` environment variable when
...
-correlate to metrics and incidents, for ease of debugging.
+IOTA Docker images include [heaptrack](https://github.com/KDE/heaptrack), a heap memory profiler that tracks all memory allocations and can help identify memory leaks and excessive memory usage patterns. Heaptrack provides detailed information about where memory is allocated and can generate comprehensive reports for analysis.
 
-For the memory profiling to work, you need to set the environment variable `_RJEM_MALLOC_CONF=prof:true`. If you use the [Docker image](https://hub.docker.com/r/iotaledger/iota-node) they are set automatically.
+For optimal profiling results with readable stack traces and function names, build Docker images with debug symbols by setting RUSTFLAGS before building:
 
-Running some allocator-based heap profilers such as [Bytehound](https://github.com/koute/bytehound) will essentially disable automatic jemalloc profiling, because they interfere with or don't implement `jemalloc_ctl` stats APIs.
+```bash
+export RUSTFLAGS="-C debuginfo=2"
```
semenov-vladyslav, line 205:
```suggestion
```
`--build-arg RUSTFLAGS` should be enough.

semenov-vladyslav (approved):

docker/iota-indexer/Dockerfile
```diff
@@ -29,6 +31,9 @@ COPY external-crates external-crates
 COPY iota-execution iota-execution
 COPY Cargo.toml Cargo.lock ./
 
+# Set RUSTFLAGS if provided via build argument
+ENV RUSTFLAGS=${RUSTFLAGS}
```
bingyanglin, line 35: This will directly overwrite the flag `-C link-arg=-fuse-ld=lld` in Line 21. Do you think something like the following might be better?

```bash
ARG RUSTFLAGS=""
ENV RUSTFLAGS="-C link-arg=-fuse-ld=lld ${RUSTFLAGS}"

# config.toml keeps only the linker, no rustflags:
RUN mkdir -p ~/.cargo && \
    echo -e "[target.x86_64-unknown-linux-gnu]\nlinker = \"clang\"" > ~/.cargo/config.toml
```

muXxer:
Turns out that heaptrack seems to be much slower than jemalloc. We decided to not do this change.

PR #10105, opened 2026-02-05
fix(ui-kit): Fix toggle right side spacing

marc2332:
also changed the animation because I like it more


evavirseda (approved):
finally, thank u!👏🏼

msarcev (changes requested):
Probably needs a changeset

msarcev (approved):

PR #10124, opened 2026-02-06
Support runtime ogd parameters via fullnode.yaml

salaheldinsoliman:
This PR adds support for runtime parameters of `OGD` implementation in `CongestionTracker`.

### DIFF
--- a/crates/iota-config/data/fullnode-template.yaml
+++ b/crates/iota-config/data/fullnode-template.yaml
@@ -24,3 +24,12 @@ authority-store-pruning-config:
   max-checkpoints-in-batch: 10
   max-transactions-in-batch: 1000
   pruning-run-delay-seconds: 60
+
+# Optional congestion tracker tuning. All fields are optional and
+# have defaults defined in the node config (see `CongestionTrackerConfig` in `node.rs`).
+# The defaults are:
+# congestion-tracker-config:
+#   cache-capacity: 10000
+#   hotness-cutoff: 1.0
+#   hotness-adjustment-factor: 1.0
+#   max-decay-factor: 2.0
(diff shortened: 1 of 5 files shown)


crates/iota-config/data/fullnode-template.yaml
```diff
@@ -24,3 +24,12 @@ authority-store-pruning-config:
   max-checkpoints-in-batch: 10
   max-transactions-in-batch: 1000
   pruning-run-delay-seconds: 60
+
+# Optional congestion tracker tuning. All fields are optional and
+# default to reasonable values if omitted.
```
vekkiokonio: What's a reasonable value? Either mention the default values or refer to `congestion_tracker.rs`

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -34,21 +35,52 @@ use crate::{
...
 
-/// Capacity of the congestion tracker's cache.
-const CONGESTION_TRACKER_CACHE_CAPACITY: u64 = 10_000;
-
-/// Threshold for hotness below which an object is considered cold.
-/// Values should be > 0.0. If HOTNESS_CUTOFF = 0.0, then no pruning will
-/// happen.
-const HOTNESS_CUTOFF: f64 = 1.0;
+/// Default capacity of the congestion tracker's cache.
+const DEFAULT_CONGESTION_TRACKER_CACHE_CAPACITY: u64 = 10_000;
```
vekkiokonio: These default values are now only used in the tests, so I don't see why they should be defined here. Maybe just define them in node.rs where they are actually set so that code is clean.

vekkiokonio (approved):
PR works as expected. Just left a couple of comments to clean up the code.

PR #10130, opened 2026-02-08
refactor(iota-data-ingestion): use gRPC for all workers

### LINKED ISSUE #10099: [data-ingestion]: use gRPC in all workers
With gRPC support form fullnodes we need to refactor the worker implementation in the `iota-data-ingestion`.

sergiupopescu199:
# Description of change

This PR replaces the REST API checkpoint source in favor of gRPC for the `iota-data-ingestion` worker implementation. 

- The `Blob` Worker responsible for the `live` service during initialization uses gRPC API to gather first epoch checkpoint sequence number instead of REST API.
- The remaining workers responsible for `kv-store` & `historical` services are not directly affected by the change since they're not using the gRPC API internally as the `live` service does, instead the ingestion framework uses the CheckpointReader V2 for downloading checkpoints from gRPC stream instead of REST API to forward to all workers.

fixes #10099

- All services compile and work as expected, all receive checkpoints and process them correctly.
- More attention was dedicated to `Blob` worker and it's additional logic for calculating the range of checkpoints to delete during epoch change.
- used the `iota` cli to spin up a private network with a epoch duration of 30s, generated genesis
```shell
cargo r -- genesis --epoch-duration-ms 30000 --with-faucet -f
```
- modified the `fullnode.yaml` and enabled grpc
- started the network
```shell
 cargo r -- start
```
- checked that after every epoch the range was calculated correctly and all data was removed form the bucket.
- also checked the calculation for the deletion range once the service is stopped upon restart it would correctly determine the range to first cleanup the bucket before starting syncing current epoch's checkpoints.

> [!NOTE]


crates/iota-data-ingestion/src/workers/blob.rs
```diff
@@ -195,11 +194,8 @@ impl Worker for BlobWorker {
         {
             let mut current_epoch = self.current_epoch.lock().await;
             if epoch > *current_epoch {
-                let delete_start = common::epoch_first_checkpoint_sequence_number(
-                    &self.rest_client,
-                    *current_epoch,
-                )
-                .await?;
+                let (_, delete_start) =
+                    common::epoch_info(&self.grpc_client, Some(*current_epoch)).await?;
```
tomxey: It looks a bit unclear what "delete_start" is as a result of "epoch_info".

I feel it would be more clear to suggest what the return value is in the function name, maybe sth like: "get_beginning_of_epoch_info"?

But I see you are also

tomxey (approved):

crates/iota-data-ingestion/src/workers/blob.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10130/commits/4c32d491dd315bf4f96a60b649869f30fe8583e7

kodemartin (approved):
lgtm ⛵

PR #10131, opened 2026-02-08
fix(iota-data-ingestion-core): stop downloading checkpoints once shutdown is triggered

### LINKED ISSUE #10129: [iota-data-ingestion-core]: stop downloading checkpoints once shutdown is triggered
When we cancel the token triggering the graceful shutdown the CheckpoinReader V2 keeps downloading checkpoints until it reaches the `LocalRead::exceeds_capacity` threshold, once reache dit exists the `self.sync()` and checks the shutdown signal in the `tokio::select`

We need to check the shutdown signal before sending a checkpoint to the Executor, in this way it will be more responsive instead of waiting until `self.sync()` which can take quite a while.

sergiupopescu199:
# Description of change

This patch fixes a bug in the CheckpointReader V2 where upon shutting down it would not respect it and keep downloading  checkpoints until the `LocalReade::exceeds_capacity` would resolve to true, subsequently make the `self.sync().await` to finish execution when lastly the `tokio::select` would check for shutdown signal.

fixes #10129 

- started the `live` service locally, since this bug was most noticeable in this particular use case.
- while syncing checkpoints from current epoch on testnet, the user would start the shutdown process by pressing `CRL+C`, Once all worker pools finished processing checkpoints the `CheckpointReader` V2 upon receiving the shutdown signal exists the `self.sync().await` promptly and subsequently terminates its operation.

> [!NOTE]
> This patch doe snot affect the normal operation of the indexer, thus no further tests were conducted.

### DIFF
--- a/crates/iota-data-ingestion-core/src/reader/v2.rs
+++ b/crates/iota-data-ingestion-core/src/reader/v2.rs
@@ -300,14 +300,24 @@ impl CheckpointReaderActor {
                     .pipe(futures::stream::iter)
                     .buffered(batch_size);
 
-                while let Some(checkpoint_result) = checkpoint_stream.next().await {
+                while let Some(checkpoint_result) = self
+                    .token
+                    .run_until_cancelled(checkpoint_stream.next())
+                    .await
+                    .flatten()
+                {
                     let (checkpoint, size) = checkpoint_result?;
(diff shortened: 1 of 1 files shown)


kodemartin (approved):
lgtm 🌻

tomxey (approved):

PR #10141, opened 2026-02-09
upstream(core): Introduce events_2 table to store events keyed by transaction digest

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/c023028e52147f7de0c14b62b7f591806bd5ea82
  - https://github.com/MystenLabs/sui/pull/21566/changes/38f0d3285b987efee5bda496820b1d19df5bd629
  - https://github.com/MystenLabs/sui/commit/8bbdf9fa93552603dc4251bf2894c2d246748192
- Description:

In Sui TransactionEvents are not unique and multiple transactions can
produce the exact same set of TransactionEvents. Today these events are
stored in a table keyed by TransactionEventsDigest which can lead to an
unfortunate scenario when pruning is enabled. If we have two
transactions `A` and `B`, both of which produced events `E`, when the
pruner goes to prune transaction `A` it will also remove events `E` from
the database. This will lead to a situation where the node will be
unable to find locally the events produced by transaction `B`.

In order to fix this bug this patch introduces a new `events_2` table to
store events keyed by the transaction digest of the transaction that
produced them. It also updates the interface used to read events from
the db to require a `TransactionDigest` instead of a
`TransactionEventsDigest`. Internally the `events_2` table is first checked
but then there is fallback logic to lookup the events from the `events`
table. This fallback logic can eventually be removed once we've
completed a db migration and deprecate the `events` table.

[run-ci]

Part of #9768

### DIFF
--- a/crates/iota-core/src/authority/authority_store.rs
+++ b/crates/iota-core/src/authority/authority_store.rs
@@ -399,18 +399,11 @@ impl AuthorityStore {
         &self,
         digest: &TransactionDigest,
     ) -> Result<Option<TransactionEvents>, TypedStoreError> {
-        // For now, during this transition period, if we don't find events for a
-        // particular Transaction we need to fallback to try and read from the
-        // older table. Once the migration has finished and we've removed the
-        // older events table we can stop doing the fallback
         if let Some(events) = self.perpetual_tables.events_2.get(digest)? {
             return Ok(Some(events));
         }
 
-        self.get_executed_effects(digest)?
-            .and_then(|effects| effects.events_digest().copied())
-            .and_then(|events_digest| self.get_events_by_events_digest(&events_digest).transpose())
-            .transpose()
+        Ok(None)
     }
 
     pub fn get_events_by_events_digest(
--- a/crates/iota-core/src/authority/authority_store.rs
+++ b/crates/iota-core/src/authority/authority_store.rs
@@ -401,11 +401,18 @@ impl AuthorityStore {
         &self,
         digest: &TransactionDigest,
     ) -> Result<Option<TransactionEvents>, TypedStoreError> {
+        // For now, during this transition period, if we don't find events for a
+        // particular Transaction we need to fallback to try and read from the
+        // older table. Once the migration has finished and we've removed the
+        // older events table we can stop doing the fallback
         if let Some(events) = self.perpetual_tables.events_2.get(digest)? {
             return Ok(Some(events));
         }
 
-        Ok(None)
+        self.get_executed_effects(digest)?
+            .and_then(|effects| effects.events_digest().copied())
+            .and_then(|events_digest| self.get_events_by_events_digest(&events_digest).transpose())
+            .transpose()
     }
 
     pub fn get_events_by_events_digest(
--- a/crates/iota-core/src/authority/authority_store.rs
+++ b/crates/iota-core/src/authority/authority_store.rs
@@ -345,11 +345,13 @@ impl AuthorityStore {
                         .unwrap();
 
                     // Insert to events_2 table
-                    store
-                        .perpetual_tables
-                        .events_2
-                        .insert(transaction.digest(), events)
-                        .unwrap();
+                    if effects.events_digest().is_some() {
+                        store
+                            .perpetual_tables
+                            .events_2
+                            .insert(transaction.digest(), events)
+                            .unwrap();
+                    }
                 }
             }
         }
--- a/crates/iota-core/src/authority/authority_store.rs
+++ b/crates/iota-core/src/authority/authority_store.rs
@@ -333,12 +333,23 @@ impl AuthorityStore {
                         .effects
                         .insert(&effects.digest(), effects)
                         .expect("cannot insert migration effects");
-                    let events = events
+                    let events_iter = events
                         .data
                         .iter()
                         .enumerate()
                         .map(|(i, e)| ((events.digest(), i), e));
-                    store.perpetual_tables.events.multi_insert(events).unwrap();
+                    store
+                        .perpetual_tables
+                        .events
+                        .multi_insert(events_iter)
+                        .unwrap();
+
+                    // Insert to events_2 table
+                    store
+                        .perpetual_tables
+                        .events_2
+                        .insert(transaction.digest(), events)
+                        .unwrap();
                 }
             }
         }
--- a/crates/iota-core/src/authority/authority_store.rs
+++ b/crates/iota-core/src/authority/authority_store.rs
@@ -276,6 +276,13 @@ impl AuthorityStore {
             // because the genesis tx hasn't but will be executed. This is
             // important for fullnodes to be able to generate indexing data
             // right now.
+            if genesis.effects().events_digest().is_some() {
+                store
+                    .perpetual_tables
+                    .events_2
+                    .insert(transaction.digest(), genesis.events())
+                    .unwrap();
+            }
             let event_digests = genesis.events().digest();
             let events = genesis
                 .events()
@@ -379,6 +386,24 @@ impl AuthorityStore {
 
     pub fn get_events(
         &self,
+        digest: &TransactionDigest,
+    ) -> Result<Option<TransactionEvents>, TypedStoreError> {
+        // For now, during this transition period, if we don't find events for a
+        // particular Transaction we need to fallback to try and read from the
+        // older table. Once the migration has finished and we've removed the
+        // older events table we can stop doing the fallback
+        if let Some(events) = self.perpetual_tables.events_2.get(digest)? {
+            return Ok(Some(events));
+        }
+
+        self.get_executed_effects(digest)?
+            .and_then(|effects| effects.events_digest().copied())
+            .and_then(|events_digest| self.get_events_by_events_digest(&events_digest).transpose())
+            .transpose()
+    }
+
+    pub fn get_events_by_events_digest(
+        &self,
         event_digest: &TransactionEventsDigest,
     ) -> Result<Option<TransactionEvents>, TypedStoreError> {
         let data = self
@@ -392,7 +417,7 @@ impl AuthorityStore {
 
     pub fn multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
     ) -> IotaResult<Vec<Option<TransactionEvents>>> {
         Ok(event_digests
             .iter()
@@ -410,7 +435,7 @@ impl AuthorityStore {
     pub fn get_executed_effects(
         &self,
         tx_digest: &TransactionDigest,
-    ) -> IotaResult<Option<TransactionEffects>> {
+    ) -> Result<Option<TransactionEffects>, TypedStoreError> {
         let effects_digest = self.perpetual_tables.executed_effects.get(tx_digest)?;
         match effects_digest {
             Some(digest) => Ok(self.perpetual_tables.effects.get(&digest)?),
@@ -869,6 +894,15 @@ impl AuthorityStore {
 
         write_batch.insert_batch(&self.perpetual_tables.objects, new_objects)?;
 
+        // Write events into the new table keyed off of transaction_digest
+        if effects.events_digest().is_some() {
+            write_batch.insert_batch(
+                &self.perpetual_tables.events_2,
+                [(transaction_digest, events)],
+            )?;
+        }
+
+        // Continue writing events into the old table for now keyed off of events digest
         let event_digest = events.digest();
         let events = events
             .data
@@ -1167,6 +1201,7 @@ impl AuthorityStore {
             iter::once(tx_digest),
         )?;
         if let Some(events_digest) = effects.events_digest() {
+            write_batch.delete_batch(&self.perpetual_tables.events_2, [tx_digest])?;
             write_batch.schedule_delete_range(
                 &self.perpetual_tables.events,
                 &(*events_digest, usize::MIN),
--- a/crates/iota-core/src/authority/authority_store_migrations.rs
+++ b/crates/iota-core/src/authority/authority_store_migrations.rs
@@ -16,19 +16,10 @@ pub async fn migrate_events(store: Arc<AuthorityStore>) {
     tracing::info!("Starting events table migration");
 
     let result = tokio::task::spawn_blocking(move || {
-        let span = tracing::info_span!(
-            "migrate_events",
-            transaction = tracing::field::Empty,
-            effects = tracing::field::Empty,
-        );
-        let _entered = span.enter();
-
         let mut batch = store.perpetual_tables.events_2.batch();
 
         for entry in store.perpetual_tables.executed_effects.safe_iter() {
             let (txn_digest, effects_digest) = entry?;
-            span.record("transaction", tracing::field::debug(&txn_digest));
-            span.record("effects", tracing::field::debug(&effects_digest));
 
             // If there's already an entry for this transaction in the new table we can skip
             // it
--- a/crates/iota-core/src/authority/authority_store_migrations.rs
+++ b/crates/iota-core/src/authority/authority_store_migrations.rs
@@ -58,7 +58,7 @@ pub async fn migrate_events(store: Arc<AuthorityStore>) {
                     tracing::warn!(
                         expected_events_digest =? events_digest,
                         fetched_events_digest =? fetched_events_digest,
-                        "fetched events don't matched expected digest; skipping",
+                        "fetched events don't match expected digest; skipping",
                     );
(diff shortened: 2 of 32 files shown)


crates/iota-core/src/authority/authority_store.rs
bingyanglin: I guess we should also write to the new `events_2` table?

bingyanglin (changes requested):
Also note that in latter PR https://github.com/MystenLabs/sui/pull/22715 the old event table was removed (we should do it in the **next next** release version). For now I guess we should enable dual-write so we can support rolling upgrade.

crates/iota-core/src/authority/authority_store.rs
jkrvivian: Yes, it's the newly added lines from 279 ~ 285.

Thoralf-M (approved):
Approving the changes in iota-light-client and iota-types

crates/iota-core/src/authority/authority_store.rs
bingyanglin: But what I mean is at [Line 336~341](https://github.com/iotaledger/iota/pull/10141/changes#diff-db6346bc77230504609a97fecb87844b0c9e3a484fd1f764189d146efe00df47R336-R341). Where the migration transaction are stored into the old table. I guess we need to enable dual-writing for both the old and new tables (`events` and `events_2`) in the next release version, and all read from the new table, and then in the next next release version we can safely remove it (Like https://github.com/MystenLabs/sui/pull/22715). Or do I miss anything?

jkrvivian: You are right!!! Added in 7fa766e89beb02b499b342b315b67472e1a27d2d

crates/iota-core/src/authority/authority_store_tables.rs
```diff
@@ -109,6 +112,9 @@ pub struct AuthorityPerpetualTables {
     // Also we need a pruning policy for this table. We can prune this table along with tx/effects.
     pub(crate) events: DBMap<(TransactionEventsDigest, usize), Event>,
 
+    // Events keyed by the digest of the transaction that produced them.
+    pub(crate) events_2: DBMap<TransactionDigest, TransactionEvents>,
```
semenov-vladyslav, line 116: Is it allowed to contain entries with empty events?

crates/iota-core/src/authority/authority_store.rs
```diff
@@ -326,12 +333,23 @@ impl AuthorityStore {
...
+                        .perpetual_tables
+                        .events
+                        .multi_insert(events_iter)
+                        .unwrap();
+
+                    // Insert to events_2 table
+                    store
+                        .perpetual_tables
+                        .events_2
+                        .insert(transaction.digest(), events)
```
semenov-vladyslav: Maybe we should not insert entries with empty events. We can check if there are events with `effects.events_digest().is_some()`. Otherwise this table can get polluted very fast.

crates/iota-core/src/authority/authority_store.rs
```diff
@@ -378,6 +396,17 @@ impl AuthorityStore {
     }
 
     pub fn get_events(
+        &self,
+        digest: &TransactionDigest,
+    ) -> Result<Option<TransactionEvents>, TypedStoreError> {
```
semenov-vladyslav, line 403: Empty events of type `Option<TransactionEvents>` can be represented in two ways: `None`, `Some(TransactionEvents::default())`. Do we want to distinguish those?

crates/simulacrum/src/lib.rs
```diff
@@ -647,7 +647,7 @@ impl<T, V: store::SimulatorStore> ReadStore for Simulacrum<T, V> {
 
     fn try_get_events(
         &self,
-        event_digest: &iota_types::digests::TransactionEventsDigest,
+        event_digest: &iota_types::digests::TransactionDigest,
```
semenov-vladyslav:
```suggestion
        digest: &iota_types::digests::TransactionDigest,
```

crates/iota-core/src/execution_cache.rs
```diff
@@ -840,22 +840,19 @@ pub trait TransactionCacheRead: Send + Sync {
 
     fn try_multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        digests: &[TransactionDigest],
```

crates/iota-core/src/execution_cache.rs
```diff
@@ -840,22 +840,19 @@ pub trait TransactionCacheRead: Send + Sync {
...
     ) -> IotaResult<Vec<Option<TransactionEvents>>>;
 
     /// Non-fallible version of `try_multi_get_events`.
     fn multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        digests: &[TransactionDigest],
```

crates/iota-transactional-test-runner/src/simulator_persisted_store.rs
```diff
@@ -766,7 +745,7 @@ impl ReadStore for PersistedStoreInnerReadOnlyWrapper {
 
     fn try_get_events(
         &self,
-        event_digest: &TransactionEventsDigest,
+        event_digest: &TransactionDigest,
```
semenov-vladyslav:
```suggestion
        digest: &TransactionDigest,
```

crates/iota-transactional-test-runner/src/simulator_persisted_store.rs
```diff
@@ -592,7 +571,7 @@ impl ReadStore for PersistedStore {
...
```
semenov-vladyslav:
```suggestion
        digest: &TransactionDigest,
```

crates/iota-transactional-test-runner/src/test_adapter.rs
```diff
@@ -2891,7 +2891,7 @@ impl ReadStore for IotaTestAdapter {
 
     fn try_get_events(
         &self,
-        event_digest: &TransactionEventsDigest,
+        event_digest: &TransactionDigest,
```
semenov-vladyslav:
```suggestion
        digest: &TransactionDigest,
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -251,20 +248,18 @@ pub trait ReadStore: ObjectStore {
             .expect("storage access failed")
     }
 
-    fn try_get_events(
-        &self,
-        event_digest: &TransactionEventsDigest,
-    ) -> Result<Option<TransactionEvents>>;
+    fn try_get_events(&self, event_digest: &TransactionDigest)
```
semenov-vladyslav:
```suggestion
    fn try_get_events(&self, digest: &TransactionDigest)
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -251,20 +248,18 @@ pub trait ReadStore: ObjectStore {
...
+    -> Result<Option<TransactionEvents>>;
 
     /// Non-fallible version of `try_get_events`.
-    fn get_events(&self, event_digest: &TransactionEventsDigest) -> Option<TransactionEvents> {
+    fn get_events(&self, event_digest: &TransactionDigest) -> Option<TransactionEvents> {
```
semenov-vladyslav:
```suggestion
    fn get_events(&self, digest: &TransactionDigest) -> Option<TransactionEvents> {
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -251,20 +248,18 @@ pub trait ReadStore: ObjectStore {
...
         self.try_get_events(event_digest)
             .expect("storage access failed")
     }
 
     fn try_multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        digests: &[TransactionDigest],
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -275,7 +270,7 @@ pub trait ReadStore: ObjectStore {
     /// Non-fallible version of `try_multi_get_events`.
     fn multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        digests: &[TransactionDigest],
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -559,14 +547,14 @@ impl<T: ReadStore + ?Sized> ReadStore for &T {
 
     fn try_get_events(
         &self,
-        event_digest: &TransactionEventsDigest,
+        event_digest: &TransactionDigest,
```
semenov-vladyslav:
```suggestion
        digest: &TransactionDigest,
```

crates/simulacrum/src/store/in_mem_store.rs
```diff
@@ -482,7 +475,7 @@ impl ReadStore for InMemoryStore {
 
     fn try_get_events(
         &self,
-        event_digest: &iota_types::digests::TransactionEventsDigest,
+        event_digest: &iota_types::digests::TransactionDigest,
```
semenov-vladyslav:
```suggestion
        digest: &iota_types::digests::TransactionDigest,
```

semenov-vladyslav (changes requested):

muXxer:
It's important that we port `8bbdf9fa93552603dc4251bf2894c2d246748192` as well.

crates/iota-core/src/authority/authority_store.rs
jkrvivian: Added in https://github.com/iotaledger/iota/pull/10141/commits/95bdf06e225b6bcf5c324d7552267da4ff75d888

crates/iota-core/src/authority/authority_store_tables.rs
jkrvivian: No, I don't think so, and in table migration, we also skip all the empty ones.

crates/iota-core/src/authority/authority_store.rs
jkrvivian: I checked the insertions to event_2, and there is no empty entry, so returning `None` makes sense, showing that the event does not exist in the table.

crates/iota-core/src/authority/authority_store_tables.rs
bingyanglin, line 218: Just a reminder this configuration becomes dead code after this PR, and https://github.com/MystenLabs/sui/commit/bb2939ca52 removed this.

crates/iota-core/src/authority/authority_store_migrations.rs
```diff
@@ -0,0 +1,100 @@
...
+        if events_data.is_empty() {
+            continue;
+        }
+
+        let events = TransactionEvents { data: events_data };
+        if &events.digest() != events_digest {
+            tracing::warn!(
+                expected_events_digest =? events_digest,
+                fetched_events_digest =? events.digest(),
+                "fetched events don't matched expected digest; skipping",
```
bingyanglin:
```suggestion
                "fetched events don't match expected digest; skipping",
```

crates/iota-core/src/authority/authority_store_tables.rs
jkrvivian: Actually, we should include that already in this branch.

bingyanglin (approved):

crates/iota-core/src/authority/authority_store_migrations.rs
```diff
@@ -0,0 +1,84 @@
...
+            if store.perpetual_tables.events_2.contains_key(&txn_digest)? {
+                continue;
+            }
+
+            let effects = if let Some(effects) = store.get_effects(&effects_digest)? {
+                effects
+            } else {
+                // Skip this one if we can't find the effects
+                continue;
+            };
```
semenov-vladyslav:
```suggestion
            let Some(effects) = store.get_effects(&effects_digest)? else {
                // Skip this one if we can't find the effects
                continue;
            };
```

crates/iota-core/src/authority/authority_store_migrations.rs
```diff
@@ -0,0 +1,84 @@
...
+
+            let events_digest = if let Some(events_digest) = effects.events_digest() {
+                events_digest
+            } else {
+                // There are no events so we can continue to the next entry
+                continue;
+            };
```
semenov-vladyslav:
```suggestion
            let Some(events_digest) = effects.events_digest() else {
                // There are no events so we can continue to the next entry
                continue;
            };
```

crates/iota-core/src/authority/authority_store_migrations.rs
```diff
@@ -0,0 +1,84 @@
...
+                        "fetched events don't match expected digest; skipping",
+                    );
+                    continue;
+                }
+                events
+            } else {
+                // Skip this one if we can't find the events. This means they were liked already
+                // pruned
+                continue;
+            };
```
semenov-vladyslav:
```suggestion
            let Some(events) = store.get_events_by_events_digest(events_digest)? else {
                // Skip this one if we can't find the events. This means they were liked already
                // pruned
                continue;
            };
            // Check that the events we're loading do match the expected events digest for
            // this transaction
            let fetched_events_digest = events.digest();
            if &fetched_events_digest != events_digest {
                tracing::warn!(
                    expected_events_digest =? events_digest,
                    fetched_events_digest =? fetched_events_digest,
                    "fetched events don't match expected digest; skipping",
                );
                continue;
            }
```

crates/iota-core/src/authority/authority_store_migrations.rs
```diff
@@ -0,0 +1,84 @@
...
+                // Check that the events we're loading do match the expected events digest for
+                // this transaction
+                let fetched_events_digest = events.digest();
+                if &fetched_events_digest != events_digest {
+                    tracing::warn!(
+                        expected_events_digest =? events_digest,
+                        fetched_events_digest =? fetched_events_digest,
+                        "fetched events don't match expected digest; skipping",
+                    );
+                    continue;
```
semenov-vladyslav:
```suggestion
```
Weird logic to skip event when the digests don't match. If that happens that would indicate about some broken logic. Why not insert the value that is expected (ie. not skip writing the entry)?

crates/iota-core/src/authority/authority_store.rs
semenov-vladyslav: In writeback_cache.rs there is a comment:

> if events.data.is_empty(), then there are no events for this transaction. We store it anyway to avoid special cases in commint_transaction_outputs, and translate an empty events structure to None when reading.

It contradicts the implementation here: there should be no None entries, and no events should be encoded as empty events, not None.

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -938,16 +932,9 @@ impl WritebackCache {
         // commint_transaction_outputs, and translate an empty events structure
```
semenov-vladyslav:
```suggestion
        // commit_transaction_outputs, and translate an empty events structure
```

crates/iota-core/src/execution_cache/writeback_cache.rs
```diff
@@ -2053,7 +2031,7 @@ impl TransactionCacheRead for WritebackCache {
     #[instrument(level = "trace", skip_all)]
     fn try_multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        tx_digests: &[TransactionDigest],
```

crates/iota-core/src/execution_cache/passthrough_cache.rs
```diff
@@ -255,7 +255,7 @@ impl TransactionCacheRead for PassthroughCache {
     #[instrument(level = "trace", skip_all)]
     fn try_multi_get_events(
         &self,
-        event_digests: &[TransactionEventsDigest],
+        event_digests: &[TransactionDigest],
```
semenov-vladyslav:
```suggestion
        tx_digests: &[TransactionDigest],
```

semenov-vladyslav (approved):
Looks good except for a few minor nits.

kodemartin (approved):

polinikita (approved):

PR #10145, opened 2026-02-09
feat(dapp): refactor getAppsBackend

### LINKED ISSUE #9666: refactor `getAppsBackend` function out of the sdk
`getAppsBackend` does not belong to the public SDK but is rather an internal env. We should move it out and probably put it into apps core. We need to think where to put it without it being anoying so it can be maintained properly. Then we have external apps like iota names that now use it so we might need to redefine it there

KeitoTadashi:
# Description of change

fixes #9666 


marc2332 (commented):
How does one now specify a custom apps backend to the dashboad or wallet? is setting the env var in the shell the only way?

.changeset/twelve-tables-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@iota/iota-sdk': patch
+---
+
+refactor getAppsBackend
```
begonaalvarezd:
```suggestion
Remove getAppsBackend function
```

marc2332 (approved):

msarcev (approved):

begonaalvarezd (changes requested):
https://github.com/iotaledger/iota/blob/b44c8e3eda4b49c90d9adc0c667935b649a56ceb/.github/workflows/changesets_publish.yml#L20
This could be removed too right?

begonaalvarezd (approved):

PR #10146, opened 2026-02-09
feat(iota-core, iota-protocol-config): Add pass_calculated_validator_scores_to_advance_epoch flag

oliviasaa:
# Description of change

Introduces a new feature flag `pass_calculated_validator_scores_to_advance_epoch` (protocol v20) that decouples calculating validator scores from passing those calculated scores to the `advance_epoch` transaction.

Previously, when `calculate_validator_scores` was enabled, the locally computed scores were always forwarded to the epoch change transaction. This change adds a second gate: scores are still calculated and tracked, but a default score (`MAX_SCORE`) is passed to `advance_epoch` unless this new flag is also enabled.
  
### Release Notes

- [x] Protocol: Added `pass_calculated_validator_scores_to_advance_epoch` feature flag to decouple validator score calculation from score propagation to the epoch advance transaction. When disabled, a default score is passed instead of locally computed scores. This flag is enabled on Devnet only.

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -19,7 +19,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-pub const MAX_PROTOCOL_VERSION: u64 = 19;
+pub const MAX_PROTOCOL_VERSION: u64 = 20;
 
 // Record history of protocol version allocations here:
 //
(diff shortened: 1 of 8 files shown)


bingyanglin (approved):

miker83z (approved):

crates/iota-protocol-config/src/lib.rs
```diff
@@ -108,6 +108,9 @@ pub const MAX_PROTOCOL_VERSION: u64 = 19;
 //             Change epoch transaction will contain validator scores.
 //             Enable validator scoring on testnet and enable adjustment of
 //             validator rewards based on scores on Devnet.
+// Version 20: Supports the calculation of validator scores while still passing
+//             a default score value to the advance_epoch call. Enables this
+//             behavior on Testnet.
```
thibault-martinez: Devnet?

oliviasaa: maybe it's poorly phrased, but it's actually Testnet... 
right now, calculating the scores => passing them to advance epoch, so: 
- enabling the flag on devnet means that nothing changes there.
- the flag set to false on mainnet also does nothing, because calculate_validator_scores is also already false.
- this flag being false on Testnet is actually what changes something, because right now we calculate scores and pass them to advance epoch there

oliviasaa: I'll phrase this better before merging 😅

thibault-martinez: Oh ok I see, thanks for the explanation :)

kodemartin (approved):

alexsporn (approved):

PR #10183, opened 2026-02-11
feat(gRPC): selective accessor generation

muXxer:
# Description of change

This PR enables the iota-protoc-build crate to generate accessors for the proto message fields.

Fields that need accessors are annotated in the `.proto` files using a custom option:

```protobuf
import "iota/grpc/options.proto";

message ObjectRequest {
  optional ObjectReference object_ref = 1 [(iota.grpc.generate_accessors) = "set,with"];
}
```

### DIFF
--- a/crates/iota-proto-build/src/codegen/accessor_config.rs
+++ b/crates/iota-proto-build/src/codegen/accessor_config.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
-use std::{collections::HashMap, path::Path};
+use std::collections::HashMap;
 
 use bitflags::bitflags;
 use prost_types::FieldDescriptorProto;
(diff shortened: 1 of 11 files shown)


Thoralf-M (approved):
Approving the crates/iota-stardust-types/Cargo.toml change

scripts/update_grpc_types.sh
```diff
@@ -20,3 +20,13 @@ trap cleanup EXIT
 rm -Rf crates/iota-grpc-types/src/proto/generated/
 mkdir -p crates/iota-grpc-types/src/proto/generated/
 cargo run -p iota-proto-build
+exit_code=$?
```
bingyanglin, line 21: Because we have the `set -e` in the previous Line 6, the shell terminates before reaching `exit_code=$?` if the `cargo run` has errors, and the following lines will never be reached?

crates/iota-proto-build/src/codegen/accessors.rs
```diff
@@ -258,94 +439,150 @@ fn generate_accessors_functions_for_field(
...
-                pub fn #name_mut(&mut self) -> &mut #field_type_path {
-                    if self.#name_opt_mut().is_none() {
-                        self.#oneof_field = Some(#oneof_type_path::#variant(#field_type_path::default()));
+            } else {
+                quote! {
+                    #( #[doc = #name_mut_comments] )*
+                    pub fn #name_mut(&mut self) -> &mut #field_type_path {
+                        if self.#name_opt_mut().is_none() {
+                            self.#oneof_field = Some(#oneof_type_path::#variant(#field_type_path::default()));
+                        }
```
bingyanglin: Here we have a hidden rule that if the user sets the `mut` flag but does not set the "mut_opt" flag, it will have a compiling error. Do we want to automatically include the `mut_opt` when there is `mut` for the `oneof` fields? Or do we specify this rule at `crates/iota-proto-build/README.md`?

crates/iota-proto-build/src/codegen/accessor_config.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    map
+}
+
+/// Parse a proto file content to extract accessor annotations
+///
+/// Remark: We parse the file manually using regex to avoid adding a full proto
+/// parser dependency like proto-reflect, For our use case this is simpler and
+/// sufficient.
```
bingyanglin: There might be a bug here? Let's take `CheckpointData` in `ledger_service.proto` for example, 
```proto
message CheckpointData {
  message EndMarker {
    // The height of this checkpoint.
    optional uint64 sequence_number = 1;
  }

  oneof payload {
    iota.grpc.v0.checkpoint.Checkpoint checkpoint = 1;
    iota.grpc.v0.transaction.ExecutedTransactions transactions = 2;
    iota.grpc.v0.event.Events events = 3;
    EndMarker end_marker = 4;
  }
}
```

if we add an accessor annotation like 
```proto
  oneof payload {
    iota.grpc.v0.checkpoint.Checkpoint checkpoint = 1
        [(iota.grpc.generate_accessors) = "getter,set,with"];
    ...
  }
```

The regex parser finds two message starts: `CheckpointData` at position X, `EndMarker` at position Y. It processes `CheckpointData`'s content as `content[X..Y]`, which only contains the `message EndMarker {` line. The `oneof payload` fields come after EndMarker's start, so they fall into EndMarker's slice content[Y..].

In this way, the accessor map gets key `EndMarker.checkpoint` instead of `CheckpointData.checkpoint`. Then in from_field(), the code looks up `CheckpointData.checkpoint`, which finds nothing, thus silently generates no accessors.

muXxer: Fixed that in [aad728b](https://github.com/iotaledger/iota/pull/10183/commits/aad728b6facf4b31fef720022b2f3049a4720d24)

crates/iota-proto-build/src/codegen/accessors.rs
muXxer: I inlined everything in [3e6c34c](https://github.com/iotaledger/iota/pull/10183/commits/3e6c34c8cf4014ec4ec44d2bf6fc65119a9d88d8)

scripts/update_grpc_types.sh
muXxer: [7c93e5d](https://github.com/iotaledger/iota/pull/10183/commits/7c93e5d483a23932f3e678ac91378965f73b0f34)

PR #10189, opened 2026-02-11
feat(iota-core, consensus): Misbehavior report rate limit

oliviasaa:
# Description of change

- Adds information about the last submitted report (sequence number and summary) to `Scorer`.
- Rate-limit report submission: reports are only sent when all three conditions are met:
    1. Metrics have changed since the last report (according to report summary)
    2. At least 1000 checkpoints have passed since the last report
    3. The current checkpoint is at most 100 behind the highest verified checkpoint (avoids sending reports during resync)
- Forces a report to be sent when the epoch is coming to an end:
   1. A report is additionally sent when `checkpoint_timestamp >= next_reconfiguration_timestamp_ms - 2000`. This condition should be sufficient so that most authorities can have their reports included before the epoch closes.

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -4275,40 +4275,41 @@ impl AuthorityPerEpochStore {
                     );
                     return Ok(ConsensusCertificateResult::ConsensusMessage);
                 }
-                if self
+                if !self
                     .get_reconfig_state_read_lock_guard()
                     .should_accept_consensus_certs()
                 {
-                    let authority_index = self
-                        .committee
-                        .authority_index(authority)
-                        .expect("authority in committee");
-                    // Check validity of the report and update scores depending on the result. We
-                    // already have consensus on inclusion of this report in the DAG.
-                    if report.is_valid_version(self.protocol_config()) {
-                        if !report.verify(self.committee.num_members()) {
-                            self.scorer.increment_invalid_reports_count(authority_index);
-                            warn!(
-                                "Received invalid misbehavior report from {:?}",
-                                authority.concise()
-                            );
-                        } else {
-                            // Here we update all counts related to the information in the
-                            // reports.
-                            self.scorer.update_received_reports(authority_index, report);
-                        }
-                    } else {
-                        self.scorer.increment_invalid_reports_count(authority_index);
-                        warn!(
-                            "Received misbehavior report with unsupported version from {:?}",
-                            authority.concise()
-                        );
-                    }
-                } else {
                     debug!(
                         "Ignoring misbehavior report from {:?} because of end of epoch",
                         authority.concise()
                     );
+                    return Ok(ConsensusCertificateResult::ConsensusMessage);
+                }
+
+                // Safe: verify_consensus_transaction already confirmed that the
+                // report's authority matches the consensus block author, who is
+                // always a committee member.
+                let authority_index = self
+                    .committee
+                    .authority_index(authority)
+                    .expect("authority in committee");
+                // Check validity of the report and update scores depending on
+                // the result. We already have consensus on inclusion of this
+                // report in the DAG.
+                if !report.is_valid_version(self.protocol_config()) {
+                    self.scorer.increment_invalid_reports_count(authority_index);
+                    warn!(
+                        "Received misbehavior report with unsupported version from {:?}",
+                        authority.concise()
+                    );
+                } else if !report.verify(self.committee.num_members()) {
+                    self.scorer.increment_invalid_reports_count(authority_index);
+                    warn!(
+                        "Received invalid misbehavior report from {:?}",
+                        authority.concise()
+                    );
+                } else {
+                    self.scorer.update_received_reports(authority_index, report);
                 }
 
                 Ok(ConsensusCertificateResult::ConsensusMessage)
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -4268,22 +4268,54 @@ impl AuthorityPerEpochStore {
                 kind: ConsensusTransactionKind::MisbehaviorReport(authority, report, _),
                 ..
             }) => {
-                let authority_index = self
-                    .committee
-                    .authority_index(authority)
-                    .expect("authority in committee");
-                // Check validity of the report and update scores depending on the result. We
-                // already have consensus on inclusion of this report in the DAG.
-                if !report.verify(self.committee.num_members()) {
-                    self.scorer.update_invalid_reports_count(authority_index);
+                if !self.protocol_config().calculate_validator_scores() {
                     warn!(
-                        "Received invalid misbehavior report from {:?}",
+                        "Received misbehavior report from {:?} but validator scores are disabled, so the report is ignored",
                         authority.concise()
                     );
-                } else {
-                    // Here we update all counts related to the information in the reports.
-                    self.scorer.update_received_reports(authority_index, report);
+                    return Ok(ConsensusCertificateResult::ConsensusMessage);
                 }
+                if self
+                    .get_reconfig_state_read_lock_guard()
+                    .should_accept_consensus_certs()
+                {
+                    let authority_index = self
+                        .committee
+                        .authority_index(authority)
+                        .expect("authority in committee");
+
+                    // Check validity of the report and update scores depending on the result. We
+                    // already have consensus on inclusion of this report in the DAG.
+                    match (report, self.protocol_config().scorer_version_as_option()) {
+                        (VersionedMisbehaviorReport::V1(..), Some(1))
+                        | (VersionedMisbehaviorReport::V1(..), None) => {
+                            if !report.verify(self.committee.num_members()) {
+                                self.scorer.update_invalid_reports_count(authority_index);
+                                warn!(
+                                    "Received invalid misbehavior report from {:?}",
+                                    authority.concise()
+                                );
+                            } else {
+                                // Here we update all counts related to the information in the
+                                // reports.
+                                self.scorer.update_received_reports(authority_index, report);
+                            }
+                        }
+                        _ => {
+                            self.scorer.update_invalid_reports_count(authority_index);
+                            warn!(
+                                "Received misbehavior report with unsupported version from {:?}",
+                                authority.concise()
+                            );
+                        }
+                    }
+                } else {
+                    debug!(
+                        "Ignoring misbehavior report from {:?} because of end of epoch",
+                        authority.concise()
+                    );
+                }
+
                 Ok(ConsensusCertificateResult::ConsensusMessage)
             }
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -2941,7 +2941,7 @@ impl AuthorityPerEpochStore {
                         authority, transaction.certificate_author_index
                     );
                     self.scorer
-                        .update_invalid_reports_count(transaction.certificate_author_index);
+                        .increment_invalid_reports_count(transaction.certificate_author_index);
                     return None;
                 }
             }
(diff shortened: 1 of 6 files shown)


crates/iota-core/src/consensus_adapter.rs
```diff
@@ -1107,13 +1159,49 @@ impl ReconfigurationInitiator for Arc<ConsensusAdapter> {
...
         };
         if send_end_of_publish {
-            info!(epoch=?epoch_store.epoch(), "Sending EndOfPublish message to consensus");
-            if let Err(err) = self.submit(
-                ConsensusTransaction::new_end_of_publish(self.authority),
-                None,
-                epoch_store,
-            ) {
-                warn!("Error when sending end of publish message: {:?}", err);
+            if epoch_store.protocol_config().calculate_validator_scores() {
```
piotrm50: I think we don't need for the queue to be drained. It's enough that we send it whenever epoch closing process begins, not necessarily AFTER we drain the pending cert queue. 

Additionally, 
```
                eop_transactions.push(report_tx);
```

this basically creates a soft-bundle, which forces consensus to process those transaction together - put them in the same block and maintain order of those transactions. But I think it's not necessary in this case. Consensus takes transactions from the adapter in the same order as they were sent to it, but  it can insert them into earlier blocks. Since it's ok to drain all remaining certs before sending an EoP transaction, it should be also OK to send the last report separately, as long as it's submitted before the EoP.

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -867,14 +884,49 @@ impl ConsensusAdapter {
...
+                eop_transactions.push(ConsensusTransaction::new_end_of_publish(self.authority));
+
+                let checkpoint_seq = self
+                    .checkpoint_store
+                    .get_highest_executed_checkpoint_seq_number()
+                    .ok()
+                    .unwrap_or(None)
+                    .unwrap_or(0);
+                let misbehavior_report = epoch_store.scorer.current_local_metrics_count.to_report();
+                let report_tx = ConsensusTransaction::new_misbehavior_report(
```
piotrm50: I don't think this is necessary here - it should be enough to only send it in `ReconfigurationInitiator.close_epoch` function as long as we don't send it on the condition that pending cert queue is drained.

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -575,16 +575,28 @@ impl ConsensusAdapter {
...
-            // kind. The check is required because we assume this in
-            // submit_and_wait_inner.
-            for transaction in transactions {
-                fp_ensure!(
-                    matches!(
-                        transaction.kind,
-                        ConsensusTransactionKind::CertifiedTransaction(_)
-                    ),
-                    IotaError::InvalidTxKindInSoftBundle
+            // kind. The check is required because we assume this in submit_and_wait_inner.
```
piotrm50: Unnecessary complication.

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,22 +123,41 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
+            if new_report_summary != epoch_store.scorer.last_report_summary() {
+                let transaction = ConsensusTransaction::new_misbehavior_report(
+                    epoch_store.name,
+                    &misbehavior_report,
+                    checkpoint_seq,
+                );
+                info!(?transaction, "submitting misbehavior report to consensus");
+                self.sender
+                    .submit_to_consensus(&[transaction], epoch_store)?;
+                epoch_store
```
piotrm50, line 173: why are we storing last report summary here? I think this should be done post consensus.

crates/iota-core/src/consensus_adapter.rs
oliviasaa: This was reverted, as discussed

crates/iota-core/src/consensus_adapter.rs
oliviasaa: This was reverted, as discussed

crates/iota-core/src/consensus_adapter.rs
oliviasaa: This was reverted, as discussed

crates/iota-core/src/authority/scorer.rs
```diff
@@ -165,6 +175,24 @@ impl Scorer {
...
+        self.last_report_checkpoint_seq
+            .store(checkpoint_seq, Ordering::Relaxed)
+    }
+
+    pub(crate) fn last_report_summary(&self) -> u64 {
+        self.last_report_summary.load(Ordering::Relaxed)
+    }
+
+    pub(crate) fn store_last_report_summary(&self, summary: u64) {
+        self.last_report_checkpoint_seq
```
semenov-vladyslav:
```suggestion
        self.last_report_summary
```

semenov-vladyslav (changes requested):

semenov-vladyslav:
If a validator exceeds this rate-limit too often is it considered a misbehavior? Should it be reported?

crates/iota-core/src/checkpoints/checkpoint_output.rs
oliviasaa: This summary here has nothing to do with the report summary that should be added to storage, this value is supposed to be used just to rate limit the report creation. 
That being said, I have no strong opinion _against_ updating this summary after consensus, we might just end up sending some extra reports if we wait consensus (which I suppose it's not a problem)

crates/iota-core/src/authority/scorer.rs
oliviasaa: fixed in [82167e8](https://github.com/iotaledger/iota/pull/10189/commits/82167e88a10f41f117a2ba6ba27712e95dad4e27)

oliviasaa:
> If a validator exceeds this rate-limit too often is it considered a misbehavior? Should it be reported?

This PR just adds a simple rate limit to avoid wasting resources. For now, we don't count how many reports the others sent.
Now, we can definitely add this misbehavior in the future, specially because it doesn't need to be reported (all authorities receive the same reports at this stage, thus they would already share their view of the misbehavior).

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,22 +123,50 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
+        //
+        // Additionally to these periodic reports, we also send a report when the epoch
+        // is coming to an end. Since `close_epoch` is called according to local clocks,
+        // we use an analogous rule for the last reports, requiring that the checkpoint
+        // is close to the next reconfiguration timestamp.
+        let should_send_last_report =
+            checkpoint_timestamp >= self.next_reconfiguration_timestamp_ms - 2000;
+        if epoch_store.protocol_config().calculate_validator_scores()
+            && ((checkpoint_seq - epoch_store.scorer.last_report_checkpoint_seq() >= 1000
+                && Some(checkpoint_seq + 100) >= highest_verified_checkpoint)
```
cyberphysic4l: Perhaps consts for these values defined at the top of the file would be better than the magic numbers, even though their origin is explained in the comment.

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,22 +123,50 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
+        // We also require that the checkpoint for which we want to send the report is
+        // at most 100 checkpoints behind the highest verified checkpoint, to avoid
+        // sending reports during resync.
+        //
+        // Additionally to these periodic reports, we also send a report when the epoch
+        // is coming to an end. Since `close_epoch` is called according to local clocks,
+        // we use an analogous rule for the last reports, requiring that the checkpoint
+        // is close to the next reconfiguration timestamp.
+        let should_send_last_report =
+            checkpoint_timestamp >= self.next_reconfiguration_timestamp_ms - 2000;
```
cyberphysic4l: Is there a risk here that we spam lots of misbehavior reports in the last few consensus rounds of the epoch? We should consider adding a "last_report_sent" to the scorer that we set to true once the last one is sent to avoid multiple "last" reports.

oliviasaa: fixed in [ead4c7c](https://github.com/iotaledger/iota/pull/10189/commits/ead4c7ce7c4cd3b5812824c098829d2838bb6b24)

crates/iota-core/src/checkpoints/checkpoint_output.rs
oliviasaa: added in [27077e4](https://github.com/iotaledger/iota/pull/10189/commits/27077e4cd1306c2033a70fa189c876669cab76b7)

Thoralf-M (approved):
Approving the changes in crates/iota-types/src/messages_consensus.rs

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,22 +126,58 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
+        // We also require that the checkpoint for which we want to send the report is
+        // at most 100 checkpoints behind the highest verified checkpoint, to avoid
+        // sending reports during resync.
+        //
+        // Additionally to these periodic reports, we also send a report when the epoch
+        // is coming to an end. Since `close_epoch` is called according to local clocks,
+        // we use an analogous rule for the last reports, requiring that the checkpoint
+        // is close to the next reconfiguration timestamp.
+        let should_send_last_report = checkpoint_timestamp
+            >= self.next_reconfiguration_timestamp_ms - REPORT_END_OF_EPOCH_MARGIN_MS
```
piotrm50: use `saturating_sub` as next_reconfiguration_timestamp_ms might sometimes be set to a smaller value than REPORT_END_OF_EPOCH_MARGIN_MS (e.g. in tests)

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -123,22 +126,58 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
+            && !epoch_store.scorer.has_sent_end_of_epoch_report();
+        if epoch_store.protocol_config().calculate_validator_scores()
+            && ((checkpoint_seq - epoch_store.scorer.last_report_checkpoint_seq()
```
piotrm50: might as well use saturating_sub here

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -4145,34 +4145,53 @@ impl AuthorityPerEpochStore {
...
+                    .should_accept_consensus_certs()
+                {
+                    let authority_index = self
+                        .committee
+                        .authority_index(authority)
+                        .expect("authority in committee");
+
+                    // Check validity of the report and update scores depending on the result. We
+                    // already have consensus on inclusion of this report in the DAG.
+                    match report.is_valid_version(self.protocol_config()) {
```
cyberphysic4l: Seems weird to match a bool rather than just an if/else

cyberphysic4l (changes requested):

cyberphysic4l (approved):
Accidental request changes, meant to approve :)

crates/iota-core/src/checkpoints/checkpoint_output.rs
oliviasaa: fixed in [3096822](https://github.com/iotaledger/iota/pull/10189/commits/3096822174b4bfa2eae4fe7c285feff0971b9b09)

crates/iota-core/src/checkpoints/checkpoint_output.rs
oliviasaa: fixed in [3096822](https://github.com/iotaledger/iota/pull/10189/commits/3096822174b4bfa2eae4fe7c285feff0971b9b09)

crates/iota-core/src/authority/authority_per_epoch_store.rs
oliviasaa: fixed in [3096822](https://github.com/iotaledger/iota/pull/10189/commits/3096822174b4bfa2eae4fe7c285feff0971b9b09)

semenov-vladyslav (approved):

piotrm50:
### Code review

Found 1 issue:

1. Warning says "report is ignored" when scoring is disabled, but there is no early return — processing continues unconditionally into the `should_accept_consensus_certs()` block, where the report is fully validated and stored via `update_received_reports()`. The log message directly contradicts the behavior.

https://github.com/iotaledger/iota/blob/edd6585f3330e3efe1ad3129263c6e4ed98302aa/crates/iota-core/src/authority/authority_per_epoch_store.rs#L4140-L4170

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

piotrm50 (approved):

PR #10219, opened 2026-02-11
chore(ts-sdk): port upstream changes added in PR 100

### LINKED ISSUE #10203: Port upstream PR 100 (Batch #2)
Port changes from upstream MystenLabs/ts-sdks for PR 100 in batch #2.

Upstream commit: https://github.com/MystenLabs/ts-sdks/commit/d76da2208d763fd42bf8bd9b92ae80c0cd1cc739 (PR 100)

Batch: 2
PR number: 100
Commit: d76da2208d763fd42bf8bd9b92ae80c0cd1cc739

brancoder:
# Description of change

resolves #10203

### DIFF
--- a/sdk/kiosk/src/query/kiosk.ts
+++ b/sdk/kiosk/src/query/kiosk.ts
@@ -3,10 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type {
+    PaginatedObjectsResponse,
     PaginationArguments,
     IotaClient,
     IotaObjectData,
-    IotaObjectDataFilter,
     IotaObjectResponse,
 } from '@iota/iota-sdk/client';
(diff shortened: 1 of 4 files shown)


sdk/kiosk/src/query/kiosk.ts
```diff
@@ -99,53 +103,129 @@ export async function getOwnedKiosks(
         personalKioskType: string;
     },
 ): Promise<OwnedKiosks> {
-    if (!isValidIotaAddress(address))
+    // TODO: this should throw an error, but I am not going to change it right now incase it breaks existing code.
```
msarcev, line 106: Should we really copy comments like this too ? 🤔

brancoder: I left it in because it makes sense, we should throw an error for invalid address instead of returning empty.

PR #10229, opened 2026-02-11
feat(gRPC): non-exhaustive types

muXxer:
# Description of change

This updates all gRPC proto types to have non exhaustive message types and enum types, and generates and uses accessors because we need to use the builder pattern with non-exhaustive types.


crates/iota-grpc-server/src/merge.rs
```diff
@@ -295,29 +298,25 @@ impl Merge<&iota_sdk_types::Event> for Event {
...
         if mask.contains(Self::EVENT_TYPE_FIELD.name) {
-            self.event_type = Some(source.type_.to_string());
+            self.set_event_type(source.type_.to_string());
         }
 
         if mask.contains(Self::BCS_CONTENTS_FIELD.name) {
-            self.bcs_contents = Some(BcsData {
-                data: source.contents.clone().into(),
-            });
+            self.set_bcs(BcsData::default().with_data(source.contents.clone()));
```
bingyanglin: Should be `set_bcs_contents`?

crates/iota-e2e-tests/tests/grpc/v0/transaction_execution_service/simulate.rs
```diff
@@ -68,19 +71,12 @@ async fn simulate_transaction_with_gas_estimation() {
...
+        .with_bcs(BcsData::default().with_data(bcs::to_bytes(&tx_data).unwrap()));
 
-    let request = SimulateTransactionRequest {
-        transaction: Some(transaction),
-        tx_checks: vec![],
-        estimate_gas_budget: Some(true),
-        read_mask: None,
-    };
+    let request = SimulateTransactionRequest::default()
+        .with_transaction(transaction)
```
bingyanglin, line 78: Do we need `.with_estimate_gas_budget(true)` here?

crates/iota-e2e-tests/tests/grpc/v0/transaction_execution_service/header.rs
```diff
@@ -93,19 +87,12 @@ async fn test_response_headers() {
...
-            transaction: Some(transaction),
-            tx_checks: vec![],
-            estimate_gas_budget: Some(true),
-            read_mask: None,
-        };
+        let transaction = ProtoTransaction::default()
+            .with_bcs(BcsData::default().with_data(bcs::to_bytes(&tx_data).unwrap()));
+
+        let request = SimulateTransactionRequest::default()
+            .with_transaction(transaction)
```
bingyanglin, line 94: Do we need `.with_estimate_gas_budget(true)` here?

crates/iota-grpc-types/src/proto/generated/iota.grpc.v0.filter.rs
```diff
@@ -85,36 +93,42 @@ pub mod event_filter {
...
 pub struct NotTransactionFilter {
     #[prost(message, optional, boxed, tag = "1")]
     pub filter: ::core::option::Option<::prost::alloc::boxed::Box<TransactionFilter>>,
 }
 /// Filter by transaction kinds (any of the specified kinds).
+#[non_exhaustive]
 #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
 pub struct TransactionKindsFilter {
-    #[prost(enumeration = "TransactionKind", repeated, tag = "1")]
+    #[prost(enumeration = "TransactionKind", repeated, packed = "false", tag = "1")]
```
bingyanglin: The proto docs says it should be packed, (i.e., packed = "true") by default, see the section **Repeated Elements** in https://protobuf.dev/programming-guides/encoding/

The generated `packed = "false"` here means we use per-element encoding, but in the `.proto` file we didn't specify it. Somehow this is an old bug (https://github.com/tokio-rs/prost/issues/976) and I just created PR for it (https://github.com/tokio-rs/prost/pull/1399).

crates/iota-e2e-tests/tests/grpc/v0/transaction_execution_service/simulate.rs
muXxer: will be fixed in a follow up PR, I didn't want to mix that fix of the missing "estimate_gas_budget" flag in the "simulate_transaction" function signature with this big PR that changes just the exhaustive types.

crates/iota-e2e-tests/tests/grpc/v0/transaction_execution_service/header.rs
muXxer: will be fixed in a follow up PR, I didn't want to mix that fix of the missing "estimate_gas_budget" flag in the "simulate_transaction" function signature with this big PR that changes just the exhaustive types.

crates/iota-grpc-types/src/proto/generated/iota.grpc.v0.filter.rs
muXxer: good catch, I was already wondering why this flag changes. We need to see what happens with your PR before we ship the gRPC interface, or do the extra fields in a different way, if that is caused by my extra flags?

bingyanglin: For this flag It should be caused by the proto bug of the repeated fields only, but I agree we need to carefully review the generated protos when we ship it.

bingyanglin (approved):

PR #10245, opened 2026-02-12
feat(wallet): add side panel changed amplitude event

### LINKED ISSUE #10194: Add side panel changed amplitude event
Add side panel changed amplitude event
Fires when the user updates the side panel setting

Props
- `enabled`: `boolean`

evavirseda:
# Description of change

disabling the side panel closes the window immediately, so any code after the mutation never runs :see_no_evil:  
the only solution i found was to track and flush the event before the mutation when disabling, and after when enabling (where the window is still alive)

`fixes https://github.com/iotaledger/iota/issues/10194


Juligs (approved):

apps/wallet/src/ui/app/components/menu/content/WalletSettingsMenuList.tsx
```diff
@@ -101,8 +101,14 @@ export function MenuList() {
     }
 
     function onSidePanelClick() {
-        sidePanelMutation.mutateAsync(!sidePanel.data).then(() => {
-            if (!sidePanel.data) {
+        if (sidePanel.data === undefined) return;
```
msarcev: Why adding the early return if `undefined` ?
This seams to change the logic of the function.

msarcev (approved):

panteleymonchuk (changes requested):
plese, check these 2 cases:
- when we click "enable side panel" event fires 2 times
- when we click "disable side panel" event doesn't fire at all

panteleymonchuk (approved):

PR #10258, opened 2026-02-12
feat(explorer): enhance OwnedObjects component with new filters 

### LINKED ISSUE #5394: Proposal for a New "Data Objects" Section in the IOTA Explorer
Currently, the IOTA Explorer features an **ASSETS** section divided into two subsections: **NFTs** and **KIOSKS**.  

I propose adding a third subsection to list all objects that do not conform to either the NFT or KIOSK standards.  

While NFTs can be based on different standards, there are certainly **Move Objects** that cannot be classified as NFTs. From an implementation perspective, this distinction can be achieved either by introducing a specific standard or more simply by identifying objects that lack fundamental NFT properties—such as an external content URL.  

All **Move Objects** that are neither **Kiosks** nor **NFTs** should be grouped under a new section labeled **"Data Objects."**  

### Why is this distinction important?  

Applications built on Move Objects do not always require tokens. Even when tokens are used, they may come with significant restrictions tailored to specific business needs.

For example, in a prepaid service model, a provider might issue "service credit" tokens to customers. In this case, the provider wants to ensure that the credit assigned to a specific customer cannot be transferred to third parties, preventing speculation or unintended resale. This is particularly relevant in industries where prepaid services are closely tied to contractual agreements, such as cloud computing, telecommunications, or subscription-based platforms.

To enforce these constraints, the service credit token could be designed as non-transferable, meaning it can only be spent on predefined services or burned upon use. This approach ensures the token functions as a consumption-based asset rather than a tradable financial instrument. By doing so, businesses can maintain pricing stability and regulatory compliance while preventing the formation of a secondary market that could disrupt their business model.

These characteristics—or more broadly, the lack of features that typically define an asset as a token—carry regulatory implications under the MiCA framework.

MiCA imposes compliance requirements on companies using tokens within their services and applications. If a token can be transferred or traded, even indirectly, it may fall within MiCA’s regulatory scope. However, if a service credit token is strictly non-transferable and cannot be used for speculation, it is unlikely to be classified as a regulated financial instrument.

Despite this, displaying such service credit tokens as NFTs in the explorer creates confusion and inconsistency. Users and regulators might wrongly assume they belong to the same category as transferable digital assets, leading to misunderstandings about their regulatory treatment and intended functionality. Properly distinguishing these assets in the explorer would remove ambiguity, ensuring clarity for both developers and businesses.

Unlike traditional blockchain tokens, which are inherently transferable, tradable, and wrappable, a Move Object does not necessarily share these properties. More importantly, Move Objects lacking the methods required for transfer, trading, or wrapping should not be classified as tokens. Ensuring a clear separation in the explorer would help prevent misclassification and reduce regulatory uncertainty for applications that fall outside the scope of financial regulations.

### Benefits of a Clear Distinction in the Explorer  

Explicitly **differentiating Move Objects from tokens** and **providing a dedicated visualization** in the explorer would offer significant benefits to developers and businesses. It would:  

- **Streamline development** for applications that are unrelated to MiCA’s regulatory scope.  
- **Prevent misclassification** of Move Objects as financial instruments.  
- **Reduce unnecessary regulatory risks** for businesses using Move Objects in non-financial contexts.  

By implementing this separation, the IOTA ecosystem would foster a **more transparent and compliant** environment for both developers and enterprises.

Juligs:
fixes #5394


apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -131,23 +140,91 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
     useEffect(() => {
-        if (!isPending) {
-            setFilter(
-                kioskData?.list?.length && filter === FilterValue.Kiosks
-                    ? FilterValue.Kiosks
-                    : FilterValue.All,
-            );
+        if (filter === FilterValue.Other && viewMode !== ObjectViewMode.List) {
+            setViewMode(ObjectViewMode.List);
         }
```
brancoder: maybe we could simplify the filtering logic a bit, can we remove this useEffect to switch view mode to list for `OTHER` since we have the same logic calculated in `effectiveViewMode`?

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -187,14 +264,18 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
-                );
+                nextHeight = MIN_OBJECT_COUNT_TO_HEIGHT_MAP[Number(minObjectCount)];
             }
         });
-    }, [sortedDataByDisplayImages, setOwnedObjectsContainerHeight]);
+
+        if (nextHeight !== ownedObjectsContainerHeight) {
+            setOwnedObjectsContainerHeight(nextHeight);
+        }
+    }, [sortedDataByDisplayImages.length, ownedObjectsContainerHeight]);
```
brancoder: can we use `useMemo` here since this is a calculation from `sortedDataByDisplayImages.length`?
also we are adding `ownedObjectsContainerHeight` in dependency array but also setting it inside the `useEffect` causing rerenders?

brancoder (changes requested):
LG and works 👍  just a few questions

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -218,12 +287,12 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
+                    <div className="flex w-full flex-col flex-wrap items-start justify-between gap-xs sm:min-h-[72px] sm:flex-row sm:items-center md:gap-0">
                         <Title size={TitleSize.Medium} title="Assets" />
-                        {hasVisualAssets && (
-                            <div className="flex justify-between px-md--rs sm:flex-row">
+                        {hasVisualAssets && availableFilters.length > 0 && (
+                            <div className="flex flex-col gap-sm px-md--rs sm:flex-row sm:gap-0">
                                 <div className="flex items-center gap-sm">
-                                    {VIEW_MODES.map((mode) => {
+                                    {availableViewModes.map((mode) => {
                                         const selected = mode.value === viewMode;
```
brancoder, line 223: we should change this to `const selected = mode.value === effectiveViewMode;` because it wont work correctly for filter `OTHER`, right?

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -34,6 +36,8 @@ import { ListView, NoObjectsOwnedMessage, SmallThumbnailsView, ThumbnailsView }
 import { ObjectViewMode } from '~/lib/enums';
 import { Pagination } from '~/components/ui';
 import { PAGE_SIZES_RANGE_10_50 } from '~/lib/constants';
+import { getNameRegistrationType, getSubnameRegistrationType } from '@iota/iota-names-sdk';
+import type { IotaObjectResponse } from '@iota/iota-sdk/src/client';
```
brancoder: we can remove the `/src` from the import

brancoder (changes requested):
LG just minor details

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -55,8 +61,10 @@ enum OwnedObjectsContainerHeight {
 }
 
 const FILTER_OPTIONS = [
-    { label: 'NFTS', value: FilterValue.All },
-    { label: 'KIOSKS', value: FilterValue.Kiosks },
+    { label: 'NFT', value: FilterValue.Nft },
+    { label: 'NAME', value: FilterValue.Name },
+    { label: 'KIOSK', value: FilterValue.Kiosk },
+    { label: 'OTHER', value: FilterValue.Other },
```
evavirseda:
```suggestion
    { label: 'NFTS', value: FilterValue.Nft },
    { label: 'NAMES', value: FilterValue.Name },
    { label: 'KIOSKS', value: FilterValue.Kiosk },
    { label: 'OTHER', value: FilterValue.Other },
```

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
evavirseda:
```suggestion
import type { IotaObjectResponse } from '@iota/iota-sdk/client';
```

evavirseda (dismissed):
🙏

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
Juligs: I spoke with @begonaalvarezd , and we think it would be better to keep the filter labels in the singular form instead of plural

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
Juligs: I think it should still work correctly with `viewMode`, since only List is available for OTHER. Switching to `effectiveViewMode` introduces a small visual flicker where the selected state briefly changes -->
[Screencast from 2026-02-17 13-42-08.webm](https://github.com/user-attachments/assets/79b4b365-bc8a-4385-ac76-2bd714f07fa2)

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -131,28 +137,89 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
+            case FilterValue.Name:
+                return categorizedObjects.name;
+            case FilterValue.Nft:
+                return categorizedObjects.nft;
+            case FilterValue.Other:
+                return categorizedObjects.other;
+            default:
+                return [];
+        }
+    }, [filter, data?.data, categorizedObjects]);
```
marc2332: Unnecessary useMemo
```suggestion
    const filteredData = (() => {
        if (!data?.data && filter !== FilterValue.Kiosk) return [];

        switch (filter) {
            case FilterValue.Kiosk:
                return categorizedObjects.kiosk;
            case FilterValue.Name:
                return categorizedObjects.name;
            case FilterValue.Nft:
                return categorizedObjects.nft;
            case FilterValue.Other:
                return categorizedObjects.other;
            default:
                return [];
        }
    })();
```

marc2332 (dismissed):

KeitoTadashi (dismissed):
1. we have a weird behaviour counting and paginating elements:
https://github.com/user-attachments/assets/c0dea5d8-4207-483b-a85d-3e1583231fac

2. also now we are showing a big padding/margin at bottom (under pagination)
Image

3. pagination disappear when:
step 1: group by NFT
step 2: click next page
step 3: group by name
step 4: click prev page -> pagination disappear
https://github.com/user-attachments/assets/ff095128-9f49-4d30-88f0-02c544157625

apps/core/src/hooks/useGetCategorizedOwnedObjects.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+
+    const {
+        data: kioskData,
+        isFetching: kioskFetching,
+        isError: kioskError,
+    } = useGetKioskContents(address);
+
+    const { iotaNamesClient } = useIotaNamesClient();
+    const packageId = iotaNamesClient?.getPackage('packageId', 'v1');
```
VmMad: can you add a try/catch in this line? currently if you setup a localnet it will throw an error an no assets will be displayed

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -312,15 +311,17 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
-                                        Showing {start} - {end}
-                                    </span>
-                                )}
+                    <div className="flex flex-col items-center justify-between gap-sm px-sm--rs py-sm--rs md:flex-row">
+                        {showPagination && hasVisualAssets && activeCategoryData && (
+                            <Pagination {...activeCategoryData.pagination} />
+                        )}
+                        <div className="flex items-center gap-sm">
+                            {!isPending && showPagination && hasVisualAssets && (
+                                <span className="shrink-0·text-body-sm·text-iota-neutral-40·dark:text-iota-neutral-6">
```
evavirseda:
```suggestion
                                <span className="shrink-0 text-body-sm text-iota-neutral-40 dark:text-iota-neutral-6">
```

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -257,28 +253,30 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
+                                            <ButtonSegment
+                                                key={value}
+                                                type={ButtonSegmentType.Rounded}
+                                                selected={value === filter}
+                                                label={option?.label ?? value.toUpperCase()}
+                                                disabled={isPending}
+                                                onClick={() => setFilter(value)}
+                                            />
+                                        );
+                                    })}
```
evavirseda: i think we could simplify this 🙏 
```suggestion
                                    {availableCategories.map((value) => (
                                        <ButtonSegment
                                            key={value}
                                            type={ButtonSegmentType.Rounded}
                                            selected={value === filter}
                                            label={CATEGORY_LABELS[value]}
                                            disabled={isPending}
                                            onClick={() => setFilter(value)}
                                        />
                                    ))}
```

and define `CATEGORY_LABELS` like:
```
    const CATEGORY_LABELS: Record<OwnedObjectCategory, string> = {
        [OwnedObjectCategory.Nft]: 'NFT',
        [OwnedObjectCategory.Name]: 'NAME',
        [OwnedObjectCategory.Kiosk]: 'KIOSK',
        [OwnedObjectCategory.Other]: 'OTHER',
    };
```

apps/core/src/hooks/useGetCategorizedOwnedObjects.ts
```diff
@@ -0,0 +1,202 @@
...
+
+function useVirtualPagination<T>(
+    items: T[],
+    pageSize: number,
+): CategoryData<T> & { setPage: (page: number) => void } {
+    const [currentPage, setCurrentPage] = useState(0);
+
+    useEffect(() => {
+        setCurrentPage(0);
+    }, [pageSize]);
```
evavirseda: i think `useVirtualPagination` only resets to page 0 when `pageSize` changes, so if a user is on page 3 of `NFTs` and switches to `Kiosk`, the Kiosk pagination will start on page 3 as well, which may render an empty page if there are fewer items 🙈 

maybe we can add a `resetKey` parameter to reset the page when the active category changes?

```suggestion
function useVirtualPagination<T>(
    items: T[],
    pageSize: number,
    resetKey?: string,
): CategoryData<T> & { setPage: (page: number) => void } {
    const [currentPage, setCurrentPage] = useState(0);

    useEffect(() => {
        setCurrentPage(0);
    }, [pageSize, resetKey]);
```

apps/core/src/hooks/useGetCategorizedOwnedObjects.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        return { nftItems: nft, nameItems: name, otherItems: other };
+    }, [allFetchedObjects, nameTypes]);
+
+    const kioskItems = useMemo(() => kioskData?.list ?? [], [kioskData?.list]);
+
+    const nftPagination = useVirtualPagination(nftItems, pageSize);
+    const namePagination = useVirtualPagination(nameItems, pageSize);
+    const otherPagination = useVirtualPagination(otherItems, pageSize);
+    const kioskPagination = useVirtualPagination(kioskItems, pageSize);
```
evavirseda: and update this 🙏 
```suggestion
    const nftPagination   = useVirtualPagination(nftItems,   pageSize, OwnedObjectCategory.Nft);
    const namePagination  = useVirtualPagination(nameItems,  pageSize, OwnedObjectCategory.Name);
    const otherPagination = useVirtualPagination(otherItems, pageSize, OwnedObjectCategory.Other);
    const kioskPagination = useVirtualPagination(kioskItems, pageSize, OwnedObjectCategory.Kiosk);
```

evavirseda (changes requested):
🙏

evavirseda (approved):

VmMad (approved):

KeitoTadashi (approved):

apps/explorer/src/components/owned-objects/OwnedObjects.tsx
```diff
@@ -312,15 +303,17 @@ export function OwnedObjects({ id }: OwnedObjectsProps): JSX.Element {
...
-                                        Showing {start} - {end}
-                                    </span>
-                                )}
+                    <div className="flex flex-col items-center justify-between gap-sm px-sm--rs py-sm--rs md:flex-row">
+                        {showPagination && hasVisualAssets && activeCategoryData && (
+                            <Pagination {...activeCategoryData.pagination} />
+                        )}
+                        <div className="flex items-center gap-sm">
+                            {!isPending && showPagination && hasVisualAssets && (
+                                <span className="shrink-0 text-body-sm text-iota-neutral-40 dark:text-iota-neutral-6">
```
brancoder: should this be `dark:text-iota-neutral-60`

apps/core/src/hooks/useGetCategorizedOwnedObjects.ts
```diff
@@ -0,0 +1,202 @@
...
+    const availableCategories = useMemo(() => {
+        const category: OwnedObjectCategory[] = [];
+        if (nftItems.length > 0) category.push(OwnedObjectCategory.Nft);
+        if (nameItems.length > 0) category.push(OwnedObjectCategory.Name);
+        if (kioskItems.length > 0) category.push(OwnedObjectCategory.Kiosk);
+        if (otherItems.length > 0) category.push(OwnedObjectCategory.Other);
+        return category;
+    }, [nftItems.length, nameItems.length, kioskItems.length, otherItems.length]);
+
+    const isFetchingAll = ownedObjectsQuery.isFetching || hasMore;
```
brancoder: can we remove `hasMore`? 
can there be a case where `isFetchingAll = true` because `hasMore = true` but `ownedObjectsQuery.isFetching = false` ? so basically it stays flagged as loading eventhough it only has more but is not fetching?

brancoder (commented):
just couple of questions

brancoder (approved):

PR #10268, opened 2026-02-13
feat(wallet): add element copied amplitude event

### LINKED ISSUE #10174: Add element copied amplitude event
Add element copied amplitude event. 
Event should fire when a user copies anything to the clipboard. 
The element copied can be private or public, and it should be private by default. When is private, the value cant be sent to amplitude

Props:
- `value`: string | null
- `type`: `address | digest | … it can grow`

For now all element copied events should be private, so no values should be sent to amplitude

evavirseda:
# Description of change

fixes https://github.com/iotaledger/iota/issues/10174


VmMad (approved):

Juligs (approved):
Lgtm  😻

panteleymonchuk:
Just wanted to say that we can create generic plugin and pass params, as logic pretty same.
But seems it was merged before I commented it, so I'll leave it here:

```
export function privacyPlugin(eventType: string, privateTypes: Set<string>): EnrichmentPlugin {
    return {
        name: 'privacy-plugin',
        type: 'enrichment',

        async execute(event: Event) {
            if (!event.event_type?.endsWith(eventType)) {
                return event;
            }

            let props = { ...(event.event_properties ?? {}) } as Record<string, unknown>;

            const type =
                typeof props.type === 'string' && props.type.trim() ? props.type : 'unknown';

            let visibility: 'private' | 'public' =
                props.visibility === 'public' ? 'public' : 'private';

            if (privateTypes.has(type)) {
                visibility = 'private';
            }

            props.type = type;
            props.visibility = visibility;

            if (visibility === 'private') {
                const { value, ...rest } = props;
                props = rest;
            }

            return {
                ...event,
                event_properties: { ...props },
            };
        },
    };
}
```

PR #10273, opened 2026-02-13
feat(gRPC): add versioned BCS envelopes for types without native discriminants

### LINKED ISSUE #10264: gRPC: add versioned BCS envelopes for types without native discriminants
Our gRPC API uses a two-layer versioning approach:

- **BCS-level:** Types with enum versioning (`Transaction`, `TransactionEffects`, `CheckpointContents`, `UserSignature`) embed a leading discriminant byte (`0x00` = V1, `0x01` = V2, etc.) — supports up to 256 variants per type. Clients deserialize transparently via SDK `#[non_exhaustive]` enums. No proto changes needed when new variants are added.
- **API-level:** Breaking API changes (new RPCs, changed semantics) are handled by proto package versioning (`iota.grpc.v0` → `iota.grpc.v1`).

We explicitly **do not** use proto `oneof` to mirror BCS variants because:
- It duplicates the BCS discriminant already present in the bytes
- Every new BCS version would require proto file updates + code regeneration
- It adds server-side complexity (peeking at discriminant byte to select oneof variant)

### Current issue

4 types are plain structs with **no BCS discriminant byte**:

| Type | Structure |
|---|---|
| `Object` | Struct (inner `Data` is stable Move/Package enum, not a version enum) |
| `Event` | Plain struct |
| `CheckpointSummary` | Plain struct (has `version_specific_data: Vec<u8>` for inner extensibility, but the struct itself is unversioned) |
| `ValidatorAggregatedSignature` | Plain struct |

If the layout of any of these structs changes, clients have **no way to detect the mismatch**: they get silent corruption or opaque deserialization errors with no version signal.

### Possible Solution

Add version metadata alongside the BCS blob for these unversioned types so the server can communicate which schema version the bytes represent, allowing clients to select the correct deserializer or fail with a clear version mismatch error.

bingyanglin:
# Description of change

Add versioned BCS wrappers (V1 envelope) for `Object`, `Event`, `CheckpointSummary`, and `ValidatorAggregatedSignature` plain structs that lack native BCS discriminants

fixes #10264  

## Change checklist

- [x] I have followed the [contribution guidelines](../../CONTRIBUTING.md) for this project
- [x] I have made corresponding changes to the documentation


crates/iota-grpc-types/src/versioned.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// BCS format: `versioned-object = %x00 object`
+#[derive(Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum VersionedObject {
+    V1(iota_sdk_types::Object),
+}
+
+impl VersionedObject {
+    pub fn into_inner(self) -> iota_sdk_types::Object {
```
semenov-vladyslav: If a new variant is added, eg. `V2(iota_sdk_types::ObjectV2)`, then what type should this function return?

semenov-vladyslav: Maybe there should be instead `fn try_v1(&self) -> Option<&iota_sdk_types::Object>` and `fn try_into_v1(self) -> Result<iota_sdk_types::Object, Self>`?

crates/iota-grpc-types/src/proto/iota/grpc/v0/transaction.rs
```diff
@@ -80,10 +81,12 @@ impl TryFrom<&TransactionEvents> for iota_sdk_types::TransactionEvents {
                     TryFromProtoError::missing("event.bcs")
                         .nested_at(TransactionEvents::EVENTS_FIELD.name, i)
                 })?;
-                bcs.deserialize().map_err(|err| {
-                    TryFromProtoError::invalid("event.bcs", err)
-                        .nested_at(TransactionEvents::EVENTS_FIELD.name, i)
-                })
+                bcs.deserialize::<VersionedEvent>()
+                    .map(VersionedEvent::into_inner)
```
semenov-vladyslav: This conversion can be fallible: `VersionedEvent` will not always have `V1` variant.

semenov-vladyslav (changes requested):

crates/iota-grpc-types/src/versioned.rs
bingyanglin: Addressed in [ef4265d](https://github.com/iotaledger/iota/pull/10273/commits/ef4265dd7da23e66d45463b3a38e2404298b649d)

crates/iota-grpc-types/src/proto/iota/grpc/v0/transaction.rs
bingyanglin: Addressed in [ef4265d](https://github.com/iotaledger/iota/pull/10273/commits/ef4265dd7da23e66d45463b3a38e2404298b649d).
Note that the compiler complains the following so we need to box it.
```bash
error: the `Err`-variant returned from this function is very large
  --> crates/iota-grpc-types/src/versioned.rs:64:33
   |
54 |     V1(iota_sdk_types::CheckpointSummary),
   |     ------------------------------------- the largest variant contains at least 256 bytes
...
64 |     pub fn try_into_v1(self) -> Result<iota_sdk_types::CheckpointSummary, Self> {
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

crates/iota-grpc-server/src/merge.rs
```diff
@@ -294,7 +295,7 @@ impl Merge<&iota_sdk_types::Event> for Event {
         mask: &FieldMaskTree,
     ) -> Result<(), Box<dyn std::error::Error>> {
         if mask.contains(Self::BCS_FIELD.name) {
-            self.bcs = Some(BcsData::serialize(&source)?);
+            self.bcs = Some(BcsData::serialize(&VersionedEvent::V1(source.clone()))?);
```
semenov-vladyslav, line 298:
```suggestion
            self.bcs = Some(BcsData::serialize(&V1(source))?);
```

semenov-vladyslav (changes requested):

crates/iota-grpc-types/src/proto/iota/grpc/v0/transaction.rs
semenov-vladyslav: It's probably a temporary solution that should be addressed properly.

crates/iota-grpc-server/src/merge.rs
semenov-vladyslav: `V1` was removed. Using `VersionedXXX` types is clearer.

semenov-vladyslav (approved):

muXxer (approved):

PR #10276, opened 2026-02-13
fix(iota-rest-kv): fix transaction events deserialization

### LINKED ISSUE #10269: [iota-rest-kv]: flatten optional `TransactionEvents`
Currently the returned result from the `iota-rest-kv` when requesting events is to return `Option<Bytes>` the bytes are represented as `Option<TransactionEvents>`, the server side shall represent the `Bytes` as `TransactionEvents`.

sergiupopescu199:
# Description of change

This patch fixes the deserialization issue on the `iota-rest-kv` for the TransactionEvents and also fixes a runtime issue encountered in the newest version 1.17.0 of missing process-level CryptoProvider.

fixes #10269 

- started locally the `iota-rest-kv` server connected to the actual testnet BigTable database
- executed the same test Bing provided in the [commit](https://github.com/bingyanglin/iota/commit/c805cd6e9af5f0444cb577cbfb7f75f0dcd324d5#diff-0a2e6fd5dcd4514822cc6ff7d0932710d95af9fc0203d958928460e6226a6fabR358)

### DIFF
--- a/crates/iota-rest-kv/src/bigtable.rs
+++ b/crates/iota-rest-kv/src/bigtable.rs
@@ -22,7 +22,7 @@ use iota_kvstore::{
     proto::bigtable::v2::{RowFilter, row_filter::Filter},
 };
 use iota_storage::http_key_value_store::Key;
-use iota_types::storage::ObjectKey;
+use iota_types::{effects::TransactionEvents, storage::ObjectKey};
 use serde::{Deserialize, Serialize};
 use tracing::error;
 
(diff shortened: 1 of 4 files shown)


crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -235,13 +235,23 @@ impl KvStoreClient {
...
+
+                Ok(response
+                    .into_iter()
+                    .map(|value| {
+                        value.map(|bytes|
+                        // we strip the first byte which represents the Option<T>, returning only
+                        // the bytes which represent the contained value in the Option<T>
+                        bytes.slice(1..))
+                    })
+                    .collect())
```
bingyanglin:
```suggestion
                Ok(response
                    .into_iter()
                    .map(|value| {
                        value.and_then(|bytes| {
                            // The events column stores BCS-encoded Option<TransactionEvents>.
                            // 0x01 = Some(T): strip the tag and return T's bytes.
                            // 0x00 = None: no events for this transaction.
                            if bytes.first() == Some(&1) {
                                Some(bytes.slice(1..))
                            } else {
                                None
                            }
                        })
                    })
                    .collect())
```

We need to use `and_then` instead of `map` to propagate `None`? Also if the value is empty it will panic so we need `and_then` I guess?

sergiupopescu199: added a more robust check: https://github.com/iotaledger/iota/pull/10276/commits/2839727250ce3fecf8a1663c18a986d960457374

bingyanglin (approved):

kodemartin (approved):
lgtm

PR #10288, opened 2026-02-16
feat(consensus, iota-types): Remove VersionedScoringMetrics type

oliviasaa:
# Description of change

Removes the `VersionedStorageScoringMetrics` type by reusing `VersionedScoringMetrics`:
  - Before this PR, each authority's scoring metrics were stored as a separate entry keyed by `AuthorityIndex`, using `VersionedStorageScoringMetrics` (wrapping of `MisbehaviorsV1<u64>`). Now, all authorities' metrics are stored as a single blob using `VersionedScoringMetrics` directly (wrapping of `MisbehaviorsV1<Vec<AtomicU64>>`). For that, custom `Serialize/Deserialize` impls were added to `VersionedScoringMetrics`.
- Simplified eviction code: `update_scoring_metrics_on_eviction()` used to return an update per authority; the caller collected these into a `Vec` for writing. Now, the method only updates state, and a single `snapshot()` call after all evictions captures the full metrics blob for persistence.
- Simplified initialization/recovery: Before this PR, `initialize_scoring_metrics()` used a vector of `VersionedStorageScoringMetrics` that was not guaranteed to have entries for all authorities, so it had to "manually" fill in missing authorities. Now, it uses `Option<VersionedScoringMetrics>`, so either the full blob exists or it doesn't. 
- Since now the full blob is persisted after eviction (instead of only the metrics for some authorities), the migration in RocksDB is now lazy (migrated blob returned directly, v2 CF written on the first flush) instead of forced during `scan_scoring_metrics()`.

### DIFF
--- a/consensus/core/src/scoring_metrics_store.rs
+++ b/consensus/core/src/scoring_metrics_store.rs
@@ -38,7 +38,7 @@ impl MysticetiScoringMetricsStore {
     // recovered_scoring_metrics and blocks_in_cache_by_authority.
     pub(crate) fn initialize_scoring_metrics(
         &self,
-        recovered_scoring_metrics: Vec<(AuthorityIndex, VersionedStorageScoringMetrics)>,
+        recovered_scoring_metrics: Option<VersionedStorageScoringMetrics>,
         blocks_in_cache_by_authority: &Vec<BTreeSet<BlockRef>>,
         threshold_clock_round: u32,
         eviction_rounds: &Vec<u32>,
@@ -46,9 +46,9 @@ impl MysticetiScoringMetricsStore {
     ) {
         // At the beginning of the epoch (i.e., before the epoch's first eviction),
         // there is nothing to be recovered from storage, so recovered_scoring_metrics
-        // will be empty. In this case, the initialization functions will simply
+        // will be None. In this case, the initialization functions will simply
         // initialize zeroed metrics for all authorities, so we can return here.
-        if recovered_scoring_metrics.is_empty() {
+        if recovered_scoring_metrics.is_none() {
             return;
         }
 
@@ -56,7 +56,7 @@ impl MysticetiScoringMetricsStore {
         // them.
         match &context.protocol_config.scorer_version_as_option() {
             None | Some(1) => self.initialize_scoring_metrics_v1(
-                recovered_scoring_metrics,
+                recovered_scoring_metrics.unwrap(),
                 blocks_in_cache_by_authority,
                 threshold_clock_round,
                 eviction_rounds,
@@ -70,37 +70,36 @@ impl MysticetiScoringMetricsStore {
     // recovered_scoring_metrics and blocks_in_cache_by_authority.
     pub(crate) fn initialize_scoring_metrics_v1(
         &self,
-        recovered_scoring_metrics: Vec<(AuthorityIndex, VersionedStorageScoringMetrics)>,
+        recovered_scoring_metrics: VersionedStorageScoringMetrics,
         blocks_in_cache_by_authority: &Vec<BTreeSet<BlockRef>>,
         threshold_clock_round: u32,
         eviction_rounds: &Vec<u32>,
         context: &Arc<Context>,
     ) {
-        let hostnames = context
+        let indices_and_hostnames = context
             .committee
             .authorities()
-            .map(|(_, x)| x.hostname.as_str())
+            .map(|(i, x)| (i, x.hostname.as_str()))
             .collect::<Vec<_>>();
+        let VersionedStorageScoringMetrics::V1(inner) = recovered_scoring_metrics;
 
-        for ((authority_index, metrics), hostname, blocks_in_cache, &eviction_round) in izip!(
-            recovered_scoring_metrics,
-            hostnames,
+        for ((authority_index, hostname), blocks_in_cache, &eviction_round) in izip!(
+            indices_and_hostnames,
             blocks_in_cache_by_authority,
             eviction_rounds
         ) {
             // Initialize the uncached scoring metrics according to
             // recovered_scoring_metrics
-            let VersionedStorageScoringMetrics::V1(inner) = &metrics;
             self.initialize_faulty_blocks_metrics(
-                *inner.faulty_blocks_provable(),
-                *inner.faulty_blocks_unprovable(),
+                inner.faulty_blocks_provable()[authority_index.value() as usize],
+                inner.faulty_blocks_unprovable()[authority_index.value() as usize],
                 hostname,
                 authority_index,
                 &context.metrics.node_metrics,
             );
             self.update_missing_blocks_and_equivocations(
-                *inner.missing_proposals(),
-                *inner.equivocations(),
+                inner.missing_proposals()[authority_index.value() as usize],
+                inner.equivocations()[authority_index.value() as usize],
                 hostname,
                 authority_index,
                 StoreType::Uncached,
@@ -286,14 +285,11 @@ impl MysticetiScoringMetricsStore {
         last_eviction_round: u32,
         threshold_clock_round: u32,
         context: &Arc<Context>,
-    ) -> VersionedStorageScoringMetrics {
+    ) {
         let node_metrics = &context.metrics.node_metrics;
         // threshold_clock_round should be always at least 1.
         if threshold_clock_round == 0 {
-            return VersionedStorageScoringMetrics::new_from(
-                &self.uncached_metrics,
-                authority_index.value(),
-            );
+            return;
         }
 
         // Get the blocks rounds that were not evicted.
@@ -323,10 +319,7 @@ impl MysticetiScoringMetricsStore {
         // If no eviction happened, we do not update the current local metrics counts
         // and the metrics on storage.
         if eviction_round == last_eviction_round {
-            return VersionedStorageScoringMetrics::new_from(
-                &self.uncached_metrics,
-                authority_index.value(),
-            );
+            return;
         }
 
         // Get the evicted blocks rounds.
@@ -360,8 +353,6 @@ impl MysticetiScoringMetricsStore {
         // since the last eviction (e.g., faulty blocks metrics updated on block
         // receival).
         self.update_current_local_metrics_count(authority_index);
-
-        VersionedStorageScoringMetrics::new_from(&self.uncached_metrics, authority_index.value())
     }
 
     // The `authority_index` should be a valid index, otherwise the function will
@@ -537,7 +528,6 @@ mod tests {
     use std::{collections::BTreeSet, sync::Arc, vec};
 
     use consensus_config::{AuthorityIndex, NetworkKeyPair, ProtocolKeyPair};
-    use iota_common::scoring_metrics::VersionedStorageScoringMetrics;
     use parking_lot::RwLock;
     use tokio::sync::broadcast;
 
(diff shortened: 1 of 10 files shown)


consensus/core/src/scoring_metrics_store.rs
```diff
@@ -86,37 +70,48 @@ impl MysticetiScoringMetricsStore {
...
+        let equivocations = recovered_scoring_metrics.load_equivocations();
+
+        for (
+            (authority_index, authority),
+            blocks_in_cache,
+            &eviction_round,
+            &fbp,
+            &fbu,
+            &mp,
+            &eq,
```
Thoralf-M: Probably better to keep them written out, otherwise much harder to follow if one is not familiar with the code

consensus/core/src/scoring_metrics_store.rs
```diff
@@ -301,13 +294,11 @@ impl MysticetiScoringMetricsStore {
...
-        let committee_size = context.committee.size();
+    ) {
         let node_metrics = &context.metrics.node_metrics;
-        // threshold_clock_round should be always at least 1.  Analogously,
-        // authority_index should be a valid index.
-        if threshold_clock_round == 0 || authority_index.value() >= committee_size {
-            return None;
+        // threshold_clock_round should be always at least 1.
+        if threshold_clock_round == 0 {
+            return;
```
Thoralf-M, line 304: `should be always at least 1` only when updating the score or generally? Maybe good to add a warn log if it's the second case?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -164,11 +162,12 @@ impl Store for RocksDBStore {
                 )
                 .map_err(ConsensusError::RocksDBFailure)?;
         }
-        for (authority, metrics) in write_batch.scoring_metrics {
+        if let Some(metrics) = &write_batch.scoring_metrics {
             batch
-                .insert_batch(&self.scoring_metrics_v2, [(authority, metrics)])
+                .insert_batch(&self.scoring_metrics_v2, [(&0u32, metrics)])
```
Thoralf-M: Create a named const for 0u32

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -241,52 +240,35 @@ impl Store for RocksDBStore {
...
-        // the new scoring metrics field is empty, then we read from the legacy field,
-        // convert the data to the new format, and write them to the new field for
-        // future reads.
-        let mut metrics_by_author = vec![];
-        for kv in self.scoring_metrics_v2.safe_iter() {
-            metrics_by_author.push(kv?);
+        committee: &Committee,
+    ) -> ConsensusResult<Option<VersionedScoringMetrics>> {
+        // Try to read the single blob from the v2 CF first.
+        if let Some(metrics) = self.scoring_metrics_v2.get(&0u32)? {
```
Thoralf-M: Then use the named const also here

consensus/core/src/storage/mem_store.rs
```diff
@@ -31,7 +34,7 @@ struct Inner {
     commits: BTreeMap<(CommitIndex, CommitDigest), TrustedCommit>,
     commit_votes: BTreeSet<(CommitIndex, CommitDigest, BlockRef)>,
     commit_info: BTreeMap<(CommitIndex, CommitDigest), CommitInfo>,
-    scoring_metrics: BTreeMap<AuthorityIndex, VersionedStorageScoringMetrics>,
+    scoring_metrics: BTreeMap<u32, VersionedScoringMetrics>,
```
semenov-vladyslav: What is the purpose of having `scoring_metrics` as `BTreeMap`? Is it the intention to add more metrics (eg. v3, v4) and use them together?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -25,6 +23,8 @@ use crate::{
     storage::StorageScoringMetrics,
 };
 
+const SCORING_METRICS_V2_KEY: u32 = 0;
```
semenov-vladyslav: Should it be unified with the const of the same name from `mem_store`?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -40,8 +40,8 @@ pub(crate) struct RocksDBStore {
     commit_info: DBMap<(CommitIndex, CommitDigest), CommitInfo>,
     /// Legacy scoring metrics (read-only).
     scoring_metrics: DBMap<AuthorityIndex, StorageScoringMetrics>,
-    /// Stores versioned scoring metrics for each authority.
-    scoring_metrics_v2: DBMap<AuthorityIndex, VersionedStorageScoringMetrics>,
+    /// Stores versioned scoring metrics as a single blob under key 0.
+    scoring_metrics_v2: DBMap<u32, VersionedScoringMetrics>,
```
semenov-vladyslav: Should `scoring_metrics` be deprecated and removed (after a migration release) then?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -381,6 +368,60 @@ impl Store for RocksDBStore {
...
+                // We add a component with zeroed metrics for the authority with index i.
+                // This will ensure that every authority has an entry here.
+                // They are initialized as zero because if an authority does not have any
+                // recovered metrics, it means that it never misbehaved in a way that was
+                // detected by the node.
+                legacy_entries.insert(i.value(), (i, StorageScoringMetrics::new_zeroed()));
+            }
+        }
+    }
+
```
semenov-vladyslav:
```suggestion
```
It doesn't make sense to add zero entries into `legacy_entries`, the values for the new scoring metrics are zero-initialized anyway.

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -381,6 +368,60 @@ impl Store for RocksDBStore {
...
+    for (authority_index, old) in &legacy_entries {
+        let idx = authority_index.value();
+        faulty_blocks_provable[idx] = old.faulty_blocks_provable;
+        faulty_blocks_unprovable[idx] = old.faulty_blocks_unprovable;
+        missing_proposals[idx] = old.missing_proposals;
+        equivocations[idx] = old.equivocations;
+    }
+
+    let blob = VersionedScoringMetrics::V1(
```
semenov-vladyslav, line 393: The outer function is called `migrate_stored_metrics_v1_to_v2` but a `V1` value is returned. This is a bit confusing.

consensus/core/src/scoring_metrics_store.rs
```diff
@@ -378,11 +373,11 @@ impl MysticetiScoringMetricsStore {
     pub(crate) fn update_current_local_metrics_count(&self, authority_index: AuthorityIndex) {
         let uncached_metrics = &self
             .uncached_metrics
-            .iterate_over_metrics()
+            .iter()
             .map(|metric_vec| metric_vec[authority_index].load(Ordering::Relaxed))
             .collect::<Vec<u64>>();
```
semenov-vladyslav:
```suggestion
            ;
```
No need to collect, the iterator will be used a couple lines below.

semenov-vladyslav (approved):

consensus/core/src/scoring_metrics_store.rs
oliviasaa: fixed in [3efaaa8](https://github.com/iotaledger/iota/pull/10288/commits/3efaaa8f639acac4c7f4b8e181fb4ab10c3b806b)

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: fixed (?) in [3efaaa8](https://github.com/iotaledger/iota/pull/10288/commits/3efaaa8f639acac4c7f4b8e181fb4ab10c3b806b). I'm not sure if it became less confusing or more.

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: added a comment in [3efaaa8](https://github.com/iotaledger/iota/pull/10288/commits/3efaaa8f639acac4c7f4b8e181fb4ab10c3b806b) mentioning this

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: unified in [3efaaa8](https://github.com/iotaledger/iota/pull/10288/commits/3efaaa8f639acac4c7f4b8e181fb4ab10c3b806b)

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -39,9 +37,11 @@ pub(crate) struct RocksDBStore {
     /// Stores info related to Commit that helps recovery.
     commit_info: DBMap<(CommitIndex, CommitDigest), CommitInfo>,
     /// Legacy scoring metrics (read-only).
+    /// TODO: remove this field after migration is done.
+    #[deprecated]
     scoring_metrics: DBMap<AuthorityIndex, StorageScoringMetrics>,
-    /// Stores versioned scoring metrics for each authority.
-    scoring_metrics_v2: DBMap<AuthorityIndex, VersionedStorageScoringMetrics>,
+    /// Stores versioned scoring metrics as a single blob under key 0.
```
Thoralf-M: Outdated comment `key 0`

consensus/core/src/storage/mem_store.rs
```diff
@@ -31,7 +32,7 @@ struct Inner {
     commits: BTreeMap<(CommitIndex, CommitDigest), TrustedCommit>,
     commit_votes: BTreeSet<(CommitIndex, CommitDigest, BlockRef)>,
     commit_info: BTreeMap<(CommitIndex, CommitDigest), CommitInfo>,
-    scoring_metrics: BTreeMap<AuthorityIndex, VersionedStorageScoringMetrics>,
+    scoring_metrics: BTreeMap<(), VersionedScoringMetrics>,
```
Thoralf-M: Why not also just use Option<VersionedScoringMetrics> here?

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -241,52 +243,36 @@ impl Store for RocksDBStore {
         Ok(blocks)
     }
 
-    // This method is currently not only a scan, but it also performs migration from
-    // the legacy `StorageScoringMetrics` type to the new
-    // `VersionedStorageScoringMetrics`. This migration logic should be deleted
-    // after `StorageScoringMetrics` is removed. The writes in this method are
-    // only triggered once (for a one-time migration).
+    // Reads scoring metrics from the v2 CF (single blob under key 0). If not found,
```
Thoralf-M: Also outdated `key 0`

Thoralf-M (approved):

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: fixed in [d53e1fd](https://github.com/iotaledger/iota/pull/10288/commits/d53e1fd114e5e46bc548d89d66b6f2fb3ce40ce4)

consensus/core/src/storage/mem_store.rs
oliviasaa: fixed in [d53e1fd](https://github.com/iotaledger/iota/pull/10288/commits/d53e1fd114e5e46bc548d89d66b6f2fb3ce40ce4)

consensus/core/src/storage/rocksdb_store.rs
oliviasaa: fixed in [d53e1fd](https://github.com/iotaledger/iota/pull/10288/commits/d53e1fd114e5e46bc548d89d66b6f2fb3ce40ce4)

PR #10294, opened 2026-02-16
feat: page identity (part 2)

### LINKED ISSUE #9683: Trust Framework MVP: Add domain linkage
Add domain linkage. Design is not important at this point. I wouldnt overdo the design because it will be changed later on when the design team submits the design, I would just go simple with a list of domains and a verification hint

Image

alexruzenhack:
# Description of change

Implement Domain Linkage Services, fields information tooltips and rename DID to Identity with impact in the router `/identity/{did}` and some component's name.

Related issues:
- #9683

### Domain Linkage Services

image

### Tooltips Info
**Tooltips on fields and section titles:**
image

**Tooltip example from Object ID:**
image

### Renaming to Identity

**Tag in search:**
image

**Router path:**
```txt
http://localhost:3000/identity/did-iota-testnet-0xdc704ab63984d5763576c12ce5f62fe735766bc1fc9892a5e2a7be777a9af897?network=testnet
```

**Header title:**
image

## Notes

- Header **Identity** doesn't have a tooltip information because the header component doesn't support it and the effort to add this is not worthwhile.
- Section **Transaction Blocks** doesn't need tooltip because it is expected the user already know how to operate with it, because this is a common structure in other pages.


apps/explorer/src/pages/trust-framework/identity-result/views/IdentitySummaryView.tsx
marc2332: While you are at it
```suggestion
    const createdAt = didDateFormat(didDocument.metadataCreated()!.toRFC3339());
```

apps/explorer/src/pages/trust-framework/identity-result/hooks/useValidateDomainLinkage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { type IotaDocument } from '@iota/identity-wasm/web';
+import { validateDomainLinkageByEndpoint } from '../identityServiceHelper';
+import { useQuery } from '@tanstack/react-query';
+
+export const getValidateDomainLinkageQuery = (
+    issuerDocument: IotaDocument,
+    endpointUrl: string,
+) => ({
+    queryKey: ['did-jwt', issuerDocument, endpointUrl],
```
marc2332, line 12: Is this really a good idea to put the whole object here? MAybe just do this:

```suggestion
    queryKey: ['did-jwt', issuerDocument.id(), endpointUrl],
```

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,174 @@
...
+            isSuccess: false,
+            isError: true,
+            errorMsg: 'Invalid Content-Type.',
+        };
+    }
+
+    const data = await response.json();
+
+    // 3. Validate the properties of the received JSON.
+    const dataKeys = Object.keys(data);
```
marc2332: Use zod instead.

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,174 @@
...
+    // 2. Validate Content-Type to ensure we are processing JSON.
+    const contentType = response.headers.get('Content-Type');
+    if (!contentType || !contentType.startsWith('application/json')) {
+        console.warn(`Unexpected Content-Type: ${contentType}`);
+        return {
+            isSuccess: false,
+            isError: true,
+            errorMsg: 'Invalid Content-Type.',
+        };
+    }
```
marc2332, line 98: Do we really need this? Is checking the size and content type actually needed?

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,174 @@
...
+            issuerDocument,
+            domainLinkageConfiguration,
+            endpointUrl,
+            new JwtCredentialValidationOptions(),
+        );
+
+        return valid;
+    } catch {
+        console.error('Invalid Domain Linkage');
+        return invalid;
```
marc2332:
```suggestion
export async function validateDomainLinkageByEndpoint(
    issuerDocument: IotaDocument,
    endpointUrl: string,
) {
    const { data: didConfigurationJson, isError } = await fetchDidConfigurationJson(endpointUrl);
    if (isError) {
        return false;
    }

    try {
        const domainLinkageConfiguration = getFirstDomainLinkageConfiguration(
            didConfigurationJson!.linked_dids,
        );

        // Throws if invalid
        new JwtDomainLinkageValidator(new EcDSAJwsVerifier()).validateLinkage(
            issuerDocument,
            domainLinkageConfiguration,
            endpointUrl,
            new JwtCredentialValidationOptions(),
        );

        return true;
    } catch {
        console.error('Invalid Domain Linkage');
        return false;
```

marc2332 (changes requested):

apps/explorer/src/pages/trust-framework/identity-result/hooks/useValidateDomainLinkage.ts
alexruzenhack: No... it must be the whole document indeed, because this is a validation, not only a read. Also, tankstack lint claims it must be the whole object as a dependency.

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
alexruzenhack: Done.

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
alexruzenhack: I believe these 2 validations are good to have, indeed. We can debate allowed size, for sure, but it is good to put here a limit.

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
alexruzenhack: I prefer a more idiomatic code because increase readability in its semantics. But I'm ok removing it.

apps/explorer/src/pages/trust-framework/identity-result/views/IdentitySummaryView.tsx
```diff
@@ -37,9 +37,16 @@ export function DidSummaryView({
...
+                    {isActive && (
+                        <div>
+                            <DisplayStats
+                                label="Active"
+                                value={isActive ? 'Yes' : 'No'}
+                                tooltipPosition={TooltipPosition.Left}
+                                tooltipText="Whether this Identity is currently active on the ledger. Once deleted, an Identity is permanently deactivated and cannot be recovered."
+                            />
+                        </div>
+                    )}
```
marc2332: This looks weird, it will only show if `isActive` is true
```suggestion
                    <div>
                        <DisplayStats
                            label="Active"
                            value={isActive ? 'Yes' : 'No'}
                            tooltipPosition={TooltipPosition.Left}
                            tooltipText="Whether this Identity is currently active on the ledger. Once deleted, an Identity is permanently deactivated and cannot be recovered."
                        />
                    </div>
```

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,162 @@
...
+    const domainLinkageConfiguration = new DomainLinkageConfiguration([Jwt.fromJSON(didJwt)]);
+    return domainLinkageConfiguration;
+}
+
+export async function validateDomainLinkageByEndpoint(
+    issuerDocument: IotaDocument,
+    endpointUrl: string,
+) {
+    const { data: didConfigurationJson, isError } = await fetchDidConfigurationJson(endpointUrl);
+    if (isError) {
```
marc2332:
```suggestion
    if (isError || !didConfigurationJson) {
```

This way you don't have to use `!` later

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,162 @@
...
+        return false;
+    }
+
+    try {
+        const domainLinkageConfiguration = getFirstDomainLinkageConfiguration(
+            didConfigurationJson!.linked_dids,
```
marc2332:
```suggestion
            didConfigurationJson.linked_dids,
```

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
```diff
@@ -0,0 +1,162 @@
...
+        };
+    }
+
+    const data = await response.json();
+
+    // 3. Validate the structure and values of the received JSON.
+    const DomainLinkageResourceSchema = z
+        .object({
+            '@context': z.enum(SUPPORTED_DID_CONFIGURATION_CONTEXTS),
+            linked_dids: z.array(z.string()),
```
marc2332:
```suggestion
            linked_dids: z.array(z.string()).nonempty(),
```

apps/explorer/src/pages/trust-framework/identity-result/hooks/useValidateDomainLinkage.ts
marc2332: I dont understand, why wouldn't the id be a better choice? can the object change while having the same id?

marc2332 (changes requested):

apps/explorer/src/pages/trust-framework/identity-result/views/IdentitySummaryView.tsx
alexruzenhack: You totally right!

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
alexruzenhack: Applied!

apps/explorer/src/pages/trust-framework/identity-result/hooks/useValidateDomainLinkage.ts
alexruzenhack: Long story short: it doesn't harm to use the whole object of `issuerDocument` here.

Answering: In theory, yes, it can. A document can change its body and still keeps an ID. But in practice it will not happens in a render. However, this is not all the story. Tanstack has a lint and it complains when we use `issuerDocument.id()` because it recognises `issuerDocument` as a dependency, the whole object. And indeed we use the whole object in the validation operation.

apps/explorer/src/pages/trust-framework/identity-result/identityServiceHelper.ts
alexruzenhack: Done!

marc2332 (approved):

begonaalvarezd (approved):

PR #10299, opened 2026-02-16
refactor: add a transaction digest and a proof to MoveAuthenticator

### LINKED ISSUE #10271: [AA] Modify the `MoveAuthenticator` to have a `content` side and a `proof` side
Requirements:
- [ ] `content` here is an abstraction, meaning that the final implementation should have a part of data within the `MoveAuthenticator` struct which will be used to compute the `MoveAuthenticatorDigest` and another, separate, part which will include a `proof`
- [ ] Include the `TxDigest` in the `content`
- [ ] (optionally) limit the breaking of APIs due to this changes

valeriyr:
[run-ci]

# Description of change

TODO

fixes #10271

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1657,7 +1657,9 @@ impl AuthorityState {
                 auth_account_object_id,
                 auth_account_object_seq_number,
                 auth_account_object_digest,
-            ) = move_authenticator.object_to_authenticate_components()?;
+            ) = move_authenticator
+                .data()
+                .object_to_authenticate_components()?;
 
             // Since the `object_to_authenticate` components are provided, it is supposed
             // that the account object is loaded.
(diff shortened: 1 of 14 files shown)


crates/iota-core/src/authority.rs
```diff
@@ -1657,7 +1657,9 @@ impl AuthorityState {
                 auth_account_object_id,
                 auth_account_object_seq_number,
                 auth_account_object_digest,
-            ) = move_authenticator.object_to_authenticate_components()?;
+            ) = move_authenticator
+                .data()
+                .object_to_authenticate_components()?;
```
miker83z, line 1662: I would abstract this API: implement methods such as `object_to_authenticate_components` for `MoveAuthenticator` and then hide the call to `.data()` inside the method implementation.

crates/iota-types/src/move_authenticator.rs
```diff
@@ -54,17 +85,164 @@ impl Hash for MoveAuthenticator {
...
+        if author != self.data.address()? {
+            return Err(IotaError::InvalidSignature {
+                error: "Invalid author".to_string(),
+            });
+        };
+
+        let message = bcs::to_bytes(&value).expect("Message serialization should not fail");
+        let mut hasher = DefaultHash::default();
+        hasher.update(message);
+        let digest = hasher.finalize().digest;
```
miker83z, line 135: This approach is generic, implying that any data could be signed with that proof, not only `TransactionData`. 

What do you think about matching `T` exactly against `TransactionData` and read `TransactionData.digest()` from that? This would make this verification fail if the MoveAuthenticator was not used for a TX (and it will avoid performing an hash).

miker83z:
Closing it for now, keeping the branch alive

PR #10307, opened 2026-02-16
feat: Replace `Owner` with SDK version

### LINKED ISSUE #591: Replace `Owner` with SDK version

DaughterOfMars:
## Description

Replace `Owner`

Closes https://github.com/iotaledger/iota-rust-sdk/issues/591

### DIFF
--- a/crates/iota-benchmark/src/bank.rs
+++ b/crates/iota-benchmark/src/bank.rs
@@ -145,7 +145,10 @@ impl BenchmarkBank {
             .map_err(Error::msg)?;
 
         init_coin.0 = updated_gas.0;
-        init_coin.1 = updated_gas.1.get_owner_address()?;
+        init_coin.1 = *updated_gas
+            .1
+            .address()
+            .ok_or_else(|| Error::msg("not an address or object owner"))?;
         init_coin.2 = self.primary_coin.2.clone();
 
         let address_map: HashMap<IotaAddress, Arc<AccountKeyPair>> = coin_configs
@@ -157,7 +160,9 @@ impl BenchmarkBank {
             .created()
             .into_iter()
             .map(|c| {
-                let address = c.1.get_owner_address()?;
+                let address =
+                    *c.1.address()
+                        .ok_or_else(|| Error::msg("not an address or object owner"))?;
                 let keypair = address_map
                     .get(&address)
                     .ok_or("Owner address missing in the address map")
@@ -197,14 +202,20 @@ impl BenchmarkBank {
 
         self.primary_coin = (
             updated_gas.0,
-            updated_gas.1.get_owner_address()?,
+            *updated_gas
+                .1
+                .address()
+                .ok_or_else(|| Error::msg("not an address or object owner"))?,
             self.primary_coin.2.clone(),
         );
 
         match effects.created().first() {
             Some(created_coin) => Ok((
                 created_coin.0,
-                created_coin.1.get_owner_address()?,
+                *created_coin
+                    .1
+                    .address()
+                    .ok_or_else(|| Error::msg("not an address or object owner"))?,
                 self.primary_coin.2.clone(),
             )),
             None => panic!("Failed to create initialization coin for workload."),
--- a/crates/iota-genesis-builder/src/stardust/migration/migration.rs
+++ b/crates/iota-genesis-builder/src/stardust/migration/migration.rs
@@ -316,9 +316,9 @@ impl Extend<Object> for MigrationObjects {
             } else {
                 continue;
             };
-            let owner = object
+            let owner = *object
                 .owner
-                .get_owner_address()
+                .as_address_opt()
                 .expect("timelocks should have an address owner");
             owner_object_map
                 .entry(owner)
@@ -503,7 +503,7 @@ mod tests {
             .map(|move_object| {
                 Object::new_from_genesis(
                     Data::Move(move_object),
-                    Owner::AddressOwner(address),
+                    Owner::Address(address),
                     tx_context.digest(),
                 )
             });
@@ -547,7 +547,7 @@ mod tests {
             .map(|move_object| {
                 Object::new_from_genesis(
                     Data::Move(move_object),
-                    Owner::AddressOwner(owner),
+                    Owner::Address(owner),
                     tx_context.digest(),
                 )
             })
--- a/crates/iota-indexer/src/types.rs
+++ b/crates/iota-indexer/src/types.rs
@@ -307,10 +307,11 @@ impl TryFrom<i16> for OwnerType {
 // Returns owner_type, owner_address
 pub fn owner_to_owner_info(owner: &Owner) -> (OwnerType, Option<IotaAddress>) {
     match owner {
-        Owner::AddressOwner(address) => (OwnerType::Address, Some(*address)),
-        Owner::ObjectOwner(address) => (OwnerType::Object, Some(*address)),
+        Owner::Address(address) => (OwnerType::Address, Some(*address)),
+        Owner::Object(object_id) => (OwnerType::Object, Some(*object_id.as_address())),
         Owner::Shared { .. } => (OwnerType::Shared, None),
         Owner::Immutable => (OwnerType::Immutable, None),
+        _ => unimplemented!("a new enum variant was added and needs to be handled"),
     }
 }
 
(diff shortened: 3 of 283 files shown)


muXxer (approved):

crates/iota-analytics-indexer/src/tables.rs
```diff
@@ -148,8 +148,8 @@ pub enum ObjectStatus {
 // Object owner information.
 #[derive(Serialize, Clone, Display)]
 pub enum OwnerType {
-    AddressOwner,
-    ObjectOwner,
+    Address,
+    Object,
```
Thoralf-M: Isn't this change breaking the indexer?

crates/iota-indexer/src/types.rs
```diff
@@ -307,10 +307,11 @@ impl TryFrom<i16> for OwnerType {
 // Returns owner_type, owner_address
 pub fn owner_to_owner_info(owner: &Owner) -> (OwnerType, Option<IotaAddress>) {
     match owner {
-        Owner::AddressOwner(address) => (OwnerType::Address, Some(*address)),
-        Owner::ObjectOwner(address) => (OwnerType::Object, Some(*address)),
+        Owner::Address(address) => (OwnerType::Address, Some(*address)),
+        Owner::Object(address) => (OwnerType::Object, Some(*address.as_address())),
```
Thoralf-M:
```suggestion
        Owner::Object(object_id) => (OwnerType::Object, Some(*object_id.as_address())),
```

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1448,13 +1447,14 @@ mod checked {
...
         let is_mutable_input = match obj.owner {
-            Owner::AddressOwner(_) => true,
+            Owner::Address(_) => true,
             Owner::Shared { .. } => !override_as_immutable,
             Owner::Immutable => false,
-            Owner::ObjectOwner(_) => {
+            Owner::Object(_) => {
                 // protected by transaction input checker
-                invariant_violation!("ObjectOwner objects cannot be input")
+                invariant_violation!("Object objects cannot be input")
```
Thoralf-M: Maybe this?
```suggestion
                invariant_violation!("Object owned objects cannot be input")
```

crates/iota-analytics-indexer/src/tables.rs
DaughterOfMars: it was accidental, thanks

crates/iota-genesis-builder/src/stardust/migration/migration.rs
```diff
@@ -316,9 +316,9 @@ impl Extend<Object> for MigrationObjects {
             } else {
                 continue;
             };
-            let owner = object
+            let owner = *object
                 .owner
-                .get_owner_address()
+                .as_address_opt()
```
thibault-martinez, line 321: Different behaviour?

Thoralf-M (approved):

crates/iota-benchmark/src/bank.rs
```diff
@@ -145,7 +145,10 @@ impl BenchmarkBank {
             .map_err(Error::msg)?;
 
         init_coin.0 = updated_gas.0;
-        init_coin.1 = updated_gas.1.get_owner_address()?;
+        init_coin.1 = *updated_gas
+            .1
+            .address()
```
thibault-martinez, line 150: We could not just `?`?

crates/iota-genesis-builder/src/stardust/migration/migration.rs
DaughterOfMars: Nice catch

DaughterOfMars: Oh, wait, I did this one on purpose because of the error. It seemed like it must be an address

crates/iota-benchmark/src/bank.rs
DaughterOfMars: no

crates/iota-genesis-builder/src/stardust/migration/migration.rs
thibault-martinez: Please confirm with @iotaledger/vm-language, it could still have meant that it's tied to an address, so possibly Address AND Object variants

crates/iota-sdk/examples/utils.rs
```diff
@@ -163,8 +163,12 @@ pub async fn request_tokens_from_faucet(
             .await?;
 
         if owner.owner().is_some() {
-            let owner_address = owner.owner().unwrap().get_owner_address()?;
-            if owner_address == address {
+            if owner
+                .owner()
+                .unwrap()
+                .address()
+                .is_some_and(|a| a == &address)
```
thibault-martinez: Just an example but still, slightly different behaviour, maybe add an unwrap to address()

DaughterOfMars: like you said it's just an example :c

thibault-martinez: Well it's actually an example util, it could be doing something completely different now, impacting all examples

DaughterOfMars: Still, they are just examples

thibault-martinez: Please change it, we can't be providing examples that don't work as intended

crates/iota-types/src/object.rs
```diff
@@ -452,104 +452,6 @@ impl Data {
...
-#[derive(
-    Eq, PartialEq, Debug, Clone, Copy, Deserialize, Serialize, Hash, JsonSchema, Ord, PartialOrd,
-)]
-#[cfg_attr(feature = "fuzzing", derive(proptest_derive::Arbitrary))]
-pub enum Owner {
-    /// Object is exclusively owned by a single address, and is mutable.
-    AddressOwner(IotaAddress),
-    /// Object is exclusively owned by a single object, and is mutable.
-    /// The object ID is converted to IotaAddress as IotaAddress is universal.
-    ObjectOwner(IotaAddress),
```
thibault-martinez, line 464: Why do we not have an Address in the SDK? In this context maybe it makes more sense than ObjectId

DaughterOfMars: I don't think it does, it's an object owner after all

crates/iota-types/src/iota_serde.rs
thibault-martinez, line 1: This should be in the SDK in a serialisation module like the rest

DaughterOfMars: I don't agree. There are a lot of custom serialization structs for json RPC like `IotaTransactionBlock` which exist to satisfy the requirements. This is basically just one of those. The only reason I put it in `iota-types` is because it seems to be the way most of these are done.

thibault-martinez: We've been solving this issue by having types dedicated to serialisation in the SDK since the beginning, I don't understand why you would add one here now. What will we do when we have a type in the SDK that contains an Owner? It will be missing this.

DaughterOfMars: I think we shouldn't be doing that in the SDK. Since the RPC can't change, we should just have constant unchangeable types and defs in there

muXxer: Is this only used for the json-rpc? If yes, we should add it to the json-rpc-types instead, to not encourage someone else to use these types.

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2464,6 +2464,7 @@ impl TransactionBlockBytes {
     }
 }
 
+#[serde_as]
```
thibault-martinez: Just checking if this is necessary or was forgotten

crates/iota-types/src/iota_serde.rs
thibault-martinez: @muXxer We already have such types in the SDK, they are nested in a serialisation module and private, nobody can use them.

crates/iota-json-rpc-types/src/iota_transaction.rs
DaughterOfMars: ah, forgotten, thanks

thibault-martinez (approved):

Alex6323 (approved):

PR #10311, opened 2026-02-16
feat!: Replace `RandomnessStateUpdate` with SDK version

### LINKED ISSUE #584: Replace `RandomnessStateUpdate` with SDK version

DaughterOfMars:
## Description 

Replaces `RandomnessStateUpdate`

Closes https://github.com/iotaledger/iota-rust-sdk/issues/584

### DIFF
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -843,7 +843,7 @@ impl CheckpointExecutor {
                 if let TransactionKind::RandomnessStateUpdate(rsu) =
                     maybe_randomness_tx.data().transaction_data().kind()
                 {
-                    vec![rsu.randomness_round]
+                    vec![RandomnessRound(rsu.randomness_round)]
                 } else {
                     Vec::new()
                 }
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -843,7 +843,7 @@ impl CheckpointExecutor {
                 if let TransactionKind::RandomnessStateUpdate(rsu) =
                     maybe_randomness_tx.data().transaction_data().kind()
                 {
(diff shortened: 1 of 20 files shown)


muXxer (approved):

Thoralf-M (approved):

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -843,7 +843,7 @@ impl CheckpointExecutor {
                 if let TransactionKind::RandomnessStateUpdate(rsu) =
                     maybe_randomness_tx.data().transaction_data().kind()
                 {
-                    vec![rsu.randomness_round]
+                    vec![RandomnessRound(rsu.randomness_round)]
```
thibault-martinez: Maybe something to discuss with consensus when we replay on develop but we should either move this to the SDK (it was used in `RandomnessStateUpdate`) or remove it completely.

DaughterOfMars: I think it's a useful thing to have, but I dunno if it needs to be in the struct itself

crates/iota-core/tests/staged/iota.yaml
```diff
@@ -881,16 +881,12 @@ PublicKey:
       ZkLogin:
         NEWTYPE:
           TYPENAME: ZkLoginPublicIdentifier
-RandomnessRound:
-  NEWTYPESTRUCT: U64
 RandomnessStateUpdate:
   STRUCT:
     - epoch: U64
-    - randomness_round:
-        TYPENAME: RandomnessRound
```
thibault-martinez, line 890: Is this up to date? Shouldn't we still have `RandomnessRound`?

DaughterOfMars: it serializes as a u64, which I think is better because they type exists purely for the methods on it

thibault-martinez: Alright

thibault-martinez (approved):

PR #10317, opened 2026-02-16
feat(iota-types)!: replace CallArg by Input

### LINKED ISSUE #582: Replace `CallArg` with SDK version

Thoralf-M:
# Description of change

Replaces the local `CallArg` enum in iota-types with a type alias to `iota_sdk_types::Input`, continuing the effort to unify types between the IOTA node and the SDK

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/582


crates/iota-benchmark/src/lib.rs
```diff
@@ -615,18 +615,17 @@ impl From<ObjectRef> for BenchMoveCallArg {
...
+            CallArg::ImmutableOrOwned(obj_ref) => BenchMoveCallArg::ImmOrOwnedObject(obj_ref),
+            CallArg::Shared {
+                object_id,
+                initial_shared_version,
+                mutable,
+            } => BenchMoveCallArg::Shared((object_id, initial_shared_version, mutable)),
+            CallArg::Receiving(_) => {
+                unimplemented!("Receiving is not supported for benchmarks")
+            }
+            _ => unimplemented!("a new CallArg enum variant was added and needs to be handled"),
```
DaughterOfMars, line 628: Every other similar message is identical, so maybe we just make these the same?

Thoralf-M: All 6 are the same?

DaughterOfMars: I mean make them the same as the others

Thoralf-M: I think the others should be changed instead, because without the actual name it's hard to find the relevant places if changes are done https://github.com/iotaledger/iota/issues/10402

DaughterOfMars: fair enough

crates/iota-types/src/transaction.rs
```diff
@@ -678,152 +636,51 @@ impl EndOfEpochTransactionKind {
...
-        // unwrap safe because every u64 value is BCS-serializable
-        CallArg::Pure(bcs::to_bytes(&n).unwrap())
-    }
-}
-
-impl From<u128> for CallArg {
-    fn from(n: u128) -> Self {
-        // unwrap safe because every u128 value is BCS-serializable
-        CallArg::Pure(bcs::to_bytes(&n).unwrap())
+/// Returns the input object kinds for a `CallArg`, excluding receiving objects.
```
bingyanglin: Just curious why we don't use something like `pub trait CallArgExt`, and use `impl CallArgExt for CallArg { ... }` for the following functions so might be more idiomatic or do I miss something?

DaughterOfMars: That could be a bit nicer

Thoralf-M: I was able to remove some functions completely, added the trait for the remaining two https://github.com/iotaledger/iota/pull/10317/commits/bc619a7d6b59604edc50aa6cfa539373e5f0918d

crates/iota-types/src/move_authenticator.rs
```diff
@@ -26,19 +26,45 @@ use crate::{
...
+}
+
 /// MoveAuthenticator is a GenericSignature variant that enables a new
 /// method of authentication through Move code.
 /// This function represents the data received by the Move authenticate function
 /// during the Account Abstraction authentication flow.
 #[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
 pub struct MoveAuthenticator {
     /// Input objects or primitive values
+    #[schemars(with = "Vec<BinaryCallArgSchema>")]
```
DaughterOfMars: The problem with this is the type needs to actually deserialize using this schema too

Thoralf-M: Where? I think in the API it's always only serialized in the transaction as base64 string, so not actually used

DaughterOfMars: If so then we should remove the definitions entirely, because if the openrpc schema doesn't match how it actually serializes that's kind of a problem

Thoralf-M: Removing it just changes the openrpc schema though and requires updates in the generated TS files, that's why I added it

DaughterOfMars: Well either way the serialization and schema have to match

Thoralf-M: Just removed the schema completely as it's not needed https://github.com/iotaledger/iota/pull/10317/commits/94c9a17b444f924398a4d49b080247f53dc88ed5

crates/iota-types/src/transaction.rs
```diff
@@ -15,23 +15,24 @@ use anyhow::bail;
 use enum_dispatch::enum_dispatch;
 use fastcrypto::{encoding::Base64, hash::HashFunction};
 use iota_protocol_config::ProtocolConfig;
-pub use iota_sdk_types::{Argument, AuthenticatorStateUpdateV1, RandomnessStateUpdate};
+pub use iota_sdk_types::{
+    Argument, AuthenticatorStateUpdateV1, RandomnessStateUpdate,
+    SharedObjectReference as SharedInputObject,
```
thibault-martinez: I would re-export that as `SharedObjectRef`, feels more consistent

crates/iota/src/keytool.rs
```diff
@@ -661,7 +661,7 @@ impl KeyToolCommand {
                         let call_arguments: Vec<String> = move_auth.call_args().iter().map(|arg| {
                             match arg {
                                 CallArg::Pure(bytes) => format!("0x{}", Hex::encode(bytes)),
-                                CallArg::Object(obj) => serde_json::to_string(obj).unwrap_or_else(|_| format!("{obj:?}")),
+                                other => serde_json::to_string(&other).unwrap_or_else(|_| format!("{other:?}")),
```
thibault-martinez: I would still explicit the variants and have a panic placeholder

crates/transaction-fuzzer/src/programmable_transaction_gen.rs
```diff
@@ -530,9 +530,9 @@ fn create_input_calls(
             Identifier::from_static("mint_vec"),
             vec![],
             vec![
-                CallArg::from(cap),
-                CallArg::from(coin_value),
-                CallArg::from(input_size),
+                CallArg::ImmutableOrOwned(cap),
+                CallArg::pure(&coin_value),
+                CallArg::pure(&input_size),
```
thibault-martinez, line 535: Why don't we have these from anymore?

sdk/typescript/src/client/types/generated.ts
thibault-martinez, line 1: How come so much removal?

Thoralf-M: https://github.com/iotaledger/iota/pull/10317#discussion_r2857837817

crates/transaction-fuzzer/src/programmable_transaction_gen.rs
Thoralf-M: Don't see the benefit, I think it's more confusing than having the actual variant visible

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Mysten Labs, Inc.
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
-
```
thibault-martinez, line 4: 👀

crates/iota-types/src/signature.rs
```diff
@@ -102,6 +102,32 @@ pub enum GenericSignature {
     MoveAuthenticator,
 }
 
+// GenericSignature is always serialized as a base64 string over JSON-RPC,
+// so its schema should reflect that rather than expanding internal variants.
+impl JsonSchema for GenericSignature {
```
thibault-martinez, line 107: Not too sure about having JSON-RPC stuff here

crates/iota-surfer/src/surfer_state.rs
```diff
@@ -161,9 +161,7 @@ impl SurferState {
         args: Vec<CallArg>,
     ) {
         let rgp = self.cluster.get_reference_gas_price().await;
-        let use_shared_object = args
-            .iter()
-            .any(|arg| matches!(arg, CallArg::Object(ObjectArg::SharedObject { .. })));
+        let use_shared_object = args.iter().any(|arg| matches!(arg, CallArg::Shared(_)));
```
thibault-martinez:
```suggestion
        let use_shared_object = args.iter().any(CallArg::is_shared);
```

crates/transaction-fuzzer/src/programmable_transaction_gen.rs
thibault-martinez: It's generally more usability. You can, e.g. start accepting `Into<Input>` in function parameters etc. No big deal tho.

crates/iota-types/src/signature.rs
Thoralf-M: Yes but we want to move them anyways somewhere else, just couldn't find a good place now or do you see a good fit?

crates/iota-surfer/src/surfer_state.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota/pull/10317/commits/9264a2b633d0a093f4483e23c70b88bc611c45fe

crates/iota-json-rpc-types/src/iota_transaction.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota/pull/10317/commits/9264a2b633d0a093f4483e23c70b88bc611c45fe

crates/iota/src/keytool.rs
Thoralf-M: Changed it to just serialize everything as the variant is also interesting here

crates/iota-types/src/transaction.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota/pull/10317/commits/9264a2b633d0a093f4483e23c70b88bc611c45fe

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -1168,7 +1168,7 @@ async fn create_new_bear(
 
     let pt = {
         let mut builder = ProgrammableTransactionBuilder::new();
-        let name_arg = builder.input(CallArg::Pure(bcs::to_bytes(name).unwrap()))?;
+        let name_arg = builder.input(CallArg::pure(&name.to_string()))?;
```
thibault-martinez: maybe use `builder.pure()` directly, will avoid the to_string

crates/iota-transaction-builder/src/utils.rs
```diff
@@ -98,15 +98,15 @@ impl TransactionBuilder {
         Ok(obj_refs)
     }
 
-    /// Resolve a provided [`ObjectID`] to the required [`ObjectArg`] for a
+    /// Resolve a provided [`ObjectID`] to the required [`CallArg`] for a
     /// given move module.
     async fn get_object_arg(
```
thibault-martinez, line 103: Rename the function then?

thibault-martinez: Or maybe we still want to convey the idea it's Object idk

crates/iota-types/src/move_authenticator.rs
```diff
@@ -93,33 +93,37 @@ impl MoveAuthenticator {
...
+            }
+            CallArg::Receiving(_) => {
+                return Err(UserInputError::Unsupported(
+                    "MoveAuthenticator cannot authenticate receiving objects".to_string(),
+                ));
+            }
+            _ => {
+                return Err(UserInputError::Unsupported(
+                    "Unknown CallArg variant in MoveAuthenticator".to_string(),
+                ));
```
thibault-martinez: Not sure it should even be an error

crates/iota-types/src/move_authenticator.rs
```diff
@@ -128,25 +132,26 @@ impl MoveAuthenticator {
     pub fn input_objects(&self) -> Vec<InputObjectKind> {
         self.call_args
             .iter()
-            .flat_map(|arg| arg.input_objects())
-            .chain(self.object_to_authenticate().input_objects())
+            .filter_map(|a| a.input_object())
+            .chain(self.object_to_authenticate().input_object())
```
thibault-martinez: It used to be plural but returned a single object?

crates/iota-types/src/move_authenticator.rs
```diff
@@ -128,25 +132,26 @@ impl MoveAuthenticator {
...
             .collect::<Vec<_>>()
     }
 
     pub fn receiving_objects(&self) -> Vec<ObjectRef> {
         self.call_args
             .iter()
-            .flat_map(|arg| arg.receiving_objects())
+            .filter_map(|arg| arg.receiving_objects().copied())
```
thibault-martinez: Why is this method plural?

crates/iota/src/keytool.rs
```diff
@@ -659,10 +659,7 @@ impl KeyToolCommand {
                     GenericSignature::PasskeyAuthenticator(passkey) => DecodedSigOutput::Passkey(passkey),
                     GenericSignature::MoveAuthenticator(move_auth) => {
                         let call_arguments: Vec<String> = move_auth.call_args().iter().map(|arg| {
-                            match arg {
-                                CallArg::Pure(bytes) => format!("0x{}", Hex::encode(bytes)),
-                                CallArg::Object(obj) => serde_json::to_string(obj).unwrap_or_else(|_| format!("{obj:?}")),
-                            }
+                            serde_json::to_string(&arg).unwrap_or_else(|_| format!("{arg:?}"))
```
thibault-martinez, line 662: Is this really equivalent? Serde doesn't add stuff from the variant?

Thoralf-M: It's not, it's adding the variants now and I think that's better

crates/iota-types/src/move_authenticator.rs
```diff
@@ -128,25 +132,26 @@ impl MoveAuthenticator {
     pub fn input_objects(&self) -> Vec<InputObjectKind> {
         self.call_args
             .iter()
-            .flat_map(|arg| arg.input_objects())
-            .chain(self.object_to_authenticate().input_objects())
+            .filter_map(|a| a.input_object())
```
Alex6323: I would keep the closure variable named `arg` instead of `a` tbh. In line 143 you kept it.

docs/examples/rust/stardust/alias-migration.rs
```diff
@@ -148,9 +148,7 @@ async fn main() -> Result<(), anyhow::Error> {
             );
 
             // Create an NFT collection
-            let nft_collection_name = builder
-                .input(CallArg::Pure(bcs::to_bytes("Collection name").unwrap()))
-                .unwrap();
+            let nft_collection_name = builder.input(CallArg::pure(&"Collection name")).unwrap();
```
thibault-martinez: builder.pure() ?

crates/iota-types/src/move_authenticator.rs
```diff
@@ -128,25 +132,26 @@ impl MoveAuthenticator {
...
 
     /// Returns all shared input objects used by the MoveAuthenticator,
     /// including those from the object to authenticate.
-    pub fn shared_objects(&self) -> Vec<SharedInputObject> {
+    pub fn shared_objects(&self) -> Vec<SharedObjectRef> {
         self.call_args
             .iter()
-            .flat_map(|arg| arg.shared_objects())
-            .chain(self.object_to_authenticate().shared_objects())
+            .filter_map(|e| e.as_shared_opt())
```
Alex6323: `e` -> `arg` ?

crates/iota-types/src/move_authenticator.rs
Thoralf-M: Yes https://github.com/iotaledger/iota/blob/bbff812efc52ea0c6ff5d7096d75a7535bd8c29f/crates/iota-types/src/transaction.rs#L705-L727
The might have done this as they wanted to support vec of objects at some point, but this was only added to the JSON RPC and never on CallArg

crates/iota-types/src/move_authenticator.rs
Thoralf-M: Because I copied the name from the original one here, will change

crates/iota-types/src/transaction.rs
```diff
@@ -658,154 +619,50 @@ impl EndOfEpochTransactionKind {
...
-                }]
-            }
-            // Receiving objects are not part of the input objects.
-            CallArg::Object(ObjectArg::Receiving(_)) => vec![],
-        }
-    }
+/// Extension trait for [`CallArg`] providing helper methods.
+pub trait CallArgExt {
+    /// Returns the input object kind for this argument, excluding receiving
+    /// objects.
```
Alex6323, line 625: Since `Input`/`CallArg` is non-exhaustive it might be better to just state when it returns a Some?

crates/iota-types/src/move_authenticator.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

docs/examples/rust/stardust/alias-migration.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

crates/iota-types/src/move_authenticator.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

crates/iota-types/src/move_authenticator.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

crates/iota-types/src/move_authenticator.rs
Thoralf-M: Also wasn't sure, changed it to unimplemented https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

crates/iota-indexer/tests/rpc-tests/write_api.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/5d74cc0cd076b1d7f3ba34133b46b05a0f9d0202

crates/iota-types/src/transaction.rs
Thoralf-M: I think that would just make the comment unnecessarily complex

crates/iota-transaction-builder/src/utils.rs
Thoralf-M: I think object_arg still makes sense, also it's not public and only used in this module

crates/iota-core/src/generate_format.rs
thibault-martinez, line 1: I have to admit I don't always understand what's happening on this file 🙃

crates/iota-core/src/generate_format.rs
```diff
@@ -389,6 +408,16 @@ fn get_registry() -> Result<Registry> {
 
     tracer.trace_type::<CheckpointData>(&samples).unwrap();
 
+    // `CallArg` (= `iota_sdk_types::Input`) has a custom binary Serialize that maps
+    // through a private `ObjectArg` helper enum. Every `trace_type` call for a type
+    // that (transitively) contains `CallArg` may leave `ObjectArg` in
+    // `incomplete_enums`, because serde_reflection's loop only iterates for the
```
thibault-martinez: Bit of a repetition with the Struct doc already

Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/c20b6ed972dfa8d95d43285c927a60309dcb572f

crates/iota-core/src/generate_format.rs
Thoralf-M: Nobody does

crates/iota-types/src/transaction.rs
Alex6323: okay, but at least not wrong in case `Input` gets a new variant. But don't care that much...

Thoralf-M: Well it depends on the new variant that gets added, could be also another one that is not an object and then it would still be correct

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -2507,29 +2507,28 @@ impl IotaCallArg {
...
-                    digest: object_ref.digest,
-                })
-            }
+            CallArg::Receiving(object_ref) => IotaCallArg::Object(IotaObjectArg::Receiving {
+                object_id: object_ref.object_id,
+                version: object_ref.version,
+                digest: object_ref.digest,
+            }),
+            _ => unimplemented!("a new CallArg enum variant was added and needs to be handled"),
         })
```
Alex6323, line 2532: what's our safety net here to ensure we fix those instances in time after a SDK version bump?

crates/iota-types/src/move_authenticator.rs
```diff
@@ -93,33 +93,33 @@ impl MoveAuthenticator {
         &self,
     ) -> UserInputResult<(ObjectID, Option<SequenceNumber>, Option<ObjectDigest>)> {
         Ok(match self.object_to_authenticate() {
-            CallArg::Pure(_) => {
+            CallArg::Pure { .. } => {
```
Alex6323: Why do you switch to struct style matching?

crates/iota-types/src/programmable_transaction_builder.rs
```diff
@@ -66,39 +66,38 @@ impl ProgrammableTransactionBuilder {
         ))
     }
 
-    pub fn obj(&mut self, obj_arg: ObjectArg) -> anyhow::Result<Argument> {
-        let id = obj_arg.id();
+    pub fn obj(&mut self, obj_arg: impl Into<CallArg>) -> anyhow::Result<Argument> {
+        let obj_arg: CallArg = obj_arg.into();
+        let id = *obj_arg
+            .object_id()
+            .ok_or_else(|| anyhow::anyhow!("expected object CallArg, found pure argument"))?;
```
Alex6323, line 73: was this the justification for `ObjectArg` to exist? But unfortunate how it is now tbh.

crates/iota-types/src/transaction.rs
```diff
@@ -658,154 +619,50 @@ impl EndOfEpochTransactionKind {
...
+        }) = self.as_shared_opt()
+        {
+            Some(InputObjectKind::SharedMoveObject {
+                id: object_id,
+                initial_shared_version,
+                mutable,
+            })
+        } else {
+            None
         }
```
Alex6323, line 649: this will silently ignore any new input objects we might add. Better to match on the variants and explicily return `None` only for `Receiving` and panic on anything still unknown.

crates/iota-types/src/programmable_transaction_builder.rs
Thoralf-M: No, it was split at some point to allow a vec of objects as input, but ended up to only be added in the JSON RPC like that

crates/iota-json-rpc-types/src/iota_transaction.rs
Thoralf-M: There is none as for all the other unexhaustive enums, we have to think about that ourselves

crates/iota-types/src/transaction.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota/pull/10317/commits/432f3ddb7a5006347f1523ed99fcf020d0b23e5e

crates/iota-types/src/move_authenticator.rs
Thoralf-M: Wasn't done intentionally, reverted in https://github.com/iotaledger/iota/pull/10317/commits/432f3ddb7a5006347f1523ed99fcf020d0b23e5e

crates/iota-core/src/unit_tests/authority_tests.rs
```diff
@@ -78,11 +78,9 @@ impl TestCallArg {
         state: &AuthorityState,
     ) -> Argument {
         match self {
-            Self::Pure(value) => builder.input(CallArg::Pure(value)).unwrap(),
+            Self::Pure(value) => builder.pure_bytes(value, false),
```
thibault-martinez, line 81: Just call pure?

thibault-martinez: Ah, it's already BCS?

crates/iota-types/src/signature.rs
thibault-martinez: Idk, making @DaughterOfMars aware of this, may be relevant to her PR.

crates/iota-core/src/unit_tests/authority_tests.rs
Thoralf-M: Yes just pure made the test fail

crates/iota-rest-api/openapi/openapi.json
```diff
@@ -1185,50 +1185,52 @@
...
       },
       "Argument": {
-        "anyOf": [
+        "oneOf": [
           {
             "title": "Gas",
             "type": "string",
             "enum": [
-              "gas"
+              "Gas"
```
thibault-martinez, line 1193: These changes are weird. CI was happy in the Argument PR, I don't know why it changed just now. Actually looks breaking.

crates/iota-rest-api/openapi/openapi.json
```diff
@@ -4237,30 +4228,29 @@
...
       },
       "Owner": {
-        "description": "Enum of different types of ownership for an object.\n\n# BCS\n\nThe BCS serialized form for this type is defined by the following ABNF:\n\n```text owner = owner-address / owner-object / owner-shared / owner-immutable\n\nowner-address   = %x00 address owner-object    = %x01 object-id owner-shared    = %x02 u64 owner-immutable = %x03 ```",
         "oneOf": [
           {
             "description": "Object is exclusively owned by a single address, and is mutable.",
             "type": "object",
             "required": [
-              "address"
+              "AddressOwner"
```
thibault-martinez, line 4236: Yeah I think something is wrong

thibault-martinez: Oh, I'm looking at the REST API

crates/iota-types/src/transaction.rs
```diff
@@ -658,154 +619,50 @@ impl EndOfEpochTransactionKind {
     }
 }
 
-impl CallArg {
-    pub fn input_objects(&self) -> Vec<InputObjectKind> {
+/// Extension trait for [`CallArg`] providing helper methods.
+pub trait CallArgExt {
+    /// Returns the input object kind for this argument, excluding receiving
+    /// objects.
+    fn input_object(&self) -> Option<InputObjectKind>;
```
thibault-martinez: I think I'd rename to `input_object_kind`, input_object feels too generic because a call arg already is that

crates/iota-types/src/transaction.rs
```diff
@@ -658,154 +619,50 @@ impl EndOfEpochTransactionKind {
...
-                    }
-                );
-            }
-            CallArg::Object(o) => match o {
-                ObjectArg::Receiving(_)
-                | ObjectArg::ImmOrOwnedObject(_)
-                | ObjectArg::SharedObject { .. } => (),
-            },
+    fn validity_check(&self, config: &ProtocolConfig) -> UserInputResult {
+        if let Some(value) = self.as_pure_value() {
```
thibault-martinez: Kind of the same comment as Alex, we're hiding a panic here

Thoralf-M: No, this one can't panic

crates/iota-rest-api/openapi/openapi.json
Thoralf-M: Yeah but we don't care?

crates/iota-types/src/transaction.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/10317/commits/81ebb22e9a28e04e5c22325dfc322317ee96a7bc

crates/iota-types/src/transaction.rs
thibault-martinez: No I meant by doing if let Some like this, we obfuscate the fact the match is non exhaustive

crates/iota-rest-api/openapi/openapi.json
thibault-martinez: Idk, I guess

crates/iota-types/src/transaction.rs
Thoralf-M: Ah okay, changed https://github.com/iotaledger/iota/pull/10317/commits/fc9a69725a943884439fb52bdc7fef121e51fdd7

thibault-martinez (approved):

Alex6323 (approved):

crates/iota-core/src/generate_format.rs
```diff
@@ -54,6 +54,26 @@ use roaring::RoaringBitmap;
...
 use typed_store::TypedStoreError;
 
+/// Proxy for the `ObjectArg` helper enum defined in the `iota-sdk-types`
+/// serialization module. `CallArg` (= `iota_sdk_types::Input`) has a custom
+/// binary `Serialize`/`Deserialize` that maps through this local `ObjectArg`
+/// type. serde_reflection's `trace_type` loop only iterates until the
+/// **top-level** type (`CallArg`) is complete, so `ObjectArg` ends up in
+/// `incomplete_enums`. By tracing this proxy (which calls
+/// `deserialize_enum("ObjectArg", ...)`) we drive all of ObjectArg's variants
+/// to completion before tracing `CallArg`.
```
DaughterOfMars: Can't we just add traces for all the various variants of the `iota_sdk_types::Input` type instead?

crates/iota-types/src/signature.rs
DaughterOfMars: Yeah, I made a similar change in my PR.

crates/iota-types/src/transaction.rs
```diff
@@ -1320,44 +1174,27 @@ impl Display for ProgrammableTransaction {
...
-        mutable: true,
-    };
-
-    pub fn id(&self) -> ObjectID {
-        self.id
-    }
+/// Merges `other` into `this` shared input object.
+/// If there is a conflict in mutability, the resulting object will be
+/// mutable. Errors if the id or initial_shared_version do not match.
+pub fn left_union_shared_input_objects(
```
DaughterOfMars: Where is this used? It seems highly specific to one use case

DaughterOfMars: I checked myself and found that the method can be private, so I went ahead and pushed that :)

crates/iota-core/src/generate_format.rs
Thoralf-M: Will try that

Thoralf-M: Seems to work fine https://github.com/iotaledger/iota/pull/10317/commits/6ad63a86321a787d957d1304794a4afd26462407

DaughterOfMars (approved):

PR #10318, opened 2026-02-16
feat(iota-types): add get_score_per_epoch for ValidatorV1

miker83z:
# Description of change

Added the `get_score_per_epoch` in `iota_types` that allows to retrieve the new Table's entries.
This function could be placed or used in different ways. It could be moved to `mod.rs` similarly to what happens to `get_validator_from_table`, for instance.

When it comes to APIs, unfortunately `IotaSystemStateSummary` or `IotaValidatorSummary` do not have a ref to the `extra_fields` bag id. This means that currently, through APIs, you could only perform this path:
1. Get dynamic-fields of 0x5
2. Read DF with key 2
3. Parse the `0x3::iota_system_state_inner::IotaSystemStateV2` type 
4. Read the field `validators. active_validators[x]. extra_fields.id`
5. Get dynamic-fields of this last id
6. Read DF with key `0x3::validator::ScoreRecordKey{ }`
7. Parse the `0x2::table::Table<u64, u64>` type 
8. Read the `id` field
9. Get dynamic-fields of this last id -> obtained all the table entries

Having a `extra_fields_id` field within `IotaValidatorSummary` (e.g., something like `exchange_rates_id`) would be needed to skip the first 4 steps. With `score_record_id` you would just need to execute step 9.

### DIFF
--- a/crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
+++ b/crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
@@ -2,6 +2,8 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::ops::RangeInclusive;
+
 use anyhow::Result;
 use fastcrypto::traits::ToFromBytes;
 use move_core_types::{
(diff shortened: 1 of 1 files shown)


crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
```diff
@@ -397,6 +436,40 @@ impl ValidatorV1 {
...
+        // from the table.
+        // It will fail as well if the table does not contain an entry for a given
+        // epoch.
+        let mut scores = vec![];
+        for i in 0..epochs.len() {
+            let score: u64 =
+                get_dynamic_field_from_store(&object_store, score_table.id, &epochs[i]).map_err(
+                    |err| {
+                        IotaError::IotaSystemStateRead(format!(
+                            "Failed to load epoch score from table: {err:?}"
```
cyberphysic4l: I think preferred behaviour would be to return a `Vec<Option<u64>>` with `None` when there is no entry for a given epoch in the Table. The absence of an entry just means the validator was not on the committee in that epoch and the perhaps the caller doesn't know for certain what epochs the validator was on the committee.

miker83z: [af9ad0f](https://github.com/iotaledger/iota/pull/10318/commits/af9ad0fd6542b4de5ed3ea4f54de9145760d0b16)

miker83z:
I'll keep the branch, to be re-opened once needed @cyberphysic4l

PR #10323, opened 2026-02-17
fix(tooling): prevent env plugin from blocking amplitude event delivery

evavirseda:
# Description of change

The dev_ prefix enrichment plugin was failing in the browser because `@iota/core` can’t rely on process.env. This moves env detection to the wallet app and passes an isDev flag to the plugin

### DIFF
--- a/apps/core/src/amplitude/plugins/attachEnvironmentPlugin.ts
+++ b/apps/core/src/amplitude/plugins/attachEnvironmentPlugin.ts
@@ -20,14 +20,9 @@ export function attachEnvironmentPlugin(
         type: 'enrichment' as const,
         setup: async () => {},
         execute: async (event: Event) => {
-            try {
-                if (isDev && event.event_type && !event.event_type.startsWith(DEV_EVENT_PREFIX)) {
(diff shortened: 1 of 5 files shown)


apps/core/src/amplitude/plugins/attachEnvironmentPlugin.ts
```diff
@@ -10,25 +10,24 @@ import { BrowserClient, BrowserConfig, EnrichmentPlugin, Event } from '@amplitud
...
-            ) {
-                context.event_type = DEV_EVENT_PREFIX + context.event_type;
+        execute: async (event: Event) => {
+            try {
+                if (isDev && event.event_type && !event.event_type.startsWith(DEV_EVENT_PREFIX)) {
+                    return { ...event, event_type: DEV_EVENT_PREFIX + event.event_type };
+                }
+                return event;
+            } catch {
+                return event;
```
begonaalvarezd: isnt this catch a bit weird? you are silencing the error here

begonaalvarezd (approved):

KeitoTadashi (approved):

PR #10326, opened 2026-02-17
feat(wallet): add external link opened amplitude event

### LINKED ISSUE #10175: Add external link opened amplitude event
Add external link opened amplitude event. 
When a user visits any kind of url that is external. The link visited can be private or public, and it should be private by default

It probably makes sense to review the usage of `ExternalLink` and make sure that is used in all places, so we really track all links. Also, we should add the logic to make the amplitude private so its not sent. We should not track links of addresses, digests, etc... in general only some links are public, like documentation, apps, etc...

Props:
- `value`: string | null
- `type`: `address | digest | application | … it can grow`

Also, remove `viewedLedgerTutorial`, it should be covered by this event

evavirseda:
# Description of change

fixes https://github.com/iotaledger/iota/issues/10175


VmMad (changes requested):
I noticed there's 3 places that don't send the event to ampli:

1. Transaction view: any link with an address doesn't send the event, only when clicking the button `View on Explorer` sends an event.
2. Address in the homepage: when hovering the truncated address it shows a `↗`, that on click opens to the explorer, it doesn't send an event.
3. The addresses in the manage accounts page, any `↗` also doesn't send the event.

Something else to point: When clicking an app from the recognized apps list two events are send: `application opened` & `external link opened`

VmMad (approved):
lgtm

Juligs (changes requested):
Looks good, only one thing, in the GasSummary component, when the gas summary it's sponsored or owner the `eventType` is missing  in the `ExplorerLink` : 
```
  <KeyValueInfo
                        keyText="Sponsor"
                        value={
                            <ExplorerLink
                                type={ExplorerLinkType.Address}
                                address={gasSummary.owner}
                                eventType="address"
                            >
                                {formatAddress(gasSummary.owner)}
                            </ExplorerLink>
                        }
                        fullwidth
                    />
```

Juligs (approved):
⭐

PR #10327, opened 2026-02-17
chore(wallet): Switch from webpack to rspack

marc2332:
rspack vs webpack vs rolldown

### rspack
- ✅ Massive build boost:
  - For me a local dev build went from ~104s (webpack) to ~4/5s (rspack), so 130x faster.
  - Faster development
  - Faster CIs
- ✅ Rust (I feel obligated to put this)
- ✅ Created by ByteDance (TikTok), meaning a major company backs this, not some random dev from nebraska
- ✅ WebPack friendly: Very similar concepts and APIs from Webpack, hence why the migration was easy
- 🔴 Smaller ecosystem, but big enough for our current use cases
- ℹ️ It just works, I could not find any issue or code that needs be changed, or anything like that, it just works.

### webpack
- 🔴 Slow, very slow
- 🔴 seems to be going into a sort of maintenance-only mode
- 🔴 Javascript

### rolldown
- ✅ Probably very fast
- ✅ Rust 
- ✅ Created by Void0, the authors of Vite 
- 🔴 Not 100% stable, still in RC
- 🔴 Not webpack friendly at all, in fact, its actually meant to power Vite.

### conclusion
So, I would keep the PR as it is, because rspack:
- ✅ Is stable
- ✅ Webpack friendly
- ✅ Fast af
- ✅ Has a good future ahead


Dr-Electron (approved):

apps/wallet/package.json
```diff
@@ -7,16 +7,16 @@
     "scripts": {
         "dev": "pnpm build:dev --watch",
         "build": "pnpm build:prod",
-        "build:prod": "cross-env BUILD_ENV=\"production\" NODE_ENV=\"production\" TS_NODE_PROJECT=\"./configs/ts/tsconfig.webpack.json\" webpack --progress",
-        "build:dev": "cross-env BUILD_ENV=\"development\" NODE_ENV=\"development\" TS_NODE_PROJECT=\"./configs/ts/tsconfig.webpack.json\" webpack --progress",
```
msarcev, line 11: 🤩

msarcev (approved):

begonaalvarezd (approved):

PR #10328, opened 2026-02-17
fix(wallet): exclude identify event

panteleymonchuk:
**Context**: the `$identify` event is a special system event in Amplitude that's automatically sent when you call the identify() method to set user properties. When the plugin prefixes it with dev_, it becomes `dev_$identify`, which Amplitude treats as a custom event rather than the special system event


evavirseda (changes requested):
i think instead of hardcoding `$identify`, we should skip prefixing all amplitude system events ($...) 🙏

evavirseda (approved):
🚀

Juligs (approved):

PR #10330, opened 2026-02-17
feat(examples): implement onesig

valeriyr:
# Description of change

The example shows how the [OneSig Protocol](https://github.com/LayerZero-Labs/OneSig/blob/main/PROTOCOL.md) can be implemented using IOTA Account Abstraction.
It doesn't contain a full protocol implementation, but a simplified version only, which shows the main idea.


lzpap (approved):

examples/move/onesig/sources/account.move
```diff
@@ -0,0 +1,91 @@
...
+    signature: vector<u8>,
+    _: &AuthContext,
+    ctx: &TxContext,
+) {
+    verify_merkle_root(account, &merkle_root, &signature);
+
+    verify_merkle_proof(&merkle_root, &merkle_proof, ctx);
+}
+
+/// Verify the Merkle root against the provided signatures.
```
miker83z: Please, write a comment that this method/account could be extended to have a set of public keys to validate the signature against, instead of just 1.

examples/move/onesig/sources/merkle.move
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2026 IOTA Stiftung
```
miker83z, line 1: Maybe also include the author from which you derived the library from

miker83z (approved):

examples/move/onesig/sources/account.move
valeriyr: added in 276d754d48

examples/move/onesig/sources/merkle.move
valeriyr: added in a47077c422

vivekjain23 (approved):

PR #10336, opened 2026-02-17
feat: replace `TransactionEffects` with SDK Version

### LINKED ISSUE #9845: Replace `TransactionEffects` with SDK version

Alex6323:
# Description of change

Replaces `TransactionEffects` type with its SDK version.

Fixes #9845 


crates/iota-types/src/effects/test_effects_builder.rs
```diff
@@ -4,15 +4,17 @@
...
+    base_types::ObjectRef,
+    digests::ObjectDigest,
+    effects::new_from_execution_v1,
     execution::SharedInput,
-    execution_status::ExecutionStatus,
-    gas::GasCostSummary,
     message_envelope::Message,
-    object::Owner,
     transaction::{InputObjectKind, SenderSignedData, TransactionDataAPI},
 };
```
thibault-martinez: Keep importing from `iota-types`, not `iota_sdk_types`. We're re-exporting while keeping the original names to keep changes to the minimum

Alex6323: yeah, I know, slip up here I guess.

DaughterOfMars: Sometimes my IDE will refuse to import from iota-types, it's really annoying. I have to do it manually then :c

bingyanglin (approved):

crates/iota-types/src/effects/test_effects_builder.rs
Alex6323: exactly the same for me. Too bad it doesn't list all possible import options.

crates/iota-types/src/test_checkpoint_data_builder.rs
```diff
@@ -404,14 +403,16 @@ impl TestCheckpointDataBuilder {
             .map(|id| self.wrapped_objects.remove(&id).unwrap())
             .collect();
         let mut effects_builder = TestEffectsBuilder::new(tx.data())
-            .with_created_objects(created_objects.iter().map(|(id, o)| (*id, *o.owner())))
-            .with_mutated_objects(
-                mutated_objects
-                    .iter()
-                    .map(|(id, o)| (*id, o.version(), *o.owner())),
-            )
+            // TODO update owner to SDK
```
Thoralf-M: TODO

piotrm50 (approved):

Thoralf-M:
Needs to be rebased

crates/iota-types/src/effects/test_effects_builder.rs
```diff
@@ -4,15 +4,17 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 
+use iota_sdk_types::{
```
DaughterOfMars: Why are we importing base types from here now?

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -1158,35 +936,37 @@ impl From<Bls12381PublicKey> for crate::crypto::AuthorityPublicKeyBytes {
...
-            }
-            crate::effects::UnchangedSharedKind::ReadDeleted(version) => {
-                Self::ReadDeleted { version }
-            }
-            crate::effects::UnchangedSharedKind::Cancelled(version) => Self::Cancelled { version },
-            crate::effects::UnchangedSharedKind::PerEpochConfig => Self::PerEpochConfig,
-        }
-    }
-}
+// impl From<UnchangedSharedKind> for crate::effects::UnchangedSharedKind {
```
DaughterOfMars: Commented for why?

Dr-Electron (approved):
Approving example changes

crates/iota-types/src/effects/test_effects_builder.rs
Alex6323: fixed

crates/iota-types/src/iota_sdk_types_conversions.rs
Alex6323: fixed

crates/iota-types/src/test_checkpoint_data_builder.rs
Alex6323: fixed

crates/iota-types/src/effects/test_effects_builder.rs
Alex6323: fixed all instances I hope (even cleaned up some I didn't introduce)

DaughterOfMars (approved):

crates/iota-types/src/effects/mod.rs
```diff
@@ -68,81 +61,464 @@ impl Message for TransactionEffects {
...
+    }
+
+    fn epoch(&self) -> EpochId {
+        match self {
+            TransactionEffects::V1(v1) => (*v1).epoch(),
+            _ => unimplemented!(
+                "a new TransactionEffects enum variant was added and needs to be handled"
+            ),
+        }
+    }
```
Thoralf-M, line 383: All these can be simplified a lot with a simple macro like this

```Rust
/// Delegates a method call on `TransactionEffects` to the inner variant.
/// When a new variant is added, each invocation will produce a compile-time
/// exhaustiveness error instead of a runtime `unimplemented!()`.
macro_rules! delegate_tx_effects {
    ($self:ident, $method:ident $(, $arg:expr)*) => {
        match $self {
            TransactionEffects::V1(v1) => v1.$method($($arg),*),
            _ => unimplemented!(
                "a new TransactionEffects enum variant was added and needs to be handled"
            ),
        }
    };
}

impl TransactionEffectsAPI for TransactionEffects {
    fn status(&self) -> &ExecutionStatus {
        delegate_tx_effects!(self, status)
    }

    fn into_status(self) -> ExecutionStatus {
        delegate_tx_effects!(self, into_status)
    }

    fn epoch(&self) -> EpochId {
        delegate_tx_effects!(self, epoch)
    }

    fn modified_at_versions(&self) -> Vec<(ObjectID, Version)> {
        delegate_tx_effects!(self, modified_at_versions)
    }
```

crates/iota-types/src/effects/v1.rs
```diff
@@ -0,0 +1,424 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2024 IOTA Stiftung
```
Thoralf-M: 2026

crates/iota-types/src/effects/mod.rs
Alex6323: after having a closer look I think your suggestion is valid

crates/iota-grpc-server/src/merge.rs
```diff
@@ -577,20 +577,14 @@ impl Merge<iota_types::effects::TransactionEffects> for TransactionEffects {
 
     fn merge(
         &mut self,
-        source: iota_types::effects::TransactionEffects,
+        sdk_effects: iota_types::effects::TransactionEffects,
         mask: &FieldMaskTree,
     ) -> Result<(), Self::Error> {
         if !mask.contains(Self::DIGEST_FIELD.name) && !mask.contains(Self::BCS_FIELD.name) {
             // No need to convert if no field is requested
```
Thoralf-M: `convert` was only relevant with the conversion we removed?

Alex6323: oh yeah, good catch, this whole trait impl can be removed, or better just forwarded to the `&TransactionEffects` impl

Alex6323: https://github.com/iotaledger/iota/pull/10336/commits/6913e09bc1dc0521756bf29da52ecf1e80228ccd

Thoralf-M: Shouldn't the early return stay? Only the comment should be changed

Alex6323: I don't think so. The early return was only there to prevent the costly conversion.

Thoralf-M: But isn't the Merge::merge also more costly?

Alex6323: no, because there it does the same checks again and just returns `Ok(())` if none of the fields are requested.

Thoralf-M (approved):

crates/iota-types/src/effects/mod.rs
```diff
@@ -68,81 +61,419 @@ impl Message for TransactionEffects {
...
+    impl Sealed for super::TransactionEffectsV1 {}
+}
 
+/// Sealed extension trait providing aggregating queries built on top of
+/// [`TransactionEffectsAPI`].
+///
+/// Implementations are provided for [`TransactionEffects`] and
+/// [`TransactionEffectsV1`]; the methods here are derivable from the base API
+/// and live in their own trait to keep the core API minimal.
+pub trait TransactionEffectsAPIExt: transaction_effects_api_ext::Sealed {
```
thibault-martinez: Let's rename this to `TransactionEffectsExt`, it's more consistent with the other Ext traits + it's not really an extension of the API trait

thibault-martinez (changes requested):

crates/iota-types/src/effects/mod.rs
```diff
@@ -68,81 +61,419 @@ impl Message for TransactionEffects {
...
+    /// unwrapped_then_deleted, wrapped and input shared objects.
+    fn all_affected_objects(&self) -> Vec<ObjectRef>;
+    /// Returns a condensed [`TransactionEffectsDebugSummary`] suitable for
+    /// logging and inspection.
+    fn summary_for_debug(&self) -> TransactionEffectsDebugSummary;
+}
+
+/// Creates a TransactionEffects message from the results of execution,
+/// choosing the correct format for the current protocol version.
+pub fn new_from_execution_v1(
```
thibault-martinez: Why not keep this and `estimate_effects_size_upperbound_v1` in the Ext trait?

crates/iota-types/src/effects/mod.rs
```diff
@@ -4,36 +4,36 @@
...
     event::Event,
     execution::SharedInput,
-    execution_status::ExecutionStatus,
-    gas::GasCostSummary,
     message_envelope::{Envelope, Message, TrustedEnvelope, VerifiedEnvelope},
-    object::Owner,
     storage::WriteKind,
 };
 
-pub(crate) mod effects_v1;
```
thibault-martinez, line 34: you are not actually removing this file

crates/iota-transaction-builder/src/package.rs
```diff
@@ -85,11 +85,11 @@ impl TransactionBuilder {
             let mut builder = ProgrammableTransactionBuilder::new();
             let capability_arg = match capability_owner {
                 Owner::Address(_) => CallArg::ImmutableOrOwned(upgrade_capability.object_ref()),
-                Owner::Shared(initial_shared_version) => CallArg::Shared(SharedObjectRef {
-                    object_id: upgrade_capability.object_ref().object_id,
+                Owner::Shared(initial_shared_version) => CallArg::Shared(SharedObjectRef::new(
```
thibault-martinez, line 88: Not sure why this is in this PR but there is actually more to change in the repo...

crates/iota-grpc-server/src/merge.rs
```diff
@@ -572,25 +572,14 @@ impl Merge<&Checkpoint> for Checkpoint {
...
 
 // Transaction implementations
-impl Merge<iota_types::effects::TransactionEffects> for TransactionEffects {
+impl Merge<iota_sdk_types::TransactionEffects> for TransactionEffects {
     type Error = RpcError;
 
     fn merge(
         &mut self,
-        source: iota_types::effects::TransactionEffects,
+        sdk_effects: iota_types::effects::TransactionEffects,
```
thibault-martinez: just call it effects

crates/iota-core/src/execution_cache/passthrough_cache.rs
```diff
@@ -15,7 +15,6 @@ use iota_types::{
     executable_transaction::VerifiedExecutableTransaction,
     global_state_hash::GlobalStateHash,
     iota_system_state::{IotaSystemState, get_iota_system_state},
-    message_envelope::Message,
```
thibault-martinez, line 18: How come these were not having warnings?

crates/iota-types/src/storage/mod.rs
```diff
@@ -610,6 +610,7 @@ pub fn get_transaction_output_objects(
     object_store: &dyn ObjectStore,
     effects: &TransactionEffects,
 ) -> Result<Vec<Object>, StorageError> {
+    use crate::effects::TransactionEffectsExt;
```
thibault-martinez: Import like the others

Cargo.lock
```diff
@@ -7384,6 +7384,31 @@ dependencies = [
 name = "iota-sdk-types"
 version = "0.0.1-alpha.1"
 source = "git+https://github.com/iotaledger/iota-rust-sdk.git?rev=fcbc436a1da0e9f46ea22a072627feaf6c518eaf#fcbc436a1da0e9f46ea22a072627feaf6c518eaf"
```
thibault-martinez: Probably forgot to update a rev because now we have 2 versions of the SDK

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -34,7 +34,6 @@ use iota_types::{
     executable_transaction::VerifiedExecutableTransaction,
     full_checkpoint_content::CheckpointData,
     global_state_hash::GlobalStateHash,
-    message_envelope::Message,
```
thibault-martinez, line 37: Ok so you are removing this because it was using `Message::digest` and you have a `TransactionEffects::digest` but they are not the same?

thibault-martinez (dismissed):

crates/iota-types/src/effects/mod.rs
Alex6323: there's not much of a good reason to have it in the traits IMHO.

Alex6323: I mean `new_from_execution_v1` is probably fine, but the other doesn't refer to `self` at all, so I think it's better as a free function.

thibault-martinez: I don't think the lack of self is an argument, associated methods are common and we have lots like this in the repository, I think they fit better in it so that the scope is clear when calling them.

Alex6323: okay, fine

Cargo.lock
Alex6323: weird, I used the script

Alex6323: yup, re-ran it but it still kept the old SHAs. Now replaced with a good'ol find/replace.

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
Alex6323: Only thing I remember about this issue is that clippy showed me a unused warning and that I can remove the import. I'll dig into it.

crates/iota-types/src/storage/mod.rs
thibault-martinez: Think you forgot to do it

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
Alex6323: this is a bit tricky and goes into how the compiler resolves the `fn digest()` signature (same for rust-analyzer). They give the inherent impl of this fn defined in the SDK precedence over the `Message` trait method, which is why the import wasn't needed. However, currently both impls point to different implementations of the same logic of how to determine what is fed to the hasher. In the SDK we hardcode the salt to `TransactionEffects::`, but in the core repo this is resolved by tracing. We can just make the trait impl point to the inherent impl, but either way, the removal of the imports is correct logically, bc they should never differ. I verified it with Claude that the hash is identical.

crates/iota-types/src/storage/mod.rs
Alex6323: no, just not pushed because I don't want to trigger CI yet

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
Alex6323: so, currently for consumers of the SDK that enable the `serde` feature it uses the SDKs impl, and with `serde` disabled it uses the core repo's impl. Again, this doesn't fail because both impls result in the same digest, but still something we probably want to avoid.

Alex6323: The simplest thing to do would be to keep the change (removing the `Message` imports) and adding a test here to make sure that both digests impls always compute the same hash.

crates/iota-core/src/execution_cache/passthrough_cache.rs
Alex6323: Explained in the other comment, but in short: with the `serde` feature enabled on the SDK, the compiler resolves to the inherent `fn digest` impl defined in `hash.rs` (SDK), and doesn't use the `Message` one (lower priority), which is why I was able to remove the import. I don't know if we want to fix that divergence eventually (I looked into it, but it requires refactoring in the SDK), but the simplest thing to do atm is to add a test that ensures those 2 hash implementations produce the same hash.

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
Alex6323: https://github.com/iotaledger/iota/pull/10336/commits/5f029b7c0e51654e16ae44289b99619b906ff643

crates/iota-core/src/execution_cache/passthrough_cache.rs
Alex6323: https://github.com/iotaledger/iota/pull/10336/commits/5f029b7c0e51654e16ae44289b99619b906ff643

crates/iota-grpc-server/src/merge.rs
```diff
@@ -572,26 +572,15 @@ impl Merge<&Checkpoint> for Checkpoint {
 }
 
 // Transaction implementations
-impl Merge<iota_types::effects::TransactionEffects> for TransactionEffects {
+impl Merge<iota_sdk_types::TransactionEffects> for TransactionEffects {
```
thibault-martinez: Keep the import from `iota_types`, otherwise you're mixing both syntaxes with L580

examples/custom-indexer/rust/hybrid_reader.rs
```diff
@@ -22,10 +22,7 @@ impl Worker for CustomWorker {
 
     async fn process_checkpoint(&self, checkpoint: Arc<CheckpointData>) -> Result<Self::Message> {
         // custom processing logic
-        println!(
-            "Processing checkpoint: {}",
-            checkpoint.checkpoint_summary.to_string()
-        );
+        println!("Processing checkpoint: {}", *checkpoint.checkpoint_summary);
```
thibault-martinez, line 25: in this and 3 other files, shouldn't need the deref

crates/iota-core/src/authority_aggregator.rs
```diff
@@ -480,22 +479,22 @@ struct ProcessCertificateState {
 #[derive(Debug)]
 pub enum ProcessTransactionResult {
     Certified {
-        certificate: CertifiedTransaction,
+        certificate: Box<CertifiedTransaction>,
```
thibault-martinez, line 482: Why were these boxed? Clippy?

crates/iota-e2e-tests/tests/reconfiguration_tests.rs
```diff
@@ -527,7 +526,7 @@ async fn test_validator_resign_effects() {
         .into_effects_for_testing();
     // Ensure that we are able to form a new effects cert in the new epoch.
     assert_eq!(effects1.epoch(), 1);
-    assert_eq!(effects1.executed_epoch(), 0);
+    assert_eq!(effects1.data().epoch(), 0);
```
thibault-martinez, line 529: this is confusing, why did this and change and not, e.g. L515?

kodemartin (approved):

roman1e2f5p8s (approved):

crates/iota-core/src/authority_aggregator.rs
Alex6323: I'll check, but I am pretty sure that was the issue, because otherwise the enum variants vary too much in size.

crates/iota-e2e-tests/tests/reconfiguration_tests.rs
Alex6323: I'll have a look and come back to you...

crates/iota-core/src/authority_aggregator.rs
thibault-martinez: Why would it vary? Supposedly TransactionEffects in the monorepo and the SDK were the same?

crates/iota-grpc-server/src/merge.rs
Alex6323: I am kind of tempted to alias the import so I don't have to repeat the import path 4 times, but I won't do it since then I need to do that for other imports as well :angry:

crates/iota-grpc-server/src/merge.rs
Alex6323: just realized I need to call it `source` so it's consistent.

thibault-martinez: I don't know about that, isn't it called source when there is a conversion? There is none anymore

examples/custom-indexer/rust/hybrid_reader.rs
Alex6323: we need it, because this has nothing do to with the `Arc`, but with the `Deref` impl of the `Envelope`.

thibault-martinez: Pretty sure the Display of Envelope forwards to the inner but alright, as you prefer

Alex6323: `Envelope` doesn't automatically (nor manually) impl `Display` for `CheckpointSummary`, but has a blanket impl to `deref` to it. It's not like `Arc`.

crates/iota-core/src/authority_aggregator.rs
Alex6323: It's a downstream effect of our SDK impl where we boxed the inner of the `V1` variant.

examples/custom-indexer/rust/hybrid_reader.rs
thibault-martinez: Ah I see, such a mess all these types and typedefs

thibault-martinez: I removed it locally and it didn't complain but my VSCode is rekt apparently

crates/iota-core/src/authority_aggregator.rs
Alex6323: The boxed version of this enum is 32 bytes, the unboxed version is 496 bytes. If we remove the boxes then the enum varies between 488 (CertifiedTransaction) and 272 (VerifiedCertifiedTransactionEffects + TransactionEvents) bytes.

iota-execution/latest/iota-adapter/src/temporary_store.rs
```diff
@@ -201,21 +204,58 @@ impl<'backing> TemporaryStore<'backing> {
...
+        &self,
+        id: &ObjectID,
+        results: &ExecutionResultsV1,
+    ) -> EffectsObjectChange {
+        let modified_at = self
+            .get_object_modified_at(id)
+            .map(|metadata| ((metadata.version, metadata.digest), metadata.owner));
+        let written = results.written_objects.get(id);
+        let id_created = results.created_object_ids.contains(id);
+        let id_deleted = results.deleted_object_ids.contains(id);
```
miker83z, line 218: NIT: I'd still keep the results in the caller, so making it similar to the previous version (above):
```rust
.map(|id| {
    (
        *id,
       self.new_effects_object_change(
           *id,
            self.get_object_modified_at(id)
                .map(|metadata| ((metadata.version, metadata.digest), metadata.owner)),
            results.written_objects.get(id),
            results.created_object_ids.contains(id),
            results.deleted_object_ids.contains(id),
        ),
    )
})
```

miker83z (approved):

crates/iota-e2e-tests/tests/reconfiguration_tests.rs
Alex6323: `effects0` (L515) is of an `iota-json-rpc-types` defined type (`IotaTransactionBlockEffects`), and its API coming from an extension trait also defined there. I think we can also rename this for consistency, but it's not an implied change, more a consistency/cleanup thing for the owners of this crate, but of course I could also do it in this PR.

semenov-vladyslav (approved):

examples/custom-indexer/rust/hybrid_reader.rs
Alex6323: yeah, it didn't complain probably because it's a separate example binary. You see it complaining when you compile the example from within.

crates/iota-grpc-server/src/merge.rs
Alex6323: It's the trait method's (`fn merge(&mut self, source: T, ...)`) argument name, and it's kept unchanged for all implementations.

crates/iota-grpc-server/src/merge.rs
Alex6323: Changed, but not pushed yet.

iota-execution/latest/iota-adapter/src/temporary_store.rs
Alex6323: But we're not moving them, just passing a reference? We can do it of course as you suggest, but I don't see the benefit that outweighs making the signature more verbose? :thinking: 

(I mean if locking were involved that would change things)

crates/iota-grpc-server/src/merge.rs
thibault-martinez: Alright, are you changing to source then?

iota-execution/latest/iota-adapter/src/temporary_store.rs
Alex6323: since we have a `&self` we can actually do this:

```rust
...
        all_ids
            .into_iter()
            .map(|id| (*id, self.new_effects_object_change(id)))
            .collect()
    }

    fn new_effects_object_change(&self, id: &ObjectID) -> EffectsObjectChange {
        let modified_at = self
            .get_object_modified_at(id)
            .map(|metadata| ((metadata.version, metadata.digest), metadata.owner));
        let results = &self.execution_results;
        let written = results.written_objects.get(id);
        let id_created = results.created_object_ids.contains(id);
        let id_deleted = results.deleted_object_ids.contains(id);
...
```

crates/iota-grpc-server/src/merge.rs
Alex6323: already done, just waiting for more replies on how to proceed on other comments, so I don't trigger the whole CI pipeline each time.

crates/iota-core/src/authority_aggregator.rs
thibault-martinez: Cool, thanks

iota-execution/latest/iota-adapter/src/temporary_store.rs
Alex6323: @miker83z Do you still want that part reverted like before or is my last suggestion i.e. accessing the results through `&self` also fine for you?

Cargo.lock
thibault-martinez, line 1: Two versions of the SDK

crates/iota-types/src/effects/mod.rs
```diff
@@ -68,24 +61,30 @@ impl Message for TransactionEffects {
...
+/// versions.
+///
+/// This trait is sealed: it is only implemented for [`TransactionEffects`] and
+/// the concrete version structs (currently [`TransactionEffectsV1`]). The
+/// implementation on [`TransactionEffects`] dispatches to the active version
+/// variant, so callers can write version-agnostic code.
+pub trait TransactionEffectsAPI: transaction_effects_api::Sealed {
+    /// Build effects from the results of executing a transaction under the
+    /// V1 protocol shape.
+    fn new_from_execution_v1(
```
thibault-martinez: This and `estimate_size_upperbound_v1` belong in the Ext IMO, they were not in the API

iota-execution/latest/iota-adapter/src/temporary_store.rs
miker83z: I see why this change is needed, but the `new_effects_object_change` function itself is a bit odd from my point of view. If I think about a constructor of a EffectsObjectChange, I wouldn't think to pass to it the execution results. Here it is done because the constructor is now in the same file, but if it wasn't, I don't think it would have been the same (like the previous version). 

It's just a nit, so ignore it if everything else is already sorted and the PR can be merged. Otherwise, could be also worth removing the constructor and moving all the logic inside the all_ids iter map.

crates/iota-types/src/effects/mod.rs
```diff
@@ -427,3 +699,145 @@ impl CertifiedTransactionEffects {
         Ok(VerifiedCertifiedTransactionEffects::new_from_verified(self))
     }
 }
+/// This function demonstrates what's the invariant of the effects.
+/// It also documents the semantics of different combinations in object
+/// changes.
+#[cfg(debug_assertions)]
+fn check_invariant(v1: &TransactionEffectsV1) {
```
thibault-martinez: move it to the v1 module?

Alex6323: https://github.com/iotaledger/iota/pull/10336/commits/efd14fff24b9bad3e868d06abf4e133a3d98614b

crates/iota-types/src/effects/mod.rs
Alex6323: https://github.com/iotaledger/iota/pull/10336/commits/efd14fff24b9bad3e868d06abf4e133a3d98614b

crates/iota-types/src/effects/mod.rs
```diff
@@ -254,49 +235,14 @@ pub trait TransactionEffectsAPI: transaction_effects_api::Sealed {
...
+/// Test-only mutators and unchecked builders for [`TransactionEffects`] that
+/// bypass the normal invariants. Not for production use.
 pub trait TransactionEffectsAPIForTesting: TransactionEffectsAPI {
     // All of these should be #[cfg(test)], but they are used by tests in other
     // crates, and dependencies don't get built with cfg(test) set as far as I
     // can tell.
-    /// Builds empty effects for `transaction_digest` with no object changes
-    /// and no gas object. Useful for system transactions whose effects content
-    /// is irrelevant to a test.
-    fn empty_for_testing(transaction_digest: TransactionDigest) -> Self
```
thibault-martinez: I think it was better here, it creates an "invalid" Effects, it should be flagged as so

Alex6323: you mean the name, the location or both?

thibault-martinez: I mean, I guess you did it because the Test trait is called "API" but maybe it's simply badly named idk

thibault-martinez: I'm opening an issue to not further delay this but I think this is wrong atm

thibault-martinez: https://github.com/iotaledger/iota/issues/11596

thibault-martinez (approved):

iota-execution/latest/iota-adapter/src/temporary_store.rs
Alex6323: I created an issue for this https://github.com/iotaledger/iota/issues/11597

PR #10338, opened 2026-02-17
refactor(evm-bridge): replace wallet user group functions with amplitude identity tracking

### LINKED ISSUE #10095: Remove all setGroups methods in amplitude
Replace Amplitude `setGroup` / `setWhateverGroup` usage with Identity-based properties

panteleymonchuk:
Closes #10095


apps/evm-bridge/src/shared/analytics/amplitude.ts
evavirseda, line 1: quick thought: do we need to change the property names as well? we’re going from `l1_wallet_type -  l2_wallet_type - l2_chain_id` to `l1WalletType - l2WalletType - l2ChainId`  just want to make sure this won’t break existing charts 🙏

apps/evm-bridge/src/shared/analytics/amplitude.ts
```diff
@@ -76,44 +76,70 @@ export function getUrlWithDeviceId(url: URL) {
...
     }
 
-    if (walletInfo.l2WalletType) {
-        ampli.client.setGroup('l2_wallet_type', walletInfo.l2WalletType);
+    // Update current state with provided options
+    if (options?.activeNetwork !== undefined) {
+        currentIdentity.activeNetwork = options.activeNetwork;
+    } else if (!currentIdentity.activeNetwork) {
+        // Initialize activeNetwork on first call
+        currentIdentity.activeNetwork = getDefaultNetwork();
```
evavirseda, line 104: `activeNetwork` is set to a default on init, but i’m not sure we update it when the user switches networks
if we want this to stay accurate, we might want to call `setAmplitudeIdentity({ activeNetwork })` when network change right?

evavirseda (dismissed):
just a couple of minor nits 🙏

apps/evm-bridge/src/shared/analytics/amplitude.ts
panteleymonchuk: afaik, for evm-bridge we don't switch network at all, just use it from the config.

apps/evm-bridge/src/shared/analytics/amplitude.ts
panteleymonchuk: Before we used it as a groups, but groups we can't use in the reports currently. So we don't lose anything imo.

brancoder (approved):
LGTM

begonaalvarezd (approved):

PR #10350, opened 2026-02-18
fix(wallet): Deduplicate unlock password dialog in wallet

marc2332:


marc2332:
https://github.com/iotaledger/iota/pull/10351

Issue #10315, opened 2026-02-16
[indexer] replace `dev_inspect` & `dry_run` with gRPC


sergiupopescu199:
The gRPC API unified the JSON RPC `dev_inspect` & `dry_run` into a single `simulate` method. Refactor the indexer implementation accordingly.

PR #10355, opened 2026-02-18
refactor(iota-indexer): use gRPC for `execute_transaction` & `is_transaction_indexed_on_node`

### LINKED ISSUE #10316: [indexer]: replace `execute_transaction` & `isIndexedOnNode` with gRPC API
Replace the JSON RPC `execute_transaction` & `isIndexedOnNode ` with the gRPC one.

sergiupopescu199:
# Description of change

This patch refactors the `iota-indexer` to use the gRPC API in `execute_transaction` & `is_transaction_indexed_on_node` JSON RPC implementation.

fixes #10316 

- Ran indexer tests, made sure that the `execute_transaction` and `is_transaction_indexed_on_node` test cases passed.

```bash
```


crates/iota-indexer/src/apis/read_api.rs
```diff
@@ -235,10 +239,24 @@ impl ReadApiServer for ReadApi {
...
-            .map_err(error_object_from_rpc)
+        {
+            Ok(mut txns) => {
+                let executed_tx = txns.pop().ok_or_else(|| {
+                    IndexerError::Grpc("there should be one tx lookup response".into())
+                })?;
+                matches_digest(digest, executed_tx).map_err(Into::into)
+            }
+            Err(e) => {
+                if matches!(e, iota_grpc_client::Error::Server(ref e) if e.contains("not found")) {
```
tomxey: that's a bit not elegant, is this something that we could fix? (to have dedicated NotFound error type and avoid string operations)
if not maybe we could suggest an improvement to node team?

crates/iota-indexer/src/errors.rs
```diff
@@ -164,6 +165,12 @@ pub enum IndexerError {
 
     #[error("Missing data due to pruning: `{0}`")]
     DataPruned(String),
+
+    #[error("grpc error: `{0}`")]
+    Grpc(String),
+
+    #[error(transparent)]
+    SdkTyperConversion(#[from] SdkTypeConversionError),
```
tomxey: typo: SdkTypeConversion

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -210,22 +228,17 @@ impl OptimisticTransactionExecutor {
             .map(|sig| GenericSignature::from_bytes(&sig.to_vec()?))
             .collect::<Result<Vec<_>, FastCryptoError>>()?;
 
-        let transaction = Transaction::from_generic_sig_data(tx_data, sigs);
-
+        let transaction = Transaction::from_generic_sig_data(tx_data.clone(), sigs.clone());
```
tomxey: why the params need to be cloned now? "tx_data.clone(), sigs.clone()"

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -98,11 +98,12 @@ impl TryFrom<Object> for crate::object::Object {
...
-        Self::new_from_genesis(
-            value.data.try_into()?,
-            value.owner.into(),
-            value.previous_transaction.into(),
-        )
+        Self::from(ObjectInner {
+            data: value.data.try_into()?,
+            owner: value.owner.into(),
+            previous_transaction: value.previous_transaction.into(),
+            storage_rebate: value.storage_rebate,
```
tomxey, line 105: did you inspect the existing callsites that they are not broken by this change?
I wonder what was the reason it was implemented like this

crates/iota/src/iota_commands.rs
```diff
@@ -916,6 +922,15 @@ async fn start(
...
     }
 
+    let mut config = GrpcApiConfig::default();
+    config.address.set_port(fullnode_grpc_port);
+
+    let fullnode_grpc_url = config.address;
+
+    swarm_builder = swarm_builder
+        .with_fullnode_enable_grpc_api(true)
+        .with_fullnode_grpc_api_config(config);
```
tomxey: Is this only for development purposes?

I see above we already have:

> swarm_builder = swarm_builder.with_fullnode_enable_grpc_api(enable_grpc_api);

crates/iota-indexer/src/lib.rs
```diff
@@ -68,12 +69,12 @@ pub async fn build_json_rpc_server(
     builder.register_module(TransactionBuilderApi::from(reader.clone()))?;
     builder.register_module(MoveUtilsApi::new(reader.clone()))?;
     builder.register_module(GovernanceReadApi::new(reader.clone()))?;
-    builder.register_module(ReadApi::new(reader.clone(), fullnode_client.clone()))?;
+    builder.register_module(ReadApi::new(reader.clone(), grpc_client.clone()))?;
     builder.register_module(CoinReadApi::new(reader.clone())?)?;
     builder.register_module(ExtendedApi::new(reader.clone()))?;
     builder.register_module(OptimisticWriteApi::new(
         WriteApi::new(fullnode_client, reader.clone()),
```
tomxey: WriteApi still uses the old client for tx execution, is it planned for refactor in a separate issue?

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -493,3 +507,25 @@ impl<'a> TransactionExtractor<'a> {
...
+                .and_then(|o| Object::try_from(o).map_err(Into::into))
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
+
+/// Converts [`GrpcTransactionEvents`] into [`TransactionEvents`]
+fn convert_events(events: &GrpcTransactionEvents) -> IndexerResult<Option<TransactionEvents>> {
+    events
+        .events()
+        .ok()
```
tomxey: Is it not hiding potential errors? what can fail here?

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -121,15 +128,26 @@ impl OptimisticTransactionExecutor {
...
-            effects,
-            events,
-            input_objects,
-            output_objects,
-            ..
-        } = execution_response;
+        let effects = execution_response.effects()?.effects()?.try_into()?;
+        let events = execution_response
+            .events()
+            .ok()
```
tomxey: is it not hiding potential errors? what can fail here?

here and also below in 

> .input_objects().ok()
> .output_objects().ok()

crates/iota-types/src/iota_sdk_types_conversions.rs
sergiupopescu199: I didn't find any reason why the implementation was done that way. We could raise this question with the original authors.

From the rustdoc we see that it's important to preserve this field:
```rust
    /// The amount of IOTA we would rebate if this object gets deleted.
    /// This number is re-calculated each time the object is mutated based on
    /// the present storage gas price.
    pub storage_rebate: u64,
```

I didn't spot any breaking changes. This fix affects the data ingestion core since we were converting grpc data into `CheckpointData` which includes `input_objects` and `output_objects` fields - meaning all objects ingested by the pipeline were affected by this bug.

I was able to find three instances where this conversion is being used, didn't spot any side effects:

- `iota/crates/iota-data-ingestion-core/src/reader/v2.rs` (lines 715-737)
-  `iota/crates/iota-indexer/src/optimistic_indexing.rs` (lines 512-522)
-  `iota/crates/iota-types/src/iota_sdk_types_conversions.rs` (lines 1786-1812)

The change is already on develop

crates/iota-indexer/src/lib.rs
sergiupopescu199: Refactored based on our last conversation.

We have a dedicated issue: #10315 which will use the gRPC in the WriteAPI for `dev_inspect` & `dry_run`.

crates/iota-indexer/src/apis/read_api.rs
sergiupopescu199: You're right, this is not elegant. The issue is that `iota_grpc_client::Error` currently wraps "not found" errors from `ObjectResult`/`TransactionResult` proto responses as `Error::Server(String)`, which loses the structured error type.

The gRPC server already has proper typed errors (`TransactionNotFoundError`, `ObjectNotFoundError` in `iota-grpc-server/src/error.rs`) that convert to `tonic::Status::not_found()`. However, the client doesn't preserve this distinction.

I raised this question to the node team, they will provide a dedicated type in the future.

crates/iota-indexer/src/errors.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/changes/4ac361d4209df3cc2e06cfcc4a854989502ab781#diff-c1a92abf8580742c8ceba99f8bc8c8bda7bade63b2e7aa78d9d404d9ce563ffaR173

crates/iota/src/iota_commands.rs
sergiupopescu199: As discussed this part was removed.

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/b44b89f3710f9dd4a33929b21f85addd500bf6a4

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/b44b89f3710f9dd4a33929b21f85addd500bf6a4

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: fixed

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -237,11 +330,15 @@ impl OptimisticTransactionExecutor {
             Err(e) => {
                 node_timer.stop_and_discard();
                 self.metrics.optimistic_tx_failed_node_requests_count.inc();
-                return Err(IndexerError::Generic(e.to_string()));
+                return Err(IndexerError::Generic(e));
             }
         };
 
-        let tx_digest = *response.effects.transaction_digest();
+        let tx_digest = match response {
```
tomxey: do you think the `match response` and `match self.rpc_client` blocks could be functions in `impl FullnodeRpc/Response` ?

i think it could be cleaner that way

crates/iota-indexer/src/apis/read_api.rs
```diff
@@ -415,3 +455,15 @@ impl IotaRpcModule for ReadApi {
         iota_json_rpc_api::ReadApiOpenRpc::module_doc()
     }
 }
+
+/// Converts the grpc digest from bcs bytes and matches if the provided digest.
```
tomxey: typo: **with** the provided digest.?

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -76,16 +99,25 @@ impl WriteApiServer for WriteApi {
...
+                    .await
+                    .map_err(error_object_from_rpc)?;
+                Ok(IotaTransactionBlockResponseWithOptions {
+                    response: iota_transaction_response,
+                    options: options.unwrap_or_default(),
+                }
+                .into())
+            }
+            FullnodeRpc::Grpc(_) => {
+                panic!("execute_transaction_block should be called form OptimisticWriteApi")
```
tomxey: typo: from

crates/iota-indexer/tests/rpc-tests/write_api.rs
```diff
@@ -1413,3 +1413,91 @@ fn clever_errors() {
...
+            "indexer_grpc_execute_transaction",
+            Some(Box::new(move |builder| {
+                builder.with_fullnode_enable_grpc_api(true)
+            })),
+            None,
+        )
+        .await;
+
+        // give some time for the grpc server to start
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
tomxey: did you have an issue with making a request too early?

if yes I think it would be nice to have some more robust solution, other than fixed sleep, since 1 second may also not be enough if system is under heavy load and processes start to lag

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: I thought that without impls it could've been more transparent. But I do agree that in this way it look cleaner.

fixed: https://github.com/iotaledger/iota/pull/10355/commits/9acf4eb9cc9e1d1a9fb289609ca84359253a9e13

crates/iota-indexer/src/apis/read_api.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/9acf4eb9cc9e1d1a9fb289609ca84359253a9e13

crates/iota-indexer/src/apis/write_api.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/9acf4eb9cc9e1d1a9fb289609ca84359253a9e13

crates/iota-indexer/tests/rpc-tests/write_api.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/9acf4eb9cc9e1d1a9fb289609ca84359253a9e13

crates/iota-indexer/tests/rpc-tests/read_api.rs
```diff
@@ -2520,3 +2521,70 @@ fn find_transaction_for_create_and_wrap_same_ptb() -> Result<(), anyhow::Error>
...
+            "indexer_grpc_is_transaction_present",
+            Some(Box::new(move |builder| {
+                builder.with_fullnode_enable_grpc_api(true)
+            })),
+            None,
+        )
+        .await;
+
+        // give some time for the grpc server to start
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
tomxey: we could also use `indexer_wait_for_checkpoint(store, 1).await;` here probably

crates/iota-indexer/tests/rpc-tests/read_api.rs
```diff
@@ -2520,3 +2521,70 @@ fn find_transaction_for_create_and_wrap_same_ptb() -> Result<(), anyhow::Error>
...
+            "indexer_grpc_is_transaction_not_present",
+            Some(Box::new(move |builder| {
+                builder.with_fullnode_enable_grpc_api(true)
+            })),
+            None,
+        )
+        .await;
+
+        // give some time for the grpc server to start
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
tomxey: we could also use `indexer_wait_for_checkpoint(store, 1).await;` here probably

tomxey (approved):

crates/iota-indexer/tests/rpc-tests/read_api.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/5f62b2cff2a8ab0f0099206a578d7a3648ff8114

crates/iota-indexer/tests/rpc-tests/read_api.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/5f62b2cff2a8ab0f0099206a578d7a3648ff8114

crates/iota-indexer/tests/common/mod.rs
```diff
@@ -132,10 +132,11 @@ impl SimulacrumTestSetup {
 /// Start a [`TestCluster`][`test_cluster::TestCluster`] with a `Read` &
 /// `Write` indexer. Set `epochs_to_keep` (> 0) to enable indexer pruning.
 pub async fn start_test_cluster_with_read_write_indexer(
-    database_name: Option<&str>,
+    database_name: impl Into<Option<&str>>,
     builder_modifier: Option<Box<dyn FnOnce(TestClusterBuilder) -> TestClusterBuilder>>,
     pruning_options: Option<PruningOptions>,
 ) -> (TestCluster, PgIndexerStore, HttpClient) {
+    let database_name = database_name.into();
     let mut builder = TestClusterBuilder::new();
```
kodemartin: Wouldn't  this simplify things?
```rust
    let mut builder = TestClusterBuilder::new().with_fullnode_enable_grpc_api(true);
```

crates/iota-indexer/src/optimistic_indexing.rs
```diff
@@ -42,34 +51,177 @@ use crate::{
...
+    async fn execute_transaction(&self, transaction: &Transaction) -> IndexerResult<Response> {
+        match self {
+            FullnodeRpc::Grpc(ref client) => {
+                let readmask = FieldMask::from_paths(EXECUTE_TRANSACTION_READ_MASK)
+                    .display()
+                    .to_string()
+                    .pipe(Some);
+
+                client
+                    .execute_transaction(transaction.clone().try_into()?, readmask.as_deref())
```
kodemartin: I trust there is a use for `pipe` to ergonomically pass the value to a funciton and return the result, but for such cases as `Ok` or `Some` I would argue it adds verbosity and indirection making code harder to read.

Here you `pipe(Some)` and later `readmask.as_deref()`. Both are implicit. I don't know the type of `readmask.as_deref()` unless I read the signature.

As opposed to not using `pipe` and passing `Some(readmask.as_str())` which would be explicit (type is `Option<&str>` and less verbose.

crates/iota-indexer/src/optimistic_indexing.rs
kodemartin, line 1: Nit: I really can't see a benefit using `.pipe(Ok)` as opposed to wrapping values `Ok(T)`. It adds verbosity and indirection in my opinion.

crates/iota-indexer/src/lib.rs
```diff
@@ -103,3 +108,46 @@ fn get_http_client(rpc_client_url: &str) -> Result<HttpClient, IndexerError> {
...
+
+/// Abstracts the fullnode RPC API.
+///
+/// The gRPC API is still in alpha but will be the default in the future.
+/// On the otherhand, the JSON RPC API is scheduled to be deprecated in the
+/// future.
+/// We need to maintain backward compatibility with the JSON RPC API for now
+/// until the gRPC API is fully compatible.
+#[derive(Clone)]
+pub enum FullnodeRpc {
```
kodemartin: Wouldn't it be easier to maintain (and clean in the future) to have all variants (including Rest) consolidated in a single type?

In addition scattering implementations of `FullnodeRpc` across files makes it harder to track and reason about possible extensions, refactoring and so on.

sergiupopescu199: That's true, I thought that I would avoid some verbosity in non handing the rest api where we only needed the json rpc. 

Since after rebase we have full support for write api through gRPC, I've removed the `FullnodeRpc` struct and refactored the code accordingly.

fixed: https://github.com/iotaledger/iota/pull/10355/commits/4c33dc4cdce91e7da7640761f3030c61e2d912a2

crates/iota-indexer/tests/common/mod.rs
sergiupopescu199: True, but other write api methods still rely on the JSON RPC (dry_run, dev_inspect) so some test will fail. But now that I'm thinking it's ok. 

fixed: https://github.com/iotaledger/iota/commit/9f02e1d894aea4ef137de4f1edb1ea1d3321aa99

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/commit/9f02e1d894aea4ef137de4f1edb1ea1d3321aa99

crates/iota-indexer/src/optimistic_indexing.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10355/commits/9f02e1d894aea4ef137de4f1edb1ea1d3321aa99

kodemartin (approved):
lgtm (some failing tests though)

PR #10365, opened 2026-02-18
feat(ci): improve e2e localnet test to support early abort, except in nightly mode

muXxer:
# Description of change

Right now the e2e localnet test runs every single step, no matter if the former steps already failed. This was intended behaviour to make sure that CI tests for single components can still run, even if some of the components are broken on develop.

That means on the other hand that tests continue for a long time, even thought a failure in the CI was already detected. The rest of our CI tests error out early in that case, to not waste CI resources.

Looking at the tests in detail, we realised that they are feature gated by the changed files anyways, so even if for example one of the tests would be broken, it might not even be executed because the file change filters didn't cause the test to be enabled in the first place. This means we can change the behaviour to abort early on failure.

For the nightly tests, where all tests are enabled regardless of the changes, we decided to "continue-on-error", so that all tests keep being executed, even if single steps in the chain fail.


Dr-Electron (approved):
Looks good

marc2332 (approved):

PR #10366, opened 2026-02-19
feat(iota-core): Persist and restore scorer received reports state across restarts                                                                                                                                 

oliviasaa:
# Description of change

- Persists `ReportAggregator` state to DB: adds a `received_reports_state` table (`DBMap<u32, DBReceivedReportsStatePerAuthority>`) to `AuthorityEpochTables`, so that aggregated misbehavior report data survives crashes
- Snapshots report state into `ConsensusCommitOutput`: after processing misbehavior reports in a consensus commit, the affected authorities' report state is written into the output, which gets persisted when the quarantine flushes the commit batch
- Restores `ReportAggregator` from DB on startup: `restore_from_tables` is called during `AuthorityPerEpochStore` initialization, recovering the last flushed report state. Unflushed commits are replayed during consensus resync, bringing the aggregator back to its pre-crash state
- Optimizes score updates: scores are only recalculated when at least one misbehavior report was seen in the commit, avoiding unnecessary work


bingyanglin (approved):

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -283,4 +283,11 @@ impl VersionedScoringMetrics {
             }
         }
     }
+
+    // Restores metrics from a persisted snapshot.
+    pub fn update_from_snapshot(&self, other: &VersionedScoringMetrics) {
+        for (current, new) in self.iter().flatten().zip(other.iter().flatten()) {
```
Thoralf-M: Should first check if the enum version match? Otherwise if V2 gets added this might behave unexpected

oliviasaa: sure! added the check in [c527e02](https://github.com/iotaledger/iota/pull/10366/commits/c527e024ca0549c902bf90ebe054ffe2c011b710)

Thoralf-M (approved):
Approving the change in iota-types

crates/iota-core/src/authority/scorer.rs
```diff
@@ -727,4 +825,349 @@ mod tests {
...
         assert_eq!(scores, vec![40142, 0, 0]);
     }
+
+    // Test simulating the full persist/restore cycle:
+    //
+    // 1. Build a test authority with epoch store
+    // 2. Create a MisbehaviorReport consensus transaction
+    // 3. Process it through `process_consensus_transactions_for_tests` (verify →
+    //    process_consensus_transaction → push to quarantine)
+    // 4. Verify the scorer was updated by
```
cyberphysic4l: incomplete sentence

crates/iota-core/src/authority/scorer.rs
```diff
@@ -727,4 +825,349 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        // Verify the scorer was updated.
+        assert_eq!(
+            epoch_store
+                .scorer
+                .received_metrics_snapshot(authority_index)
+                .load_faulty_blocks_provable(),
+            vec![3; committee_size],
```
cyberphysic4l: I know it's a bit tedious, but maybe the full report state should be checked rather than only the faulty blocks provable field. It's probably enough to just check the faulty blocks provable, but feels a bit of a shortcut that could have a weird edge case.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -727,4 +825,349 @@ mod tests {
...
+            let tables = epoch_store.tables().unwrap();
+            let stored_state = tables.received_reports_state.get(&()).unwrap();
+            assert!(
+                stored_state.is_some(),
+                "received_reports_state should be persisted to DB"
+            );
+            let stored_state = stored_state.unwrap();
+            assert_eq!(
+                stored_state[authority_index as usize]
+                    .received_metrics
```
cyberphysic4l: Same here re full report vs faulty blocks provable field

crates/iota-core/src/authority/scorer.rs
```diff
@@ -727,4 +825,349 @@ mod tests {
...
+            )
+            .await
+            .unwrap();
+
+        // Scorer now reflects both reports, but DB only has commit 1.
+        assert_eq!(
+            epoch_store
+                .scorer
+                .received_metrics_snapshot(authority_index)
+                .load_faulty_blocks_provable(),
```
cyberphysic4l: Same comment about full report check.

cyberphysic4l (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1160,6 +1164,11 @@ impl AuthorityPerEpochStore {
 
         let voting_power = committee.members().map(|(_, v)| *v).collect::<Vec<u64>>();
 
+        let scorer = Arc::new(Scorer::new(voting_power, &protocol_config));
+        scorer
+            .restore_from_tables(&tables)
+            .expect("Failed to restore scorer state from tables");
```
piotrm50: wrong expect message. 
Perhaps we should replace .expect() with ? to be in line with other methods that read the tables. In practice it will still panic but at a higher level, so the caller can decide what to do on error.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -160,10 +222,55 @@ impl Scorer {
...
+                has_not_sent_report: AtomicBool::new(state.has_not_sent_report_value()),
+                invalid_reports_count: AtomicU64::new(state.invalid_reports_count_value()),
+            })
+            .collect()
+    }
+
+    pub(crate) fn restore_from_tables(&self, tables: &AuthorityEpochTables) -> IotaResult<()> {
+        if let Some(persisted_state) = tables.received_reports_state.get(&())? {
+            let received_reports_state_in_scorer = &self.received_reports_state;
+            if received_reports_state_in_scorer.len() != persisted_state.len() {
```
piotrm50: That's quite bad, isn't it? Should we just warn a log and continue processing? This can cause a fork on the node at the epoch switch (or during epoch after we add it to the prologue), so maybe a panic  or at least an error would be justified? Later `.zip` truncates to the shorted collection so we end up with incorrect state either way.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -737,4 +825,349 @@ mod tests {
...
+            vec![10; committee_size],
+            vec![2; committee_size],
+        ));
+        let report_2 = VersionedMisbehaviorReport::new_v1(MisbehaviorsV1::new(
+            vec![7; committee_size],
+            vec![4; committee_size],
+            vec![15; committee_size],
+            vec![5; committee_size],
+        ));
+        let consensus_tx_1 =
```
piotrm50: nit: would be even more interesting to see reports from different authorities to make sure that the first report is loaded correctly, while the second one from another validator updates the score after it's replayed.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -257,4 +283,15 @@ impl VersionedScoringMetrics {
             }
         }
     }
+
+    // Restores metrics from a persisted snapshot.
+    pub fn update_from_snapshot(&self, other: &VersionedScoringMetrics) {
+        match (self, other) {
```
piotrm50: I would add a warn log if the the types don't match here for future updates. Otherwise it  doesn't do anything without leaving any trail.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2977,10 +2986,17 @@ impl AuthorityPerEpochStore {
...
+        // Track whether any misbehavior report was seen (valid or invalid) so we
+        // only snapshot the scorer state when needed.
+        let mut misbehavior_report_seen = false;
+
         // Split transactions into different types for processing.
         let verified_transactions: Vec<_> = transactions
             .into_iter()
             .filter_map(|transaction| {
+                if transaction.is_misbehavior_report() {
+                    misbehavior_report_seen = true;
```
piotrm50: Should this be set to true BEFORE calling verify_consensus_transaction? What if verification fails. Perhaps we could perform this check in the next loop ` for tx in verified_transactions {` as it seems to fit there more.

crates/iota-core/src/authority/scorer.rs
```diff
@@ -8,34 +8,95 @@ use std::sync::{
...
+#[derive(Debug)]
+pub struct ReceivedReportsStatePerAuthority {
+    // The metrics counts received from the authority, i.e., the information contained in the
+    // MisbehaviourReports received. If the authority has not sent a report, received_metrics will
+    // be all zeroed.
+    received_metrics: VersionedScoringMetrics,
+    // Indicates whether the authority did not send any misbehavior reports in the epoch. We use
+    // this to differentiate an authority that did not send a report from another one who sent
+    // zeroed reports.
+    has_not_sent_report: AtomicBool,
```
piotrm50: `has_not_sent_report` is a bit weird and harder to understand than if it was just `sent_report`. Especially given that there is the following check `!state.has_not_sent_report_value()`  that could be made much simpler `state.sent_report()`.

crates/iota-types/src/scoring_metrics.rs
```diff
@@ -257,4 +283,15 @@ impl VersionedScoringMetrics {
...
+            (VersionedScoringMetrics::V1(current), VersionedScoringMetrics::V1(snapshot)) => {
+                for (current, snapshot) in current.iter().flatten().zip(snapshot.iter().flatten()) {
```
piotrm50: Should we at least make sure that lengths are identical or is that guaranteed somehow?

piotrm50 (changes requested):

oliviasaa:
This branch was rebased on top of the new PR for the Scorer refactoring, so it uses the types compatible with future versions and the Starfish implementation. The logic is still the same, but the PR was simplified, since it does not define any new types.

piotrm50:
Closing in favor of #11561, which lands the persistence portion of this PR rebased on the post-Starfish codebase.

Mapping of scope:

| Original bullet | Where it lives now |
|---|---|
| `received_reports_state` table on `AuthorityEpochTables` | #11561 (Task 2) |
| Snapshots into `ConsensusCommitOutput` on commit | #11561 (Tasks 4-6) — captured at \`process_report\` time, not at flush time, to preserve cross-validator determinism around \`scoreboard.update_scores\` |
| \`restore_from_tables\` on startup | #11561 (Tasks 3, 7) |
| Score-update optimization (recalc only when ≥1 report seen) | Tracked in https://github.com/iotaledger/iota-private/issues/408 |

The rest of this PR's diff (new \`misbehavior_monitor.rs\`, \`misbehavior_config.rs\`, \`report_aggregator.rs\`, \`scorer.rs\` rewrite, \`scoring_metrics.rs\` deletion, \`consensus_output_api.rs\` changes) has already landed on \`develop\` piecemeal via #10088, #10779, #11531, #11534 and predecessors, so a rebase wasn't tractable. Thanks @oliviasaa for the original design — the persistence layout in #11561 matches yours.

PR #10379, opened 2026-02-19
feat(iota-core, iota-types): Implement a PoC version of post-consensus conflict resolution

### LINKED ISSUE #10017: pcool: Implement a PoC version of white-flag conflict resolution
- Add transaction driver 
- Implement API endpoints to handle whiteflag-flow transaction 
- Add new consensus type
- Implement conflict resolution
- etc..

piotrm50:
# Description of change

This PR merges the `feat/consensus/transaction-certificate-removal` branch into `feat/consensus/transation-certificateless-flow-feature`, implementing an early Proof of Concept (PoC) for a certificateless transaction flow in IOTA consensus.

## ⚠️ Important: Early PoC Status
**This is an early proof of concept with many missing features and limited testing.** This implementation demonstrates base functionality but is not production-ready.
### What Works ✅
- ✅ Sending transactions from faucet
- ✅ Sending transactions between nodes
- ✅ Basic white flag conflict resolution for post-consensus owned object locking
### What's Not Yet Tested ⚠️
- ⚠️ Conflicting transactions scenarios
- ⚠️ High-load/stress testing
- ⚠️ Full end-to-end validator scenarios
- ⚠️ Edge cases and error recovery paths
## Summary
This PR introduces a **white flag transaction flow** that bypasses pre-consensus certification and owned object locking. Instead, conflicts are resolved deterministically post-consensus using persistent locks. This represents a significant architectural shift in how IOTA handles transaction consensus.
### Key Design Principles
1. **Transaction Exclusivity**: A transaction uses either white-flag flow OR certificate flow, never both (enforced by feature flags)
2. **Deterministic Conflict Resolution**: Conflicts are resolved in consensus order
3. **Persistent Lock Reuse**: Uses the same lock mechanism (`owned_object_locked_transactions` table) as certificate flow
4. **Atomic Conflict Detection**: If ANY owned input object is locked, the ENTIRE transaction is dropped
### New Components
1. **White Flag Conflict Resolution** (`white_flag.rs`)
   - Post-consensus owned object locking and conflict resolution
   - Deterministic transaction ordering based on consensus sequence
   - Atomic conflict detection (all-or-nothing for owned objects)
2. **Transaction Driver** (`transaction_driver/` module)
   - `EffectsCertifier`: Manages effects certificate creation and validation
   - `TransactionSubmitter`: Handles transaction submission to validators
   - `RequestRetrier`: Retry logic with exponential backoff
   - `ReconfigObserver`: Monitors and handles epoch reconfigurations
3. **Validator Client Monitor** (`validator_client_monitor/` module)
   - Tracks validator client performance and availability
   - Collects metrics on submission success/failure rates
   - Provides feedback for validator selection
4. **Status Aggregator** (`status_aggregator.rs`)
   - Aggregates transaction status from multiple validators
   - Determines quorum for transaction effects
### Enhanced Components
1. **Authority Per Epoch Store** (`authority_per_epoch_store.rs`)
   - Enhanced lock management for white flag transactions
   - Support for batch lock acquisition and release
   - Conflict tracking and resolution
2. **Consensus Handler** (`consensus_handler.rs`)
   - Integration with white flag conflict resolution
   - Pre-processing of consensus transactions
   - Support for UserTransactionV1 variant
3. **Authority Server** (`authority_server.rs`)
   - New gRPC endpoints for white flag submission 
   - Transaction status queries
   - Enhanced validator communication
4. **Transaction Orchestrator** (`transaction_orchestrator.rs`)
   - Support for certificateless transaction submission
   - Integration with TransactionDriver
   - Improved pending transaction recovery
### Protocol & Configuration Changes
1. **Protocol Config** (`protocol-config`)
   - New feature flag: `white_flag_transaction_flow`
   - Environment variable overrides for feature flags
   - Configuration versioning support
2. **Node Configuration** (`node.rs`)
   - `ValidatorClientMonitorConfig`: Client monitoring settings
   - Feature flag configurations
   - Network endpoint configurations
### Type System Enhancements
1. **New Types** (`iota-types`)
   - `SubmitTxRequest`, `SubmitTxResult`: White flag submission types
   - `TransactionDriverTypes`: Effects and finalization types
   - Enhanced `ConsensusTransactionKind` with `UserTransactionV1`
2. **Error Handling** (`error.rs`, `transaction_driver/error.rs`)
   - New error variants for white flag flow
   - Transaction driver specific errors
   - Enhanced error context and debugging
### Testing
1. **White Flag Tests** (`white_flag_tests.rs`)
   - Conflict resolution scenarios
   - Lock acquisition and release
   - Transaction ordering validation
2. **Unit Tests** (`authority_tests.rs`, `server_tests.rs`)
   - gRPC endpoint testing
   - Transaction submission validation
   - Status aggregation verification
3. **Validator Client Monitor Tests** (`validator_client_monitor/tests.rs`)
   - Client performance tracking
   - Metrics collection validation
### Metrics & Monitoring
- New Prometheus metrics for:
  - Transaction driver operations
  - White flag conflict resolution
  - Validator client performance
  - Starfish DAG state and shard reconstruction
- [x] Patch-specific tests (white flag conflict resolution, transaction driver)
- [x] Manual testing with local validator setup
  - Transaction submission from faucet
  - Transaction submission between nodes
- [ ] Conflicting transaction scenarios (not yet tested)
- [ ] Production-scale stress testing (not yet performed)
- [ ] Full integration testing across all validator scenarios (pending)

Resolves #10017 


crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -4301,6 +4347,74 @@ impl AuthorityPerEpochStore {
...
+                );
+
+                // Handle scheduling result
+                match scheduling_result {
+                    SchedulingResult::Defer(deferral_key, _deferral_reason) => {
+                        // TODO: Simplified: defer without gas price feedback for PoC
+                        Ok(ConsensusCertificateResult::Deferred {
+                            deferral_key,
+                            suggested_gas_price: None,
+                        })
```
roman1e2f5p8s, line 4392: FYI: this will prevent closing the epoch since deferrals will never be drained (no cancellations here).

But I already addressed this on my epoch-related branch: https://github.com/iotaledger/iota/blob/84fd50c11861edcd52ed93c1113952a644c3ccf5/crates/iota-core/src/authority/authority_per_epoch_store.rs#L4305.

crates/iota-types/src/messages_grpc.rs
```diff
@@ -284,3 +303,99 @@ pub struct HandleCapabilityNotificationResponseV1 {
...
+/// The result of submitting a transaction to a validator.
+#[derive(Clone, Serialize, Deserialize)]
+pub enum SubmitTxResult {
+    /// The transaction was submitted to consensus.
+    Submitted,
+    /// The transaction has already been executed (finalized).
+    Executed {
+        effects_digest: TransactionEffectsDigest,
+        /// Response should always include details for executed transactions.
+        details: Option<Box<ExecutedData>>,
```
Thoralf-M, line 357: If it should, in what case would it not be included? Or why is it optional?

kodemartin (approved):
lgtm

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: Good catch! If it's addressed in your PR then let's not create more conflicts here.

crates/iota-core/src/white_flag.rs
```diff
@@ -0,0 +1,246 @@
...
+            });
+        }
+    };
+
+    let input_objects = transaction_data.input_objects()?;
+
+    let owned_objects: Vec<ObjectRef> = input_objects
+        .into_iter()
+        .filter_map(|input| match input {
+            InputObjectKind::ImmOrOwnedMoveObject(obj_ref) => {
```
bingyanglin, line 234: This treats immutable objects the same as owned objects for locking. Two transactions that both reference the same published package (or any popular immutable object) will falsely conflict and the second one gets dropped. Even for a PoC this makes the feature unusable for real workloads, since almost every transaction touches at least one shared package. Maybe look up the object's ownership from the cache and skip anything that's `Immutable`? Or do I miss something?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3098,6 +3117,32 @@ impl AuthorityPerEpochStore {
...
+        if self.protocol_config.enable_white_flag_flow() {
+            let (dropped, owned_object_locks) =
+                white_flag::resolve_owned_object_conflicts(self, &mut sequenced_transactions)?;
+            output.set_owned_object_locks(owned_object_locks);
+            // TODO: add white_flag_dropped_transactions metric to AuthorityMetrics
+            //  authority_metrics.white_flag_dropped_transactions.inc_by(dropped.len() as
+            //  u64);
+            // TODO: possibly record dropped digests in ConsensusCommitPrologue for
+            //  consistent view
+            if !dropped.is_empty() {
```
bingyanglin, line 3137: `resolve_owned_object_conflicts` only runs on `sequenced_transactions`, but `sequenced_randomness_transactions` is a separate vector that never goes through white flag. If a `UserTransactionV1` uses randomness and has owned object conflicts, the conflict won't be caught. Maybe run white flag resolution on `sequenced_randomness_transactions` too, or reject `UserTransactionV1` that uses randomness until this is handled?

crates/iota-core/src/white_flag.rs
```diff
@@ -0,0 +1,246 @@
...
+/// INVARIANT: A transaction is issued through either white-flag OR cert flow,
+/// never both. This is enforced by the feature flag and submission endpoint.
+pub fn resolve_owned_object_conflicts(
+    epoch_store: &AuthorityPerEpochStore,
+    transactions: &mut Vec<VerifiedSequencedConsensusTransaction>,
+) -> IotaResult<(Vec<TransactionDigest>, HashMap<ObjectRef, LockDetails>)> {
+    let mut dropped = Vec::new();
+    let mut current_commit_locks: HashMap<ObjectRef, LockDetails> = HashMap::new();
+
+    transactions.retain(|tx| {
```
bingyanglin, line 78: The `retain` closure does DB reads that can fail, but `retain` only returns `bool`. There's no way to propagate the error. Right now a transient disk failure just drops the transaction and logs a warning, and the outer function always returns `Ok`. That means valid transactions can be silently lost on infra hiccups with no signal to the caller. Maybe restructure this as a manual loop that collects into a new `Vec` so you can use `?` on the DB calls.

crates/iota-core/src/authority_server.rs
```diff
@@ -1222,6 +1255,235 @@ impl ValidatorService {
...
+        // Handle ping requests
+        if wait_request.transaction_digest.is_none() || wait_request.ping_type {
+            // For ping, return a dummy response with zeroed digest
+            return Ok((
+                tonic::Response::new(WaitForEffectsResponse::Executed {
+                    effects_digest: TransactionEffectsDigest::ZERO,
+                    details: None,
+                }),
+                Weight::one(),
+            ));
```
bingyanglin, line 1407: `notify_read_executed_effects_digests` blocks until the transaction is executed, with no timeout. If the transaction was dropped by white flag resolution or the digest doesn't exist, this blocks forever. The outer gRPC handler uses `spawn_monitored_task!` which detaches the task, and dropping the `JoinHandle` on client disconnect doesn't cancel it. Maybe add a server-side timeout and return `Expired` when it fires?

crates/iota-types/src/messages_grpc.rs
piotrm50: Good point! It's always returned in full if the transaction was executed before. https://github.com/iotaledger/iota/issues/10761

crates/iota-core/src/authority_server.rs
piotrm50: This is already fixed exactly like you describe in subsequent PRs.

crates/iota-core/src/white_flag.rs
piotrm50: Good catch! I have to apply this fix (if still relevant) in a subsequent PR because logic from this file is moved elsewhere and refactored to also perform post-consensus validation. https://github.com/iotaledger/iota/issues/10763

crates/iota-core/src/white_flag.rs
piotrm50: That's actually a good point. After giving it some thought I think that this actually requires a bit more analysis about the determinism of this check. I'll create an issue to track this.  https://github.com/iotaledger/iota/issues/10762

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: Yes, this is a bug. I'll create an issue to fix that in a subsequent PR to avoid rebase conflicts.

piotrm50: https://github.com/iotaledger/iota/issues/10764

crates/iota-types/src/messages_consensus.rs
```diff
@@ -609,6 +620,17 @@ impl ConsensusTransaction {
         }
     }
 
+    pub fn new_user_transaction(transaction: Transaction) -> Self {
+        let mut hasher = DefaultHasher::new();
+        let tx_digest = transaction.digest();
+        tx_digest.hash(&mut hasher);
+        let tracking_id = hasher.finish().to_le_bytes();
```
Thoralf-M, line 627: Why hash it and don't just use the digest directly? I know it's also done for the other things above, just wondering what's the reason for this as it requires more computation for no obvious benefit

crates/iota-core/src/authority_server.rs
```diff
@@ -1222,6 +1255,235 @@ impl ValidatorService {
...
+            IotaError::UnsupportedFeature {
+                error: "White flag flow is not enabled in this protocol version".to_string()
+            }
+            .into()
+        );
+
+        let submit_request = request.into_inner();
+
+        // Handle ping requests
+        if submit_request.transaction.is_none() {
```
miker83z, line 1283: Maybe add a flag in the `SubmitTxRequest` to make the ping explicit? I mean instead of checking for `transaction.is_none()`.

crates/iota-core/src/authority_server.rs
```diff
@@ -1222,6 +1255,235 @@ impl ValidatorService {
...
+                        effects_digest,
+                        details: None, // TODO: optionally include details based on request flag
+                    }],
+                }),
+                Weight::zero(),
+            ));
+        }
+
+        // 6. Verify user signature
+        let _verified = match epoch_store.verify_transaction(transaction.clone()) {
```
miker83z, line 1341: Just a comment to remember. With this function `verify_transaction`, since we call it from the `epoch_store`, we don't have access to possible objects used by a MoveAuthenticator. So, for a TX carrying a MoveAuthenticator, the signature check performed here will always pass without checks. It will/should be checked later, when it is possible to spawn a VM and load objects.

crates/iota-core/src/authority_server.rs
```diff
@@ -1222,6 +1255,235 @@ impl ValidatorService {
...
+    async fn wait_for_effects_impl(
+        &self,
+        request: tonic::Request<iota_types::messages_grpc::WaitForEffectsRequest>,
+    ) -> WrappedServiceResponse<iota_types::messages_grpc::WaitForEffectsResponse> {
+        use iota_types::messages_grpc::WaitForEffectsResponse;
+
+        let wait_request = request.into_inner();
+
+        // Handle ping requests
+        if wait_request.transaction_digest.is_none() || wait_request.ping_type {
```
miker83z, line 1399: Same logic as https://github.com/iotaledger/iota/pull/10379/changes#r2952892012
Here it seems we already have `ping_type`, so maybe it could be enough? Following that logic, should we then treat `transaction_digest.is_none()` as error if `!wait_request.ping_type`?

crates/iota-core/src/authority_server.rs
piotrm50: Sure, that's just a first draft of how this can be done. This whole ping feature will be re-designed and re-implemented in the future.

crates/iota-core/src/authority_server.rs
piotrm50: This will be re-designed and reimplemented.

crates/iota-types/src/messages_consensus.rs
piotrm50: nice outside of the box thining :D I don't know to be honest. Perhaps we could just truncate the digest and use it as tracking_id, but that's outside of the scope of this PR.

crates/iota-core/src/authority_server.rs
piotrm50: This is already covered on a subsequent PR: https://github.com/iotaledger/iota/blob/5454115b62d769ccf8aff55bf87be0a2ded9a783/crates/iota-core/src/authority.rs#L968

miker83z (approved):

crates/iota/src/keytool.rs
thibault-martinez, line 1: Why is fmt of this file changing?

piotrm50: no clue, perhaps I ran it without +nightly flag or something. Anyway I already reverted it. I'm sorry about it.

thibault-martinez: No worries at all, thanks

bingyanglin (approved):

filipdulic (approved):

PR #10380, opened 2026-02-19
feat(core): include rest index in db snapshots

muXxer:
# Description of change

Right now the `RestIndexStore` is not included in the database snapshots. When `iota-tool download-db-snapshot --latest    --network testnet --path ./db --num-parallel-downloads 70 --no-sign-request --verbose` is executed, the `rest_index` database doesn't exist.

This PR adds db snapshot creation for the `RestIndexStore` on the node side and extends the `download-db-snapshot` in `iota-tool` to download the new database.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Include `RestIndexStore` in database snapshots
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/iota-types/src/storage/read_store.rs
+++ b/crates/iota-types/src/storage/read_store.rs
@@ -880,22 +880,27 @@ pub trait RestStateReader: ObjectStore + ReadStore + Send + Sync {
 pub type DynamicFieldIteratorItem =
     Result<(DynamicFieldKey, DynamicFieldIndexInfo), TypedStoreError>;
 pub trait RestIndexes: Send + Sync {
+    // only used in "grpc-server"
     fn get_epoch_info(&self, epoch: EpochId) -> Result<Option<EpochInfo>>;
 
+    // used in both "grpc-server" and "rest-api"
     fn get_transaction_info(&self, digest: &TransactionDigest) -> Result<Option<TransactionInfo>>;
 
+    // only used in "rest-api"
     fn account_owned_objects_info_iter(
         &self,
         owner: IotaAddress,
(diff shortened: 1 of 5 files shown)


semenov-vladyslav (approved):

crates/iota-types/src/storage/read_store.rs
```diff
@@ -880,22 +880,27 @@ pub trait RestStateReader: ObjectStore + ReadStore + Send + Sync {
 pub type DynamicFieldIteratorItem =
     Result<(DynamicFieldKey, DynamicFieldIndexInfo), TypedStoreError>;
 pub trait RestIndexes: Send + Sync {
+    // only used in "grpc-server"
```
Thoralf-M, line 883: Just wondering, have you considered splitting the trait in two, given that only half of the methods are used for a specific part?

muXxer: For now we are reusing the rest indexes because we don't want to duplicate the big database tables for transaction info in a rest index, and another grpc index. The rest index will be renamed to grpc index at some point and the unused tables will be dropped, when we replace the rest api completetly with grpc.

Thoralf-M (approved):

alexsporn (approved):

muXxer:
We did some tests:
- database snapshot was successfully created and uploaded
- database snapshot was successfully downloaded via `iota-tool` and the node didn't re-index after startup 👍

PR #10383, opened 2026-02-19
fix(iota_types): `iota_sdk_types::Object` to `iota_types::Object` type conversion

### LINKED ISSUE #10382: [iota_types]: `TryFrom<iota_sdk_types::Object> for iota_types::Object` loses `storage_rebate` causing digest mismatch
Currently, when converting an `iota_sdk_types::Object` to `iota_types::Object` in `iota_types::iota_sdk_types_conversions`, the `storage_rebate` field is lost because the implementation uses `Object::new_from_genesis()`, which hardcodes `storage_rebate` to `0`.

```rust
// Current implementation (incorrect)
// storage_rebate is always set to 0
Self::new_from_genesis(
    value.data.try_into()?,
    value.owner.into(),
    value.previous_transaction.into(),
) 
```

This causes a **digest mismatch** because `Object::digest()` is computed on-the-fly by hashing the full `ObjectInner` via BCS — which includes `storage_rebate`. As a result, the same logical object will have different digests before and after the SDK → core type conversion.

sergiupopescu199:
# Description of change

This patch fixes a conversion issue from `iota_sdk_types::Object` to `iota_types::Object` in `iota_types::iota_sdk_types_conversions`.
Before with `new_from_genesis()` the storage_rebate was hardcoded to 0 which can be wrong.

fixes #10382 

### DIFF
--- a/crates/iota-types/src/iota_sdk_types_conversions.rs
+++ b/crates/iota-types/src/iota_sdk_types_conversions.rs
@@ -49,7 +49,7 @@ use iota_sdk_types::{
 use move_core_types::language_storage::ModuleId;
 use tap::Pipe;
 
-use crate::transaction::TransactionDataAPI as _;
+use crate::{object::ObjectInner, transaction::TransactionDataAPI as _};
 
 #[derive(Debug)]
 pub struct SdkTypeConversionError(pub String);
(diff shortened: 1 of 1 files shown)


Thoralf-M (approved):

muXxer (approved):
The change looks correct to me, but I'm not sure about the implications on the whole codebase. Maybe dev-tools can have a second look?

alexsporn (approved):

DaughterOfMars (approved):

PR #10388, opened 2026-02-19
fix(wallet): Reprompt ledger connection if there is an address missmatch

### LINKED ISSUE #10386: Fix unable to use multiple ledger devices in the wallet
Reported in discord 

> I have a ledger nano x and ledger stax. I cannot use both device interchangeably. If i reset the wallet at import i can select the hid device and then I can only use that device. if i swap to another device the app doesnt request to connect to the new device. in case of metamask this is possible. I think we need to allow requestDevice() if the known device is not found during trasactions signing or allow change of hid device without reset.  For now I will bypass the issue by using 2 chrome profiles. Is there a easier fix? If not it would be nice to add multiple known HID device.

https://discord.com/channels/912718559954747442/1282249218039480434/1472207264063422677

marc2332:
Closes https://github.com/iotaledger/iota/issues/10386


KeitoTadashi (approved):
Great work!!

Image

VmMad (approved):

PR #10392, opened 2026-02-19
feat(wallet): improve `element copied` and `external link opened` amplitude events 

evavirseda:
- remove unnecessary plugins
- remove `value`and `visibility` properties


panteleymonchuk (dismissed):

Juligs (dismissed):

VmMad (approved):

apps/wallet/src/ui/app/components/external-link/index.tsx
```diff
@@ -23,17 +22,12 @@ export function ExternalLink({
     title,
     onClick,
     type,
-    isPublic = false,
     trackEvent = true,
 }: ExternalLinkProps) {
     const handleClick = () => {
         if (trackEvent && type) {
```
brancoder, line 28: we could improve this if clause because it is a bit miss leading because `trackEvent` does not actually control the tracking of the event but rather `trackEvent` and `type` together do, so its a hidden condition. 

ex. https://github.com/iotaledger/iota/blob/develop/apps/wallet/src/ui/app/components/iota-apps/IotaApp.tsx#L129-L139

we have these cases, in the codebase, where we set the `trackEvent = false` but actually just omitting `type="application"` would accomplish the same thing. right?

apps/wallet/src/ui/app/hooks/useCopyToClipboard.ts
```diff
@@ -10,12 +10,11 @@ export type CopyOptions = {
     copySuccessMessage?: string;
     textType: string;
     trackEvent?: boolean;
-    isPublic?: boolean;
 };
 
 export function useCopyToClipboard(
```
brancoder, line 15: can we remove `useCopyToClipboard` wallet hook and use the `useCopyToClipboard` from core?

apps/wallet/src/ui/app/pages/home/interstitial/index.tsx
```diff
@@ -61,7 +61,7 @@ export function Interstitial({
                         }}
                         className="h-full w-full"
                         type="application"
-                        isPublic
+                        trackEvent={false}
```
brancoder, line 63: related to prevous comment about  silent condition. ex. here just omitting `type="application"` would do the same thing as setting `trackEvent={false}`

brancoder (changes requested):

brancoder (approved):
LGTM

panteleymonchuk (approved):

apps/wallet/src/ui/app/components/accounts/ProtectAccountForm.tsx
```diff
@@ -134,14 +135,13 @@ export function ProtectAccountForm({
                                 <span className="text-label-lg text-iota-neutral-40 dark:text-iota-neutral-60">
                                     I read and agreed to the
                                 </span>
-                                <a
+                                <ExternalLink
                                     href={ToS_LINK}
                                     className="text-label-lg text-iota-primary-30 dark:text-iota-primary-80"
-                                    target="_blank"
-                                    rel="noopener noreferrer"
+                                    type="documentation"
```
begonaalvarezd:
```suggestion
                                    type="legal"
```

apps/wallet/src/ui/app/pages/home/tokens/SupplyIncreaseVestingStakingDialog.tsx
```diff
@@ -26,8 +26,7 @@ export function SupplyIncreaseVestingStakingDialog({
 
     function navigateToDashboard() {
         ampli.externalLinkOpened({
```
begonaalvarezd: should we maybe use here the dedicated event we have for applications?

apps/wallet/src/ui/app/pages/home/tokens/MigrationDialog.tsx
```diff
@@ -23,8 +23,7 @@ export function MigrationDialog({ open, setOpen }: MigrationDialogProps) {
 
     function navigateToDashboard() {
         ampli.externalLinkOpened({
-            value: WALLET_DASHBOARD_URL,
-            type: 'application',
+            type: 'wallet dashboard application',
```
begonaalvarezd: should we maybe use here the dedicated event we have for applications?

PR #10397, opened 2026-02-20
feat(wallet): add dapp connect started amplitude event

### LINKED ISSUE #10190: Add dapp connect started amplitude event
Add dapp connect started amplitude event.
This event should fire before approving or rejecting the connection, when the user sees the connection screen.

Props
- `applicationName`: `string`
- `applicationUrl`: `string`

panteleymonchuk:
Closes #10190


apps/wallet/src/ui/app/pages/site-connect/index.tsx
```diff
@@ -62,15 +70,21 @@ export function SiteConnectPage() {
...
-                    applicationName: permissionRequest.name,
+                    applicationName: resolvedName,
                     applicationUrl: permissionRequest.origin,
                     approvedConnection: allowed,
                 });
                 handleOnFinish();
             }
         },
-        [requestID, accountsToConnect, permissionRequest, dispatch],
+        [requestID, accountsToConnect, permissionRequest, dispatch, ecosystemApps],
```
begonaalvarezd: im a bit hesitant on firing the event based on an apps backend fetch, i dont think it is worth depending on this

apps/wallet/src/ui/app/pages/site-connect/index.tsx
```diff
@@ -98,6 +112,22 @@ export function SiteConnectPage() {
...
+                permissionRequest.origin,
+                permissionRequest.pagelink,
+                ecosystemApps,
+            );
+            ampli.dappConnectStarted({
+                applicationName: resolvedName,
+                applicationUrl: permissionRequest.origin,
+            });
+        }
+    }, [permissionRequest, requestID, ecosystemApps]);
```
begonaalvarezd: same here, im not sure we want this depending on apps backend,i think it is just adding a complexity we dont need and it is not worth it

apps/wallet/src/ui/app/pages/site-connect/index.tsx
```diff
@@ -98,6 +112,22 @@ export function SiteConnectPage() {
         [handleOnSubmit],
     );
 
+    useEffect(() => {
+        if (permissionRequest && requestID && !trackedRequestIDs.current.has(requestID)) {
+            trackedRequestIDs.current.add(requestID);
```
begonaalvarezd: we dont need to track request ids

apps/wallet/src/ui/app/pages/site-connect/index.tsx
begonaalvarezd: i know you had the logic to resolve the app name based on apps backend but imo we should remove that logic and just use the provided information by the request

begonaalvarezd (changes requested):

begonaalvarezd (approved):

brancoder (approved):

PR #10406, opened 2026-02-20
feat(consensus): remove block digest from transaction ref

### LINKED ISSUE #9821: fastsync: Remove BlockDigest from TransactionRef struct

VorobyevIlya:
# Description of change

Remove block_digest from TransactionRef, completing the decoupling of transaction references from block header digests. This is a step toward fully independent transaction addressing in the consensus protocol.

 Key changes:

  - Remove block_digest field from TransactionRef
  - Add block_digest field to VerifiedTransactions and ShardMessage — as an Option<BlockHeaderDigest>, set to Some when using BlockRef-based fetching, None when using TransactionRef-based fetching
  - Add tx_ref_to_block_digest_by_authority lookup table to DagState — maps (round, transactions_commitment) to BlockHeaderDigest per authority, enabling resolution of TransactionRefs to BlockRefs when needed

Fixes #9821 

### DIFF
--- a/crates/starfish/core/src/error.rs
+++ b/crates/starfish/core/src/error.rs
@@ -110,7 +110,7 @@ pub(crate) enum ConsensusError {
         peer: AuthorityIndex,
     },
     #[error(
-        "After reconstruction, the transaction commitment does not match the commitment in {}",
+        "After reconstruction, the transaction commitment does not match the commitment in transaction ref {}",
         transaction_ref
     )]
     TransactionCommitmentMismatch { transaction_ref: TransactionRef },
--- a/crates/starfish/core/src/error.rs
+++ b/crates/starfish/core/src/error.rs
@@ -11,7 +11,7 @@ use typed_store::TypedStoreError;
 use crate::{
     block_header::{BlockRef, GENESIS_ROUND, Round},
     commit::{Commit, CommitIndex},
-    transaction_ref::{GenericTransactionRef, GenericTransactionRefAPI as _},
+    transaction_ref::{GenericTransactionRef, GenericTransactionRefAPI as _, TransactionRef},
 };
 
 /// Errors that can occur when processing blocks, reading from storage, or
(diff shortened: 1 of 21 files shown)


crates/starfish/core/src/storage/rocksdb_store.rs
```diff
@@ -252,7 +244,9 @@ impl Store for RocksDBStore {
                             (
                                 transaction_ref.round,
                                 transaction_ref.author,
-                                transaction_ref.block_digest,
+                                transaction.block_digest().expect(
```
polinikita, line 247: if consensus_transaction_ref=false,

Then what do we store in db? Are we expecting block digest to be available for given transaction ref or not? I am wondering about updating software in the epoch when the flag is off. Would it be possible to correctly load the transactions from db with the updated software?

crates/starfish/core/src/error.rs
```diff
@@ -110,10 +110,10 @@ pub(crate) enum ConsensusError {
         peer: AuthorityIndex,
     },
     #[error(
-        "After reconstruction, the transaction commitment does not match the commitment in the block {}",
-        block_ref
+        "After reconstruction, the transaction commitment does not match the commitment in {}",
```
polinikita: strange ending of phrase

VorobyevIlya: changed to "After reconstruction, the transaction commitment does not match the commitment in transaction ref {}",
        transaction_ref

crates/starfish/core/src/storage/rocksdb_store.rs
VorobyevIlya: discussed in a huddle

bingyanglin (approved):
Aprove `crates/iota-core/src/consensus_types/consensus_output_api.rs`.

nikitapolyanskii (approved):

polinikita (approved):

piotrm50 (approved):

PR #10409, opened 2026-02-20
feat(starfish): gate FastCommitSyncer behind protocol and local config flags

### LINKED ISSUE #9919: fastsync: Add a protocol feature flag and a local flag to enable/disable FastSync
Nodes should be able to turn off FastSync logic by changing a local parameter. Then the node will fallback to syncing with CommitSyncer. 

In case the flag is disabled and we detect in the databaes that FastSync is ogoing (can happen if node crashes while FastSync), panic with message, that a Conseuns and Authoirty databases need to be cleaned.

piotrm50:
# Description of change

Extend the existing `transaction_ref` / `consensus_transaction_ref` protocol feature flag by renaming it to `consensus_fast_commit_sync` and using it to also gate the new `FastCommitSyncer` and fast commit sync gRPC endpoints. Bump protocol version to 21.

**Protocol-level flag (`consensus_fast_commit_sync` in `iota-protocol-config`):**
- Renamed from `consensus_transaction_ref` to `consensus_fast_commit_sync` to reflect its expanded scope
- Enabled only on devnet at protocol version 21
- Already controls V2 block/commit format (transaction-ref based); now additionally gates:
  - `handle_fetch_commits` when called with `CommitSyncType::Fast`
  - `handle_fetch_commits_and_transactions`
  - `handle_fetch_transactions` when called with `TransactionFetchMode::FastCommitSync`
- All three endpoint checks return the new `ConsensusError::FastCommitSyncNotEnabled` and carry a TODO to remove the gate once the feature is enabled on all networks

**Local config flag (`enable_fast_commit_syncer` in starfish `Parameters`):**
- New local node parameter, defaults to `false`; operators can enable it once the protocol flag is active, or disable it again if bugs are discovered
- Does **not** affect endpoint availability — only controls whether this node creates and runs the `FastCommitSyncer`
- All test setups that set the protocol flag also update `context.parameters.enable_fast_commit_syncer` to match

**Authority node:**
- `fast_commit_syncer_handle` changed from `CommitSyncerHandle` to `Option<CommitSyncerHandle>`
- `FastCommitSyncer` is only created when **both** the protocol flag and the local config flag are enabled
- `stop()` uses `take()` to correctly move out of the `Option` before calling `stop()` which consumes by value

**Protocol version bump:**
- Protocol version bumped to 21
- `consensus_fast_commit_sync` enabled on devnet at v21
- OpenRPC max supported protocol version updated to 21

**Code cleanup:**
- All internal `transaction_ref_enabled` local variables renamed to `consensus_fast_commit_sync`
- All stale `consensus_transaction_ref` references in comments, error messages, and doc strings updated to `consensus_fast_commit_sync`
- `ShardWithProof::new` parameter renamed from `transaction_ref_enabled` to `consensus_fast_commit_sync`

**Other:**
- New `ConsensusError::FastCommitSyncNotEnabled` error variant
- Fix pre-existing typo `MissingVoringBlockHeaderInStorage` → `MissingVotingBlockHeaderInStorage`
- Updated parameters and protocol config snapshots

### Release Notes

- [x] Protocol: renamed `consensus_transaction_ref` to `consensus_fast_commit_sync` and extended it to also gate fast commit sync gRPC endpoints; enabled on devnet at protocol version 21
- [x] Nodes (Validators and Full nodes): added `enable_fast_commit_syncer` local Starfish consensus parameter (default `false`); when set to `true` and the protocol flag is also enabled, the node will run the `FastCommitSyncer` for faster recovery from large commit gaps
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


alexsporn:
This needs updating to at least protocol v21, since v20 is already part of a release.

crates/starfish/core/src/authority_node.rs
```diff
@@ -227,16 +227,29 @@ impl ConsensusAuthority {
...
-            dag_state.clone(),
-        )
-        .start();
+        // FastCommitSyncer is enabled when BOTH conditions are met:
+        // 1. Protocol-level flag is enabled (controls gRPC endpoint availability)
+        // 2. Local config flag is enabled (allows operators to disable locally)
+        // This two-level control allows the protocol to expose endpoints while allowing
+        // operators to disable the syncer locally for testing or bug mitigation.
+        let fast_commit_syncer_handle = if context.protocol_config.enable_fast_commit_sync()
+            && context.parameters.enable_fast_commit_syncer
```
bingyanglin: In the tests `test_fast_commit_syncer_on_restart` and `test_fast_commit_syncer_fail_on_unfinished_fast_sync_restart` we need to add `enable_fast_commit_syncer: true`? Otherwise the two tests will not be triggered anymore?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -434,6 +435,12 @@ struct FeatureFlags {
     // If true, then TransactionRef is used in commit instead of BlockRef
     #[serde(skip_serializing_if = "is_false")]
     consensus_transaction_ref: bool,
+
+    // If true, enables the fast commit syncer in Starfish consensus for faster recovery
+    // from large commit gaps. This also enables the associated gRPC endpoints for fetching
+    // commits and transactions.
+    #[serde(skip_serializing_if = "is_false")]
+    enable_fast_commit_sync: bool,
```
polinikita: better to use prefix consensus for this new flag.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2519,6 +2530,12 @@ impl ProtocolConfig {
                             .pass_calculated_validator_scores_to_advance_epoch = true;
                     }
                 }
+                21 => {
+                    if chain != Chain::Testnet && chain != Chain::Mainnet {
+                        // Enable fast commit syncer for faster recovery in devnet
+                        cfg.feature_flags.enable_fast_commit_sync = true;
```
polinikita: This flag should not be enabled without consensus_transaction_ref. Moreover, the latter can serve the same purpose as enable_fast_commit_sync

crates/starfish/core/src/authority_node.rs
```diff
@@ -46,7 +46,7 @@ pub struct ConsensusAuthority {
     shard_reconstructor: Arc<ShardReconstructorHandle>,
     cordial_knowledge: Arc<CordialKnowledgeHandle>,
     regular_commit_syncer_handle: CommitSyncerHandle,
-    fast_commit_syncer_handle: CommitSyncerHandle,
+    fast_commit_syncer_handle: Option<CommitSyncerHandle>,
```
polinikita, line 49: If transaction_ref is not enabled in a given epoch, the handle will finish after entering.

polinikita (changes requested):
Not sure whether a new flag is really needed here. Transaction_ref flag can serve the same purpose and it should be activated at least before enable_fast_commit_sync.

crates/iota-protocol-config/src/lib.rs
piotrm50: I'll merge those two flags.

crates/starfish/core/src/authority_node.rs
piotrm50: Where is that done exactly? I only see that `try_schedule_once` wouldn't do anything because `should_schedule` would be false for Fast syncer. However, the tasks would be running unnecessarily.

bingyanglin (approved):

polinikita (approved):

Thoralf-M (approved):
Approving the changes in iota-protocol-config

tomxey (approved):

PR #10413, opened 2026-02-20
feat(iota-core, iota-types): Introduce ConsensusCommitPrologueV2 with additional state digests

oliviasaa:
# Description of change

- Introduce `ConsensusCommitPrologueV2` transaction kind with a new `p.additional_states_digests: Vec<AdditionalConsensusStateDigest>` field for recording digests of additional states, enabling early fork detection.                            
 - Introduce `AdditionalConsensusState` as an extensible struct where each field produces its own digest via a `AdditionalConsensusStateTrait` impl over `Serialize`. New state fields can be added without creating a V3 prologue.
 - Add a protocol config feature flag `record_additional_state_digest_in_prologue`, controlling V1→V2 prologue switch, and enabled on devnet in protocol version 21.

OBS: I probably need to add more release notes below (?)

TO BE MERGED WITH https://github.com/iotaledger/iota-rust-sdk/pull/976

### Release Notes

- [x] Protocol: Added `ConsensusCommitPrologueV2` transaction kind with `additional_state_digest field` on devnet in protocol version 21. 
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-grpc-server/src/transaction_filter.rs
```diff
@@ -31,6 +31,7 @@ pub enum TransactionKind {
     AuthenticatorStateUpdateV1 = 4,
     EndOfEpochTransaction = 5,
     RandomnessStateUpdate = 6,
+    ConsensusCommitPrologueV2 = 7,
```
bingyanglin, line 34: We need to add `CONSENSUS_COMMIT_PROLOGUE_V2 ` for `TransactionKind` in `crates/iota-grpc-types/proto/iota/grpc/v0/filter.proto` I think.

crates/iota-graphql-rpc/src/types/transaction_block_kind/consensus_commit_prologue.rs
```diff
@@ -22,23 +80,23 @@ pub(crate) struct ConsensusCommitPrologueTransaction {
 impl ConsensusCommitPrologueTransaction {
```
bingyanglin, line 80: Need a getter for `additional_states_digests`?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1577,6 +1587,22 @@ impl ProtocolConfig {
...
+    pub fn record_additional_states_digests_in_prologue(&self) -> bool {
+        self.feature_flags
+            .record_additional_states_digests_in_prologue
+    }
+
+    pub fn record_received_reports_state_digest_in_prologue(&self) -> bool {
+        let record = self
+            .feature_flags
+            .record_received_reports_state_digest_in_prologue;
+        assert!(
```
bingyanglin: We panic in the run time if a future protocol version mistakenly enables `record_received_reports_state_digest_in_prologue` without `record_additional_states_digests_in_prologue`, which is risky?

crates/iota-core/src/consensus_handler.rs
```diff
@@ -101,6 +103,83 @@ impl ConsensusHandlerInitializer {
...
+    // ConsensusCommitPrologueV2, or even only to track them across commits. To add
+    // a new field:
+    // - add it to the struct,
+    // - modify AdditionalConsensusStates::new() to initialize it,
+    // - implement a setter for the field.
+    // - add the field to AdditionalConsensusStates::iter() for debugging and
+    //   testing purposes.
+    //
+    // To use the field in ConsensusCommitPrologueV2:
+    // - add the field's to a new verstion of AdditionalConsensusStates::iter_v_(),
```
bingyanglin:
```suggestion
    // - add the field's to a new version of AdditionalConsensusStates::iter_v_(),
```

crates/iota-protocol-config/src/lib.rs
oliviasaa: I'm using the same checks that had been added in the past for other flags. This would panic actually in any test using the maximum supported protocol version

crates/iota-core/src/consensus_handler.rs
oliviasaa: sorry, I marked this PR as open to reviews before pushing the very last commits. This was fixed in [c4e0153](https://github.com/iotaledger/iota/pull/10413/commits/c4e0153f1733ac09aa9a7e9632a9ac4b9b10edf8)

crates/iota-grpc-server/src/transaction_filter.rs
oliviasaa: This file is generated, I didn't run any of those scripts yet. 
Added as a TO DO in the PR description

crates/iota-graphql-rpc/src/types/transaction_block_kind/consensus_commit_prologue.rs
oliviasaa: Honestly? I have no idea. Let's wait @iotaledger/infrastructure to decide
`consensus_determined_version_assignments` also doesn't have a getter, and I don't know exactly for what those getters are used.

crates/iota-protocol-config/src/lib.rs
bingyanglin: Yeah there are many places use the same asserts for the consensus related flags, but it might be a bug. The assert checks run in production, not just in tests. Normally the flag combinations are hardcoded per protocol version, so tests do catch misconfigurations. But `ProtocolConfig::get_for_version` supports overriding config values **at runtime** via `IOTA_PROTOCOL_CONFIG_OVERRIDE_*` environment variables. The override macro (`ProtocolConfigOverride`) generates an `apply_to` method that can replace any field, including feature_flags. The overrides are applied before the getters (and their asserts) run.

So if a node operator sets `IOTA_PROTOCOL_CONFIG_OVERRIDE_ENABLE` and overrides `feature_flags` with `record_received_reports_state_digest_in_prologue = true` but leaves `record_additional_states_digests_in_prologue = false`, the node will panic **at runtime** when the getter is called. Tests wouldn't catch this because the override is an operator action, not a code change.

Maybe we need to add an issue to trace it. Or do I miss something? Or it's just an edge case we don't care?

crates/iota-grpc-server/src/transaction_filter.rs
bingyanglin: No I don't think this file is generated?

oliviasaa: sorry, I was looking at the wrong file. Let me see this in detail

alexsporn (changes requested):
Please split this PR into 2, one introducing the `ConsensusCommitPrologueV2` and one with the `ReceivedReportsState` changes

crates/iota-protocol-config/src/lib.rs
oliviasaa: afaik, we don't care about this because the overrides are there just for testing/emergency purposes, so a node operator should not set `IOTA_PROTOCOL_CONFIG_OVERRIDE_ENABLE` without a really good reason

crates/iota-core/src/consensus_handler.rs
oliviasaa: This branch was rebased when the PR was split into 2; this fix was incorporated to [648496f](https://github.com/iotaledger/iota/pull/10413/commits/648496f45273e57d597e704f7cbce7da7ebdc461)

crates/iota-grpc-server/src/transaction_filter.rs
oliviasaa: done in [112a654](https://github.com/iotaledger/iota/pull/10413/commits/112a654dea6e6bcdecc6aefe6c97ee42fcf056fc)

oliviasaa:
> Please split this PR into 2, one introducing the `ConsensusCommitPrologueV2` and one with the `ReceivedReportsState` changes

done

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2528,6 +2540,11 @@ impl ProtocolConfig {
                         // randomness on testnet.
                         cfg.feature_flags
                             .separate_gas_price_feedback_mechanism_for_randomness = true;
+                        if chain != Chain::Testnet {
```
Thoralf-M: Technically correct as it's also in the != Chain::Mainnet branch, but in the protocol version before this was separated to make it clearer
`if chain != Chain::Testnet && chain != Chain::Mainnet {`, should probably also do that here

crates/iota-types/src/digests.rs
```diff
@@ -1063,6 +1063,57 @@ impl fmt::Debug for MisbehaviorReportDigest {
     }
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, JsonSchema)]
+pub struct AdditionalConsensusStatesDigest(Digest);
```
Thoralf-M, line 1071: Can you add some comment to the type what it is for?

oliviasaa: sure! added in [d3815e7](https://github.com/iotaledger/iota/pull/10413/commits/d3815e756367f4196e6efb9a6c64fc8c59b9a442)

crates/iota-protocol-config/src/lib.rs
oliviasaa: fixed in [d3815e7](https://github.com/iotaledger/iota/pull/10413/commits/d3815e756367f4196e6efb9a6c64fc8c59b9a442)

crates/iota-types/src/transaction.rs
```diff
@@ -3022,6 +3059,31 @@ impl VerifiedTransaction {
...
+        commit_timestamp_ms: CheckpointTimestamp,
+        consensus_commit_digest: ConsensusCommitDigest,
+        cancelled_txn_version_assignment: Vec<(TransactionDigest, Vec<(ObjectID, SequenceNumber)>)>,
+        additional_states_digests: Vec<AdditionalConsensusStatesDigest>,
+    ) -> Self {
+        ConsensusCommitPrologueV2 {
+            epoch,
+            round,
+            // sub_dag_index is reserved for when we have multi commits per round.
+            sub_dag_index: None,
```
Thoralf-M, line 3074: Is this planned to ever happen with ConsensusCommitPrologueV2 or will there be a V3 introduced then? Just looks a bit wasted that this was already in V1 but never used?

oliviasaa: Maybe someone from @iotaledger/consensus can confirm this?

crates/iota-types/src/messages_consensus.rs
```diff
@@ -61,6 +63,26 @@ pub struct ConsensusCommitPrologueV1 {
...
+    pub sub_dag_index: Option<u64>,
+    /// Unix timestamp from consensus commit.
+    pub commit_timestamp_ms: TimestampMs,
+    /// Digest of consensus output
+    pub consensus_commit_digest: ConsensusCommitDigest,
+    /// Stores consensus handler determined shared object version assignments.
+    pub consensus_determined_version_assignments: ConsensusDeterminedVersionAssignments,
+    /// Digest of any additional state computed by the consensus handler.
+    /// Used to detect forking bugs as early as possible.
+    pub additional_states_digests: Vec<AdditionalConsensusStatesDigest>,
```
muXxer, line 83: Instead of the vector, we could just use a merkle root as discussed.

piotrm50:
## Related orphan branches (for reference during rebase)

Two branches exist on remote without their own PRs. They contain work relevant to this PR:

### `protocol-research/feat/score-integration-merged-prologue-v2-storage`
Integration branch that merges this PR (#10413, ConsensusCommitPrologueV2) with #10366 (persist scorer state). Contains:
- ConsensusCommitPrologueV2 creation and handling gated by feature flag
- Tests: `test_user_sends_consensus_commit_prologue_v2`, `test_consensus_commit_prologue_v2_digests`, plus updates to pre-existing tests
- GRPC types generation and filter updates
- Snapshot updates

Useful as a reference for how the prologue-v2 and persistence work were intended to combine.

### `protocol-research/fix/ci-for-score-integration-prologue-v2`
CI fix branch for prologue-v2. Contains:
- `CancelledTransactionsV2` implementation and revert
- Fixes for `iota-benchmark` simtests, `iota-replay` regression test, clippy, ts-sdk tests, consensus commit prologue tests

Useful when rebasing this PR to identify CI failures and their known fixes.

PR #10414, opened 2026-02-20
feat(gRPC): do not return all entries of maps if parent wildcard was passed

muXxer:
# Description of change

Right now if a wildcard field mask is passed to `get_epoch`, all possible values for the protocol attributes and protocol flags are returned. This would create huge responses, even if those fields are mostly not relevant for clients. With the changed logic, they can either specify keys inside the map directly, or still get all values when mentioning the inner field inside the wrapper as a wildcard for the map.

```rust
// Get all feature flags for the current epoch
let epoch = client
    .get_epoch(None, Some("protocol_config.feature_flags.flags"))
    .await?;
let flags = epoch.protocol_config.unwrap().feature_flags.unwrap().flags;

// Get a single named feature flag
let epoch = client
    .get_epoch(
        None,
        Some("protocol_config.feature_flags.flags.zklogin_auth"),
    )
    .await?;

// Get a single named attribute
let epoch = client
    .get_epoch(
        None,
        Some("protocol_config.attributes.attributes.max_tx_gas"),
    )
    .await?;
```

### DIFF
--- a/crates/iota-grpc-client/src/api/ledger/epochs.rs
+++ b/crates/iota-grpc-client/src/api/ledger/epochs.rs
@@ -34,7 +34,24 @@ impl Client {
     /// - `start` - Epoch start timestamp
     /// - `end` - Epoch end timestamp
     /// - `reference_gas_price` - Reference gas price in NANOS
-    /// - `protocol_config` - Protocol configuration for this epoch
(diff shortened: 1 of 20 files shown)


bingyanglin (approved):
Some comments addressed in #10416

PR #10415, opened 2026-02-20
feat(gRPC): add `suggested_gas_price` to `simulate_transaction`

muXxer:
# Description of change

This PR adds the `suggested_gas_price` to `simulate_transaction`.

Merge after #10414

## Links to any relevant issues

#10391

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

Issue #10390, opened 2026-02-19
gRPC: add `execution_error_source` to the `simulate_transaction` response


muXxer:
It would be nice to expose the reason for a transaction error as it is done in the json-rpc.

muXxer:
Added in #10416

PR #10416, opened 2026-02-20
feat(gRPC): return `ExecutionResult` instead of `CommandResults` in `simulate_transaction`

muXxer:
# Description of change

With this PR we return the actual `ExecutionResult` of a `simulate_transaction` call, which also contains information about failed executions, like error_kind, error_source and command_index.

Merge after #10414 and #10415

#10390


crates/iota-grpc-server/src/transaction_execution_service/simulate.rs
bingyanglin:
```suggestion
    "result"
```

bingyanglin: Also the e2e testing for `simulate_transaction_readmask_scenarios` might need to be also updated.

crates/iota-grpc-server/src/transaction_execution_service/simulate.rs
```diff
@@ -225,33 +228,32 @@ pub async fn simulate_transaction(
...
-                response.command_results = Some(CommandResults::default());
-            }
-        }
+    // Only include suggested gas price if requested
+    if read_mask.contains(SimulateTransactionResponse::SUGGESTED_GAS_PRICE_FIELD.name) {
+        response.suggested_gas_price = Some(suggested_gas_price.ok_or_else(|| {
+            RpcError::new(
+                tonic::Code::Internal,
+                "suggested gas price is not available".to_string(),
+            )
```
bingyanglin: For transactions not touching shared objects the price can be `None` I guess? If in this way we can just do the following to make it infallible?
```suggestion
        response.suggested_gas_price = suggested_gas_price;
```

crates/iota-grpc-server/src/ledger_service/get_epoch.rs
```diff
@@ -43,95 +152,22 @@ pub fn get_epoch(
...
-        let committee = service
-            .reader
-            .get_committee(epoch)
-            .map_err(|e| Status::internal(format!("Failed to get committee: {e}")))?
-            .ok_or_else(|| CommitteeNotFoundError::new(epoch))?;
-        let sdk_committee: iota_sdk_types::ValidatorCommittee = committee.as_ref().clone().into();
-        message = message.with_committee(sdk_committee);
-    }
+    let message = Epoch::merge_from(&source, &read_mask)
+        .map_err(|e| Status::internal(format!("Failed to build epoch response: {e}")))?;
```
bingyanglin, line 170: I guess we need to downcast the `Box<dyn Error>` at the call site? Otherwise we cannot distinguish `Status::not_found` and `Status::internal` errors.

Alex6323 (approved):

crates/iota-grpc-server/src/transaction_execution_service/simulate.rs
muXxer: I removed the "result" layer and changed the tests in https://github.com/iotaledger/iota/pull/10416/commits/3063384967fc807c2a913f14b5d0923a31424ced

crates/iota-grpc-server/src/transaction_execution_service/simulate.rs
muXxer: No we can't return `None` because then the client can't distinguish between "read mask not passed" or "no objects congested". I now return the reference gas price instead. https://github.com/iotaledger/iota/pull/10416/commits/53527b3af9deba226c3e0f006168b9096de81d2e

crates/iota-grpc-server/src/ledger_service/get_epoch.rs
muXxer: Good catch, I would say we create another PR and extend the `Merge` trait with an associated Error.

bingyanglin (approved):

PR #10421, opened 2026-02-23
feat(ci): add `--filter-overwrite` argument to `rust_tests.py` to pass own filters to the tests

muXxer:
# Description of change

To run a specific simtest locally, the current approach is to run this command:
```bash
TEST_FILTER="package(iota-e2e-tests) and test(simulate_transaction_command_results)" scripts/simtest/simtest-run.sh
``` 

The `simtest-run.sh` script will then also run stresstests, no matter what filter is passed. In most cases, this is unwanted behaviour. 

The cleaner approach is to simply add new arguments to the `rust_tests.py` script that is used in the CI, so that we can overwrite the filters that are passed to the tests.

```bash
./rust_tests.sh --run-sim-tests --filter-overwrite "package(iota-e2e-tests) and test(simulate_transaction_command_results)"
```

### DIFF
--- a/scripts/ci_tests/rust_tests.py
+++ b/scripts/ci_tests/rust_tests.py
@@ -98,6 +98,8 @@ class RustTestOrchestrator:
                 'tests_crates_external',
                 'tests_pg_integration',
                 'tests_move_examples_rdeps',
+                'filter_overwrite',
+                'filter_overwrite_external',
             }
             
             self.logger.info("Additional command line arguments:")
@@ -589,6 +591,8 @@ class RustTestOrchestrator:
                              tests_crates_external=False,
                              tests_pg_integration=False,
                              tests_move_example_used_by_others=False,
+                             filter_overwrite: str = None,
+                             filter_overwrite_external: str = None,
                              ) -> int:
         
         if test_type not in [self.TEST_TYPE_NEXTEST, self.TEST_TYPE_SIMTEST]:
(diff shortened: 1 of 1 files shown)


jkrvivian (approved):

scripts/ci_tests/rust_tests.py
```diff
@@ -638,14 +654,26 @@ def filter_and_run_tests(self,
...
-        combined_set, tests_added = self.build_filterset_tests(
-            tests_crates_workspace, tests_pg_integration, tests_move_example_used_by_others,
-            test_only_changed_crates, changed_crates, changed_crates_given
-        )
+        combined_set, tests_added = "", False
+
+        if filter_overwrite:
+            self.logger.info(f"Using filter overwrite for main tests: \"{filter_overwrite}\"")
+            combined_set = filter_overwrite
+            tests_added = True
```
bingyanglin, line 671: When `filter_overwrite_external` is used, the `EXCLUDE_SET_EXTERNAL` is still applied at `L640-641`?

scripts/ci_tests/rust_tests.py
bingyanglin, line 683: When `--filter-overwrite` is provided alongside `--tests-crates-workspace`, `--tests-pg-integration`, or `--tests-move-examples-rdeps`, the `filter_overwrite` overwrites them, but here the `tests_pg_integration` still controls whether postgres is restarted. Maybe we log a warning for them so we will not forget this logic in the future?

scripts/ci_tests/rust_tests.py
muXxer: Yes this was intentional. These tests were always ignored, and I think even if the user passes a filter where those could somehow be included, the script should exclude them.

scripts/ci_tests/rust_tests.py
muXxer: Normally a user wouldn't pass these both flags if he uses the overwrite anyways, but at least he still has the possibility to restart postgres if he wants to for his specific test, so I would say we can leave it as it is.

bingyanglin (approved):

PR #10398, opened 2026-02-20
feat(wallet): update accounts added amplitude event

evavirseda:
# Description of change

Please write a summary of your changes and why you made them.

## Links to any relevant issues

fixes https://github.com/iotaledger/iota/issues/10167

## How the change has been tested

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

PR #10422, opened 2026-02-23
feat(wallet): update account creation started

### LINKED ISSUE #10164: Update account creation started amplitude event
Depends on https://github.com/iotaledger/iota/issues/10163

This event should fire as the starting point to an account creation
It should trigger during onboarding, when the user decides which account type to create, so in the new or import screens, when the type is chosen:
- Onboarding on the Import Wallet screen when the user taps either “Mnemonic”, “Private Key”, “Passkey”, ...
- Onboarding on the Create a new wallet screen when the user taps “Mnemonic” , “Passkey", ...
Fired during Onboarding on the Import Wallet screen when the user taps either “Mnemonic”, “Private Key”, “Passkey” or “Firefly Seed Recovery”.
- Onboarding when the user taps on either “Ledger” or “Keystone” buttons

Props:
- `accountType`: `private key | mnemonic | passkey | ledger | keystone | seed`
- `accountOrigin`: `new | import` 
- `isFirstAccount`: `boolean` 
- `sourceFlow`: `string`

Cleanup:
- Remove `clickedCreatePasskey`, it should be covered by account creation started with all types
- Remove `clickedImportKeystone`, it should be covered by account creation started with all types
- Remove `clickedImportPassphrase`, it should be covered by account creation started with all types
- Remove `clickedImportPrivateKey`, it should be covered by account creation started with all types
- Remove `clickedImportSeed`, it should be covered by account creation started with all types
- Remove `connectedHardwareWallet` , it should be covered by account creation started with all types
- Remove `connectedHardwareWallet`, unused event in the flow

panteleymonchuk:
Closes #10164
Depends on: https://github.com/iotaledger/iota/pull/10398


apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
```diff
@@ -82,6 +88,8 @@ export function AddAccountPage() {
     const [cameraPermissionStatus] = useCheckCameraPermissionStatus();
     const network = useAppSelector(({ app }) => app.network);
     const isPasskeysEnabled = useFeatureEnabledByNetwork(Feature.WalletPasskeys, network);
+    const { data: accounts } = useAccounts();
+    const isFirstAccount = !accounts || accounts.length === 0;
```
brancoder: we could extract this to a hook maybe `useIsFirstAccount` because its used in these three files? wdyt?

apps/wallet/src/ui/app/pages/accounts/ImportExistingWallet.tsx
```diff
@@ -67,15 +70,30 @@ export function ImportExistingWallet() {
     ) => {
         switch (actionType) {
             case AccountsFormType.ImportMnemonic:
-                ampli.clickedImportPassphrase({ sourceFlow });
+                ampli.accountCreationStarted({
+                    accountType: AmpliAccountType.Mnemonic,
+                    accountOrigin: AmpliAccountOrigin.Import,
+                    isFirstAccount,
+                    sourceFlow,
+                });
```
brancoder: we could extract firing ampli event before we reach switch case so the code is not duplicated since only property that changes is the account type. ex.
```
  const ACCOUNT_TYPE_MAP = {
          [AccountsFormType.ImportMnemonic]: AmpliAccountType.Mnemonic,
          [AccountsFormType.ImportPrivateKey]: AmpliAccountType.PrivateKey,
          [AccountsFormType.ImportPasskey]: AmpliAccountType.Passkey,
          [AccountsFormType.ImportSeed]: AmpliAccountType.Seed,
      } as const;
....

       const handleCardAction = async (
        actionType: (typeof profileOptions | typeof legacyOptions)[number]['actionType'],
      ) => {
          const accountType = ACCOUNT_TYPE_MAP[actionType as keyof typeof ACCOUNT_TYPE_MAP];
          if (accountType === undefined) {
              throw new Error('Unsupported action type');
          }
  
          ampli.accountCreationStarted({
              accountType,
              accountOrigin: AmpliAccountOrigin.Import,
              isFirstAccount,
              sourceFlow,
          });

and remove the default from  the switch case because we threw the error allready if theres no accountType
```

brancoder (changes requested):
LG just some minor suggestions

apps/wallet/src/ui/app/pages/accounts/ImportExistingWallet.tsx
brancoder: ah we all-ready have a map in constants `const ACCOUNT_TYPE_TO_AMPLI_ACCOUNT_TYPE` so we can just use that to determine `accountType`

apps/wallet/src/ui/app/pages/accounts/ImportExistingWallet.tsx
marc2332, line 1: Why do you need to call ampli in 4 places, just do it once no?

apps/wallet/src/ui/app/pages/accounts/CreateNewWallet.tsx
marc2332, line 1: Same

apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
marc2332, line 1: Same

marc2332 (changes requested):

apps/wallet/src/ui/app/pages/accounts/ImportExistingWallet.tsx
brancoder, line 29: we can remove this `const createAccountsMutation = useCreateAccountsMutation();` because the mutation is used just for `isDisabled` flag `createAccountsMutation.isPending`. but the actual mutation that does the creation of the account is in the `ProtectedAccountPage` and `useMutations` dont share context across hook instances. righ? this happens in AddAccountPage, CreateNewWallet and WelcomPage also. wdyt?
this was not introduced in this PR but maybe we can fix it in another PR or we can also add it to this one. how ever you like

brancoder (changes requested):

apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
```diff
@@ -114,9 +122,19 @@ export function AddAccountPage() {
     async function handleCardAction(
         actionType: (typeof hardwareWalletOptions)[number]['actionType'],
     ) {
+        const ampliData = ACCOUNT_FORM_TYPE_TO_AMPLI[actionType];
+
+        if (ampliData) {
+            ampli.accountCreationStarted({
+                accountType: ampliData.accountType,
+                accountOrigin: ampliData.accountOrigin,
```
brancoder, line 130: can we spread the `...ampliData`

apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
```diff
@@ -125,16 +143,13 @@ export function AddAccountPage() {
...
-                ampli.clickedImportKeystone({ sourceFlow });
                 if (isPopupOrSidePanel && cameraPermissionStatus === 'prompt') {
                     await openTabOnImportKeystone();
                     window.close();
                 } else {
                     navigate('/accounts/import-keystone');
                 }
                 break;
-            default:
-                throw new Error('Unsupported action type');
```
brancoder, line 137: seams we have lost the default case in all the files

brancoder (changes requested):

brancoder (approved):
LGTM

marc2332 (approved):

PR #10424, opened 2026-02-23
feat(wallet): add transaction opened event

### LINKED ISSUE #10197: Add transaction opened amplitude event
Add transaction opened amplitude event
Fires when user taps a transaction in the Activity page

Props
- `transactionType`: high level tx group. See `TransactionAction` in `apps/core/src/interfaces/transactions.interfaces.ts`. We should use this type, but the "Failed" type should not be part of it, Failed is for the next prop
- `success`: `boolean`

panteleymonchuk:
Closes #10197


apps/wallet/src/ui/app/components/transactions-card/index.tsx
```diff
@@ -84,13 +85,23 @@ export function TransactionCard({ txn, address }: TransactionCardProps) {
         ? '--'
         : formatDate(Number(txn.timestampMs), ['day', 'month', 'year', 'hour', 'minute']);
 
+    const handleTransactionClick = () => {
+        const transactionAction = getTransactionAction(txn, address);
```
brancoder: we can extract this `getTransactionAction` and the `success` condition to constants and use them in handler and also in:
```
<TransactionIcon
                        txnFailed={executionStatus !== 'success' || !!error}
                        variant={getTransactionAction(txn, address)}
                    />
```

brancoder (changes requested):
lg just a refactor 🙏

brancoder (approved):

begonaalvarezd (approved):

PR #10428, opened 2026-02-23
chore(docs): clean up aa how-tos

Dr-Electron:
Created this as a suggestion to follow [diataxis](https://diataxis.fr) and our old examples more closely.

A hot-to is closer to a reference. It is something you look up. Not something you use to learn directly. That would be a tutorial. 
A how to is this one step by step guide of your favourite recipe that you still have to check every time tho ;)


docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
```diff
@@ -1,72 +1,42 @@
 ---
-title: Basic Account Abstraction Example
-description: 'How to attach an authenticator to an account in Move'
+title: Create a Basic Account Authenticator
```
Dr-Electron: We only need one title. Either the one in the front matter or the header one in the document

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
```diff
@@ -1,72 +1,42 @@
...
-
-This how-to demonstrates the **minimum building blocks** required to use Account Abstraction:
-
-- Defining a shared account object
-- Attaching an authenticator function at runtime
-- Executing a transaction that is authorized via the authenticator
-
-This example intentionally avoids multisig, thresholds, or off-chain coordination to focus on core mechanics.
+This How-To demonstrates how to create and attach a basic authenticator to an account.
+If you want to learn more about the concepts of Account Abstraction, check out the [Account Abstraction Introduction](../../explanations/account-abstraction.mdx).
```
Dr-Electron, line 16: Made this a little shorter and less explanatory. That would be the job of the AA explanation page I now created a placeholder doc for :)

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
```diff
@@ -1,72 +1,42 @@
...
 ## Example Code
 
-
-Below is the **minimal core code** required for Account Abstraction.  
-The full package includes additional setup and comments, but this snippet highlights only the essential logic.
-
-### Account creation and authenticator attachment
-
-```move file=<rootDir>/examples/move/account/sources/account.move#L4-L26
+1. Define an account object.
```
Dr-Electron, line 20: Made this section more step-by-step quickly talking about each step and linking to references

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
```diff
@@ -1,72 +1,42 @@
...
-### Code Outcome
-
-After attaching the authenticator:
+## Verify the Result
 
-- The account becomes an abstract account
-- Any transaction sent from this account must pass the authenticate function
-- Transactions succeed only if the provided argument equals `"hello"`
+- A transaction authorized through this authenticator succeeds when the message argument is `"hello"`.
+- The same transaction fails when the message argument differs.
```
Dr-Electron: Basically the same as your expected behaviour. Copilot insisted on calling it `Verify the Result` so why not try something new? 😆 
I have no strong option as long as we keep it consistent.

Wdyt? Any preferences?

Dr-Electron: Also moved it before full example code. In the past we did it different but I think this way makes actually more sense

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
Dr-Electron, line 1: Let me know what you think about my changes. I wanted everything to be a little closer to our old stuff like https://legacy.wiki.iota.org/iota-sdk/how-tos/accounts-and-addresses/list-outputs/ but also I'm not against trying new things.

If you like this PR we should adopt the other examples to follow the same structure (roughly)

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
vivekjain23: Ok sure, I'll keep the `title:` one, just created that for formatting stuff, but no issues will fix this

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
vivekjain23: I think we should try a new thing here and keep the explanation part in the how-to's page itself, which will make it easy to go through the code example and easy for the newcomers, and although it doesn't make the docs too large, we can adopt it

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
vivekjain23: Yeah, this way makes more sense, so one thing we can try out here is mixing the sections and combining the result and outcome explanation in one section above the `Full Example Code`

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
Dr-Electron: In the end it doesn't really matter which one we keep. As long as we don't duplicate the title. This also sometimes had really weird side effects in the old Wiki 😆 
So keep whatever you prefer 👍

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
Dr-Electron: New comers should more look at Tutorials tho.
I have a small strong opinion against this, but if we keep it short I let you decide 😆 ❤️

docs/content/developer/move/how-tos/account-abstraction/basic-account-authenticator.mdx
vivekjain23: Sure, I'll keep only one of them

PR #10460, opened 2026-02-25
feat(indexer): Guard objects deletion with version check, persist objects only after tx global order on Checkpoint path

### LINKED ISSUE #10250: [indexer]: Handle race conditions with new object-based dependency checks
Previously dependency check was checking if global order of previous tx is already present in the global order table AND that it's status is NOT Started (so we checked if indexing for this tx is already fully finished)

Currently we are checking if transaction input objects are present in the DB (in other words, we check if objects created by parent transactions are already indexed), but that doesn't mean that parent transactions are indexed fully.
They still may have e.g. global order not assigned or they may still want to remove some objects or they may want to update the display table, etc.

this creates a race condition that may lead to failures like:

```
---- write_api::test_parallel_shared_object_updates stdout ----

thread 'write_api::test_parallel_shared_object_updates' (78126163) panicked at crates/iota-indexer/tests/rpc-tests/write_api.rs:656:25:
tx: TransactionDigest(CtCFeKVDVEgMqUseunL3qAA538Z8ZR2MVVGxqQfXPHka) should have bigger order than it's deps: {TransactionDigest(C1f4eFtw2R2octZa7y4o35Vb2UnL6R1gGQwCdfMHzEDW)}
```

Part of the issue could be solved by making sure global order is assigned in checkpoint indexing before changes to objects table are written.

To be investigated if this is enough to solve all the issues.

tomxey:
# Description of change

Guard objects deletion with version check, persist objects only after tx global order on Checkpoint path

fixes https://github.com/iotaledger/iota/issues/10250

reason for object version check: https://github.com/iotaledger/iota/issues/10250#issuecomment-3907617618
reason for tx_global_order->objects insertion ordering: https://github.com/iotaledger/iota/issues/10250#issuecomment-3921366687

### DIFF
--- a/crates/iota-indexer/src/ingestion/primary/persist.rs
+++ b/crates/iota-indexer/src/ingestion/primary/persist.rs
@@ -164,17 +164,25 @@ impl PrimaryWriter {
             let mut persist_tasks = vec![
                 self.state.persist_transactions(tx_batch),
                 self.state.persist_tx_indices(tx_indices_batch),
-                self.state
-                    .persist_tx_global_order(tx_global_order_batch.clone()),
                 self.state.persist_events(events_batch),
                 self.state.persist_event_indices(event_indices_batch),
                 self.state.persist_displays(display_updates_batch),
                 self.state.persist_packages(packages_batch),
-                self.state.persist_checkpoint_objects(object_changes_batch),
                 self.state
                     .persist_object_history(object_history_changes_batch.clone()),
                 self.state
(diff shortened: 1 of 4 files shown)


kodemartin (approved):
lgtm with the suggestion to add a version check while inserting new objects as well. It seems to me to be an extra guarantee with no additional cost, no?

sergiupopescu199 (approved):

PR #10463, opened 2026-02-25
upstream(core): Use pipelined execution in CheckpointExecutor

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/50346c87efe5c63b263af9469e8e734d61926ab9
  - https://github.com/MystenLabs/sui/commit/9cdfd1b041b6d4d51f9cce462bda68261975e746
- Description:

This allows us to have parallelism throughout the entire
CheckpointExecution process, rather than just during the
transaction-execution phase.

Additionally, by imposing stricter-than-necessary ordering constraints
on the execution phase (i.e. that we enqueue all transactions from
checkpoint N before enqueueing those from N+1) we achieve lower latency.

Throughput should also be improved here, by running the batch-building
and batch-writing steps in different stages of the pipeline, which
allows us to start building the batch for seq N+1 while we are still
committing the batch for seq N.

[run-ci]

Part of #9768


piotrm50 (approved):

crates/iota-core/src/checkpoints/checkpoint_executor/utils.rs
```diff
@@ -253,6 +256,221 @@ pub(super) fn assert_checkpoint_not_forked(
...
+//       }
+//
+// Then, simply run many instances of this function concurrently, and the
+// pipeline ordering will be maintained.
+//
+// Currently this code is not generic with respect to the names of the stages.
+// We can make the stage enum a generic param when we need to use this code in
+// other places.
+
+/// Names of the pipeline stages for CheckpointExecutor.
```
bingyanglin, line 333: This should be totally removed? in Line 348 we have exactly the same and this will make the docs of `PipelineStage` to have two identical sentences.

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
bingyanglin, line 666: I think we also need to port this bug fix
https://github.com/MystenLabs/sui/commit/5d456eb496
which is related to checkpoint executor

bingyanglin, line 970: Need to have `#[instrument(level = "info", skip_all)]`? This is corresponding to the `commit_index_updates_and_enqueue_to_subscription_service` in https://github.com/MystenLabs/sui/commit/50346c87efe5c63b263af9469e8e734d61926ab9

jkrvivian: I don't understand the bug fix in https://github.com/MystenLabs/sui/commit/5d456eb496 👀 
Also we marked it as wont-merge.

jkrvivian: I merged it to the upstream commit https://github.com/iotaledger/iota/pull/10463/commits/86ed7243f10a6b6c43fa11510d4c598e2946528e and force pushed.

bingyanglin: After re-investigating this in more details, I think it's fine to just make it panic if the db is corrupted and then we manually investigate it afterwards.

bingyanglin (approved):

Thoralf-M (approved):

PR #10464, opened 2026-02-25
refactor: use native functions in AuthContext

### LINKED ISSUE #9505: [AA] Improve AuthContext
If we just add a new field to a Move struct or change one that exists, we make the struct formally incompatible with its previous version according to the compatibility rules([this](https://github.com/iotaledger/iota/blob/f618ca4c386e2db5a87f7fc71289640d2eaddd39/crates/iota-framework-snapshot/tests/compatibility_tests.rs#L22) test fails).

The proposal is to follow the way implemented in Sui for TxContext (see #6229), where we have an empty Move struct and all its methods are native functions:
```move
public struct AuthContext has drop {}

public fun tx_inputs(ctx: &AuthContext): vector<CallArg> {
    native_tx_inputs()
}

native fun native_tx_inputs(): vector<CallArg>; 

public fun tx_inputs_v2(ctx: &AuthContext): vector<CallArgV2> {
    native_tx_inputs_v2()
}

native fun native_tx_inputs_v2(): vector<CallArgV2>; //<-- can be added without the `AuthContext` struct updating.

public native fun digest(): vector<u8>; //<-- can be just public without a Move wrapper.
```

This configuration would allow us to construct the AuthContext dynamically during execution: Instead of reading the PTB information and creating a AuthContext value to pass as argument.

The native functions fetch/construct the data requested by the user allows to “pay-per-call“, i.e., have a free creation of an empty AuthContext, then pay for each native function you use. Each native function should charge gas based on the number of inputs and commands.

valeriyr:
# Description of change

Move `AuthContext` is refactored to be managed by the native functions.
This allows us to extend `AuthContext` with new data without changing the struct layout and make it consistent with the following TxContext refactoring.

fixes #9505


iota-execution/latest/iota-move-natives/src/authentication_context.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+fn to_global_value<T: ?Sized + Serialize>(
+    field: &T,
+    field_move_layout: MoveTypeLayout,
+) -> PartialVMResult<GlobalValue> {
+    // Wrap in a tuple to match the expected Move layout of
+    //`struct AuthContext {
+    //     field: FieldType,
+    // }`
```
miker83z: This comment works for lines 75 and 121, because there we create the tuple `let rust_value = (auth_context.tx_inputs(),);`

valeriyr: fixed in 2df44e277d

miker83z (approved):

kodemartin (approved):
lgtm

vivekjain23 (approved):
LGTM!

Dr-Electron (approved):

alexsporn (approved):

Issue #10270, opened 2026-02-13
[AA] Enable EnrichedProgrammableMoveCall and EnrichedCallArg in AuthContext


miker83z:
Example of a RicherMoveCall command:
```rust
/// The command for calling a Move function, either an entry function or a
/// public function (which cannot return references).
#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
pub struct RichProgrammableMoveCall {
    /// The package containing the module and function.
    pub package: ObjectID,
    /// The specific module in the package containing the function.
    pub module: String,
    /// The function to be called.
    pub function: String,
    /// Marker if the function is intended as an entry function.
    pub is_entry: bool,
    /// The type arguments to the function.
    pub type_arguments: Vec<TypeName>,
    /// The arguments to the function.
    /// This points to Argument::Input or Argument::Result from which
    /// we can extract the TypeName.
    pub arguments: Vec<Argument>,
    /// The returned value of the function. 
    /// Argument::Result and Argument::NestedResult point to this.
    pub returns: Vec<TypeName>,
}
```

miker83z:
Additionally, flatten CallArgs. See https://github.com/iotaledger/iota/pull/10466#discussion_r2853483450

miker83z:
```rust
#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, JsonSchema)]
pub enum RichCallArg {
    // contains no structs or objects
    Pure {
        value: Vec<u8>,
        type_name: TypeName,  
    },
    // A Move object, either immutable, or owned mutable.
    ImmOrOwnedObject(ImmOrOwnedObjectArg),
    // A Move object that's shared.
    // SharedObject::mutable controls whether caller asks for a mutable reference to shared
    // object.
    SharedObject(SharedObjectArg),
    // A Move object that can be received in this transaction.
    Receiving(ImmOrOwnedObjectArg),
}

#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy, Serialize, Deserialize, JsonSchema)]
pub struct ImmOrOwnedObjectArg {
        id: ObjectID,
        version: SequenceNumber,
        digest: ObjectDigest,
        mutable: bool,
        digest: ObjectDigest,
        type_name: TypeName,
}

#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy, Serialize, Deserialize, JsonSchema)]
pub struct SharedObjectArg {
        id: ObjectID,
[...]

PR #10466, opened 2026-02-25
feat(types): create AuthContextCommand and AuthContextCallArg

miker83z:
# Description of change

This PR makes it so that `AuthContext` fields use dedicated types instead of reusing the `PTB::Command` and `PTB::CallArg` types.  `AuthContextCommand` and `AuthContextCallArg` are created in such a way that `AuthContext` de/serialization from/to Move is straightforward.

### DIFF
--- a/crates/iota-types/src/auth_context.rs
+++ b/crates/iota-types/src/auth_context.rs
@@ -4,20 +4,142 @@
 use move_binary_format::{CompiledModule, file_format::SignatureToken};
 use move_bytecode_utils::resolve_struct;
 use move_core_types::{account_address::AccountAddress, ident_str, identifier::IdentStr};
-use serde::{
-    Serialize, Serializer,
-    ser::{SerializeStruct, SerializeStructVariant, SerializeTupleVariant},
-};
+use serde::{Deserialize, Serialize};
 
 use crate::{
     IOTA_FRAMEWORK_ADDRESS,
+    base_types::ObjectID,
     digests::MoveAuthenticatorDigest,
-    transaction::{CallArg, Command, ProgrammableTransaction},
+    transaction::{Argument, CallArg, Command, ObjectArg, ProgrammableTransaction},
     type_input::TypeName,
 };
 
 pub const AUTH_CONTEXT_MODULE_NAME: &IdentStr = ident_str!("auth_context");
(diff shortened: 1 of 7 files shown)


crates/iota-types/src/auth_context.rs
```diff
@@ -4,21 +4,143 @@
...
+
+/// V1 call-argument layout for [`AuthContextCallArg`].
+///
+/// Mirrors [`crate::transaction::CallArg`], matching the BCS layout expected
+/// by the Move-side `ptb_call_arg::CallArg`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum AuthContextCallArgV1 {
+    Pure(Vec<u8>),
+    Object(ObjectArg),
+}
```
Thoralf-M: Can we flatten this to have all the ObjectArg variants directly here? I'm in the process of doing that for CallArg in the monorep replacement in https://github.com/iotaledger/iota-rust-sdk/pull/595 already
```Rust
pub enum Input {
    /// A move value serialized as BCS.
    ///
    /// For normal operations this is required to be a move primitive type and
    /// not contain structs or objects.
    Pure(Vec<u8>),
    /// A move object that is either immutable or address owned
    ImmutableOrOwned(ObjectReference),
    /// A move object whose owner is "Shared"
    Shared(SharedObjectReference),
    /// A move object that is attempted to be received in this transaction.
    // TODO add discussion around what receiving is
    Receiving(ObjectReference),
}
```

miker83z: Added here https://github.com/iotaledger/iota/issues/10270 

We'll have a V2 in which we will make these fields "richer" for the AuthContext. Making it flatten now, for this V1, would mean to add a custom serializer for AuthContextCallArgV1 because we currently cannot change the Move iota-framework representation of CallArg.

I imagine something like this could work once you import Input into the monorepo, no?
```rust
impl From<& Input > for AuthContextCallArg {
...
```

Thoralf-M: Does it really need to be in sync with the move representation? BCS encoding would still be same if you restructure it like this

valeriyr (approved):

PR #10469, opened 2026-02-25
feat(apps-ui-kit): add aria props

begonaalvarezd:
# Description of change

Add aria props to the ui kit for better usability to optimize it for Amplitude autocapture events


brancoder (approved):
LG but the changeset is missing

panteleymonchuk (approved):

PR #10471, opened 2026-02-25
feat(iota-indexer): replaces JSON RPC dry_run with gRPC simulate api

### LINKED ISSUE #10315: [indexer] replace `dev_inspect` & `dry_run` with gRPC

sergiupopescu199:
# Description of change

This PR replaces the iota-indexer JSON RPC `dry_run_transaction_block` with gRPC simulate api.

fixes #10315 

- Ran indexer tests, made sure that the `dry_run_transaction_block` test case passed.

 ```shell
```

> [!NOTE]
> indexer dev_inspect tests are expected to fails since they are not using grpc api, we have a dedicated issue to integrate dev_inspect with grpc simulate api. Same goes for the graphql tests.

### DIFF
--- a/crates/iota-indexer/src/apis/write_api.rs
+++ b/crates/iota-indexer/src/apis/write_api.rs
@@ -23,11 +23,15 @@ use iota_protocol_config::Chain;
 use iota_transaction_builder::TransactionBuilder;
 use iota_types::{
     base_types::IotaAddress,
-    effects::{TransactionEffects, TransactionEvents},
+    effects::{TransactionEffects, TransactionEffectsAPI, TransactionEvents},
+    error::ExecutionError,
     iota_serde::BigInt,
     quorum_driver_types::ExecuteTransactionRequestType,
     signature::GenericSignature,
-    transaction::{SenderSignedData, TransactionData},
+    transaction::{
+        GasData, SenderSignedData, TransactionData, TransactionDataV1, TransactionExpiration,
+        TransactionKind,
+    },
 };
 use jsonrpsee::{RpcModule, core::RpcResult, http_client::HttpClient};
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/iota-indexer/src/types.rs [hunk 1]
```diff
@@ -746,3 +746,39 @@ impl From<IotaTransactionBlockResponseWithOptions> for IotaTransactionBlockRespo
         }
     }
 }
+
+/// Provides conversion methods from gRPC types to iota core types.
+pub(crate) mod conversion {
```
tomxey: Now that those are moved to more general place, I think it would be nice for the name to mention that this is grpc related conversion. Right now it looks a bit too generic I think. (conversion::convert_objects,convert_events doesn't suggest what we are converting)

I think it would be nice to mention grpc either in module name or in function names.

crates/iota-indexer/src/types.rs
```diff
@@ -746,3 +746,39 @@ impl From<IotaTransactionBlockResponseWithOptions> for IotaTransactionBlockRespo
...
+    /// Converts [`GrpcTransactionEvents`] into [`TransactionEvents`]
+    pub(crate) fn convert_events(
+        tx_events: &GrpcTransactionEvents,
+    ) -> Option<IndexerResult<TransactionEvents>> {
+        // `GrpcTransactionEvents` is mostly created when we specifically request events
+        // via the read mask. When the transaction has no events, the `events` field is
+        // `None`.
+        // This part is inconsistent since it does not handle the case when `events`
+        // is `None` even if read mask requests the events. It should default to
+        // empty vec in order to make the tx_events.events()? not fail and return
```
tomxey: So this will be failing until it's fixed on the server side?

crates/iota-indexer/src/types.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10471/commits/4b67a6d272130bd0f6859059dc2944518a6f861a

crates/iota-indexer/src/types.rs
sergiupopescu199: Yes. The node team already pushed the fix to their feature branch. All our grpc-related feature branches were rebased onto their respective feature branches. The issue is fixed.

tomxey (approved):

crates/iota-indexer/src/apis/write_api.rs
```diff
@@ -287,3 +295,149 @@ impl IotaRpcModule for OptimisticWriteApi {
...
+        let readmask = FieldMask::from_paths(DRY_RUN_TRANSACTION_READ_MASK)
+            .display()
+            .to_string()
+            .pipe(Some);
+
+        let simulate_tx_response = client
+            .simulate_transaction(
+                transaction_data.clone().try_into()?,
+                false,
+                false,
```
tomxey: can this (estimate_gas_budget) be set to false if we are reading suggested_gas_price below? or are those independent?

sergiupopescu199: Yes, these are independent. `estimate_gas_budget` controls the total gas **budget** (how many NANOS the tx will cost) — when `true`, it runs an extra simulation pass to estimate and overwrite the transaction's budget. `suggested_gas_price` is the gas **price per unit** (driven by congestion), which is always returned as part of the simulation response. We don't need budget estimation here since the caller already provided a complete transaction with its own budget.

tomxey: go it, thanks 🙏

kodemartin (approved):

PR #10473, opened 2026-02-25
feat(types,json-rpc)!: expose effective commission rate for validator summary

### LINKED ISSUE #10305: [json-rpc]: expose `effectiveCommissionRate`

kodemartin:
# Description of change

This patch exposes the `effectiveCommissionRate` as part of `IotaValidatorSummary` following [IIP8](https://github.com/iotaledger/IIPs/blob/main/iips/IIP-0008/IIP-0008.md).

Though the change  breaks the Rust API for `IotaValidatorSummary` and by consequence for `IotaSystemStateSummary{,V1,V2}` the JSON representation is backwards compatible. As these types are supposed to represent the system state in the context of JSON-RPC, this is deemed a plausible change.

Nevertheless, the type has been used in various crates which are also affected.

Fixes #10305 

### Release Notes

- [x] JSON-RPC: Expose `effectiveCommissionRate` as part of the `IotaValidatorSummary` in `iotax_getLatestSystemStateV2` following [IIP8](https://github.com/iotaledger/IIPs/blob/main/iips/IIP-0008/IIP-0008.md)

### DIFF
--- a/crates/iota-indexer/src/models/system_state.rs
+++ b/crates/iota-indexer/src/models/system_state.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use fastcrypto::encoding::Base64;
+use iota_protocol_config::PROTOCOL_VERSION_IIP8;
 use iota_types::{
     base_types::{IotaAddress, ObjectID},
     iota_serde::{BigInt, Readable},
@@ -12,9 +13,6 @@ use iota_types::{
 };
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-
-/// Protocol version that IIP8 was put in effect.
-const PROTOCOL_VERSION_IIP8: u64 = 20;
 
 /// The representation of system state.
 #[non_exhaustive]
--- a/crates/iota-indexer/src/models/system_state.rs
+++ b/crates/iota-indexer/src/models/system_state.rs
@@ -1,0 +1,895 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use fastcrypto::encoding::Base64;
+use iota_types::{
+    base_types::{IotaAddress, ObjectID},
+    iota_serde::{BigInt, Readable},
+    iota_system_state::iota_system_state_summary::{
+        IotaSystemStateSummary, IotaSystemStateSummaryV1, IotaSystemStateSummaryV2,
+        IotaValidatorSummary,
+    },
+};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+/// Protocol version that IIP8 was put in effect.
+const PROTOCOL_VERSION_IIP8: u64 = 20;
+
+/// The representation of system state.
+#[non_exhaustive]
+#[derive(Debug, Deserialize, Serialize, Clone, derive_more::From)]
+pub enum StoredSystemState {
+    V1(StoredSystemStateV1),
+    V2(StoredSystemStateV2),
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct StoredSystemStateV1 {
+    /// The current epoch ID, starting from 0.
+    #[serde_as(as = "Readable<BigInt<u64>, _>")]
+    pub epoch: u64,
+    /// The current protocol version, starting from 1.
+    #[serde_as(as = "Readable<BigInt<u64>, _>")]
+    pub protocol_version: u64,
+    /// The current version of the system state data structure type.
+    #[serde_as(as = "Readable<BigInt<u64>, _>")]
+    pub system_state_version: u64,
+    /// The current IOTA supply.
+    #[serde_as(as = "Readable<BigInt<u64>, _>")]
+    pub iota_total_supply: u64,
+    /// The `TreasuryCap<IOTA>` object ID.
+    pub iota_treasury_cap_id: ObjectID,
+    /// The storage rebates of all the objects on-chain stored in the storage
+    /// fund.
+    #[serde_as(as = "Readable<BigInt<u64>, _>")]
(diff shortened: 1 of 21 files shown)


crates/iota-types/src/iota_system_state/mod.rs
```diff
@@ -332,6 +332,7 @@ pub fn get_validator_from_table<K>(
     object_store: &dyn ObjectStore,
     table_id: ObjectID,
     key: &K,
+    protocol_version: Option<u64>,
```
Alex6323, line 335: does anthing speak against `protocol_version: impl Into<Option<u64>>` here so it's a bit more convient for the caller?

crates/iota-types/src/iota_system_state/mod.rs
```diff
@@ -353,7 +354,7 @@ where
                             "Failed to load inner validator from the wrapper: {err:?}"
                         ))
                     })?;
-            Ok(validator.into_iota_validator_summary())
+            Ok(validator.into_iota_validator_summary(protocol_version))
```
bingyanglin, line 357: The callee `into_iota_validator_summary` does no support the protocol_version param yet?

crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
```diff
@@ -41,6 +41,9 @@ const E_METADATA_INVALID_NET_ADDR: u64 = 4;
 const E_METADATA_INVALID_P2P_ADDR: u64 = 5;
 const E_METADATA_INVALID_PRIMARY_ADDR: u64 = 6;
 
+/// Protocol version that IIP8 took effect.
+const PROTOCOL_VERSION_IIP8: u64 = 20;
```
bingyanglin: In `crates/iota-graphql-rpc/src/types/validator.rs we call it `IIP8_PROTOCOL_VERSION`, and in `crates/iota-indexer/src/models/system_state.rs` we call it `PROTOCOL_VERSION_IIP8` again. Do we need to make them consistent? Also do we have specific consideration to spread the same constant to three different places?

crates/iota-types/src/iota_system_state/iota_system_state_summary.rs
```diff
@@ -891,9 +891,20 @@ pub struct IotaValidatorSummary {
     #[schemars(with = "BigInt<u64>")]
     #[serde_as(as = "Readable<BigInt<u64>, _>")]
     pub gas_price: u64,
+    /// The fee set by the validator for providing staking services.
+    ///
+    /// This might be overriden by the protocol, that uses instead
```
bingyanglin:
```suggestion
    /// This might be overridden by the protocol, that uses instead
```

crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
kodemartin: good point! will consolidate

crates/iota-types/src/iota_system_state/mod.rs
kodemartin: what do you mean?

crates/iota-types/src/iota_system_state/iota_system_state_inner_v1.rs
kodemartin: 04e0d62004

crates/iota-types/src/iota_system_state/mod.rs
kodemartin: On occasion you have to extend the trait bounds with `Copy, Send, 'static` which is more implicit than pure `Option<u64>`.

I would prefer sticking with the explicit signature for that reason. Is there any convenience for the caller other than passing the raw `u64` as opposed to `Some(64)`?

crates/iota-types/src/iota_system_state/mod.rs
bingyanglin: In the file `crates/iota-types/src/iota_system_state/simtest_iota_system_state_inner.rs` the function `into_iota_validator_summary` does not have the `protocol_version` as the function parameter?

kodemartin: oh, missed that...https://github.com/iotaledger/iota/pull/10473/commits/a2ae1c4a8ba778d14c21f1301ef6f7ea168ec12e

bingyanglin (approved):

muXxer (approved):

crates/iota-types/src/iota_system_state/mod.rs
Alex6323: No, I think not really. And there are some downsides to it, one of which is what you mentionend.

Alex6323 (approved):

crates/iota-indexer/src/models/system_state.rs
```diff
@@ -0,0 +1,893 @@
...
+        }
+    }
+}
+
+impl From<IotaSystemStateSummary> for StoredSystemState {
+    fn from(native: IotaSystemStateSummary) -> Self {
+        match native {
+            IotaSystemStateSummary::V1(inner) => StoredSystemState::V1(inner.into()),
+            IotaSystemStateSummary::V2(inner) => StoredSystemState::V2(inner.into()),
+            _ => unimplemented!(),
```
tomxey: as this in theory can happen when introducing the v3 and so on, did you consider returning an Err from this function? (and to convert it to TryFrom)

kodemartin: any operators would need to update the application eitherway, otherwise the native system state won't be able to deserialize from the checkpoint data.

We should thus make sure that we cover this branch before shipping, and panicking implies that we exit the application on the point we should handle.

In the same sense, even if this would be a `TryFrom` we should `expect` it in any caller. 

Added a more explicit message: ceb4dcdff4

tomxey (approved):

valeriyr (approved):

PR #10475, opened 2026-02-25
feat(natives): use global value for AuthContext

miker83z:
# Description of change

In order to return reference values as result of native functions, the AuthContext should be handled as GlobalValue. This PR makes a refactoring to enable that.


iota-execution/latest/iota-move-natives/src/auth_context.rs
```diff
@@ -189,20 +199,6 @@ pub fn native_replace(
...
-        PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
-            .with_message(format!("Failed to serialize an input: {err}"))
-    })?;
-    Value::simple_deserialize(&bytes, input_move_layout).ok_or_else(|| {
-        PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
-            .with_message("Failed to deserialize an input to a Move value".to_string())
-    })
-}
-
 fn from_value<T: DeserializeOwned>(
```
miker83z: this could be updated as well now that we have defined MoveTypeLayouts

crates/iota-framework/packages/iota-framework/sources/auth_context.move
```diff
@@ -25,25 +25,25 @@ public struct AuthContext has drop {}
...
-    native_tx_commands()
+public fun tx_inputs(_actx: &AuthContext): &vector<CallArg> {
+    native_tx_inputs()
 }
 
-public fun tx_inputs(_: &AuthContext): vector<CallArg> {
-    native_tx_inputs()
+public fun tx_commands(_actx: &AuthContext): &vector<Command> {
+    native_tx_commands()
 }
```
valeriyr, line 38: Nit: I changed the functions' order to have them sorted.
Do we really need to have a parameter name `_actx` here?

crates/iota-framework/packages/iota-framework/sources/auth_context.move
```diff
@@ -25,25 +25,25 @@ public struct AuthContext has drop {}
...
 // === Native functions ===
 
-native fun native_digest(): vector<u8>;
+native fun native_digest(): &vector<u8>;
 
-native fun native_tx_commands<C>(): vector<C>;
+native fun native_tx_inputs(): &vector<CallArg>;
 
-native fun native_tx_inputs<I>(): vector<I>;
+native fun native_tx_commands(): &vector<Command>;
```
valeriyr: If we decide to change the signatures, we need to fix the following comments as well:
https://github.com/iotaledger/iota/blob/bfef62e654d6811eb0302086bee892271a97d739/crates/iota-protocol-config/src/lib.rs#L1255

crates/iota-types/src/auth_context/fields_v1.rs
```diff
@@ -83,6 +110,60 @@ impl From<&Command> for AuthContextCommand {
...
+    ///     MoveCall(ProgrammableMoveCall),
+    ///     TransferObjects(TransferObjectsData),
+    ///     SplitCoins(SplitCoinsData),
+    ///     MergeCoins(MergeCoinsData),
+    ///     Publish(PublishData),
+    ///     MakeMoveVec(MakeMoveVecData),
+    ///     Upgrade(UpgradeData),
+    /// }
+    /// ```
+    pub fn layout() -> MoveEnumLayout {
```
valeriyr: I tried to avoid creating Move layouts manually since it requires move code and makes it harder to support(we literally duplicate data structures), but it can be an option if it is really necessary, of course.

iota-execution/latest/iota-move-natives/src/authentication_context.rs
```diff
@@ -29,45 +34,170 @@ impl AuthenticationContext {
...
+            MoveAuthenticatorDigest::try_from(auth_digest_value.as_slice()).map_err(|err| {
+                PartialVMError::new(StatusCode::UNEXPECTED_DESERIALIZATION_ERROR)
+                    .with_message(err.to_string())
+            })?;
+
         self.auth_context
             .borrow_mut()
             .replace(auth_digest, tx_inputs, tx_commands);
+
         Ok(())
```
valeriyr, line 220: Does it make sense to drop the caches here?

crates/iota-framework/packages/iota-framework/sources/auth_context.move
miker83z: About the order, I tried to reduce the modifications from the original module. About the parameter name, I used the same approach as in TxContext changes in Sui

crates/iota-types/src/auth_context/fields_v1.rs
miker83z: Reverted

valeriyr (approved):

PR #10485, opened 2026-02-26
feat(protocol-research): set OGD as a default prediction algorithm

vekkiokonio:
# Description of change

- After experimental evaluation, we updated the following parameters to provide a more accurate prediction:
  - `HOTNESS_ADJUSTMENT_FACTOR: f64 = 2.0`
  - `MAX_DECAY_FACTOR: f64 = 1.1;`
- We improved OGD calculation by excluding system and read-only transactions from `clearing_txs_data`.
- We set OGD as the default gas price prediction algorithm. To avoid modifying `authority.rs`, we introduced this change by renaming the prediction function instead of altering the DryRun logic which is currently being restructured as part of the gRPC refactor (see https://github.com/iotaledger/iota/pull/10471).

_Note: we keep the legacy predictor renamed as `get_prediction_suggested_gas_price_legacy`._

fixes https://github.com/iotaledger/iota-private/issues/316.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Set OGD as the default algorithm for gas price prediction during dry run.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


crates/iota-core/src/congestion_tracker.rs
```diff
@@ -203,7 +210,10 @@ impl CongestionTracker {
     /// get the highest minimum clearing price, if any exists. The 'clearing'
     /// gas price means the underlying transaction was not cancelled due
     /// congestion.
-    pub fn get_prediction_suggested_gas_price(&self, transaction: &TransactionData) -> Option<u64> {
+    pub fn get_prediction_suggested_gas_price_legacy(
```
bingyanglin, line 214: Maybe `#[allow(dead_code)]`?

crates/iota-core/src/congestion_tracker.rs
```diff
@@ -215,7 +225,7 @@ impl CongestionTracker {
 
     /// Get the largest hotness value among all mutable input shared objects
     /// accessed by `transaction`.
-    pub fn get_suggested_gas_price_with_ogd(&self, transaction: &TransactionData) -> u64 {
+    pub fn get_prediction_suggested_gas_price(&self, transaction: &TransactionData) -> Option<u64> {
```
bingyanglin, line 229: Just curious why we need this to be Optional?

piotrm50 (approved):

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: Good point

crates/iota-core/src/congestion_tracker.rs
vekkiokonio: It doesn't. I'm making Optional to keep exactly the same signature as it is right now with the default implementation in order to avoid modifications to other crates (e.g., DryRun function in authority.rs).

salaheldinsoliman (approved):

bingyanglin (approved):

PR #10488, opened 2026-02-26
feat(gRPC): cleanup server side error handling

muXxer:
# Description of change

This PR removes the anti-pattern where `format!("{e}") → String → RpcError` happened, which lost error codes and type info via the From<String> catch-all.

The `Merge` trait now has an associated error type.

### DIFF
--- a/crates/iota-grpc-server/src/error.rs
+++ b/crates/iota-grpc-server/src/error.rs
@@ -20,22 +20,29 @@ pub struct RpcError {
 }
 
 impl RpcError {
-    pub fn new<T: Into<String>>(code: Code, message: T) -> Self {
+    pub fn new<T: std::fmt::Display>(code: Code, message: T) -> Self {
         Self {
             code,
-            message: Some(message.into()),
+            message: Some(message.to_string()),
             details: None,
         }
     }
 
     /// Add context to an existing error
-    pub fn with_context<T: Into<String>>(mut self, context: T) -> Self {
-        let context_str = context.into();
+    pub fn with_context<T: std::fmt::Display>(mut self, context: T) -> Self {
         self.message = Some(match self.message {
-            Some(existing) => format!("{}: {}", context_str, existing),
-            None => context_str,
+            Some(existing) => format!("{}: {}", context, existing),
+            None => context.to_string(),
         });
         self
+    }
+
+    pub fn internal() -> Self {
+        Self {
+            code: Code::Internal,
+            message: None,
+            details: None,
+        }
     }
 
     pub fn not_found() -> Self {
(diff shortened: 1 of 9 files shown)


crates/iota-grpc-server/src/ledger_service/get_epoch.rs
```diff
@@ -219,9 +219,9 @@ impl std::fmt::Display for CommitteeNotFoundError {
 
 impl std::error::Error for CommitteeNotFoundError {}
 
-impl From<CommitteeNotFoundError> for Status {
+impl From<CommitteeNotFoundError> for RpcError {
     fn from(value: CommitteeNotFoundError) -> Self {
-        Status::not_found(value.to_string())
+        RpcError::new(tonic::Code::NotFound, value.to_string())
```
semenov-vladyslav:
```suggestion
        RpcError::not_found().with_context(value)
```
This will work if the context type impls `Display`.

crates/iota-grpc-server/src/error.rs
```diff
@@ -28,6 +28,16 @@ impl RpcError {
...
 
+    /// Add context to an existing error
+    pub fn with_context<T: Into<String>>(mut self, context: T) -> Self {
+        let context_str = context.into();
+        self.message = Some(match self.message {
+            Some(existing) => format!("{}: {}", context_str, existing),
+            None => context_str,
+        });
+        self
+    }
```
semenov-vladyslav:
```suggestion
    pub fn with_context<T: std::fmt::Display>(mut self, context: T) -> Self {
        self.message = Some(match self.message {
            Some(existing) => format!("{}: {}", context, existing),
            None => context.to_string(),
        });
        self
    }
```
Types tend to implement `ToString`/`Display` rather than `Into<String>`.
Same for `RpcError::new`.

crates/iota-grpc-server/src/merge.rs
```diff
@@ -206,31 +215,33 @@ impl Merge<&UserSignatures> for UserSignatures {
...
+    ) -> Result<(), Self::Error> {
         // Use mask directly — Events is a transparent wrapper
         // TransactionEvents is a tuple struct with Vec<Event> at index 0
         self.events = source
             .0
             .iter()
-            .map(|event| -> Result<_, Box<dyn std::error::Error>> {
-                Merge::merge_from(event, mask)
-            })
+            .map(|event| Merge::merge_from(event, mask))
```
semenov-vladyslav:
```suggestion
            .map(|event| Event::merge_from(event, mask).map_err(|e| e.with_context("failed to merge event")))
```

crates/iota-grpc-server/src/types.rs
```diff
@@ -591,11 +591,11 @@ impl GrpcReader {
 
             // Use Merge to populate based on mask
             Merge::merge(&mut checkpoint_proto, &sdk_summary, &checkpoint_mask)
-                .map_err(|e| Status::internal(format!("merge error for summary: {e}")))?;
+                .map_err(|e: RpcError| Status::from(e.with_context("failed to merge summary")))?;
```
semenov-vladyslav:
```suggestion
                .map_err(|e| e.with_context("failed to merge summary"))?;
```
No need for `Status::from`, the `?` operator knows the target type.

semenov-vladyslav (approved):

crates/iota-grpc-server/src/ledger_service/get_epoch.rs
```diff
@@ -100,15 +98,17 @@ impl Merge<&EpochReadSource> for Epoch {
...
             } else if let Some(ref info) = epoch_info {
                 info.system_state.clone()
             } else {
-                return Err(format!(
-                    "Cannot get system state for historical epoch {}: epoch info not available",
-                    source.epoch
-                )
-                .into());
+                return Err(RpcError::new(
+                    tonic::Code::Internal,
```
semenov-vladyslav: Should there be a constructor similar to `RpcError::not_found` instead:
```rs
impl RpcError {
    pub fn internal() -> Self {
        Self {
            code: Code::Internal,
            message: None,
            details: None,
        }
    }
}
```
Then this part could be rewritten as `RpcError::internal().with_context(...)`.

semenov-vladyslav (approved):

PR #10496, opened 2026-02-26
fix(tx-checks): set max_auth_gas during authenticator signing phase as gas budget

miker83z:
# Description of change

Protocol's parameter `max_auth_gas` was only added to the transaction budget during the setup of the IotaGasStatus (used during VM execution). With this change, `max_auth_gas` (or the param `authentication_gas_budget` if lower) is used as gas budget for the execution of the authenticator function pre-consensus (signing phase).

### DIFF
--- a/crates/iota-transaction-checks/src/lib.rs
+++ b/crates/iota-transaction-checks/src/lib.rs
@@ -59,15 +59,13 @@ mod checked {
         if transaction.is_system_tx() {
             Ok(IotaGasStatus::new_unmetered())
         } else {
-            let transaction_gas_budget = transaction.gas_budget();
-
             check_gas(
                 objects,
                 protocol_config,
                 reference_gas_price,
                 gas,
                 transaction.gas_price(),
-                transaction_gas_budget,
+                transaction.gas_budget(),
                 authentication_gas_budget,
                 is_execute_transaction_to_effects,
             )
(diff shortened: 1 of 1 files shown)


crates/iota-transaction-checks/src/lib.rs
```diff
@@ -421,11 +426,29 @@ mod checked {
...
     ) -> IotaResult<IotaGasStatus> {
-        let gas_status =
-            IotaGasStatus::new(gas_budget, gas_price, reference_gas_price, protocol_config)?;
+        let authentication_gas_budget =
+            authentication_gas_budget.min(protocol_config.max_auth_gas());
+
+        // To be sure that we can cover the Move authenticator + transaction execution
+        // gas cost.
+        let gas_budget_to_set = if is_execute_transaction_to_effects {
+            transaction_gas_budget + authentication_gas_budget
```
miker83z:
```suggestion
            transaction_gas_budget
```
Implementing this suggestion will make it so that the gas budget set by the user will cover both authentication and TX execution. 

Currently, instead, the user sets the TX budget, and it is invisible to them that the real gas budget is TX budget + max_auth_gas.

@valeriyr @lzpap

lzpap: Most app logic uses the gas budget as the threshold to gather gas coins. If the real gas budget is what the user sets plus an invisible amount, then it can lead to aborted transactions.

I'd rather set the actual budget to what the user sets.

miker83z: I think the same. 
Still, for AA accounts, wallets should suggest additional gas budget to cover for `max_auth_gas`. So I imagine there should be a mechanism for these wallets like: `gas_budget = dry_run_estimation + wiggle_room (+ if AA then max_auth_gas)`.

miker83z: Changed here [73f39ad](https://github.com/iotaledger/iota/pull/10496/commits/73f39add6938ab18e973f17e193c6b95d2376a72)

lzpap (approved):

valeriyr (approved):

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -54,23 +54,20 @@ mod checked {
         reference_gas_price: u64,
         transaction: &TransactionData,
         authentication_gas_budget: u64,
+        is_execute_transaction_to_effects: bool,
```
Dkwcs, line 57: Do I get it correctly that with this flag we determine gas metering phase either signing or execution?
If yes , can we leave appropriate comment or even change bool to explicit enum values?

miker83z: The comment is written right in the only place where this parameter is used. I think the enum would be overkill in this case.

alexsporn (approved):

PR #10526, opened 2026-02-27
fix(consensus): fix order of fetched headers

VorobyevIlya:
# Description of change

When fetching headers or blocks from store and dag_state, it was possible to change their order, while the validator requesting headers or blocks assumed that the order matches the order of block references. PR fixes this by preserving order even when headers or blocks are fetched from two sources.


polinikita (approved):
I would also add a unit test checking this consistency such that it could detect a potential bug if we change something in the future.

piotrm50 (approved):

PR #10554, opened 2026-03-02
feat(amplitude): improve amplitude source flow

brancoder:
# Description of change

improves amplitude source flow tracking and removes unused create account mutation hooks


evavirseda (changes requested):
conflicts 🙏

apps/wallet/src/ui/app/pages/accounts/AddAccountPage.tsx
```diff
@@ -79,8 +78,15 @@ export function AddAccountPage() {
...
     const [isConnectLedgerModalOpen, setConnectLedgerModalOpen] = useState(forceShowLedger);
-    const createAccountsMutation = useCreateAccountsMutation();
-    const sourceFlow = searchParams.get('sourceFlow') || 'Unknown';
+    const { sourceFlowRef, setSourceFlow } = useSourceFlow();
+
+    // Bootstrap sourceFlow from URL when opened in a new tab (e.g. Ledger popup flow)
+    const urlSourceFlow = searchParams.get('sourceFlow');
+    if (urlSourceFlow) {
+        setSourceFlow(urlSourceFlow);
+    }
```
panteleymonchuk: I think better to use `useEffect` here, so we don't execute it every render.

Maybe even it makes sense to have generic hook that track if sourceFlow exists in the url params and set it to context.

apps/wallet/src/ui/app/pages/accounts/CreateNewPasskey.tsx
```diff
@@ -24,6 +24,15 @@ type ImportPasskeyFormValues = z.infer<typeof formSchema>;
...
 export function CreateNewPasskey() {
     const navigate = useNavigate();
+    const [searchParams] = useSearchParams();
+    const { setSourceFlow } = useSourceFlow();
+
+    // Bootstrap sourceFlow from URL when opened in a new tab (passkey popup flow)
+    const urlSourceFlow = searchParams.get('sourceFlow');
+    if (urlSourceFlow) {
+        setSourceFlow(urlSourceFlow);
+    }
```
panteleymonchuk: code duplication. Like previous comment, maybe better to use some generic solution.

apps/wallet/src/ui/app/hooks/useCreateAccountsMutation.ts
panteleymonchuk, line 182: I'd reset sourceFlow after success added account here.

apps/wallet/src/ui/app/pages/accounts/CreateNewWallet.tsx
```diff
@@ -73,7 +72,7 @@ export function CreateNewWallet() {
             case AccountsFormType.Passkey:
                 const url = '/accounts/passkey-account';
                 if (isPopupOrSidePanel) {
-                    openInNewTab(url);
+                    openInNewTab(`${url}?sourceFlow=${sourceFlow}`);
                     window.close();
                 } else {
                     navigate(url);
```
panteleymonchuk, line 72: ImportPasskey page expects to have sourceFlow query param. I think we need to either add it here by `setSourceFlow` or pass as query param
```
navigate(`${url}?sourceFlow=${sourceFlow}`);
```

apps/wallet/src/ui/app/components/accounts/AccountsFormContext.tsx
```diff
@@ -59,6 +60,8 @@ export type AccountsFormValues =
 type AccountsFormContextType = [
     MutableRefObject<AccountsFormValues>,
     (values: AccountsFormValues) => void,
+    MutableRefObject<string>,
+    (sourceFlow: string) => void,
```
panteleymonchuk:
```suggestion
    MutableRefObject<AmpliSourceFlow >,
    (sourceFlow: AmpliSourceFlow) => void,
```

panteleymonchuk (changes requested):

apps/wallet/src/ui/app/hooks/useCreateAccountsMutation.ts
```diff
@@ -33,23 +38,19 @@ function ensurePassword(password: string | undefined): string {
...
-            type,
-            password,
-            sourceFlow,
-        }: {
-            type: AccountsFormType;
-            password?: string;
-            sourceFlow: string;
-        }) => {
+        onSuccess: () => {
+            clearSourceFlow();
```
evavirseda: quick question: shouldnt be add this `clearSourceFlow` onError too?

evavirseda (changes requested):

apps/wallet/src/ui/app/hooks/useCreateAccountsMutation.ts
brancoder: hmm, what if user retries after error than sourceFlow would be Unknown? wdyt?

evavirseda: good point, makes sense to keep it in `onSuccess` then!

evavirseda (approved):

panteleymonchuk (approved):
🚀

PR #10557, opened 2026-03-02
refactor(starfish): block suspension in block manager

### LINKED ISSUE #10241: starfish(bug): Refactor the fix for block manager memory leak

filipdulic:
# Description of change

refactors block manager code to be more readable and efficient.
- renames suspended_blocks to suspended_transactions
- updates processing logic to read the dag state in a single call and use the result in subsequent processing.

fixes #10241.

### DIFF
--- a/crates/starfish/core/src/block_manager/mod.rs
+++ b/crates/starfish/core/src/block_manager/mod.rs
@@ -44,7 +44,7 @@ pub(crate) struct BlockManager {
     /// Keeps full blocks for suspended block headers
     /// TODO: this set can grow to become too big, need to add some eviction
     /// mechanism
-    suspended_blocks: BTreeMap<BlockRef, VerifiedBlock>,
+    suspended_transactions: BTreeMap<BlockRef, VerifiedTransactions>,
     block_suspender: BlockSuspender,
     /// A vector that holds a tuple of (lowest_round, highest_round) of received
     /// blocks per authority. This is used for metrics reporting purposes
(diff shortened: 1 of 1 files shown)


crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -337,23 +347,49 @@ impl BlockManager {
...
+            .into_iter()
+            .zip(present_flags)
+            .filter_map(|(block_ref, found)| found.then_some(block_ref))
+            .collect()
     }
 
     fn find_missing_ancestors(
         &self,
         incoming_headers: Vec<VerifiedBlockHeader>,
+        present_header_and_ancestor_refs_in_dag_states_in_dag_state: &BTreeSet<BlockRef>,
```
VorobyevIlya: in_dag_states_in_dag_state extra in_dag_state

VorobyevIlya (approved):

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -41,10 +41,8 @@ pub(crate) struct BlockManager {
     context: Arc<Context>,
     dag_state: Arc<RwLock<DagState>>,
 
-    /// Keeps full blocks for suspended block headers
-    /// TODO: this set can grow to become too big, need to add some eviction
```
piotrm50, line 45: Why is this TODO removed? I think we should keep it as long as this is not properly evicted.

piotrm50 (approved):

crates/starfish/core/src/block_manager/mod.rs
filipdulic: resolved in commit amendment.

crates/starfish/core/src/block_manager/mod.rs
filipdulic: resolved in commit amendment

PR #10559, opened 2026-03-02
refactor(explorer): remove ampli event tracking and replace with setAmplitudeIdentity function

panteleymonchuk:


panteleymonchuk:
we'll continue with this: https://github.com/iotaledger/iota/pull/10620

PR #10564, opened 2026-03-03
chore(wallet): cleanup amplitude events

### LINKED ISSUE #10283: Cleanup legacy web events & update amplitude generated file
After the next Wallet prod release:
- remove source `web` from all deprecated (merged) events and `viewedLedgerTutorial`
- regenerate ampli/index.ts

### LINKED ISSUE #10357: Review existing amplitude events trigger behavior
We need to review existing analytics events to ensure they are firing correctly and not being triggered multiple times unintentionally

For each event, confirm that:

- it fires only when the expected user action occurs
- it doesn't trigger on unrelated renders, navigation changes, refreshes, or background reloads
- it does not fire multiple times due to re-renders or state changes
- its name accurately reflects its actual behavior.

Example:
`opened wallet extension` → this event is triggered on navigation and may fire multiple times (e.g. due to `activeOrigin` changes or re-renders)

panteleymonchuk:
Closes: #10283
Closes: #10357


investigate-multiple-firings.md
evavirseda, line 1: we dont need this file 🫣

evavirseda (changes requested):

investigate-multiple-firings.md
panteleymonchuk: accidently commited it

evavirseda (approved):

apps/wallet/src/shared/analytics/ampli/index.ts
```diff
@@ -68,6 +68,10 @@ export type LoadOptions =
     | LoadOptionsWithApiKey
     | LoadOptionsWithClientInstance;
 
+export interface IdentifyProperties {
+    network?: string;
```
brancoder, line 72: do we need to update `IdentifyProperties` to add propas that are set in amplitude.ts
```
    walletAppMode?: string;   
    walletVersion?: string;  
```

```

    const identifyEvent = new Identify();
    identifyEvent.set('network', getNetworkName(options.network, options.customRpc));
    identifyEvent.set('walletAppMode', options.extensionViewType);
    identifyEvent.set('walletVersion', Browser.runtime.getManifest().version);
```

brancoder (approved):
LG

Issue #9925, opened 2026-01-26
gRPC: Decouple with JSON-RPC and REST API in gRPC implementation


bingyanglin:
- The logic of interaction between JSON-RPC and indexer can be reused, and gRPC can leverage it, if the logic is the same, consider sharing them (e.g., common operation flow or conversion flow)
- The logic of interaction between JSON-RPC and node should be decoupled with gRPC, if the logic is the same, consider sharing them (e.g., common operation flow or conversion flow)
- The logic of gRPC API should be fully decoupled with the REST API, design our own logics (like state reader)

bingyanglin:
After investigation I think we need to do the followings

- Currently in gRPC we still use `RestStateReader` and `RestIndexes` which are defined in `iota-types`. They are used by REST-API, gRPC, simulacrum, GraphQL, and the node. But actually they are NOT REST-specific, which is misleading. We can rename it to be the following in `iota-types`:
  - `RestStateReader` -> `StateReader `
  - `RestIndexes` -> `StateIndexes`
- The `RestReadStore` in `iota-core/src/storage.rs` wraps `AuthorityState` and `RocksDbStore`, and it's the node's implementation of the state reader, NOT REST-specific, same for `RestIndexStore` in `iota-core/src/rest_index.rs`. It makes sense to rename them to be the following in `iota-core`
  - `RestReadStore` -> `NodeReadStore`
  - `RestIndexStore`-> `NodeIndexStore`
    - Note that in `get_epoch_info()` and `get_transaction_info()` for gRPC server we need them via `StateReader.indexes()`
- After the renaming, the `GrpcStateReader` trait and `RestStateReaderAdapter` struct in `iota-grpc-server` can be deleted (which were just for delegation based on the naming consideration), and we can directly make `GrpcReader` holds `Arc<dyn StateReader> directly instead of `Arc<dyn GrpcStateReader`. Same for the `from_rest_state_reader()` and related logics which can be deleted.
- Now the `iota-core` and `iota-grpc-server` depends on `iota-json-rpc-types` because we need their `IotaEvent` for subscription, however we need to actually use `iota-sdk_types::Event`, in this wasy we decouple both `iota-core` and `iota-grpc-server` from `iota-json-rpc-types`
[...]

PR #10081, opened 2026-02-04
feat(iota-{data-ingestion, indexer, graphql}): enable usage of gRPC checkpoint stream

sergiupopescu199:
# Description of change

This PR refactors the infra related crates to use the fullnode's gRPC API as primary source for checkpoint ingestion replacing the REST API, which is subject for deprecation.

- `iota-data-ingestion-core`: integration of using gRPC checkpoint stream in the CheckpointReaderV2.
- `iota-data-ingestion`: All workers use the CheckpointReaderV2 by default and expect and gRPC connection for checkpoint sync.
- `iota-indexer`: ingest checkpoints and executes Write operation through JSON RPC using the gRPC API under the hood, tests were refactored accordingly.
- `iota-graphql-*`: refactored the code to support the gRPC connection in order to execute transaction.
- `docker & devtools`: update configs and files to use the gRPC API for infra related services.
- `docs`: updated the docs to take into account the new gRPC API requirements.

## Links to any relevant issues

fixes #7923 (already approved by @iotaledger/node)
fixes #7926 
fixes #10099
fixes #10316 
fixes #10315 
fixes #10477 (already approved by @iotaledger/dev-tools )
fixes #10648 (already approved by @iotaledger/devrel)
fixes #10760 (already approved by @iotaledger/node, @iotaledger/devops-admin)

## How the change has been tested

Ran local tests to assert that the gRPC url was used internally for the Write operation in the JSON-RPC calls, in the indexer case also the checkpoint ingestion is done through gRPC url, the data-ingestion-path is used in GraphQL cases.

- Indexer tests
```shell
cargo test -p iota-indexer --profile simulator --features shared_test_runtime
cargo nextest run -p iota-indexer --features pg_integration --test-threads 1
```
- GrpahQL tests
```shell
cargo nextest run -p iota-graphql-rpc --features pg_integration --no-fail-fast --test-threads 1
cargo nextest run -p iota-graphql-e2e-tests --features pg_integration
```
- Used `devnet` gRPC endpoint to sync checkpoints starting form genesis in the iota-indexer.
```shell
cargo r --release -- --database-url postgres://postgres:postgrespw@localhost:5432/iota_indexer  indexer --remote-store-url  https://grpc.devnet.iota.cafe --reset-db
```
once genesis was synced, took around 27 min, through the REST API it took around 30 min and sometimes more..

- Started the indexer again but without the --reset-db. All good, it resumed from latest synced checkpoint.

[...]

PR #10566, opened 2026-03-03
feat(node): REST API removal with storage cleanup

bingyanglin:
# Description of change

Remove the `iota-rest-api` crate entirely now that the gRPC server (`iota-grpc-server`) has fully replaced it. All REST-specific storage traits and index types are migrated to gRPC equivalents or extracted into a new lightweight crate.

- Delete `crates/iota-rest-api/` (all source, OpenAPI specs, and HTML docs)
- Introduce `crates/iota-node-storage/` — a thin crate defining `GrpcStateReader` and `GrpcIndexes` traits. This breaks what would otherwise be a circular dependency: `iota-core` needs to implement these traits while `iota-grpc-server` consumes them and already depends on `iota-core`. By extracting the trait definitions into a separate lightweight crate, both can depend on `iota-node-storage` without a cycle. This also lets `simulacrum` and other test harnesses implement the traits without pulling in production dependencies.
- Rename `rest_index.rs` → `grpc_indexes.rs` in `iota-core`, and rename `RestIndexStore` → `GrpcIndexesStore`, `RestReadStore` → `GrpcReadStore`, `RestStateReader` → `GrpcStateReader`, `RestIndexes` → `GrpcIndexes` throughout
- Move `GrpcStateReader` + `GrpcIndexes` trait definitions out of `iota-grpc-server/src/types.rs` and `iota-types/src/storage/read_store.rs` into the new `iota-node-storage` crate
- **Remove `RestStateReader` / `RestIndexes` traits and their associated types (`CoinInfo`, `AccountOwnedObjectInfo`, `DynamicFieldIndexInfo`, `DynamicFieldKey`) from `iota-types::storage`. The backfill readiness checks (`is_owner_v2_index_ready`, `is_coin_v2_index_ready`, `is_package_version_index_ready`) are preserved in the new `GrpcIndexes` trait so that upgrading nodes can populate new index tables in the background without forcing operators to perform a sequential two-step upgrade (first backfill, then enable queries).**
- Remove REST-specific node config fields (`enable_rest_api`) from `iota-config`
- Update `iota-node` startup to no longer mount the REST API router
- Remove `simulacrum/src/state_reader.rs` — `simulacrum` now implements the traits from `iota-node-storage` directly in `lib.rs`
- Update `iota-light-client`, `iota-data-ingestion-core`, `iota-graphql-rpc`, `iota-e2e-tests`, `iota-transactional-test-runner`, `simulacrum-server`, and `iota-tool` to drop REST dependencies and use the new trait paths
- Add legacy directory migration (`rest_index` → `grpc_indexes`) in `GrpcIndexesStore::migrate_legacy_dirs` so existing nodes upgrade seamlessly
- Remove REST-related snapshot test expectations and CODEOWNERS entries

Part of #9925 

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] REST API:  ⚠️⚠️⚠️ The REST API has been fully removed in this release. ⚠️⚠️⚠️


muXxer:
I would suggest several changes, after reading the current description of the PR:

- Do we even need the flag `enable_index_processing`? I think if a node provides the gRPC API, it should be able to process all requests. There is little value in having a node that doesn't have indexes, but provides the other endpoints. I get's even more confusing for clients, because they can only check if a node has a gRPC API, but not if they have the indexed data. And if we add that, the client gets more complicated. Not worth it IMO.
- The migration should move the  existing`rest_index` folder to the `grpc_indexes` folder, instead of keeping the old name. We can remove that migration logic after one release. 
- I wonder if we should call it `NodeIndexes` or `GrpcIndexes`. What if we add more API's in the future that have different indexes. On one hand we don't want to store the same information twice in different indexes, on the other hand it might be different information that needs to be stored. Maybe something to discuss.

crates/iota-core/src/authority.rs
```diff
@@ -3543,8 +3543,13 @@ impl AuthorityState {
             if let Some(indexes) = self.indexes.as_ref() {
                 indexes.checkpoint_db(&checkpoint_path_tmp.join("indexes"))?;
             }
-            if let Some(rest_index) = self.rest_index.as_ref() {
-                rest_index.checkpoint_db(&checkpoint_path_tmp.join("grpc_indexes"))?;
+            if let Some(node_index) = self.node_index.as_ref() {
+                // Snapshot directory name "grpc_indexes" is a legacy name that
```
muXxer: It's actually not a legacy name, my idea was to name it that way in the node. We can discuss that if you prefer the node_indexes.

dev-tools/iota-data-ingestion/README.md
```diff
@@ -28,8 +28,8 @@ An example of `config.yaml` for the Blob worker:
 # IndexerExecutor config
 #
 path: "./test-checkpoints"
-# IOTA Node Rest API URL
-remote-store-url: "http://localhost:9000/api/v1"
+# Remote object store URL (or gRPC URL for hybrid mode, see v2 reader)
+remote-store-url: "https://checkpoints.example.com"
```
muXxer: example.com?
why not keep it localhost?

crates/iota-core/src/node_index.rs
```diff
@@ -3,46 +3,51 @@
...
-};
+use crate::{authority::AuthorityStore, checkpoints::CheckpointStore};
 
+/// DB version stays at 1 because the schema of active tables (meta, watermark,
+/// epochs, transactions) has not changed. Deprecated tables (owner,
+/// dynamic_field, coin) are cleaned up automatically via `#[deprecated_db_map]`
+/// on the next DB open — no re-indexing is required.
+///
+/// Only increment this version when an active table's schema changes, as that
+/// triggers a full re-index of all checkpoints.
```
muXxer:
```suggestion
/// Only increment this version when an active table's schema changes, as that
/// triggers a full re-index of all checkpoints.
```

The rest of the comment we can keep here for reviewers, but doesn't need to be in the code:
```
DB version stays at 1 because the schema of active tables (meta, watermark,
epochs, transactions) has not changed. Deprecated tables (owner,
dynamic_field, coin) are cleaned up automatically via `#[deprecated_db_map]`
on the next DB open — no re-indexing is required.
```

crates/iota-core/src/node_index.rs
```diff
@@ -3,46 +3,51 @@
...
 const CURRENT_DB_VERSION: u64 = 1;
 
+/// On-disk directory name for the node index store.
+///
+/// This retains the historical name `"rest_index"` for backward compatibility
+/// with existing node databases. Renaming the directory on disk would require a
+/// migration step; for now we keep the legacy name so that nodes can upgrade
+/// without manual intervention.
+pub const NODE_INDEX_DIR: &str = "rest_index";
```
muXxer: I would rather do a one time migration now. If "rest_index" exists, move it to `grpc_indexes`, before the init of the tables. We can remove the migration logic after one version.

crates/iota-core/src/node_index.rs
```diff
@@ -56,57 +61,60 @@ pub enum Watermark {
     Pruned,
 }
 
+// Types retained only for the deprecated RocksDB column families below.
```
muXxer: Actually if we don't do a migration, we could maybe change the type of the DBMap, for example:

```diff
-owner: DBMap<OwnerIndexKey, OwnerIndexInfo>,
+#[allow(dead_code)]
+#[deprecated_db_map]
+owner: Option<DBMap<(), ()>>,
````

Can you try if that works? Then we could also directly get rid of the types.

File: crates/typed-store-derive/src/lib.rs [hunk 6]
```diff
@@ -21,8 +21,12 @@ const DEFAULT_DB_OPTIONS_CUSTOM_FN: &str = "typed_store::rocks::default_db_optio
...
 // Use a different name for the column than the identifier
 const DB_OPTIONS_RENAME: &str = "rename";
-// Deprecate a column family
-const DB_OPTIONS_DEPRECATE: &str = "deprecated";
+// Deprecate a column family with optional migration support
+// Usage: `#[deprecated_db_map]` or `#[deprecated_db_map(migration =
+// "migration_fn_path")]`
+// Hint: we can't use `#[deprecated]` because it doesn't allow us to specify a
+// migration parameter
+const DB_OPTIONS_DEPRECATED_TABLE: &str = "deprecated_db_map";
```
muXxer, line 28: Is that an old commit? The PR regarding deprecation that was merged refactored that file. This version doesn't support restarting of the node, because the tables will be missing on the second run. Also you need the `Option<DBMap<>>` syntax with the new commit. Maybe don't include the commit in this PR, but in your feature branch.

crates/iota-core/src/node_index.rs
```diff
@@ -56,57 +61,60 @@ pub enum Watermark {
...
+// The `#[deprecated_db_map]` attribute on their corresponding table entries
+// causes the column families to be dropped on the next DB open. These types
+// and the deprecated table entries can be fully removed once all nodes have
+// performed at least one startup cycle after this change (i.e. the CFs have
+// been cleaned up on disk).
```
muXxer:
```suggestion
```
I will leave it as a comment here, doesn't need to be in the code:
```
The `#[deprecated_db_map]` attribute on their corresponding table entries
causes the column families to be dropped on the next DB open. These types
and the deprecated table entries can be fully removed once all nodes have
performed at least one startup cycle after this change (i.e. the CFs have
been cleaned up on disk).
```

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -144,36 +152,28 @@ struct IndexStoreTables {
...
-    /// An index of Coin Types
-    ///
-    /// Allows looking up information related to published Coins, like the
-    /// ObjectID of its coorisponding CoinMetadata.
-    /// REST-API only
+    /// Deprecated: was used by the removed REST API for coin info queries.
+    #[allow(dead_code)]
+    #[deprecated_db_map]
     coin: DBMap<CoinIndexKey, CoinIndexInfo>,
-    // NOTE: Authors and Reviewers before adding any new tables ensure that they are either:
```
muXxer, line 472: I would leave that comment for future changes.

crates/iota-core/src/rest_index.rs
```diff
@@ -324,18 +306,11 @@ impl IndexStoreTables {
     fn index_checkpoint(
         &self,
         checkpoint: &CheckpointData,
-        resolver: &mut dyn LayoutResolver,
     ) -> Result<typed_store::rocks::DBBatch, StorageError> {
-        debug!(
```
muXxer, line 646: we can leave that debug print.

crates/iota-core/src/rest_index.rs
```diff
@@ -345,11 +320,6 @@ impl IndexStoreTables {
             )],
         )?;
 
-        debug!(
```
muXxer, line 665: that as well

crates/iota-core/src/node_index.rs
```diff
@@ -219,8 +217,8 @@ impl IndexStoreTables {
             .map(|c| c.saturating_add(1))
             .unwrap_or(0);
 
-        // Doing backfill requires processing objects so we have to restrict our
-        // backfill range to the range of checkpoints that we have objects for.
+        // Backfill reads input/output objects to populate TransactionInfo, so
+        // restrict the range to checkpoints that still have object data.
         let lowest_available_checkpoint =
```
muXxer: But actually, do we even access object data? Maybe we don't need to care about the `lowest_available_checkpoint_objects` anymore, if we don't index objects. Can you check that?

crates/iota-data-ingestion-core/src/reader/v1.rs
```diff
@@ -107,10 +107,15 @@ impl Default for ReaderOptions {
     }
 }
 
+/// Remote checkpoint store backends.
+///
+/// NOTE: Standalone fullnode (gRPC) mode is only available in the v2 reader
```
muXxer: But I think the v2 was created to support the gRPC. I think we shouldn't add it here, but maybe completely remove the file instead in favour of the v2? @sergiupopescu199 any take on this?

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -543,10 +545,10 @@ impl<P: ProgressStore> IndexerExecutor<P> {
 /// async fn main() {
 ///     let (executor, _) = setup_single_workflow(
 ///         CustomWorker,
-///         "http://127.0.0.1:9000/api/v1".to_string(), // fullnode REST API
-///         0,                                          // initial checkpoint number.
-///         5,                                          // concurrency.
-///         None,                                       // extra reader options.
+///         "https://checkpoints.example.com".to_string(), // remote object store URL
```
muXxer: I would leave the localhost example here, just replace the port.

crates/iota-e2e-tests/tests/rest.rs
```diff
@@ -1,399 +0,0 @@
...
-};
-use iota_sdk_types::{Argument, Command, TransactionExpiration, crypto::Intent};
-use iota_test_transaction_builder::make_transfer_iota_transaction;
-use iota_types::{
-    base_types::IotaAddress, effects::TransactionEffectsAPI, transaction::TransactionDataAPI,
-};
-use test_cluster::TestClusterBuilder;
-
-#[sim_test]
-async fn execute_transaction_transfer() {
```
muXxer: Maybe we should have these different transaction tests in the gRPC e2e tests?

crates/iota-grpc-server/src/event_filter.rs
```diff
@@ -137,17 +133,13 @@ impl TryFrom<iota_grpc_types::v0::filter::EventFilter> for EventFilter {
 }
 
 impl EventFilter {
-    pub fn matches_event(&self, state_reader: Arc<dyn GrpcStateReader>, item: &Event) -> bool {
+    pub fn matches_event(&self, item: &Event) -> bool {
```
muXxer, line 136: wow good catch, that's a leftover from some old event filters that needed the state reader for resolving I guess.

bingyanglin:
> I would suggest several changes, after reading the current description of the PR:
> 
> * Do we even need the flag `enable_index_processing`? I think if a node provides the gRPC API, it should be able to process all requests. There is little value in having a node that doesn't have indexes, but provides the other endpoints. I get's even more confusing for clients, because they can only check if a node has a gRPC API, but not if they have the indexed data. And if we add that, the client gets more complicated. Not worth it IMO.
> * The migration should move the  existing`rest_index` folder to the `grpc_indexes` folder, instead of keeping the old name. We can remove that migration logic after one release.
> * I wonder if we should call it `NodeIndexes` or `GrpcIndexes`. What if we add more API's in the future that have different indexes. On one hand we don't want to store the same information twice in different indexes, on the other hand it might be different information that needs to be stored. Maybe something to discuss.

Those are good points. I removed the `enable_index_processing` flag, renamed the folder to `node_indexes`, and added a one-time startup migration in [ad73590](https://github.com/iotaledger/iota/pull/10566/commits/ad73590b31c9ec7507105f89b3b694b86c4fd623). I added a `TO DISCUSS` section in the PR description covering the three items. The main trade-off is that lightweight/relay nodes will now pay a backfill cost at startup (an alternative would be tying index creation to `enable_grpc_api` instead, but we might also consider the indexes might be shared with future APIs or other usage).

muXxer (changes requested):
I am not done with the review yet, but here is a first batch of comments, so we can discuss.

crates/iota-core/src/node_index.rs
```diff
@@ -144,36 +156,28 @@ struct IndexStoreTables {
     /// the main database.
     transactions: DBMap<TransactionDigest, TransactionInfo>,
```
muXxer: Can you check this index? I can't find where the object types are needed after removing the rest-api. It might very well be that we only need the checkpoint in the TransactionInfo. But that means we should do a  migration of the old info, if that is the case. So a `transactions_v2` and a migration function that only keeps the checkpoint index. No need to bump the DB_VERSION, because that doesn't need re-indexing.

muXxer: So maybe it can be `transaction_checkpoints: DBMap<TxDigest, SequenceNumber>` or some other name.

crates/iota-core/src/authority.rs
```diff
@@ -796,7 +796,7 @@ pub struct AuthorityState {
     execution_lock: RwLock<EpochId>,
 
     pub indexes: Option<Arc<IndexStore>>,
-    pub rest_index: Option<Arc<RestIndexStore>>,
+    pub node_index: Option<Arc<NodeIndexStore>>,
```
muXxer: Let's rename everywhere to GrpcIndexesStore as discussed.

crates/iota-light-client/src/verifier.rs
```diff
@@ -93,6 +93,10 @@ pub async fn get_verified_object(config: &Config, object_id: ObjectID) -> Result
     Ok(object)
 }
 
+// TODO: Consider accepting pre-built clients (IotaClient /
```
muXxer: I would say that is something that the codeowners should take care of. I wouldn't add this comment here.

crates/iota-node-storage/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+};
+
+/// Trait extending [`ReadStore`] with full-node-specific queries that may
+/// require richer databases or indexes to support.
+pub trait NodeStateReader: ObjectStore + ReadStore + Send + Sync {
+    /// Lowest available checkpoint for which object data can be requested.
+    ///
+    /// Specifically this is the lowest checkpoint for which input/output object
+    /// data will be available.
+    fn get_lowest_available_checkpoint_objects(&self) -> Result<CheckpointSequenceNumber>;
```
muXxer, line 35: Maybe not needed anymore after the cleanup?

crates/iota-rest-api/src/client/sdk.rs
```diff
@@ -1,741 +0,0 @@
...
-            .inner
-            .get(url)
-            .header(reqwest::header::ACCEPT, crate::APPLICATION_JSON)
-            .send()
-            .await?;
-
-        self.json(response).await
-    }
-
-    pub async fn health_check(&self, threshold_seconds: Option<u32>) -> Result<Response<()>> {
```
muXxer: we might need a health endpoint in the gRPC. something to discuss.

crates/iota-rest-api/src/error.rs
```diff
@@ -1,171 +0,0 @@
...
-impl From<bcs::Error> for RestError {
-    fn from(value: bcs::Error) -> Self {
-        Self {
-            status: StatusCode::INTERNAL_SERVER_ERROR,
-            message: Some(value.to_string()),
-        }
-    }
-}
-
-impl From<iota_types::quorum_driver_types::QuorumDriverError> for RestError {
```
muXxer, line 76: Should we also do such error conversion in gRPC for the quorum driver errors?

crates/iota-rest-api/src/health.rs
```diff
@@ -1,92 +0,0 @@
...
-}
-
-async fn health(
-    Query(Threshold { threshold_seconds }): Query<Threshold>,
-    State(state): State<StateReader>,
-) -> Result<impl IntoResponse> {
-    let summary = state.inner().try_get_latest_checkpoint()?;
-
-    // If we have a provided threshold, check that it's close to the current time
-    if let Some(threshold_seconds) = threshold_seconds {
```
muXxer, line 78: Here is the logic of the health check which we could adapt.

crates/iota-core/src/authority.rs
bingyanglin: As we discussed, I renamed constant to `GRPC_INDEX_DIR = "grpc_indexes"` and removed the "legacy name" comment in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/node_index.rs
bingyanglin: Trimmed to just the two suggested lines above `CURRENT_DB_VERSION` in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/node_index.rs
bingyanglin: Now `GRPC_INDEX_DIR = "grpc_indexes"` is the current name. Added a one-time `rest_index` -> `grpc_indexes` rename at startup before DB init in in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/node_index.rs
bingyanglin: Works. Changed all three deprecated tables to `Option<DBMap<(), ()>>` and deleted all six old type definitions. Added a dedicated `deprecate_type_erased_test` in `typed-store/tests/macro_tests.rs` that proves opening with `(), ()` over original types drops the CF cleanly in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/node_index.rs
bingyanglin: Removed. After simplifying the transaction index to checkpoint-only and removing owner/dynamic_field/coin indexing, backfill no longer reads object data, and `sparse_checkpoint_data_for_backfill` passes empty input_objects/output_objects. So the `lowest_available_checkpoint_objects` bound is no longer needed. Fixed at [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-e2e-tests/tests/rest.rs
bingyanglin: Added `execute_transaction_transfer_outputs` in `grpc/client/execute.rs` that verifies a transfer produces expected output objects (gas coin + recipient coin). Note that the other 5 deleted REST tests were all for the `resolve` endpoint which was REST-only so we don't need them. Fixed at [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-data-ingestion-core/src/executor.rs
bingyanglin: Changed to `"http://127.0.0.1:9000"` (dropped the removed `/api/v1` path) in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/typed-store-derive/src/lib.rs
bingyanglin: Sorry I used the old commits. I cherry picked the correct two PRs in our feature branch.

crates/iota-core/src/node_index.rs
bingyanglin: Removed them in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Restored them in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/rest_index.rs
bingyanglin: Restored it in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/rest_index.rs
bingyanglin: Restored it in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/node_index.rs
bingyanglin: Checked. Only `.checkpoint` is used after REST removal, `object_types` is unused. Replaced `transactions` with `transaction_checkpoints: DBMap<TransactionDigest, CheckpointSequenceNumber>` and added a migration function to copy the checkpoint field from the old table. No `DB_VERSION` bump needed since this doesn't require re-indexing. Fixed at [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -150,27 +87,67 @@ struct IndexStoreTables {
...
+        &typed_store::rocks::ReadWriteOptions::default(),
+        false,
+    )?;
+
+    let mut batch = new.batch();
+    for item in old.safe_iter() {
+        let (digest, info) = item?;
+        batch.insert_batch(&new, std::iter::once((digest, info.checkpoint)))?;
+    }
+    batch.write()?;
```
muXxer: That would cause OOM if we add all to the batch write, instead of writing from time to time I guess?

dev-tools/iota-data-ingestion/README.md
bingyanglin: Not sure why my previous replied comment not sent. I revert back to be localhost and dropped `/api/v1` in [9017d20](https://github.com/iotaledger/iota/pull/10566/commits/9017d20a14a2780804d210549686667813346bf3).

crates/iota-data-ingestion-core/src/executor.rs
```diff
@@ -543,10 +545,10 @@ impl<P: ProgressStore> IndexerExecutor<P> {
 /// async fn main() {
 ///     let (executor, _) = setup_single_workflow(
 ///         CustomWorker,
-///         "http://127.0.0.1:9000/api/v1".to_string(), // fullnode REST API
-///         0,                                          // initial checkpoint number.
-///         5,                                          // concurrency.
-///         None,                                       // extra reader options.
+///         "http://127.0.0.1:9000".to_string(), // remote store URL
```
muXxer: but maybe we should use the default port for gRPC then?

crates/iota-grpc-server/src/types.rs
```diff
@@ -1063,14 +786,19 @@ impl GrpcReader {
         };
 
         let (checkpoint, timestamp_ms) = if fields.include_checkpoint || fields.include_timestamp {
-            let checkpoint = self.state_reader.get_transaction_checkpoint(digest)?;
+            let checkpoint = match self.state_reader.grpc_indexes() {
```
muXxer: why should the grpc_indexes not be available?

crates/iota-light-client/src/config.rs
```diff
@@ -19,8 +19,14 @@ const CHECKPOINTS_FILE_NAME: &str = "checkpoints.yaml";
...
 pub struct Config {
-    /// An RPC endpoint to a full node.
+    /// A JSON-RPC endpoint to a full node.
     pub rpc_url: Url,
+    /// A gRPC endpoint to a full node. Falls back to `rpc_url` when not set,
+    /// which works when a reverse proxy routes both protocols on the same URL.
+    /// For local development the gRPC server typically listens on a separate
+    /// port (default 50051).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub grpc_url: Option<Url>,
```
muXxer: I wouldn't assume that we can simply take the JSON-RPC endpoint. I would treat it as the graphql_url.

crates/iota-light-client/src/config.rs
```diff
@@ -79,6 +85,12 @@ impl Config {
         Ok(())
     }
 
+    /// Returns the gRPC URL, falling back to `rpc_url` if `grpc_url` is not
+    /// set.
+    pub fn grpc_url(&self) -> &Url {
+        self.grpc_url.as_ref().unwrap_or(&self.rpc_url)
+    }
```
muXxer: I wouldn't assume that we can simply take the JSON-RPC endpoint.

crates/iota-light-client/src/config.rs
```diff
@@ -113,6 +125,8 @@ impl Config {
     fn create_config_from_network_name(network: &str) -> Self {
         Self {
             rpc_url: Url::parse(&format!("https://api.{network}.iota.cafe")).unwrap(),
+            // Known-network endpoints route gRPC via the same URL (reverse proxy).
+            grpc_url: None,
```
muXxer: Let's add a TODO here until we know the real endpoints.

crates/iota-light-client/src/checkpoint.rs
```diff
@@ -298,19 +298,22 @@ pub async fn sync_and_verify_checkpoints(config: &Config) -> anyhow::Result<()>
...
-                let summary = client
-                    .get_checkpoint_summary(seq)
+                let response = client
+                    .get_checkpoint_by_sequence_number(seq, Some("checkpoint"), None, None)
                     .await
                     .context(format!("Failed to download checkpoint summary '{seq}'"))?;
-
-                write_checkpoint_summary(config, &summary)?;
+                let checkpoint_data: iota_types::full_checkpoint_content::CheckpointData =
+                    response.checkpoint_data()?.try_into()?;
```
muXxer: Do we need the checkpoint data here? I think the summaries are enough. Also the readmask can then be more specific.

dev-tools/iota-data-ingestion/README.md
```diff
@@ -28,8 +28,8 @@ An example of `config.yaml` for the Blob worker:
 # IndexerExecutor config
 #
 path: "./test-checkpoints"
-# IOTA Node Rest API URL
-remote-store-url: "http://localhost:9000/api/v1"
+# Remote store URL
+remote-store-url: "http://localhost:9000"
```
muXxer: Should we also use the gRPC port here? @sergiupopescu199

crates/iota-swarm-config/src/node_config_builder.rs
```diff
@@ -259,6 +252,7 @@ impl ValidatorConfigBuilder {
             verifier_signing_config: VerifierSigningConfig::default(),
             enable_db_write_stall: None,
             iota_names_config: None,
+            enable_index_processing: true,
```
muXxer: We need to undo the changes to "enable_index_processing". It's somehow also removed from the snapshots.

crates/iota-swarm-config/tests/snapshots/snapshot_tests__network_config_snapshot_matches.snap
```diff
@@ -423,7 +408,6 @@ validator_configs:
       max-submit-position: ~
       submit-delay-step-override-millis: ~
       parameters: ~
-    enable-index-processing: true
```
muXxer, line 426: We need to undo the changes to "enable_index_processing".

crates/iota-tool/src/lib.rs
```diff
@@ -1035,7 +1035,14 @@ pub async fn download_db_snapshot(
     files.extend(epoch_manifest.filter_by_prefix("checkpoints").lines);
     if !skip_indexes {
         files.extend(epoch_manifest.filter_by_prefix("indexes").lines);
-        files.extend(epoch_manifest.filter_by_prefix("grpc_indexes").lines);
+        // Download gRPC indexes regardless of which name the snapshot used.
+        // Current snapshots use "grpc_indexes"; old ones used "rest_index".
+        files.extend(
```
muXxer: Undo this change.

crates/iota-tool/src/lib.rs
```diff
@@ -1093,15 +1100,18 @@ pub async fn download_db_snapshot(
...
-    let grpc_indexes_dir = path.join(format!("epoch_{epoch}")).join("grpc_indexes");
-    if grpc_indexes_dir.exists() {
-        fs::rename(
-            &grpc_indexes_dir,
-            path.join(format!("epoch_{epoch}")).join("rest_index"),
-        )
-        .map_err(|e| anyhow::anyhow!("Failed to rename grpc_indexes to rest_index: {e}"))?;
+    // Rename legacy "rest_index" snapshot directory to "grpc_indexes".
+    // TODO(cleanup): Remove after one release cycle.
+    let epoch_dir = path.join(format!("epoch_{epoch}"));
```
muXxer: This can be completely removed.

crates/iota-indexer/README.md
```diff
@@ -6,7 +6,7 @@ IOTA Indexer is an off-fullnode service to serve data from the IOTA protocol, in
 
 > [!NOTE]
 >
-> - Indexer sync workers require the `NodeConfig::enable_rest_api` flag set to `true` in the node
+> - Indexer sync workers require the node to be running as a full node (indexes are always created for full nodes)
```
muXxer: We should say that the gRPC api should be enabled. I would leave the comment as it was before, but mention "full node" instead of node, and the parameter to enable the grpc API.

crates/iota-node/src/lib.rs
```diff
@@ -668,21 +666,21 @@ impl IotaNode {
...
-                    &epoch_store,
-                    &cache_traits.backing_package_store,
-                )
-                .await,
-            ))
-        } else {
-            None
-        };
+        let grpc_indexes_store =
+            if is_full_node && (config.enable_index_processing || config.enable_grpc_api) {
```
muXxer:
```suggestion
            if is_full_node && config.enable_grpc_api {
```

crates/typed-store/tests/macro_tests.rs
```diff
@@ -248,6 +248,71 @@ async fn deprecate_test() {
...
+
+    // Step 1: Create DB with original types (i32, String) and write data
+    {
+        let original_db =
+            Tables::open_tables_read_write(dbdir.clone(), MetricConf::default(), None, None);
+        original_db.table1.insert(&key, &value).unwrap();
+        original_db.table2.insert(&42, &value).unwrap();
+    }
+
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
muXxer, line 267: Can you replace my `from_secs(5)` with a constant everywhere? Maybe we should try to only wait 100ms. I don't want to block the CI for too long.

crates/simulacrum/src/grpc.rs
```diff
@@ -0,0 +1,45 @@
...
+pub async fn start_simulacrum_grpc_server(
+    simulacrum: Arc<Simulacrum>,
+    config: iota_config::node::GrpcApiConfig,
+    shutdown_token: CancellationToken,
+) -> Result<GrpcServerHandle> {
+    let chain_id = simulacrum
+        .get_chain_identifier()
+        .expect("chain identifier should be set");
+
+    let executor =
```
muXxer:
```suggestion
    // Create a transaction executor for simulacrum to enable transaction execution
    // and simulation via gRPC
    let executor =
```

crates/simulacrum/src/grpc.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use crate::{Simulacrum, transaction_executor::TransactionExecutor};
+
+/// Start a gRPC server for the given simulacrum instance.
+///
+/// This creates the full server stack with:
+/// - The [`Simulacrum`] as the state reader (implements [`GrpcStateReader`])
+/// - A [`TransactionExecutor`] for transaction execution/simulation
+/// - The real chain identifier from the simulacrum genesis
+pub async fn start_simulacrum_grpc_server(
```
muXxer: Why did you move this function to `simulacrum`? Can't we just leave it where it was? `simulacrum-server` can also be used as lib.

crates/iota-rest-api/src/metrics.rs
```diff
@@ -1,139 +0,0 @@
...
-    inflight_requests: IntGaugeVec,
-    num_requests: IntCounterVec,
-    request_latency: HistogramVec,
-}
-
-const LATENCY_SEC_BUCKETS: &[f64] = &[
-    0.001, 0.005, 0.01, 0.05, 0.1, 0.25, 0.5, 1., 2.5, 5., 10., 20., 30., 60., 90.,
-];
-
-impl RestMetrics {
```
muXxer, line 25: We could also implement such metrics in gRPC.

muXxer (changes requested):

crates/iota-node-storage/src/lib.rs
bingyanglin: I think we should keep `get_lowest_available_checkpoint_objects`. It is still used in 3 places in `iota-grpc-server`:

1. `crates/iota-grpc-server/src/ledger_service/get_service_info.rs`: returns this value in the gRPC service info response
2. `crates/iota-grpc-server/src/types.rs`: wrapper method on `GrpcReader`
3. `crates/iota-grpc-server/src/response.rs`: sets the `X_IOTA_LOWEST_AVAILABLE_CHECKPOINT_OBJECTS` HTTP header

crates/iota-swarm-config/src/node_config_builder.rs
bingyanglin: Goot catch. The config is restored and the snapshot is updated at [86dade9](https://github.com/iotaledger/iota/pull/10566/commits/86dade93a97864776aba71274e42ee54b62fd6c4).

crates/iota-swarm-config/tests/snapshots/snapshot_tests__network_config_snapshot_matches.snap
bingyanglin: Updated at [86dade9](https://github.com/iotaledger/iota/pull/10566/commits/86dade93a97864776aba71274e42ee54b62fd6c4).

crates/typed-store/tests/macro_tests.rs
bingyanglin: Replaced it at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-indexer/README.md
bingyanglin: Updated the docs accordingly at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-tool/src/lib.rs
bingyanglin: Removed at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-tool/src/lib.rs
bingyanglin: Reverted it at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-light-client/src/config.rs
bingyanglin: Added a TODO at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-grpc-server/src/types.rs
bingyanglin: It will returns `None` only in test mocks, and in production the gRPC server starts only when indexes are available. Addressed this in [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-light-client/src/verifier.rs
bingyanglin: Agreed. This is for our reminding of this PoC. I removed it at [4b98883](https://github.com/iotaledger/iota/pull/10566/commits/4b988838b58eec42cefcc2a83078dcfaa6a031b7).

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Good catch. Added periodic batch flushing every 10,000 entries in a5163e6.

crates/iota-light-client/src/config.rs
bingyanglin: Agreed. Removed the fallback to `rpc_url`. `grpc_url()` now returns `Option<&Url>` and callers error out if not configured. Fixed at a5163e6.

crates/iota-light-client/src/config.rs
bingyanglin: Agreed. Removed the fallback to rpc_url. grpc_url() now returns Option<&Url> and callers error out if not configured. Fixed at https://github.com/iotaledger/iota/commit/a5163e6625ea33ed6c67d780502603529b6ebfb3.

crates/iota-light-client/src/checkpoint.rs
bingyanglin: Good point. Narrowed the read_mask to `"checkpoint.summary"` at a5163e6.

crates/simulacrum/src/grpc.rs
bingyanglin: Moved `start_simulacrum_grpc_server` back to `simulacrum-server` and removed `iota-grpc-server` from `simulacrum` deps. Fixed at a5163e6.

crates/iota-data-ingestion-core/src/executor.rs
bingyanglin: This URL is for a remote object store, not a gRPC endpoint. Port 9000 was the old REST API port, so it's misleading to keep it here. But it's also not a gRPC path, so using 50051 would be wrong too. Maybe let the code owner to decide?

crates/iota-node-storage/src/lib.rs
muXxer: Ah true! Yeah I wasn't sure. Thanks!

crates/iota-data-ingestion-core/src/executor.rs
muXxer: @sergiupopescu199 fyi

crates/iota-rest-api/src/client/sdk.rs
muXxer: I added one in https://github.com/iotaledger/iota/pull/10644

crates/iota-rest-api/src/health.rs
muXxer: Added in https://github.com/iotaledger/iota/pull/10644

crates/iota-light-client/src/checkpoint.rs
muXxer: still you can simply call `response.summary()` instead of the `response.checkpoint_data()`

crates/iota-data-ingestion-core/src/reader/v1.rs
sergiupopescu199: @muXxer @bingyanglin 

We have already implemented the gRPC change on the  `iota-data-ingestion-core`, `iota-indexer`, `iota-graphql`

This is the draft PR which contains the overall changes: https://github.com/iotaledger/iota/pull/10081 so you could rebase on top of it. Regarding the `v1` you can replace the Rest API with the gRPC, we still need both `v1` & `v2` for backwards compatibility purposes.

crates/iota-light-client/src/checkpoint.rs
bingyanglin: Sorry after more investigation actually we need the `CertifiedCheckpointSummary`, so we need the summary and signature. Fixed it at [2642b20](https://github.com/iotaledger/iota/pull/10566/commits/2642b20326dce5366972f357d84573d146dc15ce).

crates/iota-data-ingestion-core/src/executor.rs
sergiupopescu199: @muXxer @bingyanglin 

we already have an issue in place to fix the infra related docs: https://github.com/issues?q=is:open+is:issue+author:sergiupopescu199+archived:false+user:iotaledger&issue=iotaledger%7Ciota%7C10648

dev-tools/iota-data-ingestion/README.md
sergiupopescu199: @muXxer @bingyanglin 

yeah `http://localhost:50051` should do the trick.

I will double check all the infra related docs regarding the gRPC API usage as part of this issue: https://github.com/issues?q=is:open+is:issue+author:sergiupopescu199+archived:false+user:iotaledger&issue=iotaledger%7Ciota%7C10648

crates/iota-data-ingestion-core/src/executor.rs
muXxer: alright, so either the issue will change the base branch and this will be conflicting then, so we can solve it, or this PR gets merged earlier and we can leave it as it is. 👍

crates/iota-rest-api/src/metrics.rs
bingyanglin: Addressed in #10676

crates/iota-node/src/main.rs
```diff
@@ -117,9 +116,8 @@ fn main() {
     // let iota-node signal main to shutdown runtimes
     let (runtime_shutdown_tx, runtime_shutdown_rx) = broadcast::channel::<()>(1);
 
-    let server_version = ServerVersion::new(env!("CARGO_BIN_NAME"), VERSION);
     runtimes.iota_node.spawn(async move {
-        match IotaNode::start_async(config, registry_service, server_version).await {
+        match IotaNode::start_async(config, registry_service, VERSION).await {
```
muXxer: I would keep the `ServerVersion` and just move it to some other crate where it makes sense, maybe directly here in `iota-node/lib.rs` ? 

```rust
#[derive(Clone)]
pub struct ServerVersion {
    pub bin: &'static str,
    pub version: &'static str,
}

impl ServerVersion {
    pub fn new(bin: &'static str, version: &'static str) -> Self {
        Self { bin, version }
    }
}

impl std::fmt::Display for ServerVersion {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str(self.bin)?;
        f.write_str("/")?;
        f.write_str(self.version)
    }
}
```

crates/iota-node/src/lib.rs
```diff
@@ -436,7 +429,7 @@ impl IotaNode {
     pub async fn start_async(
         config: NodeConfig,
         registry_service: RegistryService,
-        server_version: ServerVersion,
+        software_version: &'static str,
```
muXxer: As mentioned in the other comment, revert to `ServerVersion`.

crates/iota-node/src/lib.rs
```diff
@@ -2447,37 +2430,32 @@ async fn build_grpc_server(
...
-    let grpc_reader = Arc::new(GrpcReader::from_rest_state_reader(
-        rest_read_store,
-        Some(server_version.to_string()),
+    let grpc_reader = Arc::new(GrpcReader::new(
+        grpc_read_store,
+        Some(env!("CARGO_PKG_VERSION").to_string()),
     ));
 
-    // Create gRPC server metrics
-    let grpc_server_metrics = iota_grpc_server::GrpcServerMetrics::new(prometheus_registry);
```
muXxer, line 2462: That seems wrong! Revert.

crates/iota-node/src/lib.rs
```diff
@@ -2447,37 +2430,32 @@ async fn build_grpc_server(
...
     // Create cancellation token for proper shutdown hierarchy
     let shutdown_token = CancellationToken::new();
 
     // Create GrpcReader
-    let grpc_reader = Arc::new(GrpcReader::from_rest_state_reader(
-        rest_read_store,
-        Some(server_version.to_string()),
+    let grpc_reader = Arc::new(GrpcReader::new(
+        grpc_read_store,
+        Some(env!("CARGO_PKG_VERSION").to_string()),
```
muXxer: Revert and pass the `server_version` again.

dev-tools/iota-data-ingestion/README.md
```diff
@@ -82,6 +82,7 @@ tasks:
       # Minimum allowed chunk size is 5MB
       #
       checkpoint-chunk-size-mb: 100
+      node-grpc-url: "http://localhost:50051"
```
muXxer: There is no such config I think.

crates/iota-swarm-config/src/node_config_builder.rs
muXxer: Maybe move it to it's original position. Also, is there a reason to not use the `default_enable_index_processing`?

crates/iota-swarm-config/src/node_config_builder.rs
```diff
@@ -618,6 +606,7 @@ impl FullnodeConfigBuilder {
             verifier_signing_config: VerifierSigningConfig::default(),
             enable_db_write_stall: None,
             iota_names_config: self.iota_names_config,
+            enable_index_processing: true,
```
muXxer: Maybe move it to it's original position. Also, is there a reason to not use the `default_enable_index_processing`?

crates/iota-core/src/grpc_indexes.rs
muXxer: Is this still used?

muXxer: This can also be removed?

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -437,28 +434,23 @@ struct IndexStoreTables {
...
     owner_v2: DBMap<OwnerIndexKeyV2, OwnerIndexInfoV2>,
 
     /// An index of dynamic fields (children objects).
     ///
     /// Allows an efficient iterator to list all of the dynamic fields owned by
     /// a particular ObjectID.
-    // REST-API and gRPC
-    // TODO: Replace DynamicFieldIndexInfo with () once the REST-API server is
-    // deprecated — gRPC only needs the key.
     dynamic_field: DBMap<DynamicFieldKey, DynamicFieldIndexInfo>,
```
muXxer: What happened to the TODO? Do we need the `DynamicFieldIndexInfo` or can we drop it with a migration?

muXxer (changes requested):

crates/iota-core/src/grpc_indexes.rs
bingyanglin: `StateService::list_dynamic_fields` still needs it. So we need to keep it. The `value`, `value_type`, `field_object`, `child_object`) are only populated by `load_dynamic_field()`.

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Fixed in [c912584](https://github.com/iotaledger/iota/pull/10566/commits/c912584e101b38c23b6bb1fe1abcf08186a1a701).

bingyanglin: Fixed in https://github.com/iotaledger/iota/commit/c912584e101b38c23b6bb1fe1abcf08186a1a701.

crates/iota-swarm-config/src/node_config_builder.rs
bingyanglin: Fixed in https://github.com/iotaledger/iota/commit/c912584e101b38c23b6bb1fe1abcf08186a1a701.

dev-tools/iota-data-ingestion/README.md
bingyanglin: Fixed in https://github.com/iotaledger/iota/pull/10566/changes/c912584e101b38c23b6bb1fe1abcf08186a1a701.

crates/iota-node/src/lib.rs
bingyanglin: Fixed in https://github.com/iotaledger/iota/pull/10566/changes/c912584e101b38c23b6bb1fe1abcf08186a1a701.

crates/iota-node/src/lib.rs
bingyanglin: Fixed in [c912584](https://github.com/iotaledger/iota/pull/10566/commits/c912584e101b38c23b6bb1fe1abcf08186a1a701).

crates/iota-node/src/main.rs
bingyanglin: Fixed in [c912584](https://github.com/iotaledger/iota/pull/10566/commits/c912584e101b38c23b6bb1fe1abcf08186a1a701).

crates/iota-node/src/lib.rs
bingyanglin: Fixed in https://github.com/iotaledger/iota/pull/10566/changes/4b09f805554bcc7b21a760033158a16817da46a5

crates/iota-types/src/storage/read_store.rs
```diff
@@ -1011,8 +902,6 @@ pub struct CoinInfo {
     pub treasury_object_id: Option<ObjectID>,
 }
 
-/// Extended coin info from the `coin_v2` table — merges `coin` +
-/// `regulated_coin` into a single lookup.
```
Thoralf-M, line 1015: Why remove the comments?

bingyanglin: Fixed at [46fdf39](https://github.com/iotaledger/iota/pull/10566/commits/46fdf3943b3dbc69a3d1de024949dab015b703f4).

muXxer: IMO the removal of the comment was fine. The info doesn't make any sense anymore after the removal of the normal "CoinInfo" IMO.

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Now they are loaded on demand from the object store, fixed it in [e6c4546](https://github.com/iotaledger/iota/pull/10566/commits/e6c45467f67df8a06ff54e1fc7fa4fbc57f5da07). Thanks for the comment.

crates/iota-swarm-config/src/node_config_builder.rs
bingyanglin: Fixed in https://github.com/iotaledger/iota/commit/c912584e101b38c23b6bb1fe1abcf08186a1a701.

crates/iota-types/src/storage/read_store.rs
bingyanglin: I enhanced the comment in [5c5aa06](https://github.com/iotaledger/iota/pull/10566/commits/5c5aa06fd105f3e4fd90e0982532df4416057012) to make it clear and correct. Thanks for pointing this out.

crates/iota-grpc-server/src/state_service/list_dynamic_fields.rs
```diff
@@ -37,27 +36,62 @@ struct PageToken {
...
+    if read_mask.contains(DynamicField::FIELD_ID_FIELD.name) {
+        message.field_id = Some(crate::validation::object_id_proto(field_id));
+    }
+
+    if load_field {
+        if let Err(e) = load_dynamic_field(reader, field_id, read_mask, &mut message) {
+            tracing::warn!("error loading dynamic field object {field_id}: {e}");
+            // Return None to skip this item (object may be missing due to
+            // eventual consistency).
+            return Ok(None);
```
muXxer: But we could still return `Ok(Some(message))` here, because we filled the first two fields, even if the object doesn't exist anymore, so the clients would get at least that info.

muXxer: Before there was also this comment, we should re-add it.
```rust
        // Conditionally load the field object to populate heavy fields.
        // On recoverable errors (missing layout, deserialization failure),
        // the item is still returned with index-only fields populated so
        // that clients see all items and can detect partial data via the
        // absence of the requested heavy fields.
```

muXxer: And:
```
                // Return the item with index-only fields rather than
                // silently dropping it.
```

muXxer (changes requested):

crates/iota-grpc-server/src/state_service/list_dynamic_fields.rs
bingyanglin: Good catch! Fixed at [3d29ee8](https://github.com/iotaledger/iota/pull/10566/commits/3d29ee8c81c8880579b0db9e4ca9730192c11fa2). Thanks.

muXxer (approved):

filipdulic (approved):

alexsporn (approved):

miker83z (approved):

crates/iota-graphql-rpc/Cargo.toml
```diff
@@ -50,6 +50,7 @@ uuid.workspace = true
 
 # internal dependencies
 bin-version.workspace = true
+iota-config.workspace = true
```
sergiupopescu199: This import is not used anywhere, we can safely remove it.

```suggestion

```

crates/iota-node-storage/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "iota-node-storage"
+version.workspace = true
+authors = ["IOTA Foundation <info@iota.org>"]
+edition = "2021"
+license = "Apache-2.0"
+publish = false
+
```
sergiupopescu199, line 8: From my understanding we shall add it in all crates right?

```suggestion

[lints]
workspace = true

```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -812,144 +807,48 @@ impl<T: ReadStore + ?Sized> ReadStore for Arc<T> {
...
+    pub start_timestamp_ms: u64,
+    pub end_timestamp_ms: Option<u64>,
+    pub start_checkpoint: u64,
+    pub end_checkpoint: Option<u64>,
+    pub reference_gas_price: u64,
+    /// System State as of the start of the epoch
+    pub system_state: crate::iota_system_state::IotaSystemState,
+}
 
-#[derive(Clone)]
```
sergiupopescu199, line 952: Was the `Clone` removal intentional? This struct is part of the public interface (`iota-types`) removing a trait derive is a breaking change for downstream consumers. Worth keeping it for forward compatibility?

sergiupopescu199 (approved):
LGTM! 

I checked all infra related crates and all look good. I left a few non critical nits.

Also since we removed the REST API in favor of the gRPC, this means that we could enable by default the latter when starting local network in the `iota-localnet` as we did previously, since in the swarm builder we were always enabling the rest api by default, I think we shall do the same for the gRPC. 

This will allow to fix this issue: https://github.com/iotaledger/iota/issues/10777

What do you think?

muXxer:
> LGTM!
> 
> I checked all infra related crates and all look good. I left a few non critical nits.
> 
> Also since we removed the REST API in favor of the gRPC, this means that we could enable by default the latter when starting local network in the `iota-localnet` as we did previously, since in the swarm builder we were always enabling the rest api by default, I think we shall do the same for the gRPC.
> 
> This will allow to fix this issue: #10777
> 
> What do you think?

There is a new parameter for the `iota-localnet` where you can enable it `--with-grpc`, see https://github.com/iotaledger/iota/pull/11041

PR #10578, opened 2026-03-03
fix(consensus-adapter): actually insert `EndOfPublish` (if absent) for re-submission on crash recovery

roman1e2f5p8s:
# Description of change

This PR fixes a plain inversion bug in `submit_recovered` of `ConsensusAdapter`. Specifically, if the list of `recovered` pending consensus transactions does not contain an `EndOfPublish` transaction, the condition `recovered.iter().any(ConsensusTransaction::is_end_of_publish)` is `false`, which means `EndOfPublish` will never be resubmitted in the case of crash recovery, likely leading to epoch transitioning getting stuck. In contrast, if `EndOfPublish` is already present in recovered, the current code on `develop` needlessly inserts another `EndOfPublish`. However, in the actual crash-recovery scenario, we would like to insert `EndOfPublish` into `recovered` if it was absent; hence, the inverted condition `!recovered.iter().any(ConsensusTransaction::is_end_of_publish)` is correct.

Other minor changes include improving the corresponding comments and removing `#[expect(clippy::collapsible_if)]` as it is no longer needed.

> [!NOTE]
> 1. The bug was discovered during the certificate-less epoch change PoC ([branch](https://github.com/iotaledger/iota/tree/protocol-research/feat/certificate-less-epoch-boundary-coordination)).
> 2. Upstream added this fix as a part of https://github.com/MystenLabs/sui/pull/23122.

```shell
cargo test -p iota-core submit_recovered_end_of_publish_crash_recovery -- --nocapture
```

### DIFF
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -342,24 +342,24 @@ impl ConsensusAdapter {
         // be a big deal but can be optimized
         let mut recovered = epoch_store.get_all_pending_consensus_transactions();
 
-        #[expect(clippy::collapsible_if)] // This if can be collapsed but it will be ugly
         if epoch_store
             .get_reconfig_state_read_lock_guard()
             .is_reject_user_certs()
             && epoch_store.pending_consensus_certificates_empty()
         {
-            if recovered
+            // If `recovered` does not contain `EndOfPublish` yet, we need to insert it.
+            if !recovered
                 .iter()
                 .any(ConsensusTransaction::is_end_of_publish)
             {
-                // There are two cases when this is needed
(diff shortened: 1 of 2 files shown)


semenov-vladyslav (approved):
The fix looks reasonable. Maybe the rest of the upstream commit logic can also be ported in another PR.

muXxer (approved):

piotrm50 (approved):

roman1e2f5p8s:
> The fix looks reasonable. Maybe the rest of the upstream commit logic can also be ported in another PR.

Thanks. I believe some of the changes are not portable because they are tied to Sui-specific `FastpathCertified` transaction status, which we are not planning to have. But I added a comment to the upstream commit on our slipstream tool.

PR #10581, opened 2026-03-03
feat(ts-sdk): Include typedoc urls in codegen scripts

marc2332:
basically injects a `@see` comment in the typedoc comment of so thta when you hover it shows like this:

image


panteleymonchuk:
https://github.com/iotaledger/ts-packages/pull/98

PR #10584, opened 2026-03-03
fix(ts-sdk): Correct query events fetching order in json rpc transport

marc2332:
>   /** query result ordering, default to false (ascending order), oldest record first. */

### DIFF
--- a/sdk/typescript/src/client/client.ts
+++ b/sdk/typescript/src/client/client.ts
@@ -417,7 +417,7 @@ export class IotaClient {
                 } as IotaTransactionBlockResponseQuery,
                 input.cursor,
                 input.limit,
-                (input.order || 'descending') === 'descending',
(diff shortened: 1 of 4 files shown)


VmMad (approved):

sdk/typescript/src/client/client.ts
begonaalvarezd, line 1: we have the same issue in `queryTransactionBlocks` 😿

begonaalvarezd (changes requested):
If we update `queryTransactionBlocks` I see its affecting our apps, so we should explicitly define the order in `apps/core/src/hooks/useGetObjectOrPastObject.ts` & `apps/core/src/hooks/useQueryTransactionsByAddress.ts`

begonaalvarezd (approved):

PR #10591, opened 2026-03-04
fix(starfish): Fix handling of non-versioned ShardWithProofV1 in block bundles

piotrm50:
# Description of change

This pull request introduces backward compatibility for the `ShardWithProof` serialization format during a protocol upgrade. The main goal is to ensure that nodes running different software versions can interoperate smoothly while transitioning to the new versioned `ShardWithProof` enum. The code now conditionally supports both the old (`ShardWithProofV1`) and new (`ShardWithProof`) formats, based on the protocol configuration.

Backward compatibility for ShardWithProof serialization:

* When deserializing shards, the code now checks the `consensus_fast_commit_sync` flag in the protocol config. If disabled (indicating an upgrade period), it attempts to deserialize using the old `ShardWithProofV1` type and wraps it in the new enum; otherwise, it uses the new `ShardWithProof` directly.
* When serializing `shard_for_core`, the code similarly checks the protocol config and serializes as `ShardWithProofV1` if necessary, ensuring older nodes can still understand the data.

Type and import updates:

* The import list now includes both `ShardWithProofV1` and the updated `ShardWithProof` to support the new compatibility logic.

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -25,8 +25,8 @@ use crate::{
     CommitIndex, Round, Transaction, VerifiedBlockHeader,
     block_header::{
         BlockHeaderAPI, BlockHeaderDigest, BlockRef, GENESIS_ROUND, ShardWithProof,
-        ShardWithProofAPI, SignedBlockHeader, TransactionsCommitment, VerifiedBlock,
-        VerifiedOwnShard, VerifiedTransactions,
+        ShardWithProofAPI, ShardWithProofV1, SignedBlockHeader, TransactionsCommitment,
+        VerifiedBlock, VerifiedOwnShard, VerifiedTransactions,
     },
     block_verifier::BlockVerifier,
     commit::{CommitAPI as _, CommitRange, TrustedCommit},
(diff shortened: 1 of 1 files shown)


crates/starfish/core/src/authority_service.rs
```diff
@@ -334,7 +334,18 @@ impl<C: CoreThreadDispatcher> AuthorityService<C> {
         let mut verified_shards: Vec<ShardWithProof> = vec![];
         for serialized_shard in &serialized_shards {
             let shard: ShardWithProof =
-                bcs::from_bytes(serialized_shard).map_err(ConsensusError::MalformedShard)?;
+                if !self.context.protocol_config.consensus_fast_commit_sync() {
+                    // For backward compatibility, we still support ShardWithProofV1 during the
+                    // epoch during which nodes are upgraded to a new software version. Peers
+                    // running an old version will still send ShardWithProofV1 without the enum
+                    // wrapping. We can remove this support after we are sure
+                    // all peers have been updated to send versioned ShardWithProof.
```
VorobyevIlya, line 342: As far as I understand, we can remove this support only after the flag is enabled, not after the upgrade

VorobyevIlya (approved):

crates/starfish/core/src/authority_service.rs
piotrm50: yep, but even without the flag if everyone is updated they can send enum-wrapped V1 variant without problems. Anyway, it's a small detail so I'll leave it as is.

filipdulic (approved):

PR #10593, opened 2026-03-04
feat(indexer): Tune vacuuming for pruned indexer tables

### LINKED ISSUE #9888: [indexer]: Test robustness of the `watermarks` solution
Scenarios to test:
 - what happens if info in **watermarks table is incorrect** (can happen if we prune data in reality, but fail to update the pruner_hi). Would the mechanism be able to recover from such situation(s).
 - **prune all the tables** aggresively (epochs to keep=1 or 2) and see if **queries on current and old data** work as expected
 - run **indexer sync** from long running network, observe **TPS** of this sync with and without pruning
   - check how pruning affects sync TPS on heavily pruned indexer (epoch to keep ~ 1)
   - check how pruning affects sync TPS on big DB with lots of history (epochs to keep ~ 100?)
   - for both cases, see if pruning is able to keep up, or if it's lagging for some tables

Ideally tests should be automated using scripts, for easy reexecution in the future

tomxey:
# Description of change

Make vacuuming more frequent for pruned tables.
This change makes sure pruned tables are vacuumed often enough, which results in table size staying at a constant level and not evergrowing with dead tuples accumulating and space not reused. 

This needs to also be combined with setting appropriate `shm_size` parameter on docker container where postgres is running so that postgres has enough memory to actually vacuum the big tables. e.g. `shm_size: 2g` was tested to be working on tables of > 100GB size. 

fixes https://github.com/iotaledger/iota/issues/9888

this was tested on staging server, all pruned tables were vacuumed correctly and their size was kept on more or less the same level, ingestion speed was not visibly affected


kodemartin (approved):
lgtm 🧹

sergiupopescu199 (approved):
lgtm 🐘

Issue #10609, opened 2026-03-05
Port upstream node team related changes [mainnet-v1.46.3..mainnet-v1.47.1)


jkrvivian:

muXxer:
Merged with #10946

PR #10610, opened 2026-03-05
upstream(core): Preparations for a switchable storage backend

jkrvivian:
# Description of change

- Port commit: 
  - https://github.com/MystenLabs/sui/commit/e66ce415c6a160c0e6b577d84c496c72a94fcb41
- Description:

Preparation PR for a switchable storage backend.
 
Notes:
* `enum Storage` represents the storage backend with options for
RocksDB, InMemoryStore, and TH.
* the PR deprecates several unnecessary typed store methods and merges
redundant methods
* a basic InMemoryStore backend has been added as a reference
implementation, with potential future use in tests
* not all methods are covered. Iterators will be handled separately. The
returned interface will need to be changed, and additional methods will
be deprecated.
* the TH option is behind a feature flag. Extra options (e.g.,
with_rm_prefix) will be added in a followup
* the PR is essentially a noop and does not change the existing node's
behavior

### Additional fixes (review feedback)

* Generic types (`ColumnFamily`, `Storage`, `Database`, `GetResult`, `StorageWriteBatch`) moved out of `rocks/mod.rs` into the new `crates/typed-store/src/database.rs` module
* All visibility reduced to `pub(crate)` where external exposure is not needed (`RocksDB`, `ColumnFamily`, `Storage`)

Part of #10609


bingyanglin (approved):

crates/typed-store/src/rocks/mod.rs
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub mod errors;
+pub mod memstore;
```
semenov-vladyslav: This looks out of place. Why is it defined here and not in `super`?

piotrm50 (approved):

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
...
 }
 
 impl Drop for RocksDB {
     fn drop(&mut self) {
         self.underlying.cancel_all_background_work(/* wait */ true);
+    }
+}
+
+#[derive(Clone)]
+pub enum ColumnFamily {
```
semenov-vladyslav:
```suggestion
pub(crate) enum ColumnFamily {
```
It doesn't have to be `pub`.

semenov-vladyslav: The "generic" types should live outside `rocks` module, otherwise it's very confusing.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
...
+                .expect("Map-keying column family should have been checked at DB creation"),
+            _ => unreachable!("invariant is checked by the caller"),
+        }
+    }
+}
+
+pub enum Storage {
+    Rocks(RocksDB),
+    InMemory(InMemoryDB),
+}
```
semenov-vladyslav:
```suggestion
enum Storage {
    Rocks(RocksDB),
    #[allow(dead_code)]
    InMemory(InMemoryDB),
}
```
Here too. The less we expose in the public API, the easier to keep it clean.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
...
 impl Drop for RocksDB {
     fn drop(&mut self) {
         self.underlying.cancel_all_background_work(/* wait */ true);
+    }
+}
+
+#[derive(Clone)]
+pub enum ColumnFamily {
+    Rocks(String),
+    InMemory(String),
```
semenov-vladyslav:
```suggestion
    #[allow(dead_code)]
    InMemory(String),
```
It seems to be unused.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
 #[derive(Debug)]
 pub struct RocksDB {
```
semenov-vladyslav:
```suggestion
struct RocksDB {
```
This doesn't have to be public.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Storage::Rocks(db) => write!(f, "RocksDB Storage {:?}", db),
+            Storage::InMemory(db) => write!(f, "InMemoryDB Storage {:?}", db),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Database {
```
semenov-vladyslav: It seems to be the top-level type dispatching the inner implementations. Should we make it into a trait:

```rs
pub trait Database {
    fn path_for_pruning(&self) -> &Path;
    fn live_files(&self) -> Result<Vec<LiveFile>, Error>;
    fn flush(&self);
}
```
Only these three methods are in use, plus the first two should be factored into some `fn prune(&self);` method.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -609,16 +682,26 @@ impl<K, V> DBMap<K, V> {
             .unwrap_or(rocksdb::DEFAULT_COLUMN_FAMILY_NAME)
             .to_owned();
 
-        db.cf_handle(&cf_key)
-            .ok_or_else(|| TypedStoreError::UnregisteredColumn(cf_key.clone()))?;
+        Ok(DBMap::new(
+            db.clone(),
+            rw_options,
+            &cf_key,
+            ColumnFamily::Rocks(cf_key.to_string()),
```
semenov-vladyslav: Should the `ColumnFamily` arm be determined by the `db` runtime implementation?

crates/typed-store/src/rocks/mod.rs
```diff
@@ -609,16 +682,26 @@ impl<K, V> DBMap<K, V> {
...
-        Ok(DBMap::new(db.clone(), rw_options, &cf_key, is_deprecated))
+    pub fn cf_name(&self) -> &str {
+        &self.cf
     }
 
     pub fn batch(&self) -> DBBatch {
-        let batch = WriteBatch::default();
+        let batch = match &self.db.storage {
+            Storage::Rocks(_) => StorageWriteBatch::Rocks(WriteBatch::default()),
+            Storage::InMemory(_) => StorageWriteBatch::InMemory(InMemoryBatch::default()),
```
semenov-vladyslav: Again, all common/generic logic should be moved out of `rocks` module.

crates/typed-store/src/rocks/mod.rs
semenov-vladyslav: The `DBMap` is generic over internal implementation, but the `SafeIterator` is `SafeIter` which is rocksdb-specific. It should also be abstracted away.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -476,9 +565,10 @@ const METRICS_ERROR: i64 = -1;
 /// An interface to a rocksDB database, keyed by a columnfamily
 #[derive(Clone, Debug)]
 pub struct DBMap<K, V> {
-    pub rocksdb: Arc<RocksDB>,
+    pub db: Arc<Database>,
     _phantom: PhantomData<fn(K) -> V>,
-    // the rocksDB ColumnFamily under which the map is stored
+    column_family: ColumnFamily,
```
semenov-vladyslav: Having both `cf` and `column_family` is pretty confusing.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -133,213 +134,321 @@ macro_rules! reopen {
...
-        start: Option<K>,
-        end: Option<K>,
-    ) {
-        self.underlying.compact_range_cf(cf, start, end)
-    }
-
-    pub fn compact_range_to_bottom<K: AsRef<[u8]>>(
-        &self,
-        cf: &impl AsColumnFamilyRef,
+        cf_name: &str,
```
semenov-vladyslav: Shouldn't it be `cf_name: &ColumnFamily`?

semenov-vladyslav (changes requested):
I don't like this approach: runtime impl dispatch introduces only confusion. We should fix that with the rust's type system. Otherwise, it'll be pretty hard to maintain.

crates/typed-store/src/rocks/mod.rs
jkrvivian: Addressed in the fix commit `4d432e3de7`. `ColumnFamily` is now `pub(crate)` in `crates/typed-store/src/database.rs`.

crates/typed-store/src/rocks/mod.rs
jkrvivian: Addressed in the fix commit `4d432e3de7`. `RocksDB` is now `pub(crate)` in `rocks/mod.rs` (line 134).

crates/typed-store/src/rocks/mod.rs
jkrvivian: Not applied in this PR. Moving `DBMap` and its `reopen` method out of the `rocks` module into a top-level module is a structural change beyond this preparation PR's scope. The PR description explicitly notes iterators and further abstractions will be handled separately in follow-up PRs.

crates/typed-store/src/rocks/mod.rs
jkrvivian: Addressed in the fix commit `4d432e3de7`. `raw_iterator_cf` and `compact_range_cf` in `database.rs` now take `cf: &ColumnFamily` instead of `cf_name: &str`.

crates/typed-store/src/rocks/mod.rs
jkrvivian: It's fixed in https://github.com/MystenLabs/sui/commit/c232615066db1ca778acf5b46e6cc3e91abd5dbe, and there's a PR for it: https://github.com/iotaledger/iota/pull/10780/changes/5179ee5a8082ffd4e13437f04a876e7b2fa2630e

semenov-vladyslav (commented):
This approach has a few drawbacks:
- run-time enum dispatch, no static type-safety, increased code size
- the existing types and components (DBMap, SafeIter, DBBatch, perf monitoring, compaction) are not adjusted well for an alternative DB impl (for example, tidehunter natively supports safe iteration and perf monitoring and doesn't require external compaction)
- the upstream still has to use "tidehunter" feature to enable/disable certain code paths
- the code quality is not very high, it would be hard to maintain

The alternative approach would be to abstract away DB engine with a carefully designed API (trait with associated types). With such an approach we would have static type-safety, no bloated code size to support multiple DB backends in the same binary, no adhoc code patching with feature gates.

I don't see the point in merging this PR, if we want to do it properly.

semenov-vladyslav (approved):
Approving. We will introduce a nicer API later.

PR #10615, opened 2026-03-05
feat(iota-core): implement PoC version of certificate-less epoch boundary coordination

### LINKED ISSUE #10376: pcool: PoC version of certificate-less epoch boundary coordination
## Description

Prepare a PoC version of certificate-less epoch boundary coordination.

## Motivation

See details in the PR.

Yes.

roman1e2f5p8s:
# Description of change

This PR implements a PoC version of certificate-less epoch boundary coordination.

<details>
  <summary>Background on epoch boundary coordination with certificates</summary>
- Each validator tracks which certificates it has submitted to consensus, but they have not been sequenced by consensus yet - `pending_consensus_certificates`.
- Certificates are the unit of work that a validator must drain before an epoch can close.

The epoch change process has the following phases:
1. **Reconfig state `AcceptAllCerts`**
  - All certificates are accepted and tracked in `pending_consensus_certificates`.
  - Every time a checkpoint is created, its timestamp is compared against `next_reconfiguration_timestamp_ms = epoch_start_timestamp_ms + epoch_duration_ms`.
  - If the timestamp exceeds this threshold, `close_epoch()` is called - the reconfig state moves to `RejectUserCerts`.
2. **Reconfig state `RejectUserCerts`**
  - New certificates are rejected; pending ones must be processed by consensus.
  - The `pending_consensus_certificates` set must be drained; only then `EndOfEpoch` is sent to consensus.
  - When `2f+1` stake worth of `EndOfPublish` is collected, the reconfig state transitions to `RejectAllCerts`.
3. **Reconfig state `RejectAllCerts`**
  - New user and consensus certificates are rejected.
  - At this point, the exact set of transactions in the epoch is determined.
  - Only previously deferred transactions remain to be processed.
  - Once all deferred transactions have been processed, the state transitions to `RejectAllTx`.
4. **Reconfig state `RejectAllTx`**
  - The commit round is labeled as final.
  - The pending checkpoint is labeled as last in the epoch.
  - Epoch is finalized.
</details>   

### Certificate-less epoch boundary coordination

The process to close the epoch in the certificate-less flow stays almost the same as with certificates. The **fundamental difference** is that, in the white flag flow, there are no certificates, so there is nothing to track in `pending_consensus_certificates`. We do not track `UserTransactionV1` because there is no pre-consensus "promise" (like there is with certificates) that the transaction will be executed by the end of the epoch. Thus, validators can send `EndOfPublish` immediately once the reconfig state is `RejectUserCerts`. This PR implements this as a PoC version.

### Summary of changes:

- Behavior of `insert_pending_consensus_transactions`:
  - Keep tracking transactions that will be submitted to consensus in the persistent `pending_consensus_transactions` table, but track `pending_consensus_certificates` only if the white flag flow is disabled.
- Behavior of `remove_pending_consensus_transactions`:
  - Keep removing processed-by-consensus transactions from the persistent `pending_consensus_transactions` table, but remove from `pending_consensus_certificates` only if the white flag flow is disabled.
- Behavior of `revert_uncommitted_epoch_transactions`:
  - Revert uncommitted locally executed pending consensus certificates that were not included in the epoch only if the white flag flow is disabled. For the certificate-less mode, we revert nothing because owned-object transactions are not executed pre-consensus in this mode.
- Condition to send `EndOfPublish`:
  - Send immediately once the reconfig state is `RejectUserCerts`.
  - The `is_user_tx` flag is explicitly set based on the white flag flow: a user transaction is `CertifiedTransaction` in certificate mode and `UserTransactionV1` in certificate-less mode.
- Explicit handling of the `pending_consensus_certificates` set based on the `enable_white_flag_flow` protocol config feature flag wherever the set appears.
  - _This is done for clarity and optimization_: `pending_consensus_certificates` is always empty if `enable_white_flag_flow` is enabled, so we avoid minor overheads such as reading the empty set and counting elements in it.
  - `AuthorityPerEpochStore` is initialized with empty `pending_consensus_certificates` if the white flag flow is enabled - this avoids unnecessary iteration over `pending_consensus_transactions`.
- Added the `num_rejected_tx_in_epoch_boundary` metric to be used in `submit_transaction_impl` to count the number of rejected user transactions.
- New function `handle_scheduling_result` that simply gathers the code that handles `SchedulingResult`, so the function can be used for both `CertifiedTransaction` and `UserTransactionV1`. 

### Multiple fixes not directly related to this PR:

- Applied the same fix as in https://github.com/iotaledger/iota/pull/10578, considering that `pending_consensus_certificates` is always empty in the white flag mode.
- Fixed `load_and_process_deferred_transactions_for_randomness` to also match over white-flag transactions, which are also user transactions that might be deferred due to randomness unavailable; without this fix, the function would panic.
- Fixed `load_initial_object_debts` to also account for white-flag user transactions; otherwise, object debts would not be loaded in the white-flag mode.
- Fixed `await_consensus_or_checkpoint` to construct a list of transaction digests for the white-flag transactions as well; otherwise the list would be empty in the white-flag mode, leading to potential hanging or unnecessary re-submissions.

Closes #10376.

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -2240,10 +2240,10 @@ impl AuthorityPerEpochStore {
         Ok(())
     }
 
-    /// Insert certified transactions into the in-memory
-    /// `pending_consensus_certificates` set. When submitting a certificate
-    /// caller **must** provide a `ReconfigState` lock guard and verify that
-    /// it allows new user certificates.
+    /// Insert certified transactions that will be submitted to consensus
+    /// into the in-memory `pending_consensus_certificates` set. When
+    /// submitting a certificate caller **must** provide a `ReconfigState`
+    /// lock guard and verify that it allows new user certificates.
     pub fn insert_pending_consensus_certificates(
         &self,
         transactions: &[ConsensusTransaction],
@@ -2265,6 +2265,8 @@ impl AuthorityPerEpochStore {
         }
     }
 
+    /// Remove processed by consensus transactions from the persistent
+    /// `pending_consensus_transactions` table.
     pub fn remove_pending_consensus_transactions(
         &self,
         keys: &[ConsensusTransactionKey],
@@ -2272,18 +2274,19 @@ impl AuthorityPerEpochStore {
         self.tables()?
             .pending_consensus_transactions
             .multi_remove(keys)?;
+
+        Ok(())
+    }
+
+    /// Remove processed by consensus certified transactions from the in-memory
+    /// `pending_consensus_certificates` set.
+    pub fn remove_pending_consensus_certificates(&self, keys: &[ConsensusTransactionKey]) {
         // TODO: lock once for all remove() calls.
         for key in keys {
             if let ConsensusTransactionKey::Certificate(cert) = key {
                 self.pending_consensus_certificates.write().remove(cert);
             }
-            // NOTE: We do not need the
-            // `ConsensusTransactionKey::UserTransaction` branch
-            // (certificate-less scenario) here because `UserTransactionV1` are
-            // not inserted into
-            // `self.pending_consensus_certificates`.
         }
-        Ok(())
     }
 
     pub fn pending_consensus_certificates_count(&self) -> usize {
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -2031,6 +2031,9 @@ impl AuthorityPerEpochStore {
                         SequencedConsensusTransactionKey::External(
                             ConsensusTransactionKey::Certificate(digest),
                         ) => (digest, (*deferral_key, tx.suggested_gas_price)),
+                        SequencedConsensusTransactionKey::External(
+                            ConsensusTransactionKey::UserTransaction(digest),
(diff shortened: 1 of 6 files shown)


bingyanglin (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
bingyanglin: Now that handle_scheduling_result handles both CertifiedTransaction and UserTransactionV1, the name ConsensusCertificateResult is a bit misleading. In white flag mode there are no certificates involved. Maybe rename it to something like ConsensusTransactionResult to be mode-agnostic? This is a pre-existing naming issue but this PR's unification makes it more noticeable.

roman1e2f5p8s: > Now that handle_scheduling_result handles both CertifiedTransaction and UserTransactionV1, the name ConsensusCertificateResult is a bit misleading. In white flag mode there are no certificates involved. Maybe rename it to something like ConsensusTransactionResult to be mode-agnostic? This is a pre-existing naming issue but this PR's unification makes it more noticeable.

Thanks, @bingyanglin! I renamed `ConsensusCertificateResult` to `ConsensusTransactionResult` and `CancelConsensusCertificateReason` to `CancelConsensusTransactionReason`. More similar renaming will follow in a freshly opened PR https://github.com/iotaledger/iota/pull/10767.

piotrm50 (approved):

vekkiokonio (approved):

PR #10620, opened 2026-03-05
chore(explorer): cleanup amplitude events

### LINKED ISSUE #10619: Cleanup amplitude events
- [ ] Add missing aria-labels
- [ ] Remove unnecessary events from the codebase & amplitude
- [ ] Attach network to the user identitty

begonaalvarezd:

Close https://github.com/iotaledger/iota/issues/10619


apps/explorer/src/components/search/Search.tsx
panteleymonchuk, line 50:
```suggestion
renderSuggestion={(suggestion) => (
<div className="flex cursor-pointer justify-between" aria-label={`Search suggestion for type ${suggestion.type}`}>
```

apps/explorer/src/hooks/useInitialPageView.ts
```diff
@@ -14,14 +14,4 @@ export function useInitialPageView(activeNetwork: string): void {
     useEffect(() => {
         ampli.identify(undefined);
     }, [location.pathname, activeNetwork]);
```
panteleymonchuk: seems ampli.identify don't add any value if we use `setAmplitudeIdentity`. It's just create empty identity. Later we do the same, but with network param. It can be removed imo.

panteleymonchuk (commented):
A few more place that would be good to add aria-label:
- ~~apps/explorer/src/components/activity/Activity.tsx:81-89 - add aria-label `Activity category ${label}`~~
- ~~apps/explorer/src/components/activity/Activity.tsx#L94 - add aria-label `${showSystemTransactions ? "hide" : "show"} system transactions`~~
- ~~apps/explorer/src/pages/checkpoints/CheckpointDetail.tsx#L148 - add aria-label~~
- ~~apps/explorer/src/pages/checkpoints/CheckpointDetail.tsx#L154 - add aria-label~~
- ~~apps/explorer/src/components/ui/FilterList.tsx#L28 - add aria-label to know what filter was selected~~

apps/explorer/src/hooks/useInitialPageView.ts
begonaalvarezd: https://github.com/iotaledger/iota/pull/10620/changes/6b22530a4f593ff4ff72b151eb94a8d9df1e333d

apps/explorer/src/components/search/Search.tsx
begonaalvarezd: There is already visible text, no need for a custom aria label

apps/explorer/src/lib/utils/analytics/amplitude.ts
```diff
@@ -29,7 +30,7 @@ const ANTI_BOT_CONFIG = {
 
 let IS_BOT_CLEARED = false;
 
-export async function initAmplitude() {
+export async function initAmplitude(network?: string) {
```
evavirseda: i think the `network?` param doesn't seem to be used... `initAmplitude` is called in `main.tsx` before the app mounts, so the network isn't available yet and `useAmplitudeIdentity` already takes care of syncing the network, so imo we could just remove it to keep things clean 🙏

evavirseda (changes requested):

apps/explorer/src/lib/utils/analytics/amplitude.ts
begonaalvarezd: pushed

evavirseda (approved):

panteleymonchuk (approved):

PR #10630, opened 2026-03-05
upstream(core): Update name of "shared deletion" indicators across the codebase

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@43a5eba](https://github.com/MystenLabs/sui/commit/43a5ebab73f053e9de08d358fbf5d7ca455c24bd)
- Description:

These are now referred to as "consensus stream ended" (or similar).
ConsensusV2 objects may be removed from consensus without being deleted.

Note:
- GraphQL RPC updates are left for a future PR.
- Actual implementation of marker table writes when a ConsensusV2 object
is transferred to a different owner type is left for a future PR.

Part of #10609

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -5304,7 +5304,7 @@ impl AuthorityState {
     ) -> IotaResult<AuthenticatorFunctionRefForExecution> {
         let account_object = match account_object.object {
             ObjectReadResultKind::Object(object) => Ok(object),
-            ObjectReadResultKind::DeletedSharedObject(version, digest) => {
+            ObjectReadResultKind::ObjectConsensusStreamEnded(version, digest) => {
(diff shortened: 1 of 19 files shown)


bingyanglin (approved):

semenov-vladyslav (approved):

piotrm50 (approved):

miker83z (commented):
Does this name change make sense given the fact that we don't have the `ConsensusV2 ` object?

jkrvivian:
> Does this name change make sense given the fact that we don't have the `ConsensusV2 ` object?

You're absolutely right!!! 
Do we have ConsensusV2 on the roadmap? If so, I’m in favor of porting the rename now to stay aligned with the upstream codebase. Otherwise, should we defer this until we're ready to port the full mechanism?

jkrvivian:
@miker83z We decided to close this PR for now 😄

PR #10633, opened 2026-03-05
feat(iota-core,iota-types): Align transaction validation logic for certificate-less flow

### LINKED ISSUE #10369: pcool: Align transaction validation logic for certificate and certificate-less flow
Ensure that transaction validation in both certificate flow and certificate-less flow performs the same set of actions, such as checking for duplicate execution. Validation logic should also support soft-bundle validation.

### LINKED ISSUE #10634: pcool: Implement post-consensus validation of transactions
Currently each validator only validates transactions that are received in the authority_server. It should also perform similar checks on all of the transactions received from consensus, as nodes cannot trust local validation of other servers. This should be performed post-consensus so everyone evaluates against the same local state.

piotrm50:
# Description of change

Ensures the certificate-less (p-cool / white flag) transaction submission path performs the same validation as the certificate-based (`handle_transaction_impl`) path, and adds handling for edge cases introduced by the certificate-less flow.

Key changes:

- **`handle_transaction_deny_checks`** (`authority.rs`): Extracted deny/input/gas/coin/MoveAuthenticator checks from `handle_transaction_impl` into a reusable method on `AuthorityState`. Returns owned object refs without acquiring locks.
- **`validate_owned_object_versions`** (`execution_cache.rs`, `object_locks.rs`, `passthrough_cache.rs`, `writeback_cache.rs`, `proxy_cache.rs`): New method added to `ExecutionCacheWrite` that validates owned object existence and version/digest match against live objects without acquiring any locks.
- **Validation in white flag path** (`authority_server.rs`): Both the single-tx and soft-bundle paths of `handle_submit_transactions_impl` now call `handle_transaction_deny_checks` and `validate_owned_object_versions` after signature verification and before the reconfiguration lock — matching the gates in `handle_transaction_impl`.
- **Already-executed check** (`authority_server.rs`): If `validate_owned_object_versions` fails, the handler checks via `try_get_executed_effects` whether the transaction was already successfully executed (duplicate submission). If so, returns the executed result instead of an error.
- **Efficient executed-effects retrieval** (`authority_server.rs`): Replaced `get_signed_effects_and_maybe_resign` with `try_get_executed_effects` in the already-executed path to avoid unnecessary effect signing. Also populates `ExecutedData` with effects, events, input objects, and output objects so the `TransactionDriver` can skip a subsequent `get_full_effects` RPC.
- **Epoch boundary bail-out** (`authority_server.rs`): Added early return when the validator is in the middle of an epoch change, preventing user certificates from being accepted into consensus unnecessarily.
- **`SubmitTransactionsResponse` simplification** (`messages_grpc.rs`, `transaction_submitter.rs`): Replaced `results: Vec<SubmitTransactionResult>` with `result: SubmitTransactionResult` since the endpoint handles at most one logical outcome per call.
- **Soft-bundle homogeneity validation** (`consensus_adapter.rs`): Added check that all transactions in a soft bundle share the same gas sponsor.
- **Comprehensive unit tests** (`server_tests.rs`): ~480 lines of new unit tests covering invalid inputs, re-submissions, oversized payloads, gas price mismatches, epoch boundary rejections, and soft-bundle scenarios.

fixes #10369


crates/iota-core/src/consensus_adapter.rs
```diff
@@ -579,14 +579,16 @@ impl ConsensusAdapter {
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<JoinHandle<()>> {
         if transactions.len() > 1 {
-            // In soft bundle, we need to check if all transactions are of UserTransaction
-            // kind. The check is required because we assume this in
-            // submit_and_wait_inner.
+            // Soft-bundle batches must be homogeneous: either all
+            // CertifiedTransaction (certificate flow) or all
+            // UserTransactionV1 (white-flag flow). submit_and_wait_inner
```
piotrm50, line 584: Soft bundles will be removed

crates/iota-core/src/authority_server.rs
```diff
@@ -1461,14 +1461,41 @@ impl ValidatorService {
             // Verify each transaction's user signature individually.
             // TODO: switch to batch verification once a parallel helper exists.
             let tx_verif_guard = metrics.tx_verification_latency.start_timer();
+            let mut verified_transactions = Vec::with_capacity(transactions.len());
             for tx in &transactions {
-                if let Err(e) = epoch_store.verify_transaction(tx.clone()) {
-                    metrics.signature_errors.inc();
-                    return Err(e.into());
+                match epoch_store.verify_transaction(tx.clone()) {
```
piotrm50, line 1466: soft bundle path will be removed

crates/iota-core/src/authority_server.rs
```diff
@@ -1503,33 +1530,85 @@ impl ValidatorService {
                 .expect("single transaction expected: ping and soft-bundle cases already handled");
             let tx_digest = *transaction.digest();
 
+            // Helper to build an Executed response from transaction effects.
```
piotrm50, line 1533: This will be extended to support processing batches of independent transactions

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,352 @@
...
+                );
+                conflict = Some(IotaError::ObjectLockConflict {
+                    obj_ref: *obj_ref,
+                    pending_transaction: locked_by,
+                });
+                break 'conflict_check;
+            }
+
+            match epoch_store.tables() {
+                Ok(tables) => match tables.get_locked_transaction(obj_ref) {
```
piotrm50, line 216: Error handling in this function is improved in a subsequent PR, so we don't drop transactions on DB error anymore.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,352 @@
...
+                }
+            }
+        }
+        if let Some(e) = conflict {
+            dropped.push((digest, e));
+            keep[i] = false;
+            continue;
+        }
+
+        // Check #5: Deny list, gas, ownership, coin deny list, Move
```
piotrm50, line 262: deny checks are currently locally subjective. Will have to be changed before rolling out the new tx flow.

crates/iota-core/src/authority.rs
```diff
@@ -985,8 +991,22 @@ impl AuthorityState {
...
+        Ok(tx_checked_input_objects.inner().filter_owned_objects())
+    }
 
+    async fn handle_transaction_impl(
+        &self,
+        transaction: VerifiedTransaction,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+    ) -> IotaResult<VerifiedSignedTransaction> {
+        // Ensure that validator cannot reconfigure while we are signing the tx
+        let _execution_lock = self.execution_lock_for_signing();
```
filipdulic: not handling failure to aqure lock or error.

crates/iota-core/src/authority.rs
```diff
@@ -878,15 +884,15 @@ impl AuthorityState {
     /// This is a private method and should be kept that way. It doesn't check
```
filipdulic: changed to pub (crate), no longer private.

filipdulic (approved):

crates/iota-core/src/authority.rs
piotrm50: fixed

crates/iota-core/src/authority.rs
piotrm50: moved the comment to a new function to which this still applies

crates/iota-core/src/authority_server.rs
```diff
@@ -1461,14 +1461,41 @@ impl ValidatorService {
...
+                .should_accept_user_certs()
+            {
+                metrics
+                    .num_rejected_tx_in_epoch_boundary
+                    .inc_by(transactions.len() as u64);
+                return Err(IotaError::ValidatorHaltedAtEpochEnd.into());
+            }
+
+            // Content validation: deny checks + owned object version validation.
+            for verified_tx in &verified_transactions {
```
bingyanglin, line 1488: Deny checks run sequentially here, and for large bundles that adds up. Maybe parallelize with `try_join_all` or not necessary?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,351 @@
...
+    let mut dropped: Vec<(TransactionDigest, IotaError)> = Vec::new();
+    let mut seen_keys: HashSet<SequencedConsensusTransactionKey> = HashSet::new();
+    // Locks acquired within this commit. Populated for every transaction that
+    // passes all checks; used by subsequent transactions' conflict checks.
+    let mut current_commit_locks: HashMap<ObjectRef, LockDetails> = HashMap::new();
+    // Index-parallel keep flags: true = keep, false = remove.
+    let mut keep = vec![true; transactions.len()];
+
+    for (i, tx) in transactions.iter().enumerate() {
+        // Check #0: Dedup by ConsensusTransactionKey.
```
bingyanglin, line 101: The comment says "multiple validators" but `ConsensusTransactionKey` includes the authority index, so two validators submitting the same tx get different keys and won't dedup. Maybe fix the comment to say this only catches replay within a single validator's blocks?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,351 @@
...
+            ..
+        }) => transaction.data().transaction_data(),
+        _ => {
+            return Err(IotaError::GenericAuthority {
+                error: "Expected UserTransactionV1 in extract_owned_input_objects".to_string(),
+            });
+        }
+    };
+
+    let owned_objects = transaction_data
```
miker83z, line 345: Use `transaction.input_objects()` to include possible objects coming from the MoveAuthenticator. Currently no owned objects are ever present in a MoveAuthenticator, but it will be convenient for future changes. In general, it is better to use the `SenderSignedData` methods instead of the `TransactionData` ones.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,351 @@
...
+                error = ?e,
+                "UserTransactionV1 failed validity_check post-consensus, dropping"
+            );
+            dropped.push((digest, e));
+            keep[i] = false;
+            continue;
+        }
+
+        // Check #3: Extract owned input objects for lock conflict detection.
+        let owned_inputs = match extract_owned_input_objects(tx) {
```
miker83z, line 159: Here you read the object refs found in the TXs that the users provided as inputs. Immediately after, you'll start looking for conflicts. But how are you sure that the version and digest provided by the users are in sync with the latest version and digest found in the DB?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,351 @@
...
+        // Check #4: Three-tier lock conflict check.
+        // Cheap (HashMap + quarantine + DB lookups); performed before the
+        // expensive deny checks so conflicting transactions are filtered first.
+        //
+        // Tier 1: Local HashMap (current commit).
+        // Tier 2: Consensus quarantine (previous uncommitted commits).
+        // Tier 3: Persistent DB (committed data).
+        let mut conflict: Option<IotaError> = None;
+        'conflict_check: for obj_ref in &owned_inputs {
+            if let Some(&pending_transaction) = current_commit_locks.get(obj_ref) {
```
miker83z, line 188: if the lock is based on the ref instead of the id only ([and it does NOT check against the DB wether the TX is using the latest version](https://github.com/iotaledger/iota/pull/10633/changes#r2976375179)), it means that potentially in the same commit I could refer to the same object twice with different versions.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -0,0 +1,351 @@
...
+            keep[i] = false;
+            continue;
+        }
+
+        // Check #5: Deny list, gas, ownership, coin deny list, Move
+        // authenticator. Only reached if all locks are free — skips the
+        // expensive object loading for transactions that would be dropped
+        // by the lock conflict check.
+        //
+        // Safe to skip signature re-verification: the consensus block verifier
```
miker83z, line 267: Then, I guess, `VerifiedSequencedConsensusTransaction` should contain `VerifiedTransaction` instead of being created here below.

crates/iota-core/src/authority.rs
```diff
@@ -875,18 +881,15 @@ impl AuthorityState {
...
 
-    /// This is a private method and should be kept that way. It doesn't check
-    /// whether the provided transaction is a system transaction, and hence
-    /// can only be called internally.
+    /// Runs deny list, input object validation, gas checks, coin deny list, and
+    /// MoveAuthenticator checks. Returns the owned object refs for optional
+    /// version validation. Does NOT acquire locks or sign the transaction.
     #[instrument(level = "trace", skip_all, fields(tx_digest = ?transaction.digest()))]
-    async fn handle_transaction_impl(
+    pub(crate) async fn handle_transaction_deny_checks(
```
miker83z: `deny_checks` reminds me of something very specific, i.e., tied to `deny list`. Maybe could be renamed in a more generic way? Like `handle_transaction_validation`

crates/iota-core/src/post_consensus_validation.rs
miker83z: I think this is done later, when `handle_transaction_deny_check` is called, `read_objects_for_signing` should perform the operation I describe in here. But is it too late?

crates/iota-core/src/authority.rs
```diff
@@ -875,18 +881,15 @@ impl AuthorityState {
...
         &self,
-        transaction: VerifiedTransaction,
+        transaction: &VerifiedTransaction,
         epoch_store: &Arc<AuthorityPerEpochStore>,
-    ) -> IotaResult<VerifiedSignedTransaction> {
-        // Ensure that validator cannot reconfigure while we are signing the tx
-        let _execution_lock = self.execution_lock_for_signing()?;
```
miker83z, line 888: Removing the lock. Are we sure that `handle_transaction_deny_checks` will always be called sequentially in the new flow?

crates/iota-core/src/authority_server.rs
miker83z: Then maybe we need to reintroduce the lock? (see https://github.com/iotaledger/iota/pull/10633/changes#r2976488655)

crates/iota-core/src/authority.rs
```diff
@@ -911,7 +914,7 @@ impl AuthorityState {
         // call if there is a sender `MoveAuthenticator` signature present in the
         // transaction.
         let (tx_input_objects, tx_receiving_objects, auth_input_objects, account_object) =
-            self.read_objects_for_signing(&transaction, epoch)?;
+            self.read_objects_for_signing(transaction, epoch)?;
```
miker83z: nit: I guess this function would need to be renamed because it is not "for signing" anymore

crates/iota-core/src/authority_server.rs
piotrm50: @miker83z above we check 
```// Early bail-out during epoch boundary, before running expensive deny checks.
            if !epoch_store
                .get_reconfig_state_read_lock_guard()
                .should_accept_user_certs()
```
and after those checks we actually acquire the reconfiguration lock and check `should_accept_user_certs`. As far as I understand the execution lock is there to handle edge cases with fast-path execution. You sign and execute a transaction (and return the results to the user), but it never makes it into a checkpoint by the end of an epoch for whatever reason.

piotrm50: but it's a good idea with executing validation or maybe even whole transactions concurrently in when the client sends multiple transactions. I will not change this here because this branch will be removed, but I will create an issue to parallelise it when handling batches of transaction.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: That's incorrect. Key for `UserTransactionV1` is `ConsensusTransactionKey::UserTransaction(*tx.digest())` and only depends on the transaction digest. Two validators submitting the same tx produce identical keys. This is actually the case for other variants such as EndOfPublish or CheckpointSignature, but there it actually makes sense because each authority sends its own message and they need to be differentiated.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: Good point, however that would require refactoring consensus handler types which is out of scope of this PR. Also, I think `handle_transaction_deny_checks` only requires a VerifiedTransaction because we're re-using existing function. After this is rolled out and the old flow is cleaned up, we can change the signature of `handle_transaction_deny_checks` to use `VerifiedSequencedConsensusTransaction` type directly.

crates/iota-core/src/authority.rs
piotrm50: Lock is moved to the wrapper function `handle_transaction_impl` for the certificate flow. For the pcool flow this lock is not necessary because we don't have a signing step and pre-consensus validation step doesn't need this lock as it checks accept_user_transactions reconfiguration lock. Post consensus it also doens't matter because we will sequence, execute and include in the checkpoint any transaction that arrives before a quorum of EndOfPublish transactions.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: Hm, good point. The transactions are not exactly conflicting on the same resource because they want to use different versions. But if they are processed in a single consensus commit, treat them as conflicting and let the first one through it might still be dropped on the version check and we end up not executing any transaction. 
If we check the version freshness before checking for conflicts then we anyway have to go to RocksDB to load them. However we'd do one rocksdb check less because we would drop the stale object transactions before checking locks on rocksdb. I'll create an issue for this and will improve that in a later PR.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: good point. will change that.

crates/iota-core/src/authority.rs
piotrm50: Renamed this.

crates/iota-core/src/authority.rs
piotrm50: Renamed this and a bunch of other `_for_signing` functions.

crates/iota-core/src/post_consensus_validation.rs
bingyanglin: Might be good (but might not necessary) to clarify this in the code?

bingyanglin (approved):

miker83z (approved):
LGTM!

PR #10641, opened 2026-03-06
feat(indexer): Ensure read after write consistency

### LINKED ISSUE #10465: bug(indexer): Possibility of reading old state after optimistic execution
Currently in optimistic execution pipeline in indexer, if for some reason we don't index tx optimistically but fallback to checkpoint indexing, we return response to the user right after we are able to read a tx from DB.

This doesn't guarantee however that checkpoint indexing is fully complete for this tx, as it is still possible that only transactions table was written but e.g. not the objects table or display table, and in the subsequent request when user would try to construct transaction on newly created objects he would still see and use old objects, which is an error.

tomxey:
# Description of change

This PR introduces a guarantee that after optimistic indexing returns response to the user, the transaction effects like objects and display table are already indexed, and user can create followup transactions right away.

This was most likely already the case most of the time, but right now we have an explicit check to have this guaranteed.

fixes https://github.com/iotaledger/iota/issues/10465

### DIFF
--- a/crates/iota-indexer/src/models/transactions.rs
+++ b/crates/iota-indexer/src/models/transactions.rs
@@ -100,10 +100,11 @@ pub struct CheckpointTxGlobalOrder {
 
 /// Index status.
 #[derive(Clone, Debug, Copy, AsExpression, PartialEq, Eq)]
+#[repr(i64)]
 #[diesel(sql_type = BigInt)]
 pub enum IndexStatus {
-    Started,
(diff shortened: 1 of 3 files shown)


kodemartin (approved):
lgtm 🌅

sergiupopescu199 (approved):
lgtm 🐧

nit: could you please refactor error messages to be  lower-case.

PR #10645, opened 2026-03-06
feat(explorer): Date rendering preference

### LINKED ISSUE #57: Improve date management
We should improve the date display in the explorer, use local storage fot user preferences on displayed time, so that we render time ago or exact date, based on user preference. We could do what IOTAScan does, which is when you render a date that renders based on user preference, the user can change the preference by clicking on i. I would fix the dates in the following places
- Tables like transactions, epochs, checkpoints (in homepage, package, object, address, epoch, checkpoint...)
- In checkpoint point, in the "Checkpoint Timestamp" I would also display the time ago in brackets
- In transaction page, in the "Date" I would also display the time ago in brackets

marc2332:
Closes https://github.com/iotaledger/ts-packages/issues/57


panteleymonchuk:
Closed as we created: https://github.com/iotaledger/ts-packages/pull/64

PR #10644, opened 2026-03-06
feat(gRPC): add `get_health` endpoint

muXxer:
# Description of change

This PR adds the missing `get_health` endpoint, that is available in the REST-API.
The optional threshold value is ms based, instead of seconds, to have finer grained control.

It also contains a field `estimated_validator_latency_ms` reserved for future use.

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #10650, opened 2026-03-06
feat(gRPC): add `MetadataEnvelope` to provide additional headers

muXxer:
# Description of change

This change makes it possible for the client to access the additional metadata that is provided in the headers, to always have the information about the current sync state of the node.

Merge after #10644 


jkrvivian (approved):

crates/iota-grpc-client/src/api/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+        f: impl FnOnce(T) -> Result<U, E>,
+    ) -> Result<MetadataEnvelope<U>, E> {
+        Ok(MetadataEnvelope {
+            inner: f(self.inner)?,
+            metadata: self.metadata,
+        })
+    }
+}
+
+impl<T> Deref for MetadataEnvelope<T> {
```
bingyanglin: I am consider replacing `Deref`/`DerefMut` with explicit `body()`/`body_mut()` methods on `MetadataEnvelope`. Currently `response.chain_id` reads from the protobuf body and `response.chain_id()` reads from the gRPC metadata header. These two can return different values: body fields depend on the `read_mask`, while metadata headers are always set by the server. 

With a `body()` method, body access becomes `response.body().chain_id` and header access stays `response.chain_id()` which might be easier to follow?

crates/iota-grpc-client/src/api/ledger/objects.rs
```diff
@@ -67,9 +70,9 @@ impl Client {
         &self,
         refs: &[(ObjectId, Option<Version>)],
         read_mask: Option<&str>,
-    ) -> Result<Vec<Object>> {
+    ) -> Result<MetadataEnvelope<Vec<Object>>> {
         if refs.is_empty() {
-            return Ok(vec![]);
+            return Ok(MetadataEnvelope::new(vec![], Default::default()));
```
bingyanglin: If the input is empty then the metadata is also empty, but what if the user will use the metadata in this case?

crates/iota-grpc-client/src/api/ledger/transactions.rs
```diff
@@ -113,9 +116,9 @@ impl Client {
         &self,
         digests: &[Digest],
         read_mask: Option<&str>,
-    ) -> Result<Vec<ExecutedTransaction>> {
+    ) -> Result<MetadataEnvelope<Vec<ExecutedTransaction>>> {
         if digests.is_empty() {
-            return Ok(vec![]);
+            return Ok(MetadataEnvelope::new(vec![], Default::default()));
```
bingyanglin: If the input is empty then the metadata is also empty, but what if the user will use the metadata in this case?

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -344,9 +348,13 @@ impl Client {
         }
 
         let mut client = self.ledger_service_client();
-        let stream = client.stream_checkpoint_data(request).await?.into_inner();
+        let response = client.stream_checkpoint_data(request).await?;
+        let (stream, metadata) = MetadataEnvelope::from(response).into_parts();
```
bingyanglin, line 356: Maybe good to add the note in the code doc: the metadata in the returned envelope is captured only when the stream is opened.

crates/iota-grpc-client/src/api/metadata.rs
muXxer: yeah that's much cleaner, thx!

crates/iota-grpc-client/src/api/ledger/objects.rs
muXxer: We return an error now 👍

crates/iota-grpc-client/src/api/ledger/transactions.rs
muXxer: We return an error now 👍

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
muXxer: Added a comment

bingyanglin (approved):

PR #10652, opened 2026-03-08
feat(gRPC): add `filter_checkpoints` flag to skip checkpoints in the stream that don't match the given filters

muXxer:
# Description of change

This PR adds the option to skip checkpoints in the checkpoint stream that don't match any given transaction or events filter. That makes is possible for users to dramatically reduce the amount of checkpoints that are sent over the wire, if they are only interested in the data for their own use case anyways.

Merge after #10644 and #10650 

### DIFF
--- a/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
+++ b/crates/iota-grpc-client/src/api/ledger/checkpoints.rs
@@ -97,8 +97,8 @@ use iota_sdk_types::{CheckpointSequenceNumber, Digest};
 use crate::{
     Client, Error,
     api::{
-        CheckpointResponse, GET_CHECKPOINT_READ_MASK, MetadataEnvelope, Result, TryFromProtoError,
-        field_mask_with_default,
+        CheckpointResponse, CheckpointStreamItem, GET_CHECKPOINT_READ_MASK, MetadataEnvelope,
+        Result, TryFromProtoError, field_mask_with_default,
     },
 };
 
@@ -274,11 +274,17 @@ impl Client {
         let reassembled = Self::reassemble_checkpoint_data_stream(stream);
         futures::pin_mut!(reassembled);
 
-        let checkpoint = reassembled
-            .next()
-            .await
-            .ok_or_else(|| TryFromProtoError::missing("checkpoint data").into())
-            .and_then(|r| r)?;
+        // Skip any progress messages and find the first checkpoint
+        let checkpoint = loop {
+            match reassembled.next().await {
+                Some(Ok(CheckpointStreamItem::Checkpoint(cp))) => break *cp,
+                Some(Ok(CheckpointStreamItem::Progress { .. })) => continue,
+                Some(Err(e)) => return Err(e),
+                None => {
+                    return Err(TryFromProtoError::missing("checkpoint data").into());
+                }
+            }
+        };
 
         Ok(MetadataEnvelope::new(checkpoint, metadata))
     }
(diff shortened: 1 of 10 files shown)


crates/iota-grpc-types/proto/iota/grpc/v0/ledger_service.proto
```diff
@@ -34,6 +37,27 @@ service LedgerService {
   rpc GetEpoch(GetEpochRequest) returns (GetEpochResponse);
 }
 
+message GetHealthRequest {
+  option (iota.grpc.message_accessors) = "with";
+
+  // Optional threshold in milliseconds. If provided, the node is considered
+  // healthy only if the latest executed checkpoint timestamp is within this many
+  // milliseconds of the current system time. If not provided, the node is
+  // healthy if it can fetch the latest checkpoint.
```
bingyanglin: The comments are not very clear to me, maybe:
If not provided, defaults to 5000ms (i.e. the node is considered healthy if the latest checkpoint was executed within the last 5 seconds).
How do you think?

jkrvivian (approved):

crates/iota-grpc-server/src/types.rs
```diff
@@ -962,8 +1061,36 @@ impl GrpcReader {
...
+                                        }
+
+                                        // no filter match, skip processing this checkpoint
+                                        return;
+                                    }
+                                }
+                            }
+
+                            // Pass 2 (or normal mode): full processing
                             let transaction_stream = state_reader_inner.stream_checkpoint_transactions(checkpoint_contents.clone());
```
bingyanglin, line 1096: I am thinking can we just do one DB read? Now we need twice DB reads for every matching checkpoint.

bingyanglin: Maybe use a flag to track whether anything matched, and yield `Progress` instead of `EndMarker` if nothing did?

crates/iota-grpc-server/src/types.rs
```diff
@@ -910,6 +920,88 @@ impl GrpcReader {
...
+            transaction_stream,
+            transaction_filter,
+            event_filter,
+        )
+        .await?
+        {
+            *last_msg_time.lock().unwrap() = tokio::time::Instant::now();
+            Ok(FilterCheckResult::Matched)
+        } else {
+            let elapsed = last_msg_time.lock().unwrap().elapsed();
```
bingyanglin: The mutex is locked twice here. Once to check elapsed(), then again to update the time. Between the two locks another task could also see the interval as elapsed and send a duplicate progress message. Maybe use a single lock for both the check and update?

crates/iota-grpc-server/src/ledger_service/get_checkpoint.rs
```diff
@@ -353,6 +360,24 @@ pub(crate) fn stream_checkpoint_data(
     let (transaction_filter, event_filter) =
         convert_and_validate_filters(req.transactions_filter, req.events_filter)?;
 
+    // Validate filter_checkpoints constraints
```
bingyanglin, line 373: This validation is only added to `stream_checkpoint_data` but not `get_checkpoint_data`. The same filter+mask combination would succeed on one endpoint but fail on the other. Maybe apply this validation to both endpoints for consistency?

crates/iota-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -323,7 +337,131 @@ impl Client {
...
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// # use iota_grpc_client::{Client, CheckpointStreamItem};
+    /// # use futures::StreamExt;
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let client = Client::connect("http://localhost:9000").await?;
+    /// let mut stream = client
+    ///     .stream_checkpoints_filtered(Some(0), None, None, None, None, None)
```
bingyanglin: Need at least one filter?

crates/iota-grpc-server/src/types.rs
muXxer: I would like to address this in another PR.

bingyanglin (approved):

PR #10657, opened 2026-03-09
upstream(core): Use weak ref for metrics thread

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@35156](https://github.com/MystenLabs/sui/commit/35156d3707d589ce202666e0c3d4dcc1be4a6c0c)
- Description:

Regular Arc causes `safe_drop_db` call to fail with `lock hold by current process` exception. Uses a weak reference (`Arc::downgrade`) instead of a strong `Arc` clone for the RocksDB metrics reporting thread to prevent this.

Part of #10609

### DIFF
--- a/crates/typed-store/src/database.rs
+++ b/crates/typed-store/src/database.rs
@@ -478,7 +478,7 @@ impl<K, V> DBMap<K, V> {
         column_family: ColumnFamily,
         is_deprecated: bool,
     ) -> Self {
-        let db_cloned = db.clone();
+        let db_cloned = Arc::downgrade(&db);
         let db_metrics = DBMetrics::get();
         let db_metrics_cloned = db_metrics.clone();
         let cf = column_family.name().to_string();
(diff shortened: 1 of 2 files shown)


bingyanglin (approved):

crates/typed-store/src/rocks/mod.rs
```diff
@@ -498,7 +498,7 @@ impl<K, V> DBMap<K, V> {
         opt_cf: &str,
         is_deprecated: bool,
     ) -> Self {
-        let db_cloned = db.clone();
+        let db_cloned = Arc::downgrade(&db.clone());
```
semenov-vladyslav:
```suggestion
        let db_cloned = Arc::downgrade(&db);
```
No need to create a temporary Arc.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -510,13 +510,14 @@ impl<K, V> DBMap<K, V> {
...
                     tokio::select! {
                         _ = interval.tick() => {
-                            let db = db_cloned.clone();
-                            let cf = cf.clone();
-                            let db_metrics = db_metrics.clone();
-                            if let Err(e) = tokio::task::spawn_blocking(move || {
-                                Self::report_metrics(&db, &cf, &db_metrics);
-                            }).await {
-                                error!("Failed to log metrics with error: {}", e);
+                            if let Some(db) = db_cloned.upgrade() {
```
semenov-vladyslav: Should there be `else` branch with `break;` in it? If the object is dropped weak pointer cannot be upgraded, so the outer loop will continue to be useless.

semenov-vladyslav (changes requested):
I think it could be improved a bit.

semenov-vladyslav (approved):
Looks good, but tests fail now

crates/typed-store/src/rocks/mod.rs
thibault-martinez: I reopened this because I'm wondering why it was resolved without being done

bingyanglin (approved):

crates/typed-store/src/rocks/mod.rs
jkrvivian: Weird, I remember I added them... Thanks for pointing this out!!

thibault-martinez: Thanks, careful when doing rebases 🙏🏻

PR #10662, opened 2026-03-09
chore(dapp-kit): Move `@iota/iota-names-sdk` as peer dep in `@iota/dapp-kit`

marc2332:
This way the dapp importing `@iota/dapp-kit`  decides what version is used (by importing it depending `@iota/iota-names-sdk` itself), and thus not needing to release the dapp-kit for every bump in the names-sdk

[run-ci]


marc2332:
Closing for now

PR #10679, opened 2026-03-10
feat(dashboard): add amplitude `element copied` amplitude event.

panteleymonchuk:


apps/wallet-dashboard/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    return useCallback<MouseEventHandler>(
+        async (e) => {
+            e.stopPropagation();
+            e.preventDefault();
+            await copyToClipboardCore(textToCopy);
+        },
+        [textToCopy, copyToClipboardCore],
+    );
```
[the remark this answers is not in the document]
panteleymonchuk: I wouldn't add it as properties for now. I'll do a check if function exists - call it.

apps/wallet-dashboard/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,102 @@
...
+     * Type of element being copied for analytics tracking
+     * When provided, automatically tracks ampli.elementCopied event
+     */
+    textType?: string;
+    /**
+     * Whether to track the copy event in analytics
+     * @default true
+     */
+    trackEvent?: boolean;
+};
```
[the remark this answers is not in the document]
panteleymonchuk: this type uses for analytic only, better to make it more specifics what it really use for

apps/wallet-dashboard/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { useCallback, type MouseEventHandler } from 'react';
```
evavirseda: i'm not sure that adding the hook here again is the best approach 🙏 it seems that the current approach duplicates the same logic across apps, which makes it harder to maintain 🫣

what i propose is:

1. extend the `useCopyToClipboard` hook in `@iota/core` to handle the toast automatically, with an optional `successMessage` parameter. then we can use it across all apps.

2. create a tracking util in each app (`copyTracking.ts`) that handles `ampli.elementCopied()` locally.

this way:

* we eliminate the duplicated toast + `ampli.elementCopied()` pattern that currently exists in wallet-dashboard and wallet.
* each app keeps control of its own analytics tracking.

evavirseda (changes requested):
🙏

apps/core/src/hooks/useCopyToClipboard.ts
```diff
@@ -24,3 +27,73 @@ export function useCopyToClipboard(onSuccessCallback?: () => void) {
...
+ * Enables decoupling from specific toast/analytics implementations.
+ *
+ * @example
+ * ```typescript
+ * export const { useCopyToClipboard, useCopySuccessCallback } = createCopyToClipboardHooks({
+ *     toast: (message) => toast(message),
+ *     trackCopy: (type) => ampli.elementCopied({ type }),
+ * });
+ * ```
+ */
```
evavirseda: unnecessary 🙏

apps/core/src/hooks/useCopyToClipboard.ts
```diff
@@ -24,3 +27,73 @@ export function useCopyToClipboard(onSuccessCallback?: () => void) {
...
+    const triggerSuccess = (options?: CopyOptions) => {
+        if (options?.successMessage && dependencies.toast) {
+            dependencies.toast(options.successMessage);
+        }
+
+        if (options?.trackEvent !== false && options?.analyticType && dependencies.trackCopy) {
+            dependencies.trackCopy(options.analyticType);
+        }
+    };
+
```
evavirseda: this function is unnecessary, we already have `react-hot-toast` in core, so we don't need to pass dependecies 🙈

evavirseda: also, i didn’t say that we should add `trackEvent` here 🙏 just create a util in the app (wallet, dashboard..) like `trackElementCopied`, and then pass it to the `onSuccess` of this hook wherever we need it

apps/wallet-dashboard/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2024 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { createCopyToClipboardHooks, toast } from '@iota/core';
+import { ampli } from '@/lib/utils/analytics';
+
+/**
+ * Wallet Dashboard-specific copy-to-clipboard hooks with integrated toast and analytics.
+ * Created using dependency injection to decouple from specific implementations.
+ */
```
evavirseda: 🧹

apps/wallet-dashboard/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2024 IOTA Stiftung
```
evavirseda:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

apps/wallet/src/ui/app/hooks/useCopyToClipboard.ts
```diff
@@ -2,35 +2,14 @@
...
-            e.preventDefault();
-            await copyToClipboardCore(textToCopy);
-        },
-        [textToCopy, copyToClipboardCore],
-    );
-}
+/**
+ * Wallet-specific copy-to-clipboard hooks with integrated toast and analytics.
+ * Created using dependency injection to decouple from specific implementations.
+ */
```
evavirseda: 🧹

evavirseda (changes requested):
and please take a look at the copilot comments, and if they don’t make sense, resolve them🙏

apps/core/src/hooks/useCopyToClipboard.ts
panteleymonchuk: imo better to use here dependency injection, so we do not focus on a specific implementation but keep the hook flexible, reusable, and decoupled from any concrete libraries

apps/wallet-dashboard/lib/utils/trackElementCopied.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2024 IOTA Stiftung
```
evavirseda:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

evavirseda (changes requested):
just one detail 🙏

brancoder (changes requested):
conflicts 🙏

evavirseda (approved):

apps/explorer/src/components/ui/InternalLink.tsx
```diff
@@ -48,16 +52,13 @@ function createInternalLink<T extends string>(
...
             event.stopPropagation();
-            if (!navigator.clipboard) {
-                return;
-            }
             if (copyText) {
                 try {
-                    await navigator.clipboard.writeText(copyText);
-                    onCopySuccess();
+                    await copyToClipboard(copyText);
                 } catch (error) {
```
brancoder: this catch can never be reached since the copyToClipboard swallows the error and returns false. 
we can change it to something similar that we have in the PageHeader
```
const success = await copyToClipboard(copyText);
if (!success && onCopyError) {
    onCopyError(new Error('Clipboard write failed'), copyText);
}
```

apps/wallet-dashboard/components/dialogs/ReceiveFundsDialog.tsx
```diff
@@ -64,7 +61,11 @@ export function ReceiveFundsDialog({
                     </div>
                 </DialogBody>
                 <div className="flex w-full flex-row justify-center gap-2 px-md--rs pb-md--rs">
-                    <Button onClick={handleCopyToClipboard} fullWidth text="Copy Address" />
+                    <Button
+                        onClick={() => copyToClipboard(address)}
```
brancoder: maybe we can memoize this function also, wdyt?
const handleCopyAddress = useCallback(() => {
    copyToClipboard(address);
}, [copyToClipboard, address]);

apps/wallet-dashboard/app/(protected)/(feature-controlled)/migration/page.tsx
```diff
@@ -181,10 +180,9 @@ function MigrationDashboardPage(): JSX.Element {
...
     }
 
-    const copyToClipBoard = useCopyToClipboard(() => toast('Address copied'));
-
-    async function handleCopy() {
-        await copyToClipBoard(bech32Address);
+    function onCopySuccess() {
+        toast('Address copied');
+        trackElementCopied('stardust-address');
     }
```
brancoder: we could add callback to all onCopySuccess so the components dont get rerendered on each update.
this is applicable to all files where we have this pattern. wdyt? 
```
const onCopySuccess = useCallback(() => {
    toast('Address copied');
    trackElementCopied('address');
}, []);
```

brancoder (changes requested):

brancoder (approved):
LGTM

PR #10695, opened 2026-03-11
feat(dashboard): update collectible card clicked events for amplitude

panteleymonchuk:


brancoder (approved):
LGTM

evavirseda (approved):

apps/wallet-dashboard/lib/utils/analytics/ampli/index.ts
```diff
@@ -197,6 +188,14 @@ export class ClickedUnstakeIota implements BaseEvent {
     }
 }
 
+export class CollectibleCardClicked implements BaseEvent {
+    event_type = 'collectible card clicked';
```
begonaalvarezd: why was this event renamed? it will be considered a new event in amplitude, lets please keep the same name it had before, we are not renaming events in the dashbaord

begonaalvarezd (changes requested):

apps/wallet-dashboard/lib/utils/analytics/ampli/index.ts
panteleymonchuk: to be consisted with event names that we have in the wallet. Reverted

begonaalvarezd (approved):

PR #10716, opened 2026-03-16
fix(workflow): set libpq library path for macOS arm64 release build

junwei0117:
# Description of change

After the `20260303` runner image update the workflow started failing, but the root cause is both yes and no related to the version update, the original workflow’s macOS PostgreSQL installation steps had some issues, the reason it worked before was likely that the older runner image has PostgreSQL pre-installed and had `libpq.dylib` placed at `/usr/local/lib`, but the new image remove that pre-installation, which exposed the problem, the reason I say likely is that the `20260303` release notes don’t mention removing PostgreSQL either, in fact, they never documented having it pre-installed.

---

Homebrew's `libpq` is `keg-only`, so `brew install postgresql` alone does not place `libpq.dylib` in any default linker search path.

`env:` block also had two issues:
- `$(brew --prefix libpq)` was not shell-expanded
- Variables were not propagated to subsequent steps

Fix by using `$GITHUB_ENV` to persist `LIBRARY_PATH`, `PQ_LIB_DIR`, and `PKG_CONFIG_PATH` with the correct libpq prefix across all steps.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -132,13 +132,14 @@ jobs:
       - name: Install postgres (MacOS arm64)
         if: ${{ matrix.os_type == 'macos-arm64' }}
         shell: bash
-        env:
-          PQ_LIB_DIR: "$(brew --prefix libpq)/lib"
-          LIBRARY_PATH: "/opt/homebrew/lib:$LIBRARY_PATH"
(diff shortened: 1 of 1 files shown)


junwei0117:
@claude review

alexsporn (approved):

muXxer (approved):

PR #10735, opened 2026-03-16
feat(iota-types, iota-framework, iota-execution): Enable EnrichedProgrammableMoveCall and EnrichedCallArg in AuthContext

### LINKED ISSUE #10270: [AA] Enable EnrichedProgrammableMoveCall and EnrichedCallArg in AuthContext

Dkwcs:
[run-ci]

# Description of change

This PR improves the AuthContext that is passed to Move authenticators. It adds type information that the VM already knows, so authenticators can make more precise access-control decisions without needing additional on-chain storage lookups.

###  **IMPORTANT** VM is invoked during authentication
The enriched ProgrammableMoveCall carries two fields that are not present in a PTB:

```
pub is_entry: bool, 
pub returns: Vec<TypeName>,  
```

These are function-level metadata - they live in the module bytecode, not in the transaction. The only way to resolve them is to ask the Move VM, which already knows how to deserialize function definitions from on-chain bytecode:

`vm.load_function(package, module, function)`

**Linkage context requirement**
vm.load_function triggers full module verification for packages that are not yet in the VM's module cache (i.e. freshly published packages that have never been executed in this session). 
During verification, LinkageView::relocate is called to translate a module's runtime ID to its current storage ID - which requires a linkage context to be set.

In the normal execution path, the linkage context is set per-MoveCall just before execution. In the auth enrichment path we must do the same: fetch the package and call set_linkage before invoking vm.load_function.

### Issues are left:

1. The issue is - a PTB may include receiving objects that belong to the main transaction, not to the authenticator. 
The function `authenticate_transaction` (signature verification path) creates its own TemporaryStore, but only with authenticator-related objects, specifically:
authenticator_input_objects → derived solely from MoveAuthenticator.call_args
This does NOT include all objects referenced in the PTB.
As a result:
If the PTB references an object such as object(3,1) (e.g., PackageMetadataV1),
and this object is part of the transaction inputs, not the authenticator inputs.
2. The second issue is -  if the pure ptb input is not used in any command at all, it fails with the error "No type information for pure input at index ...". In real PTBs, this situation is valid - unused inputs are allowed.
Currently it's solved by returning the `TypeTag::Vector(Box::new(TypeTag::U8))).into()` for that input.

Fixes #10270 .

Move unit tests (all enriched-type and existing framework tests):
`cargo run -p iota -- move test --path crates/iota-framework/packages/iota-framework`

Example tests:
`cargo run -p iota -- move test --path examples/move/abstract_iota_accounts/entry_only_account
`
Rust unit tests (BCS round-trip for all new enriched types):
`cargo test -p iota-types --lib auth_context`

Framework build consistency:
`UPDATE=1 cargo test -p iota-framework --test build-system-packages`


iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1742,4 +1748,332 @@ mod checked {
...
+                Some(id) => id,
+                None => continue,
+            };
+            let function_name =
+                match move_core_types::identifier::Identifier::new(move_call.function.clone()) {
+                    Ok(id) => id,
+                    Err(_) => continue,
+                };
+
+            // Load function with empty type arguments (sufficient for
```
bingyanglin: `load_function` with `&[]` type args fails entirely. The VM rejects mismatched type arg counts, so any generic function hits `NUMBER_OF_TYPE_ARGUMENTS_MISMATCH` and the `else { continue }` skips it. No inputs get type names or mutability, even concrete ones like `address`. This affects most real transactions (`public_transfer<T>`, `coin::split<T>`, etc). Maybe convert `move_call.type_arguments` to `Type` values and pass them instead of `&[]`? Or do I miss anything?

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1742,4 +1748,332 @@ mod checked {
...
+            Some(id) => id,
+            None => return EnrichedCommand::from(cmd),
+        };
+        let function_name =
+            match move_core_types::identifier::Identifier::new(move_call.function.clone()) {
+                Ok(id) => id,
+                Err(_) => return EnrichedCommand::from(cmd),
+            };
+
+        let (is_entry, returns) = match load_function_instantiation_and_is_entry(
```
bingyanglin: `load_function_instantiation_and_is_entry` is called for the same function twice (here and Line 1875). Maybe cache the result from the first call.

crates/iota-types/src/auth_context/fields_v1.rs
```diff
@@ -221,150 +243,194 @@ mod tests {
         bcs::from_bytes(&bytes).unwrap()
     }
 
-    // ── MoveCallArg ───────────────────────────────────────────────────
+    // ── EnrichedCallArg round-trips ────────────────────────────────────────
```
bingyanglin: The Move structs say "BCS layout must stay in sync with Rust" but only Rust-side round-trips are tested. Maybe add a test that checks bytes match what Move expects.

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: 4b9d1e511316b4b5f071cd6cc2a93821abc89473

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: 4b9d1e511316b4b5f071cd6cc2a93821abc89473

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+    // -------------------------------------------------------------------------
+    // Auth-context enrichment helpers
+    // -------------------------------------------------------------------------
+
+    /// Converts a runtime `Type` to a `TypeName`.
+    ///
+    /// References are stripped before conversion so that `&T` and `&mut T`
+    /// both yield the name of `T`.  Returns `None` for unresolvable types such
+    /// as open type parameters.
+    fn type_to_type_name(vm: &MoveVM, ty: &Type) -> Option<TypeName> {
```
miker83z: We don't need this function as it is. In `iota-adapter/src/programmable_transactions/context.rs` let's try to not use TypeName. It is ok to return the TypeTag from this function, then on the caller site we convert TypeTag into TypeName.

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+            other => other,
+        };
+        vm.get_runtime()
+            .get_type_tag(base)
+            .ok()
+            .map(|tag| TypeName::from(&tag))
+    }
+
+    /// Returns `true` when `ty` is a mutable-reference type (`&mut T`).
+    fn is_mutable_runtime_ref(ty: &Type) -> bool {
```
miker83z: This function is redundant, just use the matches!

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+
+    /// Loads a function from the VM and returns the loaded function
+    /// instantiation, whether it is marked `entry` in the compiled bytecode,
+    /// and the list of return-type names.
+    ///
+    /// `linkage_view` must **not** have a linkage context set when called.
+    /// On success (and on most errors) it is left in the reset state.
+    fn load_function_instantiation_and_is_entry(
+        vm: &MoveVM,
+        linkage_view: &mut LinkageView,
```
miker83z: We don't need to pass the `linkage_view: &mut LinkageView,` here. Just pass `state_view: &dyn ExecutionState,` as the other functions already defined above (see `load_object`, `load_call_arg`, `load_object_arg`).

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+    /// Builds enriched inputs and commands for `AuthContext` from a PTB.
+    ///
+    /// Uses the Move VM to resolve function types (`is_entry`, return types,
+    /// and parameter types for pure-arg type names and object mutability).
+    /// Object type names are resolved from `objects`, which should be the
+    /// `authenticator_input_objects` map.
+    ///
+    /// Errors during VM function loading are silently ignored; the affected
+    /// command is enriched with empty/default metadata.
+    pub fn create_enriched_auth_context_components(
```
miker83z: This function should not be part of this module. It should be moved to `iota-adapter/src/execution_engine.rs`

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+                })
+            }
+        }
+    }
+
+    /// Converts a `Command` to an `EnrichedCommand`.
+    ///
+    /// `call_enrichment` carries the `(is_entry, returns)` pair resolved by
+    /// the VM for `MoveCall` commands; pass `None` to fall back to defaults.
+    fn enrich_command(
```
miker83z: This function should not be part of this module. It should be moved to `iota-types/src/auth_context/enriched_fields.rs`

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1722,338 @@ mod checked {
...
+            .filter_map(|ty| type_to_type_name(vm, ty))
+            .collect();
+
+        linkage_view.reset_linkage();
+        Ok((loaded_fn, is_entry, returns))
+    }
+
+    /// Enriches an `ObjectArg` into an `EnrichedCallArg`, filling type name
+    /// from the pre-built `objects` map and mutability from `input_is_mutable`.
+    fn enrich_object_arg(
```
miker83z: This function should not be part of this module. It should be moved to `iota-types/src/auth_context/enriched_fields.rs`

Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: f0cebcc6e0044f2bd947d625860898e0bebd1b9c

iota-execution/latest/iota-move-natives/src/auth_context.rs
```diff
@@ -240,6 +240,188 @@ pub fn native_replace(
...
 
+/// ****************************************************************************
+/// native fun native_enriched_tx_inputs
+/// Implementation of the Move native function
+/// `fun native_enriched_tx_inputs<I>(): &vector<I>`
+///
+/// Reuses `auth_context_tx_inputs_cost_params` — the work is structurally
+/// identical to `native_tx_inputs` (BCS-serialize a vector of enriched inputs).
+/// ****************************************************************************
+pub fn native_enriched_tx_inputs(
```
miker83z: instead of adding the 2 new native enriched functions, I think we can reuse `native_tx_inputs` and `native_tx_commands`. 
For instance, in `native_tx_inputs`, when this is obtained: `let input_move_layout = resolve_move_layout(context, &input_type)?;`, then we can check 
`if input_move_layout is iota::ptb_call_arg::CallArg then auth_context.tx_inputs_ref(input_move_layout)?` else `if input_move_layout is iota::enriched_call_arg::EnrichedCallArg then auth_context.enriched_tx_inputs_ref(input_move_layout)?`

crates/iota-framework-snapshot/bytecode_snapshot/22/0x0000000000000000000000000000000000000000000000000000000000000002
miker83z, line 1: Revert this v22 changes and move it to v23

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2589,7 +2589,6 @@ impl ProtocolConfig {
                     cfg.auth_context_tx_inputs_cost_per_byte = Some(2);
                     cfg.auth_context_replace_cost_base = Some(30);
                     cfg.auth_context_replace_cost_per_byte = Some(2);
-
```
miker83z, line 2625: revert this

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+
+/// Enriched representation of an immutable-or-owned object argument.
+///
+/// Extends the plain `(ObjectID, SequenceNumber, ObjectDigest)` tuple with
+/// mutability information and the fully-qualified type name of the object.
+///
+/// `mutable` is `true` when the object is borrowed as `&mut T` in at least one
+/// command of the PTB.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct ImmOrOwnedObjectArg {
```
miker83z: All the types representing their Move counterparts should start with `Move...`
```suggestion
pub struct MoveImmOrOwnedObjectArg {
```

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+    pub mutable: bool,
+    pub type_name: TypeName,
+}
+
+/// Enriched representation of a shared object argument.
+///
+/// Adds the current object digest (looked up from state) and the
+/// fully-qualified type name on top of the plain shared-object fields.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct SharedObjectArg {
```
miker83z:
```suggestion
pub struct MoveSharedObjectArg {
```

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+
+// ---------------------------------------------------------------------------
+// EnrichedCallArg
+// ---------------------------------------------------------------------------
+
+/// Enriched counterpart of CallArg.
+/// Adds type names for pure values and objects, mutability information for
+/// immutable/owned objects, and the current digest for shared objects.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, JsonSchema)]
+pub enum EnrichedCallArg {
```
miker83z:
```suggestion
pub enum MoveEnrichedCallArg {
```

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+// ---------------------------------------------------------------------------
+// EnrichedProgrammableMoveCall
+// ---------------------------------------------------------------------------
+
+/// Enriched counterpart of [`crate::auth_context::MoveProgrammableMoveCall`].
+///
+/// Adds `is_entry` and the list of return-type names on top of the base call
+/// data.
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+pub struct EnrichedProgrammableMoveCall {
```
miker83z:
```suggestion
pub struct MoveEnrichedProgrammableMoveCall {
```

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+/// and the existing `MoveCommand` / `Command` ordering):
+/// * 0 – `MoveCall`
+/// * 1 – `TransferObjects`
+/// * 2 – `SplitCoins`
+/// * 3 – `MergeCoins`
+/// * 4 – `Publish`
+/// * 5 – `MakeMoveVec`
+/// * 6 – `Upgrade`
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+pub enum EnrichedCommand {
```
miker83z:
```suggestion
pub enum MoveEnrichedCommand {
```

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+) -> EnrichedCallArg {
+    let mutable = *input_is_mutable.get(&input_idx).unwrap_or(&false);
+    match obj_arg {
+        ObjectArg::ImmOrOwnedObject((id, version, digest)) => {
+            let type_name = objects
+                .get(id)
+                .and_then(|o| o.struct_tag())
+                .map(|tag| TypeName {
+                    name: tag.to_canonical_string(false),
+                })
```
miker83z: It exists the function from TypeTag into TypeName, here it should be used

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -566,7 +573,21 @@ mod checked {
             let TransactionKind::ProgrammableTransaction(ptb) = &transaction_kind else {
                 unreachable!("Only programmable transactions are allowed");
             };
-            AuthContext::new_from_components(authenticator.digest(), ptb)
+            let input_objects_map: HashMap<_, _> = authenticator_input_objects
+                .iter_objects()
+                .map(|o| (o.id(), o))
+                .collect();
```
miker83z: This is redundant, you can pass `authenticator_input_objects` directly into `create_enriched_auth_context_components`

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2043,4 +2064,115 @@ mod checked {
...
+                let idx = i as u16;
+                match arg {
+                    CallArg::Pure(bytes) => EnrichedCallArg::Pure {
+                        value: bytes.clone(),
+                        type_name: input_type_names.remove(&idx).unwrap_or_else(|| TypeName {
+                            name: String::new(),
+                        }),
+                    },
+                    CallArg::Object(obj_arg) => {
+                        enrich_object_arg(obj_arg, objects, &input_is_mutable, idx)
```
miker83z: You don't need to pass the full hashmap of objects here. 
If you pass authenticator_input_objects into create_enriched_auth_context_components   (see https://github.com/iotaledger/iota/pull/10735/changes#r2961545860), then you can use `obj_arg.id()` to get the unique `Object` you need from authenticator_input_objects. Then you pass this object  to enrich_object_arg. If you don't find the object in authenticator_input_objects then that's an invariant error

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2043,4 +2064,115 @@ mod checked {
...
+        let enriched_inputs = ptb
+            .inputs
+            .iter()
+            .enumerate()
+            .map(|(i, arg)| {
+                let idx = i as u16;
+                match arg {
+                    CallArg::Pure(bytes) => EnrichedCallArg::Pure {
+                        value: bytes.clone(),
+                        type_name: input_type_names.remove(&idx).unwrap_or_else(|| TypeName {
```
miker83z: This is an invariant error. We must be strict, if we do not find the type then it is a big issue. Covering it with an empty string does not solve the issue.

iota-execution/latest/iota-adapter/src/execution_engine.rs
miker83z: (btw I am not against the hashmap, it can be created around here to make the search inside authenticator_input_objects more efficient)

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2043,4 +2064,115 @@ mod checked {
...
+                        enrich_object_arg(obj_arg, objects, &input_is_mutable, idx)
+                    }
+                }
+            })
+            .collect();
+
+        // Pass 3: build enriched commands.
+        let enriched_commands = ptb
+            .commands
+            .iter()
```
miker83z: Why iterate 2 times? Cannot call the enrich command directly in the first iteration (pass 1)?

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+                .unwrap_or_else(|| TypeName {
```
miker83z: Never use empty string, see https://github.com/iotaledger/iota/pull/10735/changes#r2961599638. In this function we should never have an unwrap or a fail. Entering this function we should already found all possible invariant error if the object does not exist.

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+                        TypeName {
+                            name: tag.to_canonical_string(false),
+                        },
+                        digest,
+                    ))
+                })
+                .unwrap_or_else(|| {
+                    (
+                        TypeName {
+                            name: String::new(),
```
miker83z: See https://github.com/iotaledger/iota/pull/10735/changes#r2961641767

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+            mutable: shared_mutable,
+        } => {
+            let (type_name, digest) = objects
+                .get(id)
+                .and_then(|o| {
+                    let tag = o.struct_tag()?;
+                    let digest = o.digest();
+                    Some((
+                        TypeName {
+                            name: tag.to_canonical_string(false),
```
miker83z: See https://github.com/iotaledger/iota/pull/10735/changes#r2961536165. I won't add more of these comments for all the occurrences below.

crates/iota-types/src/auth_context/enriched_fields.rs
miker83z: I won't add more of these comments for all the occurrences below.

crates/iota-types/src/auth_context/enriched_fields.rs
```diff
@@ -0,0 +1,534 @@
...
+/// an empty returns list.
+pub fn enrich_command(
+    cmd: &Command,
+    call_enrichment: Option<&(bool, Vec<TypeName>)>,
+) -> EnrichedCommand {
+    match cmd {
+        Command::MoveCall(call) => {
+            let (is_entry, returns) = call_enrichment
+                .map(|(e, r)| (*e, r.clone()))
+                .unwrap_or((false, vec![]));
```
miker83z: Nope

crates/iota-types/src/auth_context/mod.rs
```diff
@@ -43,25 +43,36 @@ pub const AUTH_CONTEXT_STRUCT_NAME: &IdentStr = ident_str!("AuthContext");
...
+    tx_commands: Vec<EnrichedCommand>,
 }
 
 impl AuthContext {
+    /// Construct an `AuthContext` from pre-built enriched inputs and commands.
+    ///
+    /// Callers (e.g. the execution engine) are responsible for building the
+    /// enriched data, which may include object-type resolution via the backing
+    /// store and function-metadata resolution via the VM.
     pub fn new_from_components(
```
miker83z: should be renamed to `new_with_enriched`

crates/iota-types/src/auth_context/mod.rs
```diff
@@ -77,11 +88,11 @@ impl AuthContext {
         &self.auth_digest
     }
 
-    pub fn tx_inputs(&self) -> &Vec<MoveCallArg> {
+    pub fn tx_inputs(&self) -> &Vec<EnrichedCallArg> {
```
miker83z: Should be kept as `pub fn tx_inputs(&self) -> &Vec<MoveCallArg>` that invokes `&self.tx_inputs.into()`. Then create another function `pub fn enriched_tx_inputs(&self) -> &Vec<EnrichedCallArg>`.

crates/iota-types/src/auth_context/mod.rs
```diff
@@ -77,11 +88,11 @@ impl AuthContext {
...
         &self.tx_inputs
     }
 
-    pub fn tx_commands(&self) -> &Vec<MoveCommand> {
+    pub fn tx_commands(&self) -> &Vec<EnrichedCommand> {
```
miker83z: See https://github.com/iotaledger/iota/pull/10735/changes#r2961667740

crates/iota-types/src/type_input.rs
```diff
@@ -11,11 +11,12 @@ use move_core_types::{
     identifier::Identifier,
     language_storage::{StructTag, TypeTag},
 };
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use crate::parse_iota_type_tag;
 
-#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize, JsonSchema)]
```
miker83z: Where is it needed?

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1717,6 +1717,88 @@ mod checked {
...
+    // -------------------------------------------------------------------------
+
+    /// Converts a runtime [`Type`] to a [`TypeTag`].
+    ///
+    /// References are stripped before conversion so that `&T` and `&mut T`
+    /// both yield the tag of `T`.  Returns `None` for unresolvable types such
+    /// as open type parameters.
+    pub fn type_to_type_tag(vm: &MoveVM, ty: &Type) -> Option<TypeTag> {
+        let base = match ty {
+            Type::Reference(inner) | Type::MutableReference(inner) => inner.as_ref(),
```
miker83z: Try with:
```suggestion
            Type::Reference(inner) | Type::MutableReference(inner) => type_to_type_tag(vm, inner.as_ref()),
```
If it happens that we get a double reference.

miker83z: Additionally, for `Type::TyParam(_)` it should directly return `None`

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2043,4 +2064,115 @@ mod checked {
...
+                    call.package,
+                    &call.module,
+                    &call.function,
+                    &type_tags,
+                )
+            else {
+                continue;
+            };
+
+            let returns = return_tags.iter().map(TypeName::from).collect();
```
miker83z: This is a bit more complex. Ping me and we will discuss how to implement it.

iota-execution/latest/iota-move-natives/src/auth_context.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/mod.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/mod.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/type_input.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/enriched_fields.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-types/src/auth_context/mod.rs
Dkwcs: 8ab71267eb3252b57720e50d4ee8fdb7e704a248

crates/iota-framework-snapshot/bytecode_snapshot/22/0x0000000000000000000000000000000000000000000000000000000000000002
Dkwcs: d1a986b0a77f93752a2ee0334cc8656d1c49278c

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2087,4 +2106,261 @@ mod checked {
...
+                    )
+                })?;
+
+            // Zip call arguments with their parameter types.
+            for (j, arg) in call.arguments.iter().enumerate() {
+                let Argument::Input(i) = arg else { continue };
+                let i = *i;
+
+                let Some(param_ty) = loaded_fn.parameters.get(j) else {
+                    continue;
```
miker83z: This should be an invariant error. It means that the index of the arguments and the parameters do not match, but this should not happen

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2087,4 +2106,261 @@ mod checked {
...
+        // arguments via `type_to_type_tag_with_subst`.
+        //
+        // For built-in commands (`TransferObjects`, `SplitCoins`, etc.): the
+        // VM has no knowledge of them (they are not Move functions). Types
+        // are inferred from protocol constants or from PTB-supplied type
+        // annotations via `index_pure_inputs_for_builtin`.
+        let mut enriched_commands = Vec::with_capacity(ptb.commands.len());
+        for cmd in ptb.commands.iter() {
+            let Command::MoveCall(call) = cmd else {
+                index_pure_inputs_for_builtin(cmd, &mut input_type_names);
```
miker83z:
```suggestion
                for (arg, tag_opt) in pure_param_types_for_cmd(cmd) {
                    if let Some(tag) = tag_opt {
                        if let Argument::Input(i) = arg {
                            input_type_names
                                .entry(*i)
                                .or_insert_with(|| TypeName::from(&tag));
                        }
                    }
                }
```

`index_pure_inputs_for_builtin` and `set_pure_input_type` are not needed, these just make it more difficult to read the code.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2087,4 +2106,261 @@ mod checked {
...
+    ///   `TransferObjects(_, recipient)` → `Some(address)` `SplitCoins(_,
+    ///   amounts)`        → `Some(u64)` per amount
+    ///
+    /// - **PTB-supplied type annotation** — provided explicitly by the sender:
+    ///   `MakeMoveVec(Some<T>, items)`   → `Some(T)` per item
+    ///   `MakeMoveVec(None, items)`      → `None` per item (type unknown)
+    ///
+    /// Returns `None` when the type cannot be determined; the argument will
+    /// receive an empty-name sentinel in the enriched output.
+    fn pure_param_types_for_cmd(cmd: &Command) -> Vec<(&Argument, Option<TypeTag>)> {
```
miker83z: Move it to `iota-types`

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2087,4 +2106,261 @@ mod checked {
...
+                            }
+                        }
+                    }
+                    Some(CallArg::Object(_)) => {
+                        // Track whether this object input is passed mutably.
+                        // Pass 2 can set the correct flag in the enriched arg.
+                        let entry = input_is_mutable.entry(i).or_insert(false);
+                        if matches!(param_ty, Type::MutableReference(_)) {
+                            *entry = true;
+                        }
```
miker83z: This is wrong. The mutability could be inferred using references, but that is not the way to to it here. First of all, even if we pass the object by value, then it is mutable as well (and this code does not cover this case). 
Second, we don't need the parameter information; we just need to know if the `CallArg::Object` is either a `SharedObject` with `mutable` flag or `!object.is_immutable();`. 
You don't need to do this here, those information can be obtained after line 2356. You don't need the `input_is_mutable` map either.

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2087,4 +2106,261 @@ mod checked {
...
+                        // but can present in the PTB inputs.
+                        let object = objects.get(&obj_arg.id()).copied().ok_or_else(|| {
+                            ExecutionError::new_with_source(
+                                ExecutionErrorKind::VMInvariantViolation,
+                                format!(
+                                    "Object {} not found in authenticator input objects",
+                                    obj_arg.id()
+                                ),
+                            )
+                        })?;
```
miker83z, line 2309: Check mutability here (see https://github.com/iotaledger/iota/pull/10735/changes#r2993402240)

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1722,6 +1722,139 @@ mod checked {
...
+    /// instantiation, whether it is marked `entry` in the compiled bytecode,
+    /// and the list of return-type tags.
+    pub fn load_function_instantiation_and_is_entry(
+        vm: &MoveVM,
+        state_view: &dyn ExecutionState,
+        package_id: ObjectID,
+        module_name: &str,
+        function_name: &str,
+        type_tags: &[TypeTag],
+    ) -> VMResult<(LoadedFunctionInstantiation, bool, Vec<Option<TypeTag>>)> {
```
miker83z: The `returns` must not have `None`s, so it must be `Vec<TypeTag>` instead of `Vec<Option<TypeTag>>`.

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1722,6 +1722,139 @@ mod checked {
...
+        let compiled = vm.load_module(&module_id, &linkage_view)?;
+        let is_entry = compiled.function_defs().iter().any(|fdef| {
+            let fhandle = compiled.function_handle_at(fdef.function);
+            compiled.identifier_at(fhandle.name) == fn_ident.as_ident_str() && fdef.is_entry
+        });
+
+        let returns = loaded_fn
+            .return_
+            .iter()
+            .map(|ty| type_to_type_tag_with_subst(vm, ty, type_tags))
```
miker83z: Throw an error if you find a None (see https://github.com/iotaledger/iota/pull/10735/changes#r2993423363)

Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: 3e31ff7daf38fc97dd37520b16dcde05be0eb03e

miker83z:
To be re-opened later

Issue #9965, opened 2026-01-28
Improve readability and flow of IOTA Identity Workshop


salaheldinsoliman:
I have identified the following issues and tasks list:

Issues for readability:
- Since the logic of DID/VC creation, signing and verification happens in the Rust back-end, having the whole project scaffold and explained right at the start might hinder readability

I suggest the following:

- [ ] The core of the workshop is in the Rust backend. It should be a standalone module/part under which there is a clear separation between the functionalities (VC creation, Issuance, Validation)
- [ ] Project pre-requisites should strictly be Rust/Cargo, any other tool is optional and shouldn't be mentioned in the workshop.
- [ ] For every core part explained, the Rust backend should be called via a POST/GET request for simplicity (A terminal or Postman maybe)
- [ ] The front-end part should be functionally complete at the start of the workshop, with only the specific API calls being filled in to both test the backend service and reinforce the connection to the use-case. Displaying the response object and also any blockchain changes via RPC could be useful here.
- [ ] For every core part explained and executed, the effects (Tx hashes, DIDs, etc...) should be printed and shown, preferably via the IOTA explorer

Ginowine:
Thanks, @salaheldinsoliman, for suggesting these updates. Let us discuss some of this in a meeting with Richard in the meeting. I understand the approach you are suggesting, but this was what Richard asked us to do. I look forward to discussing a few of these updates with Richard in the workshop coordination meeting today.

PR #10736, opened 2026-03-16
feat(docs): restructure identity workshop

salaheldinsoliman:
# Description of change

Restructured the IOTA Identity workshop.

#9965 

Docs built locally

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


richardmands (changes requested):
For each of the main steps, I would try to include a clear reference back to the use case when you mention them.

E.g.

Step 1 - Create unique identities for the student and university on the IOTA network.
Step 2 - University issues a degree to the student. This is in the form of a VC...
Step 3 - Student now wants to show their degree to a prospective employer. They can do this  with a VP...
Step 4 - Employer confirms that the student's degree, was issued by the university.

Just to keep it extremely clear what the use case is that's being described in each step. 

This is already apparent in the language on the FE buttons, e.g. 'Validate Degree' or 'Show Degree to Employer' but probably worth including it in the main text as well.

Other than that, looks good to me.

PR #10753, opened 2026-03-17
feat(indexer): Simplify tx_global_order semantics and track object finality via checkpoint

### LINKED ISSUE #10462: bug(indexer): Checkpoint indexing can create ghost objects
> such ghost object can only be created on optimistic path,
> deletion of object on optimistic path followed by creation of ghost object on checkpoint path is impossible, since for object to be deleted on optimistic path, it's creation needs to be already indexed, and this will prevent checkpoint indexing pipeline from indexing the same change again
> 
> Actually there is a possibility of indexing the same object change on checkpoint path more than once.
> 
> If indexer would create **ObjectA** in scope of **TX1** but then crash right away without accounting that given checkpoint is indexed, then it is possible that followup optimistic transaction **TX2** will be indexed after that that depends on **ObjectA** and that deletes **ObjectA**.
> 
> Checkpoint indexing then, when resumed will restart from the checkpoint that was last processed and write **ObjectA** again in scope of **TX1**, but later when processing the **TX2** that is supposed to delete this object, it will not do it because **TX2** will already be marked as optimistically indexed and checkpoint indexing is skipping such changes. So we will be left out with a ghost **ObjectA**.
> 
> I find it way less likely to happen than the previous case, since it requires indexer to crash and those are not normal operations. That's a theoretical possibility however.
> 
> The solutions pointed out above (version check on delete, ordering `object` updates after `tx_global_order` updates) will not solve this.
> Ideally we could check on insertion whether such object already existed or was deleted at a higher version, but it's not a trivial check, since we don't have this information directly available in the `objects` table.
> 
> My suggestion would be to move this to a separate issue. 

 _Originally posted by @tomxey in [#10250](https://github.com/iotaledger/iota/issues/10250#issuecomment-3928016017)_

tomxey:
# Description of change

  - **Remove `IndexStatus` enum and dual semantics from `tx_global_order`**: The `optimistic_sequence_number` column no longer encodes indexing status (`Started=0`, `Completed=-1`). Checkpoint transactions simply use `-1`, optimistic transactions use positive auto-generated values.
  - **Track object finality via `finalized_in_cp` column**: Objects store the checkpoint sequence number at which they were (or will be) indexed. Readers consider an object finalized when `finalized_in_cp IS NULL` (optimistic/already finalized) or when the referenced checkpoint has been fully indexed.
  This removes the need for an explicit finalization step after checkpoint persistence.
  - **Eliminate the `update_tx_seq_for_checkpoint_transactions` step**: `chk_tx_sequence_number` is now set via upsert during `persist_tx_global_order`, removing a separate post-persist update step.
  - **Update dependency check in optimistic indexing**: `count_existing_object_keys` now checks `finalized_in_cp IS NULL OR finalized_in_cp <= max indexed checkpoint` ensuring optimistic transactions only proceed when their input objects are fully indexed.
  - **Merge `CheckpointTxGlobalOrder` into `TxGlobalOrder`**: The two structs were identical after removing `IndexStatus`, so they are unified into one.
  - **Update `is_transaction_fully_indexed`**: For optimistic txs, checks `optimistic_sequence_number > 0`. For checkpoint txs, checks whether the latest indexed checkpoint covers the transaction.

fixes https://github.com/iotaledger/iota/issues/10462

### DIFF
--- a/crates/iota-indexer/src/ingestion/primary/persist.rs
+++ b/crates/iota-indexer/src/ingestion/primary/persist.rs
@@ -155,18 +155,7 @@ impl PrimaryWriter {
         let packages_batch = packages_batch.into_iter().flatten().collect::<Vec<_>>();
         let checkpoint_num = checkpoint_batch.len();
         let tx_count = tx_batch.len();
-
-        // Collect object IDs from the batch before they are consumed, so we can
-        // finalize them after checkpoint persistence is complete.
-        let object_ids_to_finalize: Vec<Vec<u8>> = object_changes_batch
-            .iter()
-            .flat_map(|changes| {
-                changes
-                    .changed_objects
-                    .iter()
-                    .map(|o| o.object().id().to_vec())
-            })
-            .collect();
+        let max_checkpoint_seq = committer_watermark.max_committed_cp;
 
         {
             let _step_1_guard = self
@@ -192,7 +181,7 @@ impl PrimaryWriter {
                         .persist_tx_global_order(tx_global_order_batch.clone())
                         .await?;
                     self.state
-                        .persist_checkpoint_objects(object_changes_batch)
+                        .persist_checkpoint_objects(object_changes_batch, max_checkpoint_seq)
                         .await
                 }),
             ];
@@ -211,22 +200,6 @@ impl PrimaryWriter {
                 .collect::<IndexerResult<Vec<_>>>()
                 .expect("persisting data into DB should not fail.");
         }
-
-        self.state
-            .update_status_for_checkpoint_transactions(tx_global_order_batch)
-            .await
-            .inspect_err(|e| {
-                error!("failed to update tx global order as indexed with error: {e}");
-            })
-            .expect("updating tx global order as indexed should not fail.");
-
-        self.state
-            .finalize_objects(object_ids_to_finalize)
-            .await
-            .inspect_err(|e| {
-                error!("failed to finalize objects with error: {e}");
-            })
-            .expect("finalizing objects should not fail.");
 
         let is_epoch_end = epoch.is_some();
 
--- a/crates/iota-indexer/src/ingestion/primary/persist.rs
+++ b/crates/iota-indexer/src/ingestion/primary/persist.rs
@@ -156,6 +156,18 @@ impl PrimaryWriter {
         let checkpoint_num = checkpoint_batch.len();
         let tx_count = tx_batch.len();
 
+        // Collect object IDs from the batch before they are consumed, so we can
+        // finalize them after checkpoint persistence is complete.
+        let object_ids_to_finalize: Vec<Vec<u8>> = object_changes_batch
+            .iter()
+            .flat_map(|changes| {
+                changes
+                    .changed_objects
+                    .iter()
+                    .map(|o| o.object().id().to_vec())
+            })
+            .collect();
+
         {
             let _step_1_guard = self
                 .metrics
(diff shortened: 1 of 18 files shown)


crates/iota-indexer/src/models/objects.rs
```diff
@@ -52,6 +52,13 @@ pub struct StoredObject {
     // TODO deal with overflow
     pub coin_balance: Option<i64>,
     pub df_kind: Option<i16>,
+    /// The checkpoint sequence number at which this object was, or will be,
+    /// indexed. Readers can consider the object finalized when this
+    /// checkpoint has been marked as fully indexed.
+    ///
+    /// `None` means the object is already finalized (default for existing
+    /// objects and objects written by the optimistic path).
+    pub finalized_in_cp: Option<i64>,
```
kodemartin, line 61: Why not plain `chk_sequence_number`?

crates/iota-indexer/src/models/objects.rs
kodemartin: Is there a reason we don't use this instead of `finalized_in_cp`?

crates/iota-indexer/src/ingestion/primary/persist.rs
```diff
@@ -180,7 +181,7 @@ impl PrimaryWriter {
                         .persist_tx_global_order(tx_global_order_batch.clone())
                         .await?;
                     self.state
-                        .persist_checkpoint_objects(object_changes_batch)
+                        .persist_checkpoint_objects(object_changes_batch, max_checkpoint_seq)
```
kodemartin: Why do we use `max_checkpoint_seq` and not the actual sequence number of the checkpoint that included the object?

With the latter we register also the relation between checkpoints and objects in this table, which could come handy. E.g. in queries about the checkpoint the included the object.

crates/iota-indexer/src/models/objects.rs
kodemartin: The changes in this PR might even warrant refactoring `checkpoint_sequence_number` to an `Option<_>` in `IndexedObject`.

To make the conversions in the optimistic path more transparent.

crates/iota-indexer/src/models/objects.rs
tomxey: This would give a feeling that this is a general purpose connection between object and cp it was last modified in, but in fact we are not supporting such semantics right now.

`finalized_in_cp` is assigned only for objects commited through checkpoint path and is NULL for old objects or optimistic objects, so if someone would try to use it for more general purpose filtering (e.g. give me objects modified before cp X), then this will not give correct results

crates/iota-indexer/src/models/objects.rs
tomxey: I think your suggestion makes things cleaner, done in d66b9c396a

crates/iota-indexer/src/ingestion/primary/persist.rs
tomxey: done in d66b9c396a

crates/iota-indexer/src/read.rs
```diff
@@ -1079,7 +1079,9 @@ impl IndexerReader {
 
         let query = format!(
             "SELECT COUNT(*) as count FROM objects \
-             WHERE (object_id, object_version) IN (VALUES {})",
+             WHERE (object_id, object_version) IN (VALUES {}) \
+             AND (finalized_in_cp IS NULL \
+                  OR finalized_in_cp <= (SELECT COALESCE(MAX(sequence_number), -1) FROM checkpoints))",
```
kodemartin: You can use a CTE here to evaluate the` max_checkpoint` once for all rows. Something like

```sql
WITH max_chk AS (
    SELECT COALESCE(MAX(sequence_number), -1) AS max_sn FROM checkpoints
)
```
and then use it as a scalar`OR finalized_in_cp <= (SELECT max_sn FROM max_chk)`

kodemartin (approved):
Looks good with a minor suggestion

crates/iota-indexer/src/read.rs
tomxey: done [413ba2f](https://github.com/iotaledger/iota/pull/10753/commits/413ba2fb669e9fbf8c8df18727d41963993529c6)

sergiupopescu199 (approved):
lgtm 🦫

PR #10765, opened 2026-03-18
feat(apps): Versioned feature flags in apps-backend

### LINKED ISSUE #50: Implement versioned feature flags standard
We need a way to be able to enable features in the wallet (and any other app with numeric versioning) based on version number. We should define a standard knowing our numeric versions in the wallet are `x.y.z` for production and `x.y.z.r` for RCs.

It should be additive only, so that we dont break wallet versions in production

marc2332:
Closes https://github.com/iotaledger/ts-packages/issues/50


marc2332:
Closing for now

PR #10766, opened 2026-03-18
feat(explorer): improve amplitude antibot protection

begonaalvarezd:
## Summary

  The previous antibot measure relied solely on a time delay before flushing Amplitude events. Bots were surviving this window, causing their two autocaptured SDK events (`[Amplitude] Start Session` and `[Amplitude] Page Viewed`) to be submitted.

  ## Changes

  - **Increased detection delay** from 2s to 5s, giving more time for real user interactions to accumulate before the bot check runs
  - **Increased initial queue size** from 50 to 500, matching Amplitude's default and preventing premature flushes during the detection window
  - **Added session event counter** via an enrichment plugin that counts all non-special events (excluding `$identify`, `$groupidentify`, `revenue_amount`) as they
   pass through the Amplitude pipeline
  - **Added content-based bot check**: at the detection deadline, if the session has only 1 tracked event (the automatic `[Amplitude] Page Viewed`), it is treated
  as a bot — no flush occurs and the queued data is silently discarded. The existing `pagehide` beacon flush is also implicitly blocked since `isBotCleared`
  remains `false`

  ## How it works

  Every page load produces exactly 1 event visible to the counter: `[Amplitude] Page Viewed`. `[Amplitude] Start Session` fires during `ampli.load()` before the
  plugin is registered, so it is never counted. Any real user interaction — search, navigation, clicking a validator, etc. — pushes the count above the threshold
  of 1, allowing the session to flush normally.

  ## Trade-off

  Sessions where a user opens the explorer in a background tab and never interacts within the 5s window will be misclassified as bots and their data discarded.
  This is an accepted trade-off to reduce bot-generated noise in Amplitude.


begonaalvarezd (commented):
fancy, me like it

brancoder:
| # | Scenario | DOM signal fires? | `handler()` | `controller.abort()` | `enableFlushing()` | `isBotCleared` | `flushInterval` | `pagehide` | Final result |
|---|---|---|---|---|---|---|---|---|---|
| 1 | **True bot — leaves quickly** | ❌ never | never called | never called | never called | `false` | `null` | `controller.abort()` (no-op), `isBotCleared=false` → skip flush | ✅ Events discarded |
| 2 | **True bot — stays, never interacts** | ❌ never | never called | never called | never called | `false` | `null` | `controller.abort()` (no-op), `isBotCleared=false` → skip flush | ✅ Events discarded |
| 3 | **Human scrolls/moves at any point** | ✅ fires | called once | all 4 listeners removed atomically | `isBotCleared=false` → sets `true`, calls `ampli.flush()`, starts interval | `true` | running (1s flush) | `controller.abort()` (no-op), `isBotCleared=true` → `setTransport('beacon')` + `ampli.flush()` | ✅ Events flushed |
| 4 | **Human interacts, stays on page** | ✅ fires | called once | all 4 listeners removed | `enableFlushing()` runs | `true` | running — flushes every 1s for the rest of the session | tab still open, `pagehide` pending | ✅ Events flushed continuously |
| 5 | **Human interacts, then closes tab** | ✅ fires | called once | all 4 listeners removed | `enableFlushing()` runs | `true` | may or may not have ticked yet | `clearInterval(flushInterval)`, then beacon flush | ✅ Events flushed via beacon |
| 6 | **Human interacts before Amplitude fully loads** | ✅ may fire before `ampli.isLoaded` | called | controller aborted | `isBotCleared=true`, `ampli.flush()` called | `true` | interval starts but `if (ampli.isLoaded)` guards each tick | beacon flush | ✅ Safe — guarded |
| 7 | **Truly passive user** — no scroll, no mouse, no key, no touch | ❌ never | never called | never called | never called | `false` | `null` | `isBotCleared=false` → skip flush | ❌ Events discarded — indistinguishable from a bot |

apps/explorer/src/lib/utils/analytics/amplitude.ts
```diff
@@ -69,44 +67,57 @@ export async function initAmplitude() {
     setupAntiBotProtection();
 }
 
+const HUMAN_SIGNAL_EVENTS = ['scroll', 'mousemove', 'keydown', 'touchstart'] as const;
```
panteleymonchuk, line 70: I'd add 'click' here too

panteleymonchuk (approved):

brancoder (approved):
LGTM

PR #10768, opened 2026-03-18
feat(release): allow homebrew formula update when Windows build fails

thibault-martinez:
# Description of change

## Problem

The `update-homebrew-formula` job depends on `release-build`, a matrix job that builds binaries for Linux (x86_64, arm64), macOS (arm64), and Windows (x86_64). If the Windows build failed for any reason, GitHub would mark the entire `release-build` matrix as failed and block the homebrew job from running — even though homebrew only ships macOS and Linux binaries. A flaky Windows build would prevent a valid release from being published to homebrew.

## Fix

Added `continue-on-error: true` to the Windows matrix entry. GitHub Actions propagates this to the job level via `continue-on-error: ${{ matrix.continue-on-error == true }}`, which means a Windows failure is recorded (the leg shows as orange in the UI) but does not fail the overall matrix job. The Linux and macOS builds must still fully succeed for the homebrew update to proceed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,8 +82,10 @@ jobs:
             os_type: macos-arm64
           - os: windows-latest
             os_type: windows-x86_64
+            continue-on-error: true
       fail-fast: false
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.continue-on-error == true }}
     needs: generate-tag
 
     steps:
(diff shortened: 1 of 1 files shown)


alexsporn (approved):

.github/workflows/release.yml
```diff
@@ -258,7 +260,8 @@ jobs:
     name: Update the homebrew-tap formula
     needs: release-build
     runs-on: ubuntu-latest
-    # Releasing iota cli on Testnet and Mainnet releases which are equally stable
+    # Releasing iota cli on Testnet and Mainnet releases which are equally stable.
+    # Windows uses continue-on-error so its failure won't affect the matrix result.
```
Thoralf-M, line 264: Why don't just skip it for windows entirely?

thibault-martinez: Not sure what you mean but this is a different job, the `needs` clause currently requires the whole build matrix to succeed to trigger this job.

Thoralf-M: It sounded like it's trying to build something for Windows what we don't care about, but now I understand that it's just unrelated to this part :+1:

Thoralf-M (approved):

.github/workflows/release.yml
thibault-martinez: Windows still need to build but it shouldn't be a failure criteria to run the homebrew job

PR #10770, opened 2026-03-18
feat(ts-sdk): add support for effective commission rate

### LINKED ISSUE #10310: Update TS SDK after infra changes for IIP8
Update TS SDK after infra changes for IIP8
See https://github.com/iotaledger/iota/issues/10304

- [ ] Run code autogeneration
- [ ] Implement a graceful fallback similar to what we have already in this function https://github.com/iotaledger/iota/blob/fed560424e9c619dc2c5d3d247449669ebafa521/sdk/typescript/src/client/client.ts#L599 to fallback effective commission to the commission set by the validator for protocol versions < 20 (IIP gets enabled in v20)

Work against `feat/update-rpc-effective-commission-rate`

brancoder:
# Description of change

resolves #10310


sdk/typescript/src/client/client.ts
```diff
@@ -644,28 +644,44 @@ export class IotaClient {
...
-                  ...iotaSystemStateSummary.V1,
-                  committeeMembers: iotaSystemStateSummary.V1.activeValidators,
-                  safeModeComputationCharges: iotaSystemStateSummary.V1.safeModeComputationRewards,
-                  safeModeComputationChargesBurned:
-                      iotaSystemStateSummary.V1.safeModeComputationRewards,
-              };
+        if ('V2' in iotaSystemStateSummary) {
+            const validators = iotaSystemStateSummary.V2.activeValidators.map((v) => ({
+                ...v,
+                effectiveCommissionRate: isEffectiveCommissionRateSupported
```
msarcev: Not sure we need to do this.
Consider just (...) spreading and either effectiveCommissionRate is either there or not. 
No real reason to set `commissionRate = effectiveCommissionRate` right ?

msarcev: Switching from ternary to if/else also seams unncessary

msarcev (changes requested):

msarcev (approved):

marc2332 (approved):

PR #10775, opened 2026-03-18
feat(consensus): DAG visualizer server + frontend

muXxer:
# Description of change

This PR adds a feature-flag to starfish, which adds a gRPC server that exposes information about the DAG state in the consensus. This information can then stored in the `dag-visualizer-server` and exposed to a frontend with WebGL.

This makes it possible to visualize the consensus rounds for easier debugging.


dev-tools/dag-visualizer/frontend/package-lock.json
```diff
@@ -0,0 +1,2878 @@
+{
```
bingyanglin: Personally I would use `pnpm` instead of `npm`. `pnpm` is faster and used for other projects (see `pnpm-lock.yaml`) in our monorepo.

Thoralf-M (approved):
Approving file changes owned by dev-tools

miker83z:
Can the `/dev-tools/dag-visualizer/` go in its own repo? Just to not introduce new TS code to the monorepo 😅

dev-tools/dag-visualizer/frontend/package-lock.json
muXxer: Done in https://github.com/iotaledger/iota/pull/10775/commits/35cf22c1bd5c83deedeb089b0d4abbfaafb49235

muXxer:
> Can the `/dev-tools/dag-visualizer/` go in its own repo? Just to not introduce new TS code to the monorepo 😅

We could, but then this whole setup gets more complicated. It's not that this typescript code is running tests or included in CI, so I am not sure this is even a problem ;)

It's just a dev frontend, nothing exploitable.

bingyanglin (approved):

DaughterOfMars:
> > Can the `/dev-tools/dag-visualizer/` go in its own repo? Just to not introduce new TS code to the monorepo 😅
> 
> We could, but then this whole setup gets more complicated. It's not that this typescript code is running tests or included in CI, so I am not sure this is even a problem ;)
> 
> It's just a dev frontend, nothing exploitable.

It does feel like a separate project. Not sure why we would bloat the monorepo even more with it.

crates/starfish/core/src/dag_state.rs
```diff
@@ -531,6 +659,11 @@ impl DagState {
             .accepted_block_headers
             .with_label_values(&[author_label])
             .inc();
+
+        #[cfg(feature = "dag-visualizer")]
+        if let Some(event) = viz_event {
+            self.emit_dag_visualizer_event(event);
+        }
```
polinikita: I would include this part to the above to leave them logically adjacent

crates/starfish/core/src/dag_state.rs
```diff
@@ -518,6 +602,50 @@ impl DagState {
...
+                ),
+                acknowledgments: Arc::from(
+                    block_header
+                        .acknowledgments()
+                        .iter()
+                        .map(|r| {
+                            (
+                                r.round,
+                                r.author.value() as u8,
+                                r.digest.as_ref().try_into().unwrap_or([0u8; 32]),
```
polinikita: This part is very verbose - can we make a method for this? Another idea is to send a cloned Arc to receiver and make such a parsing there - moreover you make another parsing from DagVisualizerEvent to DagEvent at the receiver side, which looks as unnecessary code. I would create a enum with events DagVisualizerEvent with wrappers for BlockHeader, committed slot and skipped slot and do one parsing at the reception.

polinikita:
1. There are many inline imports - while they might be useful in some contexts (e.g. Sender to know how it works), I don't find those helpful in many cases (e.g. crate::dag_visualizer::grpc_streamer::DagVisualizerEvent) when the reader can understand the uniqueness.
2. Could you explain please - what happens if the channel between core and dag_visualizer is in lag and we don't update the dag in the visualizer with some blocks? - would not it panic if some of the ancestors/acknowledgments are missing or handle gracefully?

muXxer:
> * There are many inline imports - while they might be useful in some contexts (e.g. Sender to know how it works), I don't find those helpful in many cases (e.g. crate::dag_visualizer::grpc_streamer::DagVisualizerEvent) when the reader can understand the uniqueness.

Fixed.

> * Could you explain please - what happens if the channel between core and dag_visualizer is in lag and we don't update the dag in the visualizer with some blocks? - would not it panic if some of the ancestors/acknowledgments are missing or handle gracefully?

The broadcast channel between core and the gRPC streamer handles lag gracefully. `broadcast::RecvError::Lagged(n)` is caught and logged as a warning, then the stream continues with the next available event (no panic). On the frontend side, if ancestors/acknowledgments reference blocks that were never received (due to dropped events), they render as dangling edge. The block grid simply doesn't show the missing block, but there's no crash or assertion. The DAG visualizer is a best-effort debugging tool; losing some events during high load is acceptable. If needed, the standalone server can always re-fetch the full DAG window via the REST `/api/v1/dag` endpoint, which reads from persistent storage and is independent of the streaming channel.

muXxer:
Moved the typescript frontend and the server now to it's own repository.

polinikita (approved):

PR #10779, opened 2026-03-18
feat(consensus, iota-core, iota-type): Refactoring of the Scorer module

oliviasaa:
# Description of change

The current implementation of the Scorer module includes the definition of a `VersionedScoringMetrics` enum in `iota-types`, used in `consensus` and `iota-core` as a wrapper around a `ScoringMetrics` type. `ScoringMetricsV1` is a struct with named fields, which makes any change in the misbehaviors tracked an extensive and manual modification. One of the goals of this PR is to replace named-field structs with Vec-based types, so adding or removing tracked misbehaviors doesn't require manual changes across multiple crates.
Furthermore, the Scorer currently handles several independent responsibilities. We break the scorer into separate independent components with more clear roles. 
Finally, we remove the shared `Arc<VersionedScoringMetrics>` between consensus and the Scorer, and instead, the metrics collected by the consensus module are exposed to iota-core via the `ConsensusOutputAPI`. This change makes most of the Atomics currently used unnecessary.

Changes in the types used:

  - Deletes `VersionedScoringMetrics` and `ScoringMetricsV1` from `iota-types` and removes the re-export from `iota-common`. Those types are moved as-is to the consensus crate (renamed as `MysticetiVersionedScoringMetrics` and `MysticetiScoringMetricsV1`) , keeping consensus-internal changes minimal given the upcoming move to Starfish.
  - Introduces `MisbehaviorCounts` (a wrapper around `Vec<Vec<u64>>`) as the domain type for misbehavior data within `iota-core`, replacing direct use of `VersionedMisbehaviorReport` for internal operations. Conversion to/from the wire format (`VersionedMisbehaviorReport`) happens at system boundaries.      

 Breaking the Scorer into several components:

  - Decouples the `Scorer` into three focused components: `MisbehaviorMonitor` (tracks local misbehavior observations from consensus output), `ReportAggregator` (validates and merges incoming peer reports), and `Scorer` (pure score computation engine that reads from the aggregator)                                                                                                                       
  - Introduces `MisbehaviorConfig` as a single source of truth for which misbehavior categories are tracked, loaded from `ProtocolConfig`. All three components (`MisbehaviorMonitor`, `ReportAggregator`, `Scorer`) derive    
  their schema from it                                               
 
  

crates/iota-core/src/authority/misbehavior_config.rs
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    }
+
+    pub(crate) fn get_value(&self, metric_index: usize, authority: usize) -> u64 {
+        self.0
+            .get(metric_index)
+            .and_then(|metric_counts| metric_counts.get(authority))
+            .copied()
+            .expect("Invalid metric index {metric_index} or authority index {authority} for misbehavior counts")
```
bingyanglin: In this way, if this panics you'll see the curly braces instead of actual values. Maybe use `unwrap_or_else(|| panic!("Invalid metric index {metric_index} ..."))` so the values show up in the message.

crates/iota-core/src/authority/misbehavior_config.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Converts the local counts into the wire/storage representation that is
+    /// broadcast to peers as a `MisbehaviorReport` transaction.
+    pub fn to_report(&self, version: Version) -> VersionedMisbehaviorReport {
+        match version {
+            Version::V1 => {
+                let payload = LegacyReportPayload {
+                    faulty_blocks_provable: self.0[0].clone(),
+                    faulty_blocks_unprovable: self.0[1].clone(),
+                    missing_proposals: self.0[2].clone(),
+                    equivocations: self.0[3].clone(),
```
bingyanglin: Might be risky if in the future someone reorders the metrics? Maybe use named constants or derive the indices from ReportedMisbehaviors so they stay in sync?

crates/iota-core/src/authority/report_aggregator.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    #[cfg(test)]
+    pub fn to_serializable(&self) -> DBReceivedReportsStatePerAuthority {
+        DBReceivedReportsStatePerAuthority {
+            received_metrics: self.received_metrics_snapshot(),
+            invalid_reports_count: self.invalid_reports_count_snapshot(),
+        }
+    }
+
+    #[expect(dead_code)]
```
bingyanglin: Wondering why we need a new but `dead_code` code?

crates/iota-core/src/authority/report_aggregator.rs
```diff
@@ -0,0 +1,347 @@
...
+            report_version,
+            received_reports_state,
+        }
+    }
+
+    /// Validates an incoming report: checks that the report version matches the
+    /// expected version, and that the payload structure is correct for the
+    /// committee size (derived from internal state, so caller and aggregator
+    /// can't disagree).
+    pub(crate) fn validate_report(
```
bingyanglin, line 65: This returns a `Result` that callers could accidentally ignore. Maybe add `#[must_use]` so the compiler catches that?

piotrm50: `Result<T, E>` already has `#[must_use]` built into the type itself. Adding #[must_use] there is redundant — the compiler already warns if you ignore a `Result`.

crates/iota-core/src/authority/report_aggregator.rs
piotrm50: removed the dead code. It was scaffolding for future PR but it can be implemented once we need it.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -255,84 +250,109 @@ impl ConsensusTransactionKind {
...
     /// Returns the digest of the misbehavior report, caching it if it has not
     /// been computed yet.
     pub fn digest(&self) -> &MisbehaviorReportDigest {
-        match self {
-            VersionedMisbehaviorReport::V1(_, digest) => {
-                digest.get_or_init(|| MisbehaviorReportDigest::new(default_hash(self)))
-            }
-        }
+        self.digest
+            .get_or_init(|| MisbehaviorReportDigest::new(default_hash(self)))
```
Thoralf-M, line 305: This caching is actually fine? If some of the fields were mutated after the first time it was called it would be wrong

VorobyevIlya (approved):

crates/iota-types/src/messages_consensus.rs
```diff
@@ -255,84 +250,109 @@ impl ConsensusTransactionKind {
...
+    ) -> Self {
+        Self {
+            authority,
+            payload: MisbehaviorObservations::V1(observations),
+            generation,
+            digest: OnceCell::new(),
+        }
     }
 
     pub fn verify(&self, committee_size: usize) -> bool {
```
roman1e2f5p8s: this was previously used, but not anymore after the refactor - just checking if we still want to keep it or remove it?

crates/iota-core/src/consensus_handler.rs
```diff
@@ -253,6 +253,10 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         // last_consensus_stats, so that it won't be re-executed in the future.
         self.last_consensus_stats.index = execution_index;
 
+        self.epoch_store
+            .misbehavior_monitor
+            .update_from_consensus_output(consensus_output.misbehavior_counts());
+
```
roman1e2f5p8s:
```suggestion
        if self
            .epoch_store
            .protocol_config()
            .calculate_validator_scores()
        {
            self.epoch_store
                .misbehavior_monitor
                .update_from_consensus_output(consensus_output.misbehavior_counts());
        }
```
I think this should be feature-gated (at least until finally enabled on mainnet) to avoid unnecessary allocations

crates/iota-types/src/messages_consensus.rs
roman1e2f5p8s: side question to this PR: why not `||` here?

roman1e2f5p8s (approved):

bingyanglin (approved):

crates/iota-core/src/authority/misbehavior_monitor.rs
```diff
@@ -0,0 +1,119 @@
...
+    committee_size: usize,
+    // The current metrics counts collected by the authority, i.e., the local view of the node
+    // about the behaviour of the rest of the committee, according to the blocks received.
+    current_local_observations: ArcSwap<MisbehaviorObservations>,
+    // Single-writer: the three rate-limit fields below are only mutated by
+    // SubmitCheckpointToConsensus::checkpoint_created. Don't add additional writers without
+    // revisiting the atomicity story — `last_report_summary` and `last_report_checkpoint_seq`
+    // form a logical tuple but are stored as independent Relaxed atomics, safe today only
+    // because reads and writes happen from the single CheckpointBuilder task.
+    // `has_sent_end_of_epoch_report` is an independent epoch-once flag.
```
cyberphysic4l: Why don't we wrap these three connected atomics in a single struct with a mutex around it to make sure it's safe if any other writers are created rather than just having this comment?

crates/iota-core/src/authority/scorer.rs
```diff
@@ -1,773 +1,657 @@
...
+/// owns its parameters and the math for the matching `MisbehaviorObservations`
+/// version. `Scoreboard` delegates here and stores the result.
+///
+/// Per-version variants are tied to the corresponding observations variant:
+/// `V1(ScorerV1)` consumes only `MisbehaviorObservationsV1`. The unwrap from
+/// the `MisbehaviorObservations` enum happens here at the dispatch boundary,
+/// so per-version scorer impls take typed inputs and never re-match.
+enum VersionedScorer {
+    V1(ScorerV1),
+}
```
cyberphysic4l, line 72: Now that so much has been removed from this and it only deals with calculating scores, how about calling it `VersionedScoreCalculator` and `ScoreCalculatorV1`?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -666,9 +676,11 @@ pub struct AuthorityPerEpochStore {
     randomness_manager: OnceCell<tokio::sync::Mutex<RandomnessManager>>,
     randomness_reporter: OnceCell<RandomnessReporter>,
 
-    /// Component including the local view about the other authorities'
-    /// misbehavior metrics, and received reports.
-    pub(crate) scorer: Arc<Scorer>,
+    pub(crate) misbehavior_monitor: MisbehaviorMonitor,
```
cyberphysic4l, line 680:
```suggestion
    /// Monitors local observations of misbehaviors and sends reports.
    pub(crate) misbehavior_monitor: MisbehaviorMonitor,
```

crates/iota-core/src/checkpoints/checkpoint_output.rs
```diff
@@ -147,37 +147,38 @@ impl<T: SubmitToConsensus + ReconfigurationInitiator> CheckpointOutput
...
                     .store_last_report_summary(new_report_summary);
                 epoch_store
-                    .scorer
+                    .misbehavior_monitor
                     .store_last_report_checkpoint_seq(checkpoint_seq);
                 if should_send_last_report {
-                    epoch_store.scorer.mark_end_of_epoch_report_sent();
+                    epoch_store
+                        .misbehavior_monitor
+                        .mark_end_of_epoch_report_sent();
```
cyberphysic4l: Perhaps wrap the three values in a single mutex instead of three separate atomics? See same comment in misbehavior monitor.

crates/iota-types/src/messages_consensus.rs
piotrm50: The struct is created locally and broadcast to other validators over the wire — recipients deserialize and consume it read-only. No code path mutates fields after construction. Since the inputs to the digest can never change, caching it in a `OnceCell` on first access is safe.

Tbh I wanted to get rid of this digest altogether because  (authority,generation) pair is sufficient, but it's already used on the testnet so removing it would not be compatible.

crates/iota-types/src/messages_consensus.rs
piotrm50: removed it.

crates/iota-core/src/consensus_handler.rs
piotrm50: applied

crates/iota-types/src/messages_consensus.rs
piotrm50: replaced with `||` - not sure why a single pipe was originally used.

crates/iota-core/src/authority/scorer.rs
piotrm50: I'd keep Scorer / VersionedScorer as-is — it's short, the module is already called scorer.rs, and Scoreboard holding a VersionedScorer reads naturally. ScoreCalculator is more explicit but doesn't really resolve any ambiguity, just adds verbosity.

crates/iota-core/src/authority/misbehavior_monitor.rs
piotrm50: applied

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: added

crates/iota-core/src/checkpoints/checkpoint_output.rs
piotrm50: added a new mutex'ed struct

Thoralf-M (approved):
Approving the changes in my owned files

cyberphysic4l (approved):

PR #10780, opened 2026-03-19
upstream(core): Enable iterator methods for in_memory_db

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@c232615](https://github.com/MystenLabs/sui/commit/c232615066db1ca778acf5b46e6cc3e91abd5dbe)
- Description:

PR enables iterator methods for in_memory_db.

Note: tidehunter-related changes were skipped as that backend is not present.

Part of #10609

### DIFF
--- a/crates/typed-store/src/database.rs
+++ b/crates/typed-store/src/database.rs
@@ -22,15 +22,17 @@ use tracing::{debug, error, instrument, warn};
 use typed_store_error::TypedStoreError;
 
 use crate::{
+    DbIterator,
     memstore::{InMemoryBatch, InMemoryDB},
     metrics::{DBMetrics, RocksDBPerfContext, SamplingInterval},
     rocks::{
-        RocksDB, be_fix_int_ser,
+        RocksDB,
         errors::{typed_store_err_from_bcs_err, typed_store_err_from_rocks_err},
-        rocks_cf,
+        rocks_cf, rocks_util,
         safe_iter::{SafeIter, SafeRevIter},
     },
     traits::{Map, TableSummary},
+    util::{be_fix_int_ser, iterator_bounds, iterator_bounds_with_range},
 };
 
 #[derive(Clone)]
@@ -105,8 +107,6 @@ pub enum StorageWriteBatch {
     Rocks(rocksdb::WriteBatch),
     InMemory(InMemoryBatch),
 }
-
-pub type RawIter<'a> = rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
 #[derive(Debug, Default)]
 pub struct MetricConf {
@@ -340,19 +340,6 @@ impl Database {
 
         #[allow(clippy::let_and_return)]
         ret
-    }
-
-    pub(crate) fn raw_iterator_cf<'a: 'b, 'b>(
-        &'a self,
-        cf: &ColumnFamily,
-        readopts: ReadOptions,
-    ) -> RawIter<'b> {
-        match &self.storage {
-            Storage::Rocks(rocksdb) => rocksdb
-                .underlying
-                .raw_iterator_cf_opt(&rocks_cf(rocksdb, cf.name()), readopts),
-            _ => unimplemented!("iterators not yet supported for other backends"),
-        }
     }
 
     pub(crate) fn compact_range_cf<K: AsRef<[u8]>>(
@@ -597,8 +584,8 @@ impl<K, V> DBMap<K, V> {
     }
 
     pub fn compact_range<J: Serialize>(&self, start: &J, end: &J) -> Result<(), TypedStoreError> {
-        let from_buf = be_fix_int_ser(start)?;
-        let to_buf = be_fix_int_ser(end)?;
+        let from_buf = be_fix_int_ser(start);
+        let to_buf = be_fix_int_ser(end);
         self.db
             .compact_range_cf(&self.column_family, Some(from_buf), Some(to_buf));
         Ok(())
@@ -638,13 +625,10 @@ impl<K, V> DBMap<K, V> {
         } else {
             None
         };
-        let keys_bytes: Result<Vec<_>, TypedStoreError> = keys
-            .into_iter()
-            .map(|k| be_fix_int_ser(k.borrow()))
-            .collect();
+        let keys_bytes = keys.into_iter().map(|k| be_fix_int_ser(k.borrow()));
         let results: Result<Vec<_>, TypedStoreError> = self
             .db
-            .multi_get(&self.column_family, keys_bytes?, &self.opts.readopts())
+            .multi_get(&self.column_family, keys_bytes, &self.opts.readopts())
             .into_iter()
             .collect();
         let entries = results?;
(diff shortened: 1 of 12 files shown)


crates/typed-store/src/util.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+        Bound::Unbounded => None,
+    };
+    let iterator_upper_bound = match range.end_bound() {
+        Bound::Included(upper_bound) => {
+            let mut key_buf = be_fix_int_ser(&upper_bound);
+
+            // If the key is already at the limit, there's nowhere else to go, so no upper
+            // bound
+            if !is_max(&key_buf) {
```
bingyanglin, line 61: The change is the same as upstream. However, the comment says "no upper bound" but we still return `Some(key_buf)` on the next line. RocksDB upper bounds are exclusive, so when `is_max` is `true`, this sets an exclusive upper bound at the max key — which excludes it. Maybe split the Some into the if branch so `is_max` returns None? Like 

```rust
if !is_max(&key_buf) {
    big_endian_saturating_add_one(&mut key_buf);
    Some(key_buf)
} else {
    None
}
```

jkrvivian: fixed in https://github.com/iotaledger/iota/pull/10780/commits/e975486f21f3e093f613754841f6f6ac4ea2a301

crates/typed-store/src/memstore.rs
```diff
@@ -101,4 +105,38 @@ impl InMemoryDB {
...
+        reverse: bool,
+    ) -> Box<dyn Iterator<Item = Result<(K, V), TypedStoreError>> + '_>
+    where
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let config = bincode::DefaultOptions::new()
+            .with_big_endian()
+            .with_fixint_encoding();
+        let lower_bound = lower_bound.map(Bound::Included).unwrap_or(Bound::Unbounded);
```
bingyanglin, line 123: `InMemoryDB::iterator` wraps the upper bound as `Bound::Included` but RocksDB's `set_iterate_upper_bound` is exclusive. So the same `safe_iter_with_bounds` call returns different results depending on the backend. We might need to correct this?

crates/typed-store/src/memstore.rs
```diff
@@ -101,4 +105,38 @@ impl InMemoryDB {
...
+            .get(cf_name)
+            .unwrap_or(&BTreeMap::new())
+            .range((lower_bound, upper_bound))
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+        if reverse {
+            section.reverse();
+        }
+        Box::new(section.into_iter().map(move |(raw_key, raw_value)| {
+            let key = config.deserialize(&raw_key).unwrap();
```
bingyanglin: The return type is `Result<(K, V), TypedStoreError>`, may be we just map errors to `Err(TypedStoreError::Deserialization(...))` instead of panicking?

crates/typed-store/src/util.rs
```diff
@@ -0,0 +1,130 @@
...
+    range: impl RangeBounds<K>,
+) -> (Option<Vec<u8>>, Option<Vec<u8>>)
+where
+    K: Serialize,
+{
+    let iterator_lower_bound = match range.start_bound() {
+        Bound::Included(lower_bound) => {
+            // Rocksdb lower bound is inclusive by default so nothing to do
+            Some(be_fix_int_ser(&lower_bound))
+            // readopts.set_iterate_lower_bound(key_buf);
```
bingyanglin: We might not need this? I checked the latest `mainnet` of upstream and looks like this is still dead code.

bingyanglin (approved):

crates/typed-store/src/util.rs
semenov-vladyslav: This issue is so bad. The fix that Bing proposed preserves the old logic. The upstream change breaks it. But the question here is: "what should be the correct logic"? If we need to correctly set an inclusive upper bound we need to increment it (since rocksdb iterator upper bound is always exclusive). This increment logic is implemented in big_endian_saturating_add_one, but it is broken. Eg. on input [0xff; 32] it will just keep this value as is; and it that key happens to be present in the DB it will be omitted. This is incorrect buggy behaviour. I don't get what prevented the devs from correctly incrementing key in big_endian_saturating_add_one (if is_max(key) is true, then just append 0 to the key; for [0xff; 32] it should be [0xff, .. 0xff, 0] -- 33 bytes in total). If we return None here we remove the upper bound entirely. The only scenario in which it doesn't matter if the DB does not contain keys larger than [0xff; key_buf.len()] (Move VM side seem to be pretty careful to preserve this invariant), and hopefully this is indeed the case.

We have the following options:
1. stick to the old incorrect behavior (the current code);
2. stick to the upstream slightly corrected code
```rs
            if !is_max(&key_buf) {
                // Since we want exclusive, we need to increment the key to get the upper bound
                big_endian_saturating_add_one(&mut key_buf);
                // readopts.set_iterate_upper_bound(key_buf);
            }
            Some(key_buf)
```
3. stick to a better variant of upstream code (big_endian_saturating_add_one already checks for is_max, no need to do it twice):
```rs
            // Since we want exclusive, we need to increment the key to get the upper bound
            big_endian_saturating_add_one(&mut key_buf);
            Some(key_buf)
```
4. fix the issue with the big_endian_saturating_add_one, return the proper upper bound in case is_max(key_buf) is true.

I would choose the case 4 since it fixes a possible (and hopefully unlikely) edge-case that can result in incorrect behavior.

nikitapolyanskii (approved):

polinikita (approved):

crates/typed-store/src/database.rs
```diff
@@ -1274,54 +1197,82 @@ where
...
-            self.cf_name().to_string(),
-            db_iter,
-            _timer,
-            _perf_ctx,
-            bytes_scanned,
-            keys_scanned,
-            Some(self.db_metrics.clone()),
-        )
+    ) -> DbIterator<'a, (K, V)> {
+        let (lower_bound, upper_bound) = iterator_bounds(lower_bound, upper_bound);
```
semenov-vladyslav, line 1227: This is another issue with the iterators. Suppose that the both bounds are provided. Since bounds are transparently passed to the rocksdb options, they will correspond to a semi-open range [lower..upper), or Included(lower), Excluded(upper). Reversing an iterator should return the same values but in the reverse order. However, the logic in reversed_safe_iter_with_bounds sets bounds as Included(lower), Included(upper) (if both bounds are provided). This range can't be identical to the range built in safe_iter_with_bounds.

Not including the upper_bound in safe_iter_with_bounds makes users create unrepresentable/invalid values of type K that serve the only purpose of being "upper bound" in certain sense.

semenov-vladyslav (approved):
Approving with a few important comments.

crates/typed-store/src/util.rs
jkrvivian: I applied Bing's suggestion and didn't see an issue with the example of key `[0xff; 32]`, None is returned for the upper bound, so the key `[0xff; 32]` should be included in the result. 👀

PR #10789, opened 2026-03-19
refactor(indexer): use historical storage as checkpoint source

### LINKED ISSUE #10723: feat(indexer): Ingest data from hybrid checkpoint source
With the ability of fullnodes to prune their internal database, it becomes problematic for the indexer to process historical checkpoints form a fullnode.

In this regard we need to refactor the iota-indexer to accept historical & live URLs in order to sync older checkpoints which might been pruned on the fullnode.

sergiupopescu199:
# Description of change

This PR extends the `iota-indexer` to support checkpoint ingestion from S3-compatible historical and live storages as an alternative to a fullnode gRPC endpoint. Since fullnodes can prune their internal databases, syncing the indexer from genesis via a fullnode gRPC URL may result in errors when historical checkpoints are no longer available.

To address this, we extended the `--remote-store-url` argument to accept either a fullnode gRPC URL or an S3-compatible historical checkpoint store URL. When provided, the indexer automatically determines the source type by first attempting a gRPC health check, then falling back to checking for a MANIFEST file on the object store. If neither probe succeeds after retrying with exponential backoff, the indexer returns an error and exits.

For minimal latency at the network tip when using historical storage, the user can optionally provide `--current-epoch-store-url`, which points to a live object store serving individual checkpoint files for the current epoch only.

fixes #10723 

- Ran the indexer to make sure that it actually was syncing from historical and live storage, used testnet data.
```shell
cargo r -- --database-url postgres://postgres:postgrespw@localhost:5432/iota_indexer  indexer --remote-store-url=https://checkpoints.testnet.iota.cafe/ingestion/historical  --current-epoch-store-url=https://checkpoints.testnet.iota.cafe/ingestion/live --reset-db
```

- Verified that `--current-epoch-store-url` requires `--remote-store-url` and cannot be used standalone
```shell
cargo r -- --database-url postgres://postgres:postgrespw@localhost:5432/iota_indexer  indexer --current-epoch-store-url=https://checkpoints.testnet.iota.cafe/ingestion/live  --reset-db

error: the following required arguments were not provided:
  --remote-store-url <REMOTE_STORE_URL>
```

### Release Notes

- [x] Indexer: Add the ability to sync checkpoints form historical & live storages as an alternative checkpoint source to fullnode.

### DIFF
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -176,8 +176,8 @@ pub struct IngestionSources {
     #[arg(long)]
     pub data_ingestion_path: Option<PathBuf>,
 
-    /// Primary remote checkpoint source. This flag accepts either a fullnode gRPC URL (e.g. http://0.0.0.0:50051) or an
-    /// S3-compatible object store URL hosting batched checkpoint files (e.g. https://checkpoints.mainnet.iota.cafe/ingestion/historical).
+    /// Primary remote checkpoint source. This flag accepts either a fullnode gRPC URL (e.g. `http://0.0.0.0:50051`) or an
+    /// S3-compatible object store URL hosting batched checkpoint files (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/historical`).
     /// When pointing to an object store, this provides complete checkpoint
     /// coverage from genesis. When pointing to a fullnode, checkpoint
     /// availability depends on the node's pruning configuration.
@@ -185,7 +185,7 @@ pub struct IngestionSources {
     pub remote_store_url: Option<Url>,
 
     /// Optional S3-compatible object store URL serving current epoch
-    /// checkpoints for low-latency ingestion at the network tip (e.g. https://checkpoints.mainnet.iota.cafe/ingestion/live).
+    /// checkpoints for low-latency ingestion at the network tip (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/live`).
     /// Use alongside `--remote-store-url` pointing to a historical store for
     /// complete coverage with minimal latency.
     #[arg(long, requires = "remote_store_url")]
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -176,20 +176,29 @@ pub struct IngestionSources {
     #[arg(long)]
     pub data_ingestion_path: Option<PathBuf>,
 
-    /// Primary remote checkpoint source. This flag accepts either a fullnode gRPC URL (e.g. `http://0.0.0.0:50051`) or an
-    /// S3-compatible object store URL hosting batched checkpoint files (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/historical`).
+    /// Primary remote checkpoint source.
+    ///
+    /// Accepts either a fullnode gRPC URL (e.g. `http://0.0.0.0:50051`) or an
+    /// S3-compatible object store URL hosting batched checkpoint files
+    /// (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/historical`).
+    ///
     /// When pointing to an object store, this provides complete checkpoint
     /// coverage from genesis. When pointing to a fullnode, checkpoint
     /// availability depends on the node's pruning configuration.
     #[arg(long)]
     pub remote_store_url: Option<Url>,
 
-    /// Optional S3-compatible object store URL serving current epoch
-    /// checkpoints for low-latency ingestion at the network tip (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/live`).
+    /// Optional live checkpoint store for low-latency ingestion at the network
+    /// tip.
+    ///
+    /// S3-compatible object store URL serving individual checkpoint files for
+    /// the current epoch only
+    /// (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/live`).
+    ///
     /// Use alongside `--remote-store-url` pointing to a historical store for
-    /// complete coverage with minimal latency.
+    /// complete coverage with minimal tip latency.
     #[arg(long, requires = "remote_store_url")]
-    pub current_epoch_store_url: Option<Url>,
+    pub live_checkpoints_store_url: Option<Url>,
 }
 
 #[derive(Args, Debug, Clone)]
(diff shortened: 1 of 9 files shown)


tomxey (approved):

crates/iota-indexer/src/ingestion/common/connection.rs
```diff
@@ -0,0 +1,93 @@
...
+) -> IndexerResult<Option<RemoteUrl>> {
+    let Some(url) = ingestion_sources
+        .remote_store_url
+        .as_ref()
+        .map(ToString::to_string)
+    else {
+        return Ok(None);
+    };
+
+    let grpc_client = GrpcClient::connect(url.clone()).await?;
```
kodemartin: This is a fallible expression: In what situations can it fail? Can it fail with a valid checkpoint store url?

crates/iota-indexer/src/ingestion/common/connection.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    // Use a lightweight S3 client to check if the MANIFEST file exists.
+    // We avoid HistoricalReader here as its internal manifest fetch retries
+    // with a 15-minute default backoff and does not have a timeout.
+    let historical =
+        create_remote_store_client(url.clone(), Default::default(), timeout.as_secs())?;
```
kodemartin: Same here: This is fallible. In what situations can it fail?

Can it be that a Fullnode::URL passes above and fails here?

crates/iota-indexer/src/ingestion/common/connection.rs
```diff
@@ -0,0 +1,93 @@
...
+                    duration.as_millis()
+                );
+                tokio::time::sleep(duration).await;
+            }
+            None => {
+                return Err(IndexerError::Generic(format!(
+                    "unable to resolve remote store URL after {timeout:?}: {url}"
+                )));
+            }
+        }
```
kodemartin: I have the following remarks on this block:
1. It will be more robust to create the clients on each iteration instead of creating them in lines 46 and 51 and coupling the failure possibilities. Creating clients should be cheap, and at any case this is a startup operation so we can tolerate the cost. 
2. Wouldn't be more concise and idiomatic if you used `backoff::future::retry` directly here as opposed to looping and `next_backoff + sleep`?
3. doesn't make sense to assume that `remote-store-url` is historical if `live` is `Some`?

crates/iota-indexer/src/config.rs
```diff
@@ -172,11 +172,24 @@ impl Default for HistoricFallbackOptions {
...
 pub struct IngestionSources {
+    /// Ingest checkpoints from the given path.
     #[arg(long)]
     pub data_ingestion_path: Option<PathBuf>,
 
+    /// Primary remote checkpoint source. This flag accepts either a fullnode gRPC URL (e.g. `http://0.0.0.0:50051`) or an
+    /// S3-compatible object store URL hosting batched checkpoint files (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/historical`).
+    /// When pointing to an object store, this provides complete checkpoint
+    /// coverage from genesis. When pointing to a fullnode, checkpoint
+    /// availability depends on the node's pruning configuration.
```
kodemartin: Consider splitting to make it more readable (summary + additional paragraphs)
```rust
    /// Primary remote checkpoint source.
    ///
    /// This flag accepts either a fullnode gRPC URL (e.g. `http://0.0.0.0:50051`) or an
    /// S3-compatible object store URL hosting batched checkpoint files (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/historical`).
    ///
    /// When pointing to an object store, this provides complete checkpoint
    /// coverage from genesis. When pointing to a fullnode, checkpoint
    /// availability depends on the node's pruning configuration.
```

crates/iota-indexer/src/config.rs
```diff
@@ -172,11 +172,24 @@ impl Default for HistoricFallbackOptions {
...
     #[arg(long)]
     pub remote_store_url: Option<Url>,
+
+    /// Optional S3-compatible object store URL serving current epoch
+    /// checkpoints for low-latency ingestion at the network tip (e.g. `https://checkpoints.mainnet.iota.cafe/ingestion/live`).
+    /// Use alongside `--remote-store-url` pointing to a historical store for
+    /// complete coverage with minimal latency.
```
kodemartin: Same here. Consider revising into summary + paragraph with details.

crates/iota-indexer/src/config.rs
```diff
@@ -172,11 +172,24 @@ impl Default for HistoricFallbackOptions {
...
+    #[arg(long, requires = "remote_store_url")]
+    pub current_epoch_store_url: Option<Url>,
```
kodemartin: Nit: `current_epoch_store_url` is a bit ambiguous as it hints me to a store that holds epoch information in general.

Maybe `live_checkpoints_store_url`?

crates/iota-indexer/src/ingestion/common/connection.rs
sergiupopescu199: `GrpcClient::connect` is lazy it only parses the URL into a valid URI, no network connection is made. It would only fail on a malformed URL, which would also fail for the S3 client. A valid checkpoint store URL will pass this step; the actual probe happens in the `get_health` call inside the loop.

fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

crates/iota-indexer/src/ingestion/common/connection.rs
sergiupopescu199: Since we pass empty options, it takes the HTTP builder path. `HttpBuilder::build()` only fails on malformed URLs. A valid fullnode URL like `http://localhost:50051` builds successfully, no network call is made at construction time very similar to `GrpcClient::connect`

fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

crates/iota-indexer/src/config.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

crates/iota-indexer/src/config.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

crates/iota-indexer/src/config.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

crates/iota-indexer/src/ingestion/common/connection.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/10789/commits/8825e56997364547bee5956148b4ce075f014002

kodemartin (approved):
lgtm

PR #10792, opened 2026-03-19
upstream(core): Provide the tls_single_cert convenience method

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@3933733](https://github.com/MystenLabs/sui/commit/39337332ac84ef76f5740d641840f725e9d6fc9b)
- Description:

Provide the `tls_single_cert` convenience method for instantiating a TLS
configuration by loading PEM formatted files for the certificate chain
and private key material.

Part of #10609

### DIFF
--- a/crates/iota-http/src/lib.rs
+++ b/crates/iota-http/src/lib.rs
@@ -10,7 +10,7 @@ use http::{Request, Response};
 use hyper_util::service::TowerToHyperService;
 use io::ServerIo;
 use tokio::task::JoinSet;
-use tokio_rustls::TlsAcceptor;
+use tokio_rustls::{TlsAcceptor, rustls};
 use tower::{Service, ServiceBuilder, ServiceExt};
 use tracing::trace;
 
@@ -37,7 +37,7 @@ const ALPN_H1: &[u8] = b"http/1.1";
 #[derive(Default)]
 pub struct Builder {
     config: Config,
-    tls_config: Option<tokio_rustls::rustls::ServerConfig>,
+    tls_config: Option<rustls::ServerConfig>,
 }
 
 impl Builder {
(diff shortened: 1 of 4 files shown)


semenov-vladyslav (approved):
The change looks good. But we have iota-tls to deal with tls config. And we have `iota_config::node::TlsConfig` that already warps `cert` and `key`. So, maybe this change should belong to some other place?

bingyanglin (approved):

crates/iota-http/src/lib.rs
```diff
@@ -50,7 +50,27 @@ impl Builder {
         self
     }
 
-    pub fn tls_config(mut self, tls_config: tokio_rustls::rustls::ServerConfig) -> Self {
+    // Convenience method for configuring TLS with a single server cert
+    //
+    // Attempts to load PEM formatted files for the certificate chain and private
+    // key material from the provided file system paths.
+    pub fn tls_single_cert(
```
bingyanglin, line 57: This method has no callers anywhere in the codebase, so it's dead code right now. But more importantly, we already have `iota_config::node::TlsConfig` (exactly cert + key file paths) and `iota-tls` is where TLS config construction lives, and `create_rustls_server_config` is already there. Maybe add a function to `iota-tls` that takes a `&TlsConfig` and returns a `rustls::ServerConfig`, and call it via `Builder::tls_config()`? There's also a crypto provider inconsistency: `iota-tls` explicitly passes `rustls::crypto::ring::default_provider()`, but this method just calls `rustls::ServerConfig::builder()` with whatever the default is.

jkrvivian: I moved the TLS construction logic into iota-tls `create_rustls_server_config_from_pem()`, and use it in `tls_single_cert`. But taking `&TlsConfig` directly would require iota-tls to depend on iota-config, which is a heavy crate (pulls in consensus-config, iota-types, etc.). Using impl AsRef<Path> keeps iota-tls lightweight, and callers can simply pass `tls_config.cert() / tls_config.key()`. 


Also, this method will be used in the next upstream commit https://github.com/MystenLabs/sui/commit/0316d747db35c796c5b3de099dfcedcafb0bb87d .

bingyanglin (approved):

PR #10810, opened 2026-03-20
feat(iota-types,iota-network,iota-core): use protobuf wire format for new validator gRPC endpoints

piotrm50:
# Description of change

Switch the 3 new validator endpoints (`submit_transaction`, `wait_for_effects`, `validator_health`) from BCS to native Protobuf (prost) wire format, following Sui's dual-codec approach.

**Key changes:**
- Add `Raw*` prost message types that wrap BCS-serialized inner data as `Bytes` fields
- Bidirectional `TryFrom` conversions between native types and `Raw*` prost types
- Refactor `build.rs` to use `MethodDef` struct with `use_prost` flag — legacy endpoints stay on `BcsCodec`, new endpoints use `ProstCodec`
- Add handler stubs in `authority_server.rs` with ping request handling
- Add native types (`ExecutedData`, `SubmitTxRequest/Response`, `WaitForEffectsRequest/Response`, `ValidatorHealthRequest/Response`) needed by the white-flag flow

## Links to any relevant issues

Relates to #10741

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

6 round-trip unit tests for all `Raw*` ↔ native type conversions. All existing server tests continue to pass. Clippy clean.

PR #10814, opened 2026-03-20
feat(iota-core,iota-types): [P-COOL] batch response and multi-transaction wait-for-effects

### LINKED ISSUE #10719: pcool: Remove soft-bundle-specific restrictions from PCOOL batch path
The current \`handle_submit_transactions_impl\` carries over validation constraints originally designed for the cert-flow soft-bundle endpoint:
- All transactions must reference shared objects
- All transactions must have the same gas price

These restrictions do not apply to the general PCOOL batch endpoint, which is a pure network optimization. Remove them from the PCOOL path.

When the request contains more than 1 UserTransactions they should be handled individually like in a single-transaction flow. 

> **Note:** the cert-flow \`handle_soft_bundle_certificates_v1\` endpoint is separate and must **not** be changed.

**Files:** \`authority_server.rs\`, \`messages_grpc.rs\`

This goes in an opposite direction from #10368, but that's ok because the decision to remove this was done already after it was implemented along with other improvements.

### LINKED ISSUE #10748: pcool: Batch response + multi-transaction wait-for-effects
Two related changes to the PCOOL batch API. Batch submission is a **network optimization only** — no ordering or atomicity guarantees between transactions in a batch.

**1. Response shape:** change `SubmitTransactionsResponse` from a single `TransactionSubmissionResult` to `Vec<TransactionSubmissionResult>` — one entry per submitted transaction in submission order. Each entry independently reports its result (submitted, already executed, or error).

**2. Wait-for-effects:** extend `WaitForEffectsRequest` to accept `Vec<TransactionDigest>` and `WaitForEffectsResponse` to return `Vec<WaitForEffectsResult>`, one per digest. The server waits for all requested effects in parallel.

Depends on: #10719, #10746

**Files:** `messages_grpc.rs`, `authority_server.rs`, `build.rs` (iota-network)

filipdulic:
# Description of change

  Changed `SubmitTransactionsResponse` to return `Vec<SubmitTransactionResult>` and
  `WaitForEffectsResponse` to return `Vec<WaitForEffectResponse>`, one entry per transaction.
  Both handlers process requests in parallel via `join_all` through dedicated single-item
  methods. Removed soft-bundle code as part of deprecation.

 ## Note
 PR #10810 was merged into this PR, it defines protobuf wrappers for validator wire format.

  closes #10748, #10719 

### DIFF
--- a/crates/iota-core/src/authority_server.rs
+++ b/crates/iota-core/src/authority_server.rs
@@ -22,7 +22,6 @@ use iota_network::{
 };
 use iota_network_stack::server::IOTA_TLS_SERVER_NAME;
 use iota_types::{
-    base_types::TransactionEffectsDigest,
     effects::{TransactionEffects, TransactionEffectsAPI},
     error::*,
     fp_ensure,
@@ -1299,26 +1298,20 @@ impl ValidatorService {
 
         let request = request.into_inner();
 
-        // Handle ping.
+        // Handle ping (empty request → empty response).
         if request.is_ping() {
-            // TODO: handle ping response better. Do we even need ping requests?
             return Ok((
-                tonic::Response::new(SubmitTransactionsResponse {
-                    results: vec![SubmitTransactionResult::Submitted],
-                }),
+                tonic::Response::new(SubmitTransactionsResponse { results: vec![] }),
                 Weight::zero(),
             ));
         }
 
         let transactions = request.transactions;
 
-        // Per-tx validity checks. For soft bundles this also accumulates total
-        // serialized size for the bundle size limit.
-        // TODO: check if size check needed?
-        let mut _total_size_bytes: u64 = 0;
+        // Per-tx validity and overload checks. These are validator-wide
+        // conditions so we short-circuit before spawning parallel futures.
         for tx in &transactions {
-            _total_size_bytes +=
-                tx.validity_check(epoch_store.protocol_config(), epoch_store.epoch())? as u64;
+            tx.validity_check(epoch_store.protocol_config(), epoch_store.epoch())?;
 
             // System overload check per transaction: check_execution_overload examines
             // per-transaction input objects (shared objects, object queue depth).
@@ -1339,9 +1332,7 @@ impl ValidatorService {
         // where the timer also starts after the overload check.
         let _handle_tx_metrics_guard = metrics.handle_transaction_latency.start_timer();
 
-        // ── Single-transaction path ─────────────────────────────────────────
-        // TODO fix above
-        // NOTO: executed in parallel
+        // Process each transaction independently in parallel.
         let futures = transactions
             .into_iter()
             .map(|tx| self.handle_submit_transaction_impl(tx, &epoch_store));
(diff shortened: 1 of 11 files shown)


crates/iota-core/src/authority_server.rs
```diff
@@ -1404,28 +1298,20 @@ impl ValidatorService {
...
 
-        // Per-tx validity checks. For soft bundles this also accumulates total
-        // serialized size for the bundle size limit.
-        let mut total_size_bytes: u64 = 0;
+        // Per-tx validity and overload checks. These are validator-wide
+        // conditions so we short-circuit before spawning parallel futures.
         for tx in &transactions {
-            total_size_bytes +=
-                tx.validity_check(epoch_store.protocol_config(), epoch_store.epoch())? as u64;
+            tx.validity_check(epoch_store.protocol_config(), epoch_store.epoch())?;
```
filipdulic: @iotaledger/consensus  should we error out on a single tx validity fail, or return rejected?

crates/iota-core/src/unit_tests/server_tests.rs
bingyanglin: The test name and docstring still describe the old rejection behavior, but the assertions now verify the opposite. Maybe update both to reflect that gas price mismatch is intentionally allowed in white-flag mode?

crates/iota-core/src/authority_server.rs
filipdulic: confirmed in team sync, we should not error out

crates/iota-core/src/unit_tests/server_tests.rs
filipdulic: fixed in [commit](https://github.com/iotaledger/iota/pull/10814/commits/1fbfb22ad08ca23a3697b3237d146b82f34a1b06)

crates/iota-types/src/messages_grpc.rs
```diff
@@ -426,3 +422,628 @@ pub enum WaitForEffectsResponse {
...
     Expired { epoch: EpochId },
 }
+
+/// Batch response from a validator to a wait for effects request.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct WaitForEffectsResponse {
+    pub results: Vec<WaitForEffectResponse>,
+}
+
+// =========== Raw prost wire-format types ===========
```
bingyanglin, line 432: I am a little bit confused: here we manually define `Raw* prost structs`, but #10924 generates types from `.proto` files for the new `ValidatorV2/ValidatorPeer` services. Which one is correct or I miss anything?

piotrm50: It's an evolution. Here we adjust the logic with in-code definitions so so that the client and server can handle batches and in the later PR we'll move the definitions to .proto files.

crates/iota-core/src/unit_tests/server_tests.rs
```diff
@@ -39,6 +42,27 @@ use crate::{
     consensus_adapter::MockConsensusClient,
 };
 
+/// Helper to make a tonic request with a native SubmitTransactionsRequest
+/// converted to Raw* for the protobuf wire format.
+fn make_raw_submit_request(
```
piotrm50, line 47: Is this method for testing only? If not then we should either rename `make_tonic_request_for_testing` or rename this function to `make_raw_submit_request_for_testing`.  Is till still relevant if we move to .proto files?

piotrm50 (approved):

crates/iota-core/src/unit_tests/server_tests.rs
filipdulic: will be deleted once proto file based types are included in a subsequent PR

miker83z (approved):

PR #10817, opened 2026-03-20
fix(iota-core): skip pending_consensus_transactions persistence for UserTransactionV1

### LINKED ISSUE #10720: pcool: Remove in-flight buffers and do best-effort transaction issuing
## Problem

In the pcool/whiteflag flow, transaction submission is best-effort: there are no guarantees that a submitted transaction will be executed. If it's dropped or garbage collected in consensus, the validator endpoint request returns an error or simply times out. If a fullnode crashes mid-processing, the wallet should simply retry transaction sending.

Despite this, `UserTransactionV1` transactions are currently being persisted to the `pending_consensus_transactions` DB table (WAL) on the validator side, just like `CertifiedTransaction` in the certificate flow. This persistence is unnecessary for fire-and-forget transactions and wastes disk I/O.

## Scope

Remove usage of pending transaction persistence buffers from the pcool/whiteflag flow. The certificate flow still needs them (for `CertifiedTransaction`, `EndOfPublish`, checkpoint signatures, DKG messages, etc.) until pcool is fully rolled out on all networks.

### Buffers involved

1. **`pending_consensus_transactions` DBMap** (validator, `authority_per_epoch_store.rs`) — persists transactions to disk for crash recovery. Currently writes `UserTransactionV1` unnecessarily. Should skip them.
2. **`pending_consensus_certificates` in-memory HashSet** — already correct: only tracks `CertifiedTransaction`, not `UserTransactionV1`.
3. **`WritePathPendingTransactionLog`** (fullnode, `transaction_orchestrator.rs`) — already correct: cleanup/recovery only runs in QuorumDriver mode. Cleanup deferred to later.
4. **`reexecute_pending_consensus_certs`** (node startup, `iota-node/lib.rs`) — only applies to optimistically-executed owned-object CertifiedTransactions. `UserTransactionV1` always goes through consensus first, so this is a no-op for pcool. Keep with comment for now.

### What changes

- `insert_pending_consensus_transactions`: filter out `UserTransactionV1` internally (matches Sui's approach)
- `submit_and_wait_inner`: skip DB removal for `UserTransactionV1` (never inserted)
- `submit_recovered`: filter out stale `UserTransactionV1` entries (upgrade safety)
- Add clarifying comments to all read/write sites marking them as "certificate-flow only; will be removed once pcool completely takes over"

piotrm50:
# Description of change

In the pcool/whiteflag flow, transaction submission is best-effort — `UserTransactionV1` is fire-and-forget with no crash recovery guarantees. Despite this, these transactions were being persisted to the `pending_consensus_transactions` DB table (WAL) unnecessarily.

This PR is rebased on top of #10615 (certificate-less epoch boundary coordination).

This change:
- Filters out `UserTransactionV1` inside `insert_pending_consensus_transactions` so they are never persisted to the WAL
- Adds clarifying comments to all read/write sites marking them as "certificate-flow only; will be removed once pcool completely takes over"
- Fixes a bug in `submit_and_wait_inner` where the `is_user_tx` check branched on `enable_white_flag_flow` but was redundant with the existing `is_soft_bundle` flag — simplified to only check `CertifiedTransaction`
- Updates comments in `iota-node/src/lib.rs` replacing TODOs with accurate descriptions of pcool behavior

The certificate flow still needs these buffers until pcool is fully rolled out.

Closes #10720

New test:
- `user_transaction_v1_not_persisted_to_pending_consensus` — verifies `UserTransactionV1` is skipped while `CertifiedTransaction` is persisted

All consensus tests pass (`cargo nextest run -p iota-core consensus_tests`).


bingyanglin (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2277,8 +2294,10 @@ impl AuthorityPerEpochStore {
         Ok(())
     }
 
-    /// Insert transactions that will be submitted to consensus into
-    /// the persistent `pending_consensus_transactions` table.
+    /// Persists consensus transactions to the WAL for crash recovery.
+    /// UserTransactionV1 (pcool flow) is fire-and-forget and skipped
+    /// internally. Certificate-flow only; will be removed once pcool
+    /// completely takes over.
```
roman1e2f5p8s, line 2300: This is different from what the code actually does: `UserTransactionV1` is still persisted to `pending_consensus_transactions`. Did you mean to modify the code as well? Otherwise, this part "UserTransactionV1 (pcool flow) is fire-and-forget and skipped internally" should be rewritten.

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -859,15 +863,11 @@ impl ConsensusAdapter {
...
             || if epoch_store.protocol_config().enable_white_flag_flow() {
-                // In the certificate-less mode, `UserTransactionV1` kind corresponds
-                // to user transactions.
                 matches!(
                     transactions[0].kind,
                     ConsensusTransactionKind::UserTransactionV1(_)
                 )
             } else {
-                // In the certificate mode, `CertifiedTransaction` kind corresponds
-                // to user transactions.
```
roman1e2f5p8s, line 870: Since we are constructing the `is_user_tx` flag here, I think it is nice to have an explanation of which kind corresponds to the user tx depending on the flow. It might be clear from `UserTransactionV1`, but not that obvious from `CertifiedTransaction`. So I would keep the comments.

PR #10841, opened 2026-03-20
feat(gRPC): add gRPC client features for `tls-native-roots` and `tls-webpki-roots`

muXxer:
# Description of change

Clients can choose what to use, and the default is set to `tls-native-roots`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15332,6 +15332,7 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
+ "webpki-roots 1.0.4",
  "zstd",
 ]
 
--- a/crates/iota-grpc-client/Cargo.toml
+++ b/crates/iota-grpc-client/Cargo.toml
@@ -17,8 +17,14 @@ futures.workspace = true
 http.workspace = true
 serde.workspace = true
 thiserror.workspace = true
-tonic = { workspace = true, features = ["tls-ring"] }
+tonic.workspace = true
 
 # internal dependencies
 iota-grpc-types.workspace = true
 iota-sdk-types.workspace = true
+
+[features]
+default = ["tls-ring", "tls-native-roots"]
+tls-ring = ["tonic/tls-ring"]
+tls-native-roots = ["tonic/tls-native-roots"]
+tls-webpki-roots = ["tonic/tls-webpki-roots"]
--- a/crates/iota-grpc-client/src/client.rs
+++ b/crates/iota-grpc-client/src/client.rs
@@ -8,7 +8,7 @@ use iota_grpc_types::v0::{
     ledger_service::ledger_service_client::LedgerServiceClient,
     transaction_execution_service::transaction_execution_service_client::TransactionExecutionServiceClient,
 };
-use tonic::{codec::CompressionEncoding, transport::channel::ClientTlsConfig};
+use tonic::codec::CompressionEncoding;
 
 use crate::{api::Result, interceptors::HeadersInterceptor};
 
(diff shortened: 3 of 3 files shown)


semenov-vladyslav (approved):
The change seems trivial: it just provides a mechanism to enable tls-related features in tonic crate. However, it should be noted, that without any of these features being enabled tonic cannot support tls at all (and `ClientTlsConfig` type from `tonic::transport::channel` would be undefined resulting in a compile error). So either one of these features must be made mandatory to enforce TLS support in tonic, or the use of `ClientTlsConfig` in `iota_grpc_client::client` must be feature guarded.

On related issue, it is not specified which TLS keys and certificate iota-grpc-server should use, those are just specified in TlsConfig, but there are no doc entries on this topic. In iota it is common to use trusted self-signed certificates (nodes use NetworkKeyPair and generate a self-signed certificate on the fly). It makes sense to use non-trusted keys in iota-grpc-server because it is a user-facing component, and users cannot trust a public key from iota-node's NetworkKeyPair. But the docs do not explain how iota-grpc-server's TLS certificates are issued, so that it even makes sense to enable "tls-native-roots" or "tls-webpki-roots" features in tonic at all.

In short:
1. guard uses of `ClientTlsConfig` in `iota_grpc_client::client` with `#[cfg(any(feature = "tls-ring", feature = "tls-native-roots", feature = "tls-webpki-roots"))]`
2. add a doc comment that at least one of these features must be enabled to support TLS
3. explain in the docs how iota-grpc-server generates TLS certificate so that users understand which root certificate they need to trust and where to look it up

jkrvivian (approved):

muXxer:
> The change seems trivial: it just provides a mechanism to enable tls-related features in tonic crate. However, it should be noted, that without any of these features being enabled tonic cannot support tls at all (and `ClientTlsConfig` type from `tonic::transport::channel` would be undefined resulting in a compile error). So either one of these features must be made mandatory to enforce TLS support in tonic, or the use of `ClientTlsConfig` in `iota_grpc_client::client` must be feature guarded.
> 
> On related issue, it is not specified which TLS keys and certificate iota-grpc-server should use, those are just specified in TlsConfig, but there are no doc entries on this topic. In iota it is common to use trusted self-signed certificates (nodes use NetworkKeyPair and generate a self-signed certificate on the fly). It makes sense to use non-trusted keys in iota-grpc-server because it is a user-facing component, and users cannot trust a public key from iota-node's NetworkKeyPair. But the docs do not explain how iota-grpc-server's TLS certificates are issued, so that it even makes sense to enable "tls-native-roots" or "tls-webpki-roots" features in tonic at all.
> 
> In short:
> 
> 1. guard uses of `ClientTlsConfig` in `iota_grpc_client::client` with `#[cfg(any(feature = "tls-ring", feature = "tls-native-roots", feature = "tls-webpki-roots"))]`
> 2. add a doc comment that at least one of these features must be enabled to support TLS
> 3. explain in the docs how iota-grpc-server generates TLS certificate so that users understand which root certificate they need to trust and where to look it up

I agree with the feature gating. I added it in https://github.com/iotaledger/iota/pull/10841/commits/0223b8755048ca4faae38a13aecb475a2df6fb30.

I don't think we need to add docs how to set this up. This here is about gRPC clients that connect to fullnodes. Fullnode operators will certainly use a reverse proxy and some certificates signed by Let's Encrypt etc. They won't use self signed certificates for their public node setups.

semenov-vladyslav (approved):

jkrvivian (approved):

PR #10816, opened 2026-03-20
feat(indexer): Add pruning documentation to README

tomxey:
# Description of change

Documents how to configure indexer pruning via TOML config, including retention policies, prunable tables, and the delayed pruning mechanism.

## Links to any relevant issues

fixes https://github.com/iotaledger/iota/issues/9830

## How the change has been tested

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

- [ ] Synchronization of the indexer from genesis for a network including migration objects.
- [ ] Restart of indexer synchronization locally without resetting the database.
- [ ] Restart of indexer synchronization on a production-like database.
- [ ] Deployment of services using Docker.
- [ ] Verification of API backward compatibility.

PR #10468, opened 2026-02-25
feat(indexer): Add tx seq indexes to pruned tables, increase pruned batch for cp pruning

tomxey:
# Description of change

Increase pruning performance by adding indexes to tx seq column, and by increasing batch size for cp based pruning

## Links to any relevant issues

fixes https://github.com/iotaledger/iota/issues/10467

## How the change has been tested

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)

Performance was tested by running benchmark on staging server, all tables were pruning with good performance.

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

- [ ] Synchronization of the indexer from genesis for a network including migration objects.
- [ ] Restart of indexer synchronization locally without resetting the database.
- [ ] Restart of indexer synchronization on a production-like database.
- [ ] Deployment of services using Docker.
- [ ] Verification of API backward compatibility.

PR #10875, opened 2026-03-23
feat(indexer): Watermarks-based pruning

### LINKED ISSUE #9824: [indexer]: Make pruning fully use the `watermarks` table
Right now we have jobs that are filling the `watermarks` table, but the actual pruning job is not reading from it.

We should adapt the [Pruner::start](https://github.com/iotaledger/iota/blob/8fe4d49fd573c45965cc69ee43541fa3dd7db056/crates/iota-indexer/src/pruning/pruner.rs#L146) job to read `watermarks` and prune tables based on it.

We should also confirm that `watermarks` table indeed contains all necessary information for all tables. Some documentation of the mechanism could be nice.

tomxey:
# Description of change

Modifies the indexer's pruning system to use per-table watermarks with configurable retention policies, replacing the previous monolithic pruning approach.

  Commits included:

  1. **Watermarks-based pruner** — Each table is pruned independently using its own watermark entry, with a 2-hour delay to protect in-flight reads (#9883)
  2. **Background watermark cache** — Periodic background task reads watermarks into memory to avoid per-query DB overhead (#9909)
  3. **Watermarks table cleanup** — Renamed columns for clarity (e.g. `checkpoint_hi_inclusive` → `max_committed_cp`, `reader_lo` → `min_available_cp`/`min_available_tx`) (#9926)
  4. **Watermark-aware reads** — RPC reads detect when querying pruned ranges (#9963)
  5. **Object-based dependency checking** — Optimistic indexing checks input objects instead of previous transactions, making it resilient to pruning (#9975)
  6. **Unified optimistic tx pruning** — Optimistic transactions pruned by the same generic mechanism as all other tables (#10020)
  7. **Prune tx_global_order** — Include `tx_global_order` table in the generic pruner so it doesn't grow unbounded (#10049)
  8. **Checkpoint range pruning** — Prune checkpoint-based tables by range instead of one-by-one for better efficiency (#10102)
  9. **Object write guards** — Version checks on object deletions, ordered persistence (tx_global_order before objects) to prevent race conditions (#10460)
  10. **Pruning indexes** — Added tx_sequence_number indexes on all pruned index tables for efficient DELETE operations (#10468)
  11. **Read-after-write consistency** — Optimistic indexing guarantees transaction effects are persisted before returning response to user (#10641)
  12. **Vacuum tuning** — Configured aggressive autovacuum settings on pruned tables to prevent dead tuple accumulation (#10593)
  13. **Simplified tx_global_order semantics** — Removed `IndexStatus` enum, track object finality via `finalized_in_cp` column instead of separate update
  step, unified `CheckpointTxGlobalOrder` into `TxGlobalOrder` (#10753)
  14. **Pruning documentation** — Added README docs for configuring pruning via TOML config, including retention policies, prunable tables, and delayed pruning
   mechanism (#10816)
  15. **E2E test feature propagation** — Forward `pg_integration` feature flag through `iota-graphql-e2e-tests` → `iota-graphql-rpc` → `iota-indexer` so e2e
  tests use the 1s pruning delay instead of 2h production value (#10878)

fixes https://github.com/iotaledger/iota/issues/9824

- [x] Patch-specific tests (correctness, functionality coverage): Tested on staging server: all tables pruned correctly, table sizes stable, ingestion speed unaffected

### Release Notes

- [x] Indexer:
 - Pruning mechanism switched to use the per-table watermarks, allowing for independent pruning of every table with retention policies configured separately for every table (using already existing `--pruning-config-path` argument)
 - Optimistic indexing now uses object based dependency check, making it resilient to pruning of old (dependent) transactions.
 - The `--optimistic-pruner-batch-size` CLI argument is deprecated and no longer used. Optimistic transactions are now pruned by the unified pruner.
 - RPC reads now detect querying pruned data ranges when possible and return appropriate error.
 - More aggressive autovacuuming has been configured for pruned tables. Operators using large databases should ensure sufficient `shm_size` (e.g. `2g`) on the Postgres container for effective vacuuming.


sergiupopescu199 (approved):
lgtm 🦫🐧🐌

kodemartin (approved):

PR #10883, opened 2026-03-23
feat(gRPC): add StateService and MovePackageService endpoints with background index backfill

bingyanglin:
# Description of change

Implements `StateService` and `MovePackageService` gRPC services with all proto-defined endpoints, plus supporting index tables with non-disruptive background backfill.

### New endpoints

- `StateService.ListOwnedObjects`: streaming objects by owner, with optional type filter and read-mask
- `StateService.ListDynamicFields`: streaming dynamic fields by parent, with conditional field-object loading
- `StateService.GetCoinInfo`: coin metadata, treasury cap, supply state, and regulated-coin metadata
- `MovePackageService.ListPackageVersions`: streaming all on-chain versions of a Move package

### Indexing (`iota-core` / `iota-types`)

- Add `package_version` and `regulated_coin` index tables, so there is **no DB version bump**; populated via background backfill on existing nodes, inline on fresh `init()`
- Endpoints return `Unavailable` with `RetryInfo` while backfill runs; `FailedPrecondition` on nodes without indexes
- Add `object_type` filter to `account_owned_objects_info_iter`
- Extend `RestIndexes` and `GrpcStateReader` traits with new methods

### Client (`iota-grpc-client`)

- Add `state` and `move_package` API modules with high-level methods
- Extract `collect_stream` helper; refactor existing streaming methods to use it

### Refactoring

- Extract shared validation helpers into `validation.rs`; refactor all `ledger_service` handlers to use them
- Extract shared test utilities (`publish_example_package`, `assert_tonic_error`, `collect_streaming_responses!`) into `grpc/utils.rs`

### Proto (`iota-grpc-types`)

- Refine `dynamic_field.proto` and `state_service.proto` docs (can be discussed, but I guess the refining make them clearer)
- Add `LIST_DYNAMIC_FIELDS_READ_MASK` and `LIST_OWNED_OBJECTS_READ_MASK` constants

None.


crates/iota-core/src/rest_index.rs
```diff
@@ -969,6 +1201,139 @@ impl LiveObjectIndexer for RestLiveObjectIndexer<'_> {
...
+}
+
+fn backfill_package_version(
+    tables: &IndexStoreTables,
+    authority_store: &AuthorityStore,
+    flag: &AtomicBool,
+) {
+    info!("Starting package_version backfill");
+    let indexer = PackageVersionBackfillIndexer { tables };
+    match crate::par_index_live_object_set::par_index_live_object_set(authority_store, &indexer) {
```
muXxer: Wouldn't it make sense for the backfill to loop only once over the live object set, but backfill package version and regulated coins at the same time? Right now we loop two times over the live objects.

For later, when we fix the indexes that come from a formal snapshot, we could directly re-use such improved logic to index everything we need in one live object loop.

muXxer: Basically we would need a LiveObjectIndexer that combines both LiveObjectIndexer implementations we have now, and then only call `crate::par_index_live_object_set::par_index_live_object_set` once, right?

crates/iota-core/src/rest_index.rs
```diff
@@ -969,6 +1201,139 @@ impl LiveObjectIndexer for RestLiveObjectIndexer<'_> {
...
+    }
+}
+
+impl LiveObjectIndexer for RegulatedCoinBatchIndexer<'_> {
+    fn index_object(&mut self, object: Object) -> Result<(), StorageError> {
+        if let Some((key, object_id)) = try_create_regulated_coin_info(&object) {
+            self.batch
+                .insert_batch(&self.tables.regulated_coin, [(key, object_id)])?;
+        }
+        if self.batch.size_in_bytes() >= 1 << 27 {
```
muXxer, line 1675:
```suggestion
        // If the batch size grows to greater that 128MB then write out to the DB so
        // that the data we need to hold in memory doesn't grown unbounded.
        if self.batch.size_in_bytes() >= 1 << 27 {
```

crates/iota-core/src/rest_index.rs
```diff
@@ -969,6 +1201,139 @@ impl LiveObjectIndexer for RestLiveObjectIndexer<'_> {
...
+    }
+}
+
+impl LiveObjectIndexer for PackageVersionBatchIndexer<'_> {
+    fn index_object(&mut self, object: Object) -> Result<(), StorageError> {
+        if let Some((key, info)) = try_create_package_version_info(&object) {
+            self.batch
+                .insert_batch(&self.tables.package_version, [(key, info)])?;
+        }
+        if self.batch.size_in_bytes() >= 1 << 27 {
```
muXxer, line 1675:
```suggestion
        // If the batch size grows to greater that 128MB then write out to the DB so
        // that the data we need to hold in memory doesn't grown unbounded.
        if self.batch.size_in_bytes() >= 1 << 27 {
```

muXxer (changes requested):

crates/iota-core/src/rest_index.rs
bingyanglin: You are right. I combined the two loops in [f31c075](https://github.com/iotaledger/iota/pull/10883/commits/f31c075ba20033a4ad09ef6930a5be728c46a77d).

crates/iota-core/src/rest_index.rs
muXxer, line 450: The comment is wrong, it is now used by both. Also, we need to add a comment that we might be able to remove the `DynamicFieldIndexInfo` and replace it with `()` later.

muXxer, line 429: The comment is wrong, it is now used by both.

crates/iota-core/src/rest_index.rs
```diff
@@ -164,6 +181,24 @@ struct IndexStoreTables {
     /// ObjectID of its coorisponding CoinMetadata.
     /// REST-API only
     coin: DBMap<CoinIndexKey, CoinIndexInfo>,
```
muXxer, line 457: Actually our `CoinIndexKey` doesn't allow to sort by balance, which is a nice feature. If we are reindexing everything for the `regulated_coin` anyways, we could directly introduce a `coin_v2` that has all these features, and drop the `coin` in the next version. We should introduce the `CoinIndexKeyV2` for this new table then.

crates/iota-core/src/rest_index.rs
```diff
@@ -164,6 +181,24 @@ struct IndexStoreTables {
...
+    package_version: DBMap<PackageVersionKey, PackageVersionInfo>,
+
+    /// An index of RegulatedCoinMetadata objects by coin type.
+    ///
+    /// Maps coin type to the ObjectID of the `RegulatedCoinMetadata<T>` object.
+    /// Bounded by the live object set (at most one entry per regulated coin).
+    /// Kept separate from the `coin` table so adding this index does not
+    /// require a DB version bump.
+    /// gRPC-server only
+    regulated_coin: DBMap<CoinIndexKey, ObjectID>,
```
muXxer: See comment above about the `coin` table.

crates/iota-core/src/rest_index.rs
muXxer: I think we need to introduce a new `owners_2` table, and use the new `OwnerIndexKeyV2` and the new `OwnerIndexInfoV2`. This would allow us to filter by Address+ObjectType (like specifc coins) and their balance descending.

```rust
pub struct OwnerIndexKeyV2 {
    pub owner: SuiAddress,

    pub object_type: StructTag,

    // If this object is coin-like (eg 0x2::coin::Coin) then this will be the balance of the coin
    // inverted `!coin.balance` in order to force sorting of coins to be from greatest to least
    pub inverted_balance: Option<u64>,

    pub object_id: ObjectID,
}

pub struct OwnerIndexInfoV2 {
    // object_id and type of this object are a part of the key
    pub version: SequenceNumber,
}
```

muXxer: Ah but instead of the StructTag, we should use a hash of the struct tag actually! This is much more robust. Maybe 8 bytes hash? And then store the actual struct tag in the OwnerIndexInfo in case we need it? If not, then we can skip that.

muXxer (changes requested):

crates/iota-grpc-types/proto/iota/grpc/v0/move_package_service.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// A simplified representation of a package version
+message PackageVersion {
+  option (iota.grpc.message_accessors) = "with";
+
+  // The storage ID of this package version
+  optional iota.grpc.v0.types.ObjectId package_id = 1;
+  // The version number
+  optional uint64 version = 2;
```
muXxer: Actually we also index the `optional iota.grpc.v0.types.ObjectId storage_id = 3;`.

crates/iota-core/src/rest_index.rs
muXxer: But on the other hand we would loose the ability to scan for types like described in the proto file.
```
  // Providing an object type with no type params will return objects of that
  // type with any type parameter, e.g. `0x2::coin::Coin` will return all
  // `Coin<T>` objects regardless of the type parameter `T`. Providing a type
  // with a type param will restrict the returned objects to only those objects
  // that match the provided type parameters, e.g.
  // `0x2::coin::Coin<0x2::iota::IOTA>` will only return `Coin<IOTA>` objects.
```

Looking at the struct tag: 
```
pub struct StructTag {
    pub address: AccountAddress,
    pub module: Identifier,
    pub name: Identifier,
    // alias for compatibility with old json serialized data.
    #[serde(rename = "type_args", alias = "type_params")]
    pub type_params: Vec<TypeTag>,
}
```
Maybe we can create a hash out of address+module+name and another hash out of `Vec<TypeTag>` and concatenate those. Then we could still do the same filtering 🤔

bingyanglin: I agree using Hash is a good idea, but I am afraid 8-byte hash might collide easily (2^32 operations, GPU can attack it easily?)? I guess 32-byte hash (birthday resistance) is more robust?

muXxer: But how does that attack vector look like? Someone mining packages? Then he has to pay for it and publish that much packages

crates/iota-grpc-types/proto/iota/grpc/v0/move_package_service.proto
bingyanglin: Fixed in [7c5a43b](https://github.com/iotaledger/iota/pull/10883/commits/7c5a43ba149326f6bb3ce60d3f5ab72e2007b536).

crates/iota-core/src/rest_index.rs
bingyanglin: Fixed in [7c5a43b](https://github.com/iotaledger/iota/pull/10883/commits/7c5a43ba149326f6bb3ce60d3f5ab72e2007b536).

crates/iota-core/src/rest_index.rs
bingyanglin: Fixed in [7c5a43b](https://github.com/iotaledger/iota/pull/10883/commits/7c5a43ba149326f6bb3ce60d3f5ab72e2007b536).

crates/iota-core/src/rest_index.rs
bingyanglin: Fixed in [7c5a43b](https://github.com/iotaledger/iota/pull/10883/commits/7c5a43ba149326f6bb3ce60d3f5ab72e2007b536) per our discussion.

muXxer (approved):

PR #10886, opened 2026-03-23
upstream(core): Add frozen/shared objects support to test checkpoint builder

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@c9f63c6](https://github.com/MystenLabs/sui/commit/c9f63c6fbf09e50f6120b6a0d42edf3f21c99030)
- Description:

Add support for frozen objects, shared object modification and reading
shared objects to the test checkpoint builder, and use that to write a
test just for `checkpoint_input_objects`.

Only the `iota-types` changes were ported (test effects builder and test checkpoint data builder).

Part of #10609

### DIFF
--- a/crates/iota-types/src/test_checkpoint_data_builder.rs
+++ b/crates/iota-types/src/test_checkpoint_data_builder.rs
@@ -30,7 +30,8 @@ use crate::{
     object::{GAS_VALUE_FOR_TESTING, MoveObject, Object, Owner},
     programmable_transaction_builder::ProgrammableTransactionBuilder,
     transaction::{
-        EndOfEpochTransactionKind, SenderSignedData, Transaction, TransactionData, TransactionKind,
+        EndOfEpochTransactionKind, ObjectArg, SenderSignedData, Transaction, TransactionData,
+        TransactionKind,
     },
 };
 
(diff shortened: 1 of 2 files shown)


Thoralf-M (approved):

crates/iota-types/src/test_checkpoint_data_builder.rs
```diff
@@ -332,6 +347,30 @@ impl TestCheckpointDataBuilder {
...
 
+    /// Add an immutable object as an input to the transaction.
+    ///
+    /// Fails if the object is not live or if its owner is not
+    /// [Owner::Immutable]).
+    pub fn read_frozen_object(mut self, object_id: u64) -> Self {
+        let tx_builder = self.checkpoint_builder.next_transaction.as_mut().unwrap();
+        let object_id = Self::derive_object_id(object_id);
+
+        let Some(obj) = self.live_objects.get(&object_id) else {
```
bingyanglin: `mutate_owned_object` uses `.expect("Mutating an object that doesn't exist")` for the missing-object case but this one uses `panic!`. Maybe just use `.expect("Frozen object not found")` here too so it reads the same way?

jkrvivian: Fixed in https://github.com/iotaledger/iota/pull/10886/commits/f7d834e11e1a233b8c4af2c68702e16ab68344b5.

bingyanglin (approved):

PR #10894, opened 2026-03-23
refactor(explorer): generalizes Identity

alexruzenhack:
# Description of change

- Extract the view `SideBySidePanelsView` as an UI component to be reused as `SideBySidePanels`
- Extract a more specialized helper `objectOwnerHelper.ts`
- Move view `TransactionsView` to `/common` to let it be reused among Trust Framework pages
- Move component `IdentityContent` a level down to let `/views` cluster only view components
- Move `headerMetadataHelper` a level down to let it be reused among Trust Framework pages

N/A


apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (isOwnerType(owner, 'Shared')) {
+        return 'Shared';
+    }
+
+    if (isOwnerType(owner, 'Immutable')) {
+        return 'Immutable';
+    }
+
+    throw new Error(
+        'The provided owner do not match a type variation. It is either not an OwnerObject or the type list has grown.',
```
VmMad: nit: in my opinion there's no need to overexplain the error
```suggestion
        `Unsupported owner type: ${owner}`.,
```

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
VmMad, line 1: This code looks agnostic to trust-framework. Do you think it could be placed outside of it?

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    switch (getOwnerType(owner)) {
+        case 'AddressOwner':
+            return getAddressOwner(owner);
+        case 'ObjectOwner':
+            return getObjectOwner(owner);
+        case 'Shared':
+            return objectId || null;
+        case 'Immutable': // NOTE: What to do with immutable?
```
VmMad: If possible it would be nice to track things todo outside of the code, for example a backlog task

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/**
+ * Gets the type of the object owner.
+ * @param owner The object owner.
+ * @returns The type of the owner as a string.
+ */
+export function getOwnerType(owner?: ObjectOwner | null): string {
+    if (owner == null) {
+        return 'Undefined';
```
VmMad: Wouldn't it be better to return null in this case? so that you don't have to manually handle "Undefined" as a string in cases where you want to reuse this function?

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+        return null;
+    }
+
+    switch (getOwnerType(owner)) {
+        case 'AddressOwner':
+            return getAddressOwner(owner);
+        case 'ObjectOwner':
+            return getObjectOwner(owner);
+        case 'Shared':
+            return objectId || null;
```
VmMad: Just wanted to check if this is correct, because in some other places we currently display a static string `Shared` as owner

Image

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: Not right now. The only use in currently here, under trust-framework. If we see the need for it later than we can extract it to somewhere else.

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: Oh! it is not necessary now because owners are either an address or an object. But there is this possibility that may worth to handle in the future — who knows? Also, I write these comments to provoke comments like this :) For now, I'm removing it.

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: I did this way because it is used as a label at `ControllerView.tsx`. However, by returning null I turn the `ControllerView.tsx` usage more explicit, which is also desirable.

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: Well, maybe I have a limited understanding of it. My rational was to appoint to the object itself when it has a "shared" ownership. But, yeah, maybe the best course is to just show "Shared".

alexruzenhack: And perhaps I could do the same to "Immutable"...

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: Done at: https://github.com/iotaledger/iota/pull/10894/commits/dcc352cad429080c8033c17d5cbbaa6d383ae195

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: See: https://github.com/iotaledger/iota/pull/10894/commits/8a1fc4111dc76566ade31cb2b0d7f013d784e753

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: See: https://github.com/iotaledger/iota/pull/10894/commits/8a1fc4111dc76566ade31cb2b0d7f013d784e753

apps/explorer/src/pages/trust-framework/objectOwnerHelper.ts
alexruzenhack: See: https://github.com/iotaledger/iota/pull/10894/commits/8a1fc4111dc76566ade31cb2b0d7f013d784e753

VmMad (approved):

PR #10905, opened 2026-03-24
feat(indexer): Bail early from dependency wait when objects are superseded

### LINKED ISSUE #10742: feat(indexer): Reduce wait time if checkpoint indexing wins over optimistic indexing
Currently if tx is executed optimistically through indexer, but checkpoint indexing indexes it faster, optimistic indexing will be stuck for full backoff time on dependency check (3s currently), waiting for input objects to be at correct versions.

This could be detected earlier. If optimistic dependency checker would see that versions in the DB are greater than the requested ones it means that there is no point in waiting and we can fail the dependency check right away. (and effectively fall back to reading the tx from DB)

This situation should be rather rare.

tomxey:
# Description of change

When optimistic indexing waits for input object dependencies, it retries with exponential backoff until all objects exist at exact versions. If checkpoint indexing (or another optimistic tx) has already advanced an object past the requested version, the exact match will never succeed, but the retry
loop continues for the full timeout (~3s).

This adds a cheap `SELECT EXISTS(...)` check on the retry path: if any dependency object already has a newer version in the DB, stop retrying immediately and fall back to checkpoint-indexed data.

fixes #10742

### DIFF
--- a/crates/iota-indexer/src/optimistic_indexing.rs
+++ b/crates/iota-indexer/src/optimistic_indexing.rs
@@ -36,7 +36,7 @@ use crate::{
         display::StoredDisplay,
         transactions::{OptimisticTransaction, StoredTransaction, TxGlobalOrder},
     },
-    read::IndexerReader,
+    read::{IndexerReader, InputObjectsStatus},
     store::{IndexerStore, PgIndexerStore},
     transactional_blocking_with_retry_with_conditional_abort,
     types::{
@@ -88,7 +88,6 @@ impl OptimisticTransactionExecutor {
         &self,
         input_obj_keys: Vec<(ObjectID, SequenceNumber)>,
     ) -> Result<(), IndexerError> {
-        let expected_count = input_obj_keys.len();
         let backoff = backoff::ExponentialBackoff {
             initial_interval: Duration::from_millis(100),
             max_elapsed_time: Some(WAIT_FOR_DEPS_MAX_ELAPSED_TIME),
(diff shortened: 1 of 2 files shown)


kodemartin (commented):
You could consider consolidating `count_existing_object_keys...` and `has_superseded_` into a single query `input_objects_exist` (or something similar) as follows (please verify correctness)

```sql
WITH input_objects(id, version) AS (
    VALUES {}
),
max_chk AS (
    SELECT COALESCE(MAX(sequence_number), -1) AS max_sn FROM checkpoints
),
matches AS (
    SELECT
        i.id                AS input_id,
        i.version           AS input_version,
        o.object_version    AS actual_version,
        t.finalized_in_cp   AS chk
    FROM input_objects i
    LEFT JOIN objects o ON o.object_id = i.id
)
SELECT
    CASE
        WHEN COUNT(*) FILTER (WHERE actual_version IS NULL)         > 0 THEN NULL     -- missing object
        WHEN COUNT(*) FILTER (WHERE actual_version < input_version) > 0 THEN NULL     -- at least one object not there yet
        WHEN COUNT(*) FILTER (WHERE actual_version > input_version) > 0 THEN FALSE    -- at least one object superseded
        WHEN COUNT(*) FILTER (WHERE chk > (SELECT max_sn FROM max_seq)) > 0 THEN NULL -- chk not fully indexed yet
        ELSE TRUE
    END AS result
FROM matches;
```

tomxey:
> You could consider consolidating `count_existing_object_keys...` and `has_superseded_` into a single query `input_objects_exist` (or something similar) as follows (please verify correctness)
> 
> ```sql
> WITH input_objects(id, version) AS (
>     VALUES {}
> ),
> max_chk AS (
>     SELECT COALESCE(MAX(sequence_number), -1) AS max_sn FROM checkpoints
> ),
> matches AS (
>     SELECT
>         i.id                AS input_id,
>         i.version           AS input_version,
>         o.object_version    AS actual_version,
>         t.finalized_in_cp   AS chk
>     FROM input_objects i
>     LEFT JOIN objects o ON o.object_id = i.id
> )
> SELECT
>     CASE
>         WHEN COUNT(*) FILTER (WHERE actual_version IS NULL)         > 0 THEN NULL     -- missing object
>         WHEN COUNT(*) FILTER (WHERE actual_version < input_version) > 0 THEN NULL     -- at least one object not there yet
>         WHEN COUNT(*) FILTER (WHERE actual_version > input_version) > 0 THEN FALSE    -- at least one object superseded
>         WHEN COUNT(*) FILTER (WHERE chk > (SELECT max_sn FROM max_seq)) > 0 THEN NULL -- chk not fully indexed yet
>         ELSE TRUE
>     END AS result
> FROM matches;
> ```

@kodemartin done in [b6f03be](https://github.com/iotaledger/iota/pull/10905/commits/b6f03be3d9a82381ca098fe5496808f444c10210)

kodemartin (approved):
lgtm 🌸

sergiupopescu199 (approved):
lgtm!

PR #10915, opened 2026-03-24
chore(ts-sdk): Use `indexer` instead `api` urls in the ts packages

marc2332:


begonaalvarezd (commented):
I would add a changeset

begonaalvarezd (approved):

KeitoTadashi (approved):

PR #10930, opened 2026-03-24
fix(ts-sdk): Change default network in `.env.defaults` to `mainnet`

marc2332:
This will also make the explorer previews use mainnet (or another one if wanted) by default


marc2332:
not needed anymore

PR #10932, opened 2026-03-24
fix(gRPC): check for lowest available checkpoint

muXxer:
# Description of change

The genesis checkpoint is never pruned, so without a check if the requested checkpoint is below the lowest available checkpoint, that causes some inconsistencies. The REST API also does exactly this check, so it's better to add it. 

### DIFF
--- a/crates/iota-grpc-server/src/types.rs
+++ b/crates/iota-grpc-server/src/types.rs
@@ -847,6 +847,18 @@ impl GrpcReader {
             while start <= end {
                 // Try fetching historical data from the DB first
                 if start <= latest {
+                    // Check if the checkpoint has been pruned since we started
+                    // (e.g. genesis checkpoint 0 is always in DB but may be
+                    // below the pruning watermark).
+                    let lowest_available = state_reader
+                        .get_lowest_available_checkpoint()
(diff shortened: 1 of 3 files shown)


crates/iota-grpc-server/src/types.rs
```diff
@@ -849,6 +849,18 @@ impl GrpcReader {
             while start <= end {
                 // Try fetching historical data from the DB first
                 if start <= latest {
+                    // Check if the checkpoint has been pruned since we started
+                    // (e.g. genesis checkpoint 0 is always in DB but may be
+                    // below the pruning watermark).
+                    let lowest_available = state_reader
```
bingyanglin, line 853: If a client requests an already-pruned start, the stream opens fine and then immediately errors on the first item. `get_checkpoint_data` already rejects pruned checkpoints before doing any work. Maybe add the same eager check in stream_checkpoint_data before creating the stream?

bingyanglin (approved):

crates/iota-grpc-server/src/types.rs
muXxer: Added in https://github.com/iotaledger/iota/pull/10932/commits/281a6e37d7ffac3d0628e186fa721583715559c1

PR #10944, opened 2026-03-25
feat(gRPC): add gRPC-Web support for browser clients

bingyanglin:
# Description of change

Enables gRPC-Web support on the gRPC server so browser-based clients can
communicate over HTTP/1.1 using the `application/grpc-web` content type.

Ported https://github.com/MystenLabs/sui/commit/8c91fcd8b3905aae4136d929e0cc62e1adfb4704 and added a test.

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6746,11 +6746,13 @@ dependencies = [
  "prometheus",
  "prost 0.14.1",
  "prost-types 0.14.1",
+ "reqwest",
  "serde",
  "tokio",
  "tokio-stream",
  "tokio-util 0.7.18",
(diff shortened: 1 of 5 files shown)


semenov-vladyslav (approved):

bingyanglin:
From @Alex6323 
> seems like we don't need grpc-web in the node, because we can use this in the reverse proxies https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/grpcweb/
also, regarding the https support in the node, alex said that no one should run a public node without a reverse proxy anyways, so there might be no point in adding that functionality

Thus we close this PR.

PR #10963, opened 2026-03-26
fix(iota-data-ingestion-core): refine clients initialization

### LINKED ISSUE #10962: [iota-data-ingestion-core]: remove gRPC connection backoff
This [commit](https://github.com/iotaledger/iota/pull/10754/changes/8555036d3057860604801ec1b5034552001b347e) was added  by the node team to fix the flaky tests encountered, with the iota indexer.

The `get_health` call during gRPC client initialization in `iota-data-ingestion-core` causes flaky indexer tests, if the gRPC server isn't ready yet, the health check fails and the client falls back to REST (which no longer exists after the REST removal). 

The fix is to remove the `get_health` call entirely. Errors are already caught during checkpoint download, which has built-in backoff and retry. This matches the behavior of the hybrid (V1) & historical (V2) client path in CheckpointReader, which doesn't perform a health check on initialization.

sergiupopescu199:
# Description of change

This PR removes the `get_health` check when instantiating a gRPC fullnode client, previously it was needed to choose between a gRPC or a REST API fullnode connection. With the removal of the fullnode REST API this check is no longer needed.

fixes #10962

- Ran the indexer using alphanet checkpoints and made sure that the CheckpointReaderV2 was working as expected
- Ran all indexer tests
```shell
cargo nextest run -p iota-indexer --features pg_integration --test-threads 1
```

### Release Notes

- [x] Indexer:  From version `v1.21.0`, the `CheckpointReaderV1` in `iota-data-ingestion-core` will be deprecated. ⚠️⚠️⚠️ Operators maintaining custom indexers are advised to migrate to `CheckpointReaderV2` via [`IndexerExecutor::run_with_config`](https://iotaledger.github.io/iota/iota_data_ingestion_core/struct.IndexerExecutor.html#method.run_with_config). With pruning enabled on fullnodes, custom indexers will need `CheckpointReaderV2` to use hybrid historical storage for syncing from genesis, making `CheckpointReaderV1` obsolete. ⚠️⚠️⚠️

### DIFF
--- a/crates/iota-data-ingestion-core/src/reader/v1.rs
+++ b/crates/iota-data-ingestion-core/src/reader/v1.rs
@@ -11,7 +11,7 @@ use std::{
 };
 
 use backoff::backoff::Backoff;
-use futures::{StreamExt, TryFutureExt};
+use futures::StreamExt;
 use iota_grpc_client::Client as GrpcClient;
 use iota_metrics::spawn_monitored_task;
 use iota_types::{
@@ -24,9 +24,9 @@ use tokio::{
         mpsc::{self, error::TryRecvError},
         oneshot,
     },
-    time::{sleep, timeout},
+    time::timeout,
 };
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info};
 
 #[cfg(not(target_os = "macos"))]
 use crate::reader::fetch::init_watcher;
@@ -111,8 +111,6 @@ impl Default for ReaderOptions {
 
 /// Remote checkpoint store backends.
 enum RemoteStore {
-    ObjectStore(Box<dyn ObjectStore>),
-    Fullnode(Box<iota_grpc_client::Client>),
     Hybrid(Box<dyn ObjectStore>, Box<iota_grpc_client::Client>),
 }
 
@@ -122,10 +120,6 @@ impl CheckpointReader {
         checkpoint_number: CheckpointSequenceNumber,
     ) -> CheckpointResult {
         match store {
-            RemoteStore::ObjectStore(store) => {
-                fetch_from_object_store(store, checkpoint_number).await
-            }
-            RemoteStore::Fullnode(client) => fetch_from_full_node(client, checkpoint_number).await,
             RemoteStore::Hybrid(store, client) => {
                 match fetch_from_full_node(client, checkpoint_number).await {
                     Ok(result) => Ok(result),
@@ -163,7 +157,7 @@ impl CheckpointReader {
         }
     }
 
-    fn start_remote_fetcher(&mut self) -> mpsc::Receiver<CheckpointResult> {
+    async fn start_remote_fetcher(&mut self) -> mpsc::Receiver<CheckpointResult> {
         let batch_size = self.options.batch_size;
         let start_checkpoint = self.current_checkpoint_number;
         let (sender, receiver) = mpsc::channel(batch_size);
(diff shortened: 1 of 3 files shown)


kodemartin (approved):
lgtm

crates/iota-data-ingestion-core/src/reader/v1.rs
```diff
@@ -174,67 +168,25 @@ impl CheckpointReader {
...
-                            "failed to connect to gRPC fullnode after retries, falling back to object store: {err:?}"
-                        );
-                        let object_store =
-                            create_remote_store_client(url, remote_store_options, timeout_secs)
-                                .expect("failed to create remote store client");
-                        RemoteStore::ObjectStore(object_store)
-                    }
-                }
-            };
+        let (fullnode_url, object_store_url) = url.split_once('|').unwrap_or((&url, &url));
```
bingyanglin, line 171: When a single URL is passed (no `|`), `unwrap_or((&url, &url))` uses the same URL for both gRPC and object store. The object store client pointed at a gRPC endpoint will never fetch anything. Maybe skip creating the object store client for single-URL case, since it's always gRPC now? Or do I miss anything?

crates/iota-data-ingestion-core/src/reader/v1.rs
```diff
@@ -174,67 +168,25 @@ impl CheckpointReader {
...
+            timeout_secs,
+        )
+        .expect("failed to create remote store client");
+
+        let grpc_client = GrpcClient::connect(fullnode_url)
+            .await
+            .map(|client| {
+                client.with_max_decoding_message_size(GRPC_MAX_DECODING_MESSAGE_SIZE_BYTES)
+            })
+            .expect("failed to connect to gRPC fullnode");
```
bingyanglin, line 185: I am just confused for v2 we use `?` to `propagate` the error but here we panic, any reason? Or we could make it consistent?

crates/iota-data-ingestion-core/src/reader/v1.rs
```diff
@@ -174,67 +168,25 @@ impl CheckpointReader {
...
+        let (fullnode_url, object_store_url) = url.split_once('|').unwrap_or((&url, &url));
 
+        let object_store = create_remote_store_client(
+            object_store_url.to_string(),
+            remote_store_options,
+            timeout_secs,
+        )
+        .expect("failed to create remote store client");
+
+        let grpc_client = GrpcClient::connect(fullnode_url)
```
bingyanglin, line 180: The old retry loop with `get_health()` logged `"not ready, retrying..."` so operators knew the fullnode wasn't up yet. Now those failures only show up as checkpoint fetch errors, is it descriptive?

crates/iota-data-ingestion-core/src/reader/v1.rs
sergiupopescu199: Good point. The approach is intentional when a single URL is provided, the object store client is created but the gRPC path is attempted first. If the URL is a gRPC endpoint, gRPC succeeds and the object store client is never used. If the URL is an object store endpoint, the gRPC call fails and it falls back to the object store, which then succeeds. So both URL types work correctly without the `|` separator. That said, V1 is being deprecated in `v1.21.0` in favor of V2, so further optimization here isn't warranted.

- https://github.com/iotaledger/iota/issues/10967

crates/iota-data-ingestion-core/src/reader/v1.rs
sergiupopescu199: Yes, the errors are descriptive when the gRPC stream fails to fetch a checkpoint, the worker pool logs a `transient worker execution error` with the full gRPC status and error message. The retry is handled automatically by the worker pool's backoff mechanism.

crates/iota-data-ingestion-core/src/reader/v1.rs
sergiupopescu199: V1 is legacy code inherited from Sui — the `expect` pattern was already there before this change. V2 uses `?` propagation which is the correct approach. Since V1 is being deprecated in `v1.21.0`, aligning the error handling isn't worth the churn.

bingyanglin (approved):

tomxey (approved):

PR #10978, opened 2026-03-26
feat(iota-core): [P-COOL] wire format server side implementation

### LINKED ISSUE #10974: pcool(Wire Format): Server-side Implementation

filipdulic:
# Description of change

Implements the server-side ValidatorV2 gRPC service for the certificate-less (P-COOL) transaction flow.

### What's new

- **`SubmitTx` RPC** — accepts up to 256 transactions, streams `TxStatusUpdate` per tx (Submitted, Executed, Rejected, Expired)
- **`GetTxStatus` RPC** — queries finality status for previously submitted transactions with optional execution details
- **`NotifyCapabilities` RPC** — accepts signed capability notifications from non-committee validators (moved from ValidatorPeer to the public ValidatorV2 service)
- **Proto + conversions** — `validator_v2.proto` with full proto ↔ domain round-trip conversions
- **Refactored authority server** — replaced `handle_with_decoration!` macro with typed `pre_handle`/`post_handle_unary`/`post_handle_stream` methods

### Key design decisions

- White-flag-only flow: no certificate APIs referenced
- Per-tx errors return `TxStatusUpdate::Rejected` (TODOs added for remaining `Err(Status)` inconsistencies)
- Traffic weight is per-request not per-tx (TODO to scale with batch size)

Closes #10974.

### DIFF
--- a/crates/iota-core/src/authority_server/validator_v2.rs
+++ b/crates/iota-core/src/authority_server/validator_v2.rs
@@ -5,7 +5,10 @@ use std::sync::Arc;
 
 use futures::future::Either;
 use iota_network::{
-    api::{GetTxStatusRequest, SubmitTxRequest, TxStatus, ValidatorV2},
+    api::{
+        GetTxStatusRequest, NotifyCapabilitiesRequest, NotifyCapabilitiesResponse, SubmitTxRequest,
+        TxStatus, ValidatorV2,
+    },
     tonic::{Request, Response, Status},
 };
 use iota_types::{
@@ -16,8 +19,9 @@ use iota_types::{
     message_envelope::Message,
     messages_consensus::ConsensusTransaction,
     messages_grpc::{
-        ExecutedData, GetTxStatusRequest as DomainGetTxStatusRequest, SubmitTransactionsRequest,
-        TxStatusUpdate,
+        ExecutedData, GetTxStatusRequest as DomainGetTxStatusRequest,
+        HandleCapabilityNotificationRequestV1, HandleCapabilityNotificationResponseV1,
+        SubmitTransactionsRequest, TxStatusUpdate,
     },
     traffic_control::Weight,
     transaction::Transaction,
@@ -35,6 +39,8 @@ const GET_TX_STATUS_TIMEOUT_SECS: u64 = 30;
 
 /// A single streamed item in a V2 RPC response.
 type TxUpdateItem = Result<(TransactionDigest, TxStatusUpdate), Status>;
+
+use iota_metrics::spawn_monitored_task;
 
 use crate::{
     authority::{AuthorityState, authority_per_epoch_store::AuthorityPerEpochStore},
@@ -82,7 +88,7 @@ impl ValidatorService {
             let consensus_adapter = consensus_adapter.clone();
             let metrics = metrics.clone();
             let tx_sender = tx_sender.clone();
-            tokio::spawn(async move {
+            spawn_monitored_task!(async move {
                 let tx_digest = *transaction.digest();
                 let result = Self::submit_single_tx(
                     &state,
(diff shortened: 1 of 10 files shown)


crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -1,31 +1,536 @@
...
 
-use crate::authority_server::ValidatorService;
+/// Maximum number of transactions allowed in a single `submit_tx` request.
+const MAX_TRANSACTIONS_PER_SUBMIT: usize = 256;
+
+/// Maximum number of queries allowed in a single `get_tx_status` request.
+const MAX_QUERIES_PER_GET_TX_STATUS: usize = 256;
+
+/// Timeout for waiting on transaction execution in `get_tx_status`.
+const GET_TX_STATUS_TIMEOUT_SECS: u64 = 30;
```
bingyanglin, line 38: Any reference for the 256? Also should them be configurable in protocol config?

piotrm50: It doesn't need to be a protocol parameter and each validator can set it locally. Perhaps we can migrate it to local node config. Each transaction in the batch is handled individually.

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -1,31 +1,536 @@
...
+        // This matches V1's handle_wait_for_effect_impl behavior. Client retry
+        // after timeout gets a fresh epoch store. Consider listening for epoch
+        // change to return Expired early if cross-epoch awareness is needed.
+        let (tx_sender, rx) = tokio::sync::mpsc::channel(request.queries.len().max(1));
+
+        for query in request.queries {
+            let state = state.clone();
+            let epoch_store = epoch_store.clone();
+            let tx_sender = tx_sender.clone();
+            tokio::spawn(async move {
```
piotrm50, line 274: why not a `spawn_monitored_task`?

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -1,31 +1,536 @@
...
+            ))
+        );
+
+        // TODO(#11111): epoch_store is captured once and used for the full 30s wait.
+        // If an epoch change occurs mid-wait, notify_read_dropped_digests
+        // watches the old epoch's cache and the timeout reports a stale epoch.
+        // This matches V1's handle_wait_for_effect_impl behavior. Client retry
+        // after timeout gets a fresh epoch store. Consider listening for epoch
+        // change to return Expired early if cross-epoch awareness is needed.
+        let (tx_sender, rx) = tokio::sync::mpsc::channel(request.queries.len().max(1));
```
piotrm50, line 268: epoch_store has `within_alive_epoch` method that returns if the epoch ends, so we don't need to wait for a timeout.

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -1,31 +1,536 @@
...
+        state: &Arc<AuthorityState>,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+        tx_digest: TransactionDigest,
+        include_details: bool,
+    ) -> TxStatusUpdate {
+        // Fast path: already executed.
+        let cache = state.get_transaction_cache_reader();
+        match cache.try_get_executed_effects(&tx_digest) {
+            Ok(Some(effects)) => {
+                return Self::build_executed_update(state, effects, include_details);
```
piotrm50, line 301: worth having a separate method just for use here?

crates/iota-core/src/authority_server/validator_v2.rs
filipdulic: fixed in later pr

crates/iota-core/src/authority_server/validator_v2.rs
filipdulic: is later (subsequent PR)refactored to be used by both submit and get

crates/iota-core/src/authority_server/validator_v2.rs
filipdulic: will take a note of this and fix in subsequent PR.

piotrm50 (approved):

vekkiokonio (approved):

PR #11002, opened 2026-03-27
fix(ts-sdk): Avoid using pnpm in build scripts, use tsc directly

marc2332:
A jump of faith.

Hopefully fixes https://github.com/iotaledger/iota/actions/runs/23657260477/job/68918377469, the theory is that concurrent build scripts might be breaking pnpm sometimes.


brancoder (approved):
LGTM 🤞

begonaalvarezd (approved):

PR #11004, opened 2026-03-29
feat: replace `GenesisTransaction` with SDK version

### LINKED ISSUE #586: Replace `GenesisTransaction` with SDK version

thibault-martinez:
As title suggests
Closes https://github.com/iotaledger/iota-rust-sdk/issues/586


crates/iota-types/src/transaction.rs
```diff
@@ -18,8 +18,8 @@ use iota_protocol_config::ProtocolConfig;
 pub use iota_sdk_types::{
     Argument, AuthenticatorStateExpire, AuthenticatorStateUpdateV1, ChangeEpoch, ChangeEpochV2,
     ChangeEpochV3, ChangeEpochV4, Command, EndOfEpochTransactionKind, GasPayment as GasData,
-    GenesisObject, MakeMoveVector, MergeCoins, MoveCall as ProgrammableMoveCall,
-    ProgrammableTransaction, Publish, RandomnessStateUpdate,
+    GenesisObject, GenesisTransaction, MakeMoveVector, MergeCoins,
+    MoveCall as ProgrammableMoveCall, ProgrammableTransaction, Publish, RandomnessStateUpdate,
```
Thoralf-M, line 22: Btw there is no need for `MoveCall as ProgrammableMoveCall` anymore?

Thoralf-M (approved):

crates/iota-types/src/transaction.rs
thibault-martinez: Why not?

Thoralf-M: There is no other MoveCall type here?

thibault-martinez: I'm not exactly sure what you are asking. You want to remove the `as ProgrammableMoveCall`?

Thoralf-M: Yes unless there is a good reason to keep it like that

thibault-martinez: We've been importing names with a rename to match the monorepo type name to avoid too many unnecessary changes

thibault-martinez: I'd rather not do it but I remember some TLs asking us to do that at the beginning

Alex6323 (approved):

Issue #10746, opened 2026-03-17
pcool: Separate ping endpoint from batch submission


piotrm50:
Pings are currently encoded as an empty `SubmitTransactionsRequest` (`is_ping()` = `transactions.is_empty()`). This conflates health checks with transaction submission and makes `SubmitTransactionsRequest` semantics ambiguous.

Design and implement a clean separation between ping/health-check and transaction submission.

Depends on: #10719

**Files:** `authority_server.rs`, `messages_grpc.rs`, `build.rs` (iota-network)

semenov-vladyslav:
Ping requests have a number of drawbacks:
- add unnecessary CPU and network overhead and code complexity
- do not properly estimate transaction finalization latency since they contain no transactions to execute
- health-check requests estimate RTT latency more accurately making ping requests redundant

Thus, the decision was made to remove ping requests entirely.

Ping requests were meant to make transaction finalization latency estimates more accurate.
To do so we use more advanced methods instead:
- use time-delayed log-exponential weighted moving average latency estimator
- use exploitation-exploration scoring strategy; the score is based on the latency estimates, failures frequency, effective sample size, recency and frequency of observations

PR #11007, opened 2026-03-30
feat(iota-core): improve ValidatorClientMonitor scoring function

### LINKED ISSUE #10371: pcool: Design and implement proper ping-request handling in TransactionDriver
Design robust connectivity measurement logic for TransactionDriver to assess connectivity to all validators. Refactor existing simple ping-pong request and remove unnecessary ping logic from transaction submission.

semenov-vladyslav:
# Description of change

This PR improves validator performance scoring by fullnodes. The main changes include:
- properly handle failed Consensus operation;
- simplify and refactor ValidatorClientMonitor and its internal components;
- change/refactor internal metrics/stats to use time-decayed EWMA of log-latency;
- a validator score includes:
  - Exploitation -- the performance score;
  - Exploration -- the score allowing to consider new or outdated validators;
  - the scores are based on latency and its variance estimate, effective sample size and number of failures, frequency of observations and recency of the last observation;
- updated the selection function:
  - choose top performing validators with lowest Exploitation score;
  - choose validators with lowest Exploration score;
  - mix them together;
  - return the rest of the validators in order of combined score (Exploitation + Exploration);
- added tests;
- removed ping request:
  - it adds a pretty heave useless network and validator load;
  - its only purpose was to add more steady stream of Submit/Effects/Consensus operations (and corresponding feedbacks); however we already have a steady stream of HealthCheck feedbacks, and byzantine validator would be able to fake Submit/Effects/Consensus anyway; plus improved scoring mechanism accounts for less observations / more variance;

Fixes #10371, #10746.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): improvements in ValidatorClientMonitor scoring mechanism
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:


piotrm50:
needs to be rebased

roman1e2f5p8s:
typo in the PR title? "funciton"

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,450 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{BTreeMap, HashMap, HashSet, btree_map::Entry},
-    time::Duration,
+    collections::{HashMap, HashSet},
+    time::Instant,
 };
 
-use iota_common::moving_window::MovingWindow;
```
bingyanglin, line 10: Maybe remove the `crates/iota-common/src/moving_window.rs` if it is unused now?

crates/iota-core/src/transaction_driver/mod.rs
```diff
@@ -320,24 +311,25 @@ where
...
                 name,
                 submit_txn_result,
                 options,
             )
             .await;
 
+        let feedback_builder = OperationFeedback::builder(
+            name,
+            auth_agg.get_display_name(&name),
+            OperationType::Consensus,
```
bingyanglin, line 329: The `OperationType::Consensus` feedback is recorded against the validator that submitted the transaction, but consensus latency depends on the whole quorum. Maybe add a comment to clarify that this measures "end-to-end consensus latency via this validator" rather than that validator's individual consensus performance?

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,450 @@
...
-/// very slow validators.
-const MAX_LATENCY: Duration = Duration::from_secs(10);
+/// Ok(latency in sec) or Err(high failure latency in sec)
+type Observation = Result<f64, f64>;
+
+/// Exponential weighted moving average (EWMA) estimator.
+///
+/// Each new observation is weighted by α and the prior estimate by (1 − α):
+/// μ_t ​= α x_t​ + (1−α) μ_{t−1}​
+/// σ_t^2​ = (1−α) (σ_{t−1}^2 ​+ α (x_t​−μ_{t−1}​)^2)
```
bingyanglin: Is the formula correct?  In the implementation at line 60 we do `(1−α)σ² + α(x−μ)²`? the tailing `)` is wrong?

crates/iota-core/src/validator_client_monitor/metrics.rs
```diff
@@ -12,21 +12,24 @@ use prometheus::{
...
-    pub performance: GaugeVec,
+    /// Current performance score per validator. It is based
+    /// on the average latency, risk, staleness over operation types.
+    pub(super) performance: GaugeVec,
+
+    /// Current exploration score per validator. It is based
+    /// on the effective sample size.
+    pub(super) exploration: GaugeVec,
 
     /// Number of low latency validators that got shuffled.
```
piotrm50: Comment and metric description is outdated.

crates/iota-core/src/transaction_driver/request_retrier.rs
```diff
@@ -39,41 +36,45 @@ const SELECT_LATENCY_DELTA: f64 = 0.02;
...
-                // `auth_agg.authority_clients` if those are coming from the list
-                // of `allowed_validators`, as the provided `auth_agg` might have been updated
-                // with a new committee that doesn't contain the validator in question.
+            .map(|(name, _display_name)| name);
+
+        let mut ranked_clients = client_monitor
+            .select_shuffled_preferred_validators(committee)
+            .into_iter()
+            .filter_map(|name| {
+                // There is a guarantee that the `name` is in the
```
piotrm50, line 68: If there is a guarantee then `filter_map` silently hides inconsistency between those two structs.

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,450 @@
...
-    pub fn update_average_latency(&mut self, operation: OperationType, new_latency: Duration) {
-        match self.average_latencies.entry(operation) {
-            Entry::Occupied(mut entry) => {
-                entry.get_mut().add_value(new_latency);
+    fn record_interaction_result(
+        &mut self,
+        config: &ValidatorClientMonitorConfig,
+        feedback: &OperationFeedback,
+    ) {
+        const TAU: f64 = 60.0;
```
piotrm50: Duplicated const here and in line R290

crates/iota-core/src/validator_client_monitor/monitor.rs
```diff
@@ -1,25 +1,21 @@
 // Copyright (c) Mysten Labs, Inc.
-// Modifications Copyright (c) 2024 IOTA Stiftung
+// Modifications Copyright (c) 2025 IOTA Stiftung
```
piotrm50:
```suggestion
// Modifications Copyright (c) 2026 IOTA Stiftung
```

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,450 @@
...
+        const VARIANCE_PENALTY_EFF: f64 = 3.0;
+        const VARIANCE_PENALTY_CON: f64 = 4.0;
+        const EMPTY_SCORE_HC: f64 = 0.5;
+        const EMPTY_SCORE_SUB: f64 = 2.0;
+        const EMPTY_SCORE_EFF: f64 = 3.0;
+        const EMPTY_SCORE_CON: f64 = 4.0;
+        const FAILURE_COEFF: f64 = 10.0;
+        let tau_hc = config.health_check_interval.as_secs_f64();
+
+        let (score_hc, _sample_size_hc, failures_hc, _alpha_hc, _interval_hc) = self
```
piotrm50, line 341: why are we discarding part of values for healthcheck?

crates/iota-core/src/validator_client_monitor/mod.rs
```diff
@@ -45,8 +45,68 @@ pub struct OperationFeedback {
     pub display_name: String,
     /// The operation type
     pub operation: OperationType,
-    /// The ping type. If it's not a ping request, then this is None.
-    pub ping: bool,
     /// Result of the operation: Ok(latency) if successful, Err(()) if failed.
     pub result: Result<Duration, ()>,
+    /// The timestamp when the operation feedback was received.
```
piotrm50:
```suggestion
    /// The timestamp when the operation feedback was observed.
```

crates/iota-core/src/transaction_driver/mod.rs
semenov-vladyslav: Right, this is a sensitive and imprecise topic. The `name` returned by `submit_transaction` may even be not the one that was used to actually disseminate the transaction. For example, valA and valB are the two validators selected by `submit_transaction`. A malicious valA will respond with a success but will not propagate the request further (essentially, the request should have failed but the fullnode has no way of knowing that) and valA's name will be returned, while the honest valB will take more time to disseminate the transaction and reply. So valA will get its stats/score updated even though it was valB who did the job.

crates/iota-core/src/validator_client_monitor/stats.rs
semenov-vladyslav: Both formulas are used. The one that is implemented is a "standard" one, while the one in this comment can also be used, but I can't justify its use here.

crates/iota-core/src/validator_client_monitor/metrics.rs
semenov-vladyslav: Right, removed it entirely, it's hard to track the number of validators that are actually used in submit and effects operation.

crates/iota-core/src/transaction_driver/request_retrier.rs
semenov-vladyslav: Sorry, which inconsistency between which two structs? Do you mean that there should be `map()` instead of `filter_map()` and `unwrap()` instead of `get()`?

crates/iota-core/src/validator_client_monitor/stats.rs
semenov-vladyslav: Because the health-check operation is very steady/regular. In case of failures (timeouts) we will observe high failures_hc values. The discarded stats are ignore so as not to interfere with the corresponding stats of other operations which might be more sensitive.

crates/iota-core/src/transaction_driver/request_retrier.rs
piotrm50: I mean that if the guarantee is broken, then this code returns None and filter_map silently hides the error. We could have at least a log to indicate the problem if it ever happens.

```
auth_agg
                    .authority_clients
                    .get(name)
                    .map(|client| (*name, client.clone()))
```

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,449 @@
...
-    pub latency_moving_window_size: usize,
+struct ValidatorClientStats {
+    /// Health-check operation stats.
+    stats_hc: LogLatencyEwma,
+    /// Submit operation stats.
+    stats_sub: LogLatencyEwma,
+    /// Effects operation stats.
+    stats_eff: LogLatencyEwma,
+    /// Consensus operation stats.
+    stats_con: LogLatencyEwma,
```
muXxer: I think the abbreviations make the code less readable.
```suggestion
    /// Health-check operation stats.
    stats_health_check: LogLatencyEwma,
    /// Submit operation stats.
    stats_submit: LogLatencyEwma,
    /// Effects operation stats.
    stats_effects: LogLatencyEwma,
    /// Consensus operation stats.
    stats_consensus: LogLatencyEwma,
```

crates/iota-core/src/validator_client_monitor/stats.rs
```diff
@@ -1,148 +1,449 @@
...
+        const MAX_INTERVAL: f64 = 3600.0;
+        const VARIANCE_PENALTY_HC: f64 = 0.5;
+        const VARIANCE_PENALTY_SUB: f64 = 2.0;
+        const VARIANCE_PENALTY_EFF: f64 = 3.0;
+        const VARIANCE_PENALTY_CON: f64 = 4.0;
+        const EMPTY_SCORE_HC: f64 = 0.5;
+        const EMPTY_SCORE_SUB: f64 = 2.0;
+        const EMPTY_SCORE_EFF: f64 = 3.0;
+        const EMPTY_SCORE_CON: f64 = 4.0;
+        const FAILURE_COEFF: f64 = 10.0;
```
muXxer: Would it make sense to collect the constants in one place in the file and add comments on what the const is used for? Maybe also a reasoning for the chosen values?

crates/iota-core/src/validator_client_monitor/monitor.rs
```diff
@@ -31,239 +27,152 @@ use crate::{
...
+        // into the global default registry.
         Self::new(
             ValidatorClientMonitorConfig::default(),
-            Arc::new(ValidatorClientMetrics::new(&Registry::default())),
-            Arc::new(ArcSwap::new(authority_aggregator)),
+            Arc::new(ValidatorClientMetrics::new(&prometheus::Registry::new())),
         )
     }
 
     /// Background task that runs periodic health checks on all validators.
```
muXxer: That comment should be moved to `run_health_checks` where it was before.

crates/iota-core/src/validator_client_monitor/mod.rs
muXxer, line 42: Can this be `pub(crate)` instead?

crates/iota-core/src/validator_client_monitor/mod.rs
```diff
@@ -45,8 +46,64 @@ pub struct OperationFeedback {
...
+        OperationFeedbackBuilder {
+            authority_name,
+            display_name,
+            operation,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct OperationFeedbackBuilder {
```
muXxer, line 70: Can this be `pub(crate)` instead?

muXxer (changes requested):
LGTM so far, just a small nit.

Also, can you add tests for the decaying of values? I didn't see that.
Can we also test epoch boundaries in case validators change?

crates/iota-core/src/validator_client_monitor/mod.rs
semenov-vladyslav: The whole `validator_client_monitor` modules is only used in `iota-core` crate, so its entire contents can be made `pub(crate)` or just the module declaration in `lib.rs`. However, `OperationFeedback` is a part of "public API" of this validator client monitor components and is used outside of `validator_client_monitor` module. So, I would keep it `pub`.

crates/iota-core/src/validator_client_monitor/mod.rs
semenov-vladyslav: Same as above.

crates/iota-core/src/transaction_driver/request_retrier.rs
semenov-vladyslav: Fixed, `map` and `expect` are used instead now.

piotrm50 (approved):

muXxer (approved):

PR #11010, opened 2026-03-30
chore: bump rust toolchain to `1.93`

thibault-martinez:
# Description of change

Bump rust toolchain to `1.93`.
Addresses the https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#unnecessary_unwrap lint.

The Rust 1.93 toolchain bump causes `garbage_inputs` property tests in `bytecode-verifier-prop-tests` and `serializer-tests` to overflow the default 2 MiB test-thread stack. The tests use proptest to generate arbitrary `CompiledModule` values with deeply-nested recursive `SignatureToken` types (`Vector`, `Reference`, `MutableReference` up to 8 levels deep). Rust 1.93 produces slightly larger stack frames in debug builds, which is enough to tip these already-borderline tests over the limit.

The fix runs both `garbage_inputs` tests on a dedicated thread with an 8 MiB stack via `std::thread::Builder::new().stack_size(...)`. This requires moving them out of the `proptest!` macro block (which expands to `#[test]` functions that run on the default test-runner thread with no way to control its stack size) and using the `proptest!()` function-call form instead. Generation depth and case count remain unchanged, so test coverage is preserved.


marc2332 (approved):

alexsporn (approved):
LGTM

Thoralf-M (approved):

crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
```diff
@@ -399,21 +399,20 @@ impl GasPriceFeedbackTester {
...
-            } else if data.2.is_some() && data.3.is_none() {
-                self.build_access_one_counter_transaction(gas_data, data.2.unwrap(), true)
-                    .await
-            } else if data.2.is_none() && data.3.is_some() {
-                self.build_access_one_counter_transaction(gas_data, data.3.unwrap(), false)
-                    .await
-            } else {
-                panic!("At least one counter must be accessed in transactions.");
+            let transaction = match (data.2, data.3) {
+                (Some(data_2), Some(data_3)) => {
```
Alex6323: I don't like `data_2` and `data_3` tbh. Let's use names that read better?

Alex6323 (approved):

crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
Alex6323: changed my mind

piotrm50 (approved):

kodemartin (approved):

Dr-Electron (approved):

miker83z (approved):

crates/iota-core/src/unit_tests/gas_price_feedback_tests.rs
thibault-martinez: that was a fair remark tbh, will do

PR #11026, opened 2026-03-30
upstream(core): Re-enqueue pending consensus transactions at startup

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@c2e769e](https://github.com/MystenLabs/sui/commit/c2e769e6e8370cfebae374f0fec636495df82a0a)
  - [MystenLabs/sui@0f0063c](https://github.com/MystenLabs/sui/commit/0f0063cb80dddcbee3bd4fe649055486c5756de1)
  - [MystenLabs/sui@68113e1](https://github.com/MystenLabs/sui/commit/68113e14ae371146856a01614c89cf30e3ee0589)
- Description:

**Commit 1 (c2e769e):** Do not call `reexecute_pending_consensus_certs` at startup.

All transactions in `pending_consensus_transactions` will eventually be executed once they come out of consensus. There is no strong need to proactively execute them upfront. This causes the `debug_fatal` in the function to be triggered, causing both simtest nightly and antithesis to fail.

**Commit 2 (0f0063c):** Re-execute all pending consensus transactions at startup.

Partially reverts the previous change because later on the code will also attempt to rebuild all checkpoints, which depends on all pending transactions being executed. Antithesis starts to report failures after the previous change. Key differences from the original implementation:
1. Instead of just re-scheduling transactions with locally signed effects, we schedule all fast-path pending consensus transactions.
2. We no longer wait here. Instead, when it rebuilds all checkpoints, it will wait anyway.

**Commit 3 (68113e1):** Fix `reexecute_pending_consensus_transactions` and address TODOs.

- Track `expected_effects_digest` for fork detection on signed transactions
- Split pending certs into those with signed effects (re-executed with expected digest) and additional certs (enqueued without)
- Run re-execution concurrently with `construct_validator_components`
- Wait for re-executed transactions to complete with timeout and `debug_fatal` on failure
- Address comment TODOs in checkpoints/mod.rs

Part of #10909

### DIFF
--- a/crates/iota-node/src/lib.rs
+++ b/crates/iota-node/src/lib.rs
@@ -25,6 +25,7 @@ use fastcrypto_zkp::bn254::zk_login::{JWK, JwkId, OIDCProvider};
 use futures::future::BoxFuture;
 pub use handle::IotaNodeHandle;
 use iota_archival::{reader::ArchiveReaderBalancer, writer::ArchiveWriter};
+use iota_common::debug_fatal;
 use iota_config::{
     ConsensusConfig, NodeConfig,
     node::{DBCheckpointConfig, RunWithRange},
@@ -114,6 +115,7 @@ use iota_types::{
     crypto::{AuthoritySignature, IotaAuthoritySignature, KeypairTraits, RandomnessRound},
     digests::ChainIdentifier,
     error::{IotaError, IotaResult},
+    executable_transaction::VerifiedExecutableTransaction,
     execution_config_utils::to_binary_config,
     full_checkpoint_content::CheckpointData,
     iota_system_state::{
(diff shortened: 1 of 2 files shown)


crates/iota-node/src/lib.rs
```diff
@@ -1608,113 +1604,41 @@ impl IotaNode {
...
-    ///    consensus transactions, or vice versa), we must re-execute pending
-    ///    consensus transactions to ensure that both processes can complete.
-    /// 3. Also note that for any pending consensus transactions for which we
-    ///    wrote a signed effects digest to disk, we must re-execute using that
-    ///    digest as the expected effects digest, to ensure that we cannot
-    ///    arrive at different effects than what we previously signed.
-    async fn reexecute_pending_consensus_certs(
+    /// Re-enqueue pending consensus certificates for execution.
+    // TODO: Add comments explain why we need to do this at start up.
+    async fn reenqueue_pending_consensus_certs(
```
bingyanglin: Note that Sui later have:
1. Enhanced it in [68113e14ae](https://github.com/MystenLabs/sui/commit/68113e14ae) with expected_effects_digest tracking (important for fork detection on signed transactions)
2. Removed it entirely in [10436c976a](https://github.com/MystenLabs/sui/commit/10436c976a) after removing preconsensus locking

Should we port the enhanced version instead, or skip this entirely if we're planning to remove preconsensus locking?

jkrvivian: I'd port the 1. commit you mentioned here.

semenov-vladyslav (approved):
We could have ignored all these three commits, the initial logic didn't change much except that a few errors now cause panics.

bingyanglin (approved):

piotrm50 (approved):

jkrvivian:
> We could have ignored all these three commits, the initial logic didn't change much except that a few errors now cause panics.

Agreed, after all the upstream modifications, it went back to the current behavior, and just made the fallible functions to non-fallible, I'll close this PR and keep the existing code as-is.

PR #11027, opened 2026-03-30
feat(explorer): show transaction type label on address page

evavirseda:
image


apps/core/src/utils/transaction/getTransactionAction.ts
KeitoTadashi, line 59: I made a pubish TX in devnet

Digest: `9oprWQ4rRSNbkPUDs6waFsU6edBFM9HYinvMbvaHTRCD`

and i expected it to be type [Transaction](https://github.com/iotaledger/iota/pull/11027/changes#diff-c3e18a6942c0d686d005712893c829895af716da8d135b8dd836975db2f1df8eR44)

but instead is showing "Sent"

imagen

KeitoTadashi (changes requested):
left a comment, rest LGTM

apps/core/src/utils/transaction/isUnlockTimelockedObjectTransaction.ts
```diff
@@ -21,6 +21,22 @@ export function isUnlockTimelockedObjectTransaction(
     return isUnlockTimelockedObject;
 }
 
+// Detects a collect all vesting transaction
+export function isCollectAllTimelocksTransaction(
```
begonaalvarezd: I would probably make it compatible retroactively, so all the collects that happened until now should also be considered here. I would rename this to something like 
```suggestion
export function isUnlockTimelockedObjectTransaction(
```

apps/core/src/utils/transaction/isUnlockTimelockedObjectTransaction.ts
```diff
@@ -21,6 +21,22 @@ export function isUnlockTimelockedObjectTransaction(
...
+): boolean {
+    if (!transaction || transaction.data.transaction.kind !== 'ProgrammableTransaction')
+        return false;
+    const moveCallTxs = transaction.data.transaction.transactions
+        .filter(isMoveCall)
+        .filter((tx) => tx.MoveCall.module === TIMELOCK_MODULE);
+    // At least one call to unlock_with_clock
+    const isCollectAll =
+        moveCallTxs.length > 0 &&
+        moveCallTxs.some((tx) => tx.MoveCall.function === 'unlock_with_clock');
```
begonaalvarezd: I would add here the `unlock` function too

apps/explorer/src/lib/ui/utils/generateTransactionsTableColumns.tsx
```diff
@@ -17,6 +23,19 @@ import {
 } from '@iota/iota-sdk/utils';
 import { getElapsedTime } from '~/pages/epochs/utils';
 
+const ACTION_LABELS: Record<TransactionAction, string> = {
```
begonaalvarezd: Dont we have these labels already in the wallet? Cant we unify these for all apps?

apps/core/src/utils/transaction/isUnlockTimelockedObjectTransaction.ts
begonaalvarezd: We should use the oportunity to improve this filter here too and check against the right package too, looking only at the module is very generic

apps/core/src/utils/transaction/isUnlockTimelockedObjectTransaction.ts
```diff
@@ -21,6 +21,22 @@ export function isUnlockTimelockedObjectTransaction(
...
 
+// Detects a collect all vesting transaction
+export function isCollectAllTimelocksTransaction(
+    transaction: IotaTransactionBlockResponse['transaction'],
+): boolean {
+    if (!transaction || transaction.data.transaction.kind !== 'ProgrammableTransaction')
+        return false;
+    const moveCallTxs = transaction.data.transaction.transactions
+        .filter(isMoveCall)
+        .filter((tx) => tx.MoveCall.module === TIMELOCK_MODULE);
```
begonaalvarezd: We should check against the right package too, looking only at the module is very generic

begonaalvarezd (changes requested):

apps/core/src/utils/transaction/isUnlockTimelockedObjectTransaction.ts
begonaalvarezd: And we can extend it to look at the timelocked_staking module too and then we detect re-stakes too

KeitoTadashi (approved):

apps/core/src/components/transaction/TransactionIcon.tsx
```diff
@@ -32,18 +38,26 @@ const icons = {
     [TransactionAction.TimelockedCollect]: <Vesting />,
 };
 
+const ICON_SIZES: Record<TransactionIconSize, string> = {
+    [TransactionIconSize.Small]: '[&_svg]:h-4 [&_svg]:w-4',
+    [TransactionIconSize.Medium]: '[&_svg]:h-5 [&_svg]:w-5',
+    [TransactionIconSize.Large]: '[&_svg]:h-6 [&_svg]:w-6',
```
VmMad:
```suggestion
    [TransactionIconSize.Small]: '[&_svg]:size-4',
    [TransactionIconSize.Medium]: '[&_svg]:size-5',
    [TransactionIconSize.Large]: '[&_svg]:size-6',
```

VmMad (approved):
just a small nit, but the list looks better now

evavirseda:
Closed in a favor of: https://github.com/iotaledger/ts-packages/pull/100

PR #11044, opened 2026-03-31
upstream(core): Upgrade rocksdb to 0.24.0

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@28dc3](https://github.com/MystenLabs/sui/commit/28dc33fc8fc43e50819c42c22b0d557b889c107e)
- Description:

The upstream commit upgrades rocksdb from 0.21.0 to 0.22.0. In this port, we jumped directly to rocksdb 0.24.0 (the latest available version).

The previous upgrade attempt triggered write stalling and stopping on archival full nodes. Since write stalling has been disabled on full node indexes and perpetual tables (via `disable_write_throttling()`), it should no longer be an issue.

The `bindgen-runtime` feature is explicitly enabled for `rocksdb` so that `clang-sys` uses `dlopen` with its own platform-aware search logic (which knows macOS Command Line Tools and Xcode paths) instead of relying on `dyld` `@rpath` resolution, which fails on macOS due to SIP stripping `DYLD_*` environment variables.

Part of #10909

### Release Notes

- [x] Protocol:
  - Bump RocksDB version to 0.24.0
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1820,21 +1820,18 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.65.1"
+version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfdf7b466f9a4903edc73f95d6d2bcd5baf8ae620638762244d3f60143643cc5"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
- "bitflags 1.3.2",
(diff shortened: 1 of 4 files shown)


bingyanglin (approved):

alexsporn:
That macOS quirk is not good. We should not have weird quirks required in the setup just to be able to build.

semenov-vladyslav (approved):

PR #11050, opened 2026-03-31
upstream(core): Cleanup serde

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@f17a550](https://github.com/MystenLabs/sui/commit/f17a550d0471616e0b1db1fd68a9ca5b6478d675)
- Description:

Improve RoaringBitmap ser/de. And a few minor cleanups.

Part of #10909

### DIFF
--- a/crates/iota-types/src/crypto.rs
+++ b/crates/iota-types/src/crypto.rs
@@ -1318,13 +1318,7 @@ impl<const STRONG_THRESHOLD: bool> AuthoritySignInfoTrait
             .get_mut(message_index)
             .ok_or(IotaError::InvalidAuthenticator)?;
 
-        let mut seen = std::collections::BTreeSet::new();
         for authority_index in self.signers_map.iter() {
-            if !seen.insert(authority_index) {
-                continue;
-            }
-
-            // Update weight when seeing the authority for the first time.
             let authority = committee
                 .authority_by_index(authority_index)
                 .ok_or_else(|| IotaError::UnknownSigner {
--- a/crates/iota-types/src/crypto.rs
+++ b/crates/iota-types/src/crypto.rs
@@ -1318,7 +1318,13 @@ impl<const STRONG_THRESHOLD: bool> AuthoritySignInfoTrait
             .get_mut(message_index)
             .ok_or(IotaError::InvalidAuthenticator)?;
 
+        let mut seen = std::collections::BTreeSet::new();
         for authority_index in self.signers_map.iter() {
+            if !seen.insert(authority_index) {
+                continue;
+            }
+
+            // Update weight when seeing the authority for the first time.
             let authority = committee
                 .authority_by_index(authority_index)
                 .ok_or_else(|| IotaError::UnknownSigner {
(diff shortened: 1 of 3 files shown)


alexsporn:
This is probably getting replaced with the SDK anyways. @thibault-martinez can you confirm? Maybe some of the improvements are also relevant for the SDK types.

thibault-martinez:
> This is probably getting replaced with the SDK anyways. @thibault-martinez can you confirm? Maybe some of the improvements are also relevant for the SDK types.

`AccountAddress` is a move core type that we can't replace so I guess this is still relevant.
We do have the exact same `RoaringBitmap` serde impl in the SDK so maybe when time comes we can remove this one.

bingyanglin (approved):

crates/iota-types/src/iota_serde.rs
```diff
@@ -369,3 +313,84 @@ impl<'de> DeserializeAs<'de, ProtocolVersion> for AsProtocolVersion {
...
+    // Ensure there are no duplicated entries in the bitmap.
+    let mut seen = std::collections::BTreeSet::new();
+    let mut new_bitmap = roaring::RoaringBitmap::new();
+    for v in orig_bitmap.iter() {
+        if seen.insert(v) {
+            new_bitmap.insert(v);
+        }
+    }
+    Ok(new_bitmap)
+}
```
semenov-vladyslav: I don't see the point in this, roaring-0.11+ already deduplicates entries. "defense-in-depth" argument is weak, we rely on a number of crates doing various verifications (including cryptographic) without any defense-in-depth. It's cheap but it is executed for every CertifiedTransaction.

crates/iota-types/src/crypto.rs
```diff
@@ -1318,16 +1318,20 @@ impl<const STRONG_THRESHOLD: bool> AuthoritySignInfoTrait
             .get_mut(message_index)
             .ok_or(IotaError::InvalidAuthenticator)?;
 
+        let mut seen = std::collections::BTreeSet::new();
         for authority_index in self.signers_map.iter() {
+            if !seen.insert(authority_index) {
+                continue;
+            }
```
semenov-vladyslav: `signers_map` has type `RoaringBitmap` which guarantees that the entries are unique. Or am I missing something?

crates/iota-types/src/crypto.rs
```diff
@@ -1318,16 +1318,20 @@ impl<const STRONG_THRESHOLD: bool> AuthoritySignInfoTrait
...
+
+            // Update weight when seeing the authority for the first time.
```
semenov-vladyslav: What does it mean "for the first time"? `authority_index` is unique, and authorities in a committee are unique. There can't be a second time.

semenov-vladyslav (commented):
The changes add unnecessary invariant verification logic that should be enforced statically at compile time, not at run-time.

If there is a concern about updating to a compromised version of a crate (and there is such a concern), there should be other means of verification that the new version is safe. We have a lot of deps (including cryptographic ones) which we can't defend-in-depth.

semenov-vladyslav (approved):

PR #10912, opened 2026-03-24
feat(wallet-dashboard): collect vesting ptb

evavirseda:
# Description of change

example of collect [transaction](https://explorer.iota.org/txblock/HjKDYUXLjAnS9DGNK6Cqu2RrT7jcQwBWsS4zLMBuqKxV?network=devnet)

- Regular timelocks → unlock → convert to coin → transfer
- Timelocked stakes → unlock first, then group by `(pool_id, stake_activation_epoch)` and merge with `staking_pool::join_staked_iota`
- Unlocks before joining to avoid `timelocked_staking::join_staked_iota`, which also checks `expiration_timestamp_ms` and would abort for stakes from different vesting periods
- If the user already has a regular stake in the same pool/epoch, merges into it instead of creating a new object
-  Shows a summary dialog after a successful collect

## Links to any relevant issues
:warning: nested PR from https://github.com/iotaledger/iota/pull/10923

fixes https://github.com/iotaledger/iota/issues/10800

## How the change has been tested

1. Get a [vesting scenario](https://github.com/iotaledger/tooling-resources/pull/101) 
2. Change `SUPPLY_INCREASE_STARTING_VESTING_YEAR` from `2023` to `2024` because the testing objects that you created expire in `2026`, but the logic only considers staker up to `2025`
3. Wait a few minutes until all are unlocked

## Testing Collect Transaction with Existing Stake Objects

To reproduce this scenario:
- Run the [vesting scenario](https://github.com/iotaledger/tooling-resources/pull/101)  from the tooling-resources repository.
- In the same epoch, add regular stakes using the dashboard or wallet.
 ❗  Make sure that the validator is the same for timelocked staking (vesting fast example) and regular stake 

This test ensures that the collect transaction correctly unlocks timelocked stake and adds it to the existing staking object in the pool

PR #11058, opened 2026-03-31
fix(wallet): approval popup crash on network mismatch

evavirseda:
context [here](https://github.com/iotaledger/iota/pull/10912#issuecomment-4153676538)  

When the dashboard is on `testnet` and the wallet is on `devnet`, 
the transaction approval popup crashed because `useTransactionData` 
was trying to look up testnet objects on devnet — and those don't exist there.

### Changes
- **Removed** `useTransactionData` from the approval flow — it was redundant 
  and caused the crash by ignoring the dapp's network (`chain` param)
- **Switched** to `useTransactionDryRun` — already supports the `chain` parameter 
  and provides the same loading/error states
- **Used** `transaction.getData()` for the details tab — no RPC call needed, 
  data is already available in the transaction object


apps/wallet/src/ui/app/pages/approval-request/transaction-request/transaction-details/index.tsx
```diff
@@ -47,18 +48,19 @@ const DETAILS_CATEGORIES = [
...
+
+    const transactionData = transaction.getData();
     const {
-        data: transactionData,
+        data: dryRunData,
         isPending,
         isError,
         error,
-    } = useTransactionData(sender, transaction);
+    } = useTransactionDryRun(sender, transaction);
```
brancoder: we should add chain to `useTransactionDryRun(sender, transaction, chain);`

brancoder (changes requested):
we could update the  `useTransactionDryRun` queryKey to contain sender and chain: `queryKey: ['dryRunTransaction', sender, chain, transaction.getData()],` wdyt?

marc2332 (approved):

msarcev (approved):

brancoder (approved):

PR #11062, opened 2026-03-31
feat(iota-replay): Enable gas profiling move based transaction authentication

### LINKED ISSUE #11098: [iota-replay]: Support for sponsor move authentication
## Description

iota-replay (exposed through the `iota-tool` binary) just added support for replaying nad profiling transactions where sender is authenticated with MoveAuthetnicators in https://github.com/iotaledger/iota/pull/11062.

With https://github.com/iotaledger/iota/pull/10947 however we also enable gas sponsors to be authenticated via MoveAuthenticator. The changes in transaction processing on the authority side + executor side will have to be replicated in `iota-replay`, since it does local move vm execution.

lzpap:
# Description of change

`iota-tool` provides the transaction replay and profiling utilities, but they currently only can work on PTBs. This PR enables `iota-tool` to create gas profiles for the authentication phase too. 

Local tests. The replay tool currently has only one integration test against the live testnet, that still works after the change.

The `iota-tool` binary must be built with the `tracing` feature enabled to be able to carry out gas profiling.

### DIFF
--- a/crates/iota-replay/src/replay.rs
+++ b/crates/iota-replay/src/replay.rs
@@ -877,7 +877,8 @@ impl LocalExec {
                 transaction_kind.clone(),
                 tx_info.sender,
                 *tx_digest,
-                bcs::to_bytes(tx_info.sender_signed_data.transaction_data()).expect("TransactionData serialization cannot fail"),
+                bcs::to_bytes(tx_info.sender_signed_data.transaction_data())
+                    .expect("TransactionData serialization cannot fail"),
                 &mut None,
             )
         };
@@ -1171,7 +1172,8 @@ impl LocalExec {
                 kind,
                 signer,
                 *executable.digest(),
-                bcs::to_bytes(sender_signed_data.transaction_data()).expect("TransactionData serialization cannot fail"),
+                bcs::to_bytes(sender_signed_data.transaction_data())
+                    .expect("TransactionData serialization cannot fail"),
                 &mut None,
             )
         };
--- a/crates/iota-replay/src/replay.rs
+++ b/crates/iota-replay/src/replay.rs
@@ -876,6 +876,7 @@ impl LocalExec {
                 transaction_kind.clone(),
                 tx_info.sender,
                 *tx_digest,
+                bcs::to_bytes(tx_info.sender_signed_data.transaction_data()).expect("TransactionData serialization cannot fail"),
                 &mut None,
             )
         };
@@ -1169,6 +1170,7 @@ impl LocalExec {
                 kind,
                 signer,
                 *executable.digest(),
+                bcs::to_bytes(sender_signed_data.transaction_data()).expect("TransactionData serialization cannot fail"),
                 &mut None,
             )
         };
--- a/crates/iota-replay/src/replay.rs
+++ b/crates/iota-replay/src/replay.rs
@@ -20,9 +20,16 @@ use iota_protocol_config::{Chain, ProtocolConfig};
 use iota_sdk::{IotaClient, IotaClientBuilder};
 use iota_types::{
     IOTA_DENY_LIST_OBJECT_ID,
+    account_abstraction::{
+        account::AuthenticatorFunctionRefV1Key,
+        authenticator_function::{
+            AuthenticatorFunctionRefForExecution, AuthenticatorFunctionRefV1,
+        },
+    },
     base_types::{ObjectID, ObjectRef, SequenceNumber, VersionNumber},
     committee::EpochId,
     digests::{ObjectDigest, TransactionDigest},
+    dynamic_field::{self, Field},
     error::{ExecutionError, IotaError, IotaResult},
     executable_transaction::VerifiedExecutableTransaction,
     gas::IotaGasStatus,
@@ -30,6 +37,7 @@ use iota_types::{
     inner_temporary_store::InnerTemporaryStore,
     message_envelope::Message,
     metrics::LimitsMetrics,
+    move_authenticator::MoveAuthenticator,
     object::{Data, Object, Owner},
     storage::{
         BackingPackageStore, ChildObjectResolver, ObjectStore, PackageObject, get_module,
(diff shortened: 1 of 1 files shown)


crates/iota-replay/src/replay.rs
```diff
@@ -924,36 +997,115 @@ impl LocalExec {
...
+                .unwrap();
+
+            let auth_input_objects = auth_input_objects
+                .expect("Auth input objects must be present for MoveAuthenticator transactions");
+            let account_object = account_object
+                .expect("Account object must be present for MoveAuthenticator transactions");
+
+            let account_version = match &account_object.object {
+                ObjectReadResultKind::Object(obj) => obj.version(),
+                _ => panic!("Account object is not available"),
```
bingyanglin: Just curious why we panic here but in `execution_engine_execute_with_tx_info_impl` (the same scenario?) we return a proper error like the following?

```rust
let account_version = match &account_object.object {
    ObjectReadResultKind::Object(obj) => obj.version(),
    _ => {
        return Err(ReplayEngineError::GeneralError {
            err: format!("Account object {} is not available", account_object.id()),
        });
    }
};
```

crates/iota-replay/src/replay.rs
```diff
@@ -924,36 +997,115 @@ impl LocalExec {
...
+            };
+
+            let authenticator_function_ref =
+                load_authenticator_function_ref(move_authenticator, account_version, |id| {
+                    store.get_object(id).cloned()
+                })
+                .unwrap();
```
bingyanglin: Is it better to propagate the `ReplayEngineError` (just use `?`), or we panic here?

lzpap: When we fail to load the authenticator function ref, it is an unrecoverable error so unwrap will panic, which is I beleieve the expected behavior, right?

crates/iota-replay/src/replay.rs
lzpap: fixed in [5a59c01](https://github.com/iotaledger/iota/pull/11062/commits/5a59c01c85f5ab2c813c0a7a5e74ad31721f9825)

crates/iota-replay/src/replay.rs
```diff
@@ -770,22 +787,78 @@ impl LocalExec {
...
+                .expect("Auth input objects must be present for MoveAuthenticator transactions");
+            let account_object = account_object
+                .expect("Account object must be present for MoveAuthenticator transactions");
+
+            let account_version = match &account_object.object {
+                ObjectReadResultKind::Object(obj) => obj.version(),
+                _ => {
+                    return Err(ReplayEngineError::GeneralError {
+                        err: format!("Account object {} is not available", account_object.id()),
+                    });
```
Alex6323: Why is this returning an error while in L803 we panic?

crates/iota-replay/src/replay.rs
```diff
@@ -770,22 +787,78 @@ impl LocalExec {
...
+                }
+            };
+
+            let authenticator_function_ref =
+                load_authenticator_function_ref(move_authenticator, account_version, |id| {
+                    self.storage
+                        .live_objects_store
+                        .lock()
+                        .expect("Can't lock")
```
Alex6323: if we're sure that can't happen, then I think the message "Can't lock" is not adding enough context, and you can just call `.unwrap()` here instead, or write more details what situation might have occured that it wasn't able to get the lock?

crates/iota-replay/src/replay.rs
lzpap: refactored with [f57e09b](https://github.com/iotaledger/iota/pull/11062/commits/f57e09b629c7bb6f293666bb2169c44a92e5fa60)

crates/iota-replay/src/replay.rs
lzpap: refactored with [f57e09b](https://github.com/iotaledger/iota/pull/11062/commits/f57e09b629c7bb6f293666bb2169c44a92e5fa60)

lzpap (commented):
@valeriyr LGTM

bingyanglin (approved):

crates/iota-replay/src/replay.rs
```diff
@@ -726,9 +734,18 @@ impl LocalExec {
         // Initialize the state necessary for execution
         // Get the input objects
         let input_objects = self.initialize_execution_env_state(tx_info).await?;
+        let unique_shared_object_ids: HashSet<_> = input_objects
+            .filter_shared_objects()
+            .iter()
+            .map(|s| match s {
+                iota_types::execution::SharedInput::Existing(obj_ref) => obj_ref.0,
+                iota_types::execution::SharedInput::Deleted((id, _, _, _)) => *id,
+                iota_types::execution::SharedInput::Cancelled((id, _)) => *id,
```
Alex6323: consider importing `SharedInput`?

crates/iota-replay/src/replay.rs
```diff
@@ -770,22 +787,98 @@ impl LocalExec {
...
+            let (_, per_authenticator_inputs) = tx_info
+                .sender_signed_data
+                .split_input_objects_into_groups_for_reading(input_objects.clone())
+                .map_err(|e| ReplayEngineError::GeneralError { err: e.to_string() })?;
+
+            debug_assert_eq!(
+                move_authenticators.len(),
+                per_authenticator_inputs.len(),
+                "Move authenticators amount must match the number of authenticator inputs"
+            );
```
Alex6323, line 824: so in release builds we assume that we don't ever violate this invariant?

Alex6323 (approved):
Feel free to address or ignore my new comments. Just cosmetics :).

crates/iota-replay/src/replay.rs
lzpap: @valeriyr would be the best person to answer why it is a debug assertion

crates/iota-replay/src/replay.rs
lzpap: [6955948](https://github.com/iotaledger/iota/pull/11062/commits/6955948a946ae8350b640ec12fd727e10d4ab9fd)

PR #11067, opened 2026-04-01
feat(iota-types)!: replace TransactionData with SDK Transaction

### LINKED ISSUE #10548: Replace `TransactionData` with SDK version

Thoralf-M:
https://github.com/iotaledger/iota/issues/10548

## Summary

Part of the ongoing type migration from `iota-types` to `iota-rust-sdk`. Replaces the locally-defined `TransactionData` enum with a re-export of `iota_sdk_types::Transaction`, making the SDK the canonical source for this type.

## Key changes

- **Type replacement:** `TransactionData` is now `pub use iota_sdk_types::Transaction as TransactionData` and `TransactionDataV1` is `TransactionV1`. Removed the local enum definition and `enum_dispatch`.
- **Manual trait impl:** Since `enum_dispatch` can't be used on external types, `TransactionDataAPI` is now implemented directly on the enum with explicit `match` arms.
- **Digest/hash fix:** The SDK type is named `Transaction` not `TransactionData`, which broke `default_hash()` (uses `serde_name::trace_name`). A manual digest implementation was added to preserve the same hash.
- **Unused methods removed:** `contains_shared_object`, `uses_randomness`, `digest` were flagged as unused and cleaned up.


Dr-Electron (approved):
Do we even still need these rust examples in the docs folder? Shouldn't those mainly be part of the new repo now? 🤔

bingyanglin (approved):

crates/iota-types/src/transaction.rs
```diff
@@ -936,31 +936,419 @@ impl TransactionKindExt for TransactionKind {
     }
 }
 
-#[enum_dispatch(TransactionDataAPI)]
-#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
-pub enum TransactionData {
-    V1(TransactionDataV1),
-    // When new variants are introduced, it is important that we check version support
-    // in the validity_check function based on the protocol config.
+pub trait TransactionDataAPI {
```
Alex6323, line 952: let Claude document every method?

crates/iota-types/src/transaction.rs
```diff
@@ -936,31 +936,419 @@ impl TransactionKindExt for TransactionKind {
...
+    fn sender_mut_for_testing(&mut self) -> &mut IotaAddress;
+
+    fn gas_data_mut(&mut self) -> &mut GasData;
+
+    fn expiration_mut_for_testing(&mut self) -> &mut TransactionExpiration;
+
+    fn new_system_transaction(kind: TransactionKind) -> TransactionData;
+
+    #[allow(clippy::new_ret_no_self)]
+    fn new(
```
Alex6323, line 1045: MOve some, if not all, ctors to the SDK?

crates/iota-types/src/transaction.rs
```diff
@@ -936,31 +936,419 @@ impl TransactionKindExt for TransactionKind {
...
+    }
+
+    fn gas_data_mut(&mut self) -> &mut GasData {
+        match self {
+            Self::V1(v1) => &mut v1.gas_payment,
+            _ => unimplemented!("a new Transaction variant was added and needs to be handled"),
+        }
+    }
+
+    fn expiration_mut_for_testing(&mut self) -> &mut TransactionExpiration {
```
Alex6323, line 1397: I wonder if we should keep the production ext trait separate from the testing stuff.

crates/iota-types/src/transaction.rs
Thoralf-M: I don't think they make sense in the SDK, they are rather specific to the monorepo. The methods could be removed, but it would require changes in >200 places which I think is not worth it now. We could migrate these places later over time and then remove the methods from the trait here

crates/iota-types/src/transaction.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/11067/changes/55edf1d1d16a7e4baa9cb92a2037fe25235c85a4

piotrm50 (approved):

crates/iota-types/src/transaction.rs
Thoralf-M: Don't see the benefit of it, I think that would be just more annoying because you then need to know which trait to import, seems like right now there are also other methods only for testing used like `new_split_coin`

crates/iota-types/src/transaction.rs
thibault-martinez, line 1: Not necessarily asking you to remove but these methods are unused:
- `contains_shared_object`
- `uses_randomness`
- `digest`

crates/iota-replay/tests/sandbox_snapshots/7qq4W43TqHg9tQPMvdAFW4Tz6J88KnPppBPR1hNKmQAd.json
```diff
@@ -10,7 +10,7 @@
                         "appId": 0
                     },
                     "value": {
-                        "V1": {
+                        "1": {
```
thibault-martinez: I updated the file but what do you think about this?

crates/iota-types/src/transaction.rs
Thoralf-M: Good catch, I think we should remove as much as possible as it's hard to cleanup such things later (not that easy to see if there are multiple fn definitions with the same name)

Thoralf-M: https://github.com/iotaledger/iota/pull/11067/commits/0c1d6e7b73dc0478145ddc3c760880c5970c391d

crates/iota-replay/tests/sandbox_snapshots/7qq4W43TqHg9tQPMvdAFW4Tz6J88KnPppBPR1hNKmQAd.json
Thoralf-M: I think the V1 is a bit clearer that this is about a version if only looking at the JSON, just "1" could also be something else

thibault-martinez: I agree

crates/iota-core/tests/staged/iota.yaml
```diff
@@ -897,22 +897,12 @@ SystemPackage:
     - dependencies:
         SEQ:
           TYPENAME: ObjectId
-TransactionData:
+Transaction:
   ENUM:
     0:
-      V1:
+      "1":
```
thibault-martinez: Does this not need to be updated?

crates/iota-core/src/authority.rs
```diff
@@ -2370,7 +2367,7 @@ impl AuthorityState {
             },
             transaction,
         );
-        let transaction_digest = TransactionDigest::new(default_hash(&intent_msg.value));
+        let transaction_digest = TransactionDigest::new(intent_msg.value.digest().into_inner());
```
thibault-martinez, line 2370: So `digest` is doing the hash? The exact same hash?

DaughterOfMars (approved):

crates/iota-core/tests/staged/iota.yaml
Thoralf-M: https://github.com/iotaledger/iota/pull/11067/commits/bb08eb8214ead8b1e9fce5e09f7855d84173d65a

crates/iota-types/src/transaction.rs
```diff
@@ -936,31 +936,478 @@ impl TransactionKindExt for TransactionKind {
...
 }
 
-#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
-pub struct TransactionDataV1 {
-    pub kind: TransactionKind,
-    pub sender: IotaAddress,
-    pub gas_data: GasData,
-    pub expiration: TransactionExpiration,
-}
+impl TransactionDataAPI for TransactionData {
```
thibault-martinez, line 1247: Just curious, it used to be implemented on the variant and now it's implemented on the enum, why?

crates/iota-core/src/authority.rs
Thoralf-M: Yes the same (before the SDK replacement as the name change), I needed to change it as it was using the serde name which is now just `Transaction` instead of `TransactionData` and the `default_hash()` uses `let name = serde_name::trace_name::<Self>().expect("Self must be a struct or an enum");` so it failed after rename

crates/iota-types/src/transaction.rs
Thoralf-M: There was `#[enum_dispatch(TransactionDataAPI)]` and that's not possible for external defined types

thibault-martinez (approved):

PR #11068, opened 2026-04-01
fix(ui-kit): Align title padding

marc2332:
Before:

image

After:
image


KeitoTadashi (approved):

evavirseda (changes requested):
i was thinking and maybe we should remove the padding of the title, i mean it has no sense that a title has a padding, it should be in the wrapper/parent component

panteleymonchuk:
https://github.com/iotaledger/ts-packages/pull/93

PR #11074, opened 2026-04-01
refactor(iota-indexer-streaming, iota-graphql-rpc): use diesel for pg listen notify

### LINKED ISSUE #9950: [indexer-streaming]: use `diesel` crate instead of `tokio-postgres` for `LISTEN/NOTIFY`
In the first iteration we used `tokio-postgres` because `diesel` crate did not support the `LISTEN/NOTIFY` notifications stream. 

With `diesel 2.3.6` we see that the support is stable and should not introduce any breaking changes if upgrading to this version.

Refactor the `iota-indexer-streaming` crate to use the `diesel` `LISTEN/NOTIFY` [implementation](https://docs.rs/diesel/2.3.6/diesel/pg/struct.PgConnection.html#method.notifications_iter) and remove the `tokio-postgres` dependency.

sergiupopescu199:
# Description of change

This PR removes the `tokio-postgres` dependency in favor of the already present `diesel`. At the time of implementing the feature, `LISTEN/NOTIFY` support was not yet available in diesel. Starting from version `2.3.2`, diesel added this support, making the separate `tokio-postgres` dependency unnecessary.
 
fixes #9950 

- used `iota-localnet` to spin up a private network
```shell
RUST_LOG=iota_graphql_rpc=error,iota_indexer_streaming=info cargo r --features indexer -- start --force-regenesis --with-graphql
```
- connected to graphiQL to execute subscription queries

```graphql
subscription Transactions {
    transactions {
        __typename
        ... on TransactionBlock {
            digest,
           effects {
            checkpoint {
              sequenceNumber,
            }
          }
        }
        ... on Lagged {
            count
        }
    }
}
```

- Compared performance on the staging environment against the previous `tokio-postgres` implementation. Prometheus metrics showed no regression in notification processing, batch querying, or broadcasting latency.

> [!NOTE]
> These changes do not affect the indexer's main checkpoint ingestion pipeline, so further ingestion tests were not conducted.

### DIFF
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -52,7 +52,7 @@ const CHANNEL_NAME: &str = "checkpoint_committed";
 /// of failure.
 const RETRY_POSTGRES_CONNECTION_DELAY: Duration = Duration::from_secs(5);
 /// Interval between polling for new notifications from the Postgres NOTIFY
-/// channel when client processed then all.
+/// channel when client processed them all.
 const PG_NOTIFICATION_POLL_INTERVAL: Duration = Duration::from_millis(100);
 
 /// Notification received from PostgreSQL NOTIFY channel when a checkpoint is
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -13,13 +13,17 @@ use std::{
     fmt::Debug,
     num::{NonZeroI64, NonZeroUsize},
     pin::Pin,
-    str::FromStr,
     sync::Arc,
     task::{Context, Poll},
     time::{Duration, Instant},
 };
 
-use futures::{Stream, StreamExt, TryFutureExt, stream};
+use diesel::{
+    PgConnection, QueryResult, RunQueryDsl,
+    pg::PgNotification,
+    r2d2::{ConnectionManager, PooledConnection},
+};
+use futures::{Stream, StreamExt, TryFutureExt};
 use iota_indexer::{
     models::{
         events::StoredEvent,
@@ -31,9 +35,6 @@ use iota_types::event::Event;
 use prometheus::IntGauge;
 use serde::{Deserialize, Serialize};
 use tokio::sync::broadcast;
-use tokio_postgres::{
-    AsyncMessage, Config as PostgresConfig, Connection, NoTls, Socket, tls::NoTlsStream,
-};
 use tokio_stream::wrappers::{BroadcastStream, errors::BroadcastStreamRecvError};
 use tracing::{debug, error};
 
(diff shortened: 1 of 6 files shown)


crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -389,16 +400,13 @@ impl InMemory {
                 }
             }
         }
-        Err(IndexerStreamingError::Postgres(
```
tomxey, line 392: Earlier on error we were returning from this function, reconnecting in the caller and calling this function again.
What happens in the current implementation on connection failure? is the pool automatically reconnected and does not return any errors?

sergiupopescu199: The reconnection behavior is preserved but the mechanism is slightly different since we moved from an async stream to a synchronous polling iterator.
 
With `tokio-postgres`, there were two failure modes: (1) the stream terminates (yields `None` permanently) — we returned the "stream closed" error at the bottom of the function, and (2) an error inside the stream value (`Err(tokio_postgres::Error)`) — propagated through `filter_checkpoint_notifications`.
 
With diesel's `notifications_iter()`, `None` just means "no pending notifications right now" — it doesn't signal a closed connection. So failure mode (1) no longer exists. Connection failures now only surface through failure mode (2): as an `Err` inside `QueryResult<PgNotification>`, which flows through `filter_checkpoint_notifications` → `resolve_tx_bounds` → `?` propagation. The outer loop then acquires a new connection from the pool, re-issues `LISTEN`, and resumes.

tomxey: got it, great 👌

tomxey (approved):

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -42,12 +43,17 @@ use crate::{
...
+pub type PoolConnection = PooledConnection<ConnectionManager<PgConnection>>;
+
 /// Postgres NOTIFY channel name.
 const CHANNEL_NAME: &str = "checkpoint_committed";
 
 /// Delay in seconds before retrying to connect to the Postgres database in case
 /// of failure.
 const RETRY_POSTGRES_CONNECTION_DELAY: Duration = Duration::from_secs(5);
+/// Interval between polling for new notifications from the Postgres NOTIFY
+/// channel when client processed then all.
```
kodemartin: typo:
```suggestion
/// channel when client processed them all.
```

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -188,40 +189,40 @@ impl InMemory {
 
             async move {
                 loop {
-                    let (client, connection) = match pg_config.connect(NoTls).await {
+                    let mut pool = match indexer_reader.get_pool().get() {
```
kodemartin: This gets a connection actually: 

```rust
                    let mut connection = match indexer_reader.get_pool().get() {
```

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -321,7 +322,7 @@ impl InMemory {
     async fn process_checkpoint_notifications(
         metrics: &InMemoryStreamMetrics,
         config: &Config,
-        mut connection: Connection<Socket, NoTlsStream>,
+        pool: &mut PoolConnection,
```
kodemartin: As above: this is a connection from the pool

```rust
        connection: &mut PoolConnection,
```

kodemartin (approved):
Looks good with minor suggestions.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11074/commits/ea66bae768fbd8df13d16736726ee137e58b3c59

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11074/commits/ea66bae768fbd8df13d16736726ee137e58b3c59

PR #11089, opened 2026-04-01
feat(iota-core): PoC certificate-less pre-consensus load shedding based on consensus queue length

### LINKED ISSUE #11035: pcool: PoC version of pre-consensus load shedding without certification
## Description

Implement a PoC version of certificate-less pre-consensus load shedding based on consensus queue length.

## Motivation

With certificates, each validator checks the system's (consensus, execution) overload state before signing a transaction. If the system is overloaded, the validator refuses to sign. If `f+1` validators observe overload, no certificate can be formed, which acts as backpressure on the client—it backs off before retrying, so consensus is protected. In the certificate-less flow, this natural implicit rate limiter—the certification step—is removed, so transactions hit consensus directly, making the consensus queue check especially important.

More specifically, the certificate-less (white-flag) flow submits transactions directly to consensus with no quorum step to naturally throttle the rate. The only consensus overload protection is a binary hard cutoff in `ConsensusAdapter::check_consensus_overload()`: either under limit (accept) or over limit (reject with `TooManyTransactionsPendingConsensus`).

## Proposal

Our plan is to introduce *graduated* pre-consensus load shedding based on consensus queue length: as the queue fills, an increasing percentage of transactions are shed proportionally, using the same deterministic rejection mechanism (`XxHash64` + temporal seed) already used by the execution-based overload monitor. The plan is to reuse `overload_monitor_accept_tx()` directly with the following change: compute a different `load_shedding_percentage` (from consensus queue length instead of execution latency) and pass it in.

Another difference is that we would like to use the consensus queue length `num_inflight_transactions: AtomicU64` as a clean instantaneous signal rather than a periodic monitor used by the existing execution-based load shedding.

## Requirements

Write a list of what you want this feature to do.

1. Certificate-less pre-consensus load shedding should only be based on consensus queue length (execution queue length will be part of post-consensus load shedding).
2. Certificate-less pre-consensus load shedding should protect the validator's local queue length based on the consensus load.

Yes.

roman1e2f5p8s:
# Description of change

This PR implements the new graduated consensus queue load shedding for the white-flag flow. The core function that calculates the graduated consensus load shedding percentage is `compute_graduated_load_shedding_percentage()`.

### Summary of changes:

- Added a white-flag flow gate to `AuthorityState::check_system_overload()`:
  - In certificate-less mode, it only checks consensus overload. Execution overload checks will be post-consensus and implemented separately.
  - In certificate mode, it runs all existing checks (authority overload, execution queue, consensus queue, and writeback cache).
- New metric `consensus_queue_load_shedding_percentage` in `AuthorityMetrics`.
- Added `AuthorityState::check_consensus_queue_graduated_limits()` that calls `compute_graduated_load_shedding_percentage()`, sets the `consensus_queue_load_shedding_percentage` metric, and decides how to reject: probabilistic rejection (shedding percentage < 100%) returns `ValidatorOverloadedRetryAfter { retry_after_secs: 30 }` via `overload_monitor_accept_tx()`; unconditional rejection (shedding percentage >= 100%) returns `TooManyTransactionsPendingConsensus`.
- Added `consensus_queue_overload_monitor()` (in `overload_monitor.rs`), spawned by `iota-node` only when the white-flag protocol flag is enabled. Periodically refreshes the `consensus_queue_load_shedding_percentage` metric so it tracks queue length even when no gRPC traffic is arriving. Handle is plumbed through `ValidatorComponents` and across epoch reconfig.
- New field to `ConsensusConfig` and `ConsensusAdapter: graduated_load_shedding_soft_limit_pct` (defaults to 50, meaning graduated shedding starts at 50% of the hard limit). The hard limit matches the existing `max_pending_transactions` (default 20_000).
- Renamed `ConsensusAdapter::check_limits()` to `check_consensus_hard_limits()` (private bool helper).

### Other indirectly related changes:

- In `ValidatorService`, replaced stale references to the non-existent `handle_submit_transaction(s)_impl` with `submit_tx` (`ValidatorV2` service).
- Fixed docstrings on fields of `AuthorityOverloadConfig`.
- Added test-only `ConsensusAdapter::new_for_testing_with_authority_name()` which constructs a new consensus adapter frequently used in many tests.
- Removed unnecessary pub visibility in `overload_monitor.rs` tests.
- Added docstrings to `ValidatorService`, `ConsensusAdapter`, and `AuthorityState` methods related to this PR.
- In `ValidatorService`, split long string literals into multiple lines.

Closes #11035.

```shell
cargo test -p iota-core test_compute_graduated_load_shedding_percentage
cargo test -p iota-core test_consensus_queue_graduated_load_shedding 
```

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -808,19 +808,21 @@ impl NodeConfig {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct ConsensusConfig {
-    // Base consensus DB path for all epochs.
+    /// Base consensus DB path for all epochs.
     pub db_path: PathBuf,
 
-    // The number of epochs for which to retain the consensus DBs. Setting it to 0 will make a
-    // consensus DB getting dropped as soon as system is switched to a new epoch.
+    /// The number of epochs for which to retain the consensus DBs.
+    /// Setting it to 0 will make a consensus DB getting dropped
+    /// as soon as system is switched to a new epoch.
     pub db_retention_epochs: Option<u64>,
 
-    // Pruner will run on every epoch change but it will also check periodically on every
-    // `db_pruner_period_secs` seconds to see if there are any epoch DBs to remove.
+    /// Pruner will run on every epoch change but it will also check
+    /// periodically on every `db_pruner_period_secs` seconds to see
+    /// if there are any epoch DBs to remove.
     pub db_pruner_period_secs: Option<u64>,
 
-    /// Maximum number of pending transactions to submit to consensus, including
-    /// those in submission wait.
+    /// Maximum number of pending transactions to submit to consensus,
+    /// including those in submission wait.
     ///
     /// Default to 20_000 inflight limit, assuming 20_000 txn tps * 1 sec
     /// consensus latency.
@@ -860,6 +862,9 @@ impl ConsensusConfig {
         &self.db_path
     }
 
+    /// Returns the maximum number of pending transactions to submit to
+    /// consensus, including those in submission wait. Defaults to 20_000
+    /// inflight limit, assuming 20_000 txn tps * 1 sec consensus latency.
     pub fn max_pending_transactions(&self) -> usize {
         self.max_pending_transactions.unwrap_or(20_000)
     }
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1183,54 +1183,74 @@ pub struct TransactionKeyValueStoreWriteConfig {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct AuthorityOverloadConfig {
+    /// Maximum time a transaction can wait in the transaction manager execution
+    /// queue before it triggers an overload detection on the object it depends
+    /// on.
     #[serde(default = "default_max_txn_age_in_queue")]
     pub max_txn_age_in_queue: Duration,
 
-    // The interval of checking overload signal.
+    /// The interval of checking overload signal.
     #[serde(default = "default_overload_monitor_interval")]
     pub overload_monitor_interval: Duration,
 
-    // The execution queueing latency when entering load shedding mode.
+    /// The execution queueing latency when entering load shedding mode.
     #[serde(default = "default_execution_queue_latency_soft_limit")]
     pub execution_queue_latency_soft_limit: Duration,
 
-    // The execution queueing latency when entering aggressive load shedding mode.
+    /// The execution queueing latency when entering aggressive load shedding
+    /// mode.
     #[serde(default = "default_execution_queue_latency_hard_limit")]
     pub execution_queue_latency_hard_limit: Duration,
 
-    // The maximum percentage of transactions to shed in load shedding mode.
+    /// The maximum percentage of transactions to shed in load shedding mode.
     #[serde(default = "default_max_load_shedding_percentage")]
     pub max_load_shedding_percentage: u32,
 
-    // When in aggressive load shedding mode, the minimum percentage of
-    // transactions to shed.
+    /// When in aggressive load shedding mode, the minimum percentage of
+    /// transactions to shed.
     #[serde(default = "default_min_load_shedding_percentage_above_hard_limit")]
     pub min_load_shedding_percentage_above_hard_limit: u32,
 
-    // If transaction ready rate is below this rate, we consider the validator
-    // is well under used, and will not enter load shedding mode.
+    /// If transaction ready rate is below this rate, we consider the validator
+    /// is well under used, and will not enter load shedding mode.
     #[serde(default = "default_safe_transaction_ready_rate")]
     pub safe_transaction_ready_rate: u32,
 
-    // When set to true, transaction signing may be rejected when the validator
-    // is overloaded.
+    /// When set to true, transaction signing may be rejected when the validator
+    /// is overloaded.
     #[serde(default = "default_check_system_overload_at_signing")]
     pub check_system_overload_at_signing: bool,
 
-    // When set to true, transaction execution may be rejected when the validator
-    // is overloaded.
+    /// When set to true, transaction execution may be rejected when the
+    /// validator is overloaded.
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub check_system_overload_at_execution: bool,
 
-    // Reject a transaction if transaction manager queue length is above this threshold.
-    // 100_000 = 10k TPS * 5s resident time in transaction manager (pending + executing) * 2.
+    /// Reject a transaction if transaction manager queue length is above this
+    /// threshold. 100_000 = 10k TPS * 5s resident time in transaction
+    /// manager (pending + executing) * 2.
     #[serde(default = "default_max_transaction_manager_queue_length")]
     pub max_transaction_manager_queue_length: usize,
 
-    // Reject a transaction if the number of pending transactions depending on the object
-    // is above the threshold.
+    /// Reject a transaction if the number of pending transactions depending on
+    /// the object is above the threshold.
     #[serde(default = "default_max_transaction_manager_per_object_queue_length")]
     pub max_transaction_manager_per_object_queue_length: usize,
+
+    /// Consensus queue length at which graduated pre-consensus load shedding
+    /// begins. This is used in the certificate-less (white-flag) mode.
+    #[serde(default = "default_consensus_queue_length_soft_limit")]
+    pub consensus_queue_length_soft_limit: usize,
+
+    /// Consensus queue length at which max pre-consensus load shedding is
+    /// applied. This is used in the certificate-less (white-flag) mode.
+    #[serde(default = "default_consensus_queue_length_hard_limit")]
+    pub consensus_queue_length_hard_limit: usize,
+
+    /// Max percentage of transactions to shed due to consensus queue overload.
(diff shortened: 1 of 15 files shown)


bingyanglin (approved):

crates/iota-config/src/node.rs
```diff
@@ -885,17 +887,48 @@ pub struct ConsensusConfig {
     /// Parameters for Starfish consensus
     #[serde(skip_serializing_if = "Option::is_none")]
     pub starfish_parameters: Option<StarfishParameters>,
+
+    /// Consensus queue length at which graduated load shedding begins.
+    /// Used in the certificate-less (white-flag) mode.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub graduated_load_shedding_soft_limit: Option<usize>,
```
cyberphysic4l: I think it might make more sense to have this in AuthorityOverloadConfig alongside the thresholds for the execution related overload metrics.

crates/iota-core/src/authority.rs
```diff
@@ -1121,44 +1131,100 @@ impl AuthorityState {
...
+            // to hard limit.
+            self.check_consensus_queue_overload(consensus_adapter, tx_data)
+                .tap_err(|_| {
+                    self.update_overload_metrics("consensus");
+                })?;
+
+            // Binary hard cutoff as defense-in-depth: graduated shedding at 100%
+            // already rejects everything at or above `max_pending_transactions`,
+            // so this check is redundant but kept as a safeguard.
+            consensus_adapter.check_consensus_overload().tap_err(|_| {
```
cyberphysic4l, line 1159: I was thinking about this and I think this is actually not necessary to have this safeguard. Instead, I think max shedding percentage should be hardcoded as 100 and only parameters we set are soft limit, hard limit, and perhaps a soft limit percentage. The hard limit should always be 100% drop by definition as it is the extreme case.

cyberphysic4l (commented):
I like the graduated load shedding approach, I think it's a simple and effective improvement over a single hard limit. My only suggestion is to simplify further and clean up the parameterisation a bit. The hard limit should always correspond to 100% load shedding. The part that should be configurable should be the percentage for the soft limit imo. I also think it would be nice to move all these consensus overload parameters into the overload config instead of with the other consensus config parameters.

crates/iota-config/src/node.rs
roman1e2f5p8s: this is actually where I held the params initially. why they are not there anymore:

- `AuthorityOverloadConfig` holds zero consensus-related parameters;
- the hard limit `max_pending_transactions` already lived in `ConsensusConfig`, not in `AuthorityOverloadConfig`:
  - any caller operating on soft/hard limits and shedding percentage would need to have access to both `AuthorityOverloadConfig` and `ConsensusConfig` if we store parameters at different configs;
  -  we would like the hard limit to match `max_pending_transactions`: if `AuthorityOverloadConfig` held a separate hard limit, who would check it matches `max_pending_transactions`?

crates/iota-core/src/authority.rs
roman1e2f5p8s: right, we could remove it if the hard limit is always set to 100%. but also keep in mind that `check_consensus_overload()` internally also checks if the semaphore has available permits for parallel permissions - either we extract that part into a separate function, or leave the binary cutoff as it is.

anyway, I will update the comment to explain that `check_consensus_overload` is not fully redundant.

roman1e2f5p8s: I clarified in the comment why the previous safeguard should stay (https://github.com/iotaledger/iota/commit/d7843e399461cee882d70a1ac35a3731d6fc80f1)

roman1e2f5p8s: > max shedding percentage should be hardcoded as 100 and only parameters we set are soft limit, hard limit, and perhaps a soft limit percentage

I changed to take only the current queue size, hard limit, and percentage when load shedding starts; the max shedding is hard-coded to 100% in the function. Did I get it right? Otherwise, can you please clarify what "soft limit percentage" is and why it is needed to be passed to the function if we pass soft limit?

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -100,6 +105,7 @@ impl ValidatorService {
             &consensus_adapter,
             transaction.data(),
             state.check_system_overload_at_signing(),
+            white_flag_flow_enabled, // should be false if this point is reached
```
piotrm50: then why not just put false here if it's asserted above?

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -74,13 +76,16 @@ impl ValidatorService {
...
         let transaction = request.into_inner();
         let epoch_store = state.load_epoch_store_one_call_per_task();
+        let white_flag_flow_enabled = epoch_store.protocol_config().enable_white_flag_flow();
 
         // Reject if white flag flow is enabled - transactions should use
-        // submit_transaction instead
+        // `handle_submit_transaction_impl` instead
         fp_ensure!(
-            !epoch_store.protocol_config().enable_white_flag_flow(),
+            !white_flag_flow_enabled,
```
piotrm50: this can be read directly as before. We don't need to pass this value below if we know it's false by the time we reach that point.

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -232,11 +238,13 @@ impl ValidatorService {
 
         // 2) Verify the certificates.
         // Check system overload
+        let white_flag_flow_enabled = epoch_store.protocol_config().enable_white_flag_flow();
         for certificate in &certificates {
             let overload_check_res = self.state.check_system_overload(
                 &self.consensus_adapter,
                 certificate.data(),
                 self.state.check_system_overload_at_execution(),
+                white_flag_flow_enabled, // should be false if this point is reached
```
piotrm50: if we know it needs to be false then why even read it from the config? 

btw. is this even checked anywhere in that function or in the caller that whiteflag flow is disabled? It is explicitly checked in `handle_transaction`

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -161,6 +161,7 @@ impl ValidatorService {
             consensus_adapter,
             transaction.data(),
             state.check_system_overload_at_signing(),
+            epoch_store.protocol_config().enable_white_flag_flow(),
```
piotrm50: I think we can always put `true` here. This parameter will be removed anyway once we fully switch to pcool flow.

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -74,13 +76,16 @@ impl ValidatorService {
...
         // Reject if white flag flow is enabled - transactions should use
-        // submit_transaction instead
+        // `handle_submit_transaction_impl` instead
         fp_ensure!(
-            !epoch_store.protocol_config().enable_white_flag_flow(),
+            !white_flag_flow_enabled,
             IotaError::UnsupportedFeature {
-                error: "handle_transaction is disabled when white flag flow is enabled. Use submit_transaction instead.".to_string()
+                error: "handle_transaction is disabled when white flag flow is enabled. \
+                    Use handle_submit_transaction_impl instead."
```
piotrm50: `handle_submit_transaction_impl` does not exist

piotrm50 (changes requested):

crates/iota-core/src/authority_server/validator.rs
roman1e2f5p8s: yes, in all 3 callers

crates/iota-core/src/authority_server/validator.rs
roman1e2f5p8s: changed in https://github.com/iotaledger/iota/pull/11089/commits/5288d40e99266a31d59af2ccc1f22647499622c0

crates/iota-core/src/authority_server/validator.rs
roman1e2f5p8s: fixed in https://github.com/iotaledger/iota/pull/11089/commits/5288d40e99266a31d59af2ccc1f22647499622c0

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -100,6 +104,7 @@ impl ValidatorService {
             &consensus_adapter,
             transaction.data(),
             state.check_system_overload_at_signing(),
+            false, // white_flag_flow_enabled
```
piotrm50:
```suggestion
            false, // white flag flow disabled
```

crates/iota-core/src/authority_server/validator.rs
```diff
@@ -237,6 +242,7 @@ impl ValidatorService {
                 &self.consensus_adapter,
                 certificate.data(),
                 self.state.check_system_overload_at_execution(),
+                false, // white_flag_flow_enabled
```
piotrm50:
```suggestion
                false, // white flag flow disabled
```

crates/iota-core/src/authority_server/validator.rs
roman1e2f5p8s: what is written in the comment is the exact name of the parameter that function takes.
since now we pass "magic flags" (true/false), I tend to name them according to standard practice `f(..., /* white_flag_flow_enabled */ false)`

your suggestion might read as:
- white flag flow disabled?
- false

roman1e2f5p8s: clarified in https://github.com/iotaledger/iota/pull/11089/commits/014f91cd369666cb25452bc7733afb8cbc28a7a3

piotrm50 (approved):

bingyanglin (approved):

PR #11039, opened 2026-03-30
feat(types): add `tx_data_bytes` and `signing_digest` to AuthContext

miker83z:
# Description of change

Extends the AuthContext with BCS-serialized TransactionData bytes, enabling Move authenticators to reconstruct and verify the exact message that protocol signatures commit to.

Changes:

- Add `tx_data_bytes` field to AuthContext (Rust + Move), populated during execution from the serialized TransactionData
- Add `intent_tx_data_bytes()` — returns the intent-prefixed transaction data (IntentMessage<TransactionData> bytes)
- Add `signed_tx_bytes()` — returns the Blake2b256 hash of the intent message, i.e., the bytes that generic signatures actually sign over
- Introduce `iota::intent` Move module mirroring the Rust Intent type (scope, version, app_id constants for all known intent scopes)

This gives abstract account authenticators full access to the signed payload, allowing them to perform signature verification entirely in Move without relying on built-in scheme support.

## Links to any relevant issues

Closes #10996 

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [x] Protocol: Add `tx_data_bytes` and `signing_digest` to AuthContext. Add the signing `Intent` representation to the iota framework.
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

PR #11094, opened 2026-04-02
feat(iota-framework): Implementation of Claim Registry

### LINKED ISSUE #11021: [AA] Create a Claim Registry and a dedicated native function for deriving object ids from pub keys
- A default account can be created by claiming its address on-chain.
- The account’s object will be created with ObjectID=address.
- It is not possible to claim an address twice: there exists some form of claim registry that prevents duplicate claims.
- An address can only be claimed by proving ownership of it with a cryptographic signature.
- Once the address is claimed, a default account is set up for the address that supports move based authentication on the very same signature formats as the proof.

This prototype can be loosely based on this implementation #7632.

Dkwcs:
# Description of change 

This PR implements `iota::claim_registry` - a singleton shared object at `0x10` that lets any address owner prove ownership and receive a deterministic `UID` bound to their address.

`claim(registry, public_key, ctx)` validates that the provided `PublicKey` derives to `ctx.sender()`, marks the address as claimed, and returns a `UID` that the caller must immediately use as the `id` field of a new on-chain object. Because `UID` has no `drop` ability, leaving it unconsumed is a compile error.

Supported schemes: Ed25519 (0x00), Secp256k1 (0x01), Secp256r1 (0x02), MultiSig (0x03), Passkey (0x06).

`ClaimRegistry` is created via `EndOfEpochTransaction::ClaimRegistryCreate` at the end of the first epoch running the protocol version where `enable_claim_registry` activates.

Updated snapshots:

crates/iota-graphql-rpc/tests/snapshots/snapshot_tests__schema_sdl_export.snap
crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__Mainnet_version_26.snap
crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__Testnet_version_26.snap
crates/iota-protocol-config/src/snapshots/iota_protocol_config__test__version_26.snap
crates/iota-swarm-config/tests/snapshots/snapshot_tests__genesis_config_snapshot_matches.snap
crates/iota-swarm-config/tests/snapshots/snapshot_tests__populated_genesis_snapshot_matches-2.snap
crates/iota-framework-snapshot/bytecode_snapshot/26/0x0000...0002
crates/iota-framework/packages_compiled/iota-framework

## TODOs

- Add `ClaimRegistryCreateTransaction` variant to `iota-rust-sdk` and update the dependency (will be done in the feature branch).

Fixes #11021 .

run in iota-adapter-tx-tests:
`
cargo run -p iota -- move test --path /iota/crates/iota-framework/packages/iota-framework/tests/claim_registry_tests.move
`
`
cargo simtest --test /iota/crates/iota-e2e-tests/tests/claim_registry_tests   
`


crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+/// The caller must be the account owner: `ctx.sender() == object_id(account)`.
+/// This is guaranteed by the ObjectID == address invariant established at claim
+/// time, so only the address holder can satisfy this check.
+public fun rotate_key(
+    account: &mut IotaDefaultAccount,
+    new_public_key: vector<u8>,
+    new_scheme: u8,
+    ctx: &TxContext,
+) {
+    ensure_tx_sender_is_account(account, ctx);
```
lzpap: this check only works if transaction validation rules are modified as well. Otherwise, nothing stops me from calling this function and rotating the key on the account with the "old" signature validation!

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+///   - Secp256k1: 64-byte (r,s) signature, signed over `SHA256(digest)`.
+///   - Secp256r1: 64-byte (r,s) signature, signed over `SHA256(digest)`.
+#[authenticator]
+public fun authenticate(
+    account: &IotaDefaultAccount,
+    signature: vector<u8>,
+    _auth_ctx: &iota::auth_context::AuthContext,
+    ctx: &TxContext,
+) {
+    let digest = ctx.digest();
```
lzpap: Nope. We need need the signing digest to stay compatible with the generic signature variants already present in the protocol, such that clients don't need to change their signign flows.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+
+// === Authenticate (Move-based authenticator) ===
+
+/// Authenticate a `MoveAuthenticator` transaction for an `IotaDefaultAccount`.
+///
+/// The VM calls this function when processing a `MoveAuthenticator` transaction
+/// whose `object_to_authenticate` is an `IotaDefaultAccount`.
+///
+/// `signature` is provided by the user as `MoveAuthenticatorV1.call_args[0]`.
+/// It must be a valid signature over `auth_ctx.digest()` using the key stored
```
lzpap: Comment is wrong. It must be a valid signature over signing digest, which is `hash(Intent<TransactionData>)`. Has nothing to do with `auth_ctx.digest()`. Is this AI slop? 🤔

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+use std::ascii;
+
+// === Signature scheme flags (match Rust `SignatureScheme`) ===
+
+/// Ed25519 signature scheme flag.
+const SCHEME_ED25519: u8 = 0x00;
+/// Secp256k1 signature scheme flag.
+const SCHEME_SECP256K1: u8 = 0x01;
+/// Secp256r1 signature scheme flag.
+const SCHEME_SECP256R1: u8 = 0x02;
```
lzpap: What about passkeys?

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
lzpap, line 1: The claim registry, the `DefaultIotaAccount` and cryptographic authenticators are all in one module currently. I suggest we modularize them into their own modules.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// === Key rotation ===
+
+/// Rotate the stored public key and/or scheme of a default account.
+///
+/// The caller must be the account owner: `ctx.sender() == object_id(account)`.
+/// This is guaranteed by the ObjectID == address invariant established at claim
+/// time, so only the address holder can satisfy this check.
+public fun rotate_key(
```
lzpap: TODO: rotating to a new pubkey shall be protected, i.e. proof of ownership shall be checked. During the initial claim ctx.sender check satisfies this, but for any subsequent rotation we need a cryptographic signature over some rotation action-specific data from the new pubkey.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,324 @@
...
+    /// key existence matters.
+    claimed_addresses: Table<address, bool>,
+}
+
+// === Genesis ===
+
+/// Create and share the `ClaimRegistry` singleton.
+/// Called exactly once during genesis from address @0x0.
+#[allow(unused_function)]
+fun create(ctx: &mut TxContext) {
```
lzpap: Where is it created for already running networks when you can't call this during genesis?

lzpap (changes requested):

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Yeah, copilot :D 
It would have been removed in further commits

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: We discussed it on the weekly meeting, it will be separated

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Currently, nowhere ...

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Could you elaborate please?

Dkwcs: If I got you correctly, we had PR related to this(accessing sighing digest from the auth context)

lzpap: yes, this one: https://github.com/iotaledger/iota/pull/11039

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: I assume for implementing passkey we need this https://github.com/iotaledger/iota/pull/11039 as well.
We have to have a `AuthContext::signing_digest` to pass it as a part of ".challenge”

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: ee931f761d0b0cacd4197da39e7b0013de910960

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,160 @@
...
+// === Claim entry points (built-in authenticator) ===
+
+/// Claim with a 32-byte Ed25519 public key. Derived address must equal `ctx.sender()`.
+public entry fun claim_ed25519(
+    registry: &mut ClaimRegistry,
+    public_key: vector<u8>,
+    ctx: &mut TxContext,
+) {
+    validate_and_mark_claimed(registry, iota_default_account::scheme_ed25519(), &public_key, ctx);
+    iota_default_account::new(ctx.sender(), public_key, iota_default_account::scheme_ed25519());
```
miker83z: I propose, that, instead of creating an iota_default_account, here we can return a hot potato, like: 
```move
public struct ClaimedAddressTicket {
    account: address,
    public_key: vector<u8>,
    flag: u8,
}
```

miker83z: Then, just have only one `claim` function with params: `registry: &mut ClaimRegistry, scheme: u8, public_key: vector<u8>,`

miker83z: Finally, remove the implementation of the iota_default_account. We will deal with it in a new PR.

crates/iota-types/src/lib.rs
```diff
@@ -138,6 +139,7 @@ built_in_ids! {
     IOTA_AUTHENTICATOR_STATE_ADDRESS / IOTA_AUTHENTICATOR_STATE_OBJECT_ID = 0x7;
     IOTA_RANDOMNESS_STATE_ADDRESS / IOTA_RANDOMNESS_STATE_OBJECT_ID = 0x8;
     GENESIS_IOTA_BRIDGE_ADDRESS / GENESIS_IOTA_BRIDGE_OBJECT_ID = 0x9;
+    IOTA_CLAIM_REGISTRY_ADDRESS / IOTA_CLAIM_REGISTRY_OBJECT_ID = 0x11;
```
miker83z: We can use 0x10 or is there something hidden that uses it?

Dkwcs: ff9f471a8d970d0c86f2a3aa289ec842e25095ff

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: ff9f471a8d970d0c86f2a3aa289ec842e25095ff

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,155 @@
...
+
+fun is_valid_scheme(scheme: u8): bool {
+    scheme == SCHEME_ED25519
+        || scheme == SCHEME_SECP256K1
+        || scheme == SCHEME_SECP256R1
+        || scheme == SCHEME_MULTISIG
+        || scheme == SCHEME_PASSKEY
+}
+
+fun is_valid_public_key_length(scheme: u8, public_key: &vector<u8>): bool {
```
miker83z: This whole module needs to be adapted to the framework types introduced in #11184 (e.g., PublicKey)

Dkwcs: c6d9aeab473a69e8c2136573fa1f6747547a7d2b

crates/iota-framework-snapshot/bytecode_snapshot/23/0x0000000000000000000000000000000000000000000000000000000000000002
miker83z, line 1: revert this

crates/iota-framework-snapshot/manifest.json
```diff
@@ -590,7 +590,7 @@
     ]
   },
   "23": {
-    "git_revision": "3e2b511de52307e8513f278ee3576a1de47a267f",
+    "git_revision": "bef521b801023025587d1f0ad19ee8b63c56a613-dirty",
```
miker83z: revert this

crates/iota-framework/packages/iota-framework/sources/account_abstraction/authenticator_function.move
```diff
@@ -70,6 +70,20 @@ public fun function_name<Account: key>(self: &AuthenticatorFunctionRefV1<Account
...
 }
 
+/// Create an `AuthenticatorFunctionRefV1` for an authenticate function defined in the
+/// iota-framework package itself (package address @0x2), bypassing metadata validation.
+/// For internal system use only — the package address is known at genesis time.
+public(package) fun create_for_framework<Account: key>(
+    module_name: ascii::String,
+    function_name: ascii::String,
+): AuthenticatorFunctionRefV1<Account> {
+    AuthenticatorFunctionRefV1 {
```
miker83z: Better to use `create_auth_function_ref_v1_inner` instead of manual packing of `AuthenticatorFunctionRefV1`

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,117 @@
...
+///   0x00 Ed25519   | 0x01 Secp256k1 | 0x02 Secp256r1
+///   0x03 MultiSig  | 0x06 Passkey
+///
+/// Scheme validity and byte-length are validated via `iota::public_key::create`.
+/// Aborts with `EAddressMismatch` if the key does not derive to the sender, or
+/// `EAlreadyClaimed` if the address was already claimed.
+public fun claim(
+    registry: &mut ClaimRegistry,
+    scheme: SignatureScheme,
+    raw_bytes: vector<u8>,
```
miker83z: I think it is better to use `PublicKey` directly here, instead of `scheme` + `raw_bytes`. Then we could add an additional `claim_with_raw_bytes` which takes only `raw_bytes` as input (plus `registry`and `ctx`) and constructs `SignatureScheme` first, then `let pk = public_key::create(scheme, raw_bytes);`, and then calls `claim(registry, pk, ctx)`;

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,117 @@
...
+
+// === Internal helpers ===
+
+/// Mirrors Rust `IotaAddress::from(&PublicKey)`:
+///   Ed25519:   Blake2b256(pubkey)           — no flag prefix
+///   Secp256k1: Blake2b256([0x01] || pubkey)
+///   Secp256r1: Blake2b256([0x02] || pubkey)
+///   MultiSig:  Blake2b256([0x03] || pubkey)
+///   Passkey:   Blake2b256([0x06] || pubkey)
+fun derive_address(pk: &PublicKey): address {
```
miker83z: move this utility in the module where the `PublicKey` type is declared

crates/iota-framework/packages/iota-framework/sources/account_abstraction/authenticator_function.move
miker83z: Is this still used?

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,117 @@
...
+///
+/// Scheme validity and byte-length are validated via `iota::public_key::create`.
+/// Aborts with `EAddressMismatch` if the key does not derive to the sender, or
+/// `EAlreadyClaimed` if the address was already claimed.
+public fun claim(
+    registry: &mut ClaimRegistry,
+    scheme: SignatureScheme,
+    raw_bytes: vector<u8>,
+    ctx: &mut TxContext,
+): UID {
```
miker83z, line 65: Idea to test with a dedicated commit which could be reverted:
- here return (UID, AccountCreatorTicket) 
- the second is a value implementing the hot potato pattern
- create a function:
```move
public fun create_account_v1_destroy_account_creator<Account: key>(
    account: Account,
    authenticator: AuthenticatorFunctionRefV1<Account>,
    account_creator: AccountCreatorTicket,
) {
    assert!(account.uid() == account_creator.uid);
    account_creator.destroy();
    account_abstraction::create_account_v1(account, authenticator);
}
```
- account_creator.destroy() is a private function only called in here.

crates/iota-framework/packages/iota-framework/tests/test_account.move
```diff
@@ -0,0 +1,27 @@
...
+
+public fun create(
+    registry: &mut ClaimRegistry,
+    scheme: SignatureScheme,
+    public_key: vector<u8>,
+    ctx: &mut TxContext,
+) {
+    let scheme_flag = scheme.flag();
+    let uid = iota::claim_registry::claim(registry, scheme, public_key, ctx);
+    transfer::transfer(Account { id: uid, scheme: scheme_flag }, ctx.sender());
```
miker83z: test the happy path of [this idea](https://github.com/iotaledger/iota/pull/11094/changes#r3311955338) here

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
miker83z: Instead of claim_with_raw_bytes, we should create a function that takes as input a vector of u8 (prefixed_bytes), parses the first byte in order to validate/create a SignatureScheme and then creates the public key using the remaining bytes.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
miker83z: Consider different names for function and ticket (e.g., ClaimTicket, ClaimedTicket , ask Claude...)

miker83z: destroy could be replaced by just unpacking

crates/iota-types/src/transaction.rs
```diff
@@ -374,6 +374,9 @@ pub enum EndOfEpochTransactionKind {
     ChangeEpochV4(ChangeEpochV4),
     // IMPORTANT: new enum variants should be added at the end to preserve serialization
     // compatibility. DO NOT CHANGE THE ORDER OF EXISTING ENTRIES!
+    // ClaimRegistryCreate can be left at the end as long as `enable_claim_registry` is not
+    // enabled in the protocol config.
+    ClaimRegistryCreate,
```
Thoralf-M: move before the `IMPORTANT` comment

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,111 @@
...
+    assert!(!is_claimed(registry, derived_addr), EAlreadyClaimed);
+    df::add(&mut registry.id, derived_addr, true);
+    object::new_uid_from_hash(derived_addr)
+}
+
+/// Creates a `DummyAccount` owned by the sender.
+///
+/// Provided so that transactional tests can exercise the full `claim` flow
+/// directly from a PTB without publishing an external module.
+public fun test_claim_account(
```
Thoralf-M: Shouldn't this be test_only?

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,111 @@
...
+public struct DummyAccount has key {
+    id: UID,
+}
+
+// === Genesis ===
+
+/// Create and share the `ClaimRegistry` singleton. Called once during genesis.
+#[allow(unused_function)]
+fun create(ctx: &TxContext) {
+    assert!(ctx.sender() == @0x0, ENotGenesis);
```
Thoralf-M: I would use another error here, `ctx.sender() == @0x0` is not necessarily genesis, since it can also happen during protocol update?

piotrm50 (approved):

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: In this case `test_claim_account` function won’t be visible in iota-adapter-transcational-tests. 
This function was created like this since the original `claim` function cannot be used anymore in tests because of `public(package) modificatory `

crates/iota-types/src/transaction.rs
Dkwcs: Done

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Done

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Thoralf-M: But does it need to be in `iota-adapter-transcational-tests` can't you move tests that need it here? Feels wrong to make testing functions public

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -113,12 +133,40 @@ public fun raw_bytes(self: &PublicKey): &vector<u8> {
...
+///   Secp256r1: Blake2b256([0x02] || pubkey)
+///   MultiSig:  Blake2b256([0x03] || pubkey)
+///   Passkey:   Blake2b256([0x06] || pubkey)
+public fun to_iota_address(self: &PublicKey): address {
+    let scheme = self.scheme;
+    let raw = self.raw_bytes;
+    let data = if (scheme == signature_scheme::ed25519()) {
+        raw
+    } else {
+        let mut v = vector[scheme.flag()];
```
bingyanglin, line 147: The multisig address math here doesn't match the node. The node builds a multisig address one way, but this code builds it a different way, so the address this makes will never equal the real sender's address. That means multisig users can never claim. The test passes only because it checks the result against itself. Maybe build the multisig address the same way the node does, using what `validate_multisig_public_key` already reads?

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,110 @@
...
+#[test_only]
+public fun derive_address_for_testing(prefixed_bytes: &vector<u8>): address {
+    iota::public_key::from_prefixed_bytes(*prefixed_bytes).to_iota_address()
+}
+
+/// Creates a `DummyAccount` owned by the sender.
+///
+/// Provided so that transactional tests can exercise the full `claim` flow
+/// directly from a PTB without publishing an external module.
+public fun test_claim_account(
```
bingyanglin: These tests build the expected address with the same function they are testing. So they always pass, even if the address math is wrong, which is how the multisig bug stays hidden. The comments above already write out the real address per scheme. Maybe hardcode those addresses from values worked out separately in Rust and check against them? Or do I miss anything?

crates/iota-e2e-tests/tests/claim_registry_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+    pubkey_bytes: Vec<u8>,
+) -> anyhow::Result<iota_types::transaction::ProgrammableTransaction> {
+    let mut b = ProgrammableTransactionBuilder::new();
+    let reg = b.input(registry_arg)?;
+    let scheme_arg = b.pure(scheme)?;
+    let pk = b.pure(pubkey_bytes)?;
+    b.programmable_move_call(
+        IOTA_FRAMEWORK_PACKAGE_ID,
+        ident_str!("claim_registry").to_owned(),
+        ident_str!("claim").to_owned(),
```
bingyanglin: These tests call claim with `(registry, scheme, bytes)`, but `claim` now takes a `PublicKey` and is `public(package)`, so a PTB can't call it at all. The transaction fails early, not with the `MoveAbort` the tests wait for, so these tests will fail. Maybe rewrite them like the `.move` tests, using `from_prefixed_bytes` and `test_claim_account`, then simtest to test it again?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2775,7 +2789,7 @@ impl ProtocolConfig {
                 26 => {
                     // Introduce a module to allow Move code to query protocol
                     // feature flags at runtime.
-
+                    cfg.feature_flags.enable_claim_registry = true;
```
bingyanglin: The flag turns on at version 26, but the upgrade test goes from version 22 to 23 and then expects the registry to exist. At 23 the flag is still off, so nothing is created and the test crashes at the end. Also the flag is set for every chain, including mainnet. Maybe move the test to version 26, and check that turning it on for mainnet is what you want. Or do I miss anything?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -244,6 +244,11 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "ConsensusTransactionOrdering::is_none")]
     consensus_transaction_ordering: ConsensusTransactionOrdering,
 
+    // If true, the ClaimRegistry singleton is created (or already exists).
+    // Used to gate genesis creation and epoch-change creation for existing networks.
+    #[serde(skip_serializing_if = "is_false")]
```
bingyanglin: Why the new added flag/config is put in the middle no in the end? WIll this make it hard to trace?

crates/iota-protocol-config/src/lib.rs
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
Dkwcs: Fixed

crates/iota-protocol-config/src/lib.rs
Dkwcs: It’s an outcome of rebasing)
Fixed

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: It can be a temporary measure until we have default iota account ready.
If we make this function #[test_only] it also breaks e2e tests
@miker83z

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -80,6 +82,24 @@ public struct PublicKey has copy, drop, store {
...
+/// 32 bytes for Ed25519, 33 bytes (compressed) for Secp256k1 / Secp256r1 / Passkey,
+/// BCS-encoded `MultiSigPublicKey` for MultiSig (1–10 signers, threshold > 0, total weight ≥ threshold).
+///
+/// Aborts if `prefixed_bytes` is empty, if the flag byte is not a recognized scheme, if the
+/// remaining byte length does not match the scheme, or if a MultiSig payload fails structural validation.
+public fun from_prefixed_bytes(prefixed_bytes: vector<u8>): PublicKey {
+    assert!(!prefixed_bytes.is_empty(), EPublicKeyBytesEmpty);
+    let mut bcs = bcs::new(prefixed_bytes);
+    let flag = bcs.peel_u8();
+    let raw_bytes = bcs.into_remainder_bytes();
```
miker83z: This is unnecessary, just call `let flag = prefixed_bytes.remove(0);` and that's it

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
miker83z: Just remove this, the dummy type and the e2e tests. Create an issue to implement the tests after https://github.com/iotaledger/iota/pull/11728 is merged.

crates/iota-types/src/transaction.rs
```diff
@@ -372,8 +372,11 @@ pub enum EndOfEpochTransactionKind {
     ChangeEpochV2(ChangeEpochV2),
     ChangeEpochV3(ChangeEpochV3),
     ChangeEpochV4(ChangeEpochV4),
+    ClaimRegistryCreate,
```
thibault-martinez, line 375: Is this a pretty old branch? This should be added to the SDK

thibault-martinez: Ah, I see you have a TODO https://github.com/iotaledger/iota/pull/11094/changes#diff-1827bb0288655ef413fc7ec440d39da1b8874827bb54a358a5354b2b2145abddR707

miker83z: We will update the feature branch directly after merging the PRs that we have currently open

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -113,12 +133,76 @@ public fun raw_bytes(self: &PublicKey): &vector<u8> {
...
+
+fun scheme_from_flag(flag: u8): SignatureScheme {
+    if (flag == 0x00) signature_scheme::ed25519()
+    else if (flag == 0x01) signature_scheme::secp256k1()
+    else if (flag == 0x02) signature_scheme::secp256r1()
+    else if (flag == 0x03) signature_scheme::multisig()
+    else if (flag == 0x06) signature_scheme::passkey()
+    else abort EUnknownPublicKeyScheme
+}
+
```
valeriyr: this should be move to the `signature_scheme` module, like:
signature_scheme::try_from_byte or signature_scheme::from_byte

we shouldn't duplicate flag values

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -113,12 +131,76 @@ public fun raw_bytes(self: &PublicKey): &vector<u8> {
...
 
 // === Package Functions ===
 
 // === Private Functions ===
 
+/// Builds the hash preimage for a MultiSig address, matching the Rust node:
+///   [0x03] || threshold_le16 || (scheme_flag || pk_bytes || weight_u8)*
+///
+/// BCS stores signers before threshold; the hash puts threshold first.
+fun multisig_hash_input(raw_bytes: &vector<u8>): vector<u8> {
```
miker83z: Maybe `multisig_to_hash_input`?

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -113,12 +131,76 @@ public fun raw_bytes(self: &PublicKey): &vector<u8> {
...
 // === Admin Functions ===
 
 // === Package Functions ===
 
 // === Private Functions ===
 
+/// Builds the hash preimage for a MultiSig address, matching the Rust node:
+///   [0x03] || threshold_le16 || (scheme_flag || pk_bytes || weight_u8)*
+///
+/// BCS stores signers before threshold; the hash puts threshold first.
```
miker83z:
```suggestion
/// Creates the bytes used to derive an address from a multisig PublicKey. A MultiSig address
/// is defined as the 32-byte Blake2b hash of serializing the flag, the
/// threshold, concatenation of all n flag, public keys and
/// its weight. `flag_MultiSig || threshold || flag_1 || pk_1 || weight_1
/// || ... || flag_n || pk_n || weight_n`.
```

Taken from the rust side.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -113,12 +131,76 @@ public fun raw_bytes(self: &PublicKey): &vector<u8> {
...
+        };
+        signer_bytes.push_back(bcs.peel_u8()); // weight
+        i = i + 1;
+    };
+
+    let threshold = bcs.peel_u16();
+    let mut data = vector[0x03u8, threshold as u8, (threshold >> 8) as u8];
+    data.append(signer_bytes);
+    data
+}
```
miker83z:
```suggestion
fun multisig_to_hash_input(mut raw_bytes: vector<u8>): vector<u8> {
    let threshold_high = raw_bytes.pop_back();
    let threshold_low = raw_bytes.pop_back();

    let mut bcs = bcs::new(raw_bytes);
    let num_signers = bcs.peel_vec_length();

    // multisig_flag_u8 || threshold_le16 || (pk_flag_u8 || pk_bytes || weight_u8)*
    let mut data = vector[signature_scheme::multisig().flag(), threshold_low, threshold_high];
    num_signers.do!(|_| {
        let tag = bcs.peel_enum_tag();

        let (key_len, scheme_flag) = if (tag == MULTISIG_KEY_TAG_ED25519) (
            ED25519_PUBLIC_KEY_LENGTH,
            signature_scheme::ed25519().flag(),
        ) else if (tag == MULTISIG_KEY_TAG_SECP256K1) (
            SECP256_PUBLIC_KEY_LENGTH,
            signature_scheme::secp256k1().flag(),
        ) else if (tag == MULTISIG_KEY_TAG_SECP256R1) (
            SECP256_PUBLIC_KEY_LENGTH,
            signature_scheme::secp256r1().flag(),
        ) else if (tag == MULTISIG_KEY_TAG_PASSKEY) (
            SECP256_PUBLIC_KEY_LENGTH,
            signature_scheme::passkey().flag(),
        ) else abort EUnknownPublicKeyScheme;

        // pk_flag_u8 || pk_bytes || weight_u8
        data.push_back(scheme_flag);
        key_len.do!(|_| {
            data.push_back(bcs.peel_u8());
        });
        data.push_back(bcs.peel_u8());
    });

    data
}
```
I moved the threshold pop above such that you do not have 1 extra append at the end and spend more gas. Then, I cleaned the signature flags part.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+/// On-chain registry for claiming addresses.
```
miker83z, line 4:
```suggestion
/// On-chain registry for claiming addresses.
/// This registry is used to allow anyone owning a public-key/private-key keypair to claim the address derived from that, such that it can be used as UID of a new object on-chain. 
```

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,91 @@
...
+const EAddressMismatch: vector<u8> =
+    b"The public key does not correspond to the transaction sender address.";
+
+#[error(code = 1)]
+const EAlreadyClaimed: vector<u8> =
+    b"This address has already been claimed.";
+
+#[error(code = 4)]
+const ENotSystemAddress: vector<u8> =
+    b"ClaimRegistry can only be created by the system address (@0x0).";
```
miker83z:
```suggestion
    b"ClaimRegistry can only be created in a system transaction.";
```

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,91 @@
...
+        id: object::new_uid_from_hash(@0x10),
+    });
+}
+
+// === Claim ===
+
+/// Mark `ctx.sender()` as claimed and return a deterministic `UID` for the new
+/// account object.
+///
+/// The returned `UID` is a hot potato — it has no `drop` ability and must be
```
miker83z: Update the comment

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,91 @@
...
+
+// === Test only ===
+
+#[test_only]
+public fun create_for_testing(ctx: &mut TxContext) {
+    create(ctx);
+}
+
+#[test_only]
+public fun derive_address_for_testing(prefixed_bytes: &vector<u8>): address {
```
miker83z: move this function to public_key.move

crates/iota-framework/packages/iota-framework/tests/claim_registry_tests.move
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// ============================================================
+// to_iota_address — cross-checked against Rust-computed vectors
+// ============================================================
+
+// These constants are independent of the Move code under test: they were
+// computed by the Rust node using fastcrypto's Blake2b256.
+#[test]
+fun test_to_iota_address_vectors() {
```
miker83z: this is a copy of the one in public_key_tests and shouldn't be here

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -1937,6 +1942,24 @@ mod checked {
...
+            .move_call(
+                IOTA_FRAMEWORK_PACKAGE_ID,
+                CLAIM_REGISTRY_MODULE_NAME.to_owned(),
+                CLAIM_REGISTRY_CREATE_FUNCTION_NAME.to_owned(),
+                vec![],
+                vec![],
+            )
+            .expect("Unable to generate claim_registry_create transaction!");
+        Ok(builder.finish())
+    }
```
miker83z, line 1961:
```suggestion
    fn setup_claim_registry_create(
        mut builder: ProgrammableTransactionBuilder,
    ) -> ProgrammableTransactionBuilder {
        builder
            .move_call(
                IOTA_FRAMEWORK_PACKAGE_ID,
                CLAIM_REGISTRY_MODULE_NAME.to_owned(),
                CLAIM_REGISTRY_CREATE_FUNCTION_NAME.to_owned(),
                vec![],
                vec![],
            )
            .expect("Unable to generate claim_registry_create transaction!");
        builder
    }
```
This should be tested, but I think this is the way to do it, otherwise the. builder will conflic with the ChangeEpoch command. The caller site above should be `builder = setup_claim_registry_create(builder)`

iota-execution/latest/iota-verifier/src/id_leak_verifier.rs
```diff
@@ -103,7 +104,17 @@ const IOTA_BRIDGE_CREATE: FunctionIdent = (
     ident_str!("bridge"),
     ident_str!("create"),
 );
-const FRESH_ID_FUNCTIONS: &[FunctionIdent] = &[OBJECT_NEW, OBJECT_NEW_UID_FROM_HASH, TS_NEW_OBJECT];
+const CLAIM_REGISTRY: FunctionIdent = (
```
miker83z: CLAIM_REGISTRY is the module name. I think the not so nice sounding but correct name should be CLAIM_REGISTRY_CLAIM

Dkwcs: Claude suggested it for me as well, but it sounds really weird, doesn't it?

iota-execution/latest/iota-adapter/src/execution_engine.rs
Dkwcs: It was a first version I tried, then I changed it to `Ok(builder.finish())`

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/tests/claim_registry_tests.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
Dkwcs: Fixed

iota-execution/latest/iota-verifier/src/id_leak_verifier.rs
miker83z: It is, but just CLAIM_REGISTRY is not correct either

miker83z (approved):

crates/iota-framework/packages/iota-framework/sources/account_abstraction/signature_scheme.move
```diff
@@ -56,6 +59,19 @@ public fun passkey(): SignatureScheme {
...
+/// Aborts with `EUnknownScheme` if `flag` is not one of the recognized values
+/// (0x00 Ed25519, 0x01 Secp256k1, 0x02 Secp256r1, 0x03 MultiSig, 0x06 Passkey).
+public fun from_flag(flag: u8): SignatureScheme {
+    assert!(
+        flag == ED25519 || flag == SECP256K1 || flag == SECP256R1
+            || flag == MULTISIG || flag == PASSKEY,
+        EUnknownScheme,
+    );
+    SignatureScheme { flag }
+}
```
valeriyr, line 73: This functon needs to be covered with tests

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// === Genesis ===
+
+/// Create and share the `ClaimRegistry` singleton. Called once during genesis.
+#[allow(unused_function)]
+fun create(ctx: &TxContext) {
+    assert!(ctx.sender() == @0x0, ENotSystemAddress);
+    transfer::share_object(ClaimRegistry {
+        id: object::new_uid_from_hash(@0x10),
```
valeriyr: It would be nice at least to create a constant for the address.
I think it is not really necessary, but you can consider adding a function to the object module like it is done for some other objects:
https://github.com/iotaledger/iota/blob/2628cb29977662eb7f05d4063397f8e4a2328e02/crates/iota-framework/packages/iota-framework/sources/object.move#L107

crates/iota-genesis-builder/src/lib.rs
```diff
@@ -550,6 +551,13 @@ impl Builder {
 
         assert!(unsigned_genesis.has_coin_deny_list_object());
 
+        let protocol_config = get_genesis_protocol_config(protocol_version.into());
+        if protocol_config.create_claim_registry_in_genesis() {
+            assert!(unsigned_genesis.has_claim_registry_object());
+        } else {
+            assert!(!unsigned_genesis.has_claim_registry_object());
+        }
```
valeriyr:
```suggestion
        assert_eq!(protocol_config.create_claim_registry_in_genesis(), unsigned_genesis.has_claim_registry_object())
```

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1371,6 +1376,15 @@ impl ProtocolConfig {
...
     }
 
+    pub fn enable_claim_registry(&self) -> bool {
+        self.feature_flags.enable_claim_registry
+    }
+
+    // this function only exists for readability in the genesis code.
+    pub fn create_claim_registry_in_genesis(&self) -> bool {
+        self.enable_claim_registry()
+    }
```
valeriyr: Do we really need both of them?
Looks clear that we need to create an instance if the claim registry is enabled during genesis.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2773,10 +2787,9 @@ impl ProtocolConfig {
                     cfg.max_age_of_jwk_in_epochs = None;
                 }
                 26 => {
-                    // Introduce a module to allow Move code to query protocol
-                    // feature flags at runtime.
```
valeriyr, line 2777: This comment could be useful because the feature is not gated with a feature flag and it helps to track the changes.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -0,0 +1,84 @@
...
+
+#[error(code = 0)]
+const EAddressMismatch: vector<u8> =
+    b"The public key does not correspond to the transaction sender address.";
+
+#[error(code = 1)]
+const EAlreadyClaimed: vector<u8> = b"This address has already been claimed.";
+
+#[error(code = 4)]
+const ENotSystemAddress: vector<u8> = b"ClaimRegistry can only be created in a system transaction.";
```
valeriyr, line 30: Do we have any specific logic related to the error numbers?
I sometimes split the errors into groups by 10 to make it possible to add errors logically in the future.
It is not clear why we have 0, 1 and 4.

valeriyr (approved):
Looks good!

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: The gap exists because codes 2 and 3 were removed earlier (they belonged to DummyAccount/test_claim_account which was deleted).

crates/iota-protocol-config/src/lib.rs
Dkwcs: Fixed

crates/iota-genesis-builder/src/lib.rs
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
Dkwcs: Fixed

crates/iota-framework/packages/iota-framework/sources/account_abstraction/signature_scheme.move
Dkwcs: Fixed

DaughterOfMars (approved):

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -702,6 +702,11 @@ impl From<crate::transaction::EndOfEpochTransactionKind> for EndOfEpochTransacti
                     adjust_rewards_by_score: change_epoch_v4.adjust_rewards_by_score,
                 })
             }
+            crate::transaction::EndOfEpochTransactionKind::ClaimRegistryCreate => {
+                // ClaimRegistryCreate is not yet present in iota-sdk-types.
+                // TODO: add the variant to iota-rust-sdk.
+                todo!("ClaimRegistryCreate is not yet supported in iota-sdk-types")
```
Thoralf-M, line 708: Is it tracked somewhere that it gets added?

Thoralf-M (approved):

crates/iota-types/src/iota_sdk_types_conversions.rs
Dkwcs: [Here](https://github.com/iotaledger/iota/issues/11811)

kodemartin (approved):

PR #11102, opened 2026-04-02
upstream(core): Fix debug assert

jkrvivian:
## Description of change

- Upstream range: [v1.48.4, v1.49.2)
- Port commit:
  - [MystenLabs/sui@105d5273d3](https://github.com/MystenLabs/sui/commit/105d5273d3af474fbcecb58a40c58e561642d962)
- Description:

slow.rs workload transactions shouldn't abort, but they can be cancelled

Part of #11101

### DIFF
--- a/crates/iota-benchmark/src/lib.rs
+++ b/crates/iota-benchmark/src/lib.rs
@@ -39,6 +39,7 @@ use iota_types::{
     committee::{Committee, EpochId},
     crypto::AuthorityStrongQuorumSignInfo,
     effects::{CertifiedTransactionEffects, TransactionEffectsAPI, TransactionEvents},
+    execution_status::ExecutionFailureStatus,
     gas::GasCostSummary,
     gas_coin::GasCoin,
     iota_system_state::{IotaSystemStateTrait, iota_system_state_summary::IotaSystemStateSummary},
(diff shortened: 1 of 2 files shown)


crates/iota-benchmark/src/lib.rs
```diff
@@ -139,6 +140,28 @@ impl ExecutionEffects {
...
     }
 
+    pub fn is_cancelled(&self) -> bool {
+        match self {
+            ExecutionEffects::CertifiedTransactionEffects(effects, ..) => {
+                match effects.data().status() {
+                    iota_types::execution_status::ExecutionStatus::Success => false,
+                    iota_types::execution_status::ExecutionStatus::Failure {
+                        error:
+                            ExecutionFailureStatus::ExecutionCancelledDueToSharedObjectCongestion {
```
semenov-vladyslav, line 150: I think we need to also cover `ExecutionCancelledDueToSharedObjectCongestionV2`

crates/iota-benchmark/src/lib.rs
```diff
@@ -139,6 +140,28 @@ impl ExecutionEffects {
...
+                                ..
+                            },
+                        ..
+                    } => true,
+                    _ => false,
+                }
+            }
+            ExecutionEffects::IotaTransactionBlockEffects(iota_tx_effects) => {
+                let status = format!("{}", iota_tx_effects.status());
+                status.contains("ExecutionCancelledDueToSharedObjectCongestion")
```
semenov-vladyslav, line 163: Here too, probably

semenov-vladyslav (changes requested):

crates/iota-benchmark/src/lib.rs
jkrvivian: V2 has `ExecutionCancelledDueToSharedObjectCongestion` as substring, so here it is included.

semenov-vladyslav (approved):

muXxer (approved):

PR #11126, opened 2026-04-07
fix(gRPC): SPAM protection for metrics

bingyanglin:
# Description of change

Prevent unbounded metric cardinality in the gRPC server by labelling unknown/non-gRPC HTTP paths as `"SPAM"` instead of recording arbitrary request paths as individual metric labels. Use exact path to identify if the request is the spam.

Test steps follow:
-  Run the local node: `RUST_LOG="off,iota_node=info" cargo run --bin iota-localnet start --force-regenesis --with-faucet --with-grpc`
- Ping the node by using both valid endpoint or spam, like `curl --http2-prior-knowledge -X POST -H "content-type: application/grpc" http://localhost:50051/iota.grpc.v1.ledger_service.LedgerService/GetCheckpoint`, etc.
- Check the metrics by ` curl -s http://127.0.0.1:<YOUR_METRICS_PORT>/metrics | grep "grpc_server" | grep -v "^#"r`
- Auto-generate `VALIDATOR_METHOD_PATHS` from the service definition in build.rs so the allowlist stays in sync when methods are added or removed.
- Rename metric names to use consistent `authority_grpc_*` / `node_grpc_*` prefixes.

No

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Authority gRPC metrics were renamed for consistency (`inflight_grpc` => `authority_grpc_inflight_requests`, `grpc_requests` => `authority_grpc_requests`, `grpc_request_latency` => `authority_grpc_request_latency`)
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-grpc-server/src/metrics.rs
```diff
@@ -107,7 +128,16 @@ where
...
-        let method = req.uri().path().to_owned();
+        let raw_path = req.uri().path();
+        let method = if self
+            .service_prefixes
+            .iter()
+            .any(|prefix| raw_path.starts_with(prefix))
+        {
+            raw_path.to_owned()
+        } else {
+            SPAM_LABEL.to_owned()
```
jkrvivian: Instead of labeling to `SPAM`, can we simply ignore those and not record them?

bingyanglin: Good question. I still record them under a single "SPAM" label because I want it to be consistent with the existing JSON-RPC metrics design, which can preserves visibility into junk traffic. And use the SPAM category can fix the issue of metric cardinality explosion.

semenov-vladyslav: Good point. If it's spam we don't even need to call the inner future (and measure its latency) since it will fail anyway. Maybe just introduce a SPAM metric counter and increment it?

bingyanglin: I think it is still informative to record the spam (like knowing if the users still call deprecated APIs). Now they are all collected under the SPAM label so no cardinality explosion.

muXxer: I agree with @semenov-vladyslav. Do we even need to call the inner if we already know it's non existent? We can just increase a spam counter and return I guess?

bingyanglin: After discussion, I think no need to record the attacker vector in our gRPC metrics, so we can just use counter for it. Fixed it at [0fcae9f](https://github.com/iotaledger/iota/pull/11126/commits/0fcae9fe61aa69085c9088445ef4816a39a43882).

crates/iota-grpc-server/src/metrics.rs
```diff
@@ -89,25 +106,48 @@ impl<S> Layer<S> for GrpcMetricsLayer {
...
     fn call(&mut self, req: http::Request<ReqBody>) -> Self::Future {
-        let method = req.uri().path().to_owned();
+        let raw_path = req.uri().path();
+
+        if !self.known_methods.contains(raw_path) {
+            // SPAM: bump counter and reject immediately without calling the
+            // inner service, avoiding unnecessary router work.
+            self.metrics
+                .num_requests
+                .with_label_values(&[SPAM_LABEL, "Unimplemented"])
```
semenov-vladyslav, line 135: Why `Unimplemented`? Why not `Unknown`, `NotFound` or even `Ok` (since the response status is OK)?

crates/iota-grpc-server/src/metrics.rs
```diff
@@ -89,25 +106,48 @@ impl<S> Layer<S> for GrpcMetricsLayer {
...
+                .inc();
+
+            let response = http::Response::builder()
+                .status(http::StatusCode::OK)
+                .header("content-type", "application/grpc")
+                .header("grpc-status", "12") // Unimplemented
```
muXxer: can't we use the `GRPC_STATUS` constant here from tonic? Also there is a Code::Unimplemented available there, or even a `unimplemented` function, that returns a Status. Also, I think we shouldn't return "http::StatusCode::OK" to the caller, if it is not available.

semenov-vladyslav: Should there be a relation between status and "grpc-status"? It looks like "grpc-status" duplicates status in a way. "grpc-status" seems to be taken from status: `grpc_status_from_response(response)`.

crates/iota-grpc-server/src/metrics.rs
bingyanglin: Conventionally `Unimplemented` is what tonic itself returns when you call an unknown method. `NotFound` in gRPC means "resource not found" (like a missing DB record), not "endpoint doesn't exist". According to the [gRPC status code doc](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) I think `Unimplemented` is correct?

crates/iota-grpc-server/src/metrics.rs
bingyanglin: I guess they're not duplicates, gRPC uses HTTP 200 for all responses, and puts the real error in the `grpc-status` header. That's the gRPC protocol design. Now I use `tonic::Status::unimplemented("").into_http()` which handles both.

bingyanglin: Thanks, fixed at [2983775](https://github.com/iotaledger/iota/pull/11126/commits/2983775a83c88a7ccde805532e8f63e6a0fb9d08). Now I am using `tonic::Status::unimplemented("").into_http()` which handles everything properly. About HTTP 200: gRPC always returns HTTP 200 at the transport layer and puts the actual error code in the `grpc-status` header, so I guess `OK` is fine? (see the [gRPC doc](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), (... `This will prevent other HTTP/2 clients from interpreting a gRPC error response, which uses status 200 (OK), as successful`....))

bingyanglin: Also maybe for malicious spammer return `OK` with `unimplmemented` is better so the attacker can learn nothing?

muXxer (approved):

jkrvivian (approved):

semenov-vladyslav (approved):

muXxer (approved):

crates/iota-grpc-server/src/metrics.rs
alexsporn: No, `Unimplemented` is used as the default response for valid services that are not yet implemented on the server. Responses for weird paths should respond with `NotFound`

bingyanglin: This is a good point, after more investigation, Sui also uses `Unimplemented`, see [sui-node/src/metrics.rs](https://github.com/MystenLabs/sui/blob/main/crates/sui-node/src/metrics.rs), they use `UNKNOWN_PATH` label for metrics sanitization, but does not override the gRPC status code. The status code that reaches metrics is still whatever tonic returns which is `Unimplemented` for unknown paths.

From the official gRPC document [here](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), see the table, for the Case `Method not found at server` we should return `UNIMPLEMENTED` from the server, which is conventional. See also the following references
- [grpc/grpc-go/server.go](https://github.com/grpc/grpc-go/blob/master/server.go), search for `codes.Unimplemented`.
- [grpc/grpc-java ServerImpl.java](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ServerImpl.java), search for `UNIMPLEMENTED`,
- [hyperium/tonic/src/service/router.rs](https://github.com/hyperium/tonic/blob/master/tonic/src/service/router.rs), check the default fallback for unmatched routes.

Also see official protobuf at [google/rpc/code.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto):
```rust
// Some requested entity (e.g., file or directory) was not found.
//
// Note to server developers: if a request is denied for an entire class
// of users, such as gradual feature rollout or undocumented allowlist,
// `NOT_FOUND` may be used. If a request is denied for some users within
// a class of users, such as user-based access control, `PERMISSION_DENIED`
// must be used.
NOT_FOUND = 5;
```
So `NOT_FOUND` is about a "requested entity" (file, directory, DB record, etc.), not about a method/operation. 

So I guess `Unimplemented` is correct in our case? Or do I miss anything?

alexsporn:
```
grpc_requests{path="/secrets.js",status="Unimplemented"} 1
grpc_requests{path="/secrets.json",status="Unimplemented"} 1
grpc_requests{path="/security.txt",status="Unimplemented"} 3
grpc_requests{path="/sellers.json",status="Unimplemented"} 2
grpc_requests{path="/server/.git/config",status="Unimplemented"} 7
grpc_requests{path="/service-account.json",status="Unimplemented"} 1
grpc_requests{path="/service-worker.js",status="Unimplemented"} 1
grpc_requests{path="/serviceaccount.json",status="Unimplemented"} 1
grpc_requests{path="/settings.js",status="Unimplemented"} 1
grpc_requests{path="/settings.json",status="Unimplemented"} 1
grpc_requests{path="/settings/.env",status="Unimplemented"} 7
grpc_requests{path="/setup.cgi",status="Unimplemented"} 3
grpc_requests{path="/sitemap.xml",status="Unimplemented"} 16
grpc_requests{path="/sms_mp/.sms_aut",status="Unimplemented"} 1
grpc_requests{path="/solr/admin/cores",status="Unimplemented"} 2
grpc_requests{path="/solr/admin/info/system",status="Unimplemented"} 2
grpc_requests{path="/src/.git/config",status="Unimplemented"} 7
grpc_requests{path="/src/app.config.js",status="Unimplemented"} 1
grpc_requests{path="/src/config.js",status="Unimplemented"} 1
grpc_requests{path="/src/constants.js",status="Unimplemented"} 1
```

Not only methods/operations land on these metrics, but I don't want to push it

PR #11151, opened 2026-04-07
feat(iota-core): pre-consensus soft locking for owned objects

### LINKED ISSUE #10373: pcool: Pre-consensus soft-locking (optional)
Optional defense layer: prevent the same validator from accepting two transactions that conflict on owned objects at pre-submission time.

Post-consensus conflict resolution (3-tier lock check) already handles this correctly. This task adds a validator-local soft lock at submission time as defense-in-depth.

Design considerations:
- In-memory lock table with TTL (how long to hold before considering a tx lost)
- Garbage collection for transactions dropped by consensus or load shedding
- Epoch-freeze risk: if a transaction is never executed, the object could be soft-locked until epoch end
- "Slightly different transaction" case: same owned objects, different tx from the same wallet

Pick up only after the batch API (#10748) is stable and the team decides the complexity is justified.

Depends on: #10748

**Files:** `authority_server.rs`, `execution_cache/object_locks.rs`

piotrm50:
# Description of change

Add an in-memory defense-in-depth mechanism that prevents a validator from accepting two conflicting transactions on the same owned objects at pre-submission time in the pcool (white-flag) flow. This reduces wasted consensus bandwidth and client-visible latency from conflicts that would only be caught post-consensus.

**Key design:**
- `DashMap`-based soft lock table keyed on `ObjectRef` with TTL (12s, derived from 4× consensus GC window)
- Same-digest resubmission is idempotent (no conflict) — allows retries freely
- Active GC: locks released when tx is processed by consensus or dropped post-consensus
- Passive TTL sweep every 10s as safety net for transactions garbage-collected by consensus
- Epoch-scoped: cleared at epoch boundary

**Integration points:**
- `handle_submit_transaction_impl` — acquire locks after validation, release on submit failure
- `authority_per_epoch_store` — release locks for all user transactions (both accepted and rejected) after consensus output is quarantined, ensuring permanent locks are visible before soft locks are freed
- `PreConsensusSoftLocks` owned by `ValidatorComponents`, passed directly to epoch store and gRPC service (not stored in `ConsensusAdapter`)

Closes #10373

9 unit tests in `soft_lock.rs` covering: acquire/release lifecycle, conflict detection, idempotent same-digest, TTL expiry, partial rollback, sweep, clear.
All 12 existing `server_tests` pass (including pcool submit flow tests).


crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -212,14 +217,39 @@ impl ValidatorService {
...
+        // Different-digest conflict on same objects → rejected.
+        // Placed after the reconfig check so we never acquire locks that would
+        // need releasing on the epoch-halt path.
+        if let Err(e) = soft_locks.try_acquire(tx_digest, &owned_objects) {
+            metrics.num_rejected_tx_soft_lock_conflict.inc();
+            return Err(tonic::Status::from(e));
+        }
+        metrics
+            .soft_lock_table_size
+            .set(soft_locks.lock_count() as i64);
```
bingyanglin, line 272: How do you think to track the count with an `AtomicUsize` so you don't need the lock for metrics?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -665,6 +665,16 @@ pub struct AuthorityPerEpochStore {
...
 
+    /// Pre-consensus soft locks for owned objects (pcool / white-flag flow).
+    ///
+    /// Set via `OnceCell` rather than passed through the constructor because
+    /// the same `Arc` instance must be shared with the gRPC `ValidatorService`
+    /// (which acquires locks pre-consensus). The gRPC server is created once
+    /// and persists across epochs, while a new `AuthorityPerEpochStore` is
+    /// created on each epoch change — so the wiring happens after construction
+    /// in `start_epoch_specific_validator_components`. Left empty in tests.
+    soft_locks: OnceCell<Arc<crate::authority_server::soft_lock::PreConsensusSoftLocks>>,
```
roman1e2f5p8s: Any good reason for not importing this in the already present `use crate::{ ... }` block? There are two usages of `PreConsensusSoftLocks` in this file.

```suggestion
    soft_locks: OnceCell<Arc<PreConsensusSoftLocks>>,
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1738,6 +1749,17 @@ impl AuthorityPerEpochStore {
             .await
     }
 
+    /// Sets the pre-consensus soft lock table. Called once during validator
+    /// setup when white-flag flow is enabled.
+    pub fn set_soft_locks(
+        &self,
+        soft_locks: Arc<crate::authority_server::soft_lock::PreConsensusSoftLocks>,
```
```suggestion
        soft_locks: Arc<PreConsensusSoftLocks>,
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -665,6 +665,16 @@ pub struct AuthorityPerEpochStore {
     /// `Rejected` response instead of waiting for the gRPC deadline.
     dropped_tx_status_cache: super::dropped_tx_status_cache::DroppedTxStatusCache,
 
+    /// Pre-consensus soft locks for owned objects (pcool / white-flag flow).
+    ///
+    /// Set via `OnceCell` rather than passed through the constructor because
+    /// the same `Arc` instance must be shared with the gRPC `ValidatorService`
+    /// (which acquires locks pre-consensus). The gRPC server is created once
```
roman1e2f5p8s, line 666: Why setting via `OnceCell`: the first sentence alone reads like the full justification and stops early; the actual reason is in the second sentence, so at least glue both.

```suggestion
    /// (which acquires locks pre-consensus): the gRPC server is created once
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1738,6 +1749,17 @@ impl AuthorityPerEpochStore {
             .await
     }
 
+    /// Sets the pre-consensus soft lock table. Called once during validator
+    /// setup when white-flag flow is enabled.
```
roman1e2f5p8s: The docstring is misleading: the caller calls this unconditionally of the white-flag flow.
Also document the panic.


```suggestion
    /// Sets the pre-consensus soft lock table. Called once during validator
    /// setup when white-flag flow is enabled.
    ///
    /// # Panics
    /// Panics if called more than once on the same instance when
    /// white-flag flow is enabled.
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1738,6 +1749,17 @@ impl AuthorityPerEpochStore {
...
+    pub fn set_soft_locks(
+        &self,
+        soft_locks: Arc<crate::authority_server::soft_lock::PreConsensusSoftLocks>,
+    ) {
+        self.soft_locks
+            .set(soft_locks)
+            .expect("soft_locks should only be set once");
```
roman1e2f5p8s, line 1688: To align with the docstring

```suggestion
        if self.protocol_config().enable_white_flag_flow() {
            self.soft_locks
                .set(soft_locks)
                .expect("soft_locks should only be set once");
        }
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3309,15 +3331,19 @@ impl AuthorityPerEpochStore {
         // single pass: validates UserTransactionV1 transactions and resolves
         // lock conflicts before reordering. Deferred txs from previous commits
         // already have persistent locks, giving them natural precedence.
+        // Also collects all UserTransactionV1 digests for soft lock release
+        // after the consensus output is quarantined.
+        let mut soft_lock_release_tx_digests = Vec::new();
```
roman1e2f5p8s, line 3194: `soft_lock_release_tx_digests` does not need to be `mut` + make it explicit on white-flag flow:

```rust
let soft_lock_release_tx_digests = if enable_white_flag {
    ...

    soft_lock_digests
} else {
    Vec::new()
};
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3533,6 +3559,34 @@ impl AuthorityPerEpochStore {
...
+                }
+                None => {
+                    error!(
+                        count = soft_lock_release_tx_digests.len(),
+                        "white-flag flow produced soft-lock release digests but \
+                         soft_locks OnceCell was never set — wiring bug in \
+                         start_epoch_specific_validator_components"
+                    );
+                }
+            }
```
roman1e2f5p8s: `if let` reads better on two-arm `Option`:

```suggestion
            if let Some(soft_locks) = self.soft_locks.get() {
                for digest in &soft_lock_release_tx_digests {
                    soft_locks.release(digest);
                }
            } else {
                error!(
                    count = soft_lock_release_tx_digests.len(),
                    "white-flag flow produced soft-lock release digests but \
                         soft_locks OnceCell was never set — wiring bug in \
                         start_epoch_specific_validator_components"
                );
            }
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3533,6 +3559,34 @@ impl AuthorityPerEpochStore {
...
+        // was never wired, that's a startup bug — locks would leak until TTL
+        // expiry. Log at `error!` so alerts fire; tests that exercise the
+        // white-flag path without a validator service take this branch
+        // harmlessly.
+        if !soft_lock_release_tx_digests.is_empty() {
+            match self.soft_locks.get() {
+                Some(soft_locks) => {
+                    for digest in &soft_lock_release_tx_digests {
+                        soft_locks.release(digest);
+                    }
```
roman1e2f5p8s: `release` acquires a mutex, so this loop does `N = soft_lock_release_tx_digests.len()` lock/unlock cycles on the same mutex every single commit. Worth adding a `release_for_batch()` method that calls `let mut inner = self.inner.lock();` only once per commit and releases soft locks for the batch of txs. The call site would also become a signle `soft_locks.release_for_batch(&soft_lock_release_tx_digests);`

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3533,6 +3559,34 @@ impl AuthorityPerEpochStore {
...
+        // If the white-flag flow produced digests to release but the OnceCell
+        // was never wired, that's a startup bug — locks would leak until TTL
+        // expiry. Log at `error!` so alerts fire; tests that exercise the
+        // white-flag path without a validator service take this branch
+        // harmlessly.
+        if !soft_lock_release_tx_digests.is_empty() {
+            match self.soft_locks.get() {
+                Some(soft_locks) => {
+                    for digest in &soft_lock_release_tx_digests {
+                        soft_locks.release(digest);
```
roman1e2f5p8s: Shouldn't the `soft_lock_table_size` metric be updated after the removal? Maybe worth storing `IntGauge` inside `PreConsensusSoftLocks` and letting every mutation method update it, so the call sites would not care where the metric should be updated?

crates/iota-core/src/authority_server/metrics.rs
```diff
@@ -33,6 +34,8 @@ pub struct ValidatorServiceMetrics {
     pub forwarded_header_invalid: IntCounter,
     pub forwarded_header_not_included: IntCounter,
     pub client_id_source_config_mismatch: IntCounter,
+    pub num_rejected_tx_soft_lock_conflict: IntCounter,
+    pub soft_lock_table_size: IntGauge,
```
roman1e2f5p8s, line 38: How about adding other metrics - counter of sweep TTL expirations and counter for successful acquires - not fully sure if they could be valuable though?

crates/iota-core/src/authority_server/mod.rs
```diff
@@ -70,9 +74,49 @@ impl ValidatorService {
...
     }
 
+    /// Spawns a background task that periodically sweeps expired soft locks
+    /// and refreshes the `soft_lock_table_size` gauge, so Prometheus scrapes
+    /// see a fresh value even under low transaction load.
+    ///
+    /// The task holds only a `Weak` reference to the lock table and exits
+    /// automatically once all strong `Arc` owners have been dropped (e.g. when
+    /// the node stops being a validator). No explicit `abort()` is needed.
+    pub fn spawn_soft_lock_sweep(
```
roman1e2f5p8s: Any explanation why this function is in the `impl ValidatorService` block? I think the proper place for it to live is `soft_lock.rs`.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -96,6 +100,9 @@ pub async fn validate_and_resolve_conflicts(
     let mut current_commit_locks: HashMap<ObjectRef, LockDetails> = HashMap::new();
     // Index-parallel keep flags: true = keep, false = remove.
     let mut keep = vec![true; transactions.len()];
+    // All UserTransactionV1 digests seen in this commit (both kept and dropped),
+    // used by the caller to release pre-consensus soft locks.
+    let mut all_user_tx_digests: Vec<TransactionDigest> = Vec::new();
```
roman1e2f5p8s:
```suggestion
    let mut all_user_tx_digests = Vec::with_capacity(transactions.len());
```

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -75,19 +75,23 @@ use crate::{
...
+///   (both kept and dropped). Used by the caller to release pre-consensus soft
+///   locks.
 pub async fn validate_and_resolve_conflicts(
     authority_state: &AuthorityState,
     epoch_store: &Arc<AuthorityPerEpochStore>,
     transactions: &mut Vec<VerifiedSequencedConsensusTransaction>,
 ) -> IotaResult<(
     Vec<(TransactionDigest, IotaError)>,
     HashMap<ObjectRef, LockDetails>,
+    Vec<TransactionDigest>,
```
roman1e2f5p8s, line 94: Maybe worth adding a named struct for this three-tuple return?

crates/iota-node/src/lib.rs
```diff
@@ -161,9 +163,14 @@ pub mod metrics;
...
     validator_server_handle: SpawnOnce,
     validator_overload_monitor_handle: Option<JoinHandle<()>>,
+    /// Handle for the soft-lock expiry sweep task. The task self-terminates
+    /// via a `Weak` reference; this handle exists purely for ownership clarity.
+    soft_lock_sweep_handle: JoinHandle<()>,
     consensus_manager: ConsensusManager,
     consensus_store_pruner: ConsensusStorePruner,
     consensus_adapter: Arc<ConsensusAdapter>,
+    soft_locks: Arc<PreConsensusSoftLocks>,
+    validator_service_metrics: Arc<ValidatorServiceMetrics>,
```
roman1e2f5p8s: Could you please elaborate on why this field is needed? It seems to be dead code - the field is never accessed.

crates/iota-node/src/lib.rs
```diff
@@ -161,9 +163,14 @@ pub mod metrics;
 pub struct ValidatorComponents {
     validator_server_handle: SpawnOnce,
     validator_overload_monitor_handle: Option<JoinHandle<()>>,
+    /// Handle for the soft-lock expiry sweep task. The task self-terminates
+    /// via a `Weak` reference; this handle exists purely for ownership clarity.
+    soft_lock_sweep_handle: JoinHandle<()>,
```
roman1e2f5p8s, line 164: Why store this as a field? Seems to be dead code that can be removed?

crates/iota-node/src/lib.rs
```diff
@@ -1349,14 +1356,34 @@ impl IotaNode {
...
+        let validator_service_metrics = Arc::new(ValidatorServiceMetrics::new(&validator_registry));
+
+        // Spawn the soft-lock sweep once for the lifetime of this validator
+        // instance. The task holds only a `Weak<PreConsensusSoftLocks>` so it
+        // stops itself automatically: each iteration it tries to upgrade the
+        // weak reference, and when all strong `Arc` owners have been dropped
+        // (i.e. `ValidatorComponents` is destructured and the old epoch store
+        // is released after an epoch transition that removes us from the
+        // committee) the upgrade returns `None` and the loop exits. No explicit
+        // `abort()` is needed. The same `Arc<PreConsensusSoftLocks>` is reused
```
roman1e2f5p8s, line 1191: so if no explicit abort is needed, why keep the `soft_lock_sweep_handle` field?

crates/iota-core/src/authority_server/soft_lock.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PreConsensusSoftLocks {
+    /// Creates a new instance with the default TTL.
+    pub fn new() -> Self {
+        Self::with_ttl(DEFAULT_SOFT_LOCK_TTL)
+    }
```
roman1e2f5p8s, line 100: `DEFAULT_SOFT_LOCK_TTL` seems to be reasonably chosen for the current protocol config parameters, but it is hardcoded. What if `consensus_gs_depth` is changed in the future? We might forget to update the TTL here. I would tie it to the protocol config and have only one public API, something like `pub fn new_from_protocol(protocol_config: &ProtocolConfig) -> Self`. The safety margin factor and commit round rate could also be passed to the constructor, defaulting to 4 and 20, respectively, if none are provided.

crates/iota-core/src/authority_server/soft_lock.rs
```diff
@@ -0,0 +1,549 @@
...
+
+        Ok(())
+    }
+
+    /// Releases all soft locks held by `tx_digest`.
+    ///
+    /// Called by active GC hooks (consensus processed / tx dropped) and on
+    /// consensus submission failure.
+    pub fn release(&self, tx_digest: &TransactionDigest) {
+        let mut inner = self.inner.lock();
```
roman1e2f5p8s, line 189: Worth adding `release_for_batch(&self, tx_digests: &[TransactionDigest])` with single mutex lock acquisition per commit.

crates/iota-core/src/authority_server/soft_lock.rs
```diff
@@ -0,0 +1,549 @@
...
+            second_ts > first_ts,
+            "same-digest re-acquire must refresh the timestamp (first={first_ts:?}, second={second_ts:?})",
+        );
+    }
+
+    /// Concurrent `try_acquire` calls for different digests on the same
+    /// object must produce exactly one winner and `N-1` `ObjectLockConflict`
+    /// errors — no lost locks, no duplicate wins.
+    #[test]
+    fn test_concurrent_acquire_exactly_one_winner() {
```
roman1e2f5p8s, line 497: A similar test could be added to run a release concurrently with an acquire, as a regression guard for future refactors.

crates/iota-core/src/authority_server/soft_lock.rs
```diff
@@ -0,0 +1,549 @@
...
+        let tx = digest(1);
+
+        table.try_acquire(tx, &[obj]).unwrap();
+        // Same digest again — should succeed.
+        table.try_acquire(tx, &[obj]).unwrap();
+        assert_eq!(table.lock_count(), 1);
+    }
+
+    #[test]
+    fn test_ttl_expiry_allows_reacquisition() {
```
roman1e2f5p8s, line 421: A similar test could be added with a non-zero duration to properly test the `now.duration_since(acquired_at) >= ttl` comparison.

roman1e2f5p8s (changes requested):

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: no need to gate it. the callsite should be gated.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: true, but I'd rather avoid adding more indentation here for better readability.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: It doesn't need to be perfectly accurate because metrics are read every minute or so anyway. I tried to also update here but it results in ugly solution so I stayed with the current one which is good enough imo.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: good idea

crates/iota-core/src/authority_server/metrics.rs
piotrm50: not really needed. It's useful only to check mem usage and number of prevented conflicts.

crates/iota-core/src/authority_server/soft_lock.rs
piotrm50: will do

crates/iota-core/src/authority_server/validator_v2.rs
piotrm50: i will do that

crates/iota-core/src/post_consensus_validation.rs
piotrm50: those are unpacked and used separately, so I think its fine as is.

crates/iota-node/src/lib.rs
piotrm50: >  this handle exists purely for ownership clarity.

piotrm50: its the same as with `validator_overload_monitor_handle`

crates/iota-node/src/lib.rs
piotrm50: will check. could be a leftover from one of the approaches to update metrics on removal

crates/iota-node/src/lib.rs
piotrm50: as written above - for ownership clarity.

crates/iota-core/src/authority_server/soft_lock.rs
piotrm50: The TTL only guards against the rare edge case where a transaction is garbage-collected by consensus (i.e. forgotten without a commit/reject decision). For the happy paths - accepted, rejected, or submission failure - we release the lock explicitly, so the TTL never matters there.
Concretely:
 - gc_depth is a knob we'd only ever raise, never lower, so a 12 s TTL stays safe by construction. If we did raise it materially, we'd notice the same way we'd notice any other consensus-tuning change.
 - Even if the TTL were "too short" for some future config, the only effect is that a soft lock for a GC'd tx gets released a bit early — which is the desired behavior anyway. It doesn't break correctness; post-consensus is the authoritative resolver.
 - 12 s is also a reasonable cap for conflict-spam prevention regardless of gc_depth: we don't want a forgotten tx to block its inputs much longer than that.
Threading ProtocolConfig through PreConsensusSoftLocks::new (plus optional safety-margin / round-rate parameters) adds API surface and a config dependency for a value that doesn't need to be precise. I'd rather keep the constant and revisit if/when we actually change consensus_gc_depth.

roman1e2f5p8s (approved):

filipdulic (approved):

bingyanglin (approved):

PR #11152, opened 2026-04-07
fix(iota-core): skip duplicate EndOfPublish in white-flag mode

### LINKED ISSUE #10925: pcool: excessive duplicate EndOfPublish messages during epoch transition in white-flag mode
## Problem

In certificate mode, the `send_end_of_publish` path in `consensus_adapter.rs` (`submit_and_wait_inner` notification callback, ~line 869) was designed as a rare race condition handler — pending certificates drain while another is being processed.

In white-flag (certificate-less) mode, `pending_consensus_certificates` is always empty, so `is_pending_consensus_certificates_empty()` always returns `true`. This means **every `UserTransactionV1` processed while in `RejectUserCerts` state triggers an `EndOfPublish` submission**.

The existing comment (line 876) says duplicate EndOfPublish messages "don't affect correctness" — but that was written for the rare-race case, not for every-transaction-triggers-it. This creates unnecessary consensus overhead during epoch transition.

## Suggested Fix

Guard with an `AtomicBool` to send `EndOfPublish` only once in white-flag mode, or at minimum add a `debug!` log when duplicates are suppressed.

## Location

`crates/iota-core/src/consensus_adapter.rs` ~line 869

piotrm50:
# Description of change

In white-flag (certificate-less) mode, every `UserTransactionV1` completing in `RejectUserCerts` state triggered an `EndOfPublish` submission from `submit_and_wait_inner`, because `pending_consensus_certificates_count()` is always 0. This produced N duplicate `EndOfPublish` messages where N is the number of in-flight user transactions at epoch transition.

The `submit_and_wait_inner` EndOfPublish logic exists for the certificate drain scenario — detecting when the last pending certificate completes. In white-flag mode there's nothing to drain, so `close_epoch` is the correct and sufficient sender.

This change skips the EndOfPublish check in `submit_and_wait_inner` for white-flag mode entirely. The certificate-mode drain logic is preserved but annotated with a TODO for removal when the certificate flow is cleaned up.

Fixes #10925

### DIFF
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -874,35 +874,40 @@ impl ConsensusAdapter {
                 )
             };
         let send_end_of_publish = if is_user_tx {
-            // If we are in `RejectUserCerts` state, we need to send `EndOfPublish` to
-            // signal other validators that we are not submitting more user
-            // transactions in the epoch. Note that there could be a race condition
-            // here where we enter this check in `RejectAllCerts` state. In that case
-            // we don't need to send `EndOfPublish` because the condition to enter
-            // `RejectAllCerts` is when 2f+1 other validators already sequenced their
-            // `EndOfPublish` message. Also note that we could send multiple
-            // `EndOfPublish` messages because multiple tasks can enter here
-            // concurrently. This doesn't affect correctness.
-            if epoch_store
-                .get_reconfig_state_read_lock_guard()
-                .is_reject_user_certs()
-            {
-                if epoch_store.protocol_config().enable_white_flag_flow() {
-                    // In the certificate-less mode, there are no pending consensus
-                    // certificates, so `EndOfPublish` is always sent immediately.
-                    debug!(epoch=?epoch_store.epoch(), "Sending EndOfPublish in certificate-less mode");
-
-                    true
-                } else {
-                    // In certificate mode, `EndOfPublish` is sent only once the list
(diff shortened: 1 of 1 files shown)


bingyanglin (approved):

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -874,35 +874,40 @@ impl ConsensusAdapter {
...
-
-                    true
-                } else {
-                    // In certificate mode, `EndOfPublish` is sent only once the list
-                    // of pending consensus certificates is drained.
+            if epoch_store.protocol_config().enable_white_flag_flow() {
+                // In certificate-less mode, `EndOfPublish` is sent solely from
+                // `close_epoch`. There is no pending certificate drain to
+                // monitor here, and sending from this per-transaction callback
+                // would produce N duplicate EndOfPublish messages (one per
```
roman1e2f5p8s, line 881: It would be great to explicitly mention here that `submit_and_wait_inner` is spawned as a separate task for each `submit`/`submit_batch` call; otherwise, it is not obvious from the local context.

roman1e2f5p8s (approved):
Great catch!

For clarity, I would also docstring `submit_and_wait_inner` that it is spawned as a separate task for each `submit`/`submit_batch` call and that the `transactions` vector typically contains a single transaction or multiple transactions in the case of a soft bundle - both are not self-evident from the function signature.

filipdulic (approved):

PR #11153, opened 2026-04-07
fix(iota-core): add retry with backoff for EndOfPublish submission

### LINKED ISSUE #10926: pcool: EndOfPublish submission failure is swallowed with no retry in white-flag mode
## Problem

In `consensus_adapter.rs` (~line 906), when `EndOfPublish` fails to submit, the error is logged at `warn!` level and swallowed:

```rust
if let Err(err) = self.submit(
    ConsensusTransaction::new_end_of_publish(self.authority),
    None,
    epoch_store,
) {
    warn!("Error when sending end of publish message: {:?}", err);
}
```

This is a pre-existing pattern, but it is amplified by white-flag mode. In certificate mode, the drain-and-retry loop provides redundancy — if one EndOfPublish attempt fails, the loop retries when the next certificate drains. In white-flag mode, `EndOfPublish` is sent immediately with no retry mechanism. A transient failure here silently stalls the epoch because other validators are waiting for this validator's `EndOfPublish`.

## Suggested Fix

- Add retry logic for `EndOfPublish` in white-flag mode (at minimum a bounded retry with backoff)
- Escalate the log to `error!` level and include epoch number and authority ID for quick identification

## Location

`crates/iota-core/src/consensus_adapter.rs` ~line 906

piotrm50:
# Description of change

This pull request improves the reliability of sending `EndOfPublish` messages to consensus by introducing an indefinite retry mechanism with exponential backoff. Previously, if sending failed due to transient errors (like database write issues), the process would not retry, potentially stalling the epoch. Now, the system will keep retrying until the message is successfully sent, ensuring forward progress.

**Reliability improvements for EndOfPublish message delivery:**

* Added a new async method `submit_end_of_publish_with_retry` to `ConsensusAdapter` that retries sending the `EndOfPublish` message with exponential backoff, capped at 10 seconds, until it succeeds. This prevents epoch stalls due to transient errors.
* Updated existing code paths to use the new retry method:
  - In the main consensus adapter logic, replaced the direct call to `submit` with `submit_end_of_publish_with_retry`.
  - In the `ReconfigurationInitiator` implementation, replaced the direct call with a monitored task that invokes the retry method asynchronously.

Both `close_epoch` and `submit_and_wait_inner` (certificate-mode path) now use the retry helper.

Fixes #10926

### DIFF
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -913,7 +913,16 @@ impl ConsensusAdapter {
             false
         };
         if send_end_of_publish {
-            self.submit_end_of_publish_with_retry(epoch_store).await;
+            if epoch_store
+                .within_alive_epoch(self.submit_end_of_publish_with_retry(epoch_store))
+                .await
+                .is_err()
+            {
+                warn!(
+                    epoch = ?epoch_store.epoch(),
+                    "EndOfPublish submission cancelled: epoch has ended",
+                );
+            }
         }
         self.metrics
             .sequencing_certificate_success
@@ -1058,11 +1067,13 @@ impl ConsensusAdapter {
         ProcessedMethod::Consensus
     }
 
-    /// Submits an `EndOfPublish` message to consensus with indefinite retry
-    /// and exponential backoff (capped at `MAX_BACKOFF`). On transient
-    /// failures (e.g. DB write errors in
-    /// `insert_pending_consensus_transactions`), retries until success since
-    /// a missing `EndOfPublish` would stall the epoch.
+    /// Submits an `EndOfPublish` message to consensus with exponential
+    /// backoff (capped at `MAX_BACKOFF`). On transient failures (e.g. DB
+    /// write errors in `insert_pending_consensus_transactions`), retries
+    /// until success since a missing `EndOfPublish` would stall the epoch.
+    ///
+    /// Callers should wrap this with `epoch_store.within_alive_epoch()` to
+    /// cancel retries when the epoch terminates.
     async fn submit_end_of_publish_with_retry(
         self: &Arc<Self>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
@@ -1207,7 +1218,20 @@ impl ReconfigurationInitiator for Arc<ConsensusAdapter> {
         if send_end_of_publish {
             let adapter = self.clone();
             let epoch_store = epoch_store.clone();
-            spawn_monitored_task!(adapter.submit_end_of_publish_with_retry(&epoch_store));
+            spawn_monitored_task!(async move {
+                if epoch_store
+                    .within_alive_epoch(
+                        adapter.submit_end_of_publish_with_retry(&epoch_store),
+                    )
+                    .await
+                    .is_err()
+                {
+                    warn!(
+                        epoch = ?epoch_store.epoch(),
+                        "EndOfPublish submission cancelled: epoch has ended",
+                    );
+                }
+            });
         }
     }
 }
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -45,7 +45,7 @@ use tokio::{
         Duration, {self},
     },
 };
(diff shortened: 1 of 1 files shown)


crates/iota-core/src/consensus_adapter.rs
```diff
@@ -1065,6 +1057,59 @@ impl ConsensusAdapter {
...
+                epoch_store,
+            ) {
+                Ok(_) => return,
+                Err(err) => {
+                    if attempt == MAX_RETRIES {
+                        error!(
+                            epoch = ?epoch_store.epoch(),
+                            authority = ?self.authority,
+                            "Failed to submit EndOfPublish after {} attempts: {:?}. \
+                             Will rely on crash recovery via submit_recovered.",
```
piotrm50: what does this actually mean? submit_recovered is done upon node restart, so how can we rely on it here? Shouldn't we try to issue this indefinitely?

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -1170,14 +1205,9 @@ impl ReconfigurationInitiator for Arc<ConsensusAdapter> {
...
-            if let Err(err) = self.submit(
-                ConsensusTransaction::new_end_of_publish(self.authority),
-                None,
-                epoch_store,
-            ) {
-                warn!("Error when sending end of publish message: {:?}", err);
-            }
+            let adapter = self.clone();
+            let epoch_store = epoch_store.clone();
+            spawn_monitored_task!(adapter.submit_end_of_publish_with_retry(&epoch_store));
```
bingyanglin: I am a bit confused about the different handling. If this task panics, the epoch close proceeds without `EndOfPublish` being sent. But in Line 916 awaits the same function.

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -1065,6 +1057,49 @@ impl ConsensusAdapter {
...
+
+        let mut attempt: u32 = 0;
+        loop {
+            match self.submit(
+                ConsensusTransaction::new_end_of_publish(self.authority),
+                None,
+                epoch_store,
+            ) {
+                Ok(_) => return,
+                Err(err) => {
```
filipdulic, line 1118: This retries on every error, but not every error should be retryable as some can cause an infinite loop, and it makes sense to handle it/them. Errors like `IotaError::EpochEnded(epoch)` returned from the `tables` method.

crates/iota-core/src/consensus_adapter.rs
piotrm50: Good question! The different handling is intentional.

`close_epoch` implements `ReconfigurationInitiator`, which is a `sync` trait, so we can't `.await` here - hence the `spawn_monitored_task!`. This is safe because by the time we reach this point, the `ReconfigState` has already been set to reject user certs and transactions locally, so no new transactions will be issued regardless. The `EndOfPublish` message is needed to inform other validators about our state, but it doesn't need to complete before close_epoch returns.

In `submit_and_wait_inner` (the certificate-mode drain path at line 916), we're already in an `async` context so we can `.await` directly.

I considered using `std::thread::sleep` instead of `tokio::time::sleep` to keep the method synchronous and block `close_epoch` until the message is sent, but that would be problematic for the `submit_and_wait_inner` call site - we'd be blocking a tokio worker thread during retries. Since `submit_end_of_publish_with_retry` is shared by both paths, it needs to be async. Once the certificate-mode path is fully removed and EndOfPublish is only issued through close_epoch, we could revisit making it synchronous, though blocking for potentially long retries is still questionable.

Additionally, in the latest revision the retry loop is now wrapped with `epoch_store.within_alive_epoch()` at both call sites, so retries are automatically cancelled when the epoch terminates — addressing Filip's concern about infinite retry loops on `EpochEnded` errors.

crates/iota-core/src/consensus_adapter.rs
piotrm50: Wrapped the callsites with `epoch_store.within_alive_epoch` so the asynchronous tasks will quit when the epoch ends regardless of the underlying errors.

filipdulic (approved):

bingyanglin (approved):

PR #11156, opened 2026-04-07
refactor(iota-indexer-streaming): streams recovery

### LINKED ISSUE #11013: [graphql,indexer-streaming]: support recovery of streams
## Description

Currently users subscribed to GraphQL streams on transactions and events do not have the option to restart the stream at a specific transaction after a broken connection.

We want to support recovery of streams based on the transaction identifier (digest).

sergiupopescu199:
# Description of change

This PR adds stream recovery support to `iota-indexer-streaming` allowing clients to resume transaction and event subscriptions from a specific transaction digest after a disconnection.

The historical stream uses `futures::stream::unfold` which is demand-driven, it only advances when the consumer polls for the next item. Transactions are fetched from the database in batches and stored in an internal buffer. Each poll drains one item from the buffer, and a new database query is only issued when the buffer is empty. Memory usage is bounded by the configured batch size with no unbounded growth. Once the historical backfill catches up to the latest committed transaction, the stream ends and `chain` seamlessly switches to the live broadcast stream.

When a `start_from` represented by the `RecoveryPoint` enum is provided (can be inclusive or exclusive), the stream first backfills historical data from the database, then seamlessly transitions to live data from the broadcast channel. Without `start_from`, behavior is unchanged.

fixes #11013

>[!NOTE]
> This patch does not affect the normal operations of the iota-indexer this not further tests were conducted.

### DIFF
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -264,10 +264,10 @@ impl InMemory {
     /// filtering.
     ///
     /// When `start_from` is `None`, subscribes to live events only. When
-    /// `start_from` is `Some(digest)`, the stream first backfills
-    /// historical events from the transaction identified by the digest,
-    /// then seamlessly transitions to live events. This enables stream
-    /// recovery after a disconnection.
+    /// `start_from` is `Some(digest)`, the stream first backfills historical
+    /// events from the transaction identified by the digest inclusive) up to
+    /// the tip of the network, then seamlessly transitions to live events. This
+    /// enables stream recovery after a disconnection.
     ///
     /// # Note
     /// Since under the hood a [`tokio::sync::broadcast`] channel is used for
@@ -376,10 +376,10 @@ impl InMemory {
     /// filtering.
     ///
     /// When `start_from` is `None`, subscribes to live transactions only. When
-    /// `start_from` is `Some(digest)`, the stream first backfills
-    /// historical transactions from the one identified by the digest,
-    /// then seamlessly transitions to live transactions. This enables
-    /// stream recovery after a disconnection.
+    /// `start_from` is `Some(digest)`, the stream first backfills historical
+    /// transactions from the one identified by the digest (inclusive) up to the
+    /// tip of the network, then seamlessly transitions to live transactions.
+    /// This enables stream recovery after a disconnection.
     ///
     /// # Note
     /// Since under the hood a [`tokio::sync::broadcast`] channel is used for
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -943,8 +943,10 @@ impl HistoricalFetch {
                             Ok(cursor) => state.cursor.store(cursor, Ordering::Relaxed),
                             Err(e) => {
                                 state.should_close_stream = true;
-                                let e = IndexerStreamingError::Postgres(e.to_string());
-                                error!("failed to resolve tx digest into its sequence number: {e}");
+                                let e = IndexerStreamingError::Postgres(format!(
+                                    "unable to resolve transaction, may not exist or has been pruned: {e}"
+                                ));
+                                error!("{e}");
                                 return Some((Err(e), state));
                             }
                         }
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -360,13 +360,20 @@ impl InMemory {
                 // when switching from historical to live, the broadcast buffer may
                 // contain transactions already delivered by the historical backfill.
                 // Filter them out using the cursor to prevent duplicates.
-                .chain(live_stream.filter(move |result| {
-                    future::ready(match result {
-                        Ok(tx) => tx.tx_sequence_number >= cursor.load(Ordering::Relaxed),
-                        // we must include errors.
-                        Err(_) => true,
+                .chain({
+                    let mut initial_lagged_skipped = false;
+                    live_stream.skip_while(move |result| {
+                        future::ready(match result {
+                            Ok(tx) => tx.tx_sequence_number < cursor.load(Ordering::Relaxed),
+                            Err(IndexerStreamingError::Lagged(_)) if !initial_lagged_skipped => {
+                                initial_lagged_skipped = true;
+                                true // skip this one
+                            }
+                            // surface any other error or second Lagged
+                            Err(_) => false,
+                        })
                     })
-                })),
+                }),
         )
     }
 
@@ -458,13 +465,20 @@ impl InMemory {
                 // when switching from historical to live, the broadcast buffer may
                 // contain transactions already delivered by the historical backfill.
                 // Filter them out using the cursor to prevent duplicates.
-                .chain(live_stream.filter(move |result| {
-                    future::ready(match result {
-                        Ok(tx) => tx.tx_sequence_number >= cursor.load(Ordering::Relaxed),
-                        // we must include errors.
-                        Err(_) => true,
+                .chain({
+                    let mut initial_lagged_skipped = false;
+                    live_stream.skip_while(move |result| {
+                        future::ready(match result {
+                            Ok(tx) => tx.tx_sequence_number < cursor.load(Ordering::Relaxed),
+                            Err(IndexerStreamingError::Lagged(_)) if !initial_lagged_skipped => {
+                                initial_lagged_skipped = true;
+                                true // skip this one
+                            }
+                            // surface any other error or second Lagged
+                            Err(_) => false,
+                        })
                     })
-                })),
+                }),
         )
     }
 
--- a/crates/iota-indexer-streaming/src/memory.rs
+++ b/crates/iota-indexer-streaming/src/memory.rs
@@ -134,6 +134,34 @@ impl Default for Config {
     }
 }
 
+/// Where to start a recovery stream relative to a known transaction digest.
+///
+/// - [`Inclusive`](Self::Inclusive) — yield the identified transaction, then
+///   everything after it.
+/// - [`Exclusive`](Self::Exclusive) — skip the identified transaction; start
+///   from the next one.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Deserialize, Serialize)]
+pub enum RecoveryPoint {
+    /// Include the transaction identified by the digest.
+    Inclusive(TransactionDigest),
+    /// Start after the transaction identified by the digest.
+    Exclusive(TransactionDigest),
+}
+
+impl RecoveryPoint {
+    /// Returns the contained transaction digest.
+    fn digest(&self) -> TransactionDigest {
+        match self {
+            Self::Inclusive(d) | Self::Exclusive(d) => *d,
+        }
+    }
+
+    /// Checks if the starting transaction should be included in the stream.
+    fn is_inclusive(&self) -> bool {
+        matches!(self, Self::Inclusive(_))
+    }
+}
+
 /// Provides real-time streaming of transactions and events from the IOTA
 /// Indexer by listening to PostgreSQL NOTIFY messages triggered when new
 /// checkpoints are committed to the indexer database.
@@ -264,10 +292,19 @@ impl InMemory {
     /// filtering.
     ///
     /// When `start_from` is `None`, subscribes to live events only. When
-    /// `start_from` is `Some(digest)`, the stream first backfills historical
-    /// events from the transaction identified by the digest inclusive) up to
-    /// the tip of the network, then seamlessly transitions to live events. This
-    /// enables stream recovery after a disconnection.
+    /// `start_from` is `Some(recovery_point)`, the stream first backfills
+    /// historical events from the transaction identified by the recovery
+    /// point's digest up to the tip of the network, then seamlessly transitions
+    /// to live events. This enables stream recovery after a disconnection.
+    ///
+    /// The recovery point variant controls whether events of the starting
+    /// transaction are included:
+    /// - [`RecoveryPoint::Inclusive`]: events of the starting transaction are
+    ///   yielded.
+    /// - [`RecoveryPoint::Exclusive`]: events of the starting transaction are
+    ///   skipped; streaming begins from the next transaction onward. Useful for
+    ///   reconnection, where the client already received that transaction's
+    ///   events in the previous session.
     ///
     /// # Note
     /// Since under the hood a [`tokio::sync::broadcast`] channel is used for
@@ -276,7 +313,7 @@ impl InMemory {
     /// # Example
     ///
     /// ```rust,ignore
-    /// use iota_indexer_streaming::error::IndexerStreamingError;
+    /// use iota_indexer_streaming::{error::IndexerStreamingError, memory::RecoveryPoint};
     /// use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
     ///
     /// // live events only.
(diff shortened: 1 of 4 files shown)


crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -661,3 +834,213 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
...
+                                .await
+                                .map_err(backoff::Error::transient)
+                        })
+                        .await
+                        {
+                            Ok(cursor) => state.cursor.store(cursor, Ordering::Relaxed),
+                            Err(e) => {
+                                state.should_close_stream = true;
+                                let e = IndexerStreamingError::Postgres(e.to_string());
+                                error!("failed to resolve tx digest into its sequence number: {e}");
```
kodemartin: If the requested `startFrom` transaction is pruned we will get to this branch.

So it will be helpful to inform users that data might be pruned, optionally providing the minimum transaction cursor available.

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -242,60 +263,209 @@ impl InMemory {
...
+        );
+
+        let cursor = historical.cursor();
+
+        Box::pin(
+            historical
+                .into_stream()
+                // when switching from historical to live, the broadcast buffer may
+                // contain transactions already delivered by the historical backfill.
+                // Filter them out using the cursor to prevent duplicates.
```
kodemartin, line 536: Isn't it possible that the broadcast buffer has been filled at capacity before the historical fetch finishes?

For example, what happens if `startFrom` is a transaction digest is from one epoch ago?

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -661,3 +834,213 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
...
+        };
+
+        backoff::future::retry(backoff, f).await
+    }
+
+    /// Converts into a stream that yields historical transactions from the
+    /// database in batches. The stream ends once all transactions up to
+    /// the latest committed transaction to database have been yielded, at which
+    /// point the caller can chain a live stream.
+    fn into_stream(self) -> impl Stream<Item = IndexerStreamingResult<StoredTransaction>> {
```
kodemartin, line 1015: As far as I understand, this essentially polls for the latest committed transaction once, and stops once we reach that.

But by the time it stops the latest committed transaction in the database has likely advanced, and thus we rely on the live stream to include the batch of transaction in between.

It seems that if we keep updating `state.latest` after each batch ends we don't need the live stream at all, reducing into a polling mechanism. This seems more robust, as it is not make any assumptions about the consumption rate of the live broadcast channel. Have you considered this approach?

sergiupopescu199: The historical stream works as follows: on start, it fetches `state.latest` (the latest committed tx sequence number) from the database and downloads all transactions in batches until `cursor > state.latest`. At that point, it refreshes `state.latest` from the database. If `state.latest` has advanced (new transactions were committed while fetching), the stream continues downloading the new range. If after the refresh `cursor > state.latest` still holds, the database has no new data — the stream has reached the tip and returns `None`, handing off to live.

I did consider removing the live stream entirely. The concern is database load: with pure polling, each of N subscribers independently queries the DB for new transactions. With the broadcast approach, the DB is queried once and the result is fanned out to N subscribers, significantly less DB load at the cost of the broadcast buffer limitation. We need benchmarks to determine at what subscriber count the pure polling approach becomes a bottleneck and the database load with these two approaches.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: Yes, if the recovery point is far behind, the broadcast buffer will likely overflow during backfill since the live receiver is created at subscription time but not consumed until the historical stream finishes.
 
I considered deferring the live subscription to after backfill completes, but that introduces a silent gap transactions committed between the last `state.latest` refresh and the subscribe moment would be lost without the client knowing. With the current approach, a buffer overflow surfaces as a `Lagged` error, which is preferable to silent data loss.

In practice this converges naturally: on `Lagged`, the client reconnects with the last received digest, each retry starts closer to the tip, and the backfill gets shorter until it no longer overflows the buffer.
 
The pure polling approach using only DB fetches would eliminate this limitation entirely. The only concern is database load under a high number of concurrent subscribers, which would need benchmarking to validate.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: Good point, I'll add the necessary changes to the error message. I was thinking that in future iterations it could make sense to fetch data from the `kv-store` if configured to the indexer if data was pruned form db.

crates/iota-indexer-streaming/src/memory.rs
kodemartin: The thing I didn't get is that by the time the batch ends, cursor is larger than `state.latest` (it is incremented by 1 past the range end) so you re-poll for the latest transaction in the database.

So this 

> It seems that if we keep updating state.latest after each batch ends we don't need the live stream at all, reducing into a polling mechanism. This seems more robust, as it is not make any assumptions about the consumption rate of the live broadcast channel. Have you considered this approach?

is already done!

crates/iota-indexer-streaming/src/memory.rs
kodemartin: Following our discussion, and since we verified that the new implementation guarantees that clients will have received all messages at the tip of the network when the switch to the live stream, they shouldn't receive a `Lagged` message. 

Consumption should remain smooth. This needs to be handled.

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -242,60 +263,209 @@ impl InMemory {
     /// By default all events are received, the client shall handle the
     /// filtering.
     ///
+    /// When `start_from` is `None`, subscribes to live events only. When
+    /// `start_from` is `Some(digest)`, the stream first backfills
+    /// historical events from the transaction identified by the digest,
```
tomxey: It could be stated whether it backfills from `digest` inclusive or exclusive. (will the first message include `digest` or not?)

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/660442ccc96fe92ab3d272e1b4664d59bb4f13e6

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/23e5eb782bf2666822e12525ab342c6b18eadf24

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -242,60 +260,221 @@ impl InMemory {
...
+                Err(e) => stream::once(future::err(e)).right_stream(),
+            })
+            .flatten();
+
+        // subscribe to the broadcast channel lazily, only after the historical
+        // stream finishes. stream::once defers the subscription until chain
+        // polls the live stream, ensuring we subscribe at the tip of the network
+        // and avoid accumulating stale items in the broadcast buffer.
+        let live_stream = stream::once({
+            let event_tx = self.event_tx.clone();
```
tomxey: If this is called after historical fetch stopped reading, how do we make sure that we are not missing some events before we start the new live stream?

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: Yes it will be inclusive. Will refine the rust doc to include this information.

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/0dee593e8cf17b7bbd59635ca27c1c7f2453e004

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/23e5eb782bf2666822e12525ab342c6b18eadf24

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -661,3 +848,215 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
...
+                    }
+
+                    // drain the buffer before fetching new transactions from the database.
+                    // This is the main point of sending transactions to the caller.
+                    if let Some(tx) = state.buffer.pop_front() {
+                        return Some((Ok(tx), state));
+                    }
+
+                    // resolve the cursor from the digest if it hasn't been set yet.
+                    if state.cursor.load(Ordering::Relaxed) == 0 {
```
tomxey: What if the cursor is indeed 0 (genesis) ?
Did you consider storing lack of cursor more explicitly with a `None`?

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -242,60 +263,223 @@ impl InMemory {
     /// By default all events are received, the client shall handle the
     /// filtering.
     ///
+    /// When `start_from` is `None`, subscribes to live events only. When
+    /// `start_from` is `Some(digest)`, the stream first backfills historical
+    /// events from the transaction identified by the digest inclusive) up to
+    /// the tip of the network, then seamlessly transitions to live events. This
+    /// enables stream recovery after a disconnection.
```
tomxey: so the flow is that user is receiving transactions from the stream, and when the stream disconnects user establishes a new stream passing the last seen tx digest as the `start_from`?

but then user will get the `start_from` digest twice, once from the original stream and once from the new stream, is that intended?

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -242,60 +263,223 @@ impl InMemory {
...
+
+        Box::pin(
+            historical
+                .into_stream()
+                // when switching from historical to live, the broadcast buffer may
+                // contain transactions already delivered by the historical backfill.
+                // Filter them out using the cursor to prevent duplicates.
+                .chain({
+                    let mut initial_lagged_skipped = false;
+                    live_stream.skip_while(move |result| {
```
tomxey, line 539: is some automated test planned to test this behaviour?

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -661,3 +848,215 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
...
 }
+
+/// Fetches historical transactions from the indexer database starting from a
+/// given transaction digest up to the latest available sequence number.
+///
+/// This stream is used for stream recovery, the caller chains this stream with
+/// a live broadcast stream to provide gap-free delivery from a recovery point.
+struct HistoricalFetch {
+    /// Transaction digest to start fetching from.
+    digest: TransactionDigest,
```
tomxey: nit: could we use sth like `start_from_digest`, currently it may look a bit like current digest

crates/iota-indexer-streaming/src/memory.rs
```diff
@@ -661,3 +848,215 @@ impl<T: Clone + Send + 'static> Stream for SubscriberStream<T> {
...
+    /// Next transaction sequence number to fetch.
+    cursor: Arc<AtomicI64>,
+    /// Latest committed transaction sequence number in the database.
+    latest: i64,
```
tomxey: nit: could we use sth like `latest_tx_in_db` ? currently it may look a bit like latest processd tx

crates/iota-indexer-streaming/src/memory.rs
kodemartin: I would suggest having a dedicated error variant (e.g. `DataNotFound`), so that applications can chose whether this is a server- or user-side error.

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/5525d8cf35a68ff36805ce1fe7eb2915b5b4fbb3

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/fcffa5ce42160733a0037bfaea91b6347ee69e05

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/fcffa5ce42160733a0037bfaea91b6347ee69e05

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/fcffa5ce42160733a0037bfaea91b6347ee69e05

The library now supports explicit tx inclusion on start to support different use cases. Graphql will use the Exclusive enum variant.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11156/commits/fcffa5ce42160733a0037bfaea91b6347ee69e05

Opted to use a boolean instead of an Option moslty because updating and inspecting the cursor would've been more tedious.

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: The crate isn't currently set up for end-to-end integration tests of this path, exercising it properly needs a running postgres, a committed indexer state, and an active broker publishing to the broadcast channel. That's a non-trivial harness to add in this PR.

I did the tests manually though, which not ideal, we could thing in a future issue a way to add automated tests.

crates/iota-indexer-streaming/src/memory.rs
tomxey: Is it possible that a user streaming events receives only some events from a transaction, then the stream crashes? On reconnect they'd pass the last tx digest they saw, but Exclusive skips the entire transaction — including the events they never actually got.

  - The event stream recovery operates at transaction granularity (the skip_while drops the whole tx before extracting events)
  - If the crash happened mid-transaction (e.g. got 3 out of 7 events), the remaining 4 are silently lost

kodemartin (approved):
🍉 lgtm

crates/iota-indexer-streaming/src/memory.rs
sergiupopescu199: Good observation. We try to push atomicity to the client side: they buffer events by `transactionBlock.digest` and only commit once the digest changes that signals the previous tx is fully delivered. The digest passed back on reconnect is always the last *fully-processed* tx, so `Exclusive` is safe by construction. I've updated the GraphQL subscription docs to spell out this consumption pattern.

A possible comment for the GraphQL events stream query

```
    /// # Stream recovery
    ///
    /// Events are streamed in transaction order all events from a single
    /// transaction arrive contiguously before any events from the next one.
    ///
    /// When `start_after` is provided with a transaction digest, the stream
    /// resumes from the transaction immediately after it.
    ///
    /// Because a single transaction can emit multiple events, a
    /// disconnection may leave the client with only a partial set of events
    /// from the transaction it was last receiving. To recover safely,
    /// clients are expected to:
    ///
    /// 1. Buffer events grouped by their parent `transactionBlock.digest`.
    /// 2. Commit/process the buffered group only once the digest changes — this
    ///    signals that the previous transaction's events have all been
    ///    delivered.
    /// 3. On reconnect, pass the digest of the last **fully-processed**
    ///    transaction as `start_after`. The stream resumes from the next
    ///    transaction, so no partial transaction is ever observed.
```

tomxey: okay, let's go with this kind of doc

tomxey (approved):

PR #11174, opened 2026-04-08
feat(dapp-kit): add syncTabs property

### LINKED ISSUE #51: `@iota/dapp-kit` (specifically hooks like `useCurrentAccount`) does not natively synchronize state across different browser tabs when the connected wallet changes or disconnects.
## Description

`@iota/dapp-kit` (specifically hooks like `useCurrentAccount`) does not natively synchronize state across different browser tabs when the connected wallet changes or disconnects.

## Motivation

In decentralized applications, it is common for users to have multiple tabs of the same dApp open simultaneously. If a user connects a wallet in Tab A, and then switches to a different wallet (or disconnects entirely) in Tab B, the `@iota/dapp-kit` state in Tab A remains completely stale.

This leads to a dangerous desynchronization where the UI in Tab A shows the old account's identity and permissions, but any transaction signed from Tab A would either fail or be routed through the new account silently. This is a severe UX and potential security issue. Mature Web3 libraries (such as `wagmi`) handle cross-tab synchronization natively out-of-the-box by listening to `storage` events and updating their React contexts globally.

## Requirements

Write a list of what you want this feature to do.

1. Automatically detect changes to the active wallet/account across different browser tabs (e.g., by natively listening to the `storage` event on the specific localStorage keys used by the kit).
2. Reactively trigger updates for hooks like `useCurrentAccount()`, `useCurrentWallet()`, and the internal `WalletProvider` state in all background tabs without requiring a manual page reload.
3. Ensure that background tabs correctly fire connection/disconnection lifecycle events when the active account changes in a sibling tab.

## Open questions (optional)

- Should this cross-tab synchronization be enabled by default, or should it be an opt-in flag on the `WalletProvider` component (e.g., `<WalletProvider syncTabs={true}>`)?
- Does this architectural change need to be implemented upstream in `@mysten/dapp-kit`, or can it be handled completely within the `@iota/dapp-kit` implementation?

No.

panteleymonchuk:
Closes iotaledger/ts-packages#51


brancoder (changes requested):
tests 🙏

panteleymonchuk:
Closed in a favor of: https://github.com/iotaledger/ts-packages/pull/76

PR #11184, opened 2026-04-08
feat: implement built-in IOTA Move authenticators

### LINKED ISSUE #11015: [AA] Create a prototype of cryptographic authentication for default IOTA Accounts
Create a default account and 5 authenticators at the framework level:
- ed25519
- secp256k1
- verify_secp256r1
- verify_multisig
- verify_passkey

[This example](https://github.com/iotaledger/iota/commit/ffb2b3d4d6dc7b43f3969bb5ba5bad959d66c19a#diff-20f8eda8d3fd0826c7194f587502f206cb491ab0455cef9b3f64b3d08f6766d9) can be used as reference. 

To be defined the implementation of native functions for authenticators.

valeriyr:
[run-ci]

# Description of change

Implements the built-in IOTA authenticators for the account abstraction system, enabling accounts to authenticate using the standard IOTA signature schemes (Ed25519, Secp256k1, Secp256r1, MultiSig, Passkey).

fixes #11015


crates/iota-framework/packages/iota-framework/sources/account_abstraction/authenticator_function.move
```diff
@@ -69,6 +69,15 @@ public fun function_name<Account: key>(self: &AuthenticatorFunctionRefV1<Account
     &self.function_name
 }
 
+/// Create an `AuthenticatorFunctionRefV1` instance for built-in authenticator functions.
+public(package) fun create_auth_function_ref_v1_inner<Account: key>(
```
miker83z, line 74: `create_auth_function_ref_v1_for_testing` and `create_auth_function_ref_v1` could call this internally.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -0,0 +1,104 @@
...
+        assert!(len == ED25519_PUBLIC_KEY_LENGTH, EInvalidPublicKeyLength);
+    } else if (
+        scheme == signature_scheme::secp256k1() ||
+        scheme == signature_scheme::secp256r1() ||
+        scheme == signature_scheme::passkey()
+    ) {
+        assert!(len == SECP256_PUBLIC_KEY_LENGTH, EInvalidPublicKeyLength);
+    } else if (scheme == signature_scheme::multisig()) {
+        // MultiSig key payload is BCS-encoded and variable in length.
+        assert!(len > 0, EInvalidPublicKeyLength);
```
miker83z: what about [deserializing it](https://github.com/iotaledger/iota/commit/ffb2b3d4d6dc7b43f3969bb5ba5bad959d66c19a#diff-20f8eda8d3fd0826c7194f587502f206cb491ab0455cef9b3f64b3d08f6766d9R313) and then validate like in MultiSigPublicKey::validate?

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
```diff
@@ -0,0 +1,282 @@
...
+    "secp256r1_authenticator_function_ref_v1";
+pub const MULTISIG_AUTHENTICATOR_FUNCTION_V1_NAME: &str = "multisig_authenticator_function_ref_v1";
+pub const PASSKEY_AUTHENTICATOR_FUNCTION_V1_NAME: &str = "passkey_authenticator_function_ref_v1";
+
+/// The pre-loaded data for a built-in authenticator.
+///
+/// Constructed once during transaction input loading so the executor does not
+/// need to re-derive the scheme from the authenticator function reference.
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct PreloadedBuiltinAuthenticatorData {
```
miker83z, line 44: PreloadedBuiltinAuthenticatorData or PreloadedBuiltinAuthenticatorPublicKey? Do you see this type to be expanded in the future or will it just carry PublicKey data?

crates/iota-types/src/account_abstraction/signature_scheme.rs
miker83z, line 1: I think this can be merged on the rust side with public key. Maybe, even all in builtin_authenticator_functions.rs so that the context of MoveSignatureScheme and MovePublicKey is more clear while moving between files.

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
```diff
@@ -0,0 +1,282 @@
...
+            Ok((public_key_field_id, Some((field.value, metadata))))
+        }
+        None => Ok((public_key_field_id, None)),
+    }
+}
+
+/// If the given authenticator function reference corresponds to a built-in
+/// authenticator, returns the corresponding signature scheme. Otherwise,
+/// returns `None`.
+pub fn builtin_signature_scheme(
```
miker83z: `is_`?

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -2007,4 +2046,29 @@ mod checked {
...
+    /// built-in signature verification.
+    fn authenticate_transaction_with_builtin_signature(
+        protocol_config: &ProtocolConfig,
+        authenticator: &MoveAuthenticator,
+        builtin_authenticator_data: PreloadedBuiltinAuthenticatorData,
+        tx_data_bytes: &[u8],
+        _gas_charger: &GasCharger,
+    ) -> Result<<execution_mode::Authentication as ExecutionMode>::ExecutionResults, ExecutionError>
+    {
+        // TODO: gas charging for built-in authentication.
```
miker83z: Should be a protocol param fixed cost (initially set to 0)

crates/iota-framework/packages/iota-framework/sources/account_abstraction/authenticator_function.move
valeriyr: fixed in c970250f01

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
valeriyr: fixed in c970250f01

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: I chose this name to make it open for adding new fields if it is necessary

crates/iota-types/src/account_abstraction/signature_scheme.rs
valeriyr: These are modules that can be used independently of each other. signature_scheme and public_key were created generic enough to be used by other parts of the framework or 3rd-party developers.
So I would prefer to leave them as is.

valeriyr: The prefix Move I added to make the name different from the original types to get rid of the misleading naming.

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: Еhe `is_` prefix usually implies that the function returns a boolean value, which is not the case. I would leave the name as is or choose a different option

iota-execution/latest/iota-adapter/src/execution_engine.rs
valeriyr: implemented here 3858814ab4

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
miker83z: `extract_`? or similar? While reading the code using this it was not clear the intention of this function

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
miker83z: What about also put in the name that this comes from the chain? Like `PreloadedOnchainBuiltinAuthenticatorData`

crates/iota-types/src/account_abstraction/signature_scheme.rs
miker83z: Well, the file name is also misleading in the case of signature_scheme.rs and public_key.rs for someone not knowing the context.

valeriyr: I am not sure I understand what you propose.
Adding all these types to the same file is wrong because they can be used separately from built-in authenticators. Built-in authenticators are implemented using these utilities, and, for example, ClaimManager(and not only) can be implemented using them.
We already have a bunch of structs in iota-types with a prefix Move, which means the struct is related to the Move side(MovePackage, MoveObject, MoveAuthenticator, etc). In this case, the names are too generic and widely used in the project, so it will be necessary to create aliases everywhere anyway.

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: renamed here 737f429368

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: What is the problem with the name?
Why is it important from where the data is loaded?
The proposed name is longer and limits the data that the struct can hold; it will be misleading to add any non chain data to it.

crates/iota-types/src/account_abstraction/signature_scheme.rs
valeriyr: For people who are out of context, a lot of documentation is added. I don't think that it is a naming problem. iota-type contains a lot of Rust structs that represent Move-side ones. New structs are placed in the `account_abstraction` folder, which contains Move-side modules. For someone who is out of context, everything will be misleading, while the changes are absolutely consistent.

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
miker83z: While reviewing it, from my perspective, it was just not clear what the role of this struct was given the name. After going through the whole creation and usage flow, then I understood it was "simply" the on chain stored public key.

crates/iota-types/src/account_abstraction/signature_scheme.rs
miker83z: Ok no problem

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: It contains not only the on-chain stored public key, but also a signature scheme resolved from the attached AuthenticatorFunctionRefV1

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
```diff
@@ -0,0 +1,288 @@
...
+    .map_err(|_| UserInputError::UnableToGetAccountPublicKeyId { account_object_id })?;
+
+    match get_object(public_key_field_id)? {
+        Some(object) => {
+            let metadata = DynamicallyLoadedObjectMetadata::from(&object);
+            let field: Field<PublicKeyFieldName, MovePublicKey> = object
+                .data
+                .try_as_move()
+                .expect("dynamic field should never be a package object")
+                .to_rust()
```
lollobene, line 158: consider to add a pubkey validation step here

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
```diff
@@ -0,0 +1,183 @@
...
+        scheme == signature_scheme::secp256r1() ||
+        scheme == signature_scheme::passkey()
+    ) {
+        assert!(len == SECP256_PUBLIC_KEY_LENGTH, EInvalidPublicKeyLength);
+    } else if (scheme == signature_scheme::multisig()) {
+        validate_multisig_public_key(raw_bytes);
+    } else {
+        abort EUnknownPublicKeyScheme
+    }
+}
```
lollobene, line 138: consider add proper validation of public keys here

lollobene (approved):

miker83z (approved):

crates/iota-framework/packages/iota-framework/sources/account_abstraction/public_key.move
valeriyr: This validation will be implemented in a dedicated PR.

crates/iota-types/src/account_abstraction/builtin_authenticator_functions.rs
valeriyr: It is added in a different place already, but a public key validation will additionally be implemented in a dedicated PR

PR #11187, opened 2026-04-08
refactor(dashboard): simplify transaction summary handling

brancoder:
- removes the redundant Extended Transaction on favor of transaction summary which holds the same data
- refactors Transaction Details layout


apps/wallet-dashboard/components/dialogs/transaction/TransactionDetailsLayout.tsx
```diff
@@ -35,14 +47,34 @@ export function TransactionDetailsLayout({ transaction, onClose }: TransactionDi
...
+        return (
+            <div className="flex h-full w-full justify-center">
+                <LoadingIndicator />
+            </div>
+        );
+    }
 
     return (
         <>
             <Header title="Transaction" onClose={onClose} />
```
evavirseda, line 74: when it is loading or there is an error, the `<Header>` isn’t shown, so the user can’t close the dialog 🙈 before, the header was always visible and the loading spinner was inside the content, so maybe we could keep the header visible and move the loading/error states inside the main view so the close action is always available

apps/wallet-dashboard/components/dialogs/transaction/TransactionDetailsLayout.tsx
```diff
@@ -35,14 +47,34 @@ export function TransactionDetailsLayout({ transaction, onClose }: TransactionDi
...
+                supportingText={
+                    error?.message ?? 'An error occurred when getting the transaction info'
+                }
+                icon={<Warning />}
+                style={InfoBoxStyle.Default}
+            />
+        );
+    }
+
+    if (!transaction || !summary) {
```
evavirseda, line 64: do we need the `!summary` check here? if transaction is defined, summary will always be too right?

evavirseda (changes requested):
🙏 🙏

apps/wallet-dashboard/components/dialogs/transaction/TransactionDetailsLayout.tsx
brancoder: the idea for the `!summary` was to we show the loader while the summary is being fetched and than display the receipt

evavirseda (approved):

brancoder:
moved to https://github.com/iotaledger/ts-packages/pull/94

PR #11189, opened 2026-04-08
fix(tests): make all temp_dir usages in tests nondeterministic to avoid collisions

muXxer:
# Description of change

Replace `std::env::temp_dir()` and raw `tempfile::tempdir()` with `iota_common::tempdir()` in tests.

## Summary                                                                                                                                                                                                                                                                  

  - Replace all uses of `std::env::temp_dir()` in test code with `iota_common::tempdir()`
  - Replace all direct `tempfile::tempdir()` calls in test code with `iota_common::tempdir()`
  - Add `iota-common` dev-dependency to `iota-data-ingestion-core` and `iota-archival`                                                                                                                                                                                                 

## Motivation

After upgrading RocksDB, unrelated PRs (still on the old version) started failing with "unsupported version 6" SST file errors. This revealed that CI test runners were sharing database paths because tests used `std::env::temp_dir()` directly, which always resolves to  the same `/tmp` directory. This also explains long-standing flaky tests where RocksDB complained about databases already being opened.

Additionally, several test files had local `temp_dir()` helpers that called `tempfile::tempdir()` directly, bypassing the `nondeterministic!()` macro. In `simtests`, deterministic RNG seeding can cause `tempfile::tempdir()` to produce identical paths across test runs, leading to the same class of collisions.

`iota_common::tempdir()` is the existing utility that solves both problems: it creates a unique random directory via `tempfile::tempdir()` and wraps it in `nondeterministic!()` to ensure uniqueness even under `simtest` determinism.                                            


crates/iota-core/src/epoch/consensus_store_pruner.rs
```diff
@@ -225,6 +225,12 @@ mod tests {
 
     use crate::epoch::consensus_store_pruner::{ConsensusStorePruner, Metrics};
 
+    fn temp_dir() -> std::path::PathBuf {
```
bingyanglin: So we have this `fn temp_dir() -> std::path::PathBuf` function in many files, do we consider move them to `iota_common` and reuse?

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -1049,27 +1049,33 @@ mod tests {
...
     async fn test_pruning_objects() {
-        let path = tempfile::tempdir().unwrap().keep();
+        let path = temp_dir();
         let to_keep = run_pruner(&path, 3, 2, 1000).await;
         assert_eq!(
             HashSet::from_iter(to_keep),
             get_keys_after_pruning(&path).unwrap()
         );
-        run_pruner(&tempfile::tempdir().unwrap().keep(), 3, 2, 1000).await;
+        run_pruner(&temp_dir(), 3, 2, 1000).await;
```
alexsporn: This line looks weird (was weird before), because it is calling the pruner on a different dir

crates/iota-core/src/epoch/committee_store.rs
```diff
@@ -62,8 +60,7 @@ impl CommitteeStore {
     }
 
     pub fn new_for_testing(genesis_committee: &Committee) -> Self {
-        let dir = std::env::temp_dir();
-        let path = dir.join(format!("DB_{:?}", nondeterministic!(ObjectID::random())));
```
Thoralf-M, line 66: this DB_ part is not needed?

muXxer: No I don't see where this would be used. I guess it was just to easier identify what's in the tempdir. We can remove that IMO.

crates/iota-core/src/authority/authority_store_pruner.rs
muXxer: I removed the line, it was redundant.

crates/iota-core/src/epoch/consensus_store_pruner.rs
muXxer: I removed all helpers, and moved the "expect" directly into the `iota_common::tempdir()`

crates/iota-data-ingestion-core/src/tests.rs
```diff
@@ -68,7 +68,7 @@ async fn run(
         None => {
             indexer
                 .run(
-                    path.into().unwrap_or_else(temp_dir),
+                    path.into().unwrap_or_else(|| iota_common::tempdir().keep()),
```
Thoralf-M: Why add .keep() to all the tests here?

crates/iota-common/src/random_util.rs
```diff
@@ -42,6 +42,6 @@ pub type TempDir = tempfile::TempDir;
 
 /// Creates a temporary directory with random name.
 /// Ensure the name is randomized even in simtests.
-pub fn tempdir() -> std::io::Result<TempDir> {
-    nondeterministic!(tempfile::tempdir())
+pub fn tempdir() -> TempDir {
+    nondeterministic!(tempfile::tempdir()).expect("Failed to open temporary directory")
```
piotrm50: as per code style guidelines
```suggestion
    nondeterministic!(tempfile::tempdir()).expect("temporary directory should not fail to be created")
```

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -1052,34 +1052,33 @@ mod tests {
     // Tests pruning old version of live objects.
     #[tokio::test]
     async fn test_pruning_objects() {
-        let path = tempfile::tempdir().unwrap().keep();
-        let to_keep = run_pruner(&path, 3, 2, 1000).await;
+        let path = iota_common::tempdir();
```
piotrm50: can we change the variable name so it doesn't look awkward calling `.path()` on a `path`?

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -2366,10 +2366,8 @@ async fn diagnose_split_brain(
         Datetime: {time:?}"
     );
     let fork_logs_text = format!("{header}\n\n{diff_patches}\n\n");
-    let path = tempfile::tempdir()
-        .expect("Failed to create tempdir")
-        .keep()
-        .join(Path::new("checkpoint_fork_dump.txt"));
+    let tmp = iota_common::tempdir();
+    let path = tmp.path().join(Path::new("checkpoint_fork_dump.txt"));
```
piotrm50: shouldn't we keep the dir so that the fork dump can be inspected after the test run?

Thoralf-M: Did you ever do this?

muXxer:
Nightly check was successful with current changes https://github.com/iotaledger/iota/actions/runs/24228238912

crates/iota-replay/src/config.rs
```diff
@@ -181,7 +181,7 @@ pub fn get_rpc_url(
 fn test_yaml() {
     let mut set = ReplayableNetworkConfigSet::default();
 
-    let path = tempfile::tempdir().unwrap().path().to_path_buf();
+    let path = iota_common::tempdir().keep().join("config.yaml");
```
Thoralf-M: Where is the config.yaml coming from?

Thoralf-M: Ah I remember you told me there was some weird logic before

Thoralf-M (approved):

bingyanglin (approved):

crates/iota-core/src/checkpoints/mod.rs
muXxer: He is right, that is production code and we should keep that.

roman1e2f5p8s (approved):

piotrm50 (approved):

thibault-martinez (approved):

tomxey (approved):

muXxer:
Nightly passed with latest changes https://github.com/iotaledger/iota/actions/runs/24334465075

Dkwcs (approved):
vm/lang files LGTM

PR #11196, opened 2026-04-09
feat(iota-graphql-rpc): add subscription recovery

### LINKED ISSUE #11199: [iota-graphql-rpc]: Add Subscriptions recovery
Refactor the `iota-graphql-rpc` crate to use the newly added changes from the `iota-indexer-streaming` crate to support stream recovery.

sergiupopescu199:
# Description of change

This PR adds subscription recovery support to `iota-graphql-rpc`, allowing clients to resume transaction and event subscriptions from a specific transaction digest after a disconnection.

When a `startAfter` digest is provided, the stream first backfills historical data from the database, then seamlessly transitions to live data from the broadcast channel. Without `startAfter`, behavior is unchanged.

fixes #11199 

- started a local network using the `iota-localnet`
```shell
cargo r --features indexer  -- start --force-regenesis --with-indexer --with-graphql
```
- First started a grpahql subscription through the grpahiql on `http://localhost:9125`
```graphql
subscription Transactions {
    transactions {
        __typename
        ... on TransactionBlock {
            digest,
           effects {
            checkpoint {
              sequenceNumber,
            }
          }
        }
        ... on Lagged {
            count
        }
    }
}
```
- Stopped the subscription and got the latest digest it returned. Used that digest to start a new subscription.

```graphql
subscription Transactions {
    transactions(startAfter: "5GJHkPzZq94pn74q5Mfd1GtZPC6TXNFN8a2sXYoEVkkd") {
        __typename
        ... on TransactionBlock {
            digest,
           effects {
            checkpoint {
              sequenceNumber,
            }
          }
        }
        ... on Lagged {
            count
        }
    }
}
```

- The client received historical transactions noticeably faster than live ones. Historical data is fetched from the database in batches, limited only by query speed. Live data depends on the network's checkpoint production rate, which is inherently slower. The debug logs confirmed the transition from historical backfill to live streaming.

### Release Notes
- [x] GraphQL: Subscription queries now accept an optional `startAfter` transaction digest to resume receiving transactions and events from a specific point. For more detailed information consult the graphql schema.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/subscription/mod.rs
+++ b/crates/iota-graphql-rpc/src/types/subscription/mod.rs
@@ -8,11 +8,11 @@ use futures::{Stream, StreamExt, TryStreamExt, future};
 use iota_indexer::read::IndexerReader;
 use iota_indexer_streaming::{
     error::IndexerStreamingError,
-    memory::{Config, InMemory},
+    memory::{Config, InMemory, RecoveryPoint},
     metrics::InMemoryStreamMetrics,
 };
 use iota_json_rpc_types::Filter;
-use iota_types::supported_protocol_versions::Chain;
+use iota_types::{digests::TransactionDigest, supported_protocol_versions::Chain};
 use prometheus::Registry;
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 use tracing::warn;
@@ -21,6 +21,7 @@ use crate::{
     error::Error,
     types::{
         chain_identifier::ChainIdentifierCache,
+        digest::Digest,
         event::Event,
         subscription::filter::{SubscriptionEventFilter, SubscriptionTransactionFilter},
         transaction_block::{TransactionBlock, TransactionBlockInner},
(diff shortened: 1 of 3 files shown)


tomxey (approved):

crates/iota-graphql-rpc/src/types/subscription/mod.rs
```diff
@@ -205,10 +237,13 @@ impl GraphQLStream {
...
     pub(crate) fn subscribe_events(
         &self,
+        start_after: Option<Digest>,
         filter: Option<SubscriptionEventFilter>,
     ) -> impl Stream<Item = Result<SubscriptionItem<Event>, Error>> {
         self.streamer
-            .subscribe_events(None)
+            .subscribe_events(
+                start_after.map(|digest| RecoveryPoint::Exclusive(TransactionDigest::from(digest))),
+            )
```
kodemartin, line 246: While for transactions it make sense to `start_after`, for events I think we must `start_from` and duplicate some events.

What happens if connections drops after we got the any transaction event that is not the latest within the transaction?

kodemartin (approved):
lgtm

crates/iota-graphql-rpc/src/types/subscription/mod.rs
sergiupopescu199: I think the Rust doc covers the question. The client tracks the last tx digest whose events they’ve fully consumed, so Exclusive is safe. I wanted to avoid using Exclusive for transactions and Inclusive for events, because the asymmetric semantics would be confusing for users of the API.

```
    /// Subscribe to incoming events from the IOTA network.
    ///
    /// If no filter is provided, all events will be returned.
    ///
    /// # Stream recovery
    ///
    /// Events are streamed in transaction order, all events from a single
    /// transaction arrive contiguously before any events from the next one.
    ///
    /// When `start_after` is provided with a transaction digest, the stream
    /// resumes from the transaction immediately after it.
    ///
    /// Because a single transaction can emit multiple events, a disconnection
    /// may leave the client with only a partial set of events from the
    /// transaction it was last receiving. A transaction is considered fully
    /// received only once the client has seen an event from the following
    /// transaction (a digest change signals the previous transaction is
    /// complete).
    ///
    /// On reconnect, clients should pass the digest of the last fully
    /// received transaction as `start_after`. The stream resumes from the
    /// next transaction, so no partial transaction is ever observed.
```

PR #11197, opened 2026-04-09
feat(graphql): Use backward history to produce consistent views

### LINKED ISSUE #11018: [graphql]: Refactor consistent and historic views on objects and coins

tomxey:
# Description of change

  Switch GraphQL object/coin/IOTA-names consistent and historical views from the forward diff approach (`objects_snapshot` + `objects_history`) to the backward diff approach (`checkpointed_objects` + `objects_backward_history`).

  **New read path (`backward_view/`):**
  - `backward_view::consistent` — reconstructs object state at a given checkpoint using `checkpointed_objects` as the current base and `objects_backward_history` to undo post-checkpoint changes. Merges `consistent_checkpointed_objects` (unchanged since the viewed checkpoint) with `consistent_historical_objects` (previous state of objects modified after the viewed checkpoint).
  - `backward_view::historical` — exact id+version lookups (`objectKeys`) without consistency filtering. Merges `checkpointed_objects` with `historical_objects`.
  - `merge_and_deduplicate` combines sources via `UNION ALL` + `DISTINCT ON`

  **New result type (`StoredBackwardObject` in `object.rs`):**
  - `QueryableByName` struct matching the backward diff UNION ALL output (no backing Diesel table)
  - Converts to `StoredHistoryObject` via `into_stored_history(checkpoint_viewed_at)`, filling `checkpoint_sequence_number` with `checkpoint_viewed_at`. This field is not exposed to users. It is used in one internal guard (`Loader<LatestAtKey>`) that checks whether an object existed at the queried checkpoint — setting it to `checkpoint_viewed_at` makes the guard a no-op, which is correct because the backward diff query already guarantees consistency.

  **Callers updated:**
  - `Object::paginate_subtype`, `Loader<LatestAtKey>`, `Coin::paginate`, `IotaNames::query_name_expiration` — all switched to backward diff
  - View selection: always use `backward_view::consistent` except for explicit object-key lookups which use `backward_view::historical`

  **NOT changed:**
  - `dynamic_field.rs` — still uses forward diff (`consistency.rs`)
  - `Loader<HistoricalKey>`, `Loader<ParentVersionKey>` — still query `objects_history` directly

  **Available range:**
  - Backward diff callers use `AvailableRange::is_checkpoint_in_backward_history_range` which reads from the `objects_backward_history` watermark (`min_available_cp`) instead of `objects_snapshot`. An additional max lookback of 900 checkpoints is enforced for performance.
  - Forward diff callers (`balance.rs`, `dynamic_field.rs`) continue using `AvailableRange::result` with `objects_snapshot`.
  - Removed unused `View::Historical` variant from `consistency.rs`.

  **Behavior change:** 
  - Wrapped/deleted objects that are returned from `objects_backward_history` report a lamport-1 version approximation instead of the exact tombstone version. This version may be higher than the actual tombstone version.
  - `objects(filter: {objectKeys:` lookups by the real tombstone version of a wrapped/deleted object return empty. Only the lamport-1 version (stored in `objects_backward_history`) is findable. This could be addressed in the future by joining against `objects_version`.

fixes https://github.com/iotaledger/iota/issues/11018

    - 3 new e2e tests:
      - `consistent_view_wrapped`: verifies WRAPPED_OR_DELETED is returned for objects wrapped at the viewed checkpoint but later unwrapped (exercises backward history Source B with lamport-1 version)
      - `consistent_view_no_filter`: verifies unfiltered `objects()` query returns consistent state at a past checkpoint, not the current state
      - `object_keys_tombstone`: documents the known limitation where `objectKeys` lookup by real tombstone version returns empty, while lamport-1 version returns WRAPPED_OR_DELETED


crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+        all_past_versions(page, &filter_fn),
+    )
+}
+
+/// Returns objects from `checkpointed_objects` (including tombstones) that
+/// were not superseded after the given checkpoint.
+///
+/// Uses a LEFT JOIN against `objects_backward_history` to exclude objects
+/// that have any entry with `superseded_at_checkpoint > cv`.
+fn non_superseded_live_objects(
```
kodemartin: So these are essentially the checkpointed objects that were consistent at the given checkpoint, right?

If so, I think that referring to "checkpointed_object what were not superseded" might be a bit confusing. Why not:

 
```rust
/// Returns objects from `checkpointed_objects` (including tombstones) that
/// were consistent also at the given checkpoint.
///
/// Uses a LEFT JOIN against `objects_backward_history` to exclude objects
/// that have any entry with `superseded_at_checkpoint > cv`.
fn consistent_checkpointed_objects(
```

crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+    cv: i64,
```
kodemartin: Better to be more verbose here

```rust
    checkpoint_viewed_at: i64,
```

crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Returns previous versions of objects that were superseded after the given
+/// checkpoint.
+///
+/// Picks the earliest superseded version (`MIN(object_version)`) per object,
+/// which represents the state just before the first change after the target
+/// checkpoint. Excludes `NOT_YET_CREATED` entries (objects that didn't exist
+/// at the target checkpoint).
+fn superseded_past_versions(
```
kodemartin: Similar reasoning here. This returns objects from `backwards_history` that were consistent at the given checkpoint.  Consistent is indeed a consequence of the semantics of "superseded_at_checkpoint" but the referring to the latter is not as clear.

So why not


```rust
/// Returns objects from `backwards_history` that were consistent at the given checkpoint.
///
/// Picks the earliest superseded version (`MIN(object_version)`) per object,
/// which represents the state just before the first change after the target
/// checkpoint. Excludes `NOT_YET_CREATED` entries (objects that didn't exist
/// at the target checkpoint).
fn consistent_historical_objects(
```

crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+               AND candidates.object_version = oldest.min_version"#,
+        history_window,
+        oldest_subquery
+    );
+    page.apply::<StoredBackwardObject>(source)
+}
+
+/// Returns all objects from `checkpointed_objects` (including tombstones)
+/// without consistency filtering.
+fn all_live_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQuery) -> RawQuery {
```
kodemartin: We can simplify and be more transparent here as well


```rust
/// Returns all objects from `checkpointed_objects` (including tombstones)
/// that satisfy the provided filter.
fn checkpointed_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQuery) -> RawQuery {
```

crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+    let source = query!(
+        "SELECT candidates.* FROM ({}) candidates",
+        checkpointed_filtered
+    );
+    page.apply::<StoredBackwardObject>(source)
+}
+
+/// Returns all past versions from `objects_backward_history`, excluding
+/// `NOT_YET_CREATED` entries.
+fn all_past_versions(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQuery) -> RawQuery {
```
kodemartin: Similarly

```rust
/// Returns all objects from `objects_backward_history` that satisfy the provided filter, excluding
/// `NOT_YET_CREATED` entries.
fn historical_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQuery) -> RawQuery {
```

crates/iota-graphql-rpc/src/consistency.rs
```diff
@@ -24,6 +24,7 @@ pub(crate) enum View {
     /// Return objects that fulfill the filtering criteria, even if there are
     /// more recent versions of the object within the checkpoint range. This
     /// is used for lookups such as by `object_id` and `version`.
+    #[expect(dead_code)] // Matched in build_objects_query arms; no caller currently passes it
     Historical,
```
kodemartin: So why don't we remove it?

crates/iota-graphql-rpc/src/types/iota_names_registration.rs
```diff
@@ -520,22 +520,26 @@ impl IotaNames {
...
+                    BackwardView::Consistent {
+                        checkpoint_viewed_at,
+                    },
                     &page,
                     move |query| filter.apply(query),
-                    move |newer| newer,
                 );
 
-                let objects: Vec<StoredHistoryObject> =
+                let backward_objects: Vec<super::object::StoredBackwardObject> =
```
kodemartin: If you must use the fully qualified type, opt for the absolute module path, as it is more robust against future refactors
```rust
                let backward_objects: Vec<crate::types::object::StoredBackwardObject> =
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -871,14 +874,14 @@ impl Object {
 
         let Some((prev, next, results)) = db
             .execute_repeatable(move |conn| {
-                let Some(range) = AvailableRange::result(conn, checkpoint_viewed_at)? else {
+                if AvailableRange::result(conn, checkpoint_viewed_at)?.is_none() {
```
kodemartin: Don't we need to refactor `AvailableRange::result` to account for the actual range of the `backwards_history` table?

crates/iota-graphql-rpc/src/backward_consistency.rs
```diff
@@ -0,0 +1,192 @@
...
+) -> RawQuery {
+    merge_and_deduplicate(
+        non_superseded_live_objects(cv, page, &filter_fn),
+        superseded_past_versions(cv, page, &filter_fn),
+    )
+}
+
+/// Builds a historical view by merging all live objects with all past
+/// versions, without consistency filtering.
+fn build_historical_query(
```
kodemartin: What speaks against exposing these functions as `pub(crate)` and getting rid of the generic `build_backwards_object_query` and the associated `View`?

You could even split the logic into two distinct modules: `view::{consistent,historical}`. This would be simpler and would separate concerns.

crates/iota-graphql-rpc/src/consistency.rs
tomxey: there is still some forward consistency code that uses it that I wanted to keep for now, the code that uses this enum exists, it's just not called.
or do you prefer to delete the unused forward consistency code right away?

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/types/iota_names_registration.rs
tomxey: 0f1622c8b4

crates/iota-graphql-rpc/src/types/object.rs
tomxey: I wanted for now to keep the range as it was, and we can extend it later when we decide to do so in a separate issue,
or do you mean to handle the "warmup" case when indexer is deployed with the backward history migration and the table is initially empty?
I thought that maybe we could handle that with just a release note that indexer needs 30minutes to warm up, so the deployment should take that into account (e.g. switch to the new instance only after 30 minutes pass)

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: no specific reason, I was just keeping the old code organization where we had the view enum
maybe such single enum is better at showing what possible query types we can make, instead of a bunch of scattered functions?

if you prefer I can make the split

crates/iota-graphql-rpc/src/consistency.rs
kodemartin: yes if it is dangling either way, we might as well remove it

crates/iota-graphql-rpc/src/types/object.rs
kodemartin: No, I meant as opposed to using `objects_snapshot` to get the lower bound and check if `checkpoint_viewed_at` falls within it

crates/iota-graphql-rpc/src/backward_consistency.rs
kodemartin: I think it would benefit reflecting the different semantics between consistent and historical queries on the module structure. 

The original implementation used `consistency` in the similar way indeed coupling everything under the same functions. One could argue that this is warranted by the use of the single table, but this doesn't mean that historical queries relate to consistency. Hence it seems plausible to me to separate them.

You actually already separated them nicely through the functions, why stop there and not extend the separation into modules as well?

crates/iota-graphql-rpc/src/consistency.rs
tomxey: removed cc40493975

crates/iota-graphql-rpc/src/types/object.rs
tomxey: switched to checking backward history watermark to determine the range: b6dc316db6
I still left some cap (900 checkpoints for now, we can change it later), to avoid queries for state many thousands of checkpoints ago, which could overload the DB.

crates/iota-graphql-rpc/src/backward_consistency.rs
tomxey: split the module in cb02ddda5e

kodemartin (approved):
lgtm 🚢

crates/iota-graphql-e2e-tests/tests/consistency/consistent_view_wrapped.move
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2025 IOTA Stiftung
```
sergiupopescu199:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-graphql-e2e-tests/tests/consistency/consistent_view_no_filter.move
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2025 IOTA Stiftung
```
sergiupopescu199:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-graphql-e2e-tests/tests/consistency/object_keys_tombstone.move
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2025 IOTA Stiftung
```
sergiupopescu199:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

sergiupopescu199 (approved):
lgtm!! 🦫

crates/iota-graphql-e2e-tests/tests/consistency/consistent_view_wrapped.move
tomxey: dd958c18ae

crates/iota-graphql-e2e-tests/tests/consistency/consistent_view_no_filter.move
tomxey: dd958c18ae

crates/iota-graphql-e2e-tests/tests/consistency/object_keys_tombstone.move
tomxey: dd958c18ae

PR #11206, opened 2026-04-09
feat(iota-types)!: update to MoveStruct with contents checked

Thoralf-M:
https://github.com/iotaledger/iota-rust-sdk/issues/1043


crates/iota-types/src/object.rs
```diff
@@ -182,9 +178,9 @@ impl MoveObjectExt for MoveObject {
     fn set_coin_value_unchecked(&mut self, value: u64) {
         debug_assert!(self.object_type.is_coin());
         // 32 bytes for object ID, 8 for balance
-        debug_assert!(self.contents.len() == 40);
+        debug_assert!(self.contents().len() == 40);
 
-        self.contents.splice(ID_END_INDEX.., value.to_le_bytes());
+        self.contents_mut_unchecked()[ID_END_INDEX..].copy_from_slice(&value.to_le_bytes());
```
DaughterOfMars: This method is a bit weird. Couldn't we just construct a new MoveObject and set it like `*self = new_val;`?

piotrm50 (approved):

bingyanglin (approved):

crates/iota-types/src/object.rs
thibault-martinez: I kind of agree

Thoralf-M: Really want to create a whole new MoveObject or call set_contents() instead on it?

thibault-martinez: I would have a slight preference for using `set_contents` indeed. But not a strong no from me, we can keep this.

DaughterOfMars: set_contents is better, yes

DaughterOfMars (approved):

thibault-martinez:
CI failing

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1629,7 +1631,8 @@ mod checked {
     ) -> Result<MoveObject, ExecutionError> {
         debug_assert_eq!(
             id,
-            MoveObject::id_opt(&contents).expect("object contents should start with an id")
+            ObjectID::from_bytes(&contents[..ObjectID::LENGTH])
```
thibault-martinez, line 1634: Where does `contents` come from? This would panic if it wasn't at least `ObjectID::LENGTH`. Maybe add a verification layer like in L1584?

Thoralf-M: From the VM execution result, should be safe, the logic is same as before

thibault-martinez (approved):

crates/iota-open-rpc/src/examples.rs
thibault-martinez, line 1: Why do we need to do this?

Thoralf-M: The previous example created an invalid move object because it was just an empty vec

thibault-martinez: Ah, right

thibault-martinez (approved):

Issue #10955, opened 2026-03-25
gRPC: cleanup index table migration code


muXxer:
We need to clean up our migration code that creates the new tables like `coin_v2`, `package_versions` etc.

PR #11209, opened 2026-04-09
fix(gRPC): remove legacy index migrations and rename v2 tables to final names

bingyanglin:
# Description of change

- Remove deprecated column families (`transactions`, `owner`, `dynamic_field`, `coin`) and their `#[deprecated_db_map]` annotations, migration functions, `LegacyDynamicFieldIndexInfo`, and the legacy directory migration (`migrate_legacy_dirs`, `LEGACY_INDEX_DIR`)
- Remove background backfill infrastructure: `BackfillTask`/`BackfillIndexer`/`BackfillBatchIndexer`, readiness flags, `Watermark` backfill variants, `IndexBackfillInProgressError`, `GrpcReader` readiness guards, and `is_*_index_ready()` trait methods
- Rename all v2 types, methods, and table fields to their final names (e.g. `OwnerIndexKeyV2` -> `OwnerIndexKey`, `CoinIndexInfoV2` -> `CoinIndexInfo`, `owner_v2`/`dynamic_field_v2`/`coin_v2` -> `owner`/`dynamic_field`/`coin`, `OwnedObjectV2Cursor` -> `OwnedObjectCursor`, `iota_types::storage::CoinInfoV2` -> `CoinInfo` — the unused pre-regulated `CoinInfo` struct was removed to free the name)
- Extract a generic `migrate_cf` helper in `grpc_indexes.rs` that batch-copies entries through a caller-supplied map function; the three remaining CF migrations are now thin wrappers around it

Parts of #10955 

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Remove deprecated gRPC index tables. ⚠️⚠️⚠️ The migration may incur a downtime of several minutes upon restart of the full nodes, so operators are requested to stage it according to their SLA needs.⚠️⚠️⚠️
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-core/src/grpc_indexes.rs
+++ b/crates/iota-core/src/grpc_indexes.rs
@@ -31,6 +31,7 @@ use serde::{Deserialize, Serialize};
 use tracing::{debug, info};
 use typed_store::{
     DBMapUtils, TypedStoreError,
+    database::Database,
     rocks::{DBMap, MetricConf},
     traits::Map,
 };
@@ -50,12 +51,7 @@ use crate::{
 /// dedicated `Watermark` variants (`PackageVersionBackfilled`,
 /// `CoinV2Backfilled`, `OwnerV2Backfilled`).  While the backfill runs, affected
 /// endpoints return `Code::Unavailable` with a `RetryInfo` hint.
-///
-/// Version history:
-/// - 1: Initial version with deprecated CF migration + background backfill.
-/// - 2: Removed deprecated CFs (`transactions`, `owner`, `dynamic_field`,
-///   `coin`) and background backfill infrastructure.
-const CURRENT_DB_VERSION: u64 = 2;
+const CURRENT_DB_VERSION: u64 = 1;
 
 /// On-disk directory name for the gRPC indexes store.
 pub const GRPC_INDEXES_DIR: &str = "grpc_indexes";
@@ -361,6 +357,13 @@ struct IndexStoreTables {
     /// the main database.
     transaction_checkpoints: DBMap<TransactionDigest, CheckpointSequenceNumber>,
 
+    /// Deprecated: migrated to `owner`.
+    // TODO(cleanup): Remove this field and `migrate_owner_v2_to_owner` in the
+    // next release once all nodes have migrated.
+    #[allow(dead_code)]
+    #[deprecated_db_map(migration = "migrate_owner_v2_to_owner")]
+    owner_v2: Option<DBMap<OwnerIndexKey, OwnerIndexInfo>>,
+
     /// An index of object ownership.
     ///
     /// Uses fixed-size u64 hash keys for correct RocksDB byte-order iteration.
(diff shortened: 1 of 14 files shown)


crates/iota-core/src/grpc_indexes.rs
```diff
@@ -54,15 +50,16 @@ use crate::{
 /// dedicated `Watermark` variants (`PackageVersionBackfilled`,
 /// `CoinV2Backfilled`, `OwnerV2Backfilled`).  While the backfill runs, affected
 /// endpoints return `Code::Unavailable` with a `RetryInfo` hint.
-const CURRENT_DB_VERSION: u64 = 1;
+///
+/// Version history:
+/// - 1: Initial version with deprecated CF migration + background backfill.
+/// - 2: Removed deprecated CFs (`transactions`, `owner`, `dynamic_field`,
+///   `coin`) and background backfill infrastructure.
+const CURRENT_DB_VERSION: u64 = 2;
```
muXxer: Wait, why do we bump the db version? The whole idea to introduce these tables and backfill was that we don't need to bump and completely reindex everything.

muXxer (changes requested):

crates/iota-core/src/grpc_indexes.rs
bingyanglin: In Rust side we have `owner_v2` → `owner`, `dynamic_field_v2` → `dynamic_field`, `coin_v2` → `coin`, and then in the on-disk CF names I also renamed to make it clean so needs DB bump. Or we keep `_v2` for the on-disk CF names? But in this way we should also use `v2` in the Rust side otherwise it's confusing?

bingyanglin: Or we do migrate again for those v2 fields to the non-v2 fields, and then next next release we remove the v2 ones? So after 2 more releases we can make the CF clean and also rust field clean?

bingyanglin: After discussion we use 2 more releases to make it clean and wo version bump.

crates/iota-grpc-server/src/lib.rs
```diff
@@ -32,5 +32,5 @@ pub use state_service::StateGrpcService;
 pub use transaction_execution_service::TransactionExecutionGrpcService;
 pub use types::{
     DynamicFieldIterItem, GrpcCheckpointDataBroadcaster, GrpcReader, OwnedObjectIterItem,
-    OwnedObjectV2Cursor, OwnedObjectV2IterItem, PackageVersionIterItem,
+    OwnedObjectV2Cursor, PackageVersionIterItem,
```
muXxer: Can we rename all the V2 types as well?

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -503,26 +452,59 @@ fn migrate_dynamic_field_to_v2(db: &std::sync::Arc<Database>) -> Result<(), Type
...
+        batch.insert_batch(&new, std::iter::once((key, ())))?;
+        count += 1;
+        if count.is_multiple_of(BATCH_SIZE) {
+            batch.write()?;
+            batch = new.batch();
+        }
+    }
+    if !count.is_multiple_of(BATCH_SIZE) {
+        batch.write()?;
+    }
```
muXxer: Maybe we should extract an utility function for this, and put it somewhere? Including a map function for every entry in the iterator.

muXxer (approved):

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Addressed at [adaca9c](https://github.com/iotaledger/iota/pull/11209/commits/adaca9c3f76e141a6b1864c21336ba8b003ec3c0). Thanks.

crates/iota-grpc-server/src/lib.rs
bingyanglin: Addressed at [adaca9c](https://github.com/iotaledger/iota/pull/11209/commits/adaca9c3f76e141a6b1864c21336ba8b003ec3c0). Thanks.

muXxer (approved):

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -48,21 +45,11 @@ use crate::{
 /// Bump this when changing the serialization format of an existing table.
 /// A version mismatch triggers a full re-index via
 /// `needs_to_do_initialization`.
-///
-/// NOTE: Adding a *new* table does NOT require a version bump.  New tables
-/// start empty and are populated by a background backfill task tracked via
-/// dedicated `Watermark` variants (`PackageVersionBackfilled`,
-/// `CoinV2Backfilled`, `OwnerV2Backfilled`).  While the backfill runs, affected
-/// endpoints return `Code::Unavailable` with a `RetryInfo` hint.
 const CURRENT_DB_VERSION: u64 = 1;
```
semenov-vladyslav, line 48: Please, fix the PR description saying it bumps `CURRENT_DB_VERSION` to 2, which is not the case.

semenov-vladyslav (approved):

crates/iota-core/src/grpc_indexes.rs
bingyanglin: Good catch! I refined all the PR description to make it short and clear. Thanks!

polinikita (approved):

PR #11235, opened 2026-04-13
feat(iota-core, iota-types, iota-grpc-server): skip effects certification when waiting for checkpoint execution

### LINKED ISSUE #11252: pcool: Skip effects certification when waiting for checkpoint execution
## Description

When a caller requests `WaitForLocalExecution`, the TransactionDriver should skip the 2f+1 effects digest acknowledgments broadcast and fetch effects from a single validator instead. Finality comes from the local checkpoint executor rather than the certification round.

This eliminates redundant network round-trips for callers that already wait for checkpoint inclusion, since the checkpoint provides equivalent finality guarantees.

## Scope

- **`EffectsCertifier::get_effects_without_certification`** — fetches effects from one validator, no quorum broadcast. Response tagged `PendingCheckpointExecution`.
- **`drive_transaction`** — accepts `request_type: Option<ExecuteTransactionRequestType>`. `WaitForLocalExecution` triggers the skip-cert path.
- **`execute_transaction_v1` (gRPC path)** — accepts `request_type`. The gRPC handler derives it from `checkpoint_inclusion_timeout_ms`: positive timeout → `WaitForLocalExecution`; otherwise → `WaitForEffectsCert`.
- **`execute_transaction_block` (JSON-RPC path)** — after a successful local-execution wait, uncertified effects are replaced with the authoritative copies from the local cache. Finality upgraded to `QuorumExecuted`. Digest mismatches are logged as warnings (byzantine submitter detection).
- **Safety guard** — any `PendingCheckpointExecution` response at return time is rejected with `debug_fatal!`.

## Implementation

PR #11235

piotrm50:
# Description of change

When a caller asks the JSON-RPC `execute_transaction_block` for
`WaitForLocalExecution` (or the gRPC `execute_transactions` with
`checkpoint_inclusion_timeout_ms > 0`), the TransactionDriver now **skips
the 2f+1 effects-digest broadcast** and fetches effects from a single
validator. Finality comes from the local checkpoint executor — once the
fullnode observes the tx in a checkpoint, the cache holds authoritative
effects and the single-validator copies can be discarded.

This shaves a network round-trip per request off the fullnode tx path.

### TransactionDriver
- `EffectsCertifier::get_effects_without_certification` — fetches effects
  from one validator, no quorum broadcast. Response tagged
  `EffectsFinalityInfo::UncertifiedSingleValidator(authority)`.
- `corroborate_single_validator_error` — when the single-validator fetch
  fails, poll the rest of the committee for `TxStatusUpdate` to decide
  whether the tx was rejected by f+1 (`RejectedByValidators`) or is still
  in flight (`SubmittedButFetchFailed`, retriable).
- `drive_transaction` takes `skip_certification: bool`.

### Orchestrator (`transaction_orchestrator.rs`)
- `submit_with_checkpoint_race` — races the driver future against
  `wait_for_checkpoint_inclusion`; `biased` so the driver result is
  preferred when both are ready.
- `reconcile_effects_from_cache` — **always full-rebuilds** the response
  from the local cache (effects, events, input/output objects). Replaces
  any submitter-supplied data including events; logs a warning on
  effects-digest mismatch (possible byzantine submitter).
- `build_response_from_cache` — used when only the checkpoint future
  yielded a result; cache miss is mapped to `TimeoutBeforeFinality`
  (retriable) rather than `QuorumDriverInternal`.
- Finality is upgraded to `EffectsFinalityInfo::Checkpointed(epoch, seq)`
  before returning.

### gRPC (`transaction_execution_service`)
- `execute_transactions` derives `skip_certification` from
  `checkpoint_inclusion_timeout_ms` (positive → skip-cert).
- `finalize_item` rebuilds each skip-cert item from the executor's cache
  after the checkpoint wait. `Ok(None)` from the executor (no local
  cache, e.g. simulacrum misconfigured for skip-cert) is now an
  `Internal` error rather than silently returning uncertified data.

### Safety guard
At the end of `execute_transaction_block`, any response still carrying
`UncertifiedSingleValidator` finality is converted to
`QuorumDriverInternal` and a `debug_fatal!` fires in dev. All reachable
paths upgrade finality before this point; the guard is a refactor-proof
sentinel — see the comment around line 380.

### Infrastructure
- `#![recursion_limit = "256"]` on `iota-json-rpc` — the new
  `submit_with_checkpoint_race` adds nested generic Future types that
  push rustc's monomorphization query past the default 128. Same pattern
  already in `iota-indexer`. Comment in the file explains the chain and
  the box-pin escalation path if a future change overruns 256.

## Reviewer focus

- `reconcile_effects_from_cache` and `build_response_from_cache` in
  `transaction_orchestrator.rs` — these are the bytes the client sees;
  the byzantine-submitter argument lives here.
- The safety guard at the end of `execute_transaction_block` — please
  verify the comment about reachable paths matches your reading.
- `corroborate_single_validator_error` — f+1 rejection threshold logic.
- The recursion_limit bump in `iota-json-rpc/src/lib.rs` (cosmetic but
  flagged explicitly).

## Known gaps / follow-ups

- No e2e test for the byzantine-events-mismatch path; that requires a
  spoofed-validator harness that doesn't exist today. Tracked as a
  follow-up. Code-level argument: `reconcile_effects_from_cache` always
  rebuilds from cache, so submitter-supplied events/objects are
  unconditionally discarded.
- `submit_with_checkpoint_race` does not await the checkpoint future on
  the driver-error branch (`?` short-circuits). Narrow window (only when
  driver returns a terminal error *and* the tx already landed in a
  checkpoint); accepted for this PR.

## Base branch

This PR targets `consensus/feat/pcool-feature` rather than `develop` — it
stacks on the pcool-feature consensus work.

Closes #11252

New tests:

- `test_skip_effect_cert_reconciles_to_checkpointed` — JSON-RPC happy
  path; verifies finality is upgraded to `Checkpointed` and the request
  flags populate input/output objects.
- `test_skip_effect_cert_respects_request_flags` — JSON-RPC; verifies
  events/objects are not leaked when the read mask doesn't request them.
- `test_skip_effect_cert_timeout_without_quorum` — JSON-RPC under
  consensus quorum loss; verifies the orchestrator surfaces
  `TimeoutBeforeFinality` (not `QuorumDriverInternal`).
- `execute_transaction_v1_skip_cert_rebuilds_from_cache` — gRPC happy
  path; checkpoint + timestamp populated proves the cache rebuild ran.
- `execute_transaction_v1_skip_cert_no_quorum_yields_per_item_error` —
  gRPC under quorum loss; verifies per-item `DeadlineExceeded`/`Internal`
  rather than a success with uncertified data.
- `test_get_effects_without_certification_corroboration_*` — unit
  coverage for the corroboration error classification (in iota-core).

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): TransactionDriver can now skip
  the 2f+1 effects certification when the caller waits for local
  checkpoint execution, reducing fullnode tx-execution latency by one
  network round-trip. Behind the white-flag protocol flag.
- [ ] Indexer:
- [x] JSON-RPC: `execute_transaction_block` with `WaitForLocalExecution`
  uses the skip-cert path when white flag is enabled.
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] gRPC: `execute_transactions` with
  `checkpoint_inclusion_timeout_ms` > 0 uses the skip-cert path when
  white flag is enabled.


piotrm50:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

crates/iota-core/src/transaction_orchestrator.rs
```diff
@@ -253,54 +254,290 @@ where
...
+        // the cache has authoritative effects.
+        let needs_cache_rebuild = matches!(
+            response.as_ref().map(|r| &r.effects.finality_info),
+            None | Some(EffectsFinalityInfo::UncertifiedSingleValidator(_)),
+        );
+
+        let executed_locally = if !wait_for_local_execution {
+            false
+        } else if needs_cache_rebuild {
+            let Some(seq) = seq else {
```
bingyanglin, line 312: When `seq` is `None` and response is `Some`, this returns the effects straight to the client. Problem is those effects are `UncertifiedSingleValidator`, and they came from the single submitting validator and a byzantine one could have forged them. The comment says this "falls through to the safety guard," but a let-else else block can't fall through, it returns `Ok` right here, so the guard below never runs. The gRPC side handles the same case in `finalize_item` by returning `DeadlineExceeded` instead of leaking. Maybe just return the timeout error for the Some case too like the following?

```rust
let Some(seq) = seq else {
    // Not checkpointed in time. Any Some(response) here is uncertified
    // single-validator data: don't leak it.
    return Err(QuorumDriverError::TimeoutBeforeFinality);
};
```

crates/iota-grpc-server/src/transaction_execution_service/mod.rs
```diff
@@ -274,78 +275,330 @@ pub async fn execute_transactions(
...
+    config: &iota_config::node::GrpcApiConfig,
+    read_mask: &FieldMaskTree,
+    flags: &ReadFlags,
+    mut original: ExecuteTransactionResult,
+    digest: &TransactionDigest,
+    rebuild_ctx: Option<RebuildCtx>,
+    checkpoint_entry: Option<(u64, u64)>,
+    wait_error: Option<&str>,
+) -> ExecuteTransactionResult {
+    let Some((seq, ts)) = checkpoint_entry else {
```
bingyanglin, line 442: This same "rebuild the effects or return an error" check exists in two spots: here, and in `finalize_item` on the gRPC side. They don't match anymore: the gRPC one returns an error on timeout, this one leaks the uncertified effects. Maybe put the check in one shared helper so both paths do the same thing to make them consistent?

crates/iota-core/src/transaction_orchestrator.rs
piotrm50: ah, good catch! fixed.

crates/iota-grpc-server/src/transaction_execution_service/mod.rs
piotrm50: after the issue with JSON API leak is fixed I don't think it makes sense to extract a common function because the `finalize_item` and the JSON API check don't share that much logic and they operate on different types, so having an additional method would be a bit awkward.

bingyanglin (approved):

tomxey (approved):

VorobyevIlya (approved):

PR #11240, opened 2026-04-15
refactor(iota-core): safe renaming of certificate occurrences to transaction

piotrm50:
## Summary

Re-applies the rename originally merged as #10767, which was temporarily dropped from the feature branch to reduce rebase conflict surface during active development.

This PR restores the internal identifier rename (`certificate` → `transaction`) on the feature branch, with conflicts against develop's multi-authenticator API properly resolved.

## Background

The rename was first merged to the feature branch via #10767. During a subsequent rebase onto develop, the merge conflict in `authority.rs` around the multi-authenticator code was resolved incorrectly, producing an inverted `if let Some(...)` / `else` that regressed `unchanged_shared` tracking — this caused failures in `abstract_account/*` snapshot tests and `child_count/count_decremented_v9`. See branch `consensus/feat/transaction-certificateless-flow-feature-backup-2026-04-10` for the buggy state.

## What's here

- The same rename content as the original #10767 PR
- Conflicts against current feature branch (post-multi-authenticator API) resolved correctly:
  - `authority.rs` `prepare_transaction` multi-auth branching preserved (HEAD structure, `transaction` rename applied)
  - `process_certificate` → `process_transaction` rename
  - `iota-transactional-test-runner/src/lib.rs` uses `TxLockGuard` with develop's current tuple shape

## Test plan

- [x] `cargo +nightly fmt --check` passes
- [x] `cargo ci-clippy` passes
- [x] `cargo test -p iota-adapter-transactional-tests -- abstract_account/authenticator/deny_list_v1` passes
- [x] `cargo test -p iota-adapter-transactional-tests -- child_count/count_decremented_v9` passes

## Why draft

Keeping this separate from the feature branch reduces rebase burden during feature development. Merge once the feature branch is settling down.

Closes #10908 (same as #10767, since this re-applies that work).

PR #11205, opened 2026-04-09
feat(iota-core, iota-types, iota-network): remove V1 white flag endpoints, types and tests

filipdulic:
# Description of change

Remove the V1 white-flag endpoints, request/response types, proto service methods, and associated tests from the validator stack now that their V2 replacements are fully in place.

### What was removed

- **V1 gRPC service methods** (`SubmitTransactionsV1`, `WaitForEffectsV1`, `ValidatorHealthV1`) from the `Validator` service definition in `iota-network/build.rs`.
- **V1 server handlers** (`handle_submit_transactions_impl`, `handle_wait_for_effects_impl`, `handle_validator_health_impl`) and all supporting helper functions from `authority_server/validator.rs`.
- **V1 client trait methods** (`handle_submit_transactions`, `handle_wait_for_effects`, `handle_validator_health`) from `AuthorityAPI` in `authority_client/validator.rs`, `safe_client.rs`, and `test_authority_clients.rs`.
- **V1 domain types** (`SubmitTransactionsRequest`, `SubmitTransactionResult`, `WaitForEffectsRequest`, `WaitForEffectsResponse`, `ValidatorHealthRequest`, `ValidatorHealthResponse`, and all `Raw*` prost wrappers/conversions) from `iota-types/messages_grpc.rs` and `iota-network/convert/validator_v2.rs`.
- **V1-specific error variant** (`FullNodeCantHandleSubmitTransactions`) from `iota-types/error.rs`.
- **V1 server tests** (submit transactions, wait-for-effects, validator health, and ping tests) from `unit_tests/server_tests.rs`, replaced with V2 equivalents.
- **V1 call sites** in `transaction_driver`, `transaction_orchestrator`, and `validator_tx_finalizer` migrated to use V2 APIs (`submit_tx`, `get_tx_status`, `health_check`).

## Links to any relevant issues

fixes #11204

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ]  I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

PR #11242, opened 2026-04-15
feat(iota-core): certificate-less transaction flow (P-COOL)

piotrm50:
# Description of change

Umbrella PR for the certificate-less transaction flow (P-COOL / white-flag) feature. Aggregates the sub-PRs that have landed on `consensus/feat/transaction-certificateless-flow-feature` and brings the feature to `develop`.

## Scope

Sub-PRs included:

- PoC post-consensus conflict resolution (#10379)
- Soft bundle support in p-cool flow (#10632)
- Transaction validation alignment for certificate-less flow (#10633)
- Post-consensus validation fixes (#10774)
- ValidatorHealth metrics wiring (#10812)
- Certificate-less epoch boundary coordination (#10615)
- Batch response and multi-transaction wait-for-effects (#10814)
- ValidatorV2 / ValidatorPeer proto scaffolding (#10924)
- Dropped tx notification race fix and categorization (#10813)
- Split \`iota-core\` authority_client/server into per-service modules (#10954)
- Protocol-version-aware validation for UserTransactionV1 (#11097)
- Wire-format server-side implementation (#10978)
- Client-side integration of ValidatorV2 / ValidatorPeer (#11150)
- Remove V1 white-flag endpoints, types and tests (#11205)
- \`notify_capabilities_v2\` in test mock clients (#11151)
- Skip duplicate EndOfPublish in white-flag mode (#11152)
- Retry with backoff for EndOfPublish submission (#11199)
- Benchmark mode and white-flag env vars in \`iota-private-network\` (#11224)
- Resolution of validator gRPC migration TODOs (#11220)
- Keep already-executed transactions in white-flag checkpoint roots (#11662)
- Notify for post-consensus dropped transactions (#11672)

The internal identifier rename (#10767) is tracked separately in #11240 and will be applied after this PR lands.

Tracking: Core Consensus project board.


dev-tools/iota-private-network/README.md
```diff
@@ -91,6 +91,49 @@ To bring up 10 validators and faucet:
...
+Replace `--transfer-object 100 --shared-counter 0` with different ratios to test other workloads:
+
+- **Shared-object only:** `--transfer-object 0 --shared-counter 100 --shared-counter-hotness-factor 50`
+- **Mixed (50/50):** `--transfer-object 50 --shared-counter 50`
+
+Use `--benchmark-stats-path /tmp/bench_results.json` to save results, and `--compare-with /tmp/bench_baseline.json` to compare runs.
+
+### Optional: Selecting a Consensus Protocol
+
+You can run the network with an optional consensus protocol flag. There are two options `starfish` and `mysticeti`.
```
muXxer: We need to remove this section. There is no other consensus protocol anymore now.

dev-tools/iota-private-network/docker-compose.yaml
```diff
@@ -6,6 +6,10 @@ x-common-validator: &common-validator
     - RPC_WORKER_THREAD=12
     - NEW_CHECKPOINT_WARNING_TIMEOUT_MS=30000
     - NEW_CHECKPOINT_PANIC_TIMEOUT_MS=60000
+    - CONSENSUS_PROTOCOL=${CONSENSUS_PROTOCOL:-}
```
muXxer: `CONSENSUS_PROTOCOL` should be removed

muXxer (changes requested):

crates/iota-core/src/transaction_driver/effects_certifier.rs
```diff
@@ -0,0 +1,762 @@
...
+        // failure and gather error messages to explain the rejection.
+        let certified_digest = acknowledgments_result?;
+
+        // Retry until there is a valid full effects that matches the certified digest,
+        // or all targets have been attempted.
+        loop {
+            let display_name = authority_aggregator.get_display_name(&current_target);
+            match full_effects_result {
+                Ok((effects_digest, executed_data)) => {
+                    if effects_digest != certified_digest {
```
bingyanglin, line 168: `effects_digest` here is what the validator reports about itself, we never recompute it from the effects body we got back. So a byzantine validator can return the correct certified digest but a forged `details.effects`, and we'd hand back the forged effects as finalized. The 2f+1 quorum only certifies the digest; the body still comes from a single validator, so it needs to be checked against it (this is what `check_signed_effects_plain` does on the V1 path). Maybe recompute and compare instead of trusting the reported field?

Same gap on the fast path at line 92, but both flow through this check so fixing it here covers both. Worth verifying events too: `effects.events_digest()` vs `executed_data.events`.

piotrm50: This is already tracked in an issue https://github.com/iotaledger/iota/issues/11608 and will be resolved in a subsequent PR, on top of https://github.com/iotaledger/iota/pull/11235

dev-tools/iota-private-network/README.md
piotrm50: removed

dev-tools/iota-private-network/docker-compose.yaml
piotrm50: removed

tomxey (approved):

crates/iota-core/src/transaction_orchestrator.rs
```diff
@@ -556,6 +719,59 @@ where
...
+/// backward-compatible error reporting.
+fn map_td_error_to_qd(e: TransactionDriverError) -> QuorumDriverError {
+    match e {
+        TransactionDriverError::ValidationFailed { error } => {
+            QuorumDriverError::InvalidUserSignature(IotaError::InvalidSignature { error })
+        }
+        TransactionDriverError::TimeoutWithLastRetriableError { .. } => {
+            QuorumDriverError::TimeoutBeforeFinality
+        }
+        other => {
```
bingyanglin: `RejectedByValidators` falls into this. Two issues:
- Clients treat internal errors as transient and retry, which just resends the invalid tx and adds load, and it buries real internal errors in the metrics. 
- Also It's not even hiding anything for safety, `other.to_string()` already dumps the full aggregated validator errors into the message. 
Maybe split it: map this (and `e.categorize() == InvalidTransaction/LockConflict`) to a non-retriable client error, and keep the client-facing message short instead of echoing the validator errors?

bingyanglin (approved):
Only a tiny thing that might be good to address.

crates/iota-core/src/transaction_orchestrator.rs
piotrm50: updated as suggested. thanks!

cyberphysic4l (approved):

polinikita (approved):

miker83z (approved):

muXxer (approved):

PR #11248, opened 2026-04-16
feat(dapp-kit): add useWalletStoreSync

panteleymonchuk:
**Problem:**
- Zustand's persist middleware has a selector propagation delay during hydration. When a persisted wallet exists in
  localStorage, standard selectors return stale values briefly before hydration completes. This causes ConnectionGuard to incorrectly think no wallet is persisted and prematurely redirect users from protected routes, even though auto-connect is about to restore their session.

**Solution:**
- useWalletStoreSync uses React's useSyncExternalStore to read directly from the Zustand store, bypassing the selector
  propagation delay. This provides immediate access to persisted state during hydration.

**Security:**
- Only pre-approved selectors are exposed (lastConnectedWalletName, connectionStatus) to maintain a controlled public API surface. The selectors themselves control what data is returned and in what format.

### DIFF
--- a/sdk/dapp-kit/src/hooks/wallet/useWalletStoreSync.ts
+++ b/sdk/dapp-kit/src/hooks/wallet/useWalletStoreSync.ts
@@ -1,0 +1,46 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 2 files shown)


panteleymonchuk:
Closed in a favor of: https://github.com/iotaledger/ts-packages/pull/77

PR #11265, opened 2026-04-20
feat(docs): Add claude command and update AGENTS.md

Dr-Electron:
# Description of change

This claude commands and the AGENTS file should mention everything claude needs to create docs, understand how the work and so on. It can technically also be used by a human to understand our docs 😆

### DIFF
--- a/docs/.claude/commands/new-page.md
+++ b/docs/.claude/commands/new-page.md
@@ -1,0 +1,27 @@
+Create a new documentation page following the project conventions in CLAUDE.md.
+
+## Steps
+
+1. Ask the user for:
+   - **Diátaxis type**: `tutorial`, `how-to`, `reference`, or `explanation`
+   - **Section**: `developer`, `operator`, `users`, or `about-iota`
+   - **Topic/subdirectory** (e.g. `move`, `isc`, `identity`)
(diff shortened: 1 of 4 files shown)


docs/.claude/commands/new-page.md
```diff
@@ -0,0 +1,27 @@
+Create a new documentation page following the project conventions in CLAUDE.md.
+
+## Steps
+
+1. Ask the user for:
+   - **Diátaxis type**: `tutorial`, `how-to`, `reference`, or `explanation`
```
richardmands, line 6: Does this need the accent on 'Diataxis'?

richardmands (approved):

docs/.claude/commands/new-page.md
Dr-Electron: `Diátaxis` is how it is written on their website.
https://diataxis.fr/ so I guess so?

Ginowine (approved):

docs/.claude/commands/new-page.md
richardmands: We don't use that kind of thing in English usually. Could it affect people if they search for the term? Not a huge issue.

Issue #11009, opened 2026-03-30
starfish: Implement optimistic commit rule in Starfish


piotrm50:
TODO

PR #11267, opened 2026-04-20
feat(consensus): require strong-vote quorum for block creation

### LINKED ISSUE #11135: starfish-speed: Require strong-vote quorum for block creation readiness
## Parent
Part of #11009

## Goal
In StarfishSpeed, block creation requires a strong-vote quorum (2f+1 blocks with `is_strong_vote() == true`) in addition to the regular vote quorum.

### Modified readiness check
Add a `relaxed: bool` parameter to the block readiness function:
- When `consensus_starfish_speed()` AND `!relaxed`:
  - Count blocks at the voting round (r+1) with `is_strong_vote() == true`
  - Require 2f+1 stake from strong-vote blocks
  - If not reached, return `false` (block creation blocked)
- When `relaxed == true`: skip strong-vote quorum entirely (used by soft timeout, Step 6)

### Key logic
```rust
let count_strong_votes = !relaxed
    && context.protocol_config.consensus_starfish_speed();

// ... existing vote counting ...
if votes.is_quorum(committee) {
    if count_strong_votes && !strong_votes.is_quorum(committee) {
        return false;  // Block until strong-vote quorum
    }
}
```

### Callers
- Normal block creation → `relaxed = false`
- Soft timeout (Step 6) → `relaxed = true`

## Files
- `core/src/dag_state.rs` or `core/src/core.rs` (wherever readiness is evaluated)

## Source reference
`starfish/crates/starfish-core/src/dag_state.rs` lines 1545–1645

## Depends on
Step 4 (strong vote computation)

### LINKED ISSUE #11136: starfish-speed: Add soft block timeout for relaxed block creation
## Parent
Part of #11009

## Goal
After a configurable timeout (default 200ms), create a block even without strong-vote quorum. Prevents liveness stalls when data is slow to propagate.

### Configuration
```rust
pub soft_block_timeout: Duration,  // default: 200ms
```

### New block creation reason
Add `RelaxedTimeout` variant to `ReasonToCreateBlock`.

### Background task: soft_block_timeout_task()
```rust
async fn soft_block_timeout_task(...) {
    let mut armed_round = dag_state.threshold_clock_round();
    loop {
        // Wait for threshold clock to advance
        while dag_state.threshold_clock_round() <= armed_round {
            select! { _ = proposal_round_notify.notified() => {}, _ = stopped => return }
        }
        armed_round = dag_state.threshold_clock_round();

        // Arm timer
        select! {
            _ = sleep(soft_block_timeout) => {
                // Timeout: relaxed block creation
                syncer.try_new_block_relaxed(armed_round).await;
            }
            _ = proposal_round_notify.notified() => {
                // Round advanced — restart timer
            }
            _ = stopped => return,
        }
    }
}
```

### try_new_block_relaxed()
- Checks threshold clock round matches
- Checks last proposed < threshold round
- Calls `ready_new_block_relaxed()` (bypasses strong-vote quorum)
- Creates block with `ReasonToCreateBlock::RelaxedTimeout`

### Spawn condition
Only spawn when `consensus_starfish_speed()` is enabled.

## Files
- Parameters/config — add `soft_block_timeout`
- Authority node / networking layer — spawn task
- `core/src/core.rs` or syncer — `try_new_block_relaxed()`

## Source reference
`starfish/crates/starfish-core/src/net_sync.rs` lines 1950–1981
`starfish/crates/starfish-core/src/syncer.rs` lines 267–285

## Depends on
Step 5 (readiness check with relaxed parameter)

VorobyevIlya:
# Description of change

Gate block creation on a 2f+1 strong-vote quorum when `consensus_starfish_speed` is enabled, with a soft-leader-timeout fallback so the chain continues to make progress when strong votes are slow to arrive.

- **Quorum gate** (`core.rs`): `Core::has_strong_vote_quorum(round)` counts 2f+1 stake from blocks at `round + 1` with `is_strong_vote() == true`. `try_new_block` returns `None` when the flag is on and the quorum is not yet reached (forced reasons already bypass the readiness check upstream).
- **Soft leader timeout** (`leader_timeout.rs`, `parameters.rs`): new `ReasonToCreateBlock::SoftTimeout` variant and a timer (default `soft_leader_timeout = 100ms`, paired with a hard `leader_timeout = 200ms`) that dispatches it once per round. Active only when `consensus_starfish_speed` is enabled; resets on new round and on new own block.
- **Round-scoped skip** (`core.rs`): `Core::strong_vote_timed_out_round: Option<Round>` records the clock round on arrival of the `SoftTimeout` reason. Any subsequent block-creation attempt at the same round — `AddBlock`, `AddBlockHeader`, `MinBlockDelayTimeout`, etc. — then skips the strong-vote check regardless of its reason. This fixes the reason-scoped bypass that left non-soft attempts stuck behind the quorum gate even after the timer had fired.
- **Cleanup**: extracted `dispatch_new_block` helper in `LeaderTimeoutTask` so the three timeout arms share one response-handling path.

Part of #11009
Fixes #11135
Fixes #11136


crates/starfish/core/src/core.rs
```diff
@@ -1312,6 +1337,21 @@ impl Core {
...
 
+    /// Checks whether a 2f+1 quorum of blocks at the given round have
+    /// `is_strong_vote() == true`, indicating data availability for the
+    /// leader's acknowledged blocks.
+    fn has_strong_vote_quorum(&self, round: Round) -> bool {
+        let dag_state = self.dag_state.read();
+        let blocks = dag_state.get_last_cached_block_header_per_authority(round + 1);
+        let mut strong_votes = StakeAggregator::<QuorumThreshold>::new();
+        for (block, _equivocating) in &blocks {
+            if block.round() == round && block.is_strong_vote() {
```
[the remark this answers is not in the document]
VorobyevIlya: logic is correct, copilot is wrong. Will add some comments and better names

crates/starfish/core/src/core.rs
```diff
@@ -1312,6 +1337,22 @@ impl Core {
         Some(missing)
     }
 
+    /// Returns true when 2f+1 stake of blocks at `voting_round` have
+    /// `is_strong_vote() == true`. A block at round R with `is_strong_vote`
+    /// certifies the leader at R-1, so a quorum at `voting_round` certifies
+    /// the leader at `voting_round - 1`.
+    fn has_strong_vote_quorum(&self, voting_round: Round) -> bool {
```
polinikita, line 1360: The readiness check should include two conditions:
1) a quorum of strong votes for the leader two rounds before
2) our current block is a strong vote for the leader in the previous round

polinikita (changes requested):

crates/starfish/core/src/core.rs
VorobyevIlya: done + some refactoring

crates/starfish/core/src/core.rs
```diff
@@ -1309,23 +1350,33 @@ impl Core {
...
 
-        true
+    /// Returns the leader block header for `round` if it is present in the
+    /// DAG. Starfish has exactly one leader per round, so this is either
+    /// `Some(leader_block)` or `None`.
+    fn leader_header(&self, round: Round) -> Option<VerifiedBlockHeader> {
+        let leader = self.leaders(round).into_iter().next()?;
+        self.dag_state
+            .read()
+            .get_cached_block_header_at_slot(Slot::new(round, leader.authority))
```
polinikita, line 1379: is this same as included by ancestors?

VorobyevIlya: yes

polinikita: I would consider later a refactoring that will store a structure in dag state and compute votes incrementally  after adding blocks, which will be gated everywhere (e.g. block creation)

polinikita (approved):

PR #11276, opened 2026-04-21
feat(core): Enable implicit accounts MoveAuthenticator synthetic signature  

### LINKED ISSUE #11141: [AA] Enable iota-core authority path for implicit accounts
An EOA could be implicitly treated as IOTAccount, assuming that an object exists on chain representing that account. This would allow an easy key rotation. The authority must check that an account is already present in the ledger or create "implicit" authenticators references.

miker83z:
# Description of change

Builds on #12006: 
- That made an **implicit account**, i.e., a plain (non-`MoveAuthenticator`) signer with no on-chain object, read deterministically as a synthetic read-only `SmartAccount` object at `OBJECT_START_VERSION`. 
- This PR makes the authority authenticate that implicit account through the same built-in Move-authenticator path used for explicit (claimed) accounts, instead of a separate legacy signature-verification path.

### Details

For each implicit account referenced by a transaction, the authority now synthesizes a `MoveAuthenticator` on the fly:
- It matches the implicit account id to its originating plain signature and derives the built-in authenticator data from the signature itself.
- It builds the synthetic authenticator: `call_args[0]` carries the BCS-serialized original signature, and the auth object is the read-only synthetic `SmartAccount` object produced by consistent with the version pin, so no on-chain state is created or mutated.
- Explicit and implicit authenticators are unified behind a single `MoveAuthenticatorInputs` type (carrying the original signature for implicit ones), so both run through the same gas-metered built-in-authenticator execution and dispatch.

Closes #11141

New/extended coverage:
- `crates/iota-types/.../builtin_authenticator_functions_tests.rs` — deriving the authenticator
  data from each signature type (Ed25519/Secp256k1/Secp256r1/MultiSig/Passkey) and rejecting the
  unsupported ones.
- `crates/iota-e2e-tests/tests/implicit_account_tests.rs` — end-to-end implicit authentication
  across all `GenericSignature` types and the implicit→explicit claim transition.

### Release Notes

- [x] Protocol: Implicit accounts (plain-signed senders with no on-chain object) are authenticated
  via a synthetic built-in `MoveAuthenticator` derived from the signature, gated by the
  `enable_implicit_accounts` feature flag.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-types/src/transaction.rs
```diff
@@ -2589,6 +2589,24 @@ impl SenderSignedData {
         Ok(input_objects_set.into_iter().collect::<Vec<_>>())
     }
 
+    pub fn built_in_account_objects(&self) -> IotaResult<Vec<ObjectID>> {
```
Dkwcs: Could you clarify the invariant for the key-rotation case?
As I see:
`built_in_account_objects()` derives the account object ID from the address of the current signing key. If an explicit account has rotated its public key (e.g. claimed at `addr_A`, then rotated to `key_B` with address `addr_B`), a plain `GenericSignature` from `key_B` will resolve to `ObjectID::from(addr_B)` - a **new** object - rather than to the existing explicit account at `addr_A`.
I assume test `test_explicit_rotated_pk_old_key_fails` is about this

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -291,4 +304,30 @@ impl TransactionInputLoader {
         }
         Ok(receiving_results.into())
     }
+
+    fn read_built_in_account_objects(
```
Dkwcs: `read_built_in_account_objects` is called identically at signing time and at execution time via `try_get_object`. 
If the account object does not exist during signing as `Implicit`, but is created by concurrent certified transaction before this transaction executes as `Explicit`, the executor will load on-chain authenticator fields instead of deriving them from the signature.
So if the account was claimed and key-rotated in that window, a transaction that was valid at signing
time will fail at execution time. Is this a known, accepted behavior?

crates/iota-types/src/transaction.rs
```diff
@@ -2589,6 +2589,24 @@ impl SenderSignedData {
...
+        self.tx_signatures()
+            .iter()
+            .filter_map(|sig| match sig {
+                GenericSignature::MoveAuthenticator(_) => None,
+                _ => Some(
+                    sig.try_into()
+                        .map(|signer: IotaAddress| ObjectID::from(signer)),
+                ),
+            })
+            .collect::<IotaResult<Vec<_>>>()
```
Dkwcs, line 2212: `tx_signatures()` for a sponsored transaction has two entries. If sender and sponsor are the same key pair (same address), this method returns the same `ObjectID` twice. Will it cause that the same account is authenticated twice in one transaction?

crates/iota-core/src/authority.rs
```diff
@@ -5367,196 +5515,316 @@ impl AuthorityState {
...
+            .map(|account_object| {
+                let sig = tx_signatures
+                    .iter()
+                    .filter(|sig| !matches!(sig, GenericSignature::MoveAuthenticator(_)))
+                    .find(|sig| {
+                        IotaAddress::try_from(*sig)
+                            .map(|addr| ObjectID::from(addr) == account_object.id())
+                            .unwrap_or(false)
+                    })
+                    .expect("account objects were read from these signatures");
```
Dkwcs: Would it be better to handle this error like:
```
    .ok_or_else(|| IotaError::Unknown(format!(
        "no matching signature for built-in account object {:?}", account_object.id()
    )))?
```
If that error shouldn’t crash the validator, handled error is better

crates/iota-core/src/transaction_input_loader.rs
```diff
@@ -291,4 +304,30 @@ impl TransactionInputLoader {
...
+        &self,
+        built_in_account_objects: &[ObjectID],
+    ) -> IotaResult<BuiltInAccountObjects> {
+        let mut built_in_account_object_results =
+            Vec::with_capacity(built_in_account_objects.len());
+        for &object_id in built_in_account_objects.iter() {
+            match self.cache.try_get_object(&object_id)? {
```
bingyanglin: For an explicit account that's a shared object, we read it here with `try_get_object` (latest version) instead of going through consensus shared-version assignment. If that account gets its key rotated by another tx in flight, two validators executing this in parallel could read different versions, verify against different public keys, and disagree on whether the tx is authenticated, and creates a fork. Maybe require account objects to be immutable, or declare the shared one as a real consensus input so it gets a pinned version? Or this is not an issue?

miker83z: This is an issue, indeed, I was already looking into. I created a dedicated a tracking issue here https://github.com/iotaledger/iota/issues/11900 and I will first rebase and implement the new logic on top of P-COOL.

crates/iota-types/src/transaction.rs
miker83z: pls check https://github.com/iotaledger/iota/pull/12006, this issue continues there

crates/iota-core/src/transaction_input_loader.rs
miker83z: pls check https://github.com/iotaledger/iota/pull/12006, this issue continues there

crates/iota-types/src/transaction.rs
miker83z: The fact that you can't have sponsor and signer with the same address is checked before in the tx-checks

crates/iota-core/src/transaction_input_loader.rs
miker83z: pls check https://github.com/iotaledger/iota/pull/12006, this issue continues there

Issue #10009, opened 2026-02-02
gRPC: complete and clean up gRPC e2e testing


bingyanglin:
- The `stream_checkpoints` e2e testing (which is already high-level ones) should be possibly moved from v0 folder to the `client` folder.
- The `get_checkpoints` e2e testing in v0 (which is already using the high-level client) should also be moved to the `client` folder.
- The response header tests (`v0/ledger_service/header.rs`, `v0/transaction_execution_service/header.rs`) verify that every endpoint returns correct `x-iota-*` headers. We need to decide whether to migrate these to `client/` (using `ResponseExt`) or keep them as low-level integration tests.
- The high-level `simulate_transaction` is missing `estimate_gas_budget` support — the v0 simulate tests cover this but the `client/` tests don't. We need to add this parameter to the high-level client before we can remove the v0 simulate tests.
- The v0 simulate tests also cover `command_results` read_mask scenarios with programmable transactions (SplitCoins). There is no `client/` equivalent for this yet.
- Once the high-level Client has full coverage (field masks on `checkpoints`, `get_service_info`, `get_epoch`, `estimate_gas_budget` for simulate, `command_results` coverage, and header testing), the existing `crates/iota-e2e-tests/tests/grpc/v0/` tests can be removed, because we can just go through the high-level client instead of raw tonic service clients.

muXxer:
Merged with #11278

PR #11278, opened 2026-04-22
chore(gRPC): complete and clean up gRPC e2e testing

### LINKED ISSUE #10007: gRPC: reorganize e2e testing for consistency and add `read_mask` coverage
- The `get_checkpoint_by_digest` should also have an e2e testing, possibly inside the `get_checkpoint_scenarios`.
- The checkpoint tests in `crates/iota-e2e-tests/tests/grpc/v0/ledger_service/get_checkpoints.rs` currently use high-level client API with `#[tokio::test]`, which is inconsistent with `get_objects.rs` and `get_transactions.rs` that use low-level tonic API with `#[sim_test]`. Move the high-level tests to `crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs` and replace with proper low-level tonic tests.
- Create low-level checkpoint tests in `v0/` using `LedgerServiceClient` directly for `read_mask` scenario coverage, following the same pattern as `get_objects.rs` and `get_transactions.rs`.
- Low-level tests (`grpc/v0/`) should test server behavior including `read_mask` scenarios and field presence validation. High-level tests (`grpc/client/`) should test client ergonomics with default masks.

bingyanglin:
# Description of change

Reorganizes the gRPC checkpoint e2e tests to match the layout used by `get_objects.rs` / `get_transactions.rs`, and fills in `read_mask` coverage. Also clean up the gRPC e2e testing.

fixes #10007 #10009 

### DIFF
--- a/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
+++ b/crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
@@ -1,7 +1,11 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::time::Duration;
+
+use futures::StreamExt;
 use iota_macros::sim_test;
+use tokio::time::timeout;
 
 use super::{super::utils::setup_grpc_test, common::assert_grpc_not_found};
 
@@ -41,6 +45,23 @@ async fn get_checkpoint_scenarios() {
         "Checkpoint sequence number should match requested"
     );
 
+    // Test: get checkpoint by digest round-trips via the default mask
+    let genesis_digest = genesis
+        .body()
+        .summary()
+        .expect("genesis should have summary")
+        .digest()
+        .expect("genesis summary should have a digest");
+    let by_digest = client
+        .get_checkpoint_by_digest(genesis_digest, None, None, None)
+        .await
+        .expect("Failed to get checkpoint by digest");
+    assert_eq!(
+        by_digest.body().sequence_number(),
+        0,
+        "by-digest lookup should return the genesis checkpoint"
+    );
+
     // Test: nonexistent checkpoint returns not-found error
     let result = client
         .get_checkpoint_by_sequence_number(999_999_999, None, None, None)
@@ -53,4 +74,26 @@ async fn get_checkpoint_scenarios() {
         .get_checkpoint_by_sequence_number(future_sequence, None, None, None)
         .await;
     assert_grpc_not_found(result);
+}
+
+#[sim_test]
+async fn stream_checkpoints_live() {
+    // Live-streaming sanity check: open the stream before checkpoint 2 exists
+    // and verify the broadcaster delivers it within the 120s window. This is
+    // the one test that exercises the live (non-historical) streaming path —
+    // every other checkpoint test either replays historical checkpoints or
+    // pre-waits for the range to be produced.
+    let (_test_cluster, client) = setup_grpc_test(None, None).await;
+
+    let mut stream = client
+        .stream_checkpoints(Some(2), Some(2), None, None, None)
+        .await
+        .expect("Failed to open checkpoint stream");
+
+    let first = timeout(Duration::from_secs(120), stream.body_mut().next())
+        .await
+        .expect("waiting for live checkpoint timed out")
+        .expect("stream ended without a checkpoint")
+        .expect("stream error");
+    assert_eq!(first.sequence_number, 2, "expected checkpoint 2");
 }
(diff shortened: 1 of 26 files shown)


crates/iota-grpc-client/src/client.rs
```diff
@@ -193,3 +193,31 @@ impl_grpc_client_config!(
     StateServiceClient<InterceptedChannel>,
     MovePackageServiceClient<InterceptedChannel>,
 );
+
+#[cfg(test)]
+mod tests {
+    // E2E tests always use `http://`, so the HTTPS-without-TLS branch is only
+    // reachable from a unit test compiled with `tls-ring` disabled.
+    #[cfg(not(feature = "tls-ring"))]
```
semenov-vladyslav, line 203: Will this test run in CI? The "tls-ring" feature is enabled by default. `rust_tests.py` script has a possibility to enable certain features (`feature_set` argument to `run_cargo_nextest` function) but it is not used.

bingyanglin: Good catch. I added a no-default-features step in `_rust_tests.yml`so it runs now. I also added two small fixes in `Cargo.toml` files for `iota-grpc-types` and `iota-grpc-client`. They need the serde feature from iota-sdk-types (and hash for the client). Previously they got these features by accident from other crates in the workspace. Now they ask for them directly. 
Fixed in [c3ef808](https://github.com/iotaledger/iota/pull/11278/commits/c3ef808977187b0ce006bf7a04256d8db4eecbda).

semenov-vladyslav: Not sure `_rust_tests.yml` is the right place. Why not unify it with `rust_tests.sh` script since it can detect changed crates?

bingyanglin: `rust_tests.py`'s feature_set only adds features, no `--no-default-features` support yet. And `check-unused-deps` in this same file already runs `--no-default-features` directly.
Also, changed-crate detection skips the check when `iota-grpc-client`'s own files didn't change. But the bug this catches is a silent feature leak from other crates in the workspace. So I want this to run on every PR (instead of only changed crates), so I guess adding it in `_rust_tests.yml ` is correct?

crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
```diff
@@ -54,3 +75,25 @@ async fn get_checkpoint_scenarios() {
...
+        .stream_checkpoints(Some(2), Some(2), None, None, None)
+        .await
+        .expect("Failed to open checkpoint stream");
+
+    let first = timeout(Duration::from_secs(120), stream.body_mut().next())
+        .await
+        .expect("waiting for live checkpoint timed out")
+        .expect("stream ended without a checkpoint")
+        .expect("stream error");
+    assert_eq!(first.sequence_number, 2, "expected checkpoint 2");
```
muXxer: That sounds flaky. How do we know the checkpoint doesn't exist yet and we are really testing the live streaming? Shouldn't we get the current checkpoint number from the cluster, and then wait for the "current + 5" or something?

crates/iota-grpc-client/src/client.rs
```diff
@@ -193,3 +193,31 @@ impl_grpc_client_config!(
     StateServiceClient<InterceptedChannel>,
     MovePackageServiceClient<InterceptedChannel>,
 );
+
+#[cfg(test)]
+mod tests {
+    // E2E tests always use `http://`, so the HTTPS-without-TLS branch is only
```
muXxer: will this still be the case after this PR https://github.com/iotaledger/iota/pull/11353 ?

muXxer (changes requested):

crates/iota-grpc-client/src/client.rs
semenov-vladyslav: Ok, got it.

crates/iota-e2e-tests/tests/grpc/client/checkpoints.rs
bingyanglin: Thanks for pointing this out. Fixed by querying the cluster's latest and targeting latest + 5 in [1301915](https://github.com/iotaledger/iota/pull/11278/commits/13019155850fc32da532f215347c330fccb5cc4e). This guarantees start > latest when the stream opens.

crates/iota-grpc-client/src/client.rs
bingyanglin: I think it still holds.

PR #11353 only changes the validator gRPC server in iota-network-stack. It rewrites the multiaddr from `/http` to `/https` when TLS is enabled on ServerBuilder. That code path is not used by the public fullnode gRPC. I modified the comment in [1301915](https://github.com/iotaledger/iota/pull/11278/commits/13019155850fc32da532f215347c330fccb5cc4e) to scope it to this crate, so it should be clearer now.

muXxer (approved):

semenov-vladyslav (approved):

PR #11292, opened 2026-04-22
fix(json-rpc,open-rpc): eliminate name collision for versions

### LINKED ISSUE #11291: bug(open-rpc): mismatch between wire and spec types for versions
## Description

As reported by @daughterofmars

> I found a few mismatches between the openrpc spec and the real data format.
> 1. `Coin::version, ObjectData::version` are strings but the spec says they should be numbers
> 2. `ObjectResponseError::Deleted::version, DynamicFieldInfo::version, RawData::version` are numbers but the spec says they should be strings

Upon further investigation we found that a similar mismatch exists for `UpgradeInfo::upgraded_version`: it is a number through the wire, a string in the generated OpenRPC spec (`crates/iota-open-rpc/spec/openrpc.json`)

## Root cause

Two distinct Rust types derive `JsonSchema` and both are named as `"SequenceNumber"`:

- `iota_types::base_types::SequenceNumber(u64)` that is a plain wrapper and should correspond to an integer schema.
- `iota_types::iota_serde::SequenceNumber` that is a serde helper aliased as `AsSequenceNumber` in imports, and should be represented by a string (through `BigInt_for_uint64` in the spec).

The names collide and to resolve the ambiguity `schemars` maps one type to `SequenceNumber` and the other to `SequenceNumber2` non deterministically.

Our spec builder (`crates/iota-open-rpc/src/lib.rs`) however uses a separate generator per RPC module and then merges each module's schemas into the final spec. This implies the following:

1. Inside one generator, schemars appends `2` to the second type it encounters with a colliding name. Which type wins the unsuffixed `"SequenceNumber"` slot depends the type the generator visits first).
2. Different modules make different choices. Any module that first reaches a field using `base_types::SequenceNumber` directly (e.g. `IotaObjectResponseError::VersionNotFound(ObjectID, SequenceNumber)` in `iota_object.rs`) assigns the integer variant to `"SequenceNumber"`. Modules that only see `AsSequenceNumber`-wrapped fields assign the string variant to `"SequenceNumber"`.
3. During merge the last only the mapping of the last module persists to resolve the types for the two variants `SequenceNumber` and `SequenceNumber2`, which might not be consistent for all modules.

## Impact

- Any clients generated from the spec receive misleading type information for any field whose ref landed on the wrong variant. But since this would lead to a broken client eitherway, we assume that there is no actual impact to this respect.
- IOTA TS SDK is unaffected because the generator has a hardcoded `format: uint64` → `string` override.

## Proposed fix

As a solution, it is propose to give the serde helper a distinct `schema_name`, removing the collision:

```rust
#[schemars(rename = "SequenceNumberBigInt")]
pub struct SequenceNumber(#[schemars(with = "BigInt<u64>")] u64);
```

kodemartin:
# Description of change

This patch fixes a nasty bug that caused the mismatch between the wire type of versions and the type specified in the OpenRPC spec.

Fixes #11291 

Regenerated the openrpc schema.

### Release Notes

- [x] JSON-RPC: :warning: Fix a type mismatch between the wire representation and the JSON-RPC spec. The patch renames former `SequenceNumber2` type into `SequenceNumberBigInt` that is represented as a string. Any clients generated by the spec would be broken at this point, so this should not have an impact on existing downstream applications.

### DIFF
--- a/crates/iota-open-rpc/spec/openrpc.json
+++ b/crates/iota-open-rpc/spec/openrpc.json
@@ -6398,7 +6398,7 @@
             "$ref": "#/components/schemas/TransactionDigest"
           },
           "version": {
-            "$ref": "#/components/schemas/SequenceNumber"
+            "$ref": "#/components/schemas/SequenceNumberBigInt"
           }
         }
       },
@@ -6928,7 +6928,7 @@
             "$ref": "#/components/schemas/DynamicFieldType"
           },
           "version": {
-            "$ref": "#/components/schemas/SequenceNumber2"
+            "$ref": "#/components/schemas/SequenceNumber"
           }
         }
       },
@@ -7783,7 +7783,7 @@
             "description": "the version of the queried object.",
             "allOf": [
               {
-                "$ref": "#/components/schemas/SequenceNumber2"
+                "$ref": "#/components/schemas/SequenceNumberBigInt"
               }
             ]
           }
@@ -7836,7 +7836,7 @@
                     "$ref": "#/components/schemas/ObjectID"
                   },
                   "initial_shared_version": {
-                    "$ref": "#/components/schemas/SequenceNumber2"
+                    "$ref": "#/components/schemas/SequenceNumberBigInt"
                   },
                   "mutable": {
                     "default": true,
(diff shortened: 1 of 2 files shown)


sergiupopescu199 (approved):
lgtm 🎸

KeitoTadashi (approved):

PR #11296, opened 2026-04-22
feat(consensus): classify committed leaders by StarfishSpeed metastate

### LINKED ISSUE #11137: starfish-speed: Introduce CommitMetastate enum and determine_metastate()
## Parent
Part of #11009

## Goal
Introduce the `CommitMetastate` enum (Opt/Std/Pending) and compute it during leader commit decisions.

### Type definition
```rust
#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
pub enum CommitMetastate {
    /// Optimistic: strong-vote quorum at r+1 AND StrongQC quorum at r+2.
    /// Sequence HistDA(L) + leader acknowledgment blocks.
    Opt,
    /// Standard: strong-blame quorum observed at voting round.
    /// Sequence HistDA(L) only.
    Std,
    /// Pending: neither condition met. Block sequencing until resolved.
    Pending,
}
```

### LeaderStatus extension
```rust
pub(crate) enum LeaderStatus {
    Commit(VerifiedBlockHeader, Option<CommitMetastate>),  // Option is Some for StarfishSpeed
    Skip(Slot),
    Undecided(Slot),
}
```

### determine_metastate() in BaseCommitter
```rust
fn determine_metastate(&self, leader_block, voting_round, voter_info) -> Option<CommitMetastate> {
    if !consensus_starfish_speed() { return None; }

    // 1. Strong-blame quorum at r+1: 2f+1 voters with is_strong_blame() == true → Std
    // 2. StrongQC quorum at r+2: 2f+1 certifiers carrying StrongQC → Opt
    // 3. Otherwise → Pending
}
```

### Integration
- Called from `try_direct_decide()` after confirming 2f+1 support
- Metastate flows through `LeaderStatus::Commit` → `UniversalCommitter` → `CommitObserver` → `Linearizer`

## Files
- `core/src/consensus/mod.rs` or new file — `CommitMetastate` definition
- `core/src/base_committer.rs` — `determine_metastate()` + modify `try_direct_decide()`
- `core/src/universal_committer.rs` — propagate metastate
- `core/src/commit.rs` — extend `LeaderStatus`

## Source reference
`starfish/crates/starfish-core/src/consensus/mod.rs` lines 40–62
`starfish/crates/starfish-core/src/consensus/base_committer.rs` lines 348–396

## Depends on
Step 4 (strong vote computation — blocks must carry strong_vote for metastate determination)

VorobyevIlya:
# Description of change

Adds `CommitMetastate` classification to StarfishSpeed-committed leaders and propagates it through `LeaderStatus::Commit` and `DecidedLeader::Commit`. Three variants — `Optimistic` / `Standard` / `Pending`; metastate is `None` when `consensus_starfish_speed` is off.

## Classification rules

**Direct rule** (when 2f+1 QCs are observed locally for the slot):
- `Optimistic` — 2f+1 StrongQCs at `r+2`, each backed by 2f+1 strong-vote ancestors at `r+1`.
- `Standard` — 2f+1 strong-blame quorum at `r+1`.
- Otherwise `Pending`.

**Indirect rule** (when `decide_leader_from_anchor` resolves a previously Pending / Undecided slot via a later-committed anchor): restricted to the anchor's causal history and based on the *existence* of a StrongQC in that path — present → `Optimistic`, else `Standard`. `Skip` is unreachable once the slot has a certificate in the anchor's path (by quorum intersection).

Both rules apply an `is_vote(_, leader_block) ∧ is_strong_vote()` conjunction so a Byzantine equivocating leader cannot have its strong-vote flag count for the wrong leader block at the same slot.

## Changes

- `commit.rs`: `LeaderStatus::Commit` and `DecidedLeader::Commit` carry `Option<CommitMetastate>`; `Display` prints `optimistic` / `standard` / `pending`.
- `base_committer.rs`: `determine_metastate_direct` for the direct path; `decide_leader_from_anchor` inlines the indirect classifier with a single ancestor scan returning `(is_certificate, is_strong_certificate)`. Flag-off keeps the cheaper `is_certificate`-only walk.
- `universal_committer.rs`: metrics labels include the metastate when present.
- Existing tests: pattern matches updated for the new `Commit(_, _)` shape.
- `base_committer_metastate_tests.rs` (new, 8 tests):
  - Direct rule: Optimistic / Standard / Pending / flag-off (4).
  - Direct under equivocation: 2 tests validating the `is_vote()` conjunction is load-bearing — without it, a strong-vote/strong-blame quorum would be wrongly crossed.
  - Indirect rule: 2 tests for the anchor's-path-only semantic — StrongQC inside path → Optimistic; outside path → Standard.

Part of #11009
Fixes #11137

### DIFF
--- a/crates/starfish/core/src/base_committer.rs
+++ b/crates/starfish/core/src/base_committer.rs
@@ -117,33 +117,27 @@ impl BaseCommitter {
             .unwrap_or(LeaderStatus::Undecided(leader))
     }
 
-    /// Apply the indirect decision rule to the specified leader to see whether
-    /// we can indirect-commit or indirect-skip it.
-    #[tracing::instrument(skip_all, fields(leader = %leader_slot))]
+    /// Apply the indirect rule; return `current` unchanged if no committed
+    /// anchor is reachable.
+    #[tracing::instrument(skip_all, fields(leader = %current))]
     pub fn try_indirect_decide<'a>(
         &self,
-        leader_slot: Slot,
+        current: LeaderStatus,
         leaders: impl Iterator<Item = &'a LeaderStatus>,
     ) -> LeaderStatus {
-        // The anchor is the first committed leader with round higher than the decision
-        // round of the target leader. We must stop the iteration upon
-        // encountering an undecided leader.
-        let anchors = leaders.filter(|x| leader_slot.round + WAVE_LENGTH <= x.round());
-
+        let slot = current.slot();
+        let anchors = leaders.filter(|x| slot.round + WAVE_LENGTH <= x.round());
         for anchor in anchors {
-            tracing::trace!(
-                "[{self}] Trying to indirect-decide {leader_slot} using anchor {anchor}",
-            );
+            tracing::trace!("[{self}] Trying to indirect-decide {slot} using anchor {anchor}",);
             match anchor {
                 LeaderStatus::Commit(anchor, _) => {
-                    return self.decide_leader_from_anchor(anchor, leader_slot);
+                    return self.decide_leader_from_anchor(anchor, slot);
                 }
                 LeaderStatus::Skip(..) => (),
-                LeaderStatus::Undecided(..) => break,
+                LeaderStatus::Undecided(..) => return current,
             }
         }
-
-        LeaderStatus::Undecided(leader_slot)
+        current
     }
 
     pub fn elect_leader(&self, round: Round) -> Option<Slot> {
--- a/crates/starfish/core/src/base_committer.rs
+++ b/crates/starfish/core/src/base_committer.rs
@@ -2,7 +2,11 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashMap, fmt::Display, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+    sync::Arc,
+};
 
 use parking_lot::RwLock;
 use starfish_config::{AuthorityIndex, Stake};
@@ -10,7 +14,7 @@ use tracing::warn;
 
 use crate::{
     block_header::{BlockHeaderAPI, BlockRef, Round, Slot, VerifiedBlockHeader},
-    commit::{LeaderStatus, WAVE_LENGTH, WaveNumber},
+    commit::{CommitMetastate, LeaderStatus, WAVE_LENGTH, WaveNumber},
     context::Context,
     dag_state::DagState,
     leader_schedule::LeaderSchedule,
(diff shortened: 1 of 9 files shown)


crates/starfish/core/src/base_committer.rs
```diff
@@ -324,7 +335,10 @@ impl BaseCommitter {
         // We commit the target leader if it has a certificate that is an ancestor of
         // the anchor. Otherwise skip it.
         match certified_leader_blocks.pop() {
-            Some(certified_leader_block) => LeaderStatus::Commit(certified_leader_block),
+            Some(certified_leader_block) => {
+                let metastate = self.determine_metastate(&certified_leader_block);
```
polinikita: Here we should check whether exists a path though a strong QC from an anchor. If yes, then it is Commit-Optimistic. If only through QCm then it is Commit-Standard. If no QC  from an anchor, then it is Skip

polinikita (changes requested):

crates/starfish/core/src/base_committer.rs
VorobyevIlya: done

polinikita (changes requested):
What is still not clear is how Pending state can be changed once it is set to Pending. I think that one needs to reevaluate the metastate of "not-final" decisions like pending on state/commit level and pending on metastate level.

crates/starfish/core/src/base_committer.rs
```diff
@@ -106,33 +117,27 @@ impl BaseCommitter {
...
-            tracing::trace!(
-                "[{self}] Trying to indirect-decide {leader_slot} using anchor {anchor}",
-            );
+            tracing::trace!("[{self}] Trying to indirect-decide {slot} using anchor {anchor}",);
             match anchor {
-                LeaderStatus::Commit(anchor) => {
-                    return self.decide_leader_from_anchor(anchor, leader_slot);
+                LeaderStatus::Commit(anchor, _) => {
+                    return self.decide_leader_from_anchor(anchor, slot);
                 }
```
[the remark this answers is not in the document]
VorobyevIlya: no, that is impossible

polinikita:
One needs to update the description of the PR  (e.g. remove the paper reference and adjust to properly reflect the current diff change)

crates/starfish/core/src/base_committer.rs
```diff
@@ -106,33 +117,27 @@ impl BaseCommitter {
...
                 LeaderStatus::Skip(..) => (),
-                LeaderStatus::Undecided(..) => break,
+                LeaderStatus::Undecided(..) => return current,
```
polinikita, line 137: We can also break here as it was before, right?

polinikita (approved):

crates/starfish/core/src/base_committer.rs
VorobyevIlya: yes

PR #11297, opened 2026-04-22
fix(graphql): resolve real tombstone versions for wrapped/deleted objects

### LINKED ISSUE #11195: [indexer]: Return correct object version for wrapped/deleted objects
### Context

When a wrapped object is later unwrapped, `objects_backward_history` records its previous wrapped state with `version = unwrap_output_version - 1` (lamport-1 approximation). This may be higher than the real tombstone version. For example:

- cp1: Foo created (v2)
- cp2: Foo wrapped (real tombstone v3), Wrapper transferred 3 times → lamport pumped
- cp3: Foo unwrapped (v7), backward history stores `WrappedOrDeleted` at v6 (lamport-1)

The real tombstone version is 3, but we store and return 6.

### What needs to be done

**1. Return the correct tombstone version in GraphQL responses**

When the backward diff query returns a `WrappedOrDeleted` object from `objects_backward_history`, the reported version is the lamport-1 approximation. The correct version can be resolved at read time by looking up the actual tombstone version from `objects_version`, which stores every `(object_id, object_version)` pair including tombstones.

For a `WrappedOrDeleted` entry with lamport-1 version N, the real tombstone version is the highest version in `objects_version` that is ≤ N for that `object_id`. This could be done as a post-processing step after the backward diff query, or as a join within the query itself.

**2. Make objects findable by real tombstone version via `objectKeys`**

Currently, `objects(filter: {objectKeys: [{objectId: "...", version: 3}]})` returns empty when version 3 is the real tombstone version, because neither `checkpointed_objects` (has the object at a different version) nor `objects_backward_history` (stores lamport-1) has a row at that exact version.

Since `objects_version` has the row `(object_id, 3)`, we know version 3 exists. A possible approach is to add a third source branch in the `build_backward_objects_query` for `Historical` view: join `objects_version` against the requested version to confirm the version exists, then produce a synthetic `WrappedOrDeleted` result for versions that are present in `objects_version` but not covered by the other two sources.

Alternatively, resolve the requested version against `objects_version` before running the backward diff query — map the real tombstone version to the lamport-1 version that `objects_backward_history` actually stores, then query with that.

tomxey:
# Description of change

Correct the version reported for `WrappedOrDeleted` objects in backward diff queries and enable `objectKeys` lookups by real tombstone version.

**Problem:** When a wrapped object is later unwrapped, `objects_backward_history` records the previous wrapped state with `version = lamport - 1`, which may be higher than the real tombstone version. For example: object wrapped at v3, unwrapped at v7 → backward history stores v6, but the real tombstone version is 3.

**Version correction:**
- SQL projections tag rows with `from_backward_history` (`TRUE`/`FALSE`) so we can identify which source produced each result
- After the backward diff query, a post-processing step resolves the real tombstone version from `objects_version` for `WrappedOrDeleted` entries from backward history
- Entries from `checkpointed_objects` already have the correct version and are skipped

**objectKeys lookup by real tombstone version:**
- For keys-only queries (no type/owner filters), `objects_version` is used as a third SQL source producing synthetic tombstone rows for `(object_id, version)` pairs not found in `checkpointed_objects` or `objects_backward_history` (`NOT EXISTS` checks)
- When type/owner filters are present (`HistoricalWithFilter`), this source is skipped since tombstones have NULL data fields and would never match
- `BackwardView` enum now has `HistoricalKeysOnly` vs `HistoricalWithFilter` variants to express this distinction
- A pruning watermark (`min_available_cp` from the `objects_backward_history` watermark entry) filters out versions in pruned ranges to prevent false tombstones

**Migration:**
- Backward history migration now initializes a watermark entry from the latest committed checkpoint, with lower bounds set to `max_cp + 1` to indicate no backward history data exists yet
- `down.sql` cleans up the watermark entry on rollback

fixes https://github.com/iotaledger/iota/issues/11195

    - `consistent_view_wrapped`: version corrected from 6 (lamport-1) to 3 (real tombstone)
    - `object_keys_tombstone`: real tombstone version (3) now findable via objectKeys, old active version (2) correctly returns INDEXED (not a false tombstone), lamport-1 version (6) returns corrected version 3
    - `objects_pagination`: objectKeys with type filter uses `HistoricalWithFilter` path (no synthetic tombstones)

### DIFF
--- a/crates/iota-graphql-rpc/src/backward_view/historical.rs
+++ b/crates/iota-graphql-rpc/src/backward_view/historical.rs
@@ -35,13 +35,14 @@ pub(crate) fn query_with_filter(
 /// Includes synthetic tombstones from `objects_version` for versions not
 /// found in the other sources, since tombstones could match.
 pub(crate) fn query_keys_only(
+    checkpoint_viewed_at: u64,
     page: &Page<Cursor>,
     filter_fn: impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
     merge_and_deduplicate_three(
         checkpointed_objects(page, &filter_fn),
         historical_objects(page, &filter_fn),
-        tombstones_from_objects_version(page, &filter_fn),
+        tombstones_from_objects_version(checkpoint_viewed_at, page, &filter_fn),
     )
 }
 
@@ -80,21 +81,41 @@ fn historical_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQ
 /// or `objects_backward_history`. This allows `objectKeys` lookups to find
 /// objects by their real tombstone version.
 ///
+/// Why surviving rows are necessarily `WrappedOrDeleted`: every version in
+/// `objects_version` falls into exactly one of three cases.
+///   1. Currently-latest state of the object (active row or tombstone) — lives
+///      in `checkpointed_objects` and is filtered out by the first `NOT
+///      EXISTS`.
+///   2. Prior active state superseded by a later transaction — recorded in
+///      `objects_backward_history` at its actual `object_version` with status
+///      `ACTIVE`, and filtered out by the second `NOT EXISTS`.
+///   3. Prior wrap/delete tombstone that was later overwritten in
+///      `checkpointed_objects` (typically by an unwrap). The wrap state is
+///      recorded in `objects_backward_history`, but at `lamport - 1` of the
+///      unwrapping transaction — which generally differs from the real
+///      tombstone version. The real version in `objects_version` therefore
+///      survives both `NOT EXISTS` filters. By construction this is the only
+///      surviving case, hence the static `WrappedOrDeleted` tag.
+///
 /// Only used for keys-only queries where the filter contains only
 /// `(object_id, object_version)` pairs — the `NOT EXISTS` subqueries hit
 /// primary keys so the cost is proportional to the number of requested keys.
 ///
-/// Filters out versions below the backward history watermark to avoid
-/// returning false tombstones for pruned ranges.
+/// Filters out versions outside the `[backward-history watermark,
+/// checkpoint_viewed_at]` window to avoid false tombstones from pruned ranges
+/// and concurrent in-flight batch writes.
 fn tombstones_from_objects_version(
+    checkpoint_viewed_at: u64,
     page: &Page<Cursor>,
     filter_fn: &impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
     let wrapped_or_deleted = NativeObjectStatus::WrappedOrDeleted as i16;
+    let checkpoint_viewed_at = checkpoint_viewed_at as i64;
 
-    // Inner query: select from objects_version, excluding versions that already
-    // exist in checkpointed_objects or objects_backward_history, and filtering
-    // out pruned ranges via the backward history watermark.
+    // Inner query: select from objects_version, bounded by the backward
+    // history watermark below and `checkpoint_viewed_at` above, and excluding
+    // versions that already exist in checkpointed_objects or
+    // objects_backward_history.
     let inner = query!(format!(
         "SELECT object_id, object_version, \
          {wrapped_or_deleted}::smallint AS object_status, \
@@ -114,6 +135,7 @@ fn tombstones_from_objects_version(
          WHERE cp_sequence_number >= COALESCE(\
              (SELECT min_available_cp FROM watermarks \
               WHERE entity = '{BACKWARD_HISTORY_WATERMARK_ENTITY}'), 0) \
+         AND cp_sequence_number <= {checkpoint_viewed_at} \
          AND NOT EXISTS (\
              SELECT 1 FROM checkpointed_objects co \
              WHERE co.object_id = ov.object_id \
--- a/crates/iota-graphql-rpc/src/backward_view/historical.rs
+++ b/crates/iota-graphql-rpc/src/backward_view/historical.rs
@@ -5,7 +5,10 @@
 //! filtering. Combines all objects from `checkpointed_objects` with all
 //! past versions from `objects_backward_history`.
 
-use super::{NOT_YET_CREATED, OBJECT_COLUMNS, merge_and_deduplicate};
+use super::{
+    BACKWARD_HISTORY_WATERMARK_ENTITY, CHECKPOINTED_COLUMNS, HISTORY_COLUMNS, NOT_YET_CREATED,
+    NativeObjectStatus, merge_and_deduplicate, merge_and_deduplicate_three,
+};
 use crate::{
     filter, query,
     raw_query::RawQuery,
@@ -15,13 +18,30 @@ use crate::{
     },
 };
 
-/// Builds a historical view by merging all objects from `checkpointed_objects`
-/// with all past versions from `objects_backward_history`, without consistency
-/// filtering. Used for exact id+version lookups (`objectKeys`).
-pub(crate) fn query(page: &Page<Cursor>, filter_fn: impl Fn(RawQuery) -> RawQuery) -> RawQuery {
+/// Builds a historical view with additional filters (type, owner, etc.).
+/// Since tombstones have NULL data fields, they can never match these
+/// filters, so the `objects_version` fallback is skipped.
+pub(crate) fn query_with_filter(
+    page: &Page<Cursor>,
+    filter_fn: impl Fn(RawQuery) -> RawQuery,
+) -> RawQuery {
     merge_and_deduplicate(
         checkpointed_objects(page, &filter_fn),
         historical_objects(page, &filter_fn),
+    )
+}
+
+/// Builds a historical view with only key filters (no type/owner).
+/// Includes synthetic tombstones from `objects_version` for versions not
+/// found in the other sources, since tombstones could match.
+pub(crate) fn query_keys_only(
+    page: &Page<Cursor>,
+    filter_fn: impl Fn(RawQuery) -> RawQuery,
+) -> RawQuery {
+    merge_and_deduplicate_three(
+        checkpointed_objects(page, &filter_fn),
+        historical_objects(page, &filter_fn),
+        tombstones_from_objects_version(page, &filter_fn),
     )
 }
 
@@ -32,8 +52,7 @@ fn checkpointed_objects(
     filter_fn: &impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
     let checkpointed_filtered = filter_fn(query!(format!(
-        "SELECT {} FROM checkpointed_objects",
-        OBJECT_COLUMNS
+        "SELECT {CHECKPOINTED_COLUMNS} FROM checkpointed_objects"
     )));
     let source = query!(
         "SELECT candidates.* FROM ({}) candidates",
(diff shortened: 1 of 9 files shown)


crates/iota-graphql-rpc/src/backward_view/historical.rs
```diff
@@ -55,3 +74,59 @@ fn historical_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQ
     let source = query!("SELECT candidates.* FROM ({}) candidates", history_window);
     page.apply::<StoredBackwardObject>(source)
 }
+
+/// Returns synthetic `WrappedOrDeleted` tombstone rows from `objects_version`
+/// for versions that exist there but are NOT present in `checkpointed_objects`
+/// or `objects_backward_history`. This allows `objectKeys` lookups to find
+/// objects by their real tombstone version.
```
kodemartin, line 80: It is hard to understand based on this doc why we know that the rows returned are indeed `WRAPPED_OR_DELETED`.

Please extend the document by adding a concise explanation why that is.

crates/iota-graphql-rpc/src/backward_view/historical.rs
```diff
@@ -55,3 +74,59 @@ fn historical_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQ
...
+         NULL::text AS object_type_name, \
+         NULL::bytea AS serialized_object, \
+         NULL::text AS coin_type, \
+         NULL::bigint AS coin_balance, \
+         NULL::smallint AS df_kind, \
+         FALSE AS from_backward_history \
+         FROM objects_version ov \
+         WHERE cp_sequence_number >= COALESCE(\
+             (SELECT min_available_cp FROM watermarks \
+              WHERE entity = '{BACKWARD_HISTORY_WATERMARK_ENTITY}'), 0) \
```
kodemartin: We probably need an upper bound here as well, i.e. `cp_sequence_number <= checkpoint_viewed_at`. 

What happens for example if `objects_version` stores a new active `(object_id, version)` before it is persisted on `checkpointed_objects`?

The upper bound would work if the caller ensures that `checkpoint_viewed_at` is `<= max_available_cp` from `backwards_history`, which I think is ensured by the available-range check.

crates/iota-graphql-rpc/src/backward_view/historical.rs
```diff
@@ -55,3 +74,59 @@ fn historical_objects(page: &Page<Cursor>, filter_fn: &impl Fn(RawQuery) -> RawQ
...
+             WHERE co.object_id = ov.object_id \
+               AND co.object_version = ov.object_version) \
+         AND NOT EXISTS (\
+             SELECT 1 FROM objects_backward_history bh \
+             WHERE bh.object_id = ov.object_id \
+               AND bh.object_version = ov.object_version)"
+    ));
+
+    // Wrap in a subquery so filter_fn can apply objectKeys WHERE clause cleanly.
+    let version_filtered = filter_fn(query!("SELECT * FROM ({}) ov_filtered", inner));
```
kodemartin: Are there guarantees that the filter is applied on `inner` and not only on the `outer` query? In the latter case we face a full scan of `objects_version` on the `cp_sequence_number` condition only.

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1117,6 +1120,13 @@ fn version_for_dynamic_fields(native: &NativeObject) -> u64 {
 }
 
 impl ObjectFilter {
+    /// Returns `true` if the filter only constrains on object ID and/or version
+    /// (no type or owner filters). Such filters are safe to apply against
+    /// tables that lack type/owner columns (e.g. `objects_version`).
```
kodemartin: returns `true` even if the filter is empty.

But as I suggested in one of the other comments, I think it is best to enforce this through a new-type instead.

crates/iota-graphql-rpc/src/backward_view/historical.rs
```diff
@@ -5,7 +5,10 @@
 //! filtering. Combines all objects from `checkpointed_objects` with all
 //! past versions from `objects_backward_history`.
 
-use super::{NOT_YET_CREATED, OBJECT_COLUMNS, merge_and_deduplicate};
+use super::{
```
kodemartin: This one must have slipped preview reviews. By convention, we use absolute imports to make them more robust against refactoring.


```rust
use crate::backward_view::{
```

crates/iota-graphql-rpc/src/backward_view/historical.rs
```diff
@@ -15,25 +18,41 @@ use crate::{
...
+pub(crate) fn query_keys_only(
+    page: &Page<Cursor>,
+    filter_fn: impl Fn(RawQuery) -> RawQuery,
+) -> RawQuery {
+    merge_and_deduplicate_three(
+        checkpointed_objects(page, &filter_fn),
+        historical_objects(page, &filter_fn),
+        tombstones_from_objects_version(page, &filter_fn),
+    )
+}
```
kodemartin, line 45: I noticed that these functions require that the filter contains some sort of object keys. This is not reflected on this module at all, and instead relies on the caller satisfying the requirement. 

In order to avoid this and benefit readability and maintenance we can restrain the underlying logic on this module based on the following:

1. Create a newtype `HistoricalFilter(ObjectFilter)`  that is guaranteed to have keys enforced through a conversion imp.
2. Use a single  `pub(crate) query` method that accepts the historical filter instead of the generic `filter_fn`, and only use `filter_fn` where absolutely necessary in private functions here.

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1397,6 +1413,83 @@ impl StoredBackwardObject {
...
+
+    let values = to_resolve
+        .iter()
+        .map(|(id, ver)| format!("('\\x{}'::bytea, {}::bigint)", hex::encode(id), ver))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    let sql = format!(
+        "SELECT pairs.object_id, pairs.backward_history_version, MAX(ov.object_version) AS real_version \
+         FROM (VALUES {values}) AS pairs(object_id, backward_history_version) \
```
kodemartin: Binding values instead of inlining should be generally preferable, as it has multiple advantages from the database perspective.

1. The [extended query message flow](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY) separates parsing from binding values and then executing, allowing caching and reusing plans based on the sql text. Inlining values as opposed to binding leads to different sql test every time.
2. Encoding values into text increases the payload, which is also something to take into account
3. Parser work grows with page size. The server lexes and parses every byte of the inlined `VALUES` list, including each `::bytea / ::bigint` cast. Bound parameter values skip the parser entirely and are  delivered directly to the executor.

We use `sql_query` eitherway below so something of this sort could work
```rust
  let (ids, versions): (Vec<Vec<u8>>, Vec<i64>) = to_resolve.into_iter().unzip();
  let sql = "SELECT pairs.object_id, pairs.backward_history_version, \
                    MAX(ov.object_version) AS real_version \
             FROM unnest($1::bytea[], $2::bigint[]) \
                  AS pairs(object_id, backward_history_version) \
             LEFT JOIN objects_version ov \
               ON ov.object_id = pairs.object_id \
              AND ov.object_version <= pairs.backward_history_version \
             GROUP BY pairs.object_id, pairs.backward_history_version";
  let rows: Vec<ResolvedVersion> = conn.results(move || {
      diesel::sql_query(sql)
          .bind::<sql_types::Array<sql_types::Binary>, _>(ids.clone())
          .bind::<sql_types::Array<sql_types::BigInt>, _>(versions.clone())
  })?;
```

Further reading:

- Plan caching and the custom-vs-generic-plan heuristic for prepared statements: https://www.postgresql.org/docs/current/sql-prepare.html#SQL-PREPARE-NOTES
- unnest multi-array form used below: https://www.postgresql.org/docs/current/functions-array.html#ARRAY-FUNCTIONS-TABLE
- High-level overview of how clients interact with the server: https://www.postgresql.org/docs/current/protocol-overview.html

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1397,6 +1413,83 @@ impl StoredBackwardObject {
...
+        })
+        .collect();
+
+    Ok(results
+        .into_iter()
+        .map(|mut r| {
+            if r.from_backward_history
+                && r.object_status == NativeObjectStatus::WrappedOrDeleted as i16
+            {
+                let key = (r.object_id.clone(), r.object_version);
```
kodemartin: Is there any way to avoid cloning `object_id` here?

crates/iota-graphql-rpc/src/backward_view/mod.rs
```diff
@@ -43,3 +54,21 @@ pub(super) fn merge_and_deduplicate(source_a: RawQuery, source_b: RawQuery) -> R
...
 }
+
+/// Like `merge_and_deduplicate` but for three sources.
+pub(super) fn merge_and_deduplicate_three(
+    source_a: RawQuery,
+    source_b: RawQuery,
+    source_c: RawQuery,
+) -> RawQuery {
+    let combined = query!(
+        r#"SELECT DISTINCT ON (object_id) * FROM (({}) UNION ALL ({}) UNION ALL ({})) candidates"#,
```
kodemartin: What happens if the  `ObjectFilter` includes multiple versions for the same id?

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1397,6 +1413,83 @@ impl StoredBackwardObject {
...
+/// higher than the actual tombstone version. This function looks up the real
+/// version from `objects_version` using a single batch query with a VALUES
+/// list joined via `MAX(object_version) <= backward_history_version`. Only
+/// entries tagged with `from_backward_history = true` are resolved; entries
+/// from `checkpointed_objects` already have the correct version.
+pub(crate) fn resolve_tombstone_versions(
+    conn: &mut crate::data::pg::PgConnection<'_>,
+    results: Vec<StoredBackwardObject>,
+) -> Result<Vec<StoredBackwardObject>, diesel::result::Error> {
+    let to_resolve: Vec<(Vec<u8>, i64)> = results
```
sergiupopescu199: This part could be optimized to avoid cloning the `object_id` and to construct the string in one operation. As final step we could check the result is its empty.

```rust
let values: String = results
    .iter()
    .filter(|r| {
        r.from_backward_history
            && r.object_status == NativeObjectStatus::WrappedOrDeleted as i16
    })
    .map(|r| {
        format!(
            "('\\x{}'::bytea, {}::bigint)",
            hex::encode(&r.object_id),
            r.object_version
        )
    })
    .collect::<Vec<String>>()
    .join(", ");

    // Check string emptiness
    if to_resolve.is_empty() {
        return Ok(results);
    }
```

crates/iota-graphql-rpc/src/types/object.rs
sergiupopescu199: The clone in the final loop exists because `HashMap<(Vec<u8>, i64), i64>::get` requires a `&(Vec<u8>, i64)` and tuples don't have a Borrow impl that lets you pass `(&[u8], i64)`. On eof the possible solutions could be by keying the outer map on `object_id` alone and putting version in an inner map.

```rust
    type ObjectID = Vec<u8>;
    type BackwardHisotryVersion = i64;
    type RealVersion = i64;

    let resolved_map: HashMap<ObjectID, HashMap<BackwardHisotryVersion, RealVersion>> = rows
        .into_iter()
        .filter_map(|r| {
            r.real_version
                .map(|real| (r.object_id, r.backward_history_version, real))
        })
        .fold(HashMap::new(), |mut acc, (id, ver, real)| {
            acc.entry(id).or_default().insert(ver, real);
            acc
        });

    Ok(results
        .into_iter()
        .map(|mut r| {
            if r.from_backward_history
                && r.object_status == NativeObjectStatus::WrappedOrDeleted as i16
            {
                if let Some(&real_version) = resolved_map
                    .get(&r.object_id)
                    .and_then(|versions| versions.get(&r.object_version))
                {
                    r.object_version = real_version;
                }
            }
            r
        })
        .collect()
    )

```

crates/iota-graphql-rpc/src/backward_view/mod.rs
```diff
@@ -43,3 +54,21 @@ pub(super) fn merge_and_deduplicate(source_a: RawQuery, source_b: RawQuery) -> R
 
     query!("SELECT * FROM ({}) candidates", combined)
 }
+
+/// Like `merge_and_deduplicate` but for three sources.
+pub(super) fn merge_and_deduplicate_three(
```
sergiupopescu199: This function is very similar in structure and logic to the `merge_and_deduplicate`, it could be possible to refactor the  `merge_and_deduplicate` to accept `&[RawQuery]`, and modify the `r#"SELECT DISTINCT ON (object_id) * FROM (({}) UNION ALL ({})) candidates"#,` accordingly.

crates/iota-graphql-rpc/src/backward_view/mod.rs
tomxey: I think we will return only one result with the latest version, but this is a preexisting behaviour (it behaved the same in original consistency.rs impl), so it's not a new bug.

crates/iota-graphql-rpc/src/backward_view/historical.rs
tomxey: please check if 4572d60983 is enough

crates/iota-graphql-rpc/src/backward_view/historical.rs
tomxey: done in 4572d60983

crates/iota-graphql-rpc/src/backward_view/historical.rs
tomxey: I think it's not guaranteed, restructured the query in 7b99bcb99b

crates/iota-graphql-rpc/src/types/object.rs
tomxey: created the new type in 9e867a411e

crates/iota-graphql-rpc/src/backward_view/historical.rs
tomxey: done in 9e867a411e

crates/iota-graphql-rpc/src/backward_view/historical.rs
tomxey: done in 1507bc16b5

crates/iota-graphql-rpc/src/types/object.rs
tomxey: done in a8dcfde784

crates/iota-graphql-rpc/src/types/object.rs
tomxey: done in 359657e2f2

crates/iota-graphql-rpc/src/types/object.rs
tomxey: handled as part of a8dcfde784

crates/iota-graphql-rpc/src/backward_view/mod.rs
tomxey: done in 782eb34c88

sergiupopescu199 (approved):
lgtm 🏎️

kodemartin (approved):
lgtm 🩰

PR #11299, opened 2026-04-23
fix(grpc): cap concurrent stream_checkpoints subscribers

### LINKED ISSUE #10540: gRPC: Add streaming subscription cap to prevent resource exhaustion
`stream_checkpoint_data` creates a new `broadcast::Receiver` for every call with no subscriber limit. The existing `broadcast_buffer_size` config controls buffer depth per subscriber, NOT subscriber count.

A single client can open unlimited streaming connections, each receiving every checkpoint message. 100 streaming clients = 100x memory for buffered messages. This is a resource exhaustion vector for public-facing servers without an API gateway.

bingyanglin:
# Description of change

The `stream_checkpoints` RPC had no limit on subscribers, so a single client could open unlimited streams and overload the server. This PR adds a configurable cap (default `1024`) enforced by a semaphore in `GrpcCheckpointDataBroadcaster`; over-cap requests are rejected with `Unavailable`. A new Prometheus gauge `node_grpc_inflight_checkpoint_stream_subscribers` tracks live subscriber count for alerting.

fixes #10540 

### Release Notes

- [x] gRPC: New limit on concurrent checkpoint streaming clients (default 1024). Tune via `grpc-api-config.max-concurrent-stream-subscribers`. Monitor via Prometheus metric `node_grpc_inflight_checkpoint_stream_subscribers`.

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -5,7 +5,7 @@
 use std::{
     collections::{BTreeMap, BTreeSet},
     net::{IpAddr, Ipv4Addr, SocketAddr},
-    num::NonZeroUsize,
+    num::{NonZeroU32, NonZeroUsize},
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
@@ -311,9 +311,9 @@ pub struct GrpcApiConfig {
     /// Maximum number of concurrent subscribers to checkpoint streaming RPCs.
     /// Once the cap is reached, additional subscribe requests are rejected
     /// with `Unavailable` to protect the server from being overwhelmed by
-    /// unbounded streaming clients.
+    /// unbounded streaming clients. Must be non-zero.
     #[serde(default = "default_grpc_api_max_concurrent_stream_subscribers")]
-    pub max_concurrent_stream_subscribers: u32,
+    pub max_concurrent_stream_subscribers: NonZeroU32,
 
     /// Maximum size for Move values when rendering to JSON
     /// in bytes.
@@ -345,8 +345,8 @@ fn default_grpc_api_broadcast_buffer_size() -> u32 {
     100
 }
 
-fn default_grpc_api_max_concurrent_stream_subscribers() -> u32 {
-    1024
+fn default_grpc_api_max_concurrent_stream_subscribers() -> NonZeroU32 {
+    NonZeroU32::new(1024).unwrap()
 }
 
 fn default_grpc_api_max_message_size_bytes() -> u32 {
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -308,6 +308,13 @@ pub struct GrpcApiConfig {
     #[serde(default = "default_grpc_api_broadcast_buffer_size")]
     pub broadcast_buffer_size: u32,
 
+    /// Maximum number of concurrent subscribers to checkpoint streaming RPCs.
+    /// Once the cap is reached, additional subscribe requests are rejected
+    /// with `Unavailable` to protect the server from being overwhelmed by
+    /// unbounded streaming clients.
+    #[serde(default = "default_grpc_api_max_concurrent_stream_subscribers")]
+    pub max_concurrent_stream_subscribers: u32,
+
     /// Maximum size for Move values when rendering to JSON
     /// in bytes.
     #[serde(default = "default_grpc_api_max_json_move_value_size")]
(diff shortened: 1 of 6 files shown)


crates/iota-grpc-server/src/ledger_service/get_checkpoint.rs
```diff
@@ -408,9 +408,17 @@ pub(crate) fn stream_checkpoints(
         return Err(Status::invalid_argument("events_filter requires events in read_mask").into());
     }
 
-    let rx = service.checkpoint_data_broadcaster.subscribe();
+    let subscription = service
+        .checkpoint_data_broadcaster
+        .subscribe()
+        .ok_or_else(|| {
+            Status::unavailable(format!(
+                "maximum concurrent stream subscribers reached ({})",
```
muXxer: Not sure if we should expose the information about how many concurrent are allowed to the clients. I would remove the value.

muXxer (changes requested):

crates/iota-config/src/node.rs
```diff
@@ -308,6 +308,13 @@ pub struct GrpcApiConfig {
     #[serde(default = "default_grpc_api_broadcast_buffer_size")]
     pub broadcast_buffer_size: u32,
 
+    /// Maximum number of concurrent subscribers to checkpoint streaming RPCs.
+    /// Once the cap is reached, additional subscribe requests are rejected
+    /// with `Unavailable` to protect the server from being overwhelmed by
+    /// unbounded streaming clients.
+    #[serde(default = "default_grpc_api_max_concurrent_stream_subscribers")]
+    pub max_concurrent_stream_subscribers: u32,
```
muXxer: What if someone passes 0? Should we add a minimum of 1? Or does 0 mean `Semaphore::MAX_PERMITS` ?

muXxer (changes requested):

crates/iota-grpc-server/src/ledger_service/get_checkpoint.rs
bingyanglin: Agreed, no need to expose. Fixed at [0f10d72](https://github.com/iotaledger/iota/pull/11299/commits/0f10d7237b066a567bea963c0878730129c23b1c).

crates/iota-config/src/node.rs
bingyanglin: Good point. Treating 0 as `MAX_PERMITS` would be easy to misusI think. I make the type to be `NonZeroU32` so the minimum is 1 natively, and serde will reject 0 at config load with a clear error. Fixed in [03535b6](https://github.com/iotaledger/iota/pull/11299/commits/03535b612d5f124a3fb5e3951fa872ae4234d480).

muXxer (approved):

crates/iota-grpc-server/src/types.rs
```diff
@@ -79,23 +83,99 @@ pub type GetCheckpointStream = Pin<Box<dyn futures::Stream<Item = CheckpointStre
...
+    /// Subscribe to checkpoint data broadcasts, enforcing the configured cap
+    /// on concurrent subscribers.
+    ///
+    /// Returns `None` when the cap has been reached. Callers should surface
+    /// this as `Unavailable` to the client (transient capacity, retryable).
+    pub fn subscribe(&self) -> Option<SubscribedReceiver> {
+        let permit = self
+            .subscription_semaphore
+            .clone()
+            .try_acquire_owned()
```
semenov-vladyslav, line 163: Why do you need a Semaphore? Why not just use AtomicUsize? You are not waiting for the Semaphore to become available anyway.

semenov-vladyslav (approved):
Looks good, but it seems that Semaphore can be replaced with just AtomicUsize, as non of the Semaphore's async functionality is not used.

alexsporn (approved):

crates/iota-grpc-server/src/types.rs
bingyanglin: Good point, but we only use `try_acquire_owned()`, never await. I chose `Semaphore` because it checks and increments the count in one safe step (with `AtomicUsize` I'd need a retry loop or increment-then-rollback, might be easier to get wrong?), and it auto-releases the slot when the stream drops. The permit acts like a keycard that cleans up on its own, whereas with `AtomicUsize` I'd have to write my own drop handler to decrement the counter. `subscribe()` only runs when a new client connects, so I guess there's no perf benefit to going lock-free?

Issue #10764, opened 2026-03-18
pcool: white flag resolution skips randomness transactions


piotrm50:
\`resolve_owned_object_conflicts\` only runs on \`sequenced_transactions\`. \`UserTransactionV1\` using randomness gets routed to \`sequenced_randomness_transactions\` (via \`is_user_tx_with_randomness()\` at \`authority_per_epoch_store.rs:3010\`) and bypasses conflict resolution entirely. Fix: run resolution on randomness transactions too, or reject \`UserTransactionV1\` with randomness until handled.

PR #11301, opened 2026-04-23
feat(iota-core): Post consensus load shedding

### LINKED ISSUE #11203: pcool: PoC post-consensus load shedding

cyberphysic4l:
# Description of change

Adds a post-consensus load shedding path for the certificate-less (white-flag) flow. Validators now broadcast their measured load shedding percentage to peers via a new consensus transaction, and each validator deterministically drops user transactions during post-consensus categorization based on the stake-weighted quorum percentile of those load shedding percentages — so the entire committee sheds the same set of transactions in the same round.

Prior to white-flag changes, the load shedding percentage was only used locally and was based on execution queue latency. There was also a transactions manager queue based threshold and writeback cache threshold above which everything was dropped. Now the previous latency based load shedding percentage is combined with a transaction manager queue and writeback cache based load shedding percentage, and all the dropping for these is done after consensus.

Stess testing detailed in #11203 

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1205,25 +1205,25 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_max_transaction_manager_queue_length")]
     pub max_transaction_manager_queue_length: usize,
 
-    /// Inflight execution queue length at which graduated load shedding begins
-    /// in the certificate-less (white-flag) mode. Defaults to
-    /// `max_transaction_manager_queue_length / 2`.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_transaction_manager_queue_length_soft_limit: Option<usize>,
-
     /// Reject a transaction if the number of pending transactions depending on
     /// the object is above the threshold.
     #[serde(default = "default_max_transaction_manager_per_object_queue_length")]
     pub max_transaction_manager_per_object_queue_length: usize,
+
+    /// Inflight execution queue length at which graduated load shedding begins
+    /// in the certificate-less (white-flag) mode. Defaults to
+    /// `max_transaction_manager_queue_length / 2`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_transaction_manager_queue_length_soft_limit_pct: Option<u32>,
 }
 
 impl AuthorityOverloadConfig {
-    /// Returns the inflight execution queue length at which graduated load
-    /// shedding begins in the certificate-less (white-flag) mode. Defaults
-    /// to `max_transaction_manager_queue_length / 2`.
-    pub fn max_transaction_manager_queue_length_soft_limit(&self) -> usize {
-        self.max_transaction_manager_queue_length_soft_limit
-            .unwrap_or(self.max_transaction_manager_queue_length / 2)
+    /// Returns the percentage of the max queue length at which graduated load
+    /// shedding should start.
+    pub fn max_transaction_manager_queue_length_soft_limit_pct(&self) -> u32 {
+        self.max_transaction_manager_queue_length_soft_limit_pct
+            .unwrap_or(50)
+            .min(100)
     }
 }
 
@@ -1281,7 +1281,7 @@ impl Default for AuthorityOverloadConfig {
             check_system_overload_at_signing: true,
             check_system_overload_at_execution: false,
             max_transaction_manager_queue_length: default_max_transaction_manager_queue_length(),
-            max_transaction_manager_queue_length_soft_limit: None,
+            max_transaction_manager_queue_length_soft_limit_pct: None,
             max_transaction_manager_per_object_queue_length:
                 default_max_transaction_manager_per_object_queue_length(),
         }
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -517,10 +517,12 @@ pub struct WritebackCacheConfig {
     pub backpressure_threshold_for_rpc: Option<u64>, // defaults to backpressure_threshold
 
     /// Percentage of `backpressure_threshold` at which graduated load shedding
-    /// based on writeback-cache pending transaction count begins. Sheds
-    /// linearly from 0% to 100% between
-    /// `backpressure_threshold * backpressure_soft_limit_pct / 100` and
-    /// `backpressure_threshold`. Defaults to 50.
+    /// based on writeback-cache pending transaction count begins. The
+    /// locally-calculated shedding percentage increases linearly from 0% at
+    /// `backpressure_threshold * backpressure_soft_limit_pct / 100` up to
+    /// 100% at the `backpressure_threshold` if the cache size continues to
+    /// increase. The calculated shedding percentage is broadcast to other
+    /// validators for a coordinated response. Defaults to 50.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub backpressure_soft_limit_pct: Option<u32>,
 }
@@ -1283,19 +1285,18 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_max_transaction_manager_per_object_queue_length")]
     pub max_transaction_manager_per_object_queue_length: usize,
 
-    /// Inflight execution queue length at which graduated load shedding begins
-    /// in the certificate-less (white-flag) mode. Defaults to
-    /// `max_transaction_manager_queue_length / 2`.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_transaction_manager_queue_length_soft_limit_pct: Option<u32>,
+    /// Percentage of `max_transaction_manager_queue_length` at which graduated
+    /// load shedding begins in the certificate-less (white-flag) mode. Read
+    /// via the same-named accessor, which clamps the value to <=100.
+    #[serde(default = "default_max_transaction_manager_queue_length_soft_limit_pct")]
+    pub max_transaction_manager_queue_length_soft_limit_pct: u32,
 }
 
 impl AuthorityOverloadConfig {
-    /// Returns the percentage of the max queue length at which graduated load
-    /// shedding should start.
+    /// Returns the soft-limit percentage, clamped to <=100 to guard against
+    /// out-of-range operator-supplied values.
     pub fn max_transaction_manager_queue_length_soft_limit_pct(&self) -> u32 {
         self.max_transaction_manager_queue_length_soft_limit_pct
-            .unwrap_or(50)
             .min(100)
     }
 }
@@ -1336,6 +1337,10 @@ fn default_max_transaction_manager_queue_length() -> usize {
     100_000
 }
 
+fn default_max_transaction_manager_queue_length_soft_limit_pct() -> u32 {
+    50
+}
+
 fn default_max_transaction_manager_per_object_queue_length() -> usize {
     20
 }
@@ -1354,7 +1359,8 @@ impl Default for AuthorityOverloadConfig {
             check_system_overload_at_signing: true,
             check_system_overload_at_execution: false,
             max_transaction_manager_queue_length: default_max_transaction_manager_queue_length(),
-            max_transaction_manager_queue_length_soft_limit_pct: None,
+            max_transaction_manager_queue_length_soft_limit_pct:
+                default_max_transaction_manager_queue_length_soft_limit_pct(),
             max_transaction_manager_per_object_queue_length:
                 default_max_transaction_manager_per_object_queue_length(),
         }
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -515,6 +515,14 @@ pub struct WritebackCacheConfig {
     /// if unset.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub backpressure_threshold_for_rpc: Option<u64>, // defaults to backpressure_threshold
+
+    /// Percentage of `backpressure_threshold` at which graduated load shedding
+    /// based on writeback-cache pending transaction count begins. Sheds
+    /// linearly from 0% to 100% between
+    /// `backpressure_threshold * backpressure_soft_limit_pct / 100` and
+    /// `backpressure_threshold`. Defaults to 50.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub backpressure_soft_limit_pct: Option<u32>,
 }
 
 impl WritebackCacheConfig {
@@ -612,6 +620,15 @@ impl WritebackCacheConfig {
             .and_then(|s| s.parse().ok())
             .or(self.backpressure_threshold_for_rpc)
             .unwrap_or(self.backpressure_threshold())
+    }
+
+    pub fn backpressure_soft_limit_pct(&self) -> u32 {
+        std::env::var("IOTA_CACHE_WRITEBACK_BACKPRESSURE_SOFT_LIMIT_PCT")
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .or(self.backpressure_soft_limit_pct)
+            .unwrap_or(50)
+            .min(100)
     }
 }
 
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1205,10 +1205,26 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_max_transaction_manager_queue_length")]
     pub max_transaction_manager_queue_length: usize,
 
+    /// Inflight execution queue length at which graduated load shedding begins
+    /// in the certificate-less (white-flag) mode. Defaults to
+    /// `max_transaction_manager_queue_length / 2`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_transaction_manager_queue_length_soft_limit: Option<usize>,
+
     /// Reject a transaction if the number of pending transactions depending on
     /// the object is above the threshold.
     #[serde(default = "default_max_transaction_manager_per_object_queue_length")]
     pub max_transaction_manager_per_object_queue_length: usize,
+}
+
+impl AuthorityOverloadConfig {
+    /// Returns the inflight execution queue length at which graduated load
+    /// shedding begins in the certificate-less (white-flag) mode. Defaults
+    /// to `max_transaction_manager_queue_length / 2`.
+    pub fn max_transaction_manager_queue_length_soft_limit(&self) -> usize {
+        self.max_transaction_manager_queue_length_soft_limit
+            .unwrap_or(self.max_transaction_manager_queue_length / 2)
+    }
 }
 
 fn default_max_txn_age_in_queue() -> Duration {
@@ -1265,6 +1281,7 @@ impl Default for AuthorityOverloadConfig {
             check_system_overload_at_signing: true,
             check_system_overload_at_execution: false,
             max_transaction_manager_queue_length: default_max_transaction_manager_queue_length(),
+            max_transaction_manager_queue_length_soft_limit: None,
             max_transaction_manager_per_object_queue_length:
                 default_max_transaction_manager_per_object_queue_length(),
         }
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -470,10 +470,12 @@ pub struct WritebackCacheConfig {
     pub backpressure_threshold_for_rpc: Option<u64>, // defaults to backpressure_threshold
 
     /// Percentage of `backpressure_threshold` at which graduated load shedding
-    /// based on writeback-cache pending transaction count begins. Sheds
-    /// linearly from 0% to 100% between
-    /// `backpressure_threshold * backpressure_soft_limit_pct / 100` and
-    /// `backpressure_threshold`. Defaults to 50.
+    /// based on writeback-cache pending transaction count begins. The
+    /// locally-calculated shedding percentage increases linearly from 0% at
+    /// `backpressure_threshold * backpressure_soft_limit_pct / 100` up to
+    /// 100% at the `backpressure_threshold` if the cache size continues to
+    /// increase. The calculated shedding percentage is broadcast to other
+    /// validators for a coordinated response. Defaults to 50.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub backpressure_soft_limit_pct: Option<u32>,
 }
@@ -1227,19 +1229,18 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_max_transaction_manager_per_object_queue_length")]
     pub max_transaction_manager_per_object_queue_length: usize,
 
-    /// Inflight execution queue length at which graduated load shedding begins
-    /// in the certificate-less (white-flag) mode. Defaults to
-    /// `max_transaction_manager_queue_length / 2`.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_transaction_manager_queue_length_soft_limit_pct: Option<u32>,
+    /// Percentage of `max_transaction_manager_queue_length` at which graduated
+    /// load shedding begins in the certificate-less (white-flag) mode. Read
+    /// via the same-named accessor, which clamps the value to <=100.
+    #[serde(default = "default_max_transaction_manager_queue_length_soft_limit_pct")]
+    pub max_transaction_manager_queue_length_soft_limit_pct: u32,
 }
 
 impl AuthorityOverloadConfig {
-    /// Returns the percentage of the max queue length at which graduated load
-    /// shedding should start.
+    /// Returns the soft-limit percentage, clamped to <=100 to guard against
+    /// out-of-range operator-supplied values.
     pub fn max_transaction_manager_queue_length_soft_limit_pct(&self) -> u32 {
         self.max_transaction_manager_queue_length_soft_limit_pct
-            .unwrap_or(50)
             .min(100)
     }
 }
@@ -1280,6 +1281,10 @@ fn default_max_transaction_manager_queue_length() -> usize {
     100_000
 }
 
+fn default_max_transaction_manager_queue_length_soft_limit_pct() -> u32 {
+    50
+}
+
 fn default_max_transaction_manager_per_object_queue_length() -> usize {
     20
 }
@@ -1298,7 +1303,8 @@ impl Default for AuthorityOverloadConfig {
             check_system_overload_at_signing: true,
             check_system_overload_at_execution: false,
             max_transaction_manager_queue_length: default_max_transaction_manager_queue_length(),
-            max_transaction_manager_queue_length_soft_limit_pct: None,
+            max_transaction_manager_queue_length_soft_limit_pct:
+                default_max_transaction_manager_queue_length_soft_limit_pct(),
             max_transaction_manager_per_object_queue_length:
                 default_max_transaction_manager_per_object_queue_length(),
         }
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1261,25 +1261,25 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_max_transaction_manager_queue_length")]
     pub max_transaction_manager_queue_length: usize,
 
-    /// Inflight execution queue length at which graduated load shedding begins
-    /// in the certificate-less (white-flag) mode. Defaults to
-    /// `max_transaction_manager_queue_length / 2`.
-    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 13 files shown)


crates/iota-core/src/overload_monitor.rs
```diff
@@ -96,23 +96,47 @@ fn check_authority_overload(
...
         txn_ready_rate,
         execution_rate,
     );
 
+    let queue_based_percentage = compute_queue_load_shedding_percentage(
+        inflight_queue_len,
+        config.max_transaction_manager_queue_length_soft_limit(),
+        config.max_transaction_manager_queue_length,
+        config.max_load_shedding_percentage,
+    );
```
roman1e2f5p8s: This part is broken now after I applied your suggestions to hard-code the max shedding percentage as 100% and rebased this branch. It is fine to hard-code it to 100% for consensus queue length because we have a clear signal (`num_inflight_txs`), but in the case of execution, signals might be noisy, so that is why `max_load_shedding_percentage` is set to less than 100%. I am confused now whether should we hard-code max shedding percentage in the function or pass as a parameter.

cyberphysic4l: I think it's fine. Recall that what existed before these changes was that everything would be dropped while the queue length was above the max - that is exactly what 100% load shedding is, so this applies to the execution queue as well as the consensus queue. The max load shedding percentage parameter corresponds to the latency-based load shedding that existed before, so I will make some adjustment to account for that.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -142,6 +143,14 @@ impl Debug for ConsensusTransactionKey {
                 )
             }
             Self::UserTransaction(digest) => write!(f, "UserTransaction({digest:?})"),
+            Self::OverloadNotificationV1(name, percentage) => {
+                write!(
+                    f,
+                    "OverloadNotificationV1({:?}, {:?})",
+                    name.concise(),
+                    percentage
```
Thoralf-M:
```suggestion
                    "OverloadNotificationV1({:?}, {percentage:?})",
                    name.concise()
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2754,6 +2762,74 @@ impl AuthorityPerEpochStore {
...
+            .collect::<Result<HashMap<AuthorityName, u8>, _>>()?;
+
+        let committee = self.committee();
+
+        // Build a vec of (percentage, stake) for every committee member.
+        // Default to 0% for authorities that haven't reported.
+        let mut weighted_values: Vec<(u8, StakeUnit)> = committee
+            .members()
+            .map(|(authority, stake)| {
+                let percentage = notifications.get(authority).copied().unwrap_or(0);
```
bingyanglin, line 2900: Validators that haven't sent a notification default to 0%. A malicious validator could stay silent to keep the quorum shedding percentage low. Maybe require periodic heartbeats even when unchanged? Or do I miss something?

crates/iota-node/src/lib.rs
```diff
@@ -454,6 +455,51 @@ impl IotaNode {
...
+                .overload_info
+                .load_shedding_percentage
+                .load(std::sync::atomic::Ordering::Relaxed);
+            loop {
+                tokio::time::sleep(poll_interval).await;
+                let current = state
+                    .overload_info
+                    .load_shedding_percentage
+                    .load(std::sync::atomic::Ordering::Relaxed);
+                if current != last_notified_percentage {
```
bingyanglin, line 324: This only sends when the percentage changes. New validators or validators recovering from partitions won't get fresh data if the system is stable. Maybe send a heartbeat every N intervals even when unchanged?

crates/iota-config/src/node.rs
```diff
@@ -510,6 +510,14 @@ pub struct WritebackCacheConfig {
     /// if unset.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub backpressure_threshold_for_rpc: Option<u64>, // defaults to backpressure_threshold
+
+    /// Percentage of `backpressure_threshold` at which graduated load shedding
+    /// based on writeback-cache pending transaction count begins. Sheds
```
piotrm50: > Sheds linearly 
this reads like when this threshold is reached then the node starts shedding. I would rephrase it to make it clear that this value is then broadcast to the whole committee and then the shedding level is determinisically calculated.

crates/iota-config/src/node.rs
```diff
@@ -1363,6 +1396,7 @@ impl Default for AuthorityOverloadConfig {
             check_system_overload_at_signing: true,
             check_system_overload_at_execution: false,
             max_transaction_manager_queue_length: default_max_transaction_manager_queue_length(),
+            max_transaction_manager_queue_length_soft_limit_pct: None,
```
piotrm50: Why is this None here and does not have a corresponding `default_` function like all other fields?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2754,6 +2762,74 @@ impl AuthorityPerEpochStore {
             .collect::<Result<Vec<_>, _>>()?)
     }
 
+    pub fn record_overload_notification_v1(
+        &self,
+        authority: &AuthorityName,
+        percentage: u8,
+    ) -> IotaResult {
+        info!(
```
piotrm50: This logs is going to look confusing because there is similar log on debug level where this transaction kind is processed (L4666)

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -4537,6 +4655,28 @@ impl AuthorityPerEpochStore {
...
+                if self
+                    .get_reconfig_state_read_lock_guard()
+                    .should_accept_consensus_certs()
+                {
+                    debug!(
+                        "Received OverloadNotificationV1 from {:?} with percentage {}",
+                        authority.concise(),
+                        percentage
+                    );
+                    self.record_overload_notification_v1(authority, *percentage)?;
```
piotrm50: This cannot modify the database directly. It should go through the ConsensusQuarantine buffer zone and modify the database at the same time as all other things are flushed there. Otherwise if a node crashes during high traffic period, it will have a diverged database state after restart.

On a crash mid-commit-N, you can end up with:

  - DB has the `OverloadNotification` rows from commit N (durable from the synchronous insert).
  - DB does not have `consensus_message_processed[commit N]` (that's still in the in-memory `ConsensusCommitOutput` that died with the process).
  - On restart, replay re-runs commit N from scratch.
  - process_consensus_transactions for commit N calls `get_quorum_load_shedding_percentage()` → reads `authority_overload_notifications` straight from RocksDB → sees the leftover writes from the pre-crash run, which were not visible during the original pre-crash run of commit N.
  - `drop_percentage` for commit N comes out different → `should_reject_tx` drops a different set of user transactions → this validator's executed effects diverge from peers that didn't crash.

  That's exactly the failure the quarantine layer was built to prevent.

`record_capabilities_v1` also writes directly. That works only because capabilities are read during reconfiguration at the epoch boundary, where the protocol re-synchronizes the committee anyway, and the read is idempotent w.r.t. replay (highest generation wins, replays just write the same value back).

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3178,15 +3274,37 @@ impl AuthorityPerEpochStore {
...
                 system_transactions.push(tx);
-            } else if !enable_white_flag && tx.0.is_user_tx_with_randomness() {
-                current_commit_sequenced_randomness_transactions.push(tx);
             } else {
-                current_commit_sequenced_consensus_transactions.push(tx);
+                // For user transactions, deterministically drop based on the
+                // quorum load shedding percentage.
+                if drop_percentage > 0 {
+                    if let Some(digest) = tx.0.transaction.executable_transaction_digest() {
+                        if should_reject_tx(drop_percentage, digest, drop_seed) {
```
piotrm50, line 3385: should we make sure to not drop UserTransactions such as `OverloadNotification` or `AuthorityCapabilities`? `OverloadNotification` is more crucial I think. Or is that handled someplace else?

crates/iota-node/src/lib.rs
piotrm50: This data is only relevant during an epoch and the overload monitor is reset at the epoch boundary. I don't think it's necessary to issue every N intervals with the same status. If a node has restarted and is catching up from either clean or outdated consensus database, then all of the overload notifications live inside the DAG anyway, and will be process by consensus_handler in the correct order anyway. Could you describe the scenario where not issuing at regular intervals would be problematic?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3132,6 +3208,26 @@ impl AuthorityPerEpochStore {
                     return None;
                 }
             }
+            SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::OverloadNotificationV1(authority, percentage),
+                ..
+            }) => {
+                if &transaction.sender_authority() != authority {
```
piotrm50, line 3269: `validate_transactions_feature_gating` - does this test pass if there is no feature gating on this transaction kind? Perhaps the test is missing something, but in general we don't want to allow malicious validators to issue a transaction kind before it's supported by a feature flag despite the software version being able to deserialize it. Say, we merge this to develop and release, but mainnet/testnet does not have pcool enabled. Then all nodes would try to process this transaction kind (and potentially crash) without protocol supporting this feature. That's what the `validate_transactions_feature_gating` test was supposed to catch, so maybe it's actually gated somehow, but I figured I'd ask anyway just to make sure.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: but again - what do you do if somebody doesn't send the heartbeat? you treat  it as zero?

crates/iota-core/src/authority.rs
```diff
@@ -554,6 +556,11 @@ impl AuthorityMetrics {
                 "Percentage of transactions shed due to consensus queue length.",
                 registry)
                 .unwrap(),
+            cache_backpressure_load_shedding_percentage: register_int_gauge_with_registry!(
```
piotrm50: we should add metrics to count how many transactions were actually shedded post-consensus. Optionally we should count how many notifications we've received from each authority, but I'm not sure if that's worth polluting the metrics (we have too many already 😅).

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: Ok, I see this is checked in consensus_validator.rs, but I added explicit check here for SignedCapabilityNotification and there is a similar TODO for UserTransactionV1, so I guess we should clean this up at some point.

piotrm50 (changes requested):

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: A malicious validator could also simply broadcast a value of 0 and achieve the same effect, nothing enforces particular values to be sent, it is all based on local an unprovable measurements. The security against malicious behaviour comes from the fact that the stake weighted quorum percentile is used to compute the effective load shedding value, limiting the ability of malicious validators to bias the load shedding percentage.

crates/iota-node/src/lib.rs
cyberphysic4l: This is an important concern that you've highlighted and I have been trying to figure out exactly how different new validator or crash and recovery situations will work. If I understand correctly now, validators will always process all commits in order, and when processing a new commit (where load shedding is done) the epoch store tables will always reflect the latest overload notification from each authority. If there was no notificiation in the epoch, the value is zero. So getting fresh data is not a concern in the sense that the data comes from commits and we are guaranteed to have fresh commit data when we need it.

Although getting fresh data from commits is not a problem, broadcasting fresh data about local conditions may be what you mean. So if a validator was overloaded when it crashed and hence the committed overload notification for this validator has a high load shedding percentage, they may not update this by broadcasting a new value when they recover from the crash. I think this may be better solved by broadcasting (even a zero value) upon start up to make sure each validator has put out a fresh value since last starting up. Does that make sense or do you think there is another reason for having a heartbeat-style approach to broadcasting?

crates/iota-config/src/node.rs
cyberphysic4l: I changed that now to match the others better. It was related to the fact that we also clamp the value to 100, so I also had the default just set in the getter.

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: This was actually handled already, `executable_transaction_digest()` would return digests only for user originating transactions and system transactions, but system transactions already dealt with in earlier match arm. I've made it more explicit now by having the `user_transaction_digest()` method which will only return digests for user originated transaction, so `OverloadNotification` and `AuthorityCapabilities` etc will never be dropped by this load shedding.

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: This was actually fixed (not intentionally) by a change I pushed shortly after your review (0559d1b3d0e02202fde5e67971038c48eb4ab1c4) which ensures that the overload notification from each commit are processed right before the load shedding. However, it is definitely better to use the established mechanism for consistent storage with the consensus quarantine, so I have refactored now.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3033,15 +3128,62 @@ impl AuthorityPerEpochStore {
...
+                // never be dropped here.
+                if drop_percentage > 0 {
+                    if let Some(digest) = tx.0.transaction.user_transaction_digest() {
+                        if should_reject_tx(drop_percentage, digest, drop_seed) {
+                            continue;
+                        }
+                    }
+                }
+
+                if tx.0.is_user_tx_with_randomness() {
```
```suggestion
                if !enable_white_flag && tx.0.is_user_tx_with_randomness() {
```
see https://github.com/iotaledger/iota/issues/10764 why it should NOT be removed.

crates/iota-node/src/lib.rs
```diff
@@ -279,6 +280,51 @@ impl IotaNode {
...
+        }
+
+        let poll_interval = config.authority_overload_config.overload_monitor_interval;
+        let authority_name = state.name;
+
+        Some(spawn_monitored_task!(async move {
+            let mut last_notified_percentage: u32 = state
+                .overload_info
+                .local_load_shedding_percentage
+                .load(std::sync::atomic::Ordering::Relaxed);
```
roman1e2f5p8s: shouldn't `last_notified_percentage` be initialized from the persisted view instead of the atomic on startup? something like
```rust
let mut last_notified_percentage: u32 = epoch_store
    .load_overload_notifications()
    .expect("...")
    .get(&authority_name)
    .copied()
    .map(u32::from)
    .unwrap_or(0);
```

crates/iota-core/src/consensus_validator.rs
```diff
@@ -122,6 +122,16 @@ impl IotaTxValidator {
...
+
+                ConsensusTransactionKind::OverloadNotificationV1(_, _) => {
+                    if !self.epoch_store.protocol_config().enable_white_flag_flow() {
+                        return Err(IotaError::UnsupportedFeature {
+                            error:
+                                "OverloadNotificationV1 not supported at current protocol version"
+                                    .into(),
+                        });
+                    }
+                }
```
roman1e2f5p8s:
```suggestion
                ConsensusTransactionKind::OverloadNotificationV1(authority_name, percentage) => {
                    if !self.epoch_store.protocol_config().enable_white_flag_flow() {
                        return Err(IotaError::UnsupportedFeature {
                            error:
                                "OverloadNotificationV1 not supported at current protocol version"
                                    .into(),
                        });
                    }
                    if *percentage > 100 {
                        return Err(IotaError::HandleConsensusTransactionFailure(format!(
                            "OverloadNotificationV1 with invalid percentage {percentage} from \
                                authority {}",
                            authority_name.concise(),
                        )));
                    }
                }
```
not just feature gate, but worth doing a cheap check - some other variants do that here.

roman1e2f5p8s (commented):
one test I would add is to exercise the actual drop behavior inside `process_consensus_transactions_and_commit_boundary`.
see `test_consensus_commit_prologue_generation` at `authority_tests.rs:4671` for how such a test could be constructed

crates/iota-core/src/authority.rs
cyberphysic4l: Added some dropped tx metrics

crates/iota-types/src/messages_consensus.rs
```diff
@@ -602,6 +628,9 @@ impl ConsensusTransaction {
             ConsensusTransactionKind::UserTransactionV1(tx) => {
                 ConsensusTransactionKey::UserTransaction(*tx.digest())
             }
+            ConsensusTransactionKind::OverloadNotificationV1(authority, percentage) => {
+                ConsensusTransactionKey::OverloadNotificationV1(*authority, *percentage)
```
bingyanglin: The key includes the percentage, and `consensus_message_processed` dedup is epoch-wide — so each (authority, value) pair can only be sequenced once per epoch. After 80 → 0 → 80, the second 80 is silently skipped everywhere and the notifier never retries. Maybe add a generation field like `CapabilityNotificationV1` and key on (authority, generation)?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3205,14 +3310,72 @@ impl AuthorityPerEpochStore {
...
                 current_commit_sequenced_randomness_transactions.push(tx);
             } else {
+                // Only user-originated transactions are eligible for load shedding
+                if drop_percentage > 0 {
+                    if let Some(digest) = tx.0.transaction.user_transaction_digest() {
+                        if should_reject_tx(drop_percentage, digest, drop_seed) {
+                            authority_metrics
+                                .consensus_handler_load_shedding_dropped_transactions
+                                .inc();
+                            continue;
```
bingyanglin, line 3418: This `continue` drops the tx without `dropped_tx_status_cache.insert_and_notify`, unlike the validation-drop path below. So clients hang 30s in `wait_for_tx_finality`, and the submitter's `submit_and_wait_inner` pins its `submit_semaphore` permit until epoch end, under sustained shedding the pool drains and even `EndOfPublish` can't get a permit. Maybe collect shed digests and `insert_and_notify` with `ValidatorOverloadedRetryAfter`? Or is it not important?

crates/iota-core/src/overload_monitor.rs
```diff
@@ -134,23 +140,68 @@ fn check_authority_overload(
...
     );
 
-    let (is_overload, load_shedding_percentage) = check_overload_signals(
+    // Feedback term: the locally computed percentage this monitor produced on
+    // its previous iteration. We use the *local* value (not the quorum value
+    // enforced post-consensus) because `txn_ready_rate` reflects the load this
+    // authority itself admitted, so the controller must close the loop on its
+    // own last decision to compound shedding over multiple iterations and to
+    // ratchet down gradually as latency recovers.
+    let current_local_load_shedding_percentage = authority
```
bingyanglin: This feedback is now the max of all three signals, so a one-tick queue/cache spike decays at only 10 points per tick through the latency controller's gradual-reduction branch, phantom shedding long after the spike cleared. Maybe feed back only the latency controller's own previous output?

crates/iota-node/src/lib.rs
bingyanglin: Agreed, commits replay in order, so the table is always consistent, and I also agree no heartbeat is needed for that.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2834,6 +2842,83 @@ impl AuthorityPerEpochStore {
...
+            .collect::<Result<HashMap<AuthorityName, u8>, _>>()?;
+        self.consensus_quarantine
+            .read()
+            .overlay_queued_overload_notifications(&mut notifications);
+        Ok(notifications)
+    }
+
+    /// Returns the load shedding percentage `authority` last broadcasted or 0
+    /// if it has not sent a notification this epoch.
+    pub fn load_overload_notification(&self, authority: &AuthorityName) -> IotaResult<u8> {
```
piotrm50, line 2869: how often is this called? we are reading rocksdb for all authorities everytime. Perhaps we could cache those values and only read it upon startup? Disk storage is only necessary for persistence across restarts.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3205,14 +3310,72 @@ impl AuthorityPerEpochStore {
...
+        // the loop so user transactions can be dropped inline during categorization.
+        // `OverloadNotificationV1` transactions from this commit are layered on
+        // top of the persisted notifications now to ensure immediate response to new
+        // notifications which are not yet registered in persistent storage rather than
+        // the previous round's values. The recorder call inside
+        // `process_consensus_transactions` is gated by
+        // `should_accept_consensus_certs`; this overlay mirrors that gate so
+        // the overlaid set matches the set that will be flushed.
+        let drop_percentage = if enable_white_flag {
+            let mut notifications = self.load_overload_notifications()?;
```
piotrm50, line 3342: we hit rocksdb everytime here.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3205,14 +3310,72 @@ impl AuthorityPerEpochStore {
...
+            .consensus_handler_load_shedding_percentage
+            .set(drop_percentage as i64);
+        let drop_seed = consensus_commit_info.round;
+
         for tx in verified_transactions {
             if tx.0.is_end_of_publish() {
                 end_of_publish_transactions.push(tx);
             } else if tx.0.is_system() {
                 system_transactions.push(tx);
             } else if !enable_white_flag && tx.0.is_user_tx_with_randomness() {
```
piotrm50, line 3381: why `if !enable_white_flag`? that means that we are not separating randomness transactions in whiteflag.

crates/iota-core/src/unit_tests/authority_per_epoch_store_tests.rs
```diff
@@ -145,3 +171,146 @@ async fn wait_for_transactions_in_checkpoint_times_out_without_notify() {
...
+    assert_eq!(
+        from_queue,
+        Some(80),
+        "queued notifications must be visible via load_overload_notifications",
+    );
+
+    // Now the same logical state sits fully on disk (as it would after the
+    // queued commit drains to RocksDB). The read must produce the same
+    // notification map as before.
+    flush_overload_notification(&store, authority_name, 80);
```
piotrm50: do we remove the entry from the consensus quarantine here? I think what is happening is that we're reading 80 from rocksdb and later we're overlaying it with quarantine's 80. If we want to simulate a fresh start we should recreate the store from scratch.

crates/iota-types/src/messages_consensus.rs
cyberphysic4l: Thanks for catching this. I had actually caught this while stress testing and fixed it on the stress test branch [here](https://github.com/iotaledger/iota/commit/0309d492ea2a1666f09ca386f728528136fcdde2), but forgot to port it to this branch!

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: This is also one that was fixed on the stress testing branch [here](https://github.com/iotaledger/iota/commit/9109fdb39bdcbdd9379dc2eaaee3f6f72490efc9)

crates/iota-core/src/overload_monitor.rs
cyberphysic4l: That's true, it is more correct to feedback only the latency signal, although I don't think it is a big problem to have that slightly slowed retreat after a queue/cache spike. I will fix it anyway, thanks.

crates/iota-types/src/messages_consensus.rs
cyberphysic4l: Fixed here [253ed46](https://github.com/iotaledger/iota/pull/11301/commits/253ed466aa4117b719ce70297efd659a48b43e0a)

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: Fixed here [d3528c3](https://github.com/iotaledger/iota/pull/11301/commits/d3528c382574742dc1813b6b39da2a3e5fa94ec9)

crates/iota-core/src/overload_monitor.rs
cyberphysic4l: Fixed here [cee1341](https://github.com/iotaledger/iota/pull/11301/commits/cee1341415ad4072de6102ca6c7151124c248b77)

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: Fixed here [aeda571](https://github.com/iotaledger/iota/pull/11301/commits/aeda571d3f3d3f7097741812729ea19e637eabec)

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: Fixed here [aeda571](https://github.com/iotaledger/iota/pull/11301/commits/aeda571d3f3d3f7097741812729ea19e637eabec)

crates/iota-core/src/authority/authority_per_epoch_store.rs
cyberphysic4l: Comment added here [00b0d82](https://github.com/iotaledger/iota/pull/11301/commits/00b0d82daca4cb2854dbea0e4374e05956c5c5ca)

crates/iota-core/src/unit_tests/authority_per_epoch_store_tests.rs
cyberphysic4l: Fixed here [cdb5d77](https://github.com/iotaledger/iota/pull/11301/commits/cdb5d77038aa85fc66fa8b73d8b9a9c215228a3b)

bingyanglin (approved):

piotrm50 (approved):

crates/iota-types/src/messages_consensus.rs
```diff
@@ -239,6 +247,11 @@ pub enum ConsensusTransactionKind {
     /// directly to consensus without pre-consensus object locking.
     /// Conflicts are resolved post-consensus.
     UserTransactionV1(Box<Transaction>),
+    OverloadNotificationV1(
+        AuthorityName,
+        u64, // generation
```
piotrm50, line 252: the second parameter here is `generation`, but when you destruct it in `consensus_validator.rs:126` you treat it as percentage!

crates/iota-core/src/consensus_validator.rs
```diff
@@ -122,6 +123,23 @@ impl IotaTxValidator {
 
                 ConsensusTransactionKind::EndOfPublish(_)
                 | ConsensusTransactionKind::CapabilityNotificationV1(_) => {}
+
+                ConsensusTransactionKind::OverloadNotificationV1(authority_name, percentage, _) => {
```
piotrm50: Order of values is broken! The definition of the enum looks as following! 

```
OverloadNotificationV1(
        AuthorityName,
        u64, // generation 
        u8,  // percentage
    ),
```

crates/iota-types/src/messages_consensus.rs
```diff
@@ -602,6 +646,9 @@ impl ConsensusTransaction {
             ConsensusTransactionKind::UserTransactionV1(tx) => {
                 ConsensusTransactionKey::UserTransaction(*tx.digest())
             }
+            ConsensusTransactionKind::OverloadNotificationV1(authority, generation, _) => {
```
piotrm50, line 649: here it's correctly used as `generation`.

piotrm50 (changes requested):
`generation` is treated as `percentage` value for load shedding!

crates/iota-core/src/consensus_validator.rs
cyberphysic4l: Thanks for catching that, fixed now in [this commit](https://github.com/iotaledger/iota/commit/b2dbcf25c29f90e5ba70098c8d5aa732d20ba728)

roman1e2f5p8s (approved):

piotrm50 (approved):

PR #11311, opened 2026-04-24
feat(iota-rest-kv): add object before version support

### LINKED ISSUE #11304: [iota-rest-kv]: add object before version support
Refactor the `iota-rest-kv` to incorporate the `range_scan` from `iota-bigtable` to implement the object before version capability.

sergiupopescu199:
# Description of change

Adds an optional `?before_version=<bool>` query parameter to the object retrieval routes of `iota-rest-kv`. When `true`, the server returns the highest stored object version **strictly less than** the version encoded in the request key, instead of the exact-version match.

This unblocks callers that know the `ObjectID` but not the exact version (e.g. resolving a past object state) from using the REST KV store. Previously only exact-version lookups were possible.

### Implementation details

-  **Query parameter over dedicated route.** The existing routes are generic over `item_type` (`/{item_type}` and `/{item_type}/{key}`). Adding a dedicated route for this one retrieval mode would introduce an object-only special case in an otherwise uniform route table. Expressing the mode as an optional query parameter keeps the routes generic and makes the feature backward-compatible by default. This also aligns with common REST guidance: the resource is the same (`/ob/{key}` or `/ob`), only the retrieval mode changes, which is the canonical use case for a query parameter rather than a new path.
- **Validation in a custom extractor.** A `BeforeVersionQuery` extractor reads the flag and, when `true`, enforces that the item type is `ob`. Everything else is rejected with `400 Bad Request` before the handler runs. This keeps route handlers focused on business logic.
- **Supported routes.** Both `GET /{item_type}/{key}` and `POST /{item_type}` accept the parameter. When absent it defaults to `false`, so existing callers are unaffected.
-  **Batch semantics.** The response is a BCS-serialized `Vec<Option<Bytes>>` that preserves the order of the input `keys`. `Some(bytes)` is the highest version strictly below the requested one; `None` means not found.
- **Edge case.** `before_version=true` with a key at version 0 returns `None`.
-  **BigTable scan.** Implemented as a single reverse `range_scan` on the objects table with `rows_limit = 1`. The scan is bounded to a single object by setting the start key (inclusive) to `(object_id, version = 0)` and the end key (exclusive) to `(object_id, requested_version)`. The lower bound prevents the scan from spilling into the previous `object_id`.

fixes #11304 

- Manual tests for `BeforeVersionQuery` (valid/invalid item types, absent parameter, default value) using testnet bigtable instance.
- For testing we picked the `Clock` object, since it's always mutated and has lots of versions to work with.

```shell
# POST
curl --location 'http://0.0.0.0:3555/ob?before_version=true' \
--data '{
    "keys": [
        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAYnAQAAAAAAAA",
        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAYmAQAAAAAAAA"
    ]
}'
```

```shell
# GET
curl --location 'http://0.0.0.0:3555/ob/AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAYnAQAAAAAAAA?before_version=true'
```
- Extractor validation of the `before_version=true` with an item type which is not an `ob` one.
```shell
curl --location 'http://0.0.0.0:3555/tx/Y1OOIQeD_oRmu8-7E2yFbqrboSrprUKBu0-HTMAo_Ek?before_version=true'
```
Response
```json
{
    "error_code": "400",
    "error_message": "bad request: `before_version` query parameter is only valid for `ob` item types"
}

```

> [!NOTE]

### DIFF
--- a/crates/iota-kvstore/Cargo.toml
+++ b/crates/iota-kvstore/Cargo.toml
@@ -12,7 +12,7 @@ anyhow.workspace = true
 async-trait.workspace = true
 bcs.workspace = true
 clap.workspace = true
-iota-bigtable = { git = "https://github.com/iotaledger/iota-bigtable.git", rev = "265bc250c7772a3703ed965c9542f61d35f54821" }
+iota-bigtable = { git = "https://github.com/iotaledger/iota-bigtable.git", tag = "v0.1.0" }
 serde.workspace = true
 strum.workspace = true
 tokio.workspace = true
--- a/crates/iota-rest-kv/src/bigtable.rs
+++ b/crates/iota-rest-kv/src/bigtable.rs
@@ -10,6 +10,7 @@ use std::{
 
 use anyhow::Result;
 use bytes::Bytes;
+use futures::{StreamExt, TryStreamExt, stream};
 use iota_kvstore::{
     BigTableClient, Cell,
     client::{
@@ -19,14 +20,21 @@ use iota_kvstore::{
         TRANSACTION_COLUMN_QUALIFIER, TRANSACTION_TO_CHECKPOINT, TRANSACTIONS_TABLE,
         raw_object_key,
     },
-    proto::bigtable::v2::{RowFilter, row_filter::Filter},
+    proto::bigtable::v2::{
+        RowFilter,
+        row_filter::Filter,
+        row_range::{EndKey, StartKey},
+    },
 };
-use iota_storage::http_key_value_store::Key;
-use iota_types::effects::TransactionEvents;
+use iota_storage::http_key_value_store::{ItemType, Key};
+use iota_types::{effects::TransactionEvents, storage::ObjectKey};
 use serde::{Deserialize, Serialize};
 use tracing::error;
 
-use crate::errors::ApiError;
+use crate::errors::{ApiError, RangeKeyBoundError};
+
+/// The maximum number of concurrent futures allowed when scanning BigTableDB.
+const MAX_CONCURRENT_FUTURES: usize = 100;
 
 /// Configuration for the [`KvStoreClient`] used to access data from BigTableDB
 #[derive(Serialize, Deserialize, Clone, Debug)]
@@ -81,6 +89,22 @@ impl KvStoreClient {
         })
     }
 
+    /// Builds a [`RowFilter`] that matches only cells whose column qualifier
+    /// equals `column_qualifier` exactly.
+    ///
+    /// Internally the filter is implemented as a regex anchored with `^` and
+    /// `$` to enforce an exact byte match of the provided
+    /// `column_qualifier`, preventing partial or substring matches against
+    /// other columns whose qualifiers happen to contain `column_qualifier`
+    /// as a prefix, suffix, or substring.
+    fn column_qualifier_filter(column_qualifier: &str) -> RowFilter {
+        RowFilter {
+            filter: Some(Filter::ColumnQualifierRegexFilter(
+                format!("^{column_qualifier}$").into_bytes(),
+            )),
+        }
+    }
+
     /// Get the elapsed time from which the service was instantiated.
     pub fn get_uptime(&self) -> Duration {
         self.start_time.elapsed()
@@ -90,7 +114,7 @@ impl KvStoreClient {
     ///
     /// Based on the provided [`Key`] fetch the data from BigTableDB.
     pub async fn get(&self, key: Key) -> Result<Option<Bytes>, ApiError> {
-        let results = self.multi_get(vec![key]).await?;
+        let results = self.get_items(vec![key]).await?;
         Ok(results.into_iter().next().unwrap_or(None))
     }
 
@@ -104,49 +128,37 @@ impl KvStoreClient {
     ///
     /// All keys must be of the same type, otherwise [`ApiError::BadRequest`] is
     /// returned.
-    pub async fn multi_get(&self, keys: Vec<Key>) -> Result<Vec<Option<Bytes>>, ApiError> {
+    pub async fn get_items(&self, keys: Vec<Key>) -> Result<Vec<Option<Bytes>>, ApiError> {
         if keys.is_empty() {
             return Ok(Vec::new());
         }
 
-        let mut client = self.bigtable_client.clone();
-
         // Use the first key to determine the type - all keys should be of the same type
         match keys.first().expect("emptiness was checked earlier") {
             Key::Transaction(_) => {
-                let digests = Self::extract_keys(&keys, |k| match k {
+                let digests = extract_keys(&keys, |k| match k {
                     Key::Transaction(digest) => Some(*digest),
                     _ => None,
                 })?;
 
                 let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
-                multi_get_cell(
-                    &mut client,
-                    TRANSACTIONS_TABLE,
-                    keys,
-                    TRANSACTION_COLUMN_QUALIFIER,
-                )
-                .await
+                self.fetch_from_bigtable(TRANSACTIONS_TABLE, keys, TRANSACTION_COLUMN_QUALIFIER)
+                    .await
             }
             Key::TransactionEffects(_) => {
-                let digests = Self::extract_keys(&keys, |k| match k {
+                let digests = extract_keys(&keys, |k| match k {
                     Key::TransactionEffects(digest) => Some(*digest),
                     _ => None,
                 })?;
 
                 let keys = digests.iter().map(|tx| Some(tx.inner().to_vec())).collect();
 
-                multi_get_cell(
-                    &mut client,
-                    TRANSACTIONS_TABLE,
-                    keys,
-                    EFFECTS_COLUMN_QUALIFIER,
-                )
-                .await
+                self.fetch_from_bigtable(TRANSACTIONS_TABLE, keys, EFFECTS_COLUMN_QUALIFIER)
+                    .await
             }
             Key::CheckpointContents(_) => {
-                let seq_nums = Self::extract_keys(&keys, |k| match k {
+                let seq_nums = extract_keys(&keys, |k| match k {
                     Key::CheckpointContents(seq_num) => Some(*seq_num),
                     _ => None,
                 })?;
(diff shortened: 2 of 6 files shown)


crates/iota-kvstore/Cargo.toml
```diff
@@ -12,7 +12,7 @@ anyhow.workspace = true
 async-trait.workspace = true
 bcs.workspace = true
 clap.workspace = true
-iota-bigtable = { git = "https://github.com/iotaledger/iota-bigtable.git", rev = "265bc250c7772a3703ed965c9542f61d35f54821" }
+iota-bigtable = { git = "https://github.com/iotaledger/iota-bigtable.git", rev = "524ca2ae0b39018b0321bb9cff65536fda460886" }
```
tomxey: what would you say about introducing some version tags in `iota-bigtable` repo? to not have to reference released versions through commit hashes.

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -257,113 +259,271 @@ impl KvStoreClient {
...
-    /// Extracts specific key type from a general [`Key`] type.
+    /// Fetch multiple values from a BigTable table with a specific key and
+    /// column qualifier.
     ///
-    /// Takes:
-    /// - `keys`: The list of keys to extract from
-    /// - `extractor`: Function that returns Some(extracted_value) for the
-    ///   target variant, None otherwise
+    /// Keys wrapped in `Option<Vec<u8>>` allow chaining multiple queries: the
+    /// result from one `multi_get_cell` (which contains `None` for missing
```
tomxey: The function name needs updating, `multi_get_cell` no longer exists.

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -257,113 +259,271 @@ impl KvStoreClient {
...
+impl ObjectRangeKeyBound {
+    /// Returns `true` if the range cannot contain any rows.
+    ///
+    /// The `start_key` is always `(object_id, version=0)` inclusive, so the
+    /// range collapses to empty only when the `end_key` encodes the same
+    /// `(object_id, version=0)` exclusive, i.e. when the caller asked for
+    /// "the version strictly less than 0", which has no answer.
+    fn is_empty(&self) -> bool {
+        match (&self.start_key, &self.end_key) {
+            (StartKey::StartKeyClosed(s), EndKey::EndKeyOpen(e)) => s == e,
```
tomxey: Should we also cover case when start key is higher than end key?

sergiupopescu199: I think it's not necessary, the `ObjectRangeKeyBound` does not provide a constructor and its fields are private, so the only way of creating one is by its `From<ObjectKey>` implementation.

The start key has a zero version suffix (`ObjectKey::min_for_id`), the end key has the user-supplied version. That guarantees `start_key <= end_key`. The only edge case is `start_key == end_key` when the requested version is 0, which the `is_empty()` method already covers.

crates/iota-kvstore/Cargo.toml
sergiupopescu199: I do agree, then what I will do is when we merge https://github.com/iotaledger/iota-bigtable/pull/10 I will create a tag `v0.1.0` and update the `Cargo.toml` for the `iota-kvstore` to point the tag.

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11311/commits/2ffc5e8508e98b964ebacf185f1306525ac3cfeb

crates/iota-rest-kv/src/bigtable.rs
tomxey: Would it be a problem to cover such case just in case?
the other key combinations will fail via `unreachable` explicitly, but incorrect key comparison here could hide potential future errors silently

crates/iota-rest-kv/src/bigtable.rs
tomxey: `multi_get` is an implementation detail from the point of view of this function's interface
It's the `fetch_from_bigtable` that can be chained thanks to this behaviour.

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: Agreed, being more strict is better instead of silent errors.

fixed: https://github.com/iotaledger/iota/pull/11311/commits/0d2cc15299b99ae4ff3c7f9968335434c8b2432e

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11311/commits/0d2cc15299b99ae4ff3c7f9968335434c8b2432e

tomxey (approved):

crates/iota-rest-kv/src/extractors.rs
```diff
@@ -52,3 +53,140 @@ where
...
+/// The inner `bool` reflects the validated retrieval mode: `true` means
+/// "return the latest version strictly less than the version encoded in the
+/// key", `false` means "exact-version lookup".
+///
+/// This extractor adapts to both `/{item_type}` (POST routes) and
+/// `/{item_type}/{key}` (GET routes) by reading only the `item_type`
+/// segment, ignoring `key` if present.
+pub struct BeforeVersionQuery(pub bool);
+
+impl<S> FromRequestParts<S> for BeforeVersionQuery
```
kodemartin: It is not evident to me what is the benefit of implementing this extended extractor logic, while it can be accommodated by the builtin `Query` extractor.

Adding the validation logic for other parts of the path does not warrant this, and extends beyond the extraction scope that should be independent. 

The validation rule is `before_version` requires `item_type=ob`". That's a one-liner in the handler, where both values are already in scope.

crates/iota-rest-kv/src/extractors.rs
```diff
@@ -52,3 +53,140 @@ where
...
+/// - key decoding into [`Key`] against the path's [`ItemType`].
+///
+/// Handlers receive the resulting [`Vec`]<[`Key`]> directly.
+///
+/// # NOTE
+/// This extractor consumes the request body and should be used only once per
+/// request and shall be used as last extractor in the request pipeline.
+pub struct ExtractKeys(pub Vec<Key>);
+
+impl FromRequest<SharedRestServerAppState> for ExtractKeys {
```
kodemartin: What is the benefit of this impl as opposed to the inline version in `multi_get_data` where `Path` and `Json` were used to extract what here is duplicated?

This seems more verbose, and definitely less transparent. Previously you could just read the handler and see exactly what's validated and in what order.

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -257,113 +259,278 @@ impl KvStoreClient {
...
+            (StartKey::StartKeyClosed(start_key), EndKey::EndKeyOpen(end_key)) => {
+                // use `>=` instead of `==` defensively: the `From` impl currently
+                // guarantees `start_key <= end_key`, but if a future change ever
+                // produced a swapped range, treating it as empty here is safer
+                // than forwarding an invalid range to BigTable.
+                start_key >= end_key
             }
+            // `ObjectRangeKeyBound` has private fields and is only constructed via
+            // `From<ObjectKey>`, which always produces `(StartKeyClosed, EndKeyOpen)`.
+            _ => unreachable!(),
```
kodemartin: I would suggest covering all four possible combination to be more future proof, and remove the need for the comments. With little pattern-matching magic this might need fewer than 4 arms.

crates/iota-kvstore/Cargo.toml
sergiupopescu199: Added the tag: https://github.com/iotaledger/iota/pull/11311/commits/133409306e6e19d24258a82fd6c989eb6b17bb33

crates/iota-rest-kv/src/bigtable.rs
```diff
@@ -257,113 +259,278 @@ impl KvStoreClient {
...
+
+        let mut client = self.bigtable_client.clone();
+
+        let reversed = true;
+        // we just need one element, the most recent version strictly before the
+        // requested one.
+        let rows_limit = 1;
+        // reverse scan from `end_key` (exclusive upper bound) down to `start_key`
+        // (inclusive lower bound). The lower bound confines the scan to a single
+        // object's versions.
```
kodemartin: I don't think that the "lower bound confines the scan to a single object's version[s]" part is very clear. 

The following paragraph is more explicit about it. But this too seems to duplicate part of the `client.range_scan` doc and other info encoded in the types that we use here. 

Nit: Revisit if these two paragraphs add more value to warrant the extra reading effort.

kodemartin (commented):
looks good in terms of functionality and correctness. Left a few suggestions that could simplify the patch, and reduce maintenance.

crates/iota-rest-kv/src/extractors.rs
sergiupopescu199: Wanted to achieve separation of concerns: the goal was to keep validation out of the handler at the cost of more verbose extractors.

Agreed the simpler version is easier to follow, fixed: https://github.com/iotaledger/iota/pull/11311/commits/54f5f577971ac306ed30956f21535f7b8b289f13

crates/iota-rest-kv/src/extractors.rs
sergiupopescu199: Motivation was the same as  in : https://github.com/iotaledger/iota/pull/11311#discussion_r3167525399

fixed: https://github.com/iotaledger/iota/pull/11311/commits/54f5f577971ac306ed30956f21535f7b8b289f13

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: Handled all variants as suggested.

fixed: https://github.com/iotaledger/iota/pull/11311/commits/54f5f577971ac306ed30956f21535f7b8b289f13

crates/iota-rest-kv/src/bigtable.rs
sergiupopescu199: Since the `range_scan` already provides a comprehensive rust doc, I've opted to remove the comment entirely.

fixed: https://github.com/iotaledger/iota/pull/11311/commits/54f5f577971ac306ed30956f21535f7b8b289f13

kodemartin (approved):
lgtm 🎈

PR #11312, opened 2026-04-24
feat(iota-indexer): add fallback support for object before version

### LINKED ISSUE #11053: [indexer]: add fallback for queries on objects before version
## Description

The `historical_fallback::reader` supports queries on objects with a pinned version. But querying for objects before a given version is a TODO:

https://github.com/iotaledger/iota/blob/5faf4a2829cf5281b837b5f13c0c7a18cf26fb51/crates/iota-indexer/src/historical_fallback/reader.rs#L357-L361

In order to support `iota-replay` after pruning we must implement support for this kind of queries.

Bigtable seems to support this exact type of [reverse scans][reverse-scans].

[reverse-scans]: https://docs.cloud.google.com/bigtable/docs/reads#reverse-scan

sergiupopescu199:
# Description of change

When the indexer's database is pruned, the JSON-RPC `iota_tryGetPastObject` call cannot resolve object versions that no longer exist locally. This PR extends the `iota-indexer` historical fallback client to use the REST KV store's new `before_version=true` query parameter when fetching objects, so past versions can be served from BigTable when local data has been pruned.

### Implementation details

- **New signature for the fallback client.** `HttpRestKVClient::multi_get_objects` now takes an extra `before_version: bool` argument. When `true`, each requested `(object_id, version)` is resolved to the highest stored version strictly less than the requested one, instead of an exact match.
- **Cache isolation.** Before-version lookups use a dedicated `MokaCache<ObjectKey, Bytes>` so that exact-match and before-version results cannot collide in the same cache entry.

fixes #11053 

End-to-end verification against a local BigTable emulator with objects seeded at versions 0, 1, 2, 5 for the same `ObjectID` (gaps at 3 and 4 intentional; the test also probes version 10 to cover requests above the highest stored version).

1.  Start a local BigTable emulator. See `crates/iota-kvstore/README.md` for setup.
1.  Run `iota-rest-kv` against the emulator with the following `config.yaml`:
```yaml
# Remote KV Store REST API config
#
# The Rest Api address
server-address: "0.0.0.0:3555"

instance-id: "iota"
column-family: "iota"
timeout-secs: 60
emulator-host: "localhost:8086"
```

```shell
$(gcloud beta emulators bigtable env-init)
cargo r -- --config config.yaml
```
3. Run the verification test to assert `before_version` behavior across the seeded versions and the intentional gaps.
```rust
/// Verifies that `multi_get_objects` with `before_version = true` returns, for
/// each requested `(object_id, version)`, the highest stored version strictly
/// less than the requested one, including across gaps in the stored version
/// sequence.
///
/// Assumes a local `iota-rest-kv` server is reachable and has versions
/// 0, 1, 2 and 5 seeded for `TEST_OBJECT_ID`.
#[tokio::test]
async fn get_objects_before_version() {
    std::env::set_var("BIGTABLE_EMULATOR_HOST", "localhost:8086");
    let mut client = BigTableClient::new_local("iota", "iota").await.unwrap();

    let id = ObjectID::random();
    let owner = Owner::AddressOwner(IotaAddress::random_for_testing_only());

    let objs = &[
        &Object::with_id_owner_version_for_testing(id, SequenceNumber::from_u64(0), owner),
        &Object::with_id_owner_version_for_testing(id, SequenceNumber::from_u64(1), owner),
        &Object::with_id_owner_version_for_testing(id, SequenceNumber::from_u64(2), owner),
        &Object::with_id_owner_version_for_testing(id, SequenceNumber::from_u64(5), owner),
    ];

    client.save_objects(objs).await.unwrap();

    const REST_KV_URL: &str = "http://127.0.0.1:3555";

    let client = HttpRestKVClient::new(
        REST_KV_URL,
        // max_concurrent_requests
        100,
        // multiget_batch_size
        100,
        // request_timeout_secs
        10,
        HistoricalFallbackClientMetrics::new(&prometheus::default_registry()),
    )
    .unwrap();

    // (requested_version, expected version returned, or None if no prior exists)
    let cases: &[(u64, Option<u64>)] = &[
        (0, None), // no version strictly < 0
        (1, Some(0)),
        (2, Some(1)),
        (5, Some(2)), // 3 and 4 are gaps
        (10, Some(5)),
    ];

    let requests: Vec<_> = cases
        .iter()
        .map(|(v, _)| (id, SequenceNumber::from_u64(*v)))
        .collect();

    let before_version = true;
    let results = client
        .multi_get_objects(&requests, before_version)
        .await
        .expect("multi_get_objects should succeed");

    assert_eq!(results.len(), cases.len());
    for ((requested, expected), got) in cases.iter().zip(results.iter()) {
        let got_version = got.as_ref().map(|obj| obj.version().value());
        assert_eq!(
            got_version, *expected,
            "before_version for requested v{requested}: expected {expected:?}, got {got_version:?}",
        );
    }
}

```

### Release Notes

- [x] Indexer: JSON-RPC supports fetching objects before a requested version from the fallback KV storage when the indexer database does not contain the requested version due to pruning.

### DIFF
--- a/crates/iota-indexer/src/historical_fallback/client.rs
+++ b/crates/iota-indexer/src/historical_fallback/client.rs
@@ -596,20 +596,15 @@ impl KeyValueStoreClient for HttpRestKVClient {
         object_refs: &[(ObjectID, SequenceNumber)],
         before_version: bool,
     ) -> IndexerResult<Vec<Option<Object>>> {
-        let fetches = if before_version {
-            let keys = object_refs
-                .iter()
-                .map(|(object_id, version)| ObjectKey(*object_id, *version))
-                .collect::<Vec<ObjectKey>>();
+        let keys = object_refs
+            .iter()
+            .map(|(object_id, version)| ObjectKey(*object_id, *version));
 
+        let fetches = if before_version {
+            let keys = keys.collect::<Vec<ObjectKey>>();
             self.multi_fetch_objects_before_version(&keys).await?
         } else {
-            let keys = object_refs
-                .iter()
-                .map(|(object_id, version)| Key::ObjectKey(ObjectKey(*object_id, *version)))
-                .collect::<Vec<_>>();
-
-            self.multi_fetch(keys).await?
+            self.multi_fetch(keys.map(Key::ObjectKey).collect()).await?
         };
 
         let objects = fetches
--- a/crates/iota-indexer/src/historical_fallback/client.rs
+++ b/crates/iota-indexer/src/historical_fallback/client.rs
@@ -6,7 +6,10 @@
 use std::time::Duration;
 
 use bytes::Bytes;
-use futures::stream::{self, StreamExt};
+use futures::{
+    TryStreamExt,
+    stream::{self, StreamExt},
+};
 use iota_storage::http_key_value_store::{ItemType, Key};
 use iota_types::{
     base_types::{ObjectID, SequenceNumber},
@@ -121,6 +124,7 @@ pub(crate) trait KeyValueStoreClient {
     async fn multi_get_objects(
         &self,
         object_refs: &[(ObjectID, SequenceNumber)],
+        before_version: bool,
     ) -> IndexerResult<Vec<Option<Object>>>;
 }
 
@@ -133,6 +137,13 @@ pub(crate) struct HttpRestKVClient {
     /// Maximum number of concurrent batch requests
     max_concurrent_batches: usize,
     cache: MokaCache<Key, Bytes>,
+    /// Cache for before-version lookups, keyed by `ObjectKey`.
+    ///
+    /// Kept separate from `cache` because the cached value is the object at
+    /// some earlier version, not the exact-version match that `cache` stores.
+    /// Mixing them under the same key would let before-version results serve
+    /// exact-version requests (or vice versa).
+    cache_object_before_version: MokaCache<ObjectKey, Bytes>,
     metrics: HistoricalFallbackClientMetrics,
 }
 
@@ -162,6 +173,10 @@ impl HttpRestKVClient {
             .time_to_idle(CACHE_TIME_TO_IDLE)
             .build();
 
+        let cache_object_before_version = MokaCacheBuilder::new(cache_size)
+            .time_to_idle(CACHE_TIME_TO_IDLE)
+            .build();
+
         Ok(Self {
             base_url,
             client,
@@ -169,6 +184,7 @@ impl HttpRestKVClient {
             max_concurrent_batches,
             metrics,
             cache,
+            cache_object_before_version,
         })
     }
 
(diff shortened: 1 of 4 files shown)


crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -262,6 +275,122 @@ impl HttpRestKVClient {
             IndexerError::Serde(format!("failed to deserialize multi_get response: {e:?}"))
         })
     }
+
+    async fn multi_fetch_objects_before_version(
```
tomxey, line 282: those 2 methods seem to be quite similar to `multi_fetch` and `fetch_batch`
did you consider to have some common code for them? to keep the code DRY

sergiupopescu199: I considered it. The two paths share the chunk → buffer → merge pattern, but they differ on cache type (`Key` vs `ObjectKey`), the metrics they record, the pre-validation, the request body shape (one path uses the keys as-is, the other rebuilds `Vec<Key>` from extracted `ObjectKey`s), and the URL. A shared helper would need 4-5 generic parameters and per-call closures for these differences, which I don't think reads better than the current duplication for two call sites. Open to revisiting if a third KV-fetch path lands later, since by then the right abstraction would be clearer.

tomxey (approved):

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -262,6 +278,123 @@ impl HttpRestKVClient {
...
+        &self,
+        keys: &[Key],
```
kodemartin: Why not use `keys: &[ObjectKey]` or `keys: impl Iterator<Item = &ObjectKey>` directly and avoid the fallback-input checks below? This is a method that is dedicated to fetching objects, so no need to make it more generic.

kodemartin (commented):
looks good with a minor suggestion

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11312/commits/e8f61e53e6df338116e62b4d4fe4fc35ea0d3db2

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -475,19 +594,29 @@ impl KeyValueStoreClient for HttpRestKVClient {
     async fn multi_get_objects(
         &self,
         object_refs: &[(ObjectID, SequenceNumber)],
+        before_version: bool,
     ) -> IndexerResult<Vec<Option<Object>>> {
-        let keys = object_refs
-            .iter()
-            .map(|(object_id, version)| Key::ObjectKey(ObjectKey(*object_id, *version)))
-            .collect::<Vec<_>>();
```
kodemartin, line 482: Nit: Build the `ObjectKey` iterator here and reusing in the logical branch below
```rust
        let keys = object_refs
            .iter()
            .map(|(object_id, version)| Key::ObjectKey(ObjectKey(*object_id, *version)));
```

kodemartin (approved):
lgtm

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11312/commits/2f0d1eb1954f2418f76684f27338f63f1249afee

PR #11313, opened 2026-04-24
chore(ci): update actions

muXxer:
# Description of change

I updated all actions and pinned them (manually checked) to their latest version.

`peaceiris/actions-gh-pages` still uses v20 unfortunately.

This will fix most of the node v20 deprecation warnings.

### DIFF
--- a/.github/actions/diffs/action.yml
+++ b/.github/actions/diffs/action.yml
@@ -35,7 +35,7 @@ outputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - name: Detect Changes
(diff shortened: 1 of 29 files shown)


thibault-martinez:
Are you sure that all these major bumps don't come with breaking changes?

muXxer:
> Are you sure that all these major bumps don't come with breaking changes?

image

thibault-martinez:
Would be nice to double check the actions involved in the release workflow so we don't have a surprise last minute

muXxer:
## Breaking change notes

Most major version bumps are Node.js runtime updates (v20->v24) with no API changes. Notable exceptions:

- **lycheeverse/lychee-action v2**: `fail` defaults to `true` now - but our workflow already sets `fail: false`
- **actions/setup-node v6**: Limits automatic caching to npm only (was broader before)
- **actions/checkout v6**: Persists credentials to a separate file under `$RUNNER_TEMP` instead of local git config

All changes should be backwards-compatible with the existing workflow configurations.

muXxer:
Seems like deployments were successful https://github.com/iotaledger/iota/actions/runs/24896294167

thibault-martinez (approved):

alexsporn (approved):

PR #11314, opened 2026-04-24
feat(graphql): migrate dynamic fields to backward diff

### LINKED ISSUE #11019: [graphql]: Refactor consistent views on dynamic fields

tomxey:
# Description of change

Switches `DynamicField::paginate` (and friends) from the forward-diff path (`objects_snapshot` + `objects_history`) to the backward-diff path (`checkpointed_objects` + `objects_backward_history`).

The PR converts version based lookup into time based lookup, due to issues with artificial versions that we create for e.g. tombstones. Time based lookup is not affected by artificial tombstone versions. 

Version-pinned dynamic-field queries use a finer time axis than the consistent-view path: `superseded_at_tx_sequence_number`. 
The boundary is resolved from `(parent, parent_version)` against `objects_backward_history`, then a "consistent view at that tx_seq" query is executed — same Source-A / Source-B shape and `MIN(object_version)` aggregation as `consistent::query`, just pivoting on the tx-seq column instead of the cp column.

The no-version-pin path is unchanged — it falls through to the existing consistent-view query at the requested checkpoint.

The dedicated `View` enum and `build_objects_query` in `consistency.rs` are removed; they were only used by the now-replaced version-pinned DF path.

## Why finer-than-checkpoint granularity

A version-pinned query asks "what was the state at this `parent_version`?", but multiple transactions in the same checkpoint can produce multiple versions of the same object. 
The `superseded_at_checkpoint` axis collapses all of them onto the same value and can't distinguish between them; the `superseded_at_tx_sequence_number` axis orders them.

## Notable design choices

- **NotYetCreated rows are anchored at `lamport_create - 1`**, not `-1`.
 Since it turned out to be possible to have multiple not yet created entries for single object (in case it is deleted and then recreated), we need to have such entries properly ordered, not always put at the beginning with hardcoded `-1`
- **Resolution uses the *latest* boundary** (last tx in which the parent was still at the requested version), not the produced-at boundary. It guarantees that we always try to serve data that we actually have in the DB, trying to serve data from the first tx in which parent was at requested version could make us read already pruned data.
- **`Object` carries an optional `root_version_tx_seq`** propagated through `try_from_stored_history_object` and `ObjectLookup::UnderParent`. 
  The outermost `DynamicField::paginate` resolves it once and stamps it on result objects; nested traversals reuse it without per-level re-translation.
- **Six new indexes** on `objects_backward_history` mirror the existing
  cp-axis ones on the `superseded_at_tx_sequence_number` column —
  `tx_owner`, `tx_owner_package_module_name_full_type`,
  `tx_package_module_name_full_type`, `tx_type`, `tx_coin_only`,
  `tx_coin_owner` — supporting fast lookups for `WHERE
  superseded_at_tx_sequence_number > :tx_n` scans with the same filter shapes the cp-axis queries already use.

## Visible behaviour change

Under the latest-boundary semantics, direct queries of the form `object(Child, version: V).dynamicFields` where `V` spans multiple transactions (Move's `borrow_mut` quirk: `add_nested_child(parent, ...)` mutates a child's DOF tree without bumping the child's own version) now surface in-version-bracket DOF additions. 
Two test snapshots are re-recorded to reflect this: `nested_dof.move` and `immutable_dof.move`. 
Outer `object(Parent, v=V) → dynamicFields → ... → dynamicFields` traversals still pin to the outer's resolution and behave as before.

fixes https://github.com/iotaledger/iota/issues/11019

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [x] GraphQL: `object(addr, version: V).dynamicFields` now includes dynamic-field changes that occurred while the object was at version `V` (e.g. nested dynamic fields added without bumping the parent's version).
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -0,0 +1,102 @@
...
+//!
+//! The caller resolves `parent_version → root_version_tx_seq` (the largest
+//! tx_sequence_number at which the parent was still at the requested version)
+//! and dispatches to [`query`]. The implementation mirrors
+//! [`super::consistent::query`] but pivots on
+//! `superseded_at_tx_sequence_number` instead of `superseded_at_checkpoint` —
+//! a checkpoint can contain multiple versions of the same object, and the
+//! cp-axis cannot order them.
+
+use super::{CHECKPOINTED_COLUMNS, HISTORY_COLUMNS, NOT_YET_CREATED, merge_and_deduplicate};
```
sergiupopescu199, line 15: Same applies based on this [comment](https://github.com/iotaledger/iota/pull/11314/changes#r3195215581)

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Version-pinned dynamic-field queries: consistent view at a tx-sequence
+//! boundary corresponding to the queried `parent_version`.
+//!
+//! The caller resolves `parent_version → root_version_tx_seq` (the largest
+//! tx_sequence_number at which the parent was still at the requested version)
+//! and dispatches to [`query`]. The implementation mirrors
+//! [`super::consistent::query`] but pivots on
```
sergiupopescu199, line 10: To be robust for future refactors it would be better to use absolute imports  `crate::` instead of `super::`.

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -0,0 +1,102 @@
...
+    merge_and_deduplicate(vec![
+        consistent_checkpointed_objects(root_version_tx_seq, page, &filter_fn),
+        consistent_historical_objects(root_version_tx_seq, page, &filter_fn),
+    ])
+}
+
+/// Source A: rows from `checkpointed_objects` whose current state was already
+/// in effect at `root_version_tx_seq` — i.e. no entry exists in
+/// `objects_backward_history` for that object with
+/// `superseded_at_tx_sequence_number > root_version_tx_seq`.
```
sergiupopescu199, line 43: Nit: could we expand a bit the rust doc to be more "novice" friendly?

```suggestion
/// Builds a raw SQL query to fetch rows from `checkpointed_objects` that represent 
/// the state of objects at a specific point in time (`root_version_tx_seq`).
///
/// An object state is considered "in effect" at the root version if no subsequent 
/// update exists in the history. 
///
/// The query selects objects from the current state (checkpointed_objects) and 
/// excludes any  that were modified later (objects_backward_history), 
/// by checking for newer entries  in the history table
/// (`superseded_at_tx_sequence_number > root_version_tx_seq`).
```

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -0,0 +1,102 @@
...
+    );
+    source = filter!(source, "changed.object_id IS NULL");
+    page.apply::<StoredBackwardObject>(source)
+}
+
+/// Source B: rows from `objects_backward_history` giving each object's state
+/// at `root_version_tx_seq` — the earliest version superseded after that
+/// tx_seq (`MIN(object_version)`). Excludes `NotYetCreated` so objects that
+/// didn't exist at `root_version_tx_seq` aren't returned. Mirrors
+/// [`super::consistent::consistent_historical_objects`] on the tx_seq axis.
```
sergiupopescu199, line 71: Nit: the same ad in this [comment](https://github.com/iotaledger/iota/pull/11314/changes#r3195300710) plus to use `crate::` instead of `super::`

crates/iota-graphql-rpc/src/types/dynamic_field.rs
```diff
@@ -201,17 +210,42 @@ impl DynamicField {
...
+                    // already resolved it (nested dynamic-field chains do
+                    // this); otherwise lazy-resolve `(parent, pv) → tx_seq`
+                    // here.
+                    let tx_seq = match root_version_tx_seq {
+                        Some(t) => t,
+                        None => match resolve_root_version_tx_seq(conn, parent, pv)? {
+                            Some(t) => t,
+                            None => return Ok(None),
+                        },
+                    };
```
sergiupopescu199, line 233: Nit: we could use if let chains.

```rust
 let tx_seq = if let Some(t) = root_version_tx_seq {
        t
    } else if let Some(t) = resolve_root_version_tx_seq(conn, parent, pv)? {
        t
    } else {
        return Ok(None);
    };
```

crates/iota-graphql-rpc/src/types/move_object.rs
```diff
@@ -496,6 +514,12 @@ impl MoveObject {
     pub(crate) fn root_version(&self) -> u64 {
         self.super_.root_version()
     }
+
+    /// Tx-sequence-number anchor of the outermost root, propagated through
+    /// nested dynamic-field traversals. See [`Object::root_version_tx_seq`].
```
sergiupopescu199, line 519: Nit: we could improve the rust doc.

```suggestion
/// Returns the transaction sequence number anchor of the outermost root.
///
/// This value is propagated through nested dynamic-field traversals to maintain
/// temporal consistency. For more details, see [`Object::root_version_tx_seq`].
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -84,6 +84,13 @@ pub(crate) struct Object {
     /// that were loaded but not actually mutated don't end up having
     /// their versions updated.
     root_version: u64,
+    /// Tx-sequence-number anchor of the outermost root, resolved once at the
+    /// outer `under_parent` lookup and propagated through nested dynamic-field
+    /// traversals. When set, the version-pinned dynamic-field reader uses it
+    /// directly as the at-tx cutoff instead of re-translating the (per-level)
+    /// `root_version`. `None` when this object wasn't constructed via a
+    /// `under_parent` lookup (e.g. `latest_at`, `at_version`).
```
sergiupopescu199, line 92: Nit: we could improve the rust doc to be more novice friendly?

```suggestion
/// The transaction sequence number used as a consistent temporal anchor for 
/// dynamic field lookups.
///
/// This anchor is resolved once at the top-level `under_parent` lookup and 
/// propagated through nested traversals.  Ensures all nested dynamic fields 
/// are read from the same point in time, even if the database is being updated 
/// during the traversal. Allows the reader to use this cutoff directly, bypassing 
/// the need to re-translate the `root_version` at every level.
///
/// Represented as `None` if the object was not constructed via `under_parent` 
/// (e.g., when using `latest_at` or `at_version`).
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -214,6 +221,11 @@ pub(crate) enum ObjectLookup {
         /// for the latest version of a child object whose version is
         /// less than or equal to this upper bound.
         parent_version: u64,
+        /// Tx-sequence-number anchor of the outermost root, stamped onto
+        /// the resolved child `Object` so nested dynamic-field queries reuse
+        /// it instead of re-translating per level. `None` when the caller
+        /// hasn't resolved one.
```
sergiupopescu199, line 227: Nit: 

```suggestion
/// The transaction sequence number anchor of the outermost root.
///
/// This is "stamped" onto the resolved child [`Object`] so that 
/// nested dynamic-field  queries can reuse the anchor instead 
/// of re-translating it at every level. 
///
/// Represented as `None` if the caller has not yet resolved a root anchor.
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -823,6 +854,21 @@ impl Object {
         self.root_version
     }
 
+    /// Tx-sequence-number anchor for the outermost root (set when the object
+    /// chain originates from a `under_parent` lookup). Propagated through
+    /// nested dynamic-field traversals.
```
sergiupopescu199, line 859: Nit:

```suggestion
/// Returns the transaction sequence number anchor for the outermost root.
///
/// This anchor is set when the object chain originates from an `under_parent` 
/// lookup and is propagated through nested dynamic-field traversals. This 
/// ensures all subsequent lookups in the chain remain pinned to the same 
/// point in time.
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -823,6 +854,21 @@ impl Object {
...
+    pub(crate) fn root_version_tx_seq(&self) -> Option<u64> {
+        self.root_version_tx_seq
+    }
+
+    /// Stamps the `root_version_tx_seq` anchor on this object. Used by
+    /// `Object::query` to propagate the caller's anchor onto a freshly-
+    /// loaded object without baking the anchor into the loader cache key.
```
sergiupopescu199, line 866: Nit:

```suggestion
/// Stamps the `root_version_tx_seq` anchor on this object.
///
/// Used by [`Object::query`] to propagate the caller's anchor onto a freshly-
/// loaded object without baking the anchor into the loader cache key.
```

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -917,9 +967,18 @@ impl Object {
 
     /// Look-up the latest version of an object whose version is less than or
     /// equal to its parent's version, as of a given checkpoint.
-    pub(crate) fn under_parent(parent_version: u64, checkpoint_viewed_at: u64) -> ObjectLookup {
+    /// `root_version_tx_seq` is the outermost root's tx-sequence-number
+    /// anchor (or `None` when not yet resolved); it is stamped on the
+    /// resulting object so that nested dynamic-field traversals can reuse
+    /// it.
```
sergiupopescu199, line 973: Nit:

```suggestion
    ///
    /// The `root_version_tx_seq` acts as a temporal anchor from the 
    /// outermost root. If provided, it is "stamped" onto the resulting 
    /// object, allowing nested dynamic-field traversals to reuse the 
    /// anchor instead of re-resolving it.
```

sergiupopescu199 (approved):
LGTM 🚀

Left some nits, mostly related to Rust docs since I'm not very familiar with all the technical details of the implementation I thought that making them more "novice" friendly could add some value if we would touch the code after some months.

PR #11326, opened 2026-04-27
feat!: Replace `SequenceNumber` with SDK type

DaughterOfMars:

This PR replaces `SequenceNumber` (previously defined in `iota-types`) with the
original name `SequenceNumber`, so most existing imports continue to work
unchanged.

## What changed

- The local `pub struct SequenceNumber(u64)` definition and its inherent impls
  in [crates/iota-types/src/base_types.rs](crates/iota-types/src/base_types.rs)
  are removed. `SequenceNumber` is now a re-export:
  `pub use iota_sdk_types::Version as SequenceNumber;`

## Call-site adjustments

Because the SDK `Version` has a slightly different surface than the old
`SequenceNumber`, a few call sites had to be updated rather than relying on the
re-export:

- `lamport_increment`, `next`, `increment`, `decrement`, and
  `new_congested_with_suggested_gas_price` are now fallible on the SDK type;
  call sites add `.unwrap()` (in places where the previous code already
  asserted validity, e.g.
  [shared_object_version_manager.rs](crates/iota-core/src/authority/shared_object_version_manager.rs),
  [test_effects_builder.rs](crates/iota-types/src/effects/test_effects_builder.rs),
  [move_package.rs](crates/iota-types/src/move_package.rs)).
- `SequenceNumber::new()` is replaced by `SequenceNumber::default()`.
- The `increment_to` / `decrement_to` helpers on `MoveObject` are inlined into
  `increment_version_to` / `decrement_version_to` in
  [object.rs](crates/iota-types/src/object.rs), preserving the previous debug
  assertions.
- `Display` output changes from `SequenceNumber(N)` to `Version(N)`; one
  string-literal test in
  [object.rs](crates/iota-types/src/object.rs) is updated accordingly.
- A handful of `format!`/`Display` call sites that used `u64::from(*version)`
  or `version.value()` now print the `Version` directly (or use
  `version.as_u64()` where a numeric is genuinely required, as in
  [object_changes.rs](crates/iota-json-rpc-types/src/object_changes.rs)).

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: Sequence numbers are now displayed in decimal format in error messages
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Some methods have been removed from `SequenceNumber` due to
  crate externalization (`iota-sdk-types::Version` is replacing
  `iota-types::SequenceNumber`). Constructors and arithmetic helpers
  (`next`, `increment`, `decrement`, `lamport_increment`,
  `new_congested_with_suggested_gas_price`) are now fallible on the SDK type.
- [ ] gRPC:


crates/iota-types/src/move_package.rs
```diff
@@ -357,7 +357,7 @@ impl MovePackage {
         let runtime_id = ObjectID::from(*module.address());
         let type_origin_table = build_upgraded_type_origin_table(self, modules, storage_id)?;
         let mut new_version = self.version();
-        new_version.increment();
+        new_version.increment().unwrap();
```
valeriyr, line 360: I would suggest using `expect` in this file (decrement_version, increment_version) instead of `unwrap` to indicate that it is an invariant, to make it clearer

crates/iota-types/src/transaction.rs
```diff
@@ -3625,7 +3625,7 @@ impl InputObjects {
             })
             .chain(receiving_objects.iter().map(|object_ref| object_ref.1));
 
-        SequenceNumber::lamport_increment(input_versions)
+        SequenceNumber::lamport_increment(input_versions).unwrap()
```
valeriyr, line 3628: and here

valeriyr (approved):

crates/iota-types/src/move_package.rs
thibault-martinez: Maybe more generally use `expect` in all non-tests files?

crates/iota-core/tests/staged/iota.yaml
jkrvivian:
```suggestion
            - U64
```

piotrm50 (approved):

crates/iota-types/src/move_package.rs
Alex6323: Maybe we should have have unchecked versions of these methods and use those instead if we can be sure from the context that there can't be any under/overflows?

jkrvivian (changes requested):

crates/iota-types/src/move_package.rs
thibault-martinez: We probably shouldn't clutter the SDK with fallible AND infallible versions, the decision on what to do lies with the user

crates/iota-types/src/object.rs
```diff
@@ -1216,7 +1226,7 @@ mod tests {
         let objref = format!("{:?}", o.compute_object_reference());
         assert_eq!(
             objref,
-            "(0x0000000000000000000000000000000000000000000000000000000000000000, SequenceNumber(1), o#Ba4YyVBcpc9jgX4PMLRoyt9dKLftYVSDvuKbtMr9f4NM)"
+            "(0x0000000000000000000000000000000000000000000000000000000000000000, Version(1), o#Ba4YyVBcpc9jgX4PMLRoyt9dKLftYVSDvuKbtMr9f4NM)"
```
Alex6323, line 1229: hmm, maybe this should be changed to just show `1`?

kodemartin (approved):
lgtm for the infra crates, with the suggestion to add a release note for `JSON-RPC` that the version is now displayed in decimal format in error messages.

crates/iota-types/src/move_package.rs
Alex6323: fine with me

crates/iota-types/src/object.rs
DaughterOfMars: It's a debug impl so it's good to show this stuff

lollobene (commented):
LGTM (vm-lang files)

Cargo.lock
lollobene, line 1: Why so many modifications here? The Cargo.toml doesn't have any updates

DaughterOfMars: image
It is a mystery

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -1058,7 +1058,8 @@ mod checked {
                         ExecutionErrorKind::ExecutionCancelledDueToSharedObjectCongestionV2 {
                             congested_objects: CongestedObjects(cancelled_objects),
                             suggested_gas_price: version
-                                .get_congested_version_suggested_gas_price(),
+                                .get_congested_version_suggested_gas_price()
+                                .unwrap(),
```
thibault-martinez, line 1062: expect?

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -223,7 +223,7 @@ impl AuthorityStorePruner {
             for ObjectKey(object_id, seq_number) in object_tombstones_to_prune {
                 for result in perpetual_db.objects.safe_iter_with_bounds(
                     Some(ObjectKey(object_id, VersionNumber::MIN_VALID_INCL)),
-                    Some(ObjectKey(object_id, seq_number.next())),
+                    Some(ObjectKey(object_id, seq_number.next().unwrap())),
```
thibault-martinez, line 226: expect?

crates/iota-core/src/authority/authority_store_tables.rs
```diff
@@ -427,7 +427,7 @@ impl AuthorityPerpetualTables {
     ) -> IotaResult<Vec<ObjectKey>> {
         let mut objects = vec![];
         for result in self.objects.safe_iter_with_bounds(
-            Some(ObjectKey(object.0, object.1.next())),
+            Some(ObjectKey(object.0, object.1.next().unwrap())),
```
thibault-martinez, line 430: expect?

crates/iota-core/src/authority/shared_object_version_manager.rs
```diff
@@ -196,7 +197,7 @@ impl SharedObjVerManager {
         }
 
         let next_version =
-            SequenceNumber::lamport_increment(input_object_keys.iter().map(|obj| obj.1));
+            SequenceNumber::lamport_increment(input_object_keys.iter().map(|obj| obj.1)).unwrap();
```
thibault-martinez, line 200: expect?

Cargo.lock
thibault-martinez: If I reset it to develop and generate again, I think it reverts the unnecessary changes

thibault-martinez: Or maybe they were needed by the tokio change of @Thoralf-M

crates/iota-core/src/authority/shared_object_version_manager.rs
DaughterOfMars: I mean, these are results with defined error messages. Not sure what it would improve

thibault-martinez: What do you mean `defined error messages`?

DaughterOfMars: I mean they have error messages so why would we add another generic error message to the expect?

thibault-martinez: You did it in your previous commit, what's so different now?

DaughterOfMars: Nothing, I just didn't think about it until now

thibault-martinez: Ok well, I'd maybe revert so the messages are consistent... Maybe ask Valerii on Slack if that's fine

DaughterOfMars: done

jkrvivian (approved):

thibault-martinez (approved):

crates/iota-types/src/move_package.rs
thibault-martinez: We've kept the unwrap because there are proper error messages in the SDK already

Cargo.lock
Thoralf-M: Tokio downgrade was only +5 -5

Thoralf-M: Only very few modifications left after this https://github.com/iotaledger/iota/pull/11326/commits/ba43d7f0ff8a68501df45e7b3f722ee283f5bf5e

Dkwcs:
Just to clarify, do we have compatibility tests covering cases like the following:
```
assert_eq!(serde_json::to_string(&Version::from_u64(42)).unwrap(), "42");
assert_eq!(serde_json::from_str::<Version>("42").unwrap().as_u64(), 42);
assert_eq!(serde_json::from_str::<Version>("\"42\"").unwrap().as_u64(), 42);
```
Since we have crossed repository boundaries, the highest-risk breakage is not compilation, but wire-format or schema drift.

DaughterOfMars:
> Just to clarify, do we have compatibility tests covering cases like the following:
> 
> ```
> assert_eq!(serde_json::to_string(&Version::from_u64(42)).unwrap(), "42");
> assert_eq!(serde_json::from_str::<Version>("42").unwrap().as_u64(), 42);
> assert_eq!(serde_json::from_str::<Version>("\"42\"").unwrap().as_u64(), 42);
> ```
> 
> Since we have crossed repository boundaries, the highest-risk breakage is not compilation, but wire-format or schema drift.

We have implemented BCS schema checks across all types but we are dropping json support

PR #11330, opened 2026-04-27
feat!: Replace `IotaAddress` with SDK `Address`

DaughterOfMars:

This PR replaces `IotaAddress` (previously defined in `iota-types`) with the
original name `IotaAddress`, so most existing imports continue to work
unchanged.

## What changed

- The local `pub struct IotaAddress([u8; IOTA_ADDRESS_LENGTH])` definition
  and its inherent impls in
  [crates/iota-types/src/base_types.rs](crates/iota-types/src/base_types.rs)
  are removed. `IotaAddress` is now a re-export:
  `pub use iota_sdk_types::Address as IotaAddress;`

## Call-site adjustments

Because the SDK `Address` has a slightly different surface than the old
`IotaAddress`, a few call sites had to be updated rather than relying on the
re-export:

- `IotaAddress::random_for_testing_only()` is replaced by the SDK's
  `IotaAddress::random()` everywhere (tests, fuzzers, sample data builders).
- `IotaAddress::default()` is replaced by `IotaAddress::ZERO` in
  [effects_v1.rs](crates/iota-types/src/effects/effects_v1.rs); the SDK type
  no longer derives `Default`, but exposes the same zero-value via the
  `ZERO` associated constant.
- `IotaAddress::try_from(bytes)` / `IotaAddress::try_from(&bytes[..])` are
  replaced by `IotaAddress::from_bytes(...)`. The SDK signature is
  `from_bytes(&[u8])` rather than `from_bytes<T: AsRef<[u8]>>`, so a few call
  sites pass `&vec` / `&arr` explicitly (e.g.
  [iota-json/src/lib.rs](crates/iota-json/src/lib.rs),
  [base_types_tests.rs](crates/iota-types/src/unit_tests/base_types_tests.rs)).
- The blanket `impl<T: IotaPublicKey> From<&T> for IotaAddress` is removed
  (the SDK type cannot host downstream-trait impls). It is replaced by a
  free function `address_from_iota_pub_key` in `iota-types::base_types`,
  used from
  [crypto.rs](crates/iota-types/src/crypto.rs) and
  [iota-keys/src/key_derive.rs](crates/iota-keys/src/key_derive.rs) where
  the previous `kp.public().into()` shorthand was used.
- `From<IotaAddress> for AccountAddress` and the `AsRef<[u8]>` impl are
  removed. Conversions to `move_core_types::AccountAddress` are now spelled
  `AccountAddress::new(addr.into_bytes())` (and the reverse uses
  `IotaAddress::new(account_addr.into_bytes())` /
  `IotaAddress::from(account_addr.into_bytes())`). Affected files include
  [base_types.rs (`TxContext`)](crates/iota-types/src/base_types.rs),
  [iota-json/src/lib.rs](crates/iota-json/src/lib.rs),
  [iota_sdk_types_conversions.rs](crates/iota-types/src/iota_sdk_types_conversions.rs),
  and the Move natives crates under
  [external-crates/move/crates/iota-framework/latest/iota-move-natives](external-crates/move/crates/iota-framework/latest/iota-move-natives/).
- `addr.0` field access and `addr.to_inner()` are replaced by
  `addr.into_bytes()` (or `addr.as_bytes()` where a borrow is enough); the
  inner byte array is no longer publicly accessible as a tuple field.
- `Debug` output changes from the old custom `0x...` form to the SDK's
  derived form; one assertion in
  [base_types_tests.rs](crates/iota-types/src/unit_tests/base_types_tests.rs)
  is updated to use `Display` (`{id}`) instead of `Debug` (`{id:?}`).
- The `proptest_derive::Arbitrary` derive that gated on the local
  `fuzzing` feature is dropped; the SDK type provides its own `Arbitrary`
  impl, so the `fuzzing` feature on `iota-types` no longer needs to forward
  it for `IotaAddress`.

A large number of call sites are touched purely to rename
`random_for_testing_only` -> `random` and to swap a handful of `from`/`into`
hops for `into_bytes()` / `IotaAddress::new(...)`; those changes are mechanical
and don't alter behavior.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Some methods have been removed from `IotaAddress` due to
  crate externalization (`iota-sdk-types::Address` is replacing
  `iota-types::IotaAddress`). Notably, `random_for_testing_only` is renamed
  to `random`, `to_inner` / tuple-field access is replaced by `into_bytes`,
  `try_from(&[u8])` / `try_from(Vec<u8>)` are replaced by
  `from_bytes(&[u8])`, the blanket `From<&IotaPublicKey>` impl is replaced
  by the `address_from_iota_pub_key` free function, and `Default` is no
  longer derived (use `IotaAddress::ZERO`).
- [ ] gRPC:


crates/iota-adapter-transactional-tests/tests/deny_list_v1/coin_deny_and_undeny_receiver.snap
```diff
@@ -40,8 +40,8 @@ Epoch advanced: 1
 
 task 6, lines 50-52:
 //# run iota::pay::split_and_transfer --args object(1,0) 1 @B --type-args test::regulated_coin::REGULATED_COIN --sender A
-Error: Transaction Effects Status: Address B is denied for coin test::regulated_coin::REGULATED_COIN
-Execution Error: ExecutionError: ExecutionError { inner: ExecutionErrorInner { kind: AddressDeniedForCoin { address: B, coin_type: "test::regulated_coin::REGULATED_COIN" }, source: None, command: None } }
+Error: Transaction Effects Status: Address Address("B") is denied for coin test::regulated_coin::REGULATED_COIN
```
valeriyr: This message looks worse than before. As I see, the Display implementation has changed. Is this expected?

jkrvivian (approved):

piotrm50 (approved):

crates/iota-adapter-transactional-tests/tests/deny_list_v1/coin_deny_and_undeny_receiver.snap
DaughterOfMars: This is the debug impl. We intend to ultimately change all of these messages to use Display which will revert these changes but this is widespread

thibault-martinez: Can you remind why we're not already using Display in this PR?

DaughterOfMars: It's widespread

thibault-martinez: But we are already touching the snapshots?

DaughterOfMars: Do you want me to spend the rest of the day hunting down all of these locations and bloat the PR with hundreds more snapshot changes?

thibault-martinez: Then if that's fine with @valeriyr, we open an issue to deal with this later

kodemartin (changes requested):
looks good with the suggestion to preserve log/span shape after the `Debug` change this patch introduces. More specifically

- Replace `debug!(?owner, ...) → debug!(%owner, ...)` at `coin_api.rs:181` and `:197`.
- Replace `#[instrument(skip(self))] → #[instrument(skip(self, owner), fields(owner = %owner))]` (or `address` per signature) at 
```
- coin_api.rs:94 get_coins(owner)                                                                                                                                                                                                                         
- coin_api.rs:123 get_all_coins(owner)
- coin_api.rs:169 get_balance(owner)                                                                                                                                                                                                                      
- coin_api.rs:194 get_all_balances(owner)
- indexer_api.rs:201 get_owned_objects(address)                                                                                                                                                                                                           
- indexer_api.rs:567 iota_names_reverse_lookup(address)                                                                                                                                                                                                   
- indexer_api.rs:597 iota_names_find_all_registration_nfts(address)
- governance_api.rs:377 get_stakes(owner)                                                                                                                                                                                                                 
- governance_api.rs:392 get_timelocked_stakes(owner)
```

crates/iota-adapter-transactional-tests/tests/deny_list_v1/coin_deny_and_undeny_receiver.snap
valeriyr: yes, I am good

valeriyr (approved):

crates/iota-core/src/unit_tests/batch_transaction_tests.rs
```diff
@@ -52,7 +52,9 @@ async fn test_batch_transaction_ok() -> anyhow::Result<()> {
                 vec![],
                 vec![
                     CallArg::Pure(16u64.to_le_bytes().to_vec()),
-                    CallArg::Pure(bcs::to_bytes(&AccountAddress::from(sender)).unwrap()),
+                    CallArg::Pure(
+                        bcs::to_bytes(&AccountAddress::new(sender.into_bytes())).unwrap(),
```
Thoralf-M, line 56: Could be also just `sender.into_bytes().to_vec()` I think, same below

crates/iota-core/src/test_utils.rs
```diff
@@ -230,7 +230,7 @@ pub fn make_transfer_object_move_transaction(
 ) -> Transaction {
     let args = vec![
         CallArg::Object(ObjectArg::ImmOrOwnedObject(object_ref)),
-        CallArg::Pure(bcs::to_bytes(&AccountAddress::from(dest)).unwrap()),
+        CallArg::Pure(bcs::to_bytes(&AccountAddress::new(dest.into_bytes())).unwrap()),
```
Thoralf-M, line 233: Also here just dest.into_bytes().to_vec()

kodemartin (approved):

crates/iota-core/src/unit_tests/batch_transaction_tests.rs
DaughterOfMars: I'd rather keep using the accountaddress impl to avoid possible future incompatabilities

Thoralf-M: Sure, not a big issue anyways

Thoralf-M (approved):

Alex6323 (approved):

thibault-martinez (approved):

semenov-vladyslav (approved):

PR #11331, opened 2026-04-27
feat: Replace `ObjectID` with the SDK type

DaughterOfMars:

This PR replaces `ObjectID` (previously defined in `iota-types`) with the
original name `ObjectID`, so most existing imports continue to work
unchanged.

## Description of change

- `iota-types::base_types::ObjectID` is now a re-export of `iota_sdk_types::ObjectId`; the local definition (and its `impl` block, manual `Serialize`/`Deserialize`, `FromStr`, `From<AccountAddress>`, etc.) has been removed.
- Updated call sites that depended on idiosyncrasies of the old `ObjectID` (e.g. `derive_id` now takes the digest as `Into<...>`, `TxContext::sender` constructs an `IotaAddress` directly from bytes rather than going through `ObjectID`, removed the now-unused `From<ObjectID> for IotaAddress` impl).
- Adjusted snapshot/test fixtures that rendered the type to match the SDK type's display/serialization.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): `Debug` format of `ObjectId` was changed from `0x...` to `ObjectId("0x...")`
- [ ] Indexer:
- [x] JSON-RPC: `Debug` format of `ObjectId` was changed from `0x...` to `ObjectId("0x...")`
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: Some methods have been removed from `ObjectID` due to crate externalization (`iota-rust-sdk::ObjectId` is replacing `iota-types::ObjectID`).
- [ ] gRPC:


crates/iota-json-rpc/src/error.rs
```diff
@@ -494,7 +494,7 @@ mod tests {
             let expected_code = expect!["-32002"];
             expected_code.assert_eq(&error_object.code().to_string());
             let expected_message = expect![
-                "Transaction execution failed due to issues with transaction inputs, please review the errors and try again:\n- Could not find the referenced object 0x0000000000000000000000000000000000000000000000000000000000000000 at version None"
+                "Transaction execution failed due to issues with transaction inputs, please review the errors and try again:\n- Could not find the referenced object ObjectId(\"0x0000000000000000000000000000000000000000000000000000000000000000\") at version None"
```
valeriyr, line 497: The same question as for `IotaAddress`.  Do we expect to see the object type in user messages?

DaughterOfMars: Same answer

crates/iota-types/src/unit_tests/messages_tests.rs
```diff
@@ -1275,7 +1278,7 @@ fn test_certificate_digest() {
     let (sender2, sender2_sec): (_, AccountKeyPair) = get_key_pair();
 
     let gas_price = 10;
-    let make_tx = |sender, sender_sec| {
+    let make_tx = |sender, sender_sec: Ed25519KeyPair| {
```
Thoralf-M, line 1281: Probably not needed?
```suggestion
    let make_tx = |sender, sender_sec| {
```

DaughterOfMars: It actually is :c

Thoralf-M (approved):

valeriyr (approved):

crates/iota-core/tests/staged/iota.yaml
jkrvivian:
```suggestion
                TYPENAME: ObjectId
```

crates/iota-json-rpc/src/error.rs
kodemartin: This is caused by `UserInputError::ObjectNotFound` (`iota_types::error`) using debug formatting to generate the error message. 

Should be refactored but you can create a dedicated issue for that.

kodemartin (approved):
lgtm for the infra crates, with the suggestion about using `Display` format where `Debug` is currently used.

crates/iota-replay/src/transaction_provider.rs
```diff
@@ -61,7 +61,7 @@ impl FromStr for FuzzStartPoint {
                         "{} is not a valid checkpoint (err: {:?}) or transaction digest (err: {:?})",
                         s, u64_err, tx_err
                     );
-                    Err(tx_err)
+                    anyhow::bail!(tx_err);
```
jkrvivian: I'd prefer the original code, it's already the tail expression of the match, so no early return with bail! is needed. 👀

jkrvivian (approved):

piotrm50 (approved):

Alex6323 (approved):

crates/iota-json-rpc/src/error.rs
thibault-martinez: https://github.com/iotaledger/iota/issues/11334

crates/iota-core/src/unit_tests/authority_tests.rs
```diff
@@ -5686,7 +5690,7 @@ async fn test_function_not_found() {
     let mut builder = ProgrammableTransactionBuilder::new();
     builder
         .move_call(
-            ObjectID::from_single_byte(1),
+            Address::STD.into(),
```
thibault-martinez: Could use `ObjectID::STD_PACKAGE` here, same L5749.
Only commenting this nit because we'll need to trigger CI anyway

docs/examples/rust/stardust/address-unlock-condition.rs
```diff
@@ -154,7 +154,7 @@ async fn main() -> Result<(), anyhow::Error> {
 
             // Extract the IOTA balance.
             let iota_coin = builder.programmable_move_call(
-                IOTA_FRAMEWORK_ADDRESS.into(),
+                (*IOTA_FRAMEWORK_ADDRESS).into(),
```
thibault-martinez: Here and following, use the ID constant

crates/iota/tests/snapshots/ptb_files_tests__move_call_fun_resolution.ptb.snap
```diff
@@ -36,7 +36,7 @@ expression: "results.join(\"\\n\")"
  3 │ --assign zero_x_zero @0x0
  4 │ --move-call zero_x_zero::option::is_none<u64> a
    ·             ─────┬─────
-   ·                  ╰── Object 0x0000000000000000000000000000000000000000000000000000000000000000 does not exist
+   ·                  ╰── Object ObjectId("0x0000000000000000000000000000000000000000000000000000000000000000") does not exist
```
thibault-martinez, line 39: Is this a case of being too annoying to change in this PR?

docs/examples/rust/stardust/address-unlock-condition.rs
```diff
@@ -182,7 +182,7 @@ async fn main() -> Result<(), anyhow::Error> {
             ];
 
             let nft_output = builder.programmable_move_call(
-                STARDUST_ADDRESS.into(),
+                (*STARDUST_ADDRESS).into(),
```
semenov-vladyslav:
```suggestion
                STARDUST_PACKAGE_ID,
```
Here and below, also for `IOTA_FRAMEWORK_ADDRESS`.

semenov-vladyslav (approved):

crates/iota/tests/snapshots/ptb_files_tests__move_call_fun_resolution.ptb.snap
DaughterOfMars: yes

docs/examples/rust/stardust/address-unlock-condition.rs
```diff
@@ -215,7 +215,7 @@ async fn main() -> Result<(), anyhow::Error> {
 
                 // Extract the IOTA balance.
                 let iota_coin = builder.programmable_move_call(
-                    IOTA_FRAMEWORK_ADDRESS.into(),
+                    (*IOTA_FRAMEWORK_ADDRESS).into(),
```
thibault-martinez: You missed two of them

crates/iota/tests/snapshots/ptb_files_tests__move_call_fun_resolution.ptb.snap
thibault-martinez: I don't think it is but I opened an issue for later https://github.com/iotaledger/iota/issues/11344

roman1e2f5p8s (approved):
besides the already mentioned `{:?}` -> `{}` cleanup, the PR description should explicitly mention that the `Debug` format was changed from `0x...` to `ObjectId("0x...")` - even better to mention it in the release notes section, ticking Nodes and JSON-RPC

thibault-martinez (approved):

thibault-martinez:
> besides the already mentioned `{:?}` -> `{}` cleanup, the PR description should explicitly mention that the `Debug` format was changed from `0x...` to `ObjectId("0x...")` - even better to mention it in the release notes section, ticking Nodes and JSON-RPC

Issues were opened for the cleanup and mentions added to the release notes

PR #11337, opened 2026-04-27
feat(iota-types)!: replace GasCostSummary with SDK type

Thoralf-M:

This PR replaces `GasCostSummary` (previously defined in `iota-types::gas`)
with the SDK version from `iota-sdk-types`. To keep the diff small and avoid
touching every call site, the SDK type is re-exported from `iota-types::gas`
under the original name `GasCostSummary`, so most existing imports continue to
work unchanged.

## Description of change

- `iota-types::gas::GasCostSummary` is now a re-export of `iota_sdk_types::gas::GasCostSummary`; the local struct definition (and its `new_from_txn_effects`, `Display`, `AddAssign<&Self>` / `AddAssign<Self>` impls, and `serde_with`/`BigInt` annotations) has been removed. The SDK type provides equivalent serde behavior plus the `Add`/`Sub`/`AddAssign`/`SubAssign` operators added in iotaledger/iota-rust-sdk#550.
- `new_from_txn_effects` was not moved into the SDK — it has only one caller (`CheckpointBuilder`) and pulls in `iota-types::effects::TransactionEffects`, which is not part of the SDK surface. It is now a free function `checked::new_gas_cost_summary_from_txn_effects` colocated with its caller in `iota-core::checkpoints`, gated by `#[iota_macros::with_checked_arithmetic]` to preserve the overflow semantics that the original method had via the surrounding `checked` module.
- `Cargo.toml` bumps the `iota-sdk-types` git revision to `8b29706` (the commit that lands the `AddAssign`/`SubAssign` impls on `GasCostSummary`), so the re-export carries the operators that `iota-types` callers depend on.
- The reciprocal `From<crate::gas::GasCostSummary> for GasCostSummary` / `From<GasCostSummary> for crate::gas::GasCostSummary` conversions in `iota_sdk_types_conversions.rs` are removed — they would now be `From<T> for T`. The two `CheckpointSummary` `TryFrom` arms that called `.into()` on `epoch_rolling_gas_cost_summary` are simplified to direct field assignment.
- Adjusted snapshot/test fixtures across the transactional-test-runner suite to match the SDK type's display/serialization.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::gas::GasCostSummary` is now a re-export of `iota_sdk_types::gas::GasCostSummary`. The `new_from_txn_effects` associated function has been removed; its single caller now uses a free function colocated in `iota-core::checkpoints`. `GasCostSummary::sender_rebate` was not ported over to the external SDK.
- [ ] gRPC:


kodemartin (approved):
lgtm for the `graphql-e2e-tests`

valeriyr (approved):

piotrm50 (approved):

jkrvivian (approved):

semenov-vladyslav (approved):

Alex6323 (approved):

crates/iota-types/src/gas.rs
```diff
@@ -120,159 +116,7 @@ pub mod checked {
...
-        }
-
-        pub fn gas_used(&self) -> u64 {
-            self.computation_cost + self.storage_cost
-        }
-
-        /// Portion of the storage rebate that gets passed on to the transaction
-        /// sender. The remainder will be burned, then re-minted + added
-        /// to the storage fund at the next epoch change
-        pub fn sender_rebate(&self, storage_rebate_rate: u64) -> u64 {
```
roman1e2f5p8s, line 197: this method is not added to the SDK equivalent - fair enough, it is used nowhere in the monorepo.
but it is worth mentioning the removal in the PR for any potential external consumers

thibault-martinez: Added `GasCostSummary::sender_rebate was not ported over to the external SDK.` to the release notes.

roman1e2f5p8s (approved):

thibault-martinez (approved):

PR #11338, opened 2026-04-27
feat!: Replace `Digest` with SDK type

DaughterOfMars:

This PR replaces `Digest` (previously defined in `iota-types`) with the SDK
version from `iota-sdk-types`. To keep the diff small and avoid touching every
call site, the SDK type is re-exported from `iota-types` under the original
name `Digest`, so most existing imports continue to work unchanged. The same
applies to all the per-domain digest wrappers
(`TransactionDigest`, `CheckpointDigest`, `ObjectDigest`, etc.): they are
collapsed to type aliases for the SDK `Digest`, since each was a transparent
newtype around the same `[u8; 32]` and the BCS wire format
(`NEWTYPESTRUCT { TYPENAME: Digest }`) is identical.

## What changed

- The local `pub struct Digest([u8; 32])` definition and its inherent impls in
  [crates/iota-types/src/digests.rs](crates/iota-types/src/digests.rs) are
  removed. `Digest` is now a re-export:
  `pub use iota_sdk_types::Digest;`
- The twelve per-domain digest newtypes — `CheckpointDigest`,
  `CheckpointContentsDigest`, `CertificateDigest`, `SenderSignedDataDigest`,
  `TransactionDigest`, `TransactionEffectsDigest`, `TransactionEventsDigest`,
  `EffectsAuxDataDigest`, `ObjectDigest`, `ConsensusCommitDigest`,
  `MoveAuthenticatorDigest`, `MisbehaviorReportDigest` — are collapsed to
  type aliases: `pub type TransactionDigest = Digest;`, etc. Their
  hand-written `Display`, `Debug`, `LowerHex`, `UpperHex`, `FromStr`,
  `From`, `TryFrom`, `AsRef` and inherent impls (~1000 lines) are deleted;
  the SDK type provides equivalents.

### Single Domain vs Wrappers

The change to remove the various wrapper types means standardized functionality across all digests but it also means that there is no way to differentiate between how these digests were created and what they are related to. It is a trade-off between simplicity and explicitness.

## Call-site adjustments

Because the SDK `Digest` has a slightly different surface than the old
`iota-types` digests, a number of call sites had to be updated rather than
relying on the re-export:

- The `ObjectDigest` tombstone constants are renamed to drop the redundant
  `OBJECT_DIGEST_` prefix:
  - `OBJECT_DIGEST_DELETED` → `OBJECT_DELETED`
  - `OBJECT_DIGEST_WRAPPED` → `OBJECT_WRAPPED`
  - `OBJECT_DIGEST_CANCELLED` → `OBJECT_CANCELLED`
- The corresponding tombstone predicates are namespaced on the SDK type:
  `digest.is_deleted()` / `digest.is_wrapped()` / `digest.is_alive()` →
  `digest.is_object_deleted()` / `digest.is_object_wrapped()` / `digest.is_object_alive()` (see
  [writeback_cache.rs](crates/iota-core/src/execution_cache/writeback_cache.rs)).
- `TransactionDigest::genesis_marker()` (a `const fn`) is replaced by the
  associated constant `TransactionDigest::GENESIS_MARKER` everywhere
- `digest.base58_encode()` is replaced by `digest.to_base58()`
- `Digest::try_from(Vec<u8>)` / `Digest::try_from(&[u8])` are replaced by
  `Digest::from_bytes(...)`
- `next_lexicographical` semantics change: the SDK's
  `Digest::next_lexicographical(&self) -> Self` now wraps around at all-`FF`
  rather than returning `None`. The option-returning variant is
  `next_lexicographical_opt`. The
  [next_lexicographical_digest test](crates/iota-types/src/unit_tests/base_types_tests.rs)
  is updated to compare against `Digest::ZERO` for the wrap-around case
  rather than `None`.
- `Debug` output for the per-domain digests changes from the old
  `<Wrapper>(<inner>)` / `o#<base58>` form to the SDK's
  `Digest("<base58>")` form

A large number of touched files are purely the mechanical rename of
constants/methods (`OBJECT_DELETED`, `GENESIS_MARKER`, `to_base58`,
`from_bytes`, `is_object_deleted`/`is_object_wrapped`); those changes are
behavior-preserving.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: The `Digest` type and the per-domain digest wrappers
  (`TransactionDigest`, `CheckpointDigest`, `ObjectDigest`, etc.) are now
  re-exports/aliases of `iota_sdk_types::Digest` rather than locally defined
  newtypes. Surface changes: `genesis_marker()` is replaced by the
  `GENESIS_MARKER` associated constant; `base58_encode()` is replaced by
  `to_base58()`; `try_from(Vec<u8>)` / `try_from(&[u8])` are replaced by
  `from_bytes(...)`; `next_lexicographical(&self) -> Self` now wraps at
  all-`FF` (use `next_lexicographical_opt` for the previous
  `Option`-returning behavior); the `ObjectDigest` tombstone constants are
  renamed (`OBJECT_DIGEST_DELETED` → `OBJECT_DELETED`,
  `OBJECT_DIGEST_WRAPPED` → `OBJECT_WRAPPED`,
  `OBJECT_DIGEST_CANCELLED` → `OBJECT_CANCELLED`); the matching predicates
  are renamed (`is_deleted` → `is_object_deleted`,
  `is_wrapped` → `is_object_wrapped`); the `Debug` representation is
  unified as `Digest("<base58>")`.
- [ ] gRPC:


semenov-vladyslav (approved):

roman1e2f5p8s (commented):
Maybe I am missing something, but it feels wrong to replace all the 12 old tuple-struct wrappers with aliases to a single type for several reasons:

- All new aliases are now serializable, while old `CertificateDigest` and `SenderSignedDataDigest` were not.
- Many did not have the `Default` impl; now all new aliases have and return `ZERO`.
- Many did not have the `ZERO` const, `random()`, or `generate()` constructors; now all have them.
- `GENESIS_MARKER` was a transaction-digest-only concept; now all have it. `MisbehaviorReportDigest::GENESIS_MARKER` is now perfectly valid but makes no sense.
- `OBJECT_{DELETED,WRAPPED,CANCELLED}` and `is_object_{deleted,wrapped,alive}()` were object-digest-only concepts. Now `TransactionDigest::OBJECT_DELETED` is perfectly valid but makes no sense.
- Functions like `fn(digest: TransactionDigest)` and `fn(digest: ObjectDigest)` will not be treated by the compiler as different.
- `Debug` impl for some types like `ObjectDigest` are changed. So now debugging `Digest(...)` becomes harder; tracing logs with `?digest` lose per-domain prefix, potentially breaking grep pipelines.
- None of the aliases have docstrings, while some of the previous wrappers had them, properly explaining what each per-domain digest means.

Some of these are simply mentioned in the PR description, but none is properly explained or argued. Can you please give more context here?

Also, the `is_alive` -> `is_object_alive` rename is missing in the release notes.

kodemartin (approved):
lgtm for the infra crates

Thoralf-M (approved):

valeriyr (approved):

piotrm50 (approved):

roman1e2f5p8s (approved):

thibault-martinez (approved):

Alex6323 (approved):

PR #11342, opened 2026-04-28
fix(ci): fix several CI issues after version bumps

muXxer:
# Description of change

The new versions of the CI actions have a slightly changed behaviour, which causes some issues. This PR tries to address all known problems. It also fixes the issues with the dependabot and the two cargo workspaces (external-crates).

wiki-release fixed: https://github.com/iotaledger/iota/actions/runs/25038958773/job/73337192720
wiki-preview fixed: https://github.com/iotaledger/iota/actions/runs/25039773741
diff action doesn't report duplicated headers anymore

### DIFF
--- a/.github/workflows/_rust.yml
+++ b/.github/workflows/_rust.yml
@@ -58,6 +58,8 @@ jobs:
       components: ${{ steps.filter.outputs.changes }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         id: filter
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/_rust.yml
```diff
@@ -58,6 +58,8 @@ jobs:
       components: ${{ steps.filter.outputs.changes }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
thibault-martinez, line 62: Why only on some of them?

muXxer: because it's conflicting with the paths-filter action that follows those places. this will add it's own header again and then we get the issue.

.github/workflows/preview_wiki.yml
```diff
@@ -27,16 +27,13 @@ jobs:
...
-      - name: Install pnpm
-        uses: pnpm/action-setup@903f9c1a6ebcba6cf41d87230be49611ac97822e # v6.0.3
-        with:
-          run_install: false
-
       - name: Set up Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 24
-          cache: "pnpm"
```
thibault-martinez, line 39: no cache anymore?

thibault-martinez: Ah, we have an install step I guess

thibault-martinez (approved):

polinikita (approved):

PR #11348, opened 2026-04-28
feat(iota-benchmark): resilience bench implementation

### LINKED ISSUE #11269: [AA] Benchmark `max_auth_gas` impact on a validator during normal network workload

Dkwcs:
# Description of change

Adds a benchmark mode that characterizes per-validator handling of high-cost transactions while the rest of the network sustains a baseline workload through quorum. Lets us measure the practical effect of `max_auth_gas` and related caps on validator throughput under realistic mixed traffic and validate that those caps behave as designed.

- **`DirectValidatorProxy`** (`iota-benchmark/src/lib.rs`): a new `ValidatorProxy` implementation that sends transactions straight to one validator's gRPC `handle_transaction` endpoint instead of going through quorum. Useful for measuring the cost of a workload on a single node in isolation.

- **Concurrent workload groups** (`bench_driver.rs`): adds `BenchDriver::concurrent_groups`. When set, all groups start in parallel rather than serially. Group 0 is pinned to the `DirectValidatorProxy` (the node under measurement); group 1 uses the shared quorum proxy pool for the steady-state baseline.

- **Workload profiles**:
  - `auth-basic` — minimal AA authenticator: one `ed25519_verify`.
  - `auth-compute-heavy` — new `SuperHeavy` AA authenticator: 122 BenchObject reads plus a long `ed25519_verify` loop.
  - `std-invalid` — standard TransferObject with a corrupted (still well-formed) Ed25519 signature. 

Fixes #11269

### DIFF
--- a/crates/iota-benchmark/src/workloads/expected_failure.rs
+++ b/crates/iota-benchmark/src/workloads/expected_failure.rs
@@ -135,11 +135,8 @@ impl ExpectedFailureWorkloadBuilder {
             };
             // Expected-failure transactions (e.g. InvalidSignature) are rejected
             // before execution so gas is never consumed — the same gas objects can
-            // be reused across all in-flight slots. Capping num_payloads to
-            // num_workers avoids allocating O(target_qps * in_flight_ratio) gas
-            // coins during setup, which can take minutes at high QPS and prevent
-            // the benchmark from starting within the run interval.
-            let num_payloads = num_workers.max(1);
+            // be reused across all in-flight slots.
+            let num_payloads = (num_workers * in_flight_ratio).max(1);
             let workload_builder = Box::<dyn WorkloadBuilder<dyn Payload>>::from(Box::new(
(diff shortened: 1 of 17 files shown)


crates/iota-benchmark/src/workloads/expected_failure.rs
miker83z:
```suggestion
// Bumping byte in position 1 (skipping position 0 because it is the flag) by 1
// (mod 256) with a guaranteed-different value, makes it so that `from_bytes`
// still parses but verification fails.
for signature in tx.tx_signatures_mut_for_testing().iter_mut() {
    let mut bytes = signature.as_ref().to_vec();
    bytes[1] = bytes[1].wrapping_add(1);
    *signature = GenericSignature::from_bytes(&bytes).unwrap();
}
```

muXxer (approved):

filipdulic (approved):

PR #11360, opened 2026-04-28
feat(docs): add steps for local validator

Thoralf-M:
Tested by running the commands


Thoralf-M:
Too specific for the public docs

PR #11355, opened 2026-04-28
chore: Replace `Identifier`, `StructTag` and `TypeTag` with SDK versions (#10075)

DaughterOfMars:
This PR is part of the ongoing effort to consolidate blockchain types in the
canonical `iota-rust-sdk`. Types that are needed by both clients/SDK consumers
and the node should live in `iota-rust-sdk` and be imported from there, instead
of being defined twice (once in `iota-types`, once in the SDK).

This PR replaces `Identifier`, `StructTag` and `TypeTag` (previously sourced
from `move-core-types` and re-exported through `iota-types`) with the SDK
versions from `iota-sdk-types`. To keep the diff small and avoid touching
every call site, the SDK types are re-exported from `iota-types::base_types`
under the original names `Identifier`, `StructTag` and `TypeTag`, so most
existing imports continue to work unchanged. The `move-core-types` definitions
are intentionally not removed — they are still used internally by the Move VM
(e.g. in `MoveStructLayout` / `MoveValue` / `ModuleId`), and conversion
helpers bridge between the two representations at the boundary.

## Description of change

- `iota-types::base_types` now re-exports `iota_sdk_types::{Identifier, StructTag, TypeTag}`; imports of `move_core_types::{identifier::Identifier, language_storage::{StructTag, TypeTag}}` across the workspace have been replaced with the SDK versions.
- Helpers and constants that previously lived on the local types are gone in favour of the SDK API:
  - `Coin::type_(t)` / `Coin::is_coin(s)` → `StructTag::new_coin(t)` / `StructTag::is_coin(s)`.
  - `TreasuryCap::type_(t)` / `CoinMetadata::type_(t)` / `Balance::type_(t)` / `GAS::type_()` / `StakedIota::type_()` etc. → `StructTag::new_treasury_cap(...)` / `StructTag::new_coin_metadata(...)` / `StructTag::new_balance(...)` / `StructTag::new_gas()` / `StructTag::new_staked_iota()`, etc.
  - `COIN_MODULE_NAME` / `COIN_STRUCT_NAME` / `STAKING_POOL_MODULE_NAME` / `STAKED_IOTA_STRUCT_NAME`, … → `Identifier::COIN_MODULE` / `Identifier::COIN` / `Identifier::STAKING_POOL_MODULE` / `Identifier::STAKED_IOTA`, … (and `Identifier::from_static("…")` for the few that don't have a dedicated constant).
[...]

PR #10075, opened 2026-02-04
chore: Replace `Identifier`, `StructTag` and `TypeTag` with SDK versions

DaughterOfMars:
## Description

This PR replaces the `Identifier`, `StructTag` and `TypeTag` without removing the definitions in external crates.

Closes #9839

PR #11375, opened 2026-04-29
chore: Revert some Identifier changes from "Replace Identifier, StructTag and TypeTag with SDK versions (#10075)" (#11355)

miker83z:
# Description of change

- Reverts the move_core_types::identifier::IdentStr/Identifier → iota_sdk_types::Identifier replacement inside iota-execution, keeping SDK Identifier only at boundary points where types cross crate boundaries (e.g., Command::move_call construction, Event::new)
- Internal functions, constants, and comparisons use move_core_types IdentStr directly, eliminating unnecessary conversions at the Move VM boundary
- Changes is_test_fun and get_authenticator_version_from_fun in iota_types::move_package to accept &str instead of &Identifier, since they only need the string value
- StructTag, TypeTag, and ObjectID SDK replacements from the original commit are intentionally kept, as they are needed by external types (Event, MoveObjectType, execution results)

### DIFF
--- a/crates/iota-types/src/move_package.rs
+++ b/crates/iota-types/src/move_package.rs
@@ -628,7 +628,7 @@ impl UpgradeReceipt {
 }
 
 /// Checks if a function is annotated with one of the test-related annotations
-pub fn is_test_fun(name: &Identifier, module: &CompiledModule, fn_info_map: &FnInfoMap) -> bool {
+pub fn is_test_fun(name: &str, module: &CompiledModule, fn_info_map: &FnInfoMap) -> bool {
     let fn_name = name.to_string();
     let mod_handle = module.self_handle();
     let mod_addr = IotaAddress::new(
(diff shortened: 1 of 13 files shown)


crates/iota-types/src/move_package.rs
```diff
@@ -628,7 +628,7 @@ impl UpgradeReceipt {
 }
 
 /// Checks if a function is annotated with one of the test-related annotations
-pub fn is_test_fun(name: &Identifier, module: &CompiledModule, fn_info_map: &FnInfoMap) -> bool {
+pub fn is_test_fun(name: &str, module: &CompiledModule, fn_info_map: &FnInfoMap) -> bool {
     let fn_name = name.to_string();
```
DaughterOfMars: This isn't needed now

crates/iota-types/src/move_package.rs
DaughterOfMars: same here

crates/iota-types/src/move_package.rs
miker83z: fixed [b46ef86](https://github.com/iotaledger/iota/pull/11375/commits/b46ef86914f1e5224d19b553dd0fb00b62d2e7c3)

crates/iota-types/src/move_package.rs
miker83z: [b46ef86](https://github.com/iotaledger/iota/pull/11375/commits/b46ef86914f1e5224d19b553dd0fb00b62d2e7c3)

PR #11376, opened 2026-04-29
feat(iota-types)!: replace Event with SDK type

Thoralf-M:

This PR replaces `Event` (previously defined in `iota-types::event`) with the
every call site, the SDK type is re-exported from `iota-types::event` under the
original name `Event`, so most existing imports continue to work unchanged.

## Description of change

- `iota-types::event::Event` is now a re-export of `iota_sdk_types::Event`; the
  local struct definition and its inherent impls (`new`,
  `move_event_to_move_value`, `is_system_epoch_info_event*`,
  `random_for_testing`) have been removed. The SDK type provides equivalent
  serde behavior plus the `is_system_epoch_info_event*` helpers.
- The `transaction_module` field has been renamed to `module` to match the
  SDK type. The public JSON-RPC `IotaEvent` type keeps its
  `transaction_module` field name; the rename is handled in the conversions,
  so the external RPC schema is unchanged.
- `Event::move_event_to_move_value` was not moved into the SDK — it depends
  on `BoundedVisitor` from `iota-types::object`, which is not part of the
  SDK surface. Its single caller (`IotaEvent::try_from_event`) now calls
  `BoundedVisitor::deserialize_value` directly.
- `DisplayVersionUpdatedEvent::inner_type` / `try_from_event` and the unused
  `DisplayCreatedEvent` have been removed from `iota-types::display`. The
  only caller (`StoredDisplay::try_from_event` in the indexer) already
  reimplements the same logic locally, so no behavior changes.
- The reciprocal `TryFrom` conversions for `Event` and `TransactionEvents`
  in `iota_sdk_types_conversions.rs` are removed — they would now be
  `From<T> for T`.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::event::Event` is now a re-export of
  `iota_sdk_types::Event`. The `transaction_module` field has been renamed
  to `module`; the `Event::new` constructor and
  `Event::move_event_to_move_value` associated function have been removed.
  `DisplayVersionUpdatedEvent::inner_type` / `try_from_event` and the unused
  `DisplayCreatedEvent` have been removed from `iota-types::display`.
- [ ] gRPC:


piotrm50 (approved):

valeriyr (approved):

crates/iota-json-rpc-types/src/iota_event.rs
```diff
@@ -227,7 +228,10 @@ impl IotaEvent {
             bcs: contents.to_vec(),
         };
 
-        let move_value = Event::move_event_to_move_value(&contents, layout)?;
+        let move_value = BoundedVisitor::deserialize_value(&contents, &layout.into_layout())
+            .map_err(|e| iota_types::error::IotaError::ObjectSerialization {
```
kodemartin:
```suggestion
            .map_err(|e| iota_types::error::IotaError::ObjectDeserialization {
```

kodemartin (changes requested):
looks good with a small change request

semenov-vladyslav (approved):

crates/iota-json-rpc-types/src/iota_event.rs
Thoralf-M: Good catch!

alexsporn (approved):
LGTM, just the TODO in the grpc-server should be cleaned up before merging

kodemartin (approved):
lgtm

crates/iota-grpc-server/src/merge.rs
```diff
@@ -645,17 +645,7 @@ impl Merge<iota_types::effects::TransactionEvents> for TransactionEvents {
         source: iota_types::effects::TransactionEvents,
         mask: &FieldMaskTree,
     ) -> Result<(), Self::Error> {
-        if !mask.contains(Self::DIGEST_FIELD.name) && !mask.contains(Self::EVENTS_FIELD.name) {
-            // No need to convert if no field is requested
-            return Ok(());
-        }
```
thibault-martinez, line 651: How come we are removing this?

thibault-martinez: Ah, it was really just to avoid conversion ?

crates/iota-json-rpc-types/src/iota_event.rs
```diff
@@ -227,7 +228,10 @@ impl IotaEvent {
             bcs: contents.to_vec(),
         };
 
-        let move_value = Event::move_event_to_move_value(&contents, layout)?;
+        let move_value = BoundedVisitor::deserialize_value(&contents, &layout.into_layout())
```
thibault-martinez, line 231: Do you remember why it became `BoundedVisitor`?

crates/iota-grpc-server/src/merge.rs
Thoralf-M: Yes since the conversion is not needed anymore it all can be removed

crates/iota-json-rpc-types/src/iota_event.rs
Thoralf-M: The `move_event_to_move_value` just wrapped it:
```Rust
    pub fn move_event_to_move_value(
        contents: &[u8],
        layout: MoveDatatypeLayout,
    ) -> IotaResult<MoveValue> {
        BoundedVisitor::deserialize_value(contents, &layout.into_layout()).map_err(|e| {
            IotaError::ObjectSerialization {
                error: e.to_string(),
            }
        })
    }
```

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -945,13 +945,19 @@ mod checked {
             let user_events = user_events
                 .into_iter()
                 .map(|(module_id, tag, contents)| {
-                    Event::new(
-                        IotaAddress::new(module_id.address().into_bytes()),
-                        Identifier::new_unchecked(module_id.name().as_str()),
-                        ref_context.borrow().sender(),
-                        tag,
-                        contents,
+                    let package_id = iota_types::base_types::ObjectID::from_bytes(
```
thibault-martinez: Just use `new` so it's not even fallible ?

thibault-martinez: Also please import `ObjectID`

thibault-martinez (approved):

PR #11379, opened 2026-04-29
feat(node,doc): Add epoch start timestamps to MANIFEST and add a snapshot download picker

### LINKED ISSUE #11259: snapshots: provide an easy way to list the existing formal snapshots
We will keep several formal snapshot files in the archive, but it is intransparent for the user which files exist in the remote store.

We could create a website or a page in our node operator docs, that lists all existing formal snapshots in the remote store, and provides an straightforward way to copy the download command for `iota-tool`, depending on their setup. (existing snapshot files are already listed in https://formal-snapshot.mainnet.iota.cafe/MANIFEST). Maybe we can just add a dropdown menu here https://docs.iota.org/operator/full-node/configs/snapshots#restoring-from-formal-snapshots

So the user can choose if they use the binary setup, or docker, then they choose the formal snapshot, and copy the resulting command.

jkrvivian:
# Description of change

- Add epoch start timestamp to MANIFEST file, so the developer can easily understand the time of the available epoch.
  - Add per-epoch metadata file `epoch_<NUM>/_epoch_metadata.json` (`EpochMetadata { epoch_start_timestamp_ms: u64 }`) to record the timestamp of the epoch, they will then be read and aggregated into the MANIFEST file.
- Modify document Snapshot page:
  - Add `FormalSnapshotPicker` — Fetches `https://formal-snapshot.{network}.iota.cafe/MANIFEST` at runtime and lists every available epoch; if `epoch_start_timestamps_ms` is present, each option is labelled `<epoch> — started YYYY-MM-DD HH:mm UTC`, otherwise it falls back to the bare epoch number.
  - Add `RocksDBSnapshotPicker`

Closes #11259 .


bingyanglin (approved):

crates/iota-storage/src/object_store/util.rs
```diff
@@ -23,23 +23,39 @@ use crate::object_store::{
...
 
 pub const MANIFEST_FILENAME: &str = "MANIFEST";
+pub const EPOCH_METADATA_FILENAME: &str = "_epoch_metadata.json";
 
 #[derive(Serialize, Deserialize)]
-
 pub struct Manifest {
     pub available_epochs: Vec<u64>,
+    #[serde(default)]
+    pub epoch_start_timestamps_ms: BTreeMap<u64, u64>,
```
semenov-vladyslav: Why do you need a `BTreeMap`? Why not `pub available_epochs: Vec<(u64, Option<u64>)>`?

docs/content/operator/common/snapshots.mdx
```diff
@@ -166,37 +143,11 @@ See [Install IOTA](../../developer/getting-started/install-iota.mdx#install-from
...
-         --network testnet \
-         --path <PATH-TO-NODE-DB> \
-         --num-parallel-downloads 50 \
-         --verify normal \
-         --no-sign-request \
-         --verbose
-      ```
-      </TabItem>
-   </Tabs>
+2. Pick your setup, network, and a snapshot epoch. The dropdown lists every epoch currently available in the IOTA Foundation's public store (sourced from the network's [`MANIFEST`](https://formal-snapshot.mainnet.iota.cafe/MANIFEST)). Copy the resulting command and run it:
```
muXxer: Doesn't matter where this comes from or where it is stored. I would remove the whole "dropdown" sentence tbh.

docs/site/src/components/RocksDBSnapshotPicker/index.tsx
muXxer, line 1: And it needs to be such a react component? We can't simply build the 
```
<Tabs groupId="network" queryString>
      <TabItem label="Mainnet" value="mainnet">
```
dynamically I guess?

crates/iota-storage/src/object_store/util.rs
```diff
@@ -296,8 +312,27 @@ pub async fn run_manifest_update_loop(
...
+                            match serde_json::from_slice::<EpochMetadata>(&bytes) {
+                                Ok(metadata) => {
+                                    epoch_start_timestamps_ms
+                                        .insert(*epoch, metadata.epoch_start_timestamp_ms);
+                                }
+                                Err(err) => {
+                                    warn!("Failed to parse epoch metadata for epoch {epoch}: {err}");
+                                }
+                            }
+                        }
```
semenov-vladyslav: I would move this code into `list_all_epochs`.

muXxer (changes requested):

docs/site/src/components/RocksDBSnapshotPicker/index.tsx
jkrvivian: I changed it into a `Tabs` with 4 `TabItem` instead.

jkrvivian: 圖片

muXxer (approved):

semenov-vladyslav (approved):

PR #11392, opened 2026-04-30
feat(starfish): evict pending commit votes via quorum commit index

### LINKED ISSUE #11391: starfish(core): evict pending commit votes via CommitVoteMonitor quorum index
Bound the in-memory `pending_commit_votes` tracker in `DagState` by evicting votes the network already certified.

**Key points:**
- Switch `pending_commit_votes` from `VecDeque<CommitVote>` to `BTreeSet<CommitVote>` so eviction can use `split_off` in $O(\log n)$.
- Wire `Arc<CommitVoteMonitor>` into `DagState` via a setter called from `Core::new`.
- Add `DagState::evict_pending_commit_votes()` that drops votes with `index <= monitor.quorum_commit_index() - gc_depth`. Gated on `consensus_block_restrictions`.
- Call eviction from both `take_commit_votes` and `flush()` (mirrors the `pending_acknowledgments` pattern).

### LINKED ISSUE #11400: starfish(core): centralize Core::should_propose skip reasons
Tighten the proposal skip paths in `Core` so every reason for not proposing is observable in one place.

**Key points:**
- Move the `consensus_block_restrictions` quorum-commit-round gate from `try_new_block` into `should_propose`, with a new `behind_quorum_commit_round` skip-reason metric label.
- Move the `clock_round <= last_proposed_round` check from `try_new_block` into `should_propose`. Keep it silent (no metric, no debug log) since it fires on every accepted block in a round we have already proposed in.
- Centralize all skip reasons behind a `SkipProposalReason` enum + `skip_proposal` helper so the `core_skipped_proposals` metric label space is disjoint and visible in one place.
- `try_new_block` becomes a single line for the round read; all gating is upstream.

polinikita:
# Description of change

Bounds the in-memory `pending_commit_votes` tracker in `DagState` and tightens the proposal skip paths in `Core`.

### Eviction (issue #11391)

- Switch `pending_commit_votes` from `VecDeque<CommitVote>` to `BTreeSet<CommitVote>` so eviction uses `split_off`. Insertion order is preserved (the linearizer produces commits monotonically); insertion dedup is a free side benefit.
- New field on `DagState`: `last_known_quorum_commit_index: CommitIndex` (no `Arc<CommitVoteMonitor>` reference). `Core` is the sole holder of the monitor and pushes the value via a setter.
- `evict_pending_commit_votes()` drops votes with `index <= last_known_quorum_commit_index - gc_depth`. Gated on `consensus_block_restrictions`; no-op otherwise.
- Eviction is called both from `take_commit_votes` (proposal path) and from `flush()` (mirrors the existing `pending_acknowledgments` pattern).
- `Core` refreshes `DagState::last_known_quorum_commit_index` from `CommitVoteMonitor::quorum_commit_index()` at three places: `Core::try_commit` after every successful `commit_observer.handle_committed_leaders` (regular consensus path); `Core::handle_committed_sub_dags_from_fast_sync` (fast-sync path); `Core::should_propose` (right before we propose).

### `should_propose` cleanup (issue #11400)

- Centralised every skip reason behind a new `SkipProposalReason` enum + `skip_proposal` helper, so the `core_skipped_proposals` metric label space is disjoint and visible in one place.
- New skip reason `behind_quorum_commit_round` — the `consensus_block_restrictions` gate (formerly inside `try_new_block`) moves up to `should_propose`.
- The `clock_round <= last_proposed_round` recheck moves up too. It is silent (no metric, no debug log) since it fires on every accepted block in a round we have already proposed in — counting it would drown the genuinely interesting reasons.
- `try_new_block` now reads `clock_round` only; all gates are upstream.

Fixes #11391
Fixes #11400

- [x] New unit test `dag_state::test::test_evict_pending_commit_votes` covering: quorum-driven eviction with the flag on; default-field no-op (no setter call); flag-off no-op even with a non-zero field
- [x] Existing 280 starfish-core lib tests still pass
- [x] `cargo ci-clippy` clean across the workspace

### DIFF
--- a/crates/starfish/core/src/core.rs
+++ b/crates/starfish/core/src/core.rs
@@ -774,9 +774,9 @@ impl Core {
         Ok((None, BTreeMap::new()))
     }
 
-    /// Attempts to propose a new block for the next round. If a block has
-    /// already proposed for latest or earlier round, then no block is
-    /// created and None is returned.
+    /// Attempts to propose a new block at the current clock round. Eligibility
+    /// (round bounds, block restrictions) is enforced upstream in
+    /// `should_propose`.
     #[instrument(level = "trace", skip_all)]
     fn try_new_block(&mut self, reason: ReasonToCreateBlock) -> Option<VerifiedBlock> {
         let _s = self
@@ -1218,7 +1218,10 @@ impl Core {
         info!("Last known proposed round set to {round}");
     }
 
-    /// Whether the core should propose new blocks.
+    /// Returns true when Core should propose at the current clock round. As a
+    /// side effect, when proposal is greenlit under
+    /// `consensus_block_restrictions`, refreshes `DagState`'s last-known
+    /// quorum commit index to enable eviction for commit votes
     pub(crate) fn should_propose(&self) -> bool {
         let (clock_round, last_proposed_round, local_commit_index, local_commit_round) = {
             let dag_state = self.dag_state.read();
--- a/crates/starfish/core/src/core.rs
+++ b/crates/starfish/core/src/core.rs
@@ -155,6 +155,34 @@ impl ReasonToCreateBlock {
     }
 }
 
+/// Reason a `Core::should_propose` call returned `false`. Used as the `reason`
+/// label on the `core_skipped_proposals` metric. Keeping the variants in one
+/// enum makes the label space disjoint and centrally visible. Variant data is
+/// the per-reason context interpolated into the corresponding `debug!` line.
+///
+/// The "already proposed at this round" branch is intentionally not modeled
+/// here: it fires on every block accepted in a round we have already proposed
+/// in, which is a normal high-rate event. Counting and logging it would drown
+/// the genuinely interesting reasons below.
+#[derive(Clone, Copy)]
+pub(crate) enum SkipProposalReason {
+    NoQuorumSubscriber,
+    NoLastKnownProposedRound,
+    HigherLastKnownProposedRound { last_known: Round },
+    BehindQuorumCommitRound { approx_quorum: Round },
+}
+
+impl SkipProposalReason {
+    fn label(self) -> &'static str {
+        match self {
+            Self::NoQuorumSubscriber => "no_quorum_subscriber",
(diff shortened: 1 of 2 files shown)


crates/starfish/core/src/core.rs
```diff
@@ -1200,39 +1220,99 @@ impl Core {
 
     /// Whether the core should propose new blocks.
     pub(crate) fn should_propose(&self) -> bool {
```
piotrm50, line 1225: This now modifies dag_state so we should update the comment and perhaps update the name as well to indicate that it's not just evaluating some criteria anymore.

crates/starfish/core/src/core.rs
piotrm50: Is this comment still relevant? I see that this check has been removed from below
```
if clock_round <= min_round {
                return None;
            }
```

crates/starfish/core/src/core.rs
```diff
@@ -1200,39 +1220,99 @@ impl Core {
...
+                    },
+                );
+            }
+
+            // We are about to propose: refresh DagState's known quorum commit
+            // index so the eviction of `pending_commit_votes` is bounded.
+            self.dag_state
+                .write()
+                .set_last_known_quorum_commit_index(quorum_commit_index);
+        }
```
VorobyevIlya, line 1279: Should this part be gated by block restrictions flag? Maybe it should be always executed

crates/starfish/core/src/dag_state.rs
```diff
@@ -1893,9 +1898,13 @@ impl DagState {
...
-        let mut votes = Vec::new();
-        while !self.pending_commit_votes.is_empty() && votes.len() < limit {
-            votes.push(self.pending_commit_votes.pop_front().unwrap());
+        self.evict_pending_commit_votes();
+        let mut votes = Vec::with_capacity(limit.min(self.pending_commit_votes.len()));
+        while votes.len() < limit {
+            match self.pending_commit_votes.pop_first() {
+                Some(v) => votes.push(v),
+                None => break,
+            }
```
VorobyevIlya: do we need to pop votes one by one here? Could we use split or something like that?

VorobyevIlya (approved):

PR #10331, opened 2026-02-17
feat: replace `Argument` with SDK version

thibault-martinez:
## Links to any relevant issues

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/590

## How the change has been tested

- [ ] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #11393, opened 2026-04-30
feat: replace `Argument` with SDK version (#10331)

thibault-martinez:

This PR replaces `Argument` (previously defined in `iota-types`) with the
original name `Argument`, so most existing imports continue to work
unchanged.

## Notable changes

- The `Argument::GasCoin` variant is renamed to `Argument::Gas` to match the
  SDK definition. Call sites are updated accordingly.
- The local `Display` impl is removed in favor of the SDK's. This changes the
  textual representation of the gas argument from `GasCoin` to `Gas` (visible
  in PTB snapshot tests and CLI output).
- The bidirectional `From`/`Into` conversions between the local and SDK
  `Argument` types are removed since both sides are now the same type, which
  simplifies `iota_sdk_types_conversions.rs` for `Command` significantly.
- `scripts/update_sdk_types_rev.sh` is made cross-platform (works on macOS
  too): replaces GNU-only `sed -i` and `grep -oP` with portable equivalents,
  and falls back to a background-process timeout when neither `timeout` nor
  `gtimeout` is available.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: The `Argument::GasCoin` PTB argument now displays as `Gas` instead
  of `GasCoin` in CLI output and PTB snapshots.
- [x] Rust SDK: `iota_types::transaction::Argument` is now a re-export of
  `iota_sdk_types::Argument`. The `GasCoin` variant has been renamed to `Gas`
  to match the SDK definition.
- [ ] gRPC:


kodemartin (approved):
lgtm for the infra crates

valeriyr (approved):

semenov-vladyslav (approved):

crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
```diff
@@ -417,10 +417,11 @@ impl From<NativeArgument> for TransactionArgument {
         use NativeArgument as N;
         use TransactionArgument as A;
         match argument {
-            N::GasCoin => A::GasCoin(GasCoin { dummy: None }),
+            N::Gas => A::GasCoin(GasCoin { dummy: None }),
             N::Input(ix) => A::Input(Input { ix }),
             N::Result(cmd) => A::Result(TxResult { cmd, ix: None }),
             N::NestedResult(cmd, ix) => A::Result(TxResult { cmd, ix: Some(ix) }),
+            _ => unimplemented!("a new enum variant was added and needs to be handled"),
```
DaughterOfMars: Might as well update these

DaughterOfMars (approved):

filipdulic (approved):

Alex6323 (approved):

crates/iota-graphql-rpc/src/types/transaction_block_kind/programmable.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/11393/changes/95bbb5858741437c961c1ef61b73c9bf1508f682

roman1e2f5p8s (approved):
lgtm, ~~but maybe worth ticking JSON-RPC in release notes and briefly mentioning that the JSON serialization changed?~~

PR #11397, opened 2026-04-30
fix(json-rpc): Populate balance-changes object cache with input objects

### LINKED ISSUE #10977: showBalanceChanges causes spurious -32602 errors under concurrent shared-object load
Concurrent shared-object transactions fail with ObjectSequenceNumberTooHigh when showBalanceChanges is enabled

#  Summary                                                                                                                                                                                                                       
   
  When submitting multiple transactions concurrently via iota_executeTransactionBlock that touch the same shared objects, the RPC intermittently returns:                                                                       
                                                                                                                                                                                                                              
  code: -32602                                                                                                                                                                                                                  
  message: "Could not find the referenced object <id> as the asked version SequenceNumber(N) is higher than the latest SequenceNumber(N-1)"                                                                                     
                                                                                                                                                                                                                                
  This only occurs when IotaTransactionBlockResponseOptions includes showBalanceChanges: true (e.g. via full_content()). Disabling showBalanceChanges eliminates the error.                                                     
                                                                                                                                                                                                                                
# Root cause                                                                                                                                                                                                                    
                                                                                                                                                                                                                              
  After a transaction is successfully executed by validators and effects are returned, the fullnode's handle_post_orchestration builds the RPC response. When showBalanceChanges is requested, get_balance_changes_from_effect  
  reads each mutated object at both its previous version (modified_at_versions()) and its new version (all_changed_objects()).
                                                                                                                                                                                                                                
  An ObjectProviderCache is pre-populated with the transaction's output objects (new versions) to avoid hitting the local store. However:                                                                                       
                                                                                                                                                                                                                              
  1. The previous versions of objects are never in the cache, so they always fall through to get_past_object_read() on the fullnode's local store.                                                                              
  2. Under concurrent load, the fullnode's local store may not yet have written the results of a different recently-executed transaction that produced that previous version, causing a VersionTooHigh error.                 
  3. Even the new versions can miss the cache if the output_objects field is None in the validator response — which happens when quorum is reached before the sampled validator responds (logged as "Quorum Reached but         
  requested input/output objects were not returned"), or when get_transaction_output_objects fails silently due to .and_then(Result::ok) at authority_server.rs:700-702.                                                        
                                                                                                                                                                                                                                
  The error propagates as PastObjectRead::VersionTooHigh → UserInputError::ObjectSequenceNumberTooHigh → StateReadError::Client → JSON-RPC error code -32602, making it appear as a client input validation error when it is    
  actually a transient fullnode-internal race condition.                                                                                                                                                                      
                                                                                                                                                                                                                                
#  Reproduction                                                                                                                                                                                                                
                                                                                                                                                                                                                              
  1. Deploy a local network
  2. Submit multiple transactions concurrently that mutate the same shared objects (e.g. DEX stress test)
  3. Use IotaTransactionBlockResponseOptions::full_content() (or any options with showBalanceChanges: true)                                                                                                                     
  4. Observe intermittent -32602 errors for shared objects despite transactions containing only ObjectArg::SharedObject inputs                                                                                                  
                                                                                                                                                                                                                                
  Submitting the same transactions serially, or without showBalanceChanges, does not trigger the error.                                                                                                                         
                                                                                                                                                                                                                                
#  Relevant code path                                                                                                                                                                                                            
                                                                                                                                                                                                                              
  1. RPC handler: transaction_execution_api.rs → handle_post_orchestration → get_balance_changes_from_effect                                                                                                                    
  2. Balance changes: balance_changes.rs:97 reads `modified_at_versions()` (previous version, cache miss) and :108 reads `all_changed_objects()` (new version, should hit cache)                                                  
  3. Cache fallback: ObjectProviderCache::get_object → `Arc<S>::get_object` (authority_state.rs:564) → `get_past_object_read`                                                                                                       
  4. Error origin: authority.rs:3700 → `PastObjectRead::VersionTooHigh`                                                                                                                                                    
  5. Silent error swallowing: authority_server.rs:700-702 —` .and_then(Result::ok)` discards get_transaction_output_objects failures, leaving output_objects: None

jkrvivian:
# Description of change

- Port the fix from upstream commit: https://github.com/MystenLabs/sui/commit/cc13b433d65b9171e544c407d6703e6ec816e718
  -  Pre-populate the cache with both `response.input_objects` and `response.output_objects`. 
- Log discarded errors from `get_transaction_input_objects` / `get_transaction_output_objects` in `authority_server.rs`. 
- Port the follow-up fix upstream commit: https://github.com/MystenLabs/sui/commit/c34b812822ff7cef93ddab8a5ef91c2948bbb93c (The test is skipped)
  - Create the `ObjectProviderCache` when `show_balance_changes` or `show_object_changes` is requested, instead of only when both `input_objects` and `output_objects` are Some.
  - We do not have the bug that the upstream found bc we don't support address-balance-backed transactions with no coin inputs, so the `input_objects` is always Some. We still port the fix because it could still happen if we failed to load the input objects from the local store is reachable not via "no coin inputs" but via a transient store read failure.

## Test plan

End-to-end stress test against local network with `fullnode-1`'s `iota-node` binary hot-swapped between this branch and `6bb98867b7` (the commit before the fix), following the testing flow as the issue described. CLI temporarily patched to use `WaitForEffectsCert` because the default `WaitForLocalExecution` masks the race entirely:

  | Fullnode | Concurrent shared-object increments | Successful | -32602 errors |
  |---|---|---|---|
  | `6bb98867b7`  | 90 | 2 | **88** |
  | `6bb98867b7` (rerun) | 80 | 1 | **79** |
  | this branch | 80 | 80 | **0** |
  | this branch (rerun) | 80 | 80 | **0** |

  Sample pre-fix error (the symptom from the issue):
  ```
  ErrorObject { code: InvalidParams, message: "Could not find the referenced object
  ObjectId(\"0x6af8…\") as the asked version Version(2234) is higher than the latest
  Version(2227)", data: None }
  ```

Fixes #10977.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: Fixed "Could not find the referenced object" errors in balance/object changes for concurrent shared-object transactions.
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-core/src/authority_server.rs
+++ b/crates/iota-core/src/authority_server.rs
@@ -49,7 +49,7 @@ use tonic::{
     metadata::{Ascii, MetadataValue},
     transport::server::TcpConnectInfo,
 };
-use tracing::{Instrument, debug, error, error_span, info, trace_span};
+use tracing::{Instrument, debug, error, error_span, info, trace_span, warn};
 
 use crate::{
     authority::{AuthorityState, authority_per_epoch_store::AuthorityPerEpochStore},
@@ -695,11 +695,29 @@ impl ValidatorService {
 
                 let input_objects = include_input_objects
                     .then(|| self.state.get_transaction_input_objects(&effects))
-                    .and_then(Result::ok);
+                    .and_then(|res| {
+                        res.map_err(|e| {
+                            warn!(
+                                tx_digest = ?effects.transaction_digest(),
+                                error = ?e,
+                                "Failed to load transaction input objects requested by client",
+                            )
+                        })
+                        .ok()
+                    });
 
                 let output_objects = include_output_objects
                     .then(|| self.state.get_transaction_output_objects(&effects))
-                    .and_then(Result::ok);
+                    .and_then(|res| {
+                        res.map_err(|e| {
+                            warn!(
+                                tx_digest = ?effects.transaction_digest(),
+                                error = ?e,
+                                "Failed to load transaction output objects requested by client",
+                            )
+                        })
+                        .ok()
+                    });
 
                 let signed_effects = self.state.sign_effects(effects, epoch_store)?;
                 epoch_store.insert_tx_cert_sig(certificate.digest(), certificate.auth_sig())?;
--- a/crates/iota-json-rpc/src/balance_changes.rs
+++ b/crates/iota-json-rpc/src/balance_changes.rs
@@ -186,6 +186,31 @@ impl<P> ObjectProviderCache<P> {
         }
     }
 
+    #[instrument(level = "trace", skip_all)]
+    pub fn insert_objects_into_cache(&mut self, objects: Vec<Object>) {
+        let object_cache = self.object_cache.get_mut();
+        let last_version_cache = self.last_version_cache.get_mut();
+
+        for object in objects {
+            let object_id = object.id();
+            let version = object.version();
+
+            let key = (object_id, version);
+            object_cache.insert(key, object.clone());
+
+            match last_version_cache.get_mut(&key) {
+                Some(existing_seq_number) => {
+                    if version > *existing_seq_number {
+                        *existing_seq_number = version
+                    }
+                }
+                None => {
+                    last_version_cache.insert(key, version);
+                }
+            }
+        }
+    }
+
     pub fn new_with_cache(
         provider: P,
         written_objects: BTreeMap<ObjectID, (ObjectRef, Object, WriteKind)>,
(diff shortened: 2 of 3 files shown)


muXxer (approved):

bingyanglin (changes requested):
Could you please check if this one https://github.com/MystenLabs/sui/pull/25685 fixes some bugs that introduced in this upstream change and we might include here?

kodemartin (approved):
lgtm

bingyanglin (approved):

muXxer:
Shouldn't we port the same test that was introduced in the upstream PR 25685? Or is this only address balance related?

jkrvivian:
> Shouldn't we port the same test that was introduced in the upstream PR 25685? Or is this only address balance related?

* The fix is address-balance-only. We don't have a case where the input objects could be empty.
* The test isn't portable: it depends on a lot of implementations, and the entire test file is introduced in the later version (1.52.x).

piotrm50 (approved):

PR #11402, opened 2026-04-30
feat(starfish): GC-based eviction in BlockManager

### LINKED ISSUE #11401: starfish(block_manager): GC-based eviction of stale missing-ancestor and suspended-tx state
`BlockManager` keeps every missing ancestor it sees in `missing_ancestors` / `headers_to_fetch` and every block waiting on one in `suspended_headers`. Refs at rounds at or below the network's GC floor cannot influence any not-yet-sequenced block, so retaining them is unbounded memory pressure and wasted fetch volume — particularly visible after every fast-commit-sync handoff.

**Key points:**
- After fast sync downloads commits + leader-subdag headers but skips intermediate non-leader headers, `RegularCommitSyncer`'s next batches reference parents below fast's frontier; `find_missing_ancestors` registers them and `HeaderSynchronizer` chases them indefinitely.
- `BlockManager::reinitialize` hard-clears suspended state on fast-sync completion, discarding locally-known headers and forcing re-fetches.
- `suspended_transactions` is unbounded (existing TODO at `block_manager/mod.rs:45`).
- Introduce a periodic in-line eviction sweep keyed on `DagState::gc_round_for_last_commit()`, gated by `consensus_block_restrictions`, that drops stale `missing_ancestors`/`headers_to_fetch`/`suspended_transactions` entries and unsuspends headers whose only remaining missing ancestors fell below the floor.

polinikita:
# Description of change

Bounds `BlockManager`'s in-memory state by dropping refs the network's GC floor has already passed. An ancestor at or below `DagState::gc_round_for_last_commit()` cannot influence any not-yet-sequenced block, so retaining it as "missing" or chasing it via `HeaderSynchronizer` serves no purpose.

- Adds `BlockSuspender::evict_below_round(round_floor)`. Drops `headers_to_fetch` entries below the floor; treats every `missing_ancestors` key below the floor as resolved and reuses the existing `recursively_unsuspend_dependents` cascade. Returns headers that became fully resolved.
- Adds `BlockManager::maybe_evict_below_gc_floor`, called at the top of both `try_accept_blocks` and `try_accept_block_headers`. Gated on `consensus_block_restrictions` and on the floor having advanced since the last call (`last_gc_floor_applied` field). The floor and its gauge are committed only after the sweep finishes, so a panic mid-eviction cannot leave the field claiming work that didn't complete.
- Filters sub-floor ancestors in `find_missing_ancestors` so newly arrived headers are not suspended on refs that will never matter.
- Drops incoming headers whose own round is at or below the floor in `filter_out_already_processed_and_sort`.
- Filters cascade-unsuspended headers whose own round is at or below the floor before promoting them. Keeps the GC-unsuspend path symmetric with the regular acceptance path when the floor jumps past a header that was suspended at an earlier floor.
- Retains `suspended_transactions` entries above the floor, replacing the previous unbounded-growth TODO.
- Five new `block_manager_gc_*` metrics make sweep activity visible on Grafana.

The legacy path is preserved verbatim when the flag is off: `Option<Round>` gates make the round filters true no-ops, and `maybe_evict_below_gc_floor` short-circuits to a `vec![]` return before touching DagState.

## Follow-up

Now that BlockManager actively drops sub-floor missing ancestors instead of fetching them, callers that walk the DAG via DagState must stop traversing at `gc_round_for_last_commit()`. Headers below that floor may legitimately be absent — because BlockManager will never backfill them. A follow-up PR will audit DagState ancestor walks (linearizer, commit syncer, voter checks, …) and add a uniform `gc_round` cutoff so no path can panic / spin / report-as-missing on ancestors below the floor.

Fixes #11401

- 4 new unit tests in `block_suspender.rs`: single-dep eviction, mixed (one above/below floor), cascade chain, fetch-list pruning
- 7 new tests in `block_manager/mod.rs`: header with only old missing ancestors accepted directly, no suspension of old-block transactions, merge dedup/sort for accepted+GC-unsuspended, suspended-tx eviction below floor, sweep idempotent when floor unchanged, no-op when flag off, stale-header filter on GC-unsuspend path
- 291 starfish-core tests pass


crates/starfish/core/src/block_manager/block_suspender.rs
```diff
@@ -387,12 +397,72 @@ impl BlockSuspender {
...
+        let pivot = round_floor
+            .checked_add(1)
+            .map(|round| BlockRef::new(round, AuthorityIndex::MIN, BlockHeaderDigest::MIN));
+
+        let fetch_entries_evicted = if let Some(pivot) = pivot {
+            let kept_to_fetch = self.headers_to_fetch.split_off(&pivot);
+            std::mem::replace(&mut self.headers_to_fetch, kept_to_fetch).len()
+        } else {
+            std::mem::take(&mut self.headers_to_fetch).len()
+        };
```
VorobyevIlya: So we use checked_add that can produce None in case the round is maximal, and then we clear everything? Why? This is not obvious behavior for me. Do we ever pass maximal round?

VorobyevIlya: btw, in maybe_evict_below_gc_floor we just use ordinary +1, and not checked_add. Maybe use saturated in both cases

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -68,6 +105,53 @@ impl BlockManager {
...
+        // Gated on `consensus_block_restrictions`. Off the flag, BlockManager
+        // retains its original "fetch every missing ancestor forever" behavior.
+        if !self.context.protocol_config.consensus_block_restrictions() {
+            return vec![];
+        }
+        let gc_floor = self.dag_state.read().gc_round_for_last_commit();
+        if gc_floor <= self.last_gc_floor_applied {
+            return vec![];
+        }
+        self.last_gc_floor_applied = gc_floor;
```
VorobyevIlya: Nit: I would prefer to have this line after actual eviction

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -211,13 +303,26 @@ impl BlockManager {
         }
 
         if let Some(blocks) = blocks {
+            // Mirrors the gate in `filter_out_already_processed_and_sort`: when
+            // the hardening flag is on, a block at or below the GC floor cannot
```
VorobyevIlya: hardening flag -> consensus_block_restrictions flag

VorobyevIlya: actually it appears in several places

crates/starfish/core/src/block_manager/block_suspender.rs
VorobyevIlya: addressed

crates/starfish/core/src/block_manager/mod.rs
VorobyevIlya: addressed

crates/starfish/core/src/block_manager/mod.rs
VorobyevIlya: done

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -68,6 +105,58 @@ impl BlockManager {
...
+            .inc_by(outcome.fetch_entries_evicted as u64);
+        metrics
+            .block_manager_gc_unsuspended_total
+            .inc_by(outcome.unsuspended_headers.len() as u64);
+
+        self.last_gc_floor_applied = gc_floor;
+        metrics.block_manager_gc_floor.set(gc_floor as i64);
+
+        outcome.unsuspended_headers
     }
```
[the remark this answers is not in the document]
VorobyevIlya: done

VorobyevIlya (approved):

PR #11405, opened 2026-05-04
chore(gRPC): remove stale system state TODO

### LINKED ISSUE #10077: gRPC: `IotaSystemState` proto type definition and `ProtocolConfig` round-trip conversion
Two incomplete items in `crates/iota-grpc-types/src/proto/iota/grpc/v0/epoch.rs`:

### 1. `IotaSystemState`: needs a structured proto message

- `Epoch.bcs_system_state` currently uses the generic `BcsData` wrapper (opaque bytes)
- The Rust type `IotaSystemState` (enum V1/V2) exists in `iota-types` but is not in `iota-sdk-types`: clients outside the node codebase cannot deserialize the blob
- This is a Move object (`0x3::iota_system::SystemState`) whose layout varies by protocol version, with deeply nested types (ValidatorSet, StorageFund, SystemParameters, etc.)
- Possible solution: Define a structured proto message (e.g., `SystemState`) exposing the fields consumers need (epoch info, validator set, staking data, treasury cap, gas price, etc.)

### 2. `ProtocolConfig`: needs `From/TryFrom` conversion

- Proto message already exists with structured fields (`protocol_version`, `feature_flags: map<string, bool>`, `attributes: map<string, string>`)
- Individual getters work (`protocol_version()`, `feature_flags()`, `protocol_attributes()`)
- Missing: round-trip conversion to `iota_protocol_config::ProtocolConfig`, which has 100+ individually typed fields while the proto flattens all values into `map<string, string>`
- Possible solution: Implement `From`/`TryFrom` between proto `ProtocolConfig` and `iota_protocol_config::ProtocolConfig`

bingyanglin:
# Description of change

Remove staled comment. Reasons follow

- `IotaSystemState` is already versioned inside the BCS bytes.
- `ProtocolConfig` uses `map<string, string>` on purpose already so that adding new protocol fields doesn't break clients.

fixes #10077

### DIFF
--- a/crates/iota-grpc-types/src/proto/iota/grpc/v1/epoch.rs
+++ b/crates/iota-grpc-types/src/proto/iota/grpc/v1/epoch.rs
@@ -108,13 +108,6 @@ impl Epoch {
     /// past epochs) or the current state (for the current epoch).
     ///
     /// Requires `bcs_system_state` in the read_mask.
-    // TODO: Implement when IotaSystemState type is available in iota-sdk-types.
-    // Use `system_state_bcs()` for raw bytes access in the meantime.
-    // See https://github.com/iotaledger/iota/issues/10077
-    //
-    // pub fn system_state(&self) -> Result<iota_sdk_types::IotaSystemState,
-    // TryFromProtoError> {     ...
-    // }
     pub fn system_state_bcs(&self) -> Result<&[u8], TryFromProtoError> {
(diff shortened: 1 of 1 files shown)


jkrvivian (approved):

muXxer:
I talked to the dev-tools team. The issue is still valid. I will close the PR.

PR #11409, opened 2026-05-04
feat(iota-types)!: replace CallArg by Input

Thoralf-M:

This PR replaces `CallArg` (previously defined in `iota-types`) with the
original name `CallArg`, so most existing imports continue to work
unchanged.

## Summary

- `iota-types::CallArg` and its nested `ObjectArg` enum are removed in favor of
  a type alias `pub type CallArg = iota_sdk_types::Input;`.
- The SDK's `Input` enum is extended with the constants and helper methods
  previously living on `CallArg`/`ObjectArg` (system state, clock, authenticator
  state, pure constructor, accessors), so node code can keep using them through
  the alias.
- A new `SharedObjectReference` type is introduced in `iota-sdk-types` and used
  as the payload of `Input::Shared`, replacing the previous struct-style
  variant. `iota-types` re-exports it as `SharedObjectRef`.
- Call sites in the node, indexer, JSON-RPC, GraphQL, benchmarks, e2e tests and
  related crates are adjusted to the new variant shapes (`ImmutableOrOwned`,
  `Shared(SharedObjectReference)`, tuple-style `Pure`) and to the
  `CallArgExt` trait that carries the helpers that no longer fit on the SDK
  type directly.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `Input` gains constants (`IOTA_SYSTEM_MUTABLE`,
      `CLOCK_IMMUTABLE`, `CLOCK_MUTABLE`, `AUTHENTICATOR_STATE_MUTABLE`),
      a `pure()` constructor and `*_opt` accessors. `Input::Shared` now wraps
      the new `SharedObjectReference` struct, and `Input::Pure` is a tuple
      variant.
- [ ] gRPC:


crates/iota-types/src/auth_context/fields_v1.rs
```diff
@@ -130,23 +130,23 @@ pub enum MoveObjectArg {
...
-            } => MoveObjectArg::SharedObject {
-                id: *id,
-                initial_shared_version: *initial_shared_version,
-                mutable: *mutable,
-            },
-            ObjectArg::Receiving(r) => MoveObjectArg::Receiving(*r),
-        }
-    }
-}
+// impl From<&ObjectArg> for MoveObjectArg {
```
thibault-martinez: 👀

crates/iota-types/src/auth_context/fields_v1.rs
```diff
@@ -279,39 +289,42 @@ mod tests {
...
-    #[test]
-    fn object_arg_bcs_compatible_receiving() {
-        let tx_arg = ObjectArg::Receiving(obj_ref());
-        let ctx_arg = MoveObjectArg::from(&tx_arg);
-        assert_eq!(
-            bcs::to_bytes(&tx_arg).unwrap(),
-            bcs::to_bytes(&ctx_arg).unwrap()
-        );
-    }
+    // TODO(thibault): commented during a rebase
```
thibault-martinez: 👀

Thoralf-M: Why did you comment it?

thibault-martinez: Probably didn't compile and the change was too much for a rebase, I don't remember

crates/iota-types/src/auth_context/fields_v1.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/11409/commits/51d795139abb6eb96bcc9aedfd18cf3995b0610d

crates/iota-types/src/auth_context/fields_v1.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/11409/changes/a4aadd8c8360651eb34fdcc9ff2703eabf904833

kodemartin (approved):
looks good for infra-related changes

jkrvivian (approved):

crates/iota-types/src/transaction.rs
```diff
@@ -588,151 +557,53 @@ impl EndOfEpochTransactionKind {
...
-
-    pub fn id(&self) -> ObjectID {
-        match self {
-            ObjectArg::Receiving(object_ref) | ObjectArg::ImmOrOwnedObject(object_ref) => {
-                object_ref.object_id
+            CallArg::ImmutableOrOwned(_) | CallArg::Shared(_) | CallArg::Receiving(_) => {
+                // No validation needed for these variants
             }
-            ObjectArg::SharedObject { id, .. } => *id,
+            _ => unimplemented!("a new CallArg variant was added and needs to be handled"),
```
valeriyr, line 604: This comment is probably related not only to this PR, but I would like to discuss it.
I see a problem in adding default branches like this.

By default, the Rust compiler enforces exhaustive matching: every variant of an enum must be handled in a match expression. If any variant is omitted, `compilation` fails. However, introducing a catch-all branch (such as _) bypasses this check. As a result, missing cases are no longer caught at compile time and may instead surface as runtime errors, increasing the risk of unexpected crashes.

Do we really want to do this?

thibault-martinez: I see your point, but we don't really have a choice. Since we are going to publish the SDK, we have to make everything future-proof so we never have to break the Rust API. This means adding `#[non_exhaustive]` to public enums, and having to handle it with a catch-all branch downstream. If we don't add the attribute, we will **never** be able to add a new variant if need be.

valeriyr (approved):

thibault-martinez (approved):

semenov-vladyslav (approved):

filipdulic (approved):

DaughterOfMars (approved):

PR #10552, opened 2026-03-02
feat: replace `GasData` with SDK version

thibault-martinez:
## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

PR #11423, opened 2026-05-05
feat: replace `GasData` with SDK version (#10552)

thibault-martinez:

This PR replaces `GasData` (previously defined in `iota-types`) with the
original name `GasData` (the SDK calls it `GasPayment`), so most existing
imports continue to work unchanged.

The one unavoidable rename is the `payment` field, which is called `objects`
on the SDK type. Call sites that constructed or read `gas_data.payment`
have been updated to use `gas_data.objects`.

The `iota-sdk-types` dependency is bumped to pull in the necessary type.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: the `GasData` field `payment` has been renamed to `objects` to match the SDK type (re-exported from `iota-sdk-types::GasPayment`).
- [ ] gRPC:


DaughterOfMars (approved):

valeriyr (approved):

crates/iota-types/src/transaction.rs
```diff
@@ -19,8 +19,8 @@ use fastcrypto::{encoding::Base64, hash::HashFunction};
 use iota_protocol_config::ProtocolConfig;
 pub use iota_sdk_types::{
     Argument, ChangeEpoch, ChangeEpochV2, ChangeEpochV3, ChangeEpochV4, EndOfEpochTransactionKind,
-    RandomnessStateUpdate, SharedObjectReference as SharedObjectRef, SystemPackage,
-    TransactionExpiration,
+    GasPayment as GasData, RandomnessStateUpdate, SharedObjectReference as SharedObjectRef,
```
semenov-vladyslav, line 22: Why not rename `GasData` to `GasPayment` as well? Would be less confusing.

semenov-vladyslav (approved):

crates/iota-types/src/transaction.rs
thibault-martinez: We will eventually do it. We were initially asked to reimported with the previous name to avoid having to change all imports which would create bigger PRs.

Alex6323 (approved):

kodemartin (approved):

piotrm50 (approved):

PR #11425, opened 2026-05-05
feat: replace `ConsensusCommitPrologueV1` with SDK version

Alex6323:

This PR replaces `ConsensusCommitPrologueV1` (previously defined in
`iota-types`) with the SDK version from `iota-sdk-types`. To keep the diff
small and avoid touching every call site, the SDK type is re-exported from
`iota-types` under the original name `ConsensusCommitPrologueV1`, so most
existing imports continue to work unchanged.

The same approach is applied to the related types pulled in alongside
`ConsensusCommitPrologueV1`:

- `ConsensusDeterminedVersionAssignments`
- `CancelledTransaction` (new named type, previously an inline tuple)
- `VersionAssignment` (new named type, previously an inline `(ObjectID, SequenceNumber)` tuple)
- `TimestampMs` (previously a type alias for `u64`)

All are re-exported from `iota_types::messages_consensus`.

## Notable changes

- The SDK rev (`iota-sdk-types`) is bumped to pull in the consolidated
  `ConsensusCommitPrologueV1` definition.
- `ConsensusDeterminedVersionAssignments::CancelledTransactions` changes from
  a tuple variant wrapping
  `Vec<(TransactionDigest, Vec<(ObjectID, SequenceNumber)>)>` to a struct
  variant `{ cancelled_transactions: Vec<CancelledTransaction> }`. Each entry
  is now a named `CancelledTransaction { digest, version_assignments }`,
  with `version_assignments: Vec<VersionAssignment>` instead of the previous
  unnamed `(ObjectID, SequenceNumber)` tuples.
- Call sites in `iota-core` (epoch store, consensus handler, shared object
  version manager, transaction input loader) are updated to construct and
  destructure these named types instead of raw tuples.
- The conversion code in `iota-types/src/iota_sdk_types_conversions.rs` for
  `ConsensusCommitPrologueV1` collapses to a no-op, since the internal and
  SDK types are now identical.
- The JSON-RPC `IotaConsensusDeterminedVersionAssignments` external shape is
  preserved; only the internal `From`/`Into` conversion is updated to
  destructure the new named fields.
- The staged BCS schema (`crates/iota-core/tests/staged/iota.yaml`) is
  regenerated: `CancelledTransaction` and `VersionAssignment` now appear as
  named tuple structs, and `ConsensusDeterminedVersionAssignments` references
  them.

## Test plan

- `cargo build` and `cargo test` for the affected crates (`iota-types`,
  `iota-core`, `iota-json-rpc-types`, `iota-benchmark`,
  `iota-single-node-benchmark`).
- BCS snapshot test (`crates/iota-core/tests/staged/iota.yaml`) regenerated
  and reviewed.
- Existing unit tests for shared object version assignment, consensus commit
  prologue handling, and gas price feedback updated to use the new named
  types and pass.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `ConsensusCommitPrologueV1`, `ConsensusDeterminedVersionAssignments`,
  `CancelledTransaction`, `VersionAssignment`, and `TimestampMs` are now
  re-exported from `iota_types::messages_consensus` and originate in
  `iota-sdk-types`. Downstream consumers can depend on the SDK directly
  instead of `iota-types`.
- [ ] gRPC:


crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3288,33 +3288,32 @@ impl AuthorityPerEpochStore {
...
-            Vec::new();
+        let mut version_assignment: Vec<CancelledTransaction> = Vec::new();
 
         let mut shared_input_next_version = HashMap::new();
         for txn in transactions.iter() {
             match cancelled_txns.get(txn.digest()) {
                 Some(CancelConsensusCertificateReason::CongestionOnObjects { .. })
                 | Some(CancelConsensusCertificateReason::DkgFailed) => {
-                    let assigned_versions = SharedObjVerManager::assign_versions_for_certificate(
+                    let version_assignments = SharedObjVerManager::assign_versions_for_certificate(
```
piotrm50, line 3298: it's a bit confusing having this renamed, because above we have singular `version_assignment` and here we create plural `version_assignments` and later indirectly push it to `version_assignment`. I would revert the rename here and stick to `assigned_versions` for better readability.

valeriyr (approved):

Thoralf-M (approved):
Approving the iota-types changes

kodemartin (approved):

jkrvivian (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
Alex6323: the naming is a bit weird I agree, but are you fine with renaming `version_assignment` in L3291 to `cancelled_transactions`? That seems the logical thing to do when looking at its type.

Alex6323: So I would suggest keeping this rename because it's consistent with the name of the `CancelledTransaction` field, and instead rename the currently singular `version_assignment` variable to `cancelled_transactions`. Yes, I know we have the method parameter `cancelled_txns` as well, but I would prefer that tbh. But of course if you want to revert the rename, I'll do it. Just want to confirm!

piotrm50: `cancelled_transactions` works as well.

Alex6323: changed in https://github.com/iotaledger/iota/pull/11425/commits/4b2ba8d75e08b900b042d5c0f8e4bdcaefe2e2c1

piotrm50 (approved):

thibault-martinez (approved):

PR #10795, opened 2026-03-19
feat: replace MoveObject with SDK version

thibault-martinez:
# Description of change

Replace MoveObject with SDK version

PR #11433, opened 2026-05-06
feat(iota-types)!: replace ProgrammableTransaction

Thoralf-M:

This PR replaces `ProgrammableTransaction` (previously defined in
`iota-types::transaction`) with the SDK version from `iota-sdk-types`. To keep
the diff small and avoid touching every call site, the SDK type is re-exported
from `iota-types::transaction` under the original name `ProgrammableTransaction`,
so most existing imports continue to work unchanged.

## Description of change

- `iota-types::transaction::ProgrammableTransaction` is now a re-export of
  `iota_sdk_types::ProgrammableTransaction`; the local
  `ProgrammableTransaction { inputs: Vec<CallArg>, commands: Vec<Command> }`
  struct has been removed. The SDK type has the same fields and wire format
  and now also derives `Hash` (added in
  [iota-rust-sdk#996](https://github.com/iotaledger/iota-rust-sdk/pull/996)),
  so it matches the local definition exactly.
- Because inherent impls and `Display` cannot be defined on a foreign type,
  the methods that lived on the local `impl ProgrammableTransaction`
  (`input_objects`, `receiving_objects`, `validity_check`,
  `shared_input_objects`, `move_calls`,
  `non_system_packages_to_be_published`) plus the `Display` impl have moved
  into a new extension trait `ProgrammableTransactionExt` in
  `iota-types::transaction`. `Display` is replaced by an
  `fmt_display(&mut dyn Write)` method on the trait, called from the
  `Display` impl on `TransactionKind`. Call sites that use these methods
  now bring `ProgrammableTransactionExt` into scope (updated in
  `iota-transaction-builder` and `iota-transaction-checks`).
- The `iota_sdk_types_conversions.rs` round-trip for
  `TransactionKind::ProgrammableTransaction` no longer rebuilds the value
  field-by-field — `iota-types::transaction::ProgrammableTransaction` and
  `iota_sdk_types::ProgrammableTransaction` are now the same type, so the
  conversion is an identity passthrough (analogous to `MoveObject` /
  `MoveStruct` after
  [#10795](https://github.com/iotaledger/iota/pull/10795)).
- `MoveCallExt::input_objects` is simplified slightly to read fields off
  `self` directly instead of destructuring.
- `iota-sdk-types` rev bumped to `edf0c1fe…` in the workspace `Cargo.toml`,
  `docs/examples/rust/Cargo.toml`, and `examples/tic-tac-toe/cli/Cargo.toml`.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::transaction::ProgrammableTransaction` is now a
  re-export of `iota_sdk_types::ProgrammableTransaction`. Methods that
  previously lived on the inherent impl (`input_objects`,
  `receiving_objects`, `validity_check`, `shared_input_objects`,
  `move_calls`, `non_system_packages_to_be_published`) have moved to a new
  `ProgrammableTransactionExt` trait in `iota-types::transaction`; the
  `Display` impl is replaced by `ProgrammableTransactionExt::fmt_display`.
  Call sites need to import `ProgrammableTransactionExt` to use these
  methods.
- [ ] gRPC:


DaughterOfMars (approved):

piotrm50 (approved):

DaughterOfMars (approved):

semenov-vladyslav (approved):

kodemartin (approved):

valeriyr (approved):

Alex6323 (approved):

crates/iota-types/src/transaction.rs
```diff
@@ -613,8 +597,18 @@ impl CommandExt for Command {
     }
 }
 
-impl ProgrammableTransaction {
-    pub fn input_objects(&self) -> UserInputResult<Vec<InputObjectKind>> {
+pub trait ProgrammableTransactionExt {
+    fn input_objects(&self) -> UserInputResult<Vec<InputObjectKind>>;
+    fn receiving_objects(&self) -> Vec<ObjectRef>;
+    fn validity_check(&self, config: &ProtocolConfig) -> UserInputResult;
+    fn shared_input_objects(&self) -> impl Iterator<Item = SharedObjectRef> + '_;
```
roman1e2f5p8s:
```suggestion
    fn shared_input_objects(&self) -> impl Iterator<Item = SharedObjectRef>;
```
for consistency with `non_system_packages_to_be_published` (or add it there as well) and in line 701

crates/iota-types/src/transaction.rs
roman1e2f5p8s: maybe not for this PR (except aligning the lifetime), but how about this?

```rust
    fn shared_input_objects(&self) -> impl Iterator<Item = SharedObjectRef> {
        self.inputs.iter().filter_map(|arg| match arg {
            CallArg::Shared(shared) => Some(*shared),
            CallArg::Pure(_) | CallArg::Receiving(_) | CallArg::ImmutableOrOwned(_) => None,
            _ => unimplemented!("a new CallArg variant was added and needs to be handled"),
        })
    }
```

roman1e2f5p8s (approved):

crates/iota-types/src/transaction.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/11433/commits/4dd7220828776f9ecb527e1b79d6566b1e736241

crates/iota-types/src/transaction.rs
Thoralf-M: Sure, https://github.com/iotaledger/iota/pull/11433/commits/4dd7220828776f9ecb527e1b79d6566b1e736241

thibault-martinez (approved):

Issue #11435, opened 2026-05-06
[iota-data-ingestion-core]: better ergonomics for the `TransactionFilter` type


sergiupopescu199:
The ingestion core currently exposes the proto `TransactionFilter` type directly to consumers. Building a filter is verbose, non-intuitive, and leaks proto types into user code.

A simple "Programmable AND Success AND sender(Alice)" filter looks like this today:

```rust
use iota_grpc_types::v1::filter as proto;

// Programmable
let mut kinds = proto::TransactionKindsFilter::default();
kinds.push_kinds(proto::TransactionKind::ProgrammableTransaction);
let programmable = proto::TransactionFilter::default().with_transaction_kinds(kinds);

// Success
let success = proto::TransactionFilter::default().with_execution_status(
    proto::ExecutionStatusFilter::default().with_success(true),
);

// sender(alice)
let sender = proto::TransactionFilter::default()
    .with_sender(proto::AddressFilter::default().with_address(alice));

// AND of all three
let filter = proto::TransactionFilter::default().with_all(
    proto::AllTransactionFilter::default()
        .with_filters(vec![programmable, success, sender]),
);
```
suboptimal points
- Every filter requires constructing one or two intermediate proto types (e.g. `TransactionKindsFilter`, `AddressFilter`, `ExecutionStatusFilter`) to be consumed by the `TransactionFilter` methods.
- Most structs are marked as `#[non_exaustive]` making the API harder to use.
- The user code depends directly on `iota-grpc-types`, leaking the proto API into consumer crates.

## Proposed solution
Introduce a newtype wrapper around the proto `TransactionFilter` with a factory-method API:

```rust
pub struct TransactionFilter(proto::TransactionFilter);


TransactionFilter::all([
    TransactionFilter::kinds([TransactionKind::ProgrammableTransaction]),
    TransactionFilter::execution_status(true),
    TransactionFilter::sender(alice),
])
```

Issue #11419, opened 2026-05-04
[iota-data-ingestion-core]: skip filtered checkpoints (gRPC)


sergiupopescu199:
For the user perspective it should receive only the checkpoints which match the provided filter(s). Currently the gRPC client support these methods for checkpoint streams:

[stream_checkpoints](https://iotaledger.github.io/iota/iota_grpc_client/struct.Client.html#method.stream_checkpoints): always yields every checkpoint envelope, but applies filters to the contents (transactions/events). Filter matches still produce a checkpoint with empty contents. Useful when the consumer must observe every checkpoint (watermarks, ordering). In the ingestion framework, we use this method.

[stream_checkpoints_filtered](https://iotaledger.github.io/iota/iota_grpc_client/struct.Client.html#method.stream_checkpoints_filtered): entirely skips checkpoints with no matching transactions or events filters. This introduces gaps on the framework's level but its implementation is built around contiguous, in-order checkpoints, so gaps are not allowed. In order to support filters we would need to relax this checks only when filters apply. This would require some significant framework change.

Investigate and implement the solution to allow skipping checkpoints for users, without requiring significant framework changes unless strictly necessary.

sergiupopescu199:
This diagram aims to show the flow of the ingestion core's checkpoint lifecycle.

```mermaid
flowchart LR
    SRC["Remote Sources<br/>(Fullnode gRPC / Historical / Live)"]
    CR["CheckpointReader actor"]
    IE["IndexerExecutor"]
    PS["ProgressStore"]

    subgraph WP ["Worker Pool"]
        direction TB
        WK["Workers"]
        WT["simple_watermark_tracking<br/>or Reducer"]
        WK -->|"Processed Checkpoint<br/>(seq, Worker::Message)"| WT
    end

    SRC -->|"fetch"| CR
    CR -->|"channel:<br/>Arc&lt;CheckpointData&gt;"| IE
    IE -->|"fan-out<br/>Arc&lt;CheckpointData&gt;"| WP
    WT -.->|"WorkerPoolStatus::Running<br/>(task_name, watermark)"| IE
    IE -->|"save watermark"| PS
    IE -.->|"GC signal<br/>(min watermark)"| CR
```
[...]

PR #11436, opened 2026-05-06
feat(iota-data-ingestion-core): introduce fullnode gRPC filters

### LINKED ISSUE #11325: [data-ingestion-core]: support checkpoint filters for grpc source
With the integration of gRPC connection fetching checkpoints from the fullnode, the gRPC client provides a convenient way of filtering checkpoints on the node side allowing the ingestion core to receive checkpoints based on user specified filters. This expands multiple use case, for users developing custom indexers and also optimizes the data over the wire over gRPC.

Currently the fullnode `grpc_client` does expose two kind of filters: [transaction](https://iotaledger.github.io/iota/iota_grpc_types/v1/filter/transaction_filter/index.html) & [events](https://iotaledger.github.io/iota/iota_grpc_types/v1/filter/event_filter/index.html), the IndexerExecutor could be extended to allow the user to provide those filters.

sergiupopescu199:
# Description of change

This PR adds an optional `TransactionFilter` parameter to the `IndexerExecutor`, forwarded to the fullnode's `stream_checkpoints` gRPC call when checkpoints are streamed from a `RemoteUrl::Fullnode` source.

Filters are only supported for `RemoteUrl::Fullnode`. Configuring a filter together with `RemoteUrl::HybridHistoricalStore` is rejected at startup with `IngestionError::Unsupported`.

Introduced a `config` module in the `reader`, it contains two main types `CheckpointReaderConfig` (re-exported from `reader::v2`) and `CheckpointReaderConfigExt`, the latter supports the new filter configuration option needed for the checkpoint reader and could be extended without any breaking change in the future is configuration would require new config options.

### Behavior when a transaction filter is provided:

- **Filtering happens server-side** on the fullnode. The ingestion framework performs no client-side filtering.
- **Every checkpoint still delivered** to the executor. Sequence contiguity is preserved so the framework's watermarks, ordering and gap detection continue to work unchanged.
- **Within each checkpoint, the `transactions` vector contains only transactions matching the filter.** If no transaction in the checkpoint matches, the vector is empty.
- **`checkpoint_summary` and `checkpoint_contents` are always populated** regardless of whether any transaction matched.

### Known limitation / follow-up

- Checkpoints whose contents fully fail to match the filter still travel over the wire as empty transaction vector. A follow-up issue will explore fixing this limitation while still preserving the contiguity guarantee the executor relies on. (#11419 )
- The public API currently leans on the proto `TransactionFilter` type. From a user perspective, building a filter is verbose, non-intuitive, and leaks proto types into the consumer's code. A follow-up issue will introduce a dedicated wrapper API (#11435).

fixes #11325 

> [!NOTE]
> This patch does not affect the normal operations of the indexer, thus no further tests were conducted.

### DIFF
--- a/crates/iota-data-ingestion-core/src/config.rs
+++ b/crates/iota-data-ingestion-core/src/config.rs
@@ -1,0 +1,82 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Configuration options for the ingestion framework.
+
+use std::path::PathBuf;
+
+use crate::{
+    IngestionLimit, ReaderOptions,
+    filters::fullnode::TransactionFilter,
+    reader::v2::{CheckpointReaderConfig, RemoteUrl},
+};
+
+/// Configuration options for the ingestion framework.
+#[derive(Clone, Default)]
+pub struct IngestionConfig {
+    /// Config the checkpoint reader behavior for downloading new checkpoints.
+    pub(crate) reader_options: ReaderOptions,
+    /// Local path for checkpoint ingestion. If not provided, checkpoints will
+    /// be ingested from a temporary directory.
+    pub(crate) ingestion_path: Option<PathBuf>,
+    /// Remote source for checkpoint data stream.
+    pub(crate) remote_store_url: Option<RemoteUrl>,
+    /// Determines when the ingestion process should stop.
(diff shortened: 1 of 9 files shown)


tomxey (approved):

kodemartin (approved):
lgtm 🎈

PR #11442, opened 2026-05-06
feat(indexer): graceful shutdown on SIGINT/SIGTERM

### LINKED ISSUE #8819: feat(indexer): Use graceful shutdown capabilities of data ingestion worker in iota-indexer
We should have e.g. tokio::signal::ctrl_c() signal handler to trigger graceful shutdown instead of terminating right away.

tomxey:
# Description of change

Wire OS signals (Ctrl+C and SIGTERM) into the existing CancellationToken-based graceful shutdown of the data ingestion executor and JSON-RPC server. Previously the binary exited immediately, which could interrupt in-flight checkpoints mid-write.

A single master CancellationToken is created in main and threaded into both Indexer::start_writer_with_config (already accepted one) and Indexer::start_reader (new parameter, forwarded into build_json_rpc_server, replacing the token that function used to construct internally).

Mirrors the signal-handler pattern in iota-data-ingestion's main.

fixes #8819

### DIFF
--- a/crates/iota-indexer/tests/common/mod.rs
+++ b/crates/iota-indexer/tests/common/mod.rs
@@ -54,6 +54,7 @@ use tokio::{
     sync::{Mutex, OnceCell},
     task::JoinHandle,
 };
+use tokio_util::sync::CancellationToken;
 
 const DEFAULT_DB: &str = "iota_indexer";
 const DEFAULT_INDEXER_IP: &str = "127.0.0.1";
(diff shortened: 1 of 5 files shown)


crates/iota-indexer/tests/common/mod.rs
```diff
@@ -450,9 +450,17 @@ fn start_indexer_reader(fullnode_rpc_url: impl Into<String>, database_name: Opti
...
-        async move { Indexer::start_reader(&config, store, &registry, pool, metrics).await },
-    );
+    tokio::spawn(async move {
+        Indexer::start_reader(
+            &config,
+            store,
+            &registry,
+            pool,
+            metrics,
+            tokio_util::sync::CancellationToken::new(),
```
sergiupopescu199: We could import it on the top level and use `CancellationToken::new()` in the fn argument

sergiupopescu199 (approved):
lgtm ✨

kodemartin (approved):
lgtm 🌸

PR #11444, opened 2026-05-06
feat(move): Move view functions (compiler implementation) - Linter Warning

### LINKED ISSUE #11413: [move] View functions - linter warning
Add a linter warning for potential view functions

lollobene:
# Description of change

Linter warning implementation

This change gates the Iota view-function suggestion lint behind `--lint`.

`ViewFunctionVisitor` is no longer included in the default Iota linter set, so normal builds do not emit warnings for functions that could be annotated with` #[view]`. The visitor now only runs at `LintLevel::All`, which is enabled through the existing `--lint` flag.

fixes #11413


external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
```diff
@@ -683,6 +700,10 @@ fn view_return_ty(context: &mut Context, view_loc: Loc, name: FunctionName, retu
     }
 }
 
+fn is_valid_view_return_ty(return_type: &Type) -> bool {
```
bingyanglin, line 703: Is the reference check missing here? The original `view_return_ty` validation (see above L680) checks `contains_reference_ty`, but this helper doesn't. The linter will suggest `#[view]` for functions that return references, and then the actual annotation will fail with "View functions cannot return references."
Or do I miss anything?

external-crates/move/crates/move-compiler/tests/iota_mode/linter/view_function/suggestions.move
```diff
@@ -0,0 +1,57 @@
...
+        a * b
+    }
+
+    public fun wrapper_ref(w: &mut Wrapper): u64 {
+        let old = w.obj.value;
+        w.obj.value = 42;
+        old
+    }
+
+    public native fun returns_object(): Obj;
```
bingyanglin: The test doesn't cover functions that return references. That's exactly the case the bug above would catch. Maybe add something like `public fun get_ref(data: &Data): &u64 { &data.value }` and verify it does NOT get the warning?

external-crates/move/crates/move-compiler/src/iota_mode/linters/view_function.rs
```diff
@@ -0,0 +1,58 @@
...
+            return false;
+        }
+
+        let msg = format!(
+            "Function '{}' satisfies view constraints and can be annotated with '#[view]'",
+            fname
+        );
+        let mut d = diag!(VIEW_FUNCTION_DIAG, (fname.loc(), msg));
+        d.add_note(format!(
+            "Add '#[view]' to make the function explicitly callable as a view function."
```
bingyanglin: Not sure about why we need `format!` macro here? Can we pass the string directly?

lollobene: yes you are right!

external-crates/move/crates/move-compiler/src/iota_mode/typing.rs
lollobene: true, thanks for this!

external-crates/move/crates/move-compiler/src/iota_mode/linters/mod.rs
```diff
@@ -178,6 +187,7 @@ pub fn linter_visitors(level: LintLevel) -> Vec<Visitor> {
         LintLevel::All => {
             let mut visitors = linter_visitors(LintLevel::Default);
             visitors.extend([
+                view_function::ViewFunctionVisitor.visitor(),
```
miker83z, line 190: This is basically the "lower level", right?

miker83z (approved):
LGTM!

external-crates/move/crates/move-compiler/src/iota_mode/linters/mod.rs
lollobene: Yes, I moved the ViewFunctionVisitor to the `LintLevel::All` variant case, which means that it is executed only when the compiler is run with the `--lint` flag

PR #11450, opened 2026-05-06
feat(node): reduce log spamming

muXxer:
# Description of change

This PR massively reduces the log spamming in the `iota-node` on `"info"` level.
It introduces a periodic status message that is printed every second, that reports about progress of synced and pruned checkpoints, instead of prints per every single received, executed and pruned checkpoint.

It also removes the log spam if peers don't have the range of checkpoints we want to sync, which was repeatedly 400 lines every 10s.

The `lowest` available checkpoint is now calculated based on the information from the connected peers on the same chain. Before it was based on all connected peers, which was wrong.

We reduced the log level from `info` to `debug` for `"scheduling pruning for checkpoint"`, `"executing checkpoint"` and `"received synced checkpoint"`. The Log message `"Finished syncing and executing checkpoint"` was removed and replace by the `CheckpointProgressTracker` message.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Added periodic checkpoint progress status message and reduced log spam
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-core/src/authority/authority_store_pruner.rs
+++ b/crates/iota-core/src/authority/authority_store_pruner.rs
@@ -476,7 +476,6 @@ impl AuthorityStorePruner {
         let mut effects_to_prune = vec![];
 
         let mut pruning_start = Instant::now();
-        let mut total_checkpoints_pruned: u64 = 0;
 
         loop {
             let Some(ckpt) = checkpoint_store
@@ -519,8 +518,6 @@ impl AuthorityStorePruner {
             if effects_to_prune.len() >= config.max_transactions_in_batch
                 || checkpoints_to_prune.len() >= config.max_checkpoints_in_batch
             {
-                total_checkpoints_pruned += checkpoints_to_prune.len() as u64;
-
                 match mode {
                     PruningMode::Objects => {
                         Self::prune_objects(
@@ -565,8 +562,6 @@ impl AuthorityStorePruner {
         }
 
         if !checkpoints_to_prune.is_empty() {
-            total_checkpoints_pruned += checkpoints_to_prune.len() as u64;
-
             match mode {
                 PruningMode::Objects => {
                     Self::prune_objects(
@@ -589,9 +584,10 @@ impl AuthorityStorePruner {
                     metrics.clone(),
                 )?,
             };
-        }
 
-        if total_checkpoints_pruned > 0 {
+            // Report pruning time for this batch so the progress logger
+            // shows time alongside the checkpoint deltas it reads from the
+            // DB (which are already updated at this point).
             if let Some(tracker) = progress_tracker {
                 let elapsed = pruning_start.elapsed();
                 match mode {
@@ -600,6 +596,7 @@ impl AuthorityStorePruner {
                 }
             }
         }
+
         Ok(())
     }
 
--- a/crates/iota-core/src/authority/authority_store_pruner.rs
+++ b/crates/iota-core/src/authority/authority_store_pruner.rs
@@ -40,6 +40,7 @@ use typed_store::{
 use super::authority_store_tables::{AuthorityPerpetualTables, AuthorityPrunerTables};
 use crate::{
     authority::authority_store_types::{StoreObject, StoreObjectWrapper},
+    checkpoint_progress_tracker::CheckpointProgressTracker,
     checkpoints::{CheckpointStore, CheckpointWatermark},
     grpc_indexes::GrpcIndexesStore,
     jsonrpc_index::IndexStore,
@@ -340,6 +341,7 @@ impl AuthorityStorePruner {
         config: AuthorityStorePruningConfig,
         metrics: Arc<AuthorityStorePruningMetrics>,
         epoch_duration_ms: u64,
+        progress_tracker: Option<&Arc<CheckpointProgressTracker>>,
     ) -> anyhow::Result<()> {
         let _scope = monitored_scope("PruneObjectsForEligibleEpochs");
         let (mut max_eligible_checkpoint_number, epoch_id) = checkpoint_store
@@ -370,6 +372,7 @@ impl AuthorityStorePruner {
             max_eligible_checkpoint_number,
             config,
             metrics.clone(),
+            progress_tracker,
         )
         .await
     }
@@ -391,6 +394,7 @@ impl AuthorityStorePruner {
         metrics: Arc<AuthorityStorePruningMetrics>,
         archive_readers: ArchiveReaderBalancer,
         epoch_duration_ms: u64,
+        progress_tracker: Option<&Arc<CheckpointProgressTracker>>,
     ) -> anyhow::Result<()> {
         let _scope = monitored_scope("PruneCheckpointsForEligibleEpochs");
         let pruned_checkpoint_number = checkpoint_store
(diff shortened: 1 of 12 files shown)


crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -564,6 +590,16 @@ impl AuthorityStorePruner {
                 )?,
             };
         }
+
+        if total_checkpoints_pruned > 0 {
+            if let Some(tracker) = progress_tracker {
```
piotrm50: I think it latest rust you can write something like `if let Some(x) = var && var2 > 0 {...}`

crates/iota-core/src/checkpoint_progress_tracker.rs
```diff
@@ -0,0 +1,167 @@
...
+                    tracker.checkpoint_pruning_time_ns.load(Ordering::Relaxed);
+
+                let exec_delta = highest_executed_seq_number.saturating_sub(prev_executed);
+                let tx_delta = total_tx.saturating_sub(prev_total_tx);
+                let object_prune_delta = object_pruned_seq_number.saturating_sub(prev_obj_pruned);
+                let checkpoint_prune_delta =
+                    checkpoint_pruned_seq_number.saturating_sub(prev_ckpt_pruned);
+
+                if exec_delta > 0 || object_prune_delta > 0 || checkpoint_prune_delta > 0 {
+                    let exec_time_delta =
```
piotrm50: do we need to subtract those times? can we take the value and swap with 0 and always print what we've taken from the atomic? that way we don't ever risk overflows or anything.

bingyanglin (approved):

crates/iota-core/src/checkpoint_progress_tracker.rs
```diff
@@ -0,0 +1,167 @@
...
+                        Duration::from_nanos(exec_time_ns.saturating_sub(prev_exec_time_ns));
+                    let object_prune_time_delta = Duration::from_nanos(
+                        object_prune_time_ns.saturating_sub(prev_obj_prune_time_ns),
+                    );
+                    let checkpoint_prune_time_delta = Duration::from_nanos(
+                        checkpoint_prune_time_ns.saturating_sub(prev_ckpt_prune_time_ns),
+                    );
+                    info!(
+                        "checkpoint progress [epoch {}]: executed {}/{} (+{}, {} tx/s, {:.2?}), \
```
roman1e2f5p8s:
```suggestion
                        "checkpoint progress [epoch {}]: executed {}/{} (+{}, {} txs, {:.2?}), \
```

`tx_delta` is not a rate, right?

crates/iota-core/src/checkpoint_progress_tracker.rs
```diff
@@ -0,0 +1,167 @@
...
+                        synced_seq_number,
+                        exec_delta,
+                        tx_delta,
+                        exec_time_delta,
+                        object_pruned_seq_number,
+                        object_prune_delta,
+                        object_prune_time_delta,
+                        checkpoint_pruned_seq_number,
+                        checkpoint_prune_delta,
+                        checkpoint_prune_time_delta,
```
roman1e2f5p8s: you could write all of these identifiers directly inside `{}`, which is more idiomatic

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -564,6 +590,16 @@ impl AuthorityStorePruner {
...
+                let elapsed = pruning_start.elapsed();
+                match mode {
+                    PruningMode::Objects => tracker.add_object_pruning_time(elapsed),
+                    PruningMode::Checkpoints => tracker.add_checkpoint_pruning_time(elapsed),
+                }
+            }
```
roman1e2f5p8s, line 597: all these lines should probably be put in the `if !checkpoints_to_prune.is_empty()` block right above, removing the `if total_checkpoints_pruned > 0` block.
`total_checkpoints_pruned` is "global" and incremented across the loop, while `pruning_start` is reset in the loop.
same pattern as you have inside the loop: lines 522-557 - no additional if between pruning and reporting pruning time.

or if you meant to have the `if total_checkpoints_pruned > 0` block indeed, `pruning_start` should have been remembered before the loop, not reset inside it

crates/iota-network/src/state_sync/mod.rs
```diff
@@ -1420,6 +1513,42 @@ where
...
+    let (known_peers, connected_peers, lowest_peer_watermark, is_pruned) = {
+        let guard = peer_heights.read().unwrap();
+        let known = guard.peers_on_same_chain().count();
+        let connected_peers_lowest: Vec<CheckpointSequenceNumber> = guard
+            .peers_on_same_chain()
+            .filter(|(peer_id, _)| network.peer(**peer_id).is_some())
+            .map(|(_, info)| info.lowest)
+            .collect();
+        let connected = connected_peers_lowest.len();
+        let lowest = connected_peers_lowest.into_iter().min();
```
roman1e2f5p8s, line 1526: the `lowest` is now computed based on connected peers on the same chain, but before it was computed based on all known peers, so the numbers shown in the logs might not be the same - maybe worth adding a comment in the PR description?

roman1e2f5p8s (changes requested):
could we also mention in the PR description (maybe even in release notes) that:
 
- log level for some events changed from `info` to `debug`
- some public APIs in `iota-core` got an extra optional parameter

crates/iota-core/src/authority/authority_store_pruner.rs
muXxer: image

crates/iota-core/src/checkpoint_progress_tracker.rs
muXxer: good catch!

crates/iota-core/src/checkpoint_progress_tracker.rs
muXxer: It is. We sample/log every second, and we subtract the current value from the previous value.

crates/iota-network/src/state_sync/mod.rs
muXxer: I added a comment in the PR description.

crates/iota-core/src/checkpoint_progress_tracker.rs
muXxer: Fixed it.

crates/iota-core/src/authority/authority_store_pruner.rs
muXxer: Fixed it.

muXxer:
> could we also mention in the PR description (maybe even in release notes) that:
> 
> * log level for some events changed from `info` to `debug`
> * some public APIs in `iota-core` got an extra optional parameter
Thanks for the review!

I added in the description for the log messages that were changed.
This should not be part of the release notes IMO, otherwise we spam the release notes with unnecessary information.

The public API's in `iota-core` don't matter. `iota-core` is not a public API.

jkrvivian (approved):

roman1e2f5p8s (approved):

PR #11451, opened 2026-05-06
feat(network): improve state sync peer registration and tracking

### LINKED ISSUE #11440: node: remove old style query from `query_peer_for_latest_info`
There is leftover code, which can be removed, at the end of the function.

### LINKED ISSUE #11441: node: add endpoint to query a node for the chain_id
Instead of relying on the checkpoint summary of the genesis checkpoint, we could add an endpoint just to return the chain_id, or we add it directly to the `get_checkpoint_availability` one.

Maybe this way we can prune checkpoint 0 at some point.

muXxer:
# Description of change

  - Adds a StateSyncHandshake protocol — on connect, both sides immediately exchange chain identity (genesis checkpoint), sync height, and pruning watermark. This eliminates a race condition where push_checkpoint_summary could arrive before the peer was registered (the old flow required multiple sequential queries).
  - Fixes PeerHeights insertion — new peers are now registered with complete info (chain identity + height + lowest watermark) in a single step, and the lowest field is correctly updated on reconnect.
  - Filters disconnected peers in PeerBalancer — peers that disappeared between balancer construction and iteration are skipped, avoiding wasted requests. The RTT selection window is also widened (2 → 10).
  - Periodic PeerHeights cleanup during sync — temp maps are pruned every 10k checkpoints to prevent unbounded memory growth during long syncs.
  - Removes the legacy get_checkpoint_summary(Latest) fallback in query_peer_for_latest_info.                                                                                                                       
  - Improves error context — failed checkpoint sync now reports connected/known peer counts.        

fixes #11440 
fixes #11441 

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Hardened state sync peer tracking:
  - added a handshake protocol to eliminate race conditions during peer registration
  - fixed stale peer handling
  - added periodic cleanup to prevent memory growth during long syncs
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-network/src/state_sync/server.rs
+++ b/crates/iota-network/src/state_sync/server.rs
@@ -21,7 +21,7 @@ use iota_types::{
 use serde::{Deserialize, Serialize};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore, mpsc};
 
-use super::{PeerHeights, StateSync, StateSyncMessage};
+use super::{PeerHeights, PeerStateSyncInfo, StateSync, StateSyncMessage};
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum GetCheckpointSummaryRequest {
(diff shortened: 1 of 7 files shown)


crates/iota-network/src/state_sync/server.rs
```diff
@@ -142,6 +157,54 @@ where
...
+        let peer_id = request
+            .peer_id()
+            .copied()
+            .ok_or_else(|| Status::internal("unable to query sender's PeerId"))?;
+
+        let handshake = request.into_inner();
+
+        let our_genesis_digest = *self.genesis_checkpoint.digest();
+
+        let on_same_chain = *handshake.genesis_checkpoint.digest() == our_genesis_digest;
```
bingyanglin: The handshake trusts the peer's genesis checkpoint digest without verifying the checkpoint itself. A malicious peer could send any checkpoint with our genesis digest baked in, and we'd mark them as same-chain. Maybe call `handshake.genesis_checkpoint.verify_digest()` before accepting it? The downstream sync still verifies signatures so this isn't critical, but would it be a little bit better that we reject bad data early?

bingyanglin (approved):

jkrvivian (approved):

PR #11452, opened 2026-05-06
fix(iota-node,iota-benchmark): fix capability chain mismatch and align simtest epoch durations

### LINKED ISSUE #11438: Unignore `test_mainnet_config`
It was consistently failing and blocking me so I'm commenting it out.
https://github.com/iotaledger/iota/actions/runs/25432737305/job/74607882465?pr=11430#step:8:16757

piotrm50:
# Description of change

Fix protocol config digest mismatch in capability notifications when using `with_chain_override` in tests, and align simtest epoch durations with upstream.

**Chain mismatch bug:** When creating `AuthorityCapabilitiesV1`, the code used `epoch_store.get_chain_identifier().chain()` to determine the chain. In test clusters using `with_chain_override(Chain::Mainnet)`, the `ChainIdentifier` is derived from the genesis checkpoint digest, which doesn't match known chain digests — so `.chain()` returns `Chain::Unknown`. Meanwhile, the epoch store's protocol config is computed with the correct `Chain::Mainnet` from the override. This caused `get_validators_supporting_protocol_version` to always find 0 matching validators (different chain → different protocol config → different digest), triggering the "Eligible validators weight 0" fallback on every epoch change.

**Fix:** Use `epoch_store.get_chain()` which returns the authoritative chain value the epoch store was configured with. In production, this is identical to `get_chain_identifier().chain()` since the genesis digest matches the known chain. In tests with chain override, it correctly returns the override chain.

**Epoch duration alignment:** Several simtests had shorter epoch durations (1-5s) than upstream (10s). With short epochs and ~15s epoch transitions, the network spends most of its time in transition, leaving the surfer too few transaction windows. This caused flaky failures where all surfer transactions hit epoch boundaries.

Closes #11438

- [x] Reproduced failure with `MSIM_TEST_SEED=1778094833170` on Linux x86_64
- [x] Verified fix eliminates "Eligible validators weight 0" errors
- [x] Verified previously-failing seed `1778097890094` now passes
- [x] Seed search with 200 seeds on Linux x86_64

### DIFF
--- a/crates/iota-benchmark/tests/simtest.rs
+++ b/crates/iota-benchmark/tests/simtest.rs
@@ -111,7 +111,6 @@ mod test {
         test_simulated_load(test_cluster, 60).await;
     }
 
-    #[ignore("https://github.com/iotaledger/iota/issues/11438")]
     #[sim_test(config = "test_config()")]
     async fn test_mainnet_config() {
         chain_config_smoke_test(Chain::Mainnet).await;
@@ -124,7 +123,8 @@ mod test {
 
     async fn chain_config_smoke_test(chain: Chain) {
         iota_protocol_config::ProtocolConfig::poison_get_for_min_version();
-        let test_cluster = init_test_cluster_builder(2, 3000)
+        // 2 validators, 10 seconds per epoch.
+        let test_cluster = init_test_cluster_builder(2, 10_000)
             .with_authority_overload_config(AuthorityOverloadConfig {
                 // Disable system overload checks for the test - during tests with crashes,
                 // it is possible for overload protection to trigger due to validators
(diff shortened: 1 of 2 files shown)


muXxer (approved):

thibault-martinez (commented):
Isn't this PR supposed to unignore `test_mainnet_config` if it's closing the issue?

crates/iota-benchmark/tests/simtest.rs
```diff
@@ -111,7 +111,6 @@ mod test {
         test_simulated_load(test_cluster, 60).await;
     }
 
-    #[ignore("https://github.com/iotaledger/iota/issues/11438")]
     #[sim_test(config = "test_config()")]
```
piotrm50, line 114: @thibault-martinez the test is not currently ignored 🤔

thibault-martinez (approved):

Issue #11023, opened 2026-03-30
[indexer]: startup from formal snapshots


kodemartin:
## Description

We want to enable bootstrapping or restoring an instance of IOTA Indexer from a formal snapshot.

This will allow operators to start synchronizing the database using data from a recent epoch.

The tentative requirements for the first version of this feature are the following:

* Users should be able to define the time window for the indexed data and the network of choice
* The time window should be given in fuzzy terms between a limited set of options
   * live data: Fetch the latest formal snapshot
   * current month: Fetch the formal snapshot from the 1st of the month
   * last `n` months (`n = {2,3}`): Fetch the format snapshot from the 1st day of the month before the current (`n = 2`) , or two months before the current (`n = 3`).
* The snapshot should contain all data to populate live objects, displays,  epochs, and chain identifier
* Starting IOTA Indexer in this mode should be considered completed when ingestion has caught up with the current epoch.

> [!NOTE]
> For future iterations we may consider allowing the Indexer to start from the most recent formal snapshot, and backfill indexes for the time window request by the user.

Issue #10907, opened 2026-03-24
Update wallet for improved release flow


begonaalvarezd:
Update the apps backend in the wallet so it uses
- staging for nightly
- rc for rcs
- production for production

begonaalvarezd:
Depends on https://github.com/iotaledger/iota/issues/10904 being deployed

marc2332:
@begonaalvarezd 

This requires no changes in the code/workflows, just the secrets on github:

- Nightly: `WALLET_NIGHTLY_BUILD_APPS_BACKEND`
- RC: `WALLET_RC_APPS_BACKEND`
- Prod: `WALLET_PROD_APPS_BACKEND`

marc2332:
since we will need these in other places (explorer, dashboard) we could as well just make them more generic

- Staging/nightly: `STAGING_APPS_BACKEND`
- RC: `RC_APPS_BACKEND`
- Prod: `PROD_APPS_BACKEND`

PR #11453, opened 2026-05-07
feat(iota-core): introduce StoreObjectV2 and snapshot V2

### LINKED ISSUE #10957: gRPC: add epoch's last checkpoint summary to EpochInfo index table
The `epochs` table in `RestIndexStore` stores `EpochInfo` for each epoch. Currently it has `end_checkpoint: Option<u64>` (the sequence number) but not the actual checkpoint summary data.

When gRPC serves `GetEpoch`, it reads `EpochInfo` from the index, then does a separate DB read to load the checkpoint summary. Storing the summary directly in `EpochInfo` would remove that extra read.

See the TODO at `crates/iota-core/src/rest_index.rs` on the `epochs` table:
```
// TODO: add checkpoint summary of last checkpoint (end_of_epoch_transaction?
// epoch_commitments?)
```

### LINKED ISSUE #11254: snapshots: extended formal snapshot format
As discussed, we want to extend the information inside a formal snapshot.

We need to add another file that contains the epoch information to rebuild the epochs index (including epochs last checkpoint summary) since genesis. 

The reference file should be extended by the checkpoint sequence number that contained the transaction that created that object.

Investigation: Can the "Wrapped" enum variant in the `LiveObject` enum be removed?

bingyanglin:
# Description of change

## Summary

Bumps the formal-snapshot format to **V2** so new snapshots carry the data the indexer and other downstream consumers need to rebuild state without an archival replay:

- **Per-object previous-transaction checkpoint** inline in `.obj` records: `StoreObjectV2` on the row layer, unified `LiveObject` on the snapshot wire.
- **Per-epoch metadata** in a new `EPOCH_INFO` file alongside the bucket files.

**Scope: producer-side only.** This PR teaches a single writer node to *produce* V2 snapshots; the V1 reader path is preserved on disk and remains unchanged in behavior. The reader-side restore from `EPOCH_INFO` and the one-time backfill of `previous_transaction_checkpoint` for pre-V2 objects are the PR-2 follow-up tracked in **#10957**.

#### OBJECT files
- Magic bumped to `0x00B7EC76` (V1 was `0x00B7EC75`) — V1/V2 readers fail-fast on each other.
- Records are BCS-encoded `LiveObject` (unified type), carrying `previous_transaction_checkpoint: u64` inline alongside the live `Object`. The on-disk shape, not the Rust type name, is what changes between V1 and V2.
- `LiveObject::Wrapped` removed from the live-set view (the enum collapsed to a plain struct); `StoreObject::Wrapped` on the row layer stays (distinct `OBJECT_WRAPPED` digest).

#### REFERENCE files
- Unchanged from V1.

#### EPOCH_INFO file (new)
- Layout: `magic(0x9000C001, 4 B) | bcs(EpochInfo::V1 { entries: Vec<Option<EpochInfoEntry>> })`.
- One entry per epoch in `[0, snapshot_epoch]`.
- Integrity anchored by `FileMetadata::sha3_digest` in MANIFEST (same as `.obj`/`.ref`).

```rust
pub struct EpochInfoEntry {
    pub first_checkpoint: CheckpointSequenceNumber,
    pub start_system_state: Vec<u8>,                              // bcs(IotaSystemState)
    pub last_checkpoint_summary: Option<CertifiedCheckpointSummary>,
    pub end_of_epoch_tx_events: Option<TransactionEvents>,
}
```

`start_system_state` is opaque BCS bytes so the inner `IotaSystemStateV1/V2/…` can evolve without forcing an `EpochInfoEntry` schema change.

### Producer architecture

- Source of truth is the gRPC indexer's `epoch_info` table on `IndexStoreTables` (sibling of the existing `epochs` table). Populated by `grpc_indexes::index_epoch` from `CheckpointData` over two checkpoint boundaries:
  - **Boundary 1** (prev epoch closes): insert `first_checkpoint` + `start_system_state` for the new epoch.
  - **Boundary 2** (new epoch closes): upsert `last_checkpoint_summary` + `end_of_epoch_tx_events` and advance `Watermark::EpochIndexed`.
- Snapshot V2 writer pre-publish gate: refuses to publish unless `Watermark::EpochIndexed >= snapshot_epoch`, so emitted snapshots are complete-by-construction.

### Operator contract (writer node)

The single network node that produces the formal snapshot must:

1. Run as a **fullnode** (validators do not run grpc_indexes).
2. Have **`enable_grpc_api = true`** (so `epoch_info` is populated).
3. Run the PR-2 backfill once before publishing the first V2 snapshot (so `epoch_info` covers `[0, current_epoch]`); after that, live indexing keeps it complete forever.

All three are enforced at node startup or snapshot-publish time with operator-facing error messages.

### Compatibility / reader behavior

- V2 readers accept V2 snapshots and restore object state from `LiveObject` records that carry `previous_transaction_checkpoint` inline.
- V2 readers validate the manifest lists `EPOCH_INFO` (fail-fast on a missing entry) but **do not** download or parse the file in this PR — that's PR 2.
- V1 readers are untouched and continue to consume V1 snapshots as before.

### Follow-ups (PR 2, tracked in #10957)

~~- Reader: download + verify + parse `EPOCH_INFO`, populate the indexer's `epoch_info` table.~~
- Reader: parse `EPOCH_INFO` and dispatch each `EpochInfoEntry` through a new `Restore` trait method (mirroring the object-write generalization in #11559). Concrete consumer-side persistence lives behind the trait; the Indexer-specific impl is tracked in #11023.
- One-time backfill of `epoch_info` for pre-PR epochs (so the writer's watermark precondition can clear).
- Backfill of `previous_transaction_checkpoint` for objects in snapshots produced before this PR.

fixes #11254


crates/iota-core/src/authority/authority_store_tables.rs
```diff
@@ -541,69 +549,104 @@ impl ObjectStore for AuthorityPerpetualTables {
...
-    tables: &'a AuthorityPerpetualTables,
-    prev: Option<(ObjectKey, StoreObjectWrapper)>,
-}
-
+/// Yields the same live-object set as `LiveSetIterV2` but strips the
+/// per-object `previous_transaction_checkpoint` trailer, for callers that do
+/// not need it.
+pub struct LiveSetIter<'a>(LiveSetIterV2<'a>);
+
+/// A row that the live-set iterator surfaces. The previous `Wrapped` variant
```
semenov-vladyslav: `Wrapped` variant is no more, no need to mention it.

crates/iota-core/src/authority/authority_store_tables.rs
```diff
@@ -541,69 +549,104 @@ impl ObjectStore for AuthorityPerpetualTables {
...
+/// was removed: `LiveSetIter` filters `StoreObject::Wrapped` and
+/// `StoreObject::Deleted` from its output, so wrapped objects never reach
+/// downstream consumers (snapshot writer, state-hash accumulator, restore
+/// path).
+///
+/// Kept as a single-variant enum to bound the blast radius of this PR —
```
semenov-vladyslav: Maybe shorten this part of the comment, it's pretty trivial.

crates/iota-core/src/authority/authority_store.rs
```diff
@@ -825,10 +825,18 @@ impl AuthorityStore {
...
+    /// `previous_transaction_checkpoint` field. Callers that buffer execution
+    /// outputs until checkpoint commit time (e.g. `WritebackCache`) pass the
+    /// containing checkpoint's sequence number; callers that flush at
+    /// execution time before the containing checkpoint is known (e.g.
+    /// `PassthroughCache`) pass `SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT`.
     #[instrument(level = "debug", skip_all)]
     pub fn build_db_batch(
         &self,
         epoch_id: EpochId,
+        checkpoint_seq: CheckpointSequenceNumber,
```
semenov-vladyslav:
```suggestion
        previous_transaction_checkpoint: CheckpointSequenceNumber,
```
Can we make the naming consistent so that the meaning of the sequence number is straightforward?

crates/iota-core/src/authority/authority_store.rs
```diff
@@ -861,6 +874,7 @@ impl AuthorityStore {
         &self,
         write_batch: &mut DBBatch,
         epoch_id: EpochId,
+        checkpoint_seq: CheckpointSequenceNumber,
```
semenov-vladyslav:
```suggestion
        previous_transaction_checkpoint: CheckpointSequenceNumber,
```
Here too.

crates/iota-core/src/authority/authority_store_tables.rs
```diff
@@ -541,69 +549,104 @@ impl ObjectStore for AuthorityPerpetualTables {
...
+/// every consumer site uses `let LiveObject::Normal(o) = ...;` today, and
+/// collapsing to `pub struct LiveObject(pub Object);` (or
```
semenov-vladyslav: I would do `pub struct LiveObject(pub Object);` in this PR since all call sites are already touched after removing `Wrapped` variant.

crates/iota-core/src/authority/authority_store_types.rs
```diff
@@ -6,80 +6,93 @@ use iota_types::{
...
+/// that holds this sentinel — V1 rows lifted via `migrate()`, rows stamped
+/// on the `PassthroughCache` synchronous-write path, and rows stamped on
+/// the snapshot restore path — with the real
+/// `previous_transaction_checkpoint`.
+pub const SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT: CheckpointSequenceNumber = u64::MAX;
+
 // Versioning process:
 //
-// Object storage versioning is done lazily (at read time) - therefore we must
+// Object storage versioning is done lazily (at read time) — therefore we must
```
semenov-vladyslav: What was wrong with the old dash?

crates/iota-core/src/execution_cache/passthrough_cache.rs
```diff
@@ -368,7 +377,12 @@ impl GlobalStateHashStore for PassthroughCache {
 }
 
 impl ExecutionCacheCommit for PassthroughCache {
-    fn build_db_batch(&self, _epoch: EpochId, _digests: &[TransactionDigest]) -> Batch {
+    fn build_db_batch(
+        &self,
+        _epoch: EpochId,
+        _checkpoint_seq: CheckpointSequenceNumber,
```
semenov-vladyslav: Should the `previous_transaction_checkpoint` field of the objects written earlier be updated here. The objects are written earlier with `SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT` field value.

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -569,7 +573,9 @@ async fn test_committed() {
 
         s.assert_live(&[1, 2]);
         s.assert_dirty(&[1, 2]);
-        let batch = s.cache().build_db_batch(1, &[tx]);
+        let batch = s
+            .cache()
+            .build_db_batch(1, SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT, &[tx]);
```
semenov-vladyslav: Shouldn't the test use a proper checkpoint sequence number here and assert its value below?

crates/iota-core/src/execution_cache/passthrough_cache.rs
```diff
@@ -426,7 +440,8 @@ impl PassthroughCache {
 
         let object_ref = object.compute_object_reference();
         let object_key = ObjectKey::from(object_ref);
-        let store_object = get_store_object(object.clone());
+        let store_object =
+            get_store_object(object.clone(), SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT);
```
semenov-vladyslav: Initially PassthroughCache writes `SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT` sequence number, but it should be updated to the proper value, doesn't it? Should the tests be also updated to check for `previous_transaction_checkpoint` field value?

crates/iota-snapshot/src/lib.rs
```diff
@@ -49,35 +50,39 @@ use object_store::path::Path;
...
+/// `StateSnapshotWriterV1` and consumed by `StateSnapshotReaderV1`. The
+/// snapshot is taken at the end of every epoch and stores the live object
+/// set bucketed by the same hash function used for the global state hash
+/// accumulator, allowing a single bucket to be consumed in parallel. Each
+/// bucket is split across one or more partitions; a partition is a single
+/// `.obj` file with a maximum size of 128 MB and a matching `.ref` file
+/// listing the object references in that partition. Partition files are
+/// optionally zstd-compressed.
+///
+/// V2 additions over V1:
```
semenov-vladyslav: Clarify that V2 refers to the snapshot format.

crates/iota-types/src/epoch_info.rs
```diff
@@ -0,0 +1,124 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    effects::TransactionEvents,
+    messages_checkpoint::{CertifiedCheckpointSummary, CheckpointSequenceNumber},
+};
+
+/// Per-epoch metadata sufficient to rebuild a per-epoch summary table
```
semenov-vladyslav: Which specifically "per-epoch summary table"?

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -177,6 +178,12 @@ pub struct CheckpointStoreTables {
     /// that epoch.
     epoch_last_checkpoint_map: DBMap<EpochId, CheckpointSequenceNumber>,
 
+    /// Per-epoch metadata sufficient to rebuild a per-epoch summary table
```
semenov-vladyslav: How exactly is it used to "rebuild a per-epoch summary table"?

crates/iota-core/src/authority/authority_store_types.rs
```diff
@@ -100,6 +113,50 @@ pub struct StoreObjectValue {
...
+#[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash)]
+pub struct StoreObjectValueV2 {
+    pub data: StoreData,
+    pub owner: Owner,
+    pub previous_transaction: TransactionDigest,
+    /// Checkpoint sequence number of the checkpoint that contained
+    /// `previous_transaction`. `SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT`
+    /// indicates the row was lifted from `StoreObjectV1` by `migrate()` and
+    /// has not yet been rewritten by the migration backfill.
+    pub previous_transaction_checkpoint: CheckpointSequenceNumber,
```
semenov-vladyslav: How is this field used?

crates/iota-core/src/authority/authority_store_tables.rs
bingyanglin: Fixed the code comment in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/authority/authority_store_tables.rs
bingyanglin: The comments were shorten. The only remaining note explains why we still need the `Serialize`/`Deserialize` derives. They are used by the snapshot writer/reader, which was not obvious before. Fixed in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/authority/authority_store.rs
bingyanglin: I agree this is clearer. Fixed in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60), thanks.

crates/iota-core/src/authority/authority_store.rs
bingyanglin: I agree this is clearer. Fixed in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60), thanks.

crates/iota-core/src/authority/authority_store_types.rs
bingyanglin: I checked all my touched code, the single hyphen `-` is by far the dominant style (2894 `.rs` comment occurrences vs 208 em-dashes). To make it consistent I revert it back to `-` in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/authority/authority_store_tables.rs
bingyanglin: Good suggestion. Fixed it in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/execution_cache/passthrough_cache.rs
bingyanglin: `PassthroughCache` writes to the database the moment a transaction runs, before the checkpoint number is known, so it writes a placeholder (`u64::MAX`, named `SENTINEL_PREVIOUS_TRANSACTION_CHECKPOINT`) into the field. The `build_db_batch` you commented on is an empty stub. A follow-up PR will run a one-time backfill that rewrites every placeholder row with the real number. I added an inline comment at the stub explaining this at [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
bingyanglin: I now use `0xCAFE_F00D_BEEF_0042` to make it specific in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60). Thanks for pointing this out.

crates/iota-core/src/authority/authority_store_types.rs
bingyanglin: Added more comments (Producer/Consumer paragraphs) to make it easier to understand at [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/checkpoints/mod.rs
bingyanglin: Enhanced the code docs in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-types/src/epoch_info.rs
bingyanglin: Enhanced the code docs in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-snapshot/src/lib.rs
bingyanglin: Enhanced the code docs in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60).

crates/iota-core/src/execution_cache/passthrough_cache.rs
bingyanglin: The placeholder will be fixed later by the future backfill PR.
I also added a new test `get_store_object_stamps_provided_checkpoint`. It saves an object two times. Once with the placeholder, once with a real number. Also I check the field has the right value each time. Two earlier tests also cover this: one reads the field back through the iterator, and `test_committed` writes a real number all the way to RocksDB and reads it back.
A short comment at this test helper was also added to clarify this.
Fixed in [3fe712b](https://github.com/iotaledger/iota/pull/11453/commits/3fe712bcbc3dbbc93df4488d9601e851701c1e60). Thanks.

crates/iota-core/src/execution_cache/passthrough_cache.rs
semenov-vladyslav: Ok, got it, will be implemented later.

semenov-vladyslav (approved):

crates/iota-core/src/authority/authority_store.rs
```diff
@@ -707,8 +710,10 @@ impl AuthorityStore {
     fn insert_object_direct(&self, object_ref: ObjectRef, object: &Object) -> IotaResult {
```
jkrvivian: The `insert_object_direct` can be combined with `insert_genesis_object`, otherwise, the naming is misleading, as it looks like a general function to insert objects.

bingyanglin: This is a good finding. Just this is not directly related to my PR and the upstream still has this. I guess we can just keep it to minimize structural divergence from upstream in iota-core to keep future ports cheap. How do you think?

crates/iota-snapshot/src/reader.rs
kodemartin, line 1: As far as I understand, on the reader side of `iota-snapshot` this patch still expects a serialized `Object` in the `*.obj` files. In `StateSnapshotReaderV2::sync_live_objects` we iterate over `LiveObject` entities to then write them into the storage.

Then downstream applications/libraries that are interested on the `previous_transaction_checkpoint` would have to reconstruct this info from the data encoded in the new object reference. This seems like an indirection and mixing of concerns.

So the question is, why not encode `LiveObjectV2` in the `*.obj` files directly, and leave reference encoding as it is (i.e. maintain `*.ref` files as they were)?

This way, references will be used only for integrity checks per partition as they do now, and then streaming live objects (i.e. `LiveObjectV2`) would give writers directly all the information they want.

bingyanglin: Agreed, thanks for the explanation, I now encode `LiveObjectV2` by using (object + inline previous_transaction_checkpoint), and make `.ref` revert to the V1 72-byte layout. Fixed in [2dba611](https://github.com/iotaledger/iota/pull/11453/commits/2dba61139e282de9f7b39f9d9e04559991eb90cb).

crates/iota-core/src/authority/authority_store.rs
jkrvivian: Alright, let's keep it then.

jkrvivian (approved):

crates/iota-snapshot/src/reader.rs
kodemartin: Thanks @bingyanglin! One more question: Did I understand correctly, that you plan to expose reading and writing epoch info as part of #10957?

kodemartin:
Regarding the newly added `EpochInfo`: The IOTA Indexer requires also the system-state at the start of each epoch, in order to populate the respective table. Any plan to incorporate this data as well?

crates/iota-snapshot/src/reader.rs
bingyanglin: Good question! Almost, after discussion with @muXxer.

This series (for #10957) has two PRs so make the things easier to review and discuss:

- This PR (the first one) only handles the writer side: it changes the snapshot format and writes the new EPOCH_INFO file into the snapshot. It does not read it back.
- The second PR will do two things:
  - Backfill the epoch info on nodes that don't have it yet (nodes started from genesis before grpc_indexes existed, or nodes started from an older snapshot).
  - Add the reader side: when a node loads a snapshot, we will call `grpc_indexes` directly to insert the epoch info. We need this because `grpc_indexes` does not read from the authority store or checkpoint store.

bingyanglin:
> Regarding the newly added `EpochInfo`: The IOTA Indexer requires also the system-state at the start of each epoch, in order to populate the respective table. Any plan to incorporate this data as well?

Yes this is included in #10957

kodemartin:
> > Regarding the newly added `EpochInfo`: The IOTA Indexer requires also the system-state at the start of each epoch, in order to populate the respective table. Any plan to incorporate this data as well?
> 
> Yes this is included in #10957

Could you elaborate on this? With the current patch you introduce `EpochInfoEntry` to encode epoch information in the snapshot, which does not include the system state. The [EpochInfo](https://iotaledger.github.io/iota/iota_types/storage/struct.EpochInfo.html) that I assume you refer to in #10907 includes the system state, but this is not encoded in the new snapshot. So how do you plan to include the system-state data in the snapshot?

bingyanglin:
> > > Regarding the newly added `EpochInfo`: The IOTA Indexer requires also the system-state at the start of each epoch, in order to populate the respective table. Any plan to incorporate this data as well?
> > 
> > 
> > Yes this is included in #10957
> 
> Could you elaborate on this? With the current patch you introduce `EpochInfoEntry` to encode epoch information in the snapshot, which does not include the system state. The [EpochInfo](https://iotaledger.github.io/iota/iota_types/storage/struct.EpochInfo.html) that I assume you refer to in #10907 includes the system state, but this is not encoded in the new snapshot. So how do you plan to include the system-state data in the snapshot?

Yes, this PR adds `start_system_state: Vec<u8>` to `EpochInfoEntry`. It carries the BCS-encoded `IotaSystemState` (object `0x5`) taken from the AdvanceEpoch tx of the **previous epoch's last checkpoint**, i.e. the last system state of the former epoch, which is also the start state of the new one. The indexer decodes with `bcs::from_bytes::<IotaSystemState>(&entry.start_system_state)`. Also I updated the PR description. Thanks.

crates/iota-snapshot/src/lib.rs
```diff
@@ -117,6 +135,17 @@ use tokio::time::Instant;
...
+/// EPOCH_INFO File Disk Format
+/// ┌──────────────────────────────┐
+/// │  magic(0x9000C001) <4 byte>  │
+/// ├──────────────────────────────┤
+/// │   bcs(EpochInfo)             │
+/// └──────────────────────────────┘
+/// See [`EpochInfo`] for the schema. The snapshot crate treats
+/// `EpochInfoEntry::start_system_state` as opaque bytes; only the indexer
+/// decodes it. Integrity is anchored by `FileMetadata::sha3_digest` in
+/// the MANIFEST (matching `.obj`/`.ref`); no in-file sha3 trailer.
```
kodemartin: I think this part includes information that seem rather unfit for this level of the API. For one,  the opaqueness of the `start_system_state` representation is already documented in the `EpochInfoEntry` type as revised with https://github.com/iotaledger/iota/pull/11453/commits/65b39498a9e5c7abfba2625c0996d5bfe4a9b811. Plus we don't document other fields of `EpochInfoEntry` here and rightly so.

Also, mentioning indexer as the sole consumer can be left out. The snapshots encodes information that can be used to restore epoch information. It should be agnostic of downstream applications that could use it like the Indexer.

Moreover, `FileMetadata::sha3_digest` could be used for the integrity of the file, but objects' integrity is checked by computing the checksum of all objects' digests from `*.obj` files and comparing against the checksum of the digests from `*.ref` files. So the integrity part needs some refinement. 

Finally, "in-file sha3 trailer" is not self-explanatory. I would recommend either expanding on what that means, or removing completely.


```rust
/// See [`EpochInfo`] for the schema. 
///
/// <revise integrity part>.
```

kodemartin (approved):
looks good for the `iota-snapshot` part. Left some final minor suggestions.

kodemartin:
> ### Follow-ups (PR 2, tracked in #10957)
> 
>     * Reader: download + verify + parse `EPOCH_INFO`, populate the indexer's `epoch_info` table.

What do you mean by "populate the indexer's `epoch_info` table? Note that there is already a patch generalizing the write operations for objects (#11559). Writing epoch info should follow a similar pattern I reckon, by adding a new `Restore` trait method and using that in the reader.

More specific Indexer implementations are planned in #11023

crates/iota-snapshot/src/lib.rs
bingyanglin: Agreed, I refined the comment in [312ee38](https://github.com/iotaledger/iota/pull/11453/commits/312ee38ef54a1d64298213056b695a504082fe38). Thanks for pointing this out!

bingyanglin:
> > ### Follow-ups (PR 2, tracked in #10957)
> > ```
> > * Reader: download + verify + parse `EPOCH_INFO`, populate the indexer's `epoch_info` table.
> > ```
> 
> What do you mean by "populate the indexer's `epoch_info` table? Note that there is already a patch generalizing the write operations for objects (#11559). Writing epoch info should follow a similar pattern I reckon, by adding a new `Restore` trait method and using that in the reader.
> 
> More specific Indexer implementations are planned in #11023

Thank you for the information, I integrated your information in the PR description to make it clearer.

semenov-vladyslav (approved):

piotrm50 (approved):

muXxer (approved):

PR #11454, opened 2026-05-07
feat: replace `ObjectData` with SDK version

thibault-martinez:

This PR replaces `ObjectData` (previously defined in `iota-types` as
`object::Data`) with the SDK version from `iota-sdk-types`. To keep the diff
small and avoid touching every call site, the SDK type is re-exported from
`iota-types` under the original name `Data`, so most existing imports continue
to work unchanged.

## Changes

- Remove the local `Data` enum in [crates/iota-types/src/object.rs](crates/iota-types/src/object.rs) and re-export `iota_sdk_types::ObjectData as Data`.
- The variant `Data::Move(MoveObject)` becomes `Data::Struct(MoveObject)` (matches `ObjectData::Struct` in the SDK). All in-repo usages have been updated accordingly.
- A few helper methods now come from the SDK type rather than the local impl:
  - `try_as_move()` → `as_struct_opt()`
  - `type_()` → `object_type()`
- Drop the now-redundant `TryFrom` conversions between `crate::object::Data` and `iota_sdk_types::ObjectData` in [crates/iota-types/src/iota_sdk_types_conversions.rs](crates/iota-types/src/iota_sdk_types_conversions.rs); `Object` / `GenesisObject` conversions become trivial field moves.
- Bump the `iota-sdk-types` git rev in [Cargo.toml](Cargo.toml) to pick up the SDK side of this change.
- Refresh the BCS format snapshot in [crates/iota-core/tests/staged/iota.yaml](crates/iota-core/tests/staged/iota.yaml): the type name in the registry changes from `Data` to `ObjectData`. The on-wire BCS encoding is unchanged (same variant order, same fields).

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::object::Data` is now a re-export of
  `iota_sdk_types::ObjectData`. The Move variant is renamed
  `Data::Move(MoveObject)` → `Data::Struct(MoveObject)`, and a few helpers
  move from the inherent impl to the SDK type: `try_as_move` →
  `as_struct_opt`, `type_` → `object_type`. Call sites that pattern-match on
  the variant or use these helpers need to be updated; the BCS wire format
  is unchanged.
- [ ] gRPC:


semenov-vladyslav (approved):

valeriyr (approved):

Thoralf-M (approved):

crates/iota-types/src/gas.rs
```diff
@@ -120,7 +120,7 @@ pub mod checked {
 
     pub fn deduct_gas(gas_object: &mut Object, charge_or_rebate: i64) {
         // The object must be a gas coin as we have checked in transaction handle phase.
-        let gas_coin = gas_object.data.try_as_move_mut().unwrap();
+        let gas_coin = gas_object.data.as_struct_mut_opt().unwrap();
```
Alex6323, line 123: wouldn't `as_struct_opt_mut()` be the "right" (as in commonly expected) counterpart to `as_struct_opt`?

kodemartin: Here's what the RUST API guidelines ([NAMING](https://rust-lang.github.io/api-guidelines/naming.html)) have to say about that
> If the `mut` qualifier in the name of a conversion method constitutes part of the return type, it should appear as it would appear in the type. For example [Vec::as_mut_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_slice) returns a `mut` slice; it does what it says. This name is preferred over `as_slice_mut`.

thibault-martinez: @kodemartin so that would be `as_mut_struct` and `as_mut_struct_opt` then?

Alex6323: For general getters `mut` becomes a suffix, but with `as_` an infix right after. Thanks @kodemartin

kodemartin: probably `as_opt_mut_struct` seems more consistent. Which implies that `as_opt_struct` should be also preferred than `as_struct_opt`.

thibault-martinez: Then do you mind if I open an issue for this and keep as is for now? It would mean renaming a lot of functions unrelated to this PR and it would bloat it. These functions are macro-generated so I can't rename only the ones related to ObjectData.

kodemartin (approved):

crates/iota-types/src/gas.rs
thibault-martinez: https://github.com/iotaledger/iota/issues/11459

Alex6323 (approved):

piotrm50 (approved):

PR #11395, opened 2026-04-30
feat: use ExecutionStatus from SDK (#10079)

thibault-martinez:
This PR is part of the ongoing effort to consolidate blockchain types in the
canonical `iota-rust-sdk`. Types that are needed by both clients/SDK consumers
and the node should live in `iota-rust-sdk` and be imported from there, instead
of being defined twice (once in `iota-types`, once in the SDK).

This PR replaces `ExecutionStatus` and the related `ExecutionFailureStatus`,
`MoveLocation`, `CommandArgumentError`, `PackageUpgradeError`, and
`TypeArgumentError` types (previously defined in `iota-types`) with the SDK
versions from `iota-sdk-types`. To keep the diff small and avoid touching
every call site, the SDK types are re-exported from `iota-types` under the
original names, so most existing imports continue to work unchanged. Note
that the SDK's `ExecutionError` is re-exported as `ExecutionFailureStatus`
to preserve the in-tree name.

The duplicated `From`/`TryFrom` conversions between the internal and SDK
copies in `iota_sdk_types_conversions.rs` are deleted since they are now the
same type.

Fixes https://github.com/iotaledger/iota/issues/9843

## Notable behavior-preserving changes at call sites

A number of fields and variants were renamed in the SDK definition, which
forced mechanical updates throughout the codebase:

- `ExecutionStatus::is_ok` / `is_err` → `is_success` / `is_failure`.
- `ExecutionFailureStatus::MoveAbort(MoveLocation, u64)` →
  `MoveAbort { location, code }` (tuple variant → struct variant).
- `MovePrimitiveRuntimeError(MoveLocationOpt)` →
  `MovePrimitiveRuntimeError { location: Option<MoveLocation> }`. The
  `MoveLocationOpt` newtype is gone — the SDK uses `Option<MoveLocation>`
  directly.
- `MoveLocation` now stores `package: ObjectId` + `module: Identifier`
  instead of a single `move_core_types::ModuleId`. Conversions are done at
  the call sites that need a `ModuleId` (e.g. clever-error resolution in
  [iota_transaction.rs](crates/iota-json-rpc-types/src/iota_transaction.rs)).
- Field renames: `arg_idx` → `argument`, `idx` → `index`,
  `result_idx` / `secondary_idx` → `result` / `subresult`, and
  `upgrade_error` → `kind` on `PackageUpgradeError`.
- Variant renames to follow Rust naming conventions:
[...]

PR #11461, opened 2026-05-07
fix(iota-types,ci): bump APPROX_SIZE_OF_EXECUTION_STATUS and restore workspace test coverage

DaughterOfMars:
## Summary

- Bump `APPROX_SIZE_OF_EXECUTION_STATUS` from 120 → 144. The SDK `ExecutionStatus` re-exported in #11395 lays out at 136 bytes (up from <120), tripping `check_approx_effects_components_size` in [`iota-types`](crates/iota-types/src/unit_tests/messages_tests.rs#L1380). 144 matches the slack pattern of the other `APPROX_*` constants. The wire (BCS) format is unchanged — only Rust-level struct layout grew, since variants were flattened (tuple → struct, nested `ModuleId` inlined, newtype wrappers collapsed); the `serde-reflection` snapshot at `crates/iota-core/tests/staged/iota.yaml` shows no change for `ExecutionStatus` itself.
- Fix a CI filter-set bug that allowed the regression above to land unnoticed and that has been silently disabling ~all `iota-types` (and many other) lib-test coverage on every workspace-config-touching PR for some time.

## Why CI didn't catch it

When a PR modifies `Cargo.toml` / `Cargo.lock` / `rust-toolchain.toml`, [`_rust.yml`](.github/workflows/_rust.yml#L98) sets `testOnlyChangedCrates=false` so the full workspace runs. In that mode [`build_filterset_changed_crates`](scripts/ci_tests/rust_tests.py#L379-L381) returns `""` (intent: "no filter, run everything"). That empty string was OR'd into a filter that already contained the hardcoded `FILTERSET_TESTS_POSTGRES_PG_INTEGRATION` and `FILTERSET_TESTS_MOVE_EXAMPLES_RDEPS` items. `append_filter_item_or` silently drops empty items, so the resulting `-E` expression narrowed the run to:

```
rdeps(iota-core) | rdeps(iota-test-transaction-builder) | <pg-integration> | <move-examples>
```

`iota-types` is a *forward* dependency of `iota-core`, not a reverse-dependency, so its lib tests were filtered out. (Confirmed against the `Run rust tests` log on PR #11395.)

The fix tracks the "run all" case explicitly and skips the narrower includes when it is set, so the combined filter only carries the exclude set (which nextest treats as "all tests minus the excluded ones").

## Latent failure surfaced by the CI fix

With CI coverage of `iota-types` (and other forward-deps of `iota-core`) restored, one pre-existing failure also resurfaces: `iota-types::auth_context::fields_v1::tests::command_make_move_vec_with_type_round_trip`. This is unrelated to the `ExecutionStatus` change — it's a `TypeName` serde adapter mismatch (`to_canonical_string(false)` writes without `0x`, deserializer requires `0x`). It is being fixed independently in #11457; once that lands, this test will be green under the restored CI coverage.

## Test plan

- [x] `cargo test -p iota-types --lib check_approx_effects_components_size` — passes locally.
- [x] Simulated `build_filterset_tests` across four scenarios (Cargo-config-changed PR, typical PR, no-changes, PG-only); all produce the expected filter expressions.
- [ ] CI green on this PR (will exercise the fixed filter on a workspace-config-touching PR), modulo the latent `TypeName` failure being fixed in #11457.


Thoralf-M (approved):

piotrm50 (approved):
approving APPROX_SIZE_OF_EXECUTION_STATUS increase

muXxer (approved):

PR #11463, opened 2026-05-07
feat(node): prune `AuthorityPerEpochStorePruner` during epoch transitions

muXxer:
# Description of change

This PR removes the time-based pruning parameter `epoch-db-pruning-period-secs` for the AuthorityPerEpochStore and triggers pruning during every epoch transition instead. The reasoning is twofold: First, during fast-syncing processes, epochs may sync faster than the configured pruning interval, causing pruning to fail or behave unexpectedly. Second, since pruning configuration is purely epoch-based, it is more logical to trigger it immediately after a new epoch is created.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes):  Removed the `epoch-db-pruning-period-secs` node config parameter. Pruning is now automatically triggered during every epoch transition. This fixes a potential memory leak and ensures consistent pruning behavior during node synchronization.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-config/data/fullnode-template-with-path.yaml
+++ b/crates/iota-config/data/fullnode-template-with-path.yaml
@@ -19,7 +19,6 @@ migration-tx-data-path: "migration.blob"
 
 authority-store-pruning-config:
   num-latest-epoch-dbs-to-retain: 3
-  epoch-db-pruning-period-secs: 3600
   num-epochs-to-retain: 1
   max-checkpoints-in-batch: 10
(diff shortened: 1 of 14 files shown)


jkrvivian (approved):

piotrm50 (approved):
iota-core changes lgtm

PR #11466, opened 2026-05-07
chore: use SDK's `write_sep`

### LINKED ISSUE #10960: Use write_sep from SDK?
We ported this function to the SDK but modified it, slight duplication at the moment.

thibault-martinez:
# Description of change

Removes the local `write_sep` helper from `iota-types::transaction` and switches all callers to the canonical implementation in `iota-sdk-types::utils::write_sep`. The local copy was a stopgap (per the `TODO` referencing #10960) duplicating logic that already lives in the SDK.

To enable this without forcing every caller to add `iota-sdk-types` as a direct dependency, `iota-types` now re-exports the SDK crate as `iota_types::sdk_types`, so callers can use `iota_types::sdk_types::utils::write_sep`.

The SDK signature takes an extra optional terminator argument; existing call sites pass `None` to preserve current formatting.

Fixes #10960

- [x] Patch-specific tests (correctness, functionality coverage) — pure refactor; output of affected `Display` impls in `iota-json-rpc-types` and `iota-replay` is unchanged
- [ ] I have added tests that prove my fix is effective or that my feature works — N/A (no behavior change)


crates/iota-grpc-server/tests/checkpoint_stream.rs
```diff
@@ -931,7 +931,8 @@ fn build_checkpoint_transactions_with_events(
                         Identifier::from_static("TestEvent"),
                         vec![],
                     ),
-                    contents: vec![0u8; 64], // 64 bytes of dummy content
+                    // TODO to be reviewed by node team
+                    contents: vec![0u8; 128], // 128 bytes of dummy content
```
muXxer: Why is this change needed?

muXxer (changes requested):

crates/iota-grpc-server/tests/checkpoint_stream.rs
thibault-martinez: Honestly I don't remember, I guess a test failed at some point but it doesn't anymore... I've removed it.

muXxer (approved):

Thoralf-M (approved):

valeriyr (approved):

crates/iota-indexer/src/read.rs
```diff
@@ -1088,7 +1088,7 @@ impl IndexerReader {
             .map(|(id, version)| {
                 format!(
                     "('\\x{}'::bytea, {}::bigint)",
-                    id.to_hex(),
+                    Hex::encode(id.into_bytes()),
```
DaughterOfMars, line 1091: ?

thibault-martinez: I unfortunately don't really remember how this change landed in this PR.
It appears that `::bytea` expects an unprefixed hex string and `to_hex` was prefixing.
I guess `id.to_raw_hex()` would also work.
Maybe someone from @iotaledger/infrastructure can double-check this.

semenov-vladyslav (approved):

crates/iota-indexer/src/read.rs
kodemartin: yes the bytes should be unprefixed. the `\\x` compensates for the binary representation here.

kodemartin: I agree that `to_raw_hex` seems more transparent

kodemartin: but `Hex::encode` is used all in other parts as well in this module so better to maintain consistency and leave it as is

kodemartin (approved):

PR #11468, opened 2026-05-07
feat: replace `TransactionKind` with SDK version

Alex6323:

This PR replaces `TransactionKind` (previously defined in `iota-types`) with
the SDK version from `iota-sdk-types`. To keep the diff small and avoid
touching every call site, the SDK type is re-exported from `iota-types` under
the original name `TransactionKind`, so most existing imports continue to
work unchanged.

## Notable changes

- The SDK rev (`iota-sdk-types`) is bumped to pull in the consolidated
  `TransactionKind` definition.
- The internal `iota_types::transaction::TransactionKind` enum is removed;
  `TransactionKind` is now re-exported directly from `iota_sdk_types` along
  with the other variant payload types (`ProgrammableTransaction`,
  `GenesisTransaction`, `ConsensusCommitPrologueV1`,
  `EndOfEpochTransactionKind`, `RandomnessStateUpdate`, …).
- Two SDK variant names differ from the previous internal names — call sites
  are updated everywhere they are matched:
  - `ProgrammableTransaction(_)` → `Programmable(_)`
  - `EndOfEpochTransaction(_)` → `EndOfEpoch(_)`
- Methods that used iota-types-specific types (e.g. `InputObjectKind`,
  `ProtocolConfig`, `Command`, `SharedObjectRef`) and therefore cannot live
  in the SDK are moved into a new extension trait `TransactionKindExt` in
  `iota_types::transaction`. This covers `get_advance_epoch_tx_gas_summary`,
  `contains_shared_object`, `shared_input_objects`, `move_calls`,
  `receiving_objects`, `input_objects`, `validity_check`, `iter_commands`,
  and `name`.
- A handful of inherent methods on the old enum were dropped because the SDK
  type already provides equivalents, and call sites are updated:
  - `is_system_tx()` → `is_system()`
  - `is_programmable_transaction()` → `is_programmable()`
  - `tx_count()` → `num_transactions()`
  - `is_end_of_epoch_tx()` is now provided by matching `EndOfEpoch(_)`
    directly (still exposed via `TransactionDataAPI::is_end_of_epoch_tx`).
  - The `programmable()` shortcut constructor is removed; callers use
    `TransactionKind::Programmable(pt)` directly.
- The custom `Display` impl and the `IntoStaticStr` derive on the old enum,
  as well as `ProgrammableTransactionExt::fmt_display`, are removed (the SDK
  type does not provide them and no caller relied on them after the rename).
- The conversion impls between `crate::transaction::TransactionKind` and
  `iota_sdk_types::TransactionKind` in
  `iota-types/src/iota_sdk_types_conversions.rs` are deleted, and the
  surrounding `TransactionV1` ↔ `TransactionDataV1` conversions collapse to
  direct moves (`value.into_kind()` / `value.kind`) since the kinds are now
  the same type.
- The staged BCS schema (`crates/iota-core/tests/staged/iota.yaml`) is
  regenerated: the variant-4 name on `TransactionKind` changes from
  `EndOfEpochTransaction` to `EndOfEpoch` to match the SDK enum. The
  variant index and payload shape are unchanged, so the BCS wire format is
  unchanged.
- Wide call-site updates flow from the variant renames across `iota-core`,
  `iota-graphql-rpc`, `iota-grpc-server`, `iota-indexer`,
  `iota-json-rpc-types`, `iota-replay`, `iota-transaction-builder`,
  `iota-transaction-checks`, `iota-move-build` test adapter, the execution
  engine, fuzzers, the `iota` CLI, simulacrum, and the tic-tac-toe example.
- Because the SDK enum is `#[non_exhaustive]` from the matcher's
  perspective, several `match` sites now need an additional fallback arm
  (`unimplemented!("a new TransactionKind enum variant was added …")`) to
  stay forward-compatible with future SDK additions.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `TransactionKind` (and its variant payload types) is now
  re-exported from `iota_types::transaction` and originates in
  `iota-sdk-types`. Downstream consumers can depend on the SDK directly
  instead of `iota-types`. Methods that require iota-types-specific types
  are exposed via a new `TransactionKindExt` extension trait. Several
  inherent methods are renamed to match the SDK
  (`is_system_tx` → `is_system`, `is_programmable_transaction` →
  `is_programmable`, `tx_count` → `num_transactions`).
- [ ] gRPC:


valeriyr (approved):

jkrvivian (approved):

crates/iota-replay/tests/sandbox_snapshots/B9jn6GZsPebSEX6fp9rV7U6k6sZrgokgkGpFMEE6x7Us.json
```diff
@@ -84,38 +83,37 @@
             }
         ],
         "kind": {
-            "ProgrammableTransaction": {
-                "inputs": [
-                    {
-                        "pure": {
-                            "value": "NSnsCrJaqEOUJxUZTdTu4fwdXIPf7yjoOYOPoPCzaqA="
-                        }
+            "kind": "programmable_transaction",
```
alexsporn: Is this change expected? Shouldn't it be then `programmable`?

crates/iota-types/src/transaction.rs
```diff
@@ -932,82 +893,33 @@ impl TransactionKind {
...
-        }
-    }
-
-    pub fn name(&self) -> &'static str {
+    fn name(&self) -> &'static str {
         match self {
             Self::Genesis(_) => "Genesis",
             Self::ConsensusCommitPrologueV1(_) => "ConsensusCommitPrologueV1",
-            Self::ProgrammableTransaction(_) => "ProgrammableTransaction",
+            Self::Programmable(_) => "ProgrammableTransaction",
```
alexsporn: We should decide what to do here, because I don't like the inconsistency this is creating

crates/iota-types/src/transaction.rs
```diff
@@ -932,82 +893,33 @@ impl TransactionKind {
...
-                    "Transaction Kind : Authenticator State Update (deprecated)"
-                )?;
-            }
-            Self::RandomnessStateUpdate(_) => {
-                writeln!(writer, "Transaction Kind : Randomness State Update")?;
-            }
-            Self::EndOfEpochTransaction(_) => {
-                writeln!(writer, "Transaction Kind : End of Epoch Transaction")?;
-            }
+            Self::EndOfEpoch(_) => "EndOfEpochTransaction",
```
alexsporn: Same as above

alexsporn (approved):
LGTM but we should discuss the overall change from `ProgrammableTransaction` to `Programmable` and the same with `SystemTransaction` and `EndOfEpochTransaction`, since these changes introduce a discrepancy with the rest of the codebase and maybe some public APIs

crates/iota-types/src/transaction.rs
Alex6323: it's been discussed atm internally. Will come back to you.

thibault-martinez: This `name` function is only used here https://github.com/iotaledger/iota/blob/sdk-replacement/27-TransactionKind/crates/iota-analytics-indexer/src/handlers/transaction_handler.rs#L152 in the iota-analytics-indexer crate, do you think it's fine to change it?

crates/iota-replay/tests/sandbox_snapshots/B9jn6GZsPebSEX6fp9rV7U6k6sZrgokgkGpFMEE6x7Us.json
thibault-martinez: This test uses the JSON serialisation of the type (nothing related to JSON-RPC), we believe it's fine to change it and update this test. Probably this should also remove the suffix as well.

kodemartin (approved):

crates/iota-types/src/transaction.rs
Alex6323: https://github.com/iotaledger/iota/pull/11468/commits/d144b688e07840525af361a753a7e20557ac0f9a

crates/iota-types/src/transaction.rs
Alex6323: https://github.com/iotaledger/iota/pull/11468/commits/d144b688e07840525af361a753a7e20557ac0f9a

crates/iota-replay/tests/sandbox_snapshots/B9jn6GZsPebSEX6fp9rV7U6k6sZrgokgkGpFMEE6x7Us.json
Alex6323: https://github.com/iotaledger/iota/pull/11468/commits/d144b688e07840525af361a753a7e20557ac0f9a

crates/iota-json-rpc-types/src/iota_transaction.rs
```diff
@@ -634,17 +637,19 @@ impl IotaTransactionBlockKind {
...
                     package_resolver,
                 )
                 .await?,
+            )),
+            _ => unimplemented!(
+                "a new TransactionKind enum variant was added and needs to be handled"
             ),
-            tx => Self::try_from_inner(tx, tx_digest)?,
-        })
+        }
```
thibault-martinez: What's happening here? Is this a failed rebase?

filipdulic (approved):

crates/iota-json-rpc-types/src/iota_transaction.rs
Alex6323: hmm, that change doesn't seem right

Alex6323: good catch, dunno what happened there tbh

thibault-martinez (approved):

PR #11470, opened 2026-05-08
feat(iota-types)!: replace TransactionData with SDK Transaction

Thoralf-M:

This PR replaces `TransactionData` (previously defined in
`iota-types::transaction`) with the SDK version from `iota-sdk-types`. To keep
the diff small and avoid touching every call site, the SDK type is re-exported
from `iota-types::transaction` under the original name `TransactionData`, so
most existing imports continue to work unchanged.

## Description of change

- **Type replacement:** `TransactionData` is now `pub use iota_sdk_types::Transaction as TransactionData` and `TransactionDataV1` is `TransactionV1`. Removed the local enum definition and `enum_dispatch`.
- **Manual trait impl:** Since `enum_dispatch` can't be used on external types, `TransactionDataAPI` is now implemented directly on the enum with explicit `match` arms.
- **Digest/hash fix:** The SDK type is named `Transaction` not `TransactionData`, which broke `default_hash()` (uses `serde_name::trace_name`). A manual digest implementation was added to preserve the same hash.
- **Unused methods removed:** `contains_shared_object`, `uses_randomness`, `digest` were flagged as unused and cleaned up.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::transaction::TransactionData` is now a
  re-export of `iota_sdk_types::Transaction` (and `TransactionDataV1` of
  `TransactionV1`). The `TransactionDataV1::gas_data` field is renamed to
  `gas_payment`; the `gas_data()` / `gas_data_mut()` trait accessors keep
  their names. All inherent constructors (`new`, `new_move_call`,
  `new_programmable`, …) and the `message_version` / `execution_parts`
  helpers have moved onto `TransactionDataAPI`; calls that used to go
  through the inherent `impl` need the trait in scope. The unused
  `contains_shared_object`, `uses_randomness`, and inherent `digest`
  methods have been removed; transaction digest computation now goes
  through the SDK's `Transaction::digest()`, which uses the salt
  `"TransactionData::"` to preserve hash stability across the rename.
- [ ] gRPC:


valeriyr (approved):

semenov-vladyslav (approved):

crates/iota-core/src/authority.rs
```diff
@@ -2389,7 +2386,7 @@ impl AuthorityState {
             },
             transaction,
         );
-        let transaction_digest = TransactionDigest::new(default_hash(&intent_msg.value));
+        let transaction_digest = TransactionDigest::new(intent_msg.value.digest().into_inner());
```
thibault-martinez, line 2389: Just double checking, strictly similar ofc ?

kodemartin (approved):

crates/iota-core/src/authority.rs
Thoralf-M: Yes, same as the first time checking :trollface: 
https://github.com/iotaledger/iota/pull/11067#discussion_r3050015674

thibault-martinez: consistency :trollface:

filipdulic (approved):

thibault-martinez (approved):

Alex6323 (approved):

PR #11471, opened 2026-05-08
feat(iota-types)!: update to MoveStruct with contents checked

Thoralf-M:
This PR follows up on the `MoveObjectType` / `MoveStruct` migration in
`iota-rust-sdk`. The SDK's `MoveStruct` (re-exported in `iota-types` as
`MoveObject`) has been hardened: its fields are now private, construction
goes through a fallible `MoveStruct::new` constructor, and replacing the
contents goes through a fallible `set_contents` setter — both of which
validate that `contents` is at least `ObjectId::LENGTH` (32) bytes, since
every Move object must carry its `ObjectId` as the leading bytes of its
BCS payload. This PR adapts the `iota-types` and execution call sites to
the new API.

## Description of change

- `iota-types::object` (`MoveObjectExt`) no longer reaches into
  `MoveObject` fields directly. All `self.object_type` / `self.version`
  / `self.contents` accesses are replaced with the corresponding
  `object_type()` / `version()` / `contents()` getters; mutations go
  through `set_version()` and the fallible `set_contents()` (returning
  `MoveStructContentsError`, mapped to `ExecutionError::invariant_violation`
  on the production path).
- `MoveObjectExt::set_coin_value_unchecked` and
  `set_clock_timestamp_ms_unchecked` no longer mutate the contents buffer
  in place via `Vec::splice` — they now copy the buffer
  (`self.contents().to_vec()`), patch the trailing 8 bytes, and pass it
  back through `set_contents(...).unwrap()` (the length is unchanged, so
  the length check cannot fail).
- All `MoveObject { object_type, version, contents }` struct-literal
  constructions in `Object::*_for_testing` helpers and in
  `MoveObjectExt::new_from_execution_with_limit` are replaced with
  `MoveObject::new(...)`. Test helpers `.unwrap()` the result; the
  execution path maps the error to `ExecutionError::invariant_violation`.
- The associated helper `MoveObject::id_opt(&bytes)` (which extracted an
  `ObjectId` from the leading bytes of a raw payload) was removed in the
  SDK now that the same invariant is upheld by the constructor. The two
  call sites in `iota-adapter::programmable_transactions::context` are
  updated to call `ObjectID::from_bytes(&bytes[..ObjectID::LENGTH])`
  directly, with the missing-bytes case turning into an
  `ExecutionError::invariant_violation`.
- `iota-open-rpc::examples` had a `MoveObject::new_from_execution_with_limit`
  call that built an example object with an empty `contents: Vec::new()`
  and an empty `MoveStructLayout`. Empty contents now fail the new
  length check, so the example is updated to construct a 32-byte payload
  via `UID::new(parent_object_id).to_bcs_bytes()` and a matching layout
  with a single `id: UID` field. Generated `openrpc.json` snapshots are
  regenerated to reflect the new example payload.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `MoveStruct` (re-exported as `iota-types::object::MoveObject`)
  now has private fields. Construction must go through
  `MoveStruct::new(object_type, version, contents)` and content updates
  through `set_contents(...)`; both are fallible and reject `contents`
  shorter than `ObjectId::LENGTH` (32 bytes), since every Move object
  must carry its `ObjectId` as the leading bytes. The associated helper
  `MoveObject::id_opt(&bytes)` has been removed — call
  `ObjectID::from_bytes(&bytes[..ObjectID::LENGTH])` directly. Direct
  field access (`.object_type`, `.version`, `.contents`) is replaced by
  the equivalent getter methods. Wire formats (BCS and JSON) are
  unchanged.
- [ ] gRPC:


iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
```diff
@@ -1626,7 +1628,8 @@ mod checked {
     ) -> Result<MoveObject, ExecutionError> {
         debug_assert_eq!(
             id,
-            MoveObject::id_opt(&contents).expect("object contents should start with an id")
+            ObjectID::from_bytes(&contents[..ObjectID::LENGTH])
+                .expect("object contents should start with an id")
```
valeriyr, line 1632: Can we create a utility function for this? MoveObject::id or something similar?

Thoralf-M: `MoveObject::id()` exists, but we would need the full `MoveObject` for that, not just the contents. Could be constructed from
```Rust
        id: ObjectID,
        type_: Type,
        contents: Vec<u8>,
```
but then one might want to refactor the whole fn to just accept `MoveObject`
Do you want that?

valeriyr: I am not sure, I just would like to hide this part somewhere: `ObjectID::from_bytes(&contents[..ObjectID::LENGTH])`, probably to ObjectID?

muXxer (approved):

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Thoralf-M: So something like `ObjectID::from_vec()` and it just uses the first 32 bytes independent of the length and only panic if it's shorter?

valeriyr: Ok, as I see, we have some places where this logic is repeated. We can leave it as is. You can decide whether it makes sense to change anything.

valeriyr (approved):

semenov-vladyslav (approved):

kodemartin (approved):

iota-execution/latest/iota-adapter/src/programmable_transactions/context.rs
Thoralf-M: I don't see a big improvement of another fn and to keep changes minimal I'm not changing something now

filipdulic (approved):

crates/iota-types/src/object.rs
```diff
@@ -191,10 +189,11 @@ impl MoveObjectExt for MoveObject {
     fn set_clock_timestamp_ms_unchecked(&mut self, timestamp_ms: u64) {
         assert!(self.struct_tag().is_clock());
         // 32 bytes for object ID, 8 for timestamp
-        assert!(self.contents.len() == 40);
+        assert!(self.contents().len() == 40);
```
thibault-martinez, line 192: Nothing major but a bit inconsistent that this function uses `assert` and the other one `debug_assert`, maybe an issue to open for later

Thoralf-M: https://github.com/iotaledger/iota/issues/11482

thibault-martinez (approved):

Alex6323 (approved):

PR #11476, opened 2026-05-08
feat(iota-data-ingestion-core): add filter newtype for better ergonomics

### LINKED ISSUE #11435: [iota-data-ingestion-core]: better ergonomics for the `TransactionFilter` type

sergiupopescu199:
# Description of change

Introduces newtype wrappers around `proto::TransactionFilter` and `proto::EventFilter` that expose a chained builder API, plus a flat factory wrapper for `proto::CommandFilter`. Users build filters without interacting directly with the proto types.

### Why a wrapper

The proto API requires constructing several intermediate types (e.g. `TransactionKindsFilter`, `AddressFilter`, `ExecutionStatusFilter`) and wrapping them via `with_*` builders.

Even though the proto exposes builder-like methods, its `oneof` (enum) semantics make a chained builder unsafe on the proto types directly: calling `with_kinds(...).with_sender(...)` on the same instance silently overwrites the first variant. The wrapper avoids this by making each leaf method return a fresh combined filter, implicitly `AND`ing the new condition with everything accumulated so far. Explicit `OR` is available via `.or(...)`, and `NOT` via `.negate()` or the `!` operator.

`CommandFilter` is a flat factory rather than a builder because its proto oneof has only leaf variants (no `All` / `Any` / `Negation`). Logical composition over commands happens at the `TransactionFilter` level.

### Why not re-export the proto types

Any change on the proto side (new variant, renamed field, module reshuffle) propagates as a breaking change downstream. The newtype gives us a stable surface that can evolve independently.

### Example of using proto types directly vs newtype wrapper

```rust
use iota_grpc_types::v1::filter as proto;

// kinds([Programmable])
let mut kinds_inner = proto::TransactionKindsFilter::default();
kinds_inner.push_kinds(proto::TransactionKind::ProgrammableTransaction);
let kinds = proto::TransactionFilter::default().with_transaction_kinds(kinds_inner);

// execution_status(true)
let status = proto::TransactionFilter::default().with_execution_status(
    proto::ExecutionStatusFilter::default().with_success(true),
);

// sender(alice)
let sender = proto::TransactionFilter::default()
    .with_sender(proto::AddressFilter::default().with_address(alice));

// AND of all three
let all = proto::TransactionFilter::default().with_all(
    proto::AllTransactionFilter::default().with_filters(vec![kinds, status, sender]),
);

// NOT receiver(bob)
let receiver = proto::TransactionFilter::default()
    .with_receiver(proto::AddressFilter::default().with_address(bob));
let not_receiver = proto::TransactionFilter::default().with_negation(
    proto::NotTransactionFilter::default().with_filter(receiver),
);

// OR of [all, not_receiver]
let filter = proto::TransactionFilter::default().with_any(
    proto::AnyTransactionFilter::default().with_filters(vec![all, not_receiver]),
);

```
With the wrapper:

```rust
use iota_data_ingestion_core::filters::fullnode::{TransactionFilter, TransactionKind};

let filter = TransactionFilter::new()
    .kinds([TransactionKind::ProgrammableTransaction])
    .execution_status(true)
    .sender(alice)
    .or(!TransactionFilter::new().receiver(bob));
```

fixes #11435 

> [!NOTE]

### DIFF
--- a/crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
+++ b/crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
@@ -1,19 +1,868 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-//! Server-side filtering logic for Fullnode gRPC connections.
+//! Server-side filters for fullnode gRPC connections.
 //!
-//! These filters operate at the content level (transactions and events)
-//! within each checkpoint.
-pub use iota_grpc_types::v1::{
-    filter::{
-        AddressFilter, AllEventFilter, AllTransactionFilter, AnyEventFilter, AnyTransactionFilter,
-        CommandFilter, EventFilter, ExecutionStatusFilter, MakeMoveVecCommandFilter,
-        MergeCoinsCommandFilter, MoveCallCommandFilter, MoveEventTypeFilter,
-        MovePackageAndModuleFilter, NotEventFilter, NotTransactionFilter, ObjectIdFilter,
-        PublishCommandFilter, SplitCoinsCommandFilter, TransactionFilter, TransactionKind,
-        TransactionKindsFilter, TransferObjectsCommandFilter, UpgradeCommandFilter, command_filter,
-        event_filter, transaction_filter,
-    },
-    types::{Address, ObjectId, ObjectReference},
-};
+//! Filters tell the fullnode which transactions to include in each
+//! checkpoint payload. Filtering happens entirely on the server-side, the
+//! ingestion framework performs no client-side filtering.
+//!
+//! # Filter types
+//!
+//! - [`TransactionFilter`]: applied to a checkpoint's transactions.
+//! - [`EventFilter`]: matches a transaction by its emitted events. Used as
+//!   input to [`TransactionFilter::event`].
+//! - [`CommandFilter`]: matches a command within a programmable transaction.
+//!   Used as input to [`TransactionFilter::command`].
+//!
+//! # Composition
+//!
+//! [`TransactionFilter`] and [`EventFilter`] use a chained builder pattern:
+//! each leaf method (e.g. `TransactionFilter::new().kinds().sender()`)
+//! implicitly `AND`s the new condition with everything accumulated so far.
+//! Explicit `OR` composition is available via the [`or`](TransactionFilter::or)
+//! method on each type. `NOT` is available via
+//! [`negate`](TransactionFilter::negate), the `!` operator, or `.not()` when
+//! [`std::ops::Not`] is in scope.
+//!
+//! [`CommandFilter`] is composed only at the [`TransactionFilter`] level by
+//! passing each command to [`TransactionFilter::command`] within a chain.
+//! Multiple [`command`](TransactionFilter::command) calls implicitly `AND`,
+//! while `OR` and `NOT` follow the rules described above.
+//!
+//! # Examples
+//!
+//! Successful programmable transactions:
+//!
+//! ```rust
+//! use iota_data_ingestion_core::filters::fullnode::{TransactionFilter, TransactionKind};
+//!
+//! let filter = TransactionFilter::new()
+//!     .kinds([TransactionKind::Programmable])
+//!     .execution_status(true);
+//! ```
+//!
+//! Transactions emitting a specific event type:
+//!
+//! ```rust
+//! use iota_data_ingestion_core::filters::fullnode::{EventFilter, TransactionFilter};
+//!
+//! let filter =
+//!     TransactionFilter::new().event(EventFilter::new().event_type("0xabcd::my_module::Foo"));
+//! ```
+//!
+//! Transactions NOT sent by a specific address:
+//!
+//! ```ignore
+//! let filter = !TransactionFilter::new().sender(some_address);
+//! ```
+
+use std::ops::Not;
+
+use iota_grpc_types::v1::{filter as proto, types::ObjectReference};
+use iota_types::base_types::{IotaAddress, ObjectDigest, ObjectID, SequenceNumber};
+
+/// Available transaction kinds for filtering.
+pub type TransactionKind = proto::TransactionKind;
+
+/// Filter applied to transactions in a fullnode checkpoint stream.
+///
+/// Built with a chained builder pattern. Start from [`TransactionFilter::new`]
+/// and add leaves with the named methods (e.g. [`TransactionFilter::kinds`],
+/// [`TransactionFilter::sender`], [`TransactionFilter::execution_status`]).
+///
+/// Each leaf implicitly `AND`s with everything accumulated so far. Use
+/// [`TransactionFilter::or`] for `OR` composition and
+/// [`TransactionFilter::negate`] (or the `!` operator) for `NOT`.
+///
+/// # Example
+///
+/// Implicit AND composition
+///
+/// ```rust
+/// use iota_data_ingestion_core::filters::fullnode::{TransactionFilter, TransactionKind};
+///
+/// let filter = TransactionFilter::new()
+///     .kinds([TransactionKind::Programmable])
+///     .execution_status(true);
+/// ```
+/// Composition with AND, OR and NOT:
+///
+/// ```ignore
+/// // (Programmable AND success AND sender == Alice) OR NOT(receiver == Bob)
+/// let filter = TransactionFilter::new()
+///     .kinds([TransactionKind::Programmable])
+///     .execution_status(true)
+///     .sender(alice)
+///     .or(!TransactionFilter::new().receiver(bob));
+/// ```
+#[derive(Clone, Debug, Default)]
+pub struct TransactionFilter(proto::TransactionFilter);
+
+impl TransactionFilter {
+    /// Creates an empty filter.
+    ///
+    /// An empty filter is rejected by the fullnode. Add at least one leaf via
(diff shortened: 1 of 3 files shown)


crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,453 @@
...
-        TransactionKindsFilter, TransferObjectsCommandFilter, UpgradeCommandFilter, command_filter,
-        event_filter, transaction_filter,
-    },
-    types::{Address, ObjectId, ObjectReference},
-};
+//! Filters tell the fullnode which transactions to include in each
+//! checkpoint payload.
+
+use iota_grpc_types::v1::filter as proto;
+use iota_sdk_types::{Address, ObjectId, ObjectReference};
```
kodemartin: Better to use the reexported aliases from `iota-types` for the sake of consistency.

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,453 @@
...
+            }
+            TransactionKind::Genesis => proto::TransactionKind::Genesis,
+            TransactionKind::ConsensusCommitPrologueV1 => {
+                proto::TransactionKind::ConsensusCommitPrologueV1
+            }
+            TransactionKind::EndOfEpochTransaction => proto::TransactionKind::EndOfEpochTransaction,
+            TransactionKind::RandomnessStateUpdate => proto::TransactionKind::RandomnessStateUpdate,
+        }
+    }
+}
```
kodemartin: Why don't we use the `iota-types` type instead?

sergiupopescu199: Looking at the [TransactionKind](https://iotaledger.github.io/iota/iota_types/transaction/enum.TransactionKind.html) enum it does contain values, so constructing it from user side it's non optimal.

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11476/commits/e1018bba4216b842b5a8c2f0ffec181c1bb9d573

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
kodemartin: Why not aliasing `proto::TransactionKind` then?

kodemartin (approved):
looks meraviglioso! 

the PR descriptions needs some update though.

Plus a minor follow up comment

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
sergiupopescu199: I was thinking that exposing directly this enum could result in breaking changes in the future, mostly if one already existing variant gets renamed. After careful consideration I think this situation isn't likely to happen.

fixed: https://github.com/iotaledger/iota/pull/11476/commits/3556e68185830bc5f70d9d1d2c628dde469e6abb

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,818 @@
...
+    /// - `false` for failed transactions.
+    pub fn execution_status(self, success: bool) -> Self {
+        self.and_with(
+            proto::TransactionFilter::default().with_execution_status(
+                proto::ExecutionStatusFilter::default().with_success(success),
+            ),
+        )
+    }
+
+    /// Matches transactions sent by the given object id.
```
tomxey: typo: by given address

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,818 @@
...
+///     .kinds([TransactionKind::ProgrammableTransaction])
+///     .execution_status(true)
+///     .sender(alice)
+///     .or(!TransactionFilter::new().receiver(bob));
+/// ```
+#[derive(Clone, Debug, Default)]
+pub struct TransactionFilter(proto::TransactionFilter);
+
+impl TransactionFilter {
+    /// Creates an empty filter.
```
tomxey, line 103: could we document what is the meaning of the empty filter?
here and also for event filter

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,818 @@
...
+            Some(_) => {
+                Self(proto::TransactionFilter::default().with_any(
+                    proto::AnyTransactionFilter::default().with_filters(vec![self.0, leaf]),
+                ))
+            }
+        }
+    }
+
+    /// Matches transactions of any of the given [`TransactionKind`]s.
+    pub fn kinds(self, kinds: impl IntoIterator<Item = TransactionKind>) -> Self {
```
tomxey, line 160: could we document what is the meaning of the filter if we pass empty list of kinds?

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
```diff
@@ -1,19 +1,818 @@
...
+    /// Matches transactions whose recipient is the given address.
+    pub fn receiver(self, address: IotaAddress) -> Self {
+        self.and_with(
+            proto::TransactionFilter::default()
+                .with_receiver(proto::AddressFilter::default().with_address(address)),
+        )
+    }
+
+    /// Matches transactions that touch the given object.
+    pub fn affected_object(self, object_ref: ObjectRef) -> Self {
```
tomxey: shouldn't it be enough to pass the object id here? why do we need the version and digest, are they used?

sergiupopescu199: Even tough the server side does only use the object id, the proto definition requires an object ref as filter input.

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11476/commits/78c6aaa245d813f27648772d11688fc80211f4bf

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11476/commits/78c6aaa245d813f27648772d11688fc80211f4bf

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11476/commits/78c6aaa245d813f27648772d11688fc80211f4bf

crates/iota-data-ingestion-core/src/reader/filters/fullnode.rs
tomxey: could we make it easier for the user and accept only object id as an argument to `affected_object`?
internally we can construct the objectref object with dummy version and digest, but I think it may be strange for the user to have to come up with dummy values themself

what do you think?

sergiupopescu199: That's good that you raised this question, had a conversation with the node team and indeed the filter for now only checks for object id but it should support 3 options where user can choose:

1. check for object id
2. check for object id and version
3. check for digest

I've updated the builder api to support this options, the node team till update the grpc server in order to support them.

fixed: https://github.com/iotaledger/iota/pull/11476/commits/c1983222a1394fa54f57ce1c46ddeb6c7c26efce

tomxey (approved):

PR #11477, opened 2026-05-08
feat(graphql): migrate dynamic fields to backward diff using `objects_version`

### LINKED ISSUE #11019: [graphql]: Refactor consistent views on dynamic fields

tomxey:
# Description of change

Switch DynamicField::paginate from forward diff (objects_snapshot + objects_history) to backward diff (checkpointed_objects + objects_backward_history). The version-pinned path uses the new backward_view::version_pinned query; the no-version-pin path falls through to the existing consistent-view query.

Read-side:

- New backward_view::version_pinned query mirrors consistent::query in shape (Source A from checkpointed_objects + Source B from objects_backward_history merged via DISTINCT ON), but pivots on a version-axis cutoff rather than the cp-axis. For each candidate the target version is the largest objects_version.object_version that is <= parent_version. The state row at (object_id, target_version) is read from checkpointed_objects (current state) or objects_backward_history (prior state). Tombstone target versions exist in objects_version but have no corresponding state row in either table, so the candidate is correctly excluded.
- Source A excludes rows whose (object_id, object_version) also appears in objects_backward_history, suppressing duplicates during the brief race window where backward_history has been written but checkpointed_objects has not yet been updated.
- Behaviour matches the original forward-diff "earliest / produced-at" semantics: any DF with lamport <= parent_version is in scope, and the latest pre-cutoff state of each candidate is returned.

Tests:

- dof_add_reclaim_transfer_reclaim_add.move (delete-recreate of a Field-object with the same derived id) passes without any new schema column: the version-axis approach disambiguates the intra-checkpoint deletion gap via objects_version's tombstone versions.
- dynamic_fields.move: two queries renamed away from the snapshot-lag "outside consistent range" naming (those semantics don't apply under backward-diff retention) and re-recorded.

fixes https://github.com/iotaledger/iota/issues/11019


crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
kodemartin, line 1: `version_pinned` is referring to the version of the parent, but without the context of dynamic fields, this hints towards a more generic scope, and slightly ambiguous moreover.

I think therefore that naming the module `dynamic_fields` would be more transparent. This is after all logic to get dynamic fields specifically.

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
```diff
@@ -0,0 +1,126 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Version-pinned consistent view of the dynamic fields of a parent at the
+//! moment the parent reached the requested `parent_version`.
+//!
+//! For each candidate DF, the target version is the largest version in
+//! `objects_version` whose value is `<= parent_version`. The state row at
+//! `(object_id, target_version)` is then read from `checkpointed_objects` (if
+//! it's the current state) or `objects_backward_history` (if it's a prior
+//! state). When the target version is a tombstone, NYC marker, or synth
```
kodemartin: What is `NYC`? New York City 🏙️ ? 

If you must use abbreviations, the general rule of practice is that we provide full expansion on first mention. E.g. `NotYetCreated` (NYC). But it's ok to use the full status.

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
```diff
@@ -0,0 +1,126 @@
...
+//! WrappedOrDeleted, the row's `owner_id`/`df_kind`/etc. are NULL — the
+//! `owner_id`/`df_kind` filter applied internally drops it.
+//!
+//! Mirrors the **earliest / produced-at** semantics of the original
+//! forward-diff `df.object_version <= parent_version` filter.
```
kodemartin: Are the `earliest / produced-at` semantics documented somewhere in the RUST API? If not this segment causes more confusion than it clarifies the underlying design concepts.

kodemartin (approved):
Looks good with some minor suggestions.

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -0,0 +1,126 @@
...
+pub(crate) fn query(parent: IotaAddress, parent_version: u64, page: &Page<Cursor>) -> RawQuery {
+    let parent_version = parent_version as i64;
+    let parent_filter = parent_dynamic_field_filter(parent);
+    merge_and_deduplicate(vec![
+        version_pinned_checkpointed_objects(parent_version, page, &parent_filter),
+        version_pinned_historical_objects(parent_version, page, &parent_filter),
+    ])
+}
+
+/// Returns a filter that constrains a row to be a dynamic field of `parent`.
```
sergiupopescu199, line 48:
```suggestion
/// Returns a filter that constrains a row to be a dynamic field of `parent`.
///
```

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
```diff
@@ -0,0 +1,126 @@
...
+        iota_address::IotaAddress,
+        object::{Cursor, StoredBackwardObject},
+    },
+};
+
+/// Builds a version-pinned consistent view of the dynamic fields owned by
+/// `parent` at `parent_version`. The DF-shape filter
+/// (`owner_id = parent AND owner_type = Object AND df_kind IS NOT NULL`) is
+/// applied internally — it both narrows the candidate set and excludes
+/// non-Active rows (whose `owner_id`/`df_kind` are NULL).
```
sergiupopescu199: We could add a newline between the first sentence in order to comply better with the rust convenstions for the Rust doc

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
```diff
@@ -0,0 +1,126 @@
...
+/// "the largest known version of this object is in the requested window" —
+/// no `objects_version` lookup is needed here.
+///
+/// Excludes rows also present in `objects_backward_history` at the same
+/// `(object_id, object_version)` to keep the two sources disjoint at the
+/// row level — the same `(id, version)` can briefly appear in both tables
+/// during the race window between backward-history and checkpointed-objects
+/// writes. The merge step's `DISTINCT ON` only deduplicates within a page,
+/// so source-level disjointness is needed for correctness across page
+/// boundaries.
```
sergiupopescu199: For someone who's not very familiar with all the implementation details I suppose we could make the Rust doc a bit more "novice friendly", bysplitting the informations across paragraphs where we could mix the high level overview with technical more detailed information.

This is mostly a draft but does not represent the final revision of this rust doc.

```rust
/// Builds an SQL query representing objects from `checkpointed_objects` where
/// the current `object_version` is less than or equal to the `parent_version`.
///
/// Since `checkpointed_objects` tracks the latest version of each object, this 
/// condition implies that the largest known version of the object falls within 
/// the requested window, removing the need for an additional `objects_version` lookup.
///
/// # Constraints
///
/// Excludes rows present in `objects_backward_history` with the same 
/// `(object_id, object_version)` to ensure both sources remain disjoint. 
///
/// This deduplication is necessary because a specific `(id, version)` pair may 
/// transiently exist in both tables during the race window between 
/// `objects_backward_history` and `checkpointed_objects` writes. 
///
/// # Implementation Notes
///
/// Because the merge step utilizes `DISTINCT ON` solely for intra-page 
/// deduplication, ensuring disjointness at the source level is required 
/// to maintain correctness across page boundaries.
```

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
```diff
@@ -0,0 +1,126 @@
...
+    );
+
+    let source = query!("SELECT candidates.* FROM ({}) candidates", no_overlap);
+    page.apply::<StoredBackwardObject>(source)
+}
+
+/// Source B: rows in `objects_backward_history` whose `object_version` equals
+/// the largest `objects_version` entry `<= parent_version` for that
+/// `object_id`. This row carries the prior-state data of the object as it
+/// was at `parent_version`.
```
sergiupopescu199: Very similar to [comment](https://github.com/iotaledger/iota/pull/11477/changes#r3219290074)

sergiupopescu199 (commented):
lgtm 🍕

Added some non critical comments to better present the Rust doc.

sergiupopescu199 (approved):

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
tomxey: done 2ac96c9b0c

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
tomxey: done d623642098

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
tomxey: done d623642098

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
tomxey: done 10e55f1be5

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
tomxey: done 10e55f1be5

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
tomxey: done 10e55f1be5

crates/iota-graphql-rpc/src/backward_view/version_pinned.rs
tomxey: done 10e55f1be5

PR #11486, opened 2026-05-10
feat(ci): add bump version workflow

thibault-martinez:
# Description of change

Adds a manually-triggered `Bump version` GitHub Actions workflow plus a refactored `scripts/bump_version.sh`, so version bumps no longer need to be done by hand.

The workflow takes two structured inputs instead of a free-form version string:

- `bump` — `none`, `minor`, or `patch`. Controls which segment of `MAJOR.MINOR.PATCH` is incremented. `none` keeps the base version unchanged, which is what's used to promote a pre-release in place (e.g. `alpha → beta → rc → release`).
- `prerelease` — `alpha`, `beta`, `rc`, or `release`. Sets the pre-release suffix; `release` means no suffix.
- `base_branch` — branch the resulting PR is opened against (defaults to `develop`).

`scripts/bump_version.sh` reads the current workspace version from `Cargo.toml`, computes the new version from the two inputs, and propagates it to:

- `Cargo.toml` (only `[workspace.package].version`)
- `crates/iota-open-rpc/spec/openrpc.json`
- `crates/iota-graphql-e2e-tests/tests/call/simple.snap` (both the `*-testing-no-sha` form and the `MAJOR.MINOR` `availableVersions` entry)
- `Cargo.lock` (via `cargo update --workspace`)

The script also writes `IOTA_VERSION=<new-version>` to `$GITHUB_ENV` when running under Actions, so the workflow's PR-creation step can reuse the resolved version directly without re-parsing `Cargo.toml`.

Finally, `release` is added to the allowed PR title prefixes in `.github/workflows/pr_lint.yml`, so the auto-generated `release: bump version to ...` PR passes title linting.

Common invocations:

| Current        | Goal               | Inputs            | Result          |
| -------------- | ------------------ | ----------------- | --------------- |
| `1.24.0`       | start next cycle   | `minor` / `alpha` | `1.25.0-alpha`  |
| `1.24.0-alpha` | promote to beta    | `none`  / `beta`  | `1.24.0-beta`   |
| `1.24.0-rc`    | release            | `none`  / `release` | `1.24.0`      |
| `1.24.0`       | patch release      | `patch` / `release` | `1.24.1`      |

N/A.

- [x] Local dry-run of `scripts/bump_version.sh` against a stub workspace covering `minor alpha`, `none beta`, `patch release`, and `none release`, plus the no-op path (same args twice) and the input-validation paths (bad first arg, bad second arg, wrong arg count).
- [x] Verified the `[workspace.package]`-scoped sed leaves sibling tables with their own `version =` keys untouched.
- [x] Verified the script appends `IOTA_VERSION=<new>` to `$GITHUB_ENV` when that variable is set.
- [ ] Trigger the workflow manually against `develop` with `bump=minor`, `prerelease=alpha` and confirm a PR is opened on `release/bump-version-<new-version>` with the expected title and file changes.
- [ ] Confirm the resulting PR passes `pr_lint` with the `release:` prefix.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -21,13 +21,6 @@ jobs:
   bump-version:
     runs-on: ubuntu-latest
     steps:
-      - name: Validate version
-        run: |
-          if [[ ! "${{ inputs.version }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9.]+)?$ ]]; then
-            echo "Error: '${{ inputs.version }}' is not a valid version (expected MAJOR.MINOR.PATCH[-prerelease])" >&2
-            exit 1
-          fi
-
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: ${{ inputs.base_branch }}
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -1,0 +1,46 @@
+name: Bump version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "New version (MAJOR.MINOR.PATCH[-prerelease], e.g. 1.24.0-alpha)"
+        type: string
+        required: true
+      base_branch:
+        description: "Branch to open the PR against"
+        type: string
+        required: true
+        default: develop
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version
+        run: |
+          if [[ ! "${{ inputs.version }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9.]+)?$ ]]; then
+            echo "Error: '${{ inputs.version }}' is not a valid version (expected MAJOR.MINOR.PATCH[-prerelease])" >&2
+            exit 1
+          fi
+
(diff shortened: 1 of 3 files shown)


bingyanglin (approved):

scripts/bump_version.sh
```diff
@@ -1,29 +1,45 @@
 #!/usr/bin/env bash
 # Bump the workspace version and propagate it to dependent files.
 #
-# Usage: scripts/bump_version.sh <new-version>
-# Example: scripts/bump_version.sh 1.24.0-alpha
+# Usage: scripts/bump_version.sh <minor|patch|none> <alpha|beta|rc|release>
```
Thoralf-M, line 4: major we don't want to support in this script?

.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,53 @@
...
+          - minor
+          - patch
+      prerelease:
+        description: "Pre-release suffix to set (release = no suffix)"
+        type: choice
+        required: true
+        options:
+          - alpha
+          - beta
+          - rc
```
Thoralf-M, line 22: With the workflow we can't create more fine grained versions like `alpha.1` or `rc.2`, that's fine?

scripts/bump_version.sh
thibault-martinez: Yeah, major is significant enough to have to do it manually

.github/workflows/bump_version.yml
thibault-martinez: Yeah, we don't really do that ever in the release cycle

Thoralf-M (approved):

alexsporn (approved):

PR #11508, opened 2026-05-12
feat: only the sponsor MoveAuthenticator should be executed on the pre-consensus phase for a sponsored transaction

### LINKED ISSUE #11262: [AA] Make sponsored txs execute the sender authenticator only at post-consensus phase
Current implementation uses `max_auth_gas` as gas budget when the Authenticator Functions are executed pre-consensus (handle_transaction_impl) and the user set gas budget when executed post-consensus (process_certificate). 

With this change, at the pre-consensus phase, if a sponsor MoveAuthenticator is present (and the sender signature is a  MoveAuthenticator) only execute this authenticator. 

This means that only the sponsor authenticator will abide to the `max_auth_gas` limit. In fact, if the sponsor authenticator passes pre-consensus, then when both authenticators will be executed at the post-consensus phase the user gas budget will become the real limit.

valeriyr:
# Description of change

Only the sponsor MoveAuthenticator should be executed on the pre-consensus phase for a sponsored transaction.

fixes #11262 

### Release Notes

- [x] Protocol: Enables a flow in which, if a transaction has both a MoveAuthenticator for a sender and a sponsor, then during the pre-consensus phase only the sponsor Authenticator Function is executed. Two new fields are added to the AuthContext, namely sender_auth_digest and sponsor_auth_digest, representing the digests of both signature fields.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-types/src/signature.rs
```diff
@@ -118,6 +119,10 @@ impl GenericSignature {
         matches!(self, GenericSignature::MoveAuthenticator(_))
     }
 
+    pub fn digest(&self) -> GenericSignatureDigest {
+        GenericSignatureDigest::new(default_hash(self))
```
miker83z: How can we make sure that this and `MoveAuthenticator::digest()` will always return the same value, even if we modify the implementation of one of the two?

crates/iota-e2e-tests/tests/abstract_account_tests.rs
```diff
@@ -1347,6 +1347,186 @@ async fn test_aa_sender_and_aa_sponsor_rejected_when_sponsor_aa_fails_with_enabl
...
+
+    Ok(())
+}
+
+/// With `pre_consensus_sponsor_only_move_authentication` enabled, the flag
+/// only skips the sender's MA pre-consensus check for *sponsored* transactions.
+/// For non-sponsored transactions, the sender's MA still runs pre-consensus, so
+/// a bad sender MA is still rejected before consensus.
+#[sim_test]
+async fn test_non_sponsored_tx_sender_aa_rejected_pre_consensus_with_sponsor_only_flag()
```
lzpap, line 1505: Should we add one more test case when the sender is an MA, but sponsor is another generic sig variant? It should fail pre-consensus in that case, right?

valeriyr: In this version, it is allowed. In this case, only the sponsor signature is validated during pre-consensus.
We can discuss if we want to introduce this limitation on the protocol level.

crates/iota/tests/cli_tests.rs
```diff
@@ -6363,15 +6363,13 @@ async fn setup_move_authenticator_account(
...
+        .coin_read_api()
+        .get_coins(publisher, None, None, None)
         .await?
         .data
         .first()
         .unwrap()
-        .object()
-        .unwrap()
-        .object_id;
+        .coin_object_id;
```
valeriyr, line 6628: Used coin_read_api to filter for Coin<IOTA> objects only; get_owned_objects
returns all owned objects (including non-coin ones like UpgradeCap) and the
ordering may place non-coin objects first.

valeriyr: @Thoralf-M

crates/iota-types/src/signature.rs
valeriyr: GenericSignatureDigest was removed; digests for authentication are calculated in the following function:
https://github.com/iotaledger/iota/blob/9104ec66d58ed5a03d34b477b555d387340c9e67/crates/iota-types/src/transaction.rs#L2122

miker83z (commented):
LGTM!

crates/iota/tests/cli_tests.rs
Thoralf-M: Makes sense, good change 👍

crates/iota-types/src/transaction.rs
```diff
@@ -2093,6 +2094,65 @@ impl SenderSignedData {
...
+        #[allow(deprecated)]
+        let compute_digest = |address: IotaAddress| {
+            self.tx_signatures()
+                .iter()
+                .find(|sig| IotaAddress::try_from(*sig).ok() == Some(address))
+                .map(|sig| match sig {
+                    GenericSignature::MoveAuthenticator(authenticator) => authenticator.digest(),
+                    GenericSignature::MultiSig(_)
+                    | GenericSignature::Signature(_)
+                    | GenericSignature::ZkLoginAuthenticatorDeprecated(_)
```
Thoralf-M: ZkLoginAuthenticatorDeprecated should also return an error I guess

valeriyr: fixed in c7aa4e11c7

Thoralf-M (approved):
Approving the changes in iota-types

lzpap (approved):
lgtm!

miker83z (approved):

kodemartin (approved):

crates/iota-protocol-config/src/lib.rs
```diff
@@ -442,6 +444,10 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     enable_move_authentication_for_sponsor: bool,
 
+    // If true, only sponsor Move authentication is performed pre-consensus.
+    #[serde(skip_serializing_if = "is_false")]
+    pre_consensus_sponsor_only_move_authentication: bool,
```
muXxer: IMO we should follow the practice to insert new parameters always in chronological order. Thats much easier to read later on, and easier to understand which parameter was added when.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1647,6 +1653,23 @@ impl ProtocolConfig {
         enable_move_authentication_for_sponsor
     }
 
+    pub fn pre_consensus_sponsor_only_move_authentication(&self) -> bool {
```
muXxer: Also move the getter and the setten then to another position.

muXxer (changes requested):

crates/iota-protocol-config/src/lib.rs
valeriyr: Moved in 577ca7d765
But it would be nice to sort them if we wish, because, I think, they are not really sorted.

crates/iota-protocol-config/src/lib.rs
valeriyr: Moved in 577ca7d765

piotrm50 (approved):

semenov-vladyslav (approved):

muXxer (approved):

PR #11506, opened 2026-05-12
refactor(types,json-rpc)!: consolidate the IotaObjectResponseError types

kodemartin:
# Description of change

This patch follows #11380 and is based on the observation that we can consolidate the two `IotaObjectResponseError` variants living in `iota-types` and `iota-json-rpc-types` into the latter library. 

This is plausible because `IotaObjectResponseError` only lives in the context of JSON-RPC. Every call site that constructs, pattern-matches, or transports it sits in crates dependent on `iota-json-rpc-types`: `iota-json-rpc`, `iota-indexer`, `iota-cluster-test`, `iota-replay`,  `iota-source-validation`, the `iota` RPC client paths. `iota-types` doesn't actually need it for internal node logic.

The JSON-RPC wire format an the OpenRPC schema representation remain the same.

## Links to any relevant issues

Closes #11385

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)

### Infrastructure QA (only required for crates that are maintained by @iotaledger/infrastructure)

- [x] Verification of API backward compatibility.

PR #11523, opened 2026-05-13
chore: remove unused dependencies

thibault-martinez:
## Summary

Ran `cargo machete` and removed unused dependencies across ~45 crates. Where machete reported false positives (deps used only via macro expansion, build.rs-generated code, or feature flags), the dep is kept and added to a `[package.metadata.cargo-machete]` ignored list with a comment explaining why.

## Compilation time

Most removed deps were internal workspace crates, so a full `cargo build --workspace` still compiles every member and total work is roughly unchanged. The wins show up in:

- **Targeted builds** (`cargo check -p <crate>`): fewer transitive crates need to be compiled when working on a single package.
- **Parallelism**: each removed edge drops a "wait for B before A" constraint, so the critical path can shrink — especially around hot crates like `iota-core`, `iota-types`, `typed-store`.
- **Incremental rebuilds**: editing a crate invalidates fewer dependents. Trimming edges into widely-depended-on crates (e.g. `typed-store`) narrows the fanout when upstream crates change.
- **Fully-dropped external deps**: `fastcrypto-zkp` (was pulled in by `iota-core`, `iota-node`, `iota-transaction-checks`, `iota`) and `im` (was pulled in by `iota-core`, `iota-graphql-rpc`) now drop out of the build graph entirely — real wins on clean builds.

## Notable false positives ignored

- `iota-stardust-types` → `iota-crypto` (imported under the alias `crypto`)
- `typed-store` → `iota-metrics` (referenced from code generated by the `debug_fatal!` macro)
- `iota-faucet` → `tempfile` (referenced from code generated by the `DBMapUtils` derive macro)
- `starfish-core` → `tonic-prost` (referenced from build.rs-generated tonic stubs)
- `move-binary-format` → `getrandom` (used via the `wasm` feature)
- `move-core-types` → `enum-compat-util` (referenced from code generated by the `test_variant_order` macro)
- `move-compiler` → `stacker` (used by code generated by the `growing_stack` macro)
- `docs-examples` → `move-binary-format` (`serialize` method is gated by its `fuzzing` feature)
- `tree-sitter-move` → `tree-sitter` (consumed by `bindings/rust/lib.rs`, which is not checked in)

`proptest-derive` was also removed from `iota-types`'s `fuzzing` feature alongside the dep itself.

## Test plan

- [x] `cargo machete` reports no unused dependencies
- [x] `cargo check --workspace --all-targets` passes
- [x] `cargo check --all-targets -p iota-types --features fuzzing` passes
- [x] `cargo check` passes in each excluded standalone crate (`docs/examples/rust`, `examples/tic-tac-toe/cli`, `examples/custom-indexer/rust`)
- [x] `cargo check --all-targets` passes in the `external-crates/move` workspace


iota-execution/Cargo.toml
```diff
@@ -23,7 +23,6 @@ move-vm-config.workspace = true
 # move-bytecode-verifier-$CUT = { path = "../external-crates/move/move-execution/$CUT/crates/move-bytecode-verifier" }
 move-vm-runtime-latest = { path = "../external-crates/move/crates/move-vm-runtime", package = "move-vm-runtime" }
 # move-vm-runtime-$CUT = { path = "../external-crates/move/move-execution/$CUT/crates/move-vm-runtime" }
-move-vm-types-latest = { path = "../external-crates/move/crates/move-vm-types", package = "move-vm-types" }
```
thibault-martinez, line 26: @iotaledger/vm-language Just checking if this one has special meaning or if it can be treated as any dependency and be removed

crates/iota-json-rpc-types/Cargo.toml
```diff
@@ -22,7 +22,6 @@ serde_json.workspace = true
 serde_with.workspace = true
 strum.workspace = true
 tabled.workspace = true
-thiserror.workspace = true
```
kodemartin, line 25: yep, this should be reverted following https://github.com/iotaledger/iota/pull/11506

thibault-martinez: Just saw it 😬

muXxer (approved):

kodemartin (approved):

semenov-vladyslav (commented):
Should there be a ci workflow for `cargo machete`?
`cargo update` is great, but it seems orthogonal to the PR's main purpose.

thibault-martinez:
> Should there be a ci workflow for `cargo machete`? 

Yes, I'll add one at some point.

> `cargo update` is great, but it seems orthogonal to the PR's main purpose.

I didn't mean to do it, reverted.

iota-execution/Cargo.toml
valeriyr: `move-vm-types` is a part of a cut in IOTA, but as far as I see, not in Sui anymore.
I would revert this change for now and clean it in a dedicated PR, including the scripts related to cuts.

piotrm50 (approved):

iota-execution/Cargo.toml
thibault-martinez: reverted, would you do the changes you are talking about?

semenov-vladyslav (approved):

Alex6323 (approved):

valeriyr (approved):

iota-execution/Cargo.toml
valeriyr: I will create an issue for this, thank you!

PR #11527, opened 2026-05-13
fix(consensus): reject wrong-version commits and shards

VorobyevIlya:
# Description of change

A peer responding to consensus-sync requests can send tagged-enum
payloads whose variant does not match the local
`consensus_fast_commit_sync` flag — V2 commits to a flag-off node, V1
commits to a flag-on node, or V1-tagged `ShardWithProof` to a flag-on
node. The bytes parse cleanly today and no version check rejects them,
leaving room for silent state divergence between the receiver and the
rest of the network.

Two consistency checks at the network-reception entry points, each in
its own merged PR:

- `Commit`: `check_commit_version_matches_flags` at
  `Inner::verify_commits`. Counted by the existing
  `commit_sync_fetch_once_errors` metric.
- `ShardWithProof`: `check_shard_version_matches_flags` in
  `extract_shards_from_bundle`. Counted by the existing
  `bundles_with_invalid_parts` metric.

Store-recovery paths (local trusted history) skip both checks.

### DIFF
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -533,34 +533,6 @@ impl ShardWithProof {
     }
 }
 
-/// Rejects a deserialized `ShardWithProof` whose variant does not match the
-/// local protocol-flag configuration. The flag is uniform across the network
-/// within an epoch, so any mismatch implies either a malicious peer or a
-/// misconfigured upgrade path. In the flag-OFF (raw V1 wire form) branch the
-/// check is a tautology — the deserializer always produces V1 — but it is
-/// called there for symmetry.
-pub(crate) fn check_shard_version_matches_flags(
-    shard: &ShardWithProof,
-    protocol_config: &iota_protocol_config::ProtocolConfig,
-) -> ConsensusResult<()> {
-    let fast_commit_sync = protocol_config.consensus_fast_commit_sync();
-    let variant_matches_flags = matches!(
-        (shard, fast_commit_sync),
-        (ShardWithProof::V1(_), false) | (ShardWithProof::V2(_), true)
-    );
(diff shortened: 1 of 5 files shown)


crates/starfish/core/src/commit.rs
```diff
@@ -127,6 +127,34 @@ impl Commit {
     }
 }
 
+/// Rejects a deserialized commit whose variant does not match the local
+/// protocol-flag configuration. The flag is uniform across the network within
+/// an epoch, so any mismatch implies either a malicious peer or a misconfigured
+/// upgrade path. Called on commits received over the network in
+/// `commit_syncer::Inner::verify_commits`; recovery from local store is trusted
+/// and skips this check.
+pub(crate) fn check_commit_version_matches_flags(
```
piotrm50: Why is this implemented in this file if it's only used and tested in `commit_syncer/mod.rs`?

crates/starfish/core/src/block_header.rs
```diff
@@ -533,6 +533,34 @@ impl ShardWithProof {
     }
 }
 
+/// Rejects a deserialized `ShardWithProof` whose variant does not match the
```
piotrm50: Same here - why put this in this file if usage and testing is elsewhere?

VorobyevIlya: It is the file where ShardWithProof is defined, but I moved it now

crates/starfish/core/src/commit.rs
VorobyevIlya: Commit V1 and V2 are defined in this file, but you're right, it is probably better to move it

piotrm50 (approved):

filipdulic (approved):

PR #11528, opened 2026-05-13
feat(iota-core, iota-types): add validator attestation

vekkiokonio:
> This PR merges into `protocol-research/feat/transaction-attestation-feature`, not `develop`. That feature branch is itself branched from `consensus/feat/pcool-feature`.

# Description of change

This PR implements the **[Validator Attestation (Phase 1)](https://github.com/iotaledger/iota-private/issues/383)** — the mechanism by which a proposing validator certifies, pre-consensus, that a transaction has passed validation and provides an estimated computation cost for scheduler hints.

_Note: related [changes to sequencer and execution](https://github.com/iotaledger/iota-private/issues/386) will be part of a separate PR_.

### What changed

**New data structures** (`iota-types`)
- `attestation.rs`: `Attestation` (two variants — `Validator`, `Explicit`), `AttestationData` (versioned, carries estimated computation cost and observed shared-object versions), `AttestedTransaction` (transaction + attestation bundle). BCS round-trip tests included.
- `messages_consensus.rs`: new `UserTransactionV2(Box<AttestedTransaction>)` variant in `ConsensusTransactionKind`, carrying the bundled attestation.
- `error.rs`: `AttestationAuthorMismatch` and `ExplicitAttestationNotSupported` error variants.
- `iota-protocol-config`: `enable_validator_attestation` feature flag. Attestation is only active when `enable_white_flag_flow` is also on; the two flags are enforced as co-dependent.

**Pre-consensus attestation** (`authority.rs`, `validator_v2.rs`)
- `attest_transaction`: validates the transaction (deny checks, gas, ownership, coin deny list, full dry-run) and returns `(AttestationData, Vec<ObjectRef>)`. It subsumes `handle_transaction_validation_checks`, so the two paths are mutually exclusive.
- `submit_single_tx` restructured: when `enable_validator_attestation` is on, calls `attest_transaction` instead of `handle_transaction_validation_checks`. The pre-consensus soft lock is acquired after attestation (owned objects are returned by `attest_transaction`, so no extra pass is needed). The transaction is submitted as `UserTransactionV2` with the attestation embedded.

**Batch homogeneity** (`consensus_adapter.rs`)
- `submit_batch` enforces that all transactions in a soft-bundle are of the same kind: all-V1, all-V2, or all-certificates. Mixed batches are rejected with `InvalidTxKindInSoftBundle`.

**Post-consensus attestor verification** (`post_consensus_validation.rs`)
- `UserTransactionV2` match arm separated into pure classification; digest is pushed to `all_user_tx_digests` before any check, so dropped V2 transactions still release pre-consensus soft locks.
- New Check # 3 (attestor verification): verifies that `attestor_index` in `Attestation::Validator` matches `certificate_author_index` (block author).
- `UserTransactionV2` transactions skip Check # 6 (`handle_transaction_validation_checks`) — the attestor already performed those checks before producing the attestation.

**`consensus_validator.rs`**: comment updated to clarify that attestor verification happens in `post_consensus_validation`, not during block ingestion (where block-author context is unavailable).

**Unit tests**
- `validator_v2_tests.rs`: unit tests for the V2 submission path — happy path (transaction is attested and accepted) and rejection path (attestation fails with an invalid transaction).
- `post_consensus_validation_tests.rs`: unit tests for `UserTransactionV2` handling in post-consensus validation — correct attestor accepted, mismatched attestor rejected, and `UserTransactionV2` skipping duplicate validation checks.

**End-to-end test (iota-e2e-tests)**

- `attestation_tests.rs`: `test_aa_tx_accepted_via_v2_attestation_path` submits a Move abstract account transaction through the V2 gRPC path with both `enable_white_flag_flow` and `enable_validator_attestation` active. Exercises the `authenticate_then_execute_transaction_to_effects` branch of `attest_transaction` (MoveAuthenticator dry-run path).

Closes https://github.com/iotaledger/iota-private/issues/384
Closes https://github.com/iotaledger/iota-private/issues/385
Part of Epic https://github.com/iotaledger/iota-private/issues/383

BCS round-trip tests for `AttestationData`, `Attestation::Validator`, `Attestation::Explicit`, and `AttestedTransaction` are included in `attestation.rs`. Unit tests for the V2 submission and post-consensus validation paths are in `validator_v2_tests.rs` and `post_consensus_validation_tests.rs`. An end-to-end test exercising the full attestation pipeline is in `attestation_tests.rs`.

### Release Notes

- [x] Protocol: New `UserTransactionV2` consensus transaction kind and `enable_validator_attestation` protocol config flag (requires `enable_white_flag_flow`).
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -61,6 +61,7 @@ use iota_types::{
             AuthenticatorFunctionRefV1,
         },
     },
+    attestation::Attestation,
     base_types::{
         AuthorityName, ConciseableName, IotaAddress, ObjectID, ObjectInfo, ObjectRef, ObjectType,
         SequenceNumber, StructTag, TypeTag, VersionNumber,
@@ -180,7 +181,7 @@ use crate::{
     stake_aggregator::StakeAggregator,
     subscription_handler::SubscriptionHandler,
     transaction_input_loader::TransactionInputLoader,
-    transaction_manager::TransactionManager,
+    transaction_manager::{TransactionManager, VerifiedExecutableAttestedTransaction},
     transaction_outputs::TransactionOutputs,
     validator_tx_finalizer::ValidatorTxFinalizer,
     verify_indexes::verify_indexes,
@@ -1105,8 +1106,15 @@ impl AuthorityState {
             (inner_temp_store, effects)
         };
 
-        // Build AttestationData.
-        let estimated_computation_cost = effects.gas_cost_summary().computation_cost;
+        // Step 7: build AttestationData.
+        // `gas_cost_summary().computation_cost` is in NANOS; convert to gas
+        // units (`computation_units = computation_cost / gas_price`) so the
+        // attestation is independent of the gas price the user chose.
+        let computation_units = effects
+            .gas_cost_summary()
+            .computation_cost
+            .checked_div(tx_data.gas_price())
+            .unwrap_or(0);
 
         // `object_versions` is the evidence base for malicious-attestor detection. We
         // pull `input_objects` (resolved tx + authenticator inputs) and
@@ -1149,7 +1157,7 @@ impl AuthorityState {
 
         Ok((
             AttestationData::V1 {
-                estimated_computation_cost,
+                computation_units,
                 object_versions,
             },
             owned_objects,
@@ -1363,7 +1371,7 @@ impl AuthorityState {
     #[instrument(level = "trace", skip_all, fields(tx_digest = ?certificate.digest()))]
     pub fn try_execute_immediately(
         &self,
-        certificate: &VerifiedExecutableTransaction,
+        certificate: &VerifiedExecutableAttestedTransaction,
         expected_effects_digest: Option<TransactionEffectsDigest>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<(TransactionEffects, Option<ExecutionError>)> {
@@ -1450,11 +1458,10 @@ impl AuthorityState {
         certificate: &VerifiedCertificate,
     ) -> IotaResult<(VerifiedSignedTransactionEffects, Option<ExecutionError>)> {
         let epoch_store = self.epoch_store_for_testing();
-        let (effects, execution_error_opt) = self.try_execute_immediately(
-            &VerifiedExecutableTransaction::new_from_certificate(certificate.clone()),
-            None,
-            &epoch_store,
-        )?;
+        let executable: VerifiedExecutableAttestedTransaction =
+            VerifiedExecutableTransaction::new_from_certificate(certificate.clone()).into();
+        let (effects, execution_error_opt) =
+            self.try_execute_immediately(&executable, None, &epoch_store)?;
         let signed_effects = self.sign_effects(effects, &epoch_store)?;
         Ok((signed_effects, execution_error_opt))
     }
@@ -1523,7 +1530,7 @@ impl AuthorityState {
     pub(crate) fn process_certificate(
         &self,
         tx_guard: CertTxGuard,
-        certificate: &VerifiedExecutableTransaction,
+        certificate: &VerifiedExecutableAttestedTransaction,
         tx_input_objects: InputObjects,
         per_authenticator_inputs: Vec<(InputObjects, ObjectReadResult)>,
         expected_effects_digest: Option<TransactionEffectsDigest>,
@@ -1761,7 +1768,7 @@ impl AuthorityState {
     fn prepare_certificate(
         &self,
         _execution_guard: &ExecutionLockReadGuard<'_>,
-        certificate: &VerifiedExecutableTransaction,
+        certificate: &VerifiedExecutableAttestedTransaction,
         tx_input_objects: InputObjects,
         per_authenticator_inputs: Vec<(InputObjects, ObjectReadResult)>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
@@ -1792,6 +1799,20 @@ impl AuthorityState {
         // checks.
         let tx_data = certificate.data().transaction_data();
         tx_data.validity_check(protocol_config)?;
+
+        // The user signature of an attested (`UserTransactionV2`) transaction is
+        // verified pre-consensus in the block verifier
+        // (`IotaTxValidator::validate_transactions`), exactly as for
+        // `UserTransactionV1`, and is not re-checked here.
+        //
+        // Explicit attestations are rejected pre-consensus (block verifier) and
+        // at post-consensus (Check #3), so one reaching here is a protocol-level
+        // bug.
+        if let Some(Attestation::Explicit { .. }) = certificate.attestation() {
+            return Err(IotaError::UnsupportedFeature {
+                error: "Explicit attestation reached prepare_certificate".into(),
+            });
+        }
 
         let (kind, signer, gas_data) = tx_data.execution_parts();
 
@@ -1985,10 +2006,11 @@ impl AuthorityState {
     )> {
         let lock = RwLock::new(epoch_store.epoch());
         let execution_guard = lock.try_read().unwrap();
+        let attested: VerifiedExecutableAttestedTransaction = certificate.clone().into();
 
         self.prepare_certificate(
             &execution_guard,
-            certificate,
+            &attested,
             input_objects,
             vec![],
             epoch_store,
@@ -5403,9 +5425,10 @@ impl AuthorityState {
         let (input_objects, _) =
             self.read_objects_for_execution(&tx_lock, &executable_tx, epoch_store)?;
 
+        let attested_tx: VerifiedExecutableAttestedTransaction = executable_tx.clone().into();
         let (temporary_store, effects, _execution_error_opt) = self.prepare_certificate(
             &execution_guard,
-            &executable_tx,
+            &attested_tx,
             input_objects,
             vec![],
             epoch_store,
@@ -5713,17 +5736,21 @@ impl AuthorityState {
                 is_execute_transaction_to_effects,
             )?;
 
-        // Step 5: coin deny list.
-        let per_authenticator_checked_input_objects: Vec<&_> = per_authenticator_checked_inputs
+        // Step 5: coin deny list. Pass raw `&InputObjects` (the deny-list
+        // check only reads coin types and does not need the
+        // `CheckedInputObjects` type-state — which is still enforced fresh
+        // at `check_certificate_input` in `prepare_certificate`).
+        let per_authenticator_input_objects: Vec<&InputObjects> = per_authenticator_checked_inputs
             .iter()
-            .map(|i| &i.0)
+            .map(|i| i.0.inner())
             .collect();
         check_coin_deny_list_v1(
             tx_data.sender(),
-            &tx_checked_input_objects,
+            tx_checked_input_objects.inner(),
             &tx_receiving_objects,
-            &per_authenticator_checked_input_objects,
+            &per_authenticator_input_objects,
             &self.get_object_store(),
+            epoch,
         )?;
 
         Ok(ValidationOutputs {
@@ -5733,6 +5760,57 @@ impl AuthorityState {
             tx_checked_input_objects,
             per_authenticator_checked_inputs,
         })
+    }
+
+    /// Re-runs the sender-side coin deny-list check for an attested
+    /// (`UserTransactionV2`) transaction in the post-consensus phase.
+    ///
+    /// An honest attestor performs this check at attestation time, but the deny
+    /// list is a Move object whose state may change between attestation and
+    /// execution. Running the check at execution time would crash the
+    /// validator, so we run it during post-consensus validation where the
+    /// transaction can still be dropped.
+    pub(crate) fn check_coin_deny_list_for_attested_tx(
+        &self,
+        transaction: &VerifiedTransaction,
+        epoch: EpochId,
+    ) -> IotaResult<()> {
+        let (tx_input_objects, tx_receiving_objects, per_authenticator_inputs) =
+            self.read_objects_for_validation(transaction, epoch)?;
+        let per_authenticator_input_objects: Vec<&InputObjects> =
+            per_authenticator_inputs.iter().map(|(io, _)| io).collect();
+        check_coin_deny_list_v1(
+            transaction.data().transaction_data().sender(),
+            &tx_input_objects,
+            &tx_receiving_objects,
+            &per_authenticator_input_objects,
+            &self.get_object_store(),
+            epoch,
+        )?;
+        Ok(())
+    }
+
+    /// Re-runs the `TransactionDenyConfig` deny checks (disabled features,
+    /// denied signers, denied input/receiving objects, denied package
+    /// dependencies) for an attested (`UserTransactionV2`) transaction in the
+    /// post-consensus phase.
+    ///
+    /// NOTE: the deny config is sourced from each validator's own `NodeConfig`,
+    /// so this check is not deterministic across validators.
+    pub(crate) fn check_transaction_deny_list_for_attested_tx(
+        &self,
+        transaction: &VerifiedTransaction,
+    ) -> IotaResult<()> {
+        let tx_data = transaction.data().transaction_data();
+        iota_transaction_checks::deny::check_transaction_for_validation(
+            tx_data,
+            transaction.tx_signatures(),
+            &transaction.input_objects()?,
+            &tx_data.receiving_objects(),
+            &self.config.transaction_deny_config,
+            self.get_backing_package_store().as_ref(),
+        )?;
+        Ok(())
     }
 
     #[allow(clippy::type_complexity)]
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -837,6 +837,17 @@ pub struct AuthorityState {
 /// The authority state encapsulates all state, drives execution, and ensures
 /// safety.
 ///
+/// Outputs of the shared transaction validation pipeline used by both
+/// `handle_transaction_validation_checks` and `attest_transaction`.
+struct ValidationOutputs<'a> {
+    tx_input_objects: InputObjects,
+    per_authenticator_inputs: Vec<(InputObjects, ObjectReadResult)>,
+    move_authenticators: Vec<&'a MoveAuthenticator>,
+    gas_status: IotaGasStatus,
+    tx_checked_input_objects: CheckedInputObjects,
+    per_authenticator_checked_inputs: Vec<(CheckedInputObjects, AuthenticatorFunctionRef)>,
+}
+
 /// Note the authority operations can be accessed through a read ref (&) and do
 /// not require &mut. Internally a database is synchronized through a mutex
 /// lock.
@@ -887,67 +898,21 @@ impl AuthorityState {
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> IotaResult<Vec<ObjectRef>> {
         let protocol_config = epoch_store.protocol_config();
-        let reference_gas_price = epoch_store.reference_gas_price();
-
-        let epoch = epoch_store.epoch();
-
         let tx_data = transaction.data().transaction_data();
 
-        // Note: the deny checks may do redundant package loads but:
-        // - they only load packages when there is an active package deny map
-        // - the loads are cached anyway
-        iota_transaction_checks::deny::check_transaction_for_validation(
-            tx_data,
-            transaction.tx_signatures(),
-            &transaction.input_objects()?,
-            &tx_data.receiving_objects(),
-            &self.config.transaction_deny_config,
-            self.get_backing_package_store().as_ref(),
-        )?;
+        let ValidationOutputs {
+            move_authenticators,
+            gas_status,
+            tx_checked_input_objects,
+            per_authenticator_checked_inputs,
+            tx_input_objects: _,
+            per_authenticator_inputs: _,
+        } = self.run_validation_checks(transaction, epoch_store)?;
 
-        // Load all transaction-related input objects.
-        // Authenticator input objects and the account object are loaded in the same
-        // call if there is a sender `MoveAuthenticator` signature present in the
-        // transaction.
-        let (tx_input_objects, tx_receiving_objects, per_authenticator_inputs) =
-            self.read_objects_for_validation(transaction, epoch)?;
-
-        // Get the `MoveAuthenticator`s, if any.
-        let move_authenticators = transaction.move_authenticators();
-
-        // Check the inputs for signing.
-        // If there are `MoveAuthenticator` signatures, their input objects and the
-        // account objects are also checked and must be provided.
-        // It is also checked if there is enough gas to execute the transaction and its
-        // authenticators.
-        let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
-            .check_transaction_inputs_for_validation(
-                protocol_config,
-                reference_gas_price,
-                tx_data,
-                tx_input_objects,
-                &tx_receiving_objects,
-                &move_authenticators,
-                per_authenticator_inputs,
-            )?;
-
-        // Get the input objects for the authenticators, if there are
-        // `MoveAuthenticator`s.
-        let per_authenticator_checked_input_objects = per_authenticator_checked_inputs
+        let per_authenticator_checked_input_objects: Vec<&_> = per_authenticator_checked_inputs
             .iter()
             .map(|i| &i.0)
             .collect();
-
-        // Check if any of the sender, the transaction input objects, the receiving
-        // objects and the authenticator input objects are in the coin deny
-        // list, which would prevent the transaction from being signed.
-        check_coin_deny_list_v1(
-            tx_data.sender(),
-            &tx_checked_input_objects,
-            &tx_receiving_objects,
-            &per_authenticator_checked_input_objects,
-            &self.get_object_store(),
-        )?;
 
         // If there are `MoveAuthenticator` signatures, execute them and check if they
         // all succeed.
@@ -1025,9 +990,11 @@ impl AuthorityState {
     /// in a validator attestation and attached to the transaction before
     /// submitting it to consensus.
     ///
-    /// Subsumes `handle_transaction_validation_checks`: it runs the same
-    /// pipeline — deny checks, object loading, input validation, coin deny
-    /// list — and then replaces the final `authenticate_transaction` call with
+    /// Shares the validation pipeline with
+    /// `handle_transaction_validation_checks`
+    /// via [`Self::run_validation_checks`] (deny checks, input object loading,
+    /// input checks, coin deny list), and then replaces the
+    /// `authenticate_transaction` call with
     /// `authenticate_then_execute_transaction_to_effects` so that Move
     /// authentication and transaction execution run together and produce full
     /// `TransactionEffects` including the accurate `computation_cost`.
@@ -1040,50 +1007,16 @@ impl AuthorityState {
 
         let protocol_config = epoch_store.protocol_config();
         let reference_gas_price = epoch_store.reference_gas_price();
-        let epoch = epoch_store.epoch();
         let tx_data = transaction.data().transaction_data();
 
-        // Step 1: deny checks.
-        iota_transaction_checks::deny::check_transaction_for_validation(
-            tx_data,
-            transaction.tx_signatures(),
-            &transaction.input_objects()?,
-            &tx_data.receiving_objects(),
-            &self.config.transaction_deny_config,
-            self.get_backing_package_store().as_ref(),
-        )?;
-
-        // Step 2: load all input objects including authenticator inputs.
-        let (tx_input_objects, tx_receiving_objects, per_authenticator_inputs) =
-            self.read_objects_for_validation(transaction, epoch)?;
-
-        // Step 3: get Move authenticators.
-        let move_authenticators = transaction.move_authenticators();
-
-        // Step 4: check inputs for signing (validates gas budget covers auth + tx).
-        let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
-            .check_transaction_inputs_for_validation(
-                protocol_config,
-                reference_gas_price,
-                tx_data,
-                tx_input_objects.clone(),
-                &tx_receiving_objects,
-                &move_authenticators,
-                per_authenticator_inputs.clone(),
-            )?;
-
-        // Step 5: coin deny list.
-        let per_authenticator_checked_input_objects: Vec<&_> = per_authenticator_checked_inputs
-            .iter()
-            .map(|i| &i.0)
-            .collect();
-        check_coin_deny_list_v1(
-            tx_data.sender(),
-            &tx_checked_input_objects,
-            &tx_receiving_objects,
-            &per_authenticator_checked_input_objects,
-            &self.get_object_store(),
-        )?;
+        let ValidationOutputs {
+            tx_input_objects,
+            per_authenticator_inputs,
+            move_authenticators,
+            gas_status,
+            tx_checked_input_objects,
+            per_authenticator_checked_inputs: _,
+        } = self.run_validation_checks(transaction, epoch_store)?;
 
         let owned_objects = tx_checked_input_objects.inner().filter_owned_objects();
 
@@ -1096,9 +1029,8 @@ impl AuthorityState {
         let (kind, signer, gas_data) = tx_data.execution_parts();
         let tx_digest = *transaction.digest();
 
-        // Step 6: execute.
         let effects = if move_authenticators.is_empty() {
-            // Common path: no Move authentication, execute directly.
+            // No Move authentication, execute directly.
             let (_, _, effects, _) = epoch_store.executor().execute_transaction_to_effects(
                 backing_store.as_ref(),
                 protocol_config,
@@ -1117,9 +1049,8 @@ impl AuthorityState {
             );
             effects
         } else {
-            // MoveAuthenticator path: get AuthenticatorFunctionRefForExecution
-            // for each authenticator, then run auth + execution in one pass with an
-            // execution-mode gas status.
+            // get AuthenticatorFunctionRefForExecution for each authenticator, then run
+            // auth + execution in one pass with an execution-mode gas status.
             let tx_data_bytes =
                 bcs::to_bytes(tx_data).expect("TransactionData serialization cannot fail");
 
@@ -1183,13 +1114,13 @@ impl AuthorityState {
             effects
         };
 
-        // Step 7: build AttestationData.
+        // Build AttestationData.
         let estimated_computation_cost = effects.gas_cost_summary().computation_cost;
 
         // Collect all input object refs seen during execution. Start from the
-        // owned objects (extracted in step 5), then add shared objects as
-        // recorded in effects (includes authenticator + deny-list objects),
-        // receiving objects, and gas payment objects.
+        // owned objects, then add shared objects as recorded in effects (includes
+        // authenticator + deny-list objects), receiving objects, and gas payment
+        // objects.
         let mut seen_ids = std::collections::HashSet::new();
         let mut object_versions: Vec<ObjectRef> = Vec::new();
         let mut push = |oref: ObjectRef| {
@@ -5712,6 +5643,85 @@ impl AuthorityState {
     }
 
     #[allow(clippy::type_complexity)]
+    /// Shared validation pipeline for `handle_transaction_validation_checks`
+    /// and `attest_transaction`. Runs, in order:
+    ///   1. transaction deny checks
+    ///   2. input object loading (transaction + per-authenticator + account)
+    ///   3. `MoveAuthenticator` collection
+    ///   4. input checks (gas budget covers transaction + authenticators)
+    ///   5. coin deny list check
+    ///
+    /// Keep this helper as the single source of truth so the two callers
+    /// cannot drift. Raw inputs are cloned and returned for
+    /// `attest_transaction`'s MoveAuthenticator path;
+    /// `handle_transaction_validation_checks` ignores them.
+    fn run_validation_checks<'a>(
+        &self,
+        transaction: &'a VerifiedTransaction,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+    ) -> IotaResult<ValidationOutputs<'a>> {
+        let protocol_config = epoch_store.protocol_config();
+        let reference_gas_price = epoch_store.reference_gas_price();
+        let epoch = epoch_store.epoch();
+        let tx_data = transaction.data().transaction_data();
+
+        // Step 1: deny checks.
+        // Note: the deny checks may do redundant package loads but:
+        // - they only load packages when there is an active package deny map
+        // - the loads are cached anyway
+        iota_transaction_checks::deny::check_transaction_for_validation(
+            tx_data,
+            transaction.tx_signatures(),
+            &transaction.input_objects()?,
+            &tx_data.receiving_objects(),
+            &self.config.transaction_deny_config,
+            self.get_backing_package_store().as_ref(),
+        )?;
+
+        // Step 2: load all input objects (including authenticator + account).
+        let (tx_input_objects, tx_receiving_objects, per_authenticator_inputs) =
+            self.read_objects_for_validation(transaction, epoch)?;
+
+        // Step 3: get Move authenticators.
+        let move_authenticators = transaction.move_authenticators();
+
+        // Step 4: check inputs for signing (validates gas budget covers auth + tx).
+        // Clone raw inputs so attest_transaction's MoveAuthenticator path can
+        // reuse them in execution-mode.
+        let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
+            .check_transaction_inputs_for_validation(
+                protocol_config,
+                reference_gas_price,
+                tx_data,
+                tx_input_objects.clone(),
+                &tx_receiving_objects,
+                &move_authenticators,
+                per_authenticator_inputs.clone(),
+            )?;
+
+        // Step 5: coin deny list.
+        let per_authenticator_checked_input_objects: Vec<&_> = per_authenticator_checked_inputs
+            .iter()
+            .map(|i| &i.0)
+            .collect();
+        check_coin_deny_list_v1(
+            tx_data.sender(),
+            &tx_checked_input_objects,
+            &tx_receiving_objects,
+            &per_authenticator_checked_input_objects,
+            &self.get_object_store(),
+        )?;
+
+        Ok(ValidationOutputs {
+            tx_input_objects,
+            per_authenticator_inputs,
+            move_authenticators,
+            gas_status,
+            tx_checked_input_objects,
+            per_authenticator_checked_inputs,
+        })
+    }
+
     fn read_objects_for_validation(
         &self,
         transaction: &VerifiedTransaction,
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -704,7 +704,7 @@ impl AuthorityMetrics {
             ).unwrap(),
             consensus_handler_validation_dropped_transactions: register_int_counter_with_registry!(
                 "consensus_handler_validation_dropped_transactions",
-                "Number of UserTransactionV1 transactions dropped by post-consensus validation",
+                "Number of UserTransactionV1/V2 transactions dropped by post-consensus validation",
                 registry,
             ).unwrap(),
             consensus_handler_max_object_costs: register_int_gauge_vec_with_registry!(
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -5693,9 +5693,10 @@ impl AuthorityState {
     ///   5. coin deny list check
     ///
     /// Keep this helper as the single source of truth so the two callers
-    /// cannot drift. Raw inputs are cloned and returned for
-    /// `attest_transaction`'s MoveAuthenticator path;
-    /// `handle_transaction_validation_checks` ignores them.
+    /// cannot drift. Raw inputs are cloned and returned only on the
+    /// `attest_transaction` path (`is_execute_transaction_to_effects`), where
+    /// the MoveAuthenticator path needs them; the clone is skipped for
+    /// `handle_transaction_validation_checks`, which ignores them.
     fn run_validation_checks<'a>(
         &self,
         transaction: &'a VerifiedTransaction,
@@ -5730,9 +5731,19 @@ impl AuthorityState {
         // Step 4: check inputs (validates gas budget covers auth + tx). When
         // `is_execute_transaction_to_effects` is true (attest_transaction), the
         // gas budget is metered for full execution; otherwise (signing) it only
-        // covers authentication. Clone the raw per-authenticator inputs so
-        // attest_transaction can recover the `AuthenticatorFunctionRefForExecution`
-        // for each authenticator.
+        // covers authentication.
+        //
+        // Only the attest_transaction path consumes the raw per-authenticator
+        // inputs from `ValidationOutputs` (to recover each
+        // `AuthenticatorFunctionRefForExecution`);
+        // `handle_transaction_validation_checks` discards them. So only clone
+        // for the output when we actually need it, and move the original into
+        // the input check otherwise.
+        let per_authenticator_inputs_for_output = if is_execute_transaction_to_effects {
+            per_authenticator_inputs.clone()
+        } else {
+            Vec::new()
+        };
         let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
             .check_transaction_inputs_for_validation(
                 protocol_config,
@@ -5741,7 +5752,7 @@ impl AuthorityState {
                 tx_input_objects,
                 &tx_receiving_objects,
                 &move_authenticators,
-                per_authenticator_inputs.clone(),
+                per_authenticator_inputs,
                 is_execute_transaction_to_effects,
             )?;
 
@@ -5763,7 +5774,7 @@ impl AuthorityState {
         )?;
 
         Ok(ValidationOutputs {
-            per_authenticator_inputs,
+            per_authenticator_inputs: per_authenticator_inputs_for_output,
             move_authenticators,
             gas_status,
             tx_checked_input_objects,
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1028,7 +1028,7 @@ impl AuthorityState {
 
         // Capture the `InnerTemporaryStore` from the dry-run: its object maps are the
         // source of truth for the input versions the attestor observed.
-        let (inner_temp_store, effects) = if move_authenticators.is_empty() {
+        let (inner_temp_store, effects, authentication_failed) = if move_authenticators.is_empty() {
             // No Move authentication, execute directly.
             let (inner_temp_store, _, effects, _) =
                 epoch_store.executor().execute_transaction_to_effects(
@@ -1047,7 +1047,8 @@ impl AuthorityState {
                     tx_digest,
                     &mut None,
                 );
-            (inner_temp_store, effects)
+            // This branch has no Move authentication, so it can never fail auth.
+            (inner_temp_store, effects, false)
         } else {
             // Recover the `AuthenticatorFunctionRefForExecution` for each authenticator and
             // run auth + execution in one pass.
@@ -1083,7 +1084,7 @@ impl AuthorityState {
                 })
                 .collect::<Vec<_>>();
 
-            let (inner_temp_store, _, effects, _) = epoch_store
+            let (inner_temp_store, _, effects, _, authentication_failed) = epoch_store
                 .executor()
                 .authenticate_then_execute_transaction_to_effects(
                     backing_store.as_ref(),
@@ -1103,15 +1104,17 @@ impl AuthorityState {
                     tx_data_bytes,
                     &mut None,
                 );
-            (inner_temp_store, effects)
+            (inner_temp_store, effects, authentication_failed)
         };
 
-        // Refuse to attest a transaction whose dry-run did not succeed. This
-        // covers a failed Move authentication as well as any other execution
-        // failure.
-        if let ExecutionStatus::Failure { error, command } = effects.status() {
+        // Refuse to attest a transaction when its Move authentication did not succeed.
+        // Any other failure (out-of-gas or a Move abort in the transaction body) is the
+        // issuer's responsibility and is still attested, sequenced and charged as a
+        // failed transaction under the normal flow.
+        if authentication_failed {
             return Err(IotaError::Execution(format!(
-                "refusing to attest transaction with failing dry-run: {error:?} (command {command:?})"
+                "refusing to attest transaction with failed Move authentication: {:?}",
+                effects.status()
             )));
         }
 
@@ -1964,7 +1967,13 @@ impl AuthorityState {
                 .filter_owned_objects();
             self.check_owned_locks(&owned_object_refs)?;
 
-            epoch_store
+            let (
+                inner_temp_store,
+                gas_status,
+                effects,
+                execution_error_opt,
+                _authentication_failed,
+            ) = epoch_store
                 .executor()
                 .authenticate_then_execute_transaction_to_effects(
                     backing_store,
@@ -1985,7 +1994,8 @@ impl AuthorityState {
                     tx_digest,
                     tx_data_bytes,
                     &mut None,
-                )
+                );
+            (inner_temp_store, gas_status, effects, execution_error_opt)
         };
 
         fail_point_if!("cp_execution_nondeterminism", || {
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -1020,6 +1020,200 @@ impl AuthorityState {
         Ok(tx_checked_input_objects.inner().filter_owned_objects())
     }
 
+    /// Dry-runs a verified transaction and returns `AttestationData` ready to
+    /// be wrapped in a validator attestation and attached to the transaction
+    /// before submitting it to consensus.
+    ///
+    /// Follows the same pipeline as `handle_transaction_validation_checks`
+    /// — deny checks, object loading, input validation, coin deny list — but
+    /// replaces the final `authenticate_transaction` call with
+    /// `authenticate_then_execute_transaction_to_effects` so that Move
+    /// authentication and transaction execution run together and produce full
+    /// `TransactionEffects` including the accurate `computation_cost`.
+    pub(crate) fn attest_transaction(
+        &self,
+        transaction: &VerifiedTransaction,
+        owned_objects: &[ObjectRef],
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+    ) -> IotaResult<iota_types::attestation::AttestationData> {
+        use iota_types::attestation::AttestationData;
+
+        let protocol_config = epoch_store.protocol_config();
+        let reference_gas_price = epoch_store.reference_gas_price();
+        let epoch = epoch_store.epoch();
+        let tx_data = transaction.data().transaction_data();
+
+        // Step 1: deny checks.
+        iota_transaction_checks::deny::check_transaction_for_validation(
+            tx_data,
+            transaction.tx_signatures(),
+            &transaction.input_objects()?,
+            &tx_data.receiving_objects(),
+            &self.config.transaction_deny_config,
+            self.get_backing_package_store().as_ref(),
+        )?;
+
+        // Step 2: load all input objects including authenticator inputs.
+        let (tx_input_objects, tx_receiving_objects, per_authenticator_inputs) =
+            self.read_objects_for_validation(transaction, epoch)?;
+
+        // Step 3: get Move authenticators.
+        let move_authenticators = transaction.move_authenticators();
+
+        // Step 4: check inputs for signing (validates gas budget covers auth + tx).
+        let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
+            .check_transaction_inputs_for_validation(
+                protocol_config,
+                reference_gas_price,
+                tx_data,
+                tx_input_objects.clone(),
+                &tx_receiving_objects,
+                &move_authenticators,
+                per_authenticator_inputs.clone(),
+            )?;
+
+        // Step 5: coin deny list.
+        let per_authenticator_checked_input_objects: Vec<&_> = per_authenticator_checked_inputs
+            .iter()
+            .map(|i| &i.0)
+            .collect();
+        check_coin_deny_list_v1(
+            tx_data.sender(),
+            &tx_checked_input_objects,
+            &tx_receiving_objects,
+            &per_authenticator_checked_input_objects,
+            &self.get_object_store(),
+        )?;
+
+        let epoch_id = epoch_store.epoch_start_config().epoch_data().epoch_id();
+        let epoch_start_timestamp = epoch_store
+            .epoch_start_config()
+            .epoch_data()
+            .epoch_start_timestamp();
+        let backing_store = self.get_backing_store();
+        let (kind, signer, gas_data) = tx_data.execution_parts();
+        let tx_digest = *transaction.digest();
+
+        // Step 6: execute.
+        let effects = if move_authenticators.is_empty() {
+            // Common path: no Move authentication, execute directly.
+            let (_, _, effects, _) = epoch_store.executor().execute_transaction_to_effects(
+                backing_store.as_ref(),
+                protocol_config,
+                self.metrics.limits_metrics.clone(),
+                false, // expensive_checks
+                self.config.certificate_deny_config.certificate_deny_set(),
+                &epoch_id,
+                epoch_start_timestamp,
+                tx_checked_input_objects,
+                gas_data,
+                gas_status,
+                kind,
+                signer,
+                tx_digest,
+                &mut None,
+            );
+            effects
+        } else {
+            // MoveAuthenticator path: get AuthenticatorFunctionRefForExecution
+            // for each authenticator, then run auth + execution in one pass with an
+            // execution-mode gas status.
+            let tx_data_bytes =
+                bcs::to_bytes(tx_data).expect("TransactionData serialization cannot fail");
+
+            let per_authenticator_inputs_for_exec = move_authenticators
+                .iter()
+                .zip(per_authenticator_inputs)
+                .map(|(ma, (auth_input_objects, account_object))| {
+                    let (id, seq, digest) = ma.object_to_authenticate_components()?;
+                    let signer_addr = ma.address()?;
+                    let func_ref_for_exec =
+                        self.check_move_account(id, seq, digest, account_object, &signer_addr)?;
+                    Ok((auth_input_objects, func_ref_for_exec))
+                })
+                .collect::<IotaResult<Vec<_>>>()?;
+
+            let per_authenticator_input_objects_for_exec = per_authenticator_inputs_for_exec
+                .iter()
+                .map(|(objs, _)| objs.clone())
+                .collect();
+
+            let authenticator_gas_budget = protocol_config.max_auth_gas();
+            let (exec_gas_status, per_auth_checked, auth_and_tx_checked) =
+                iota_transaction_checks::check_transaction_and_move_authenticator_input(
+                    tx_data,
+                    tx_input_objects,
+                    per_authenticator_input_objects_for_exec,
+                    authenticator_gas_budget,
+                    protocol_config,
+                    reference_gas_price,
+                )?;
+
+            let authenticators_for_exec = move_authenticators
+                .into_iter()
+                .zip(per_authenticator_inputs_for_exec)
+                .zip(per_auth_checked)
+                .map(|((ma, (_, func_ref)), checked_inputs)| {
+                    (ma.to_owned(), func_ref, checked_inputs)
+                })
+                .collect::<Vec<_>>();
+
+            let (_, _, effects, _) = epoch_store
+                .executor()
+                .authenticate_then_execute_transaction_to_effects(
+                    backing_store.as_ref(),
+                    protocol_config,
+                    self.metrics.limits_metrics.clone(),
+                    false, // expensive_checks
+                    self.config.certificate_deny_config.certificate_deny_set(),
+                    &epoch_id,
+                    epoch_start_timestamp,
+                    gas_data,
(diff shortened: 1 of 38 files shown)


vekkiokonio:
`Attestation::Validator` currently imports `starfish_config::AuthorityIndex` (`u8`). However, elsewhere in the codebase `u32` is used for authority index. This inconsistency will be addressed once the ongoing `AuthorityIndex` refactor from validator score is complete. Tracked in https://github.com/iotaledger/iota-private/issues/404.

vekkiokonio:
`check_system_overload()` is called before `submit_single_tx` ends, but the inflight counter (`num_inflight_transactions`) is only incremented later. This means the entire attestation phase — including the full dry-run and Move authentication — runs outside the load-shedding window. Under load, a validator can accumulate many concurrent attestation dry-runs before any are counted as inflight. Without attestation this gap was cheap(er); with it, it is large enough to warrant a dedicated pre-consensus counter or an early increment at the top of `submit_single_tx`. Tracked in https://github.com/iotaledger/iota-private/issues/405.

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -271,12 +291,52 @@ impl ValidatorService {
             .soft_lock_table_size
             .set(soft_locks.lock_count() as i64);
 
+        // Build the consensus transaction.
+        let consensus_tx = if let Some(attestation_data) = attestation_data {
+            use iota_types::{
```
alexsporn: move all imports to the top of the file

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
```diff
@@ -1066,3 +1090,154 @@ async fn test_dropped_tx_does_not_acquire_locks() {
     assert!(user_tx_digests.contains(verified_tx3.digest()));
     assert!(user_tx_digests.contains(verified_tx4.digest()));
 }
+
+/// A `UserTransactionV2` whose `attestor_index` matches the block's
+/// `certificate_author_index` (both `0`) passes Check #3 and is treated the
```
alexsporn: define check 3 here

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -1,6 +1,10 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(test)]
+#[path = "../unit_tests/validator_v2_tests.rs"]
```
alexsporn: move this to the bottom of the file

crates/iota-core/src/consensus_validator.rs
```diff
@@ -120,6 +120,24 @@ impl IotaTxValidator {
                     user_tx_v1_count += 1;
                 }
 
+                ConsensusTransactionKind::UserTransactionV2(_a) => {
+                    // User signature verification is skipped: the attesting
```
alexsporn: I think this could be the correct place to check attestation signatures. Even if this PR is only handling the validator attestations that have no explicit signatures we should probably prepare the call-site.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1686,6 +1693,10 @@ impl ProtocolConfig {
     pub fn enable_white_flag_flow(&self) -> bool {
         self.feature_flags.enable_white_flag_flow
     }
+
+    pub fn enable_validator_attestation(&self) -> bool {
+        self.feature_flags.enable_white_flag_flow && self.feature_flags.enable_validator_attestation
```
alexsporn: the white_flag_flow should be asserted, see other checks above.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -485,6 +485,13 @@ struct FeatureFlags {
     // Conflicts are resolved deterministically post-consensus using persistent locks.
     #[serde(skip_serializing_if = "is_false")]
     enable_white_flag_flow: bool,
+
+    // If true, the block-proposing validator attests each received user transaction
+    // before sending it to consensus. The attestor certifies that Move authentication
+    // passed and the attestation carries a computation cost estimate for
+    // shared-object scheduling.
+    #[serde(skip_serializing_if = "is_false")]
+    enable_validator_attestation: bool,
```
alexsporn, line 494: bump the protocol version and set it to true for devnet/alphanet

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+/// The attested content carried by all [`Attestation`] variants.
+///
+/// Versioned to allow new fields to be introduced without breaking existing
+/// match arms. Both `Validator` and `Explicit` share the same `AttestationData`
+/// so any extension applies uniformly across attestation types.
+#[non_exhaustive]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum AttestationData {
+    V1 {
+        /// Expected computation cost, in MIST, used by the sequencer to improve
```
alexsporn: MIST...

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+        /// shared-object scheduling before execution.
+        estimated_computation_cost: u64,
+        /// Shared-object versions observed by the attestor during the dry-run.
+        /// Used to determine whether a discrepancy between the attested
+        /// estimate and the actual execution outcome is misbehavior or is
+        /// explained by a legitimate state change between attestation time and
+        /// execution time.
+        object_versions: Vec<ObjectRef>,
+    },
```
Thoralf-M, line 64: Don't we need a link (it's digest) to the actual transaction here? Otherwise in the `AttestedTransaction` one could just swap out the transaction with another one?

crates/iota-types/src/attestation.rs
Thoralf-M:
```suggestion
        /// Expected computation cost, in NANOS, used by the sequencer to improve
```

crates/iota-types/src/attestation.rs
alexsporn: Explicit attestations are signed, the signature would commit to this data together with the tx-digest

Thoralf-M: Ah right missed the comment

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+pub enum Attestation {
+    Validator {
+        payload: AttestationData,
+        /// Index of the attesting validator in the current epoch's committee
+        attestor_index: AuthorityIndex,
+    },
+    Explicit {
+        payload: AttestationData,
+        attestor_address: IotaAddress,
+        /// Signs over `hash(transaction.digest() || BCS(data) ||
```
Thoralf-M: Is the `data` part here `AttestationData`?
If so I think it would be good to have it written out, because `data` could be anything

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -150,7 +160,39 @@ pub async fn validate_and_resolve_conflicts(
...
+                        Some(IotaError::AttestationAuthorMismatch {
+                            expected: *attestor_index,
+                            actual: block_author,
+                        })
+                    } else {
+                        None
+                    }
+                }
+                // Reject Explicit variant as not yet implemented.
+                _ => Some(IotaError::ExplicitAttestationNotSupported),
```
bingyanglin: The `Explicit` attestation variant gets deserialized successfully but then rejected here at post-consensus. Is it late? The transaction already went through consensus. Maybe reject it earlier in `consensus_validator.rs` alongside the `UserTransactionV2` protocol check? That way we fail fast and don't waste consensus bandwidth on unsupported attestation types. Or do we have other concerns?

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +1020,205 @@ impl AuthorityState {
...
+                    signer,
+                    tx_digest,
+                    tx_data_bytes,
+                    &mut None,
+                );
+            effects
+        };
+
+        // Step 7: build AttestationData.
+        let estimated_computation_cost = effects.gas_cost_summary().computation_cost;
```
bingyanglin: This extracts `computation_cost` from effects without checking `effects.status()`. So if the transaction would abort (Move error, OOG), we still attest it and the cost reflects the abort. I guess that's intentional, so attestation certifies *validation* passed, not that execution succeeds? Just confirming that's the intended behavior?

crates/iota-core/src/authority_server/validator_v2.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/756ced81ffb1a937c99e34a8c53724ad845c36a0

crates/iota-core/src/authority_server/validator_v2.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/756ced81ffb1a937c99e34a8c53724ad845c36a0

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/0e2f31076275753296a330a4be610b702136c060

crates/iota-core/src/consensus_validator.rs
vekkiokonio: Yes, that's the plan. I prepared the call site now (https://github.com/iotaledger/iota/pull/11528/changes/faad188794d918c97bc6d38b1a6f1a37cf20c797)

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/0e2f31076275753296a330a4be610b702136c060

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/0e2f31076275753296a330a4be610b702136c060

crates/iota-types/src/attestation.rs
vekkiokonio: Good catch! It's `payload : AttestationData` indeed (not generic data). I had `data : AttestationData` before and missed to edit the comment

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: We shouldn't receive any explicit attestation right now as the logic is not yet implemented. However, I've edited `consensus_validator` as a placeholder for future attestor checks, and it now rejects explicit attestations early (https://github.com/iotaledger/iota/commit/faad188794d918c97bc6d38b1a6f1a37cf20c797)

crates/iota-core/src/authority.rs
vekkiokonio: That's a good point. The rationale here is that correctly setting gas is the issuer's responsibility. If the dry-run shows OOG, that's a user error rather than an invalid transaction. I'm also fine with changing this: the attestor could serve a dual purpose, both improving data flow for the sequencer and providing early feedback to the client by rejecting predictably-failing transactions before they reach consensus.

@alexsporn Thoughts?

alexsporn: I agree with the statements. It is a user-error and the transaction would successfully make it into a checkpoint, although as a failed tx, but if we can avoid the user loosing the gas, it might be better to abort the transaction altogether and return the user an error before we continue and pass this to consensus.
Same behavior then for the external attesters.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1686,6 +1694,15 @@ impl ProtocolConfig {
     pub fn enable_white_flag_flow(&self) -> bool {
         self.feature_flags.enable_white_flag_flow
     }
+
+    pub fn enable_validator_attestation(&self) -> bool {
+        let res = self.feature_flags.enable_validator_attestation;
```
cyberphysic4l, line 1705: I know it's used in some of the other cases above, but why the variable name `res`?

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+///   Requires a signature binding the attestation to the transaction.
+#[non_exhaustive]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Attestation {
+    Validator {
+        payload: AttestationData,
+        /// Index of the attesting validator in the current epoch's committee
+        attestor_index: AuthorityIndex,
+    },
+    Explicit {
```
cyberphysic4l, line 34: I would suggest renaming this variant to `ThirdParty` as suggested in the doc.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -588,6 +596,17 @@ impl ConsensusTransaction {
         }
     }
 
+    pub fn new_attested_transaction(attested: AttestedTransaction) -> Self {
```
cyberphysic4l:
```suggestion
    pub fn new_user_transaction_v2(attested: AttestedTransaction) -> Self {
```

crates/iota-core/src/authority.rs
cyberphysic4l: The advantage of letting OOG transaction pass is that it elimenates that as an attack vector on attestors i.e. sending transactions with insufficient gas to expend attestor resources without paying.

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +1020,205 @@ impl AuthorityState {
...
+    /// in a validator attestation and attached to the transaction before
+    /// submitting it to consensus.
+    ///
+    /// Subsumes `handle_transaction_validation_checks`: it runs the same
+    /// pipeline — deny checks, object loading, input validation, coin deny
+    /// list — and then replaces the final `authenticate_transaction` call with
+    /// `authenticate_then_execute_transaction_to_effects` so that Move
+    /// authentication and transaction execution run together and produce full
+    /// `TransactionEffects` including the accurate `computation_cost`.
+    pub(crate) fn attest_transaction(
```
cyberphysic4l: It is hard to distinguish what the exact overlap of all these checks is compared with `handle_transaction_validation_checks` even though it is outlined in the doc comment. It feels like it may be worthwhile creating a common helper containing the overlapping checks to ensure the checks remain identical in the event of changes to these checks in the future too.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -246,22 +288,31 @@ pub async fn validate_and_resolve_conflicts(
...
-            if e.is_storage_or_epoch_error() {
-                return Err(e);
+        //
+        // For `UserTransactionV2` (attested transactions), this check is skipped
+        // entirely. The attestor has already performed validation (including deny
+        // checks, gas, and ownership verification) before producing the
+        // attestation. Re-running these checks post-consensus would be redundant
+        // and could cause divergence if state changes between attestation time
+        // and execution time.
+        if attestation.is_none() {
```
cyberphysic4l, line 342: Should we really skip all the checks here for the attested transactions, or should we just skip the parts that will be re-run for execution later? As far as I understand, deny lists are not checked again, so by skipping this check, we are completely trusting the attestor with these checks.

crates/iota-core/src/consensus_handler.rs
```diff
@@ -508,6 +509,13 @@ pub(crate) fn classify(transaction: &ConsensusTransaction) -> &'static str {
                 "owned_user_transaction"
             }
         }
+        ConsensusTransactionKind::UserTransactionV2(a) => {
+            if a.transaction.contains_shared_object() {
+                "shared_user_transaction"
+            } else {
+                "owned_user_transaction"
+            }
+        }
```
cyberphysic4l, line 518:
```suggestion
        ConsensusTransactionKind::UserTransactionV2(a) => {
            if a.transaction.contains_shared_object() {
                "shared_user_transaction_v2"
            } else {
                "owned_user_transaction_v2"
            }
        }
```

crates/iota-types/src/messages_consensus.rs
```diff
@@ -248,6 +249,9 @@ pub enum ConsensusTransactionKind {
     /// directly to consensus without pre-consensus object locking.
     /// Conflicts are resolved post-consensus.
     UserTransactionV1(Box<Transaction>),
+    /// Attested user transaction. Carries a gas attestation produced either by
+    /// the proposing validator or a registered third-party attestor.
+    UserTransactionV2(Box<AttestedTransaction>),
```
piotrm50, line 254: I'm not sure this needs to be `Box`'ed. The transaction inside is already wrapped in a `Box`, so the enum size is small and limited regardless. Or perhaps the `Box` wrapping inside of `AttestedTransaction` is unnecessary?

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -271,12 +288,47 @@ impl ValidatorService {
             .soft_lock_table_size
             .set(soft_locks.lock_count() as i64);
 
+        // Build the consensus transaction.
+        let consensus_tx = if let Some(attestation_data) = attestation_data {
+            let committee = epoch_store.committee();
```
piotrm50: can use `committee.authority_index(&state.name)` directly.

piotrm50: also, since below we are using `Attestation::Validator` and we don't really do anything with the `attestor_index` value, is this even necessary? `submit_single_tx` only accepts Transactions anyway, so there is no way of to handle `ExplicitAttestation` yet. I guess we will eventually have to split this method into smaller chunks to handle ExplicitAttestations and Transactions and then this logic might be valid, but I would add it once there is an actual need for it.

crates/iota-core/src/consensus_validator.rs
```diff
@@ -120,6 +121,39 @@ impl IotaTxValidator {
                     user_tx_v1_count += 1;
                 }
 
+                ConsensusTransactionKind::UserTransactionV2(attested_tx) => {
+                    if !self
```
piotrm50, line 143: now that I think about it we should also make sure that UserTransactionV1 is not sent when `enable_validator_attestation` is enabled as well as add a similar check for `CertificateTransaction` when pcool is enabled.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -150,7 +160,39 @@ pub async fn validate_and_resolve_conflicts(
             continue;
         }
 
-        // Check #3: Extract owned input objects for lock conflict detection.
+        // Check #3: Attestor verification (UserTransactionV2 only).
+        // The block signature transitively authenticates the attestation;
+        // verify the claimed attestor matches the actual block author.
+        if let Some(attestation) = attestation {
+            let block_author =
+                starfish_config::AuthorityIndex::from(tx.0.certificate_author_index as u8);
```
piotrm50, line 175: `consensus_types::AuthorityIndex` is currently the `iota-core` native type for AuthorityIndex. We will unify this with the `starfish_config` type in the future, but for now I'd use the `iota-core` type for consistency. I see that this is also used in the Attestation::Validator transaction which is better to be u8 for the future. I also proposed in another comment that we can get rid of this field as it's redundant.

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+///   signature is needed.
+/// - [`Attestation::Explicit`]: produced by a registered third-party attestor.
+///   Requires a signature binding the attestation to the transaction.
+#[non_exhaustive]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Attestation {
+    Validator {
+        payload: AttestationData,
+        /// Index of the attesting validator in the current epoch's committee
+        attestor_index: AuthorityIndex,
```
piotrm50, line 32: Is this needed if we assume that the attestor is the one who included the transaction in a block?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -150,7 +160,39 @@ pub async fn validate_and_resolve_conflicts(
...
+            let error = match attestation {
+                Attestation::Validator { attestor_index, .. } => {
+                    if *attestor_index != block_author {
```
piotrm50, line 188: Do we need this check? Can't we assume implicitly that the block author is the attestor?

crates/iota-protocol-config/src/lib.rs
piotrm50: short for `result` I guess.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2770,6 +2787,13 @@ impl ProtocolConfig {
                     cfg.max_jwk_votes_per_validator_per_epoch = None;
                     cfg.max_age_of_jwk_in_epochs = None;
                 }
+                26 => {
```
piotrm50: I'd remove this until we actually know which version we want to release it as, otherwise we'll have to constantly rebase this.

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+        /// explained by a legitimate state change between attestation time and
+        /// execution time.
+        object_versions: Vec<ObjectRef>,
+    },
+}
+
+/// A user transaction bundled with its attestation. This is the inner payload
+/// of `ConsensusTransactionKind::UserTransactionV2`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AttestedTransaction {
```
piotrm50, line 70: this could perhaps be used as consensus type instead of UserTransactionV2? Otherwise I'd get rid of the Box<> around transaction as it doesn't bring much benefit.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: The deny checks will be checked again pre-execution, so that we're able to slash an attestor who attested wrongly. But I agree that the comment needs to be improved to make this clear as currently it feels like there is a gap in validation.

piotrm50 (approved):
Nicely scoped PR!

piotrm50 (commented):
nicely scoped PR. I left some comments.

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +1020,205 @@ impl AuthorityState {
...
+        let epoch_id = epoch_store.epoch_start_config().epoch_data().epoch_id();
+        let epoch_start_timestamp = epoch_store
+            .epoch_start_config()
+            .epoch_data()
+            .epoch_start_timestamp();
+        let backing_store = self.get_backing_store();
+        let (kind, signer, gas_data) = tx_data.execution_parts();
+        let tx_digest = *transaction.digest();
+
+        // Step 6: execute.
```
piotrm50: How compute-heavy is the execution? This whole thing runs inside of `spawn_monitored_task!` as an async task. If the transaction takes a long time to execute then we could starve Tokio workers under high load. Consider `spawn_blocking` task for the execution part. How is that done for actual tx execution later or dry run on some other endpoint?

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+/// a separate artifact that travels alongside the transaction; the transaction
+/// and the user's signature are completely unchanged.
+///
+/// Two variants are supported:
+/// - [`Attestation::Validator`]: produced by the block-proposing validator.
+///   Authenticated implicitly by the block signature — no separate attestor
+///   signature is needed.
+/// - [`Attestation::Explicit`]: produced by a registered third-party attestor.
+///   Requires a signature binding the attestation to the transaction.
+#[non_exhaustive]
```
piotrm50: Why is this non_exhaustive needed?

crates/iota-core/src/consensus_validator.rs
```diff
@@ -120,6 +121,39 @@ impl IotaTxValidator {
...
+                            // checked in post_consensus_validation.
+                        }
+                        Attestation::Explicit { .. } => {
+                            // TODO: verify explicit attestor signature against the trusted
+                            // attestor registry (Phase 2).
+                            return Err(IotaError::UnsupportedFeature {
+                                error: "Explicit attestation not yet supported".into(),
+                            });
+                        }
+                        _ => {
```
piotrm50: Why do we need this wildcard here? IMO it's better to not have it and then have a compilation error if we add a new variant in the future without having it covered in all places.

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+// TODO: change the import once the AuthorityIndex refactor is ready
```
roman1e2f5p8s, line 5:
```suggestion
// TODO: change the import once the AuthorityIndex refactor is ready
// See https://github.com/iotaledger/iota-private/issues/404
```
clarify what "the AuthorityIndex refactor" is

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+    signature::GenericSignature,
+    transaction::Transaction,
+};
+
+/// A pre-consensus claim produced by a trusted actor certifying that a specific
+/// transaction has been validated before entering consensus. The attestation is
+/// a separate artifact that travels alongside the transaction; the transaction
+/// and the user's signature are completely unchanged.
+///
+/// Two variants are supported:
```
roman1e2f5p8s, line 21: combining "Two variants are supported" with `#[non_exhaustive]` feels contradictory

```suggestion
/// The following variants are defined:
```

crates/iota-core/src/authority.rs
vekkiokonio: > The advantage of letting OOG transaction pass is that it elimenates that as an attack vector on attestors

That's actually a very good point. I think it's something we could reason more and potentially tackle on a different PR. I opened a related issue: https://github.com/iotaledger/iota-private/issues/410

crates/iota-types/src/attestation.rs
roman1e2f5p8s: same question

I would drop it, exhaustive matches would actually be safer here.
unless there is a plan to migrate this and other new enums to the `iota-rust-sdk` library crate?
there is also only one `#[non_exhaustive]` occurrence in the whole `iota-types` crate before this PR.

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+        signature: Box<GenericSignature>,
+    },
+}
+
+/// The attested content carried by all [`Attestation`] variants.
+///
+/// Versioned to allow new fields to be introduced without breaking existing
+/// match arms. Both `Validator` and `Explicit` share the same `AttestationData`
+/// so any extension applies uniformly across attestation types.
+#[non_exhaustive]
```
roman1e2f5p8s, line 49: I would drop it
https://github.com/iotaledger/iota/pull/11528#discussion_r3279868864

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Yes, just using the existing patterns/namings

crates/iota-types/src/attestation.rs
vekkiokonio: What about `External` attestation? So we avoid having two separate words (third + party)

crates/iota-types/src/messages_consensus.rs
vekkiokonio: If we call this `new_user_transaction_v2`, I would also change `new_user_transaction` to refer to `v1`. But I'm not sure it's needed tbh, I would perhaps prefer referring explicitly to attestation here

crates/iota-core/src/authority.rs
vekkiokonio: That's a good point! I had kept separate as there are relevant differences in the two implementation but it's probably worthwhile extracting the common checks. Now, I moved deny checks, input-object loading, input checks, and coin deny list into a single `run_validation_checks` helper called from both `handle_transaction_validation_checks` and `attest_transaction`. They now share steps 1–5 and only diverge at execution.

vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/75339c3604c1f9d943d638e54e516f1490a4a8d0

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+///
+/// Two variants are supported:
+/// - [`Attestation::Validator`]: produced by the block-proposing validator.
+///   Authenticated implicitly by the block signature — no separate attestor
+///   signature is needed.
+/// - [`Attestation::Explicit`]: produced by a registered third-party attestor.
+///   Requires a signature binding the attestation to the transaction.
+#[non_exhaustive]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Attestation {
```
roman1e2f5p8s, line 28: the variant names are highly asymmetric:
- `Validator` (broadly) describes "who" produced the attestation
- `Explicit` does not describe "who" but "how" (that is, explicitly) it is authenticated

the current names have other problems:
- if a validator is ever involved in any future variants, the current `Validator` variant becomes confusing
- any future "not validator" variant could literally fit into the `Explicit` category 

another problem: `payload` is repeated in every single variant, which indicates `struct` is a better model here.

`Attestation` in principle will carry only two pieces of data:
- payload - data being attested (already versioned, keep it)
- attestor - who attested that data (add a new smaller enum)

so my proposal:
```rust
pub struct Attestation {
    pub payload: AttestationData,
    pub attestor: Attestor,
}

pub enum Attestor {
    BlockSigner,
    SingleAttestor {
          address: IotaAddress,
          signature: Box<GenericSignature>,
    },
}
```

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -150,7 +160,39 @@ pub async fn validate_and_resolve_conflicts(
...
+                        Some(IotaError::AttestationAuthorMismatch {
+                            expected: *attestor_index,
+                            actual: block_author,
```
roman1e2f5p8s, line 191: `expected` should be `block_author` because it is the truth we expect to see

crates/iota-types/src/attestation.rs
roman1e2f5p8s, line 32: any justification why this is needed? couldn't this even lead to more exploitation?

crates/iota-core/src/post_consensus_validation.rs
roman1e2f5p8s, line 188: the only source of truth here is `tx.0.certificate_author_index`, derived from block signature.
why introduce a second source (`attestor_index`) - asserted by the author, not the consensus protocol - that can be wrong, just to compare it against the source of truth and error out on mismatch?

crates/iota-types/src/error.rs
```diff
@@ -530,6 +530,18 @@ pub enum IotaError {
...
     InvalidDkgMessageSize,
 
+    #[error(
+        "Attestation author mismatch: attestation claims attestor {expected:?} \
+         but the block was authored by {actual:?}"
+    )]
+    AttestationAuthorMismatch {
+        expected: starfish_config::AuthorityIndex,
+        actual: starfish_config::AuthorityIndex,
+    },
```
roman1e2f5p8s, line 540: we expect `block_author`, so it should be `expected`, not `actual`.
but the whole check that fires this error should be removed I think, in which case the error variant goes with it.

crates/iota-types/src/error.rs
```diff
@@ -530,6 +530,18 @@ pub enum IotaError {
...
+
+    #[error("Explicit attestation is not supported in the current protocol version")]
+    ExplicitAttestationNotSupported,
```
roman1e2f5p8s: I think this is not needed; we could use already present `UnsupportedFeature`

roman1e2f5p8s (commented):
PR description refers to the wrong issues

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: That's a good comment and it helped me dig deeper into these checks.
- `TransactionDenyConfig` is loaded from each validator's local `NodeConfig`, so it shouldn't run post-consensus at all (V1 included).
- Coin deny list v1 is on-chain, and I'll add the check at execution time.
- The other two checks (receiving-object validity and the Move bytecode verifier on publish) are fine to skip too as execution catches them naturally (perhaps @miker83z can confirm this :)

I'll also update the comment in `post_consensus_validation.rs` to reflect this.

crates/iota-types/src/messages_consensus.rs
```diff
@@ -248,6 +249,9 @@ pub enum ConsensusTransactionKind {
     /// directly to consensus without pre-consensus object locking.
     /// Conflicts are resolved post-consensus.
     UserTransactionV1(Box<Transaction>),
+    /// Attested user transaction. Carries a gas attestation produced either by
+    /// the proposing validator or a registered third-party attestor.
```
roman1e2f5p8s: it does not carry a gas attestation only

```suggestion
    /// Attested user transaction. Carries the transaction together with
    /// the attested data and the identity of the attestor that produced it.
```

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+        /// execution time.
+        object_versions: Vec<ObjectRef>,
+    },
+}
+
+/// A user transaction bundled with its attestation. This is the inner payload
+/// of `ConsensusTransactionKind::UserTransactionV2`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AttestedTransaction {
+    pub transaction: Box<Transaction>,
```
roman1e2f5p8s: I would remove this inner `Box`, keep the outer - the pattern is already there in the codebase to `Box` any variant carrying significant content:

```rust
  CertifiedTransaction(Box<CertifiedTransaction>),
  CheckpointSignature(Box<CheckpointSignatureMessage>),
  UserTransactionV1(Box<Transaction>),
  UserTransactionV2(Box<AttestedTransaction>),
```

crates/iota-types/src/messages_consensus.rs
roman1e2f5p8s: > Or perhaps the Box wrapping inside of AttestedTransaction is unnecessary?

this

crates/iota-core/src/post_consensus_validation.rs
roman1e2f5p8s, line 163: this applies to V2 as well, so the warning message should be updated

suggestion
```rust
            warn!(
                ?digest,
                kind = if attestation.is_some() { "UserTransactionV2" } else { "UserTransactionV1" },
                error = ?e,
                "user transaction failed validity_check post-consensus, dropping"
            );
```

crates/iota-core/src/consensus_validator.rs
```diff
@@ -404,7 +438,8 @@ mod tests {
                 | ConsensusTransactionKind::RandomnessDkgConfirmation(_, _) => None,
 
                 // Gated behind `enable_white_flag_flow`.
-                ConsensusTransactionKind::UserTransactionV1(_) => {
+                ConsensusTransactionKind::UserTransactionV1(_)
+                | ConsensusTransactionKind::UserTransactionV2(_) => {
```
roman1e2f5p8s: V2 is gated by `enable_validator_attestation`, which transitively requires `enable_white_flag_flow`, but it is not the same flag

crates/iota-types/src/messages_consensus.rs
roman1e2f5p8s: the pattern above is clear: `new_<snake_case_of_variant>()`:
- `new_user_transaction` drops conventional `_v1`
- `new_attested_transaction` does not match the variant name at all

so I agree with Andrew

rename both as

```rust
pub fn new_user_transaction_v1(transaction: Transaction) -> Self { … }
pub fn new_user_transaction_v2(attested_tx: AttestedTransaction) -> Self { … }
```

there is only one call site of `new_user_transaction`, so renaming in this PR is right and trivial

crates/iota-core/src/consensus_handler.rs
vekkiokonio: Good point, I renamed the UserTransactionV1 entries as well

crates/iota-types/src/messages_consensus.rs
```diff
@@ -638,6 +657,9 @@ impl ConsensusTransaction {
             ConsensusTransactionKind::UserTransactionV1(tx) => {
                 ConsensusTransactionKey::UserTransaction(*tx.digest())
             }
+            ConsensusTransactionKind::UserTransactionV2(attested) => {
+                ConsensusTransactionKey::UserTransaction(*attested.digest())
```
roman1e2f5p8s: `attested_tx` or just `tx` (would align with V1 above) reads better here

```suggestion
            ConsensusTransactionKind::UserTransactionV2(attested_tx) => {
                ConsensusTransactionKey::UserTransaction(*attested_tx.digest())
```

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+    pub fn digest(&self) -> &TransactionDigest {
+        self.transaction.digest()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use bcs;
+
```
roman1e2f5p8s: imported but does nothing

```suggestion
```

crates/iota-types/src/messages_consensus.rs
vekkiokonio: Good catch! I dropped the inner `Box<Transaction>` from `AttestedTransaction` in https://github.com/iotaledger/iota/pull/11528/changes/79dcbb728903528768196c84f627e8f8b0763c8b

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn attestation_data_bcs_round_trip() {
+        let data = make_attestation_data();
+        let encoded = bcs::to_bytes(&data).unwrap();
+        let decoded: AttestationData = bcs::from_bytes(&encoded).unwrap();
+        let AttestationData::V1 {
+            estimated_computation_cost,
+            object_versions,
+        } = decoded;
+        assert_eq!(estimated_computation_cost, 1_000_000);
+        assert_eq!(object_versions.len(), 1);
```
roman1e2f5p8s: why not derive `PartialEq` and maybe `Eq` on `Attestation`, `AttestationData`, `AttestedTransaction` and compare the whole data as `assert_eq!(decoded, data);` instead of checking field by field? that would be proper full coverage of bcs round trips.

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    #[test]
+    fn attestation_explicit_bcs_round_trip() {
+        let address = IotaAddress::random();
+        let attestation = Attestation::Explicit {
+            payload: make_attestation_data(),
+            attestor_address: address,
+            signature: Box::new(GenericSignature::Signature(
+                Ed25519IotaSignature::default().into(),
```
roman1e2f5p8s, line 143: all-zero bytes, worth using random-bytes signature instead

crates/iota-types/src/attestation.rs
```diff
@@ -0,0 +1,166 @@
...
+            attestor_address, ..
+        } = decoded
+        else {
+            panic!("unexpected variant");
+        };
+        assert_eq!(attestor_address, address);
+    }
+
+    #[test]
+    fn attested_transaction_bcs_round_trip() {
```
roman1e2f5p8s, line 152: this test does not verify the attestation bcs round-trips, only digest.
deriving `PartialEq` as suggested above would easily fix the test

crates/iota-core/src/authority_server/validator_v2.rs
vekkiokonio: We do need to keep `attestor_index` since the block-author info only exists on `SequencedConsensusTransaction`, which is the post-consensus wrapper. Once the tx reaches sequencer / execution, the attestor identity can only come from the attestation itself which can enable rewarding or penalizing the attestor.

crates/iota-transaction-checks/src/lib.rs
```diff
@@ -293,6 +293,50 @@ mod checked {
         ))
     }
 
+    /// Variant of [`check_certificate_and_move_authenticator_input`] for use
+    /// before consensus (attestation path), where a full
+    /// [`VerifiedExecutableTransaction`] is not yet available. Produces an
+    /// execution-mode gas status (full `transaction_gas_budget`, not the
+    /// reduced signing-time auth budget).
+    pub fn check_transaction_and_move_authenticator_input(
```
roman1e2f5p8s: the body of this function is a pure copy of the already existing `check_certificate_and_move_authenticator_input` above, why?

steps to refactor:
- `check_transaction_and_move_authenticator_input`:
  - keep the body
  - add inline comments - those present in the `check_certificate_and_move_authenticator_input` body
  - update the docstring because it will be invoked from `check_certificate_and_move_authenticator_input` too
  - maybe also annotate with `#[instrument(level = "trace", skip_all)]`
- make `check_certificate_and_move_authenticator_input` to call `check_transaction_and_move_authenticator_input` as follows

```rust
pub fn check_certificate_and_move_authenticator_input(
    cert: &VerifiedExecutableTransaction,
    tx_input_objects: InputObjects,
    per_authenticator_input_objects: Vec<InputObjects>,
    authenticator_gas_budget: u64,
    protocol_config: &ProtocolConfig,
    reference_gas_price: u64,
) -> IotaResult<(IotaGasStatus, Vec<CheckedInputObjects>, CheckedInputObjects)> {
    check_transaction_and_move_authenticator_input(
        cert.data().transaction_data(),
        tx_input_objects,
        per_authenticator_input_objects,
        authenticator_gas_budget,
        protocol_config,
        reference_gas_price,
    )
}
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2906,6 +2906,10 @@ impl AuthorityPerEpochStore {
                 // TODO: make sure that UserTransactionV1 blocks don't pass
                 //  validation if the protocol feature flag is not set
             }
+            SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::UserTransactionV2(_a),
```
```suggestion
                kind: ConsensusTransactionKind::UserTransactionV2(_attested_tx),
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2906,6 +2906,10 @@ impl AuthorityPerEpochStore {
                 // TODO: make sure that UserTransactionV1 blocks don't pass
```
roman1e2f5p8s, line 2916: there are a few places in this file where comments mention only V1 but the logic applies to V2 as well.
not critical, but misleading with V2 added - worth updating those comments

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -941,6 +941,16 @@ impl ConsensusOutputQuarantine {
                         .map(|obj| obj.object_id)
                         .collect::<Vec<_>>(),
                 ),
+                SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                    kind: ConsensusTransactionKind::UserTransactionV2(a),
+                    ..
+                }) => Some(
+                    a.transaction
```
```suggestion
                    kind: ConsensusTransactionKind::UserTransactionV2(attested_tx),
                    ..
                }) => Some(
                    attested_tx
                        .transaction
```

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -271,12 +288,47 @@ impl ValidatorService {
...
+            let attestor_index = committee
+                .names()
+                .position(|n| n == &state.name)
```
roman1e2f5p8s: this search will run for every single tx, why doing that if validator position does not change within an epoch? either cache authority index if `attestor_index` field is needed indeed, or remove the field entirely

crates/iota-core/src/consensus_validator.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/25e49fac2092762e5860ae6b90a8bd749725ae63

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: I'd keep it as it is right now and fix once the refactoring is complete. I had opened an [issue](https://github.com/iotaledger/iota-private/issues/404) for that.

crates/iota-types/src/attestation.rs
vekkiokonio: Yes, see my reply on a previous comment (https://github.com/iotaledger/iota/pull/11528#discussion_r3287562179)

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: A malicious validator may create invalid attestations with `attestor_index` of a different authority. Since after consensus the block author is lost, penalties will only be based on the attestor index field.

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -578,19 +579,22 @@ impl ConsensusAdapter {
...
-            }
+            let all_certificates = transactions
+                .iter()
+                .all(|tx| matches!(tx.kind, ConsensusTransactionKind::CertifiedTransaction(_)));
+            let all_v1 = transactions
+                .iter()
+                .all(|tx| matches!(tx.kind, ConsensusTransactionKind::UserTransactionV1(_)));
+            let all_v2 = transactions
+                .iter()
+                .all(|tx| matches!(tx.kind, ConsensusTransactionKind::UserTransactionV2(_)));
```
roman1e2f5p8s: could be done in a single scan rather than three
also could catch future user transaction V3, etc

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -848,11 +852,12 @@ impl ConsensusAdapter {
 
         let is_user_tx = is_soft_bundle
             || if epoch_store.protocol_config().enable_white_flag_flow() {
-                // In the certificate-less mode, `UserTransactionV1` kind corresponds
-                // to user transactions.
+                // In the certificate-less mode, `UserTransactionV1`/`UserTransactionV2`
```
roman1e2f5p8s, line 855: update other such comments in this file

crates/iota-core/src/consensus_handler.rs
```diff
@@ -704,6 +721,9 @@ impl SequencedConsensusTransaction {
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
                 kind: ConsensusTransactionKind::UserTransactionV1(_),
                 ..
+            }) | SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                kind: ConsensusTransactionKind::UserTransactionV2(_),
+                ..
             })
```
roman1e2f5p8s, line 725:
```suggestion
            SequencedConsensusTransactionKind::External(ConsensusTransaction {
                kind: ConsensusTransactionKind::UserTransactionV1(_)
                    | ConsensusTransactionKind::UserTransactionV2(_),
                ..
            })
```

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/da9346776b8048d97486fd410008a6e1f35b3455

crates/iota-types/src/attestation.rs
vekkiokonio: I removed the Box<> around transaction

crates/iota-core/src/authority.rs
vekkiokonio: Good point! Execution may be heavy, I've changed to `spawn_blocking` in https://github.com/iotaledger/iota/pull/11528/changes/7e6a4f2860350b0411c22d7222f2fb9e41f7bf4e

crates/iota-types/src/attestation.rs
vekkiokonio: Good point, I'll remove it. There's no plan to extend the attestor types

crates/iota-core/src/consensus_validator.rs
vekkiokonio: Fixed

crates/iota-types/src/attestation.rs
vekkiokonio: I removed the `#[non-exhaustive]` so I think we can keep the original comment

crates/iota-core/src/consensus_validator.rs
roman1e2f5p8s, line 552: V2 should be exercised here as well?

crates/iota-core/src/post_consensus_validation.rs
roman1e2f5p8s: > A malicious validator may create invalid attestations with `attestor_index` of a different authority.

but why is that even possible in the first place? isn't it because the field exists?

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s, line 4141: this is where `certificate_author_index` could be remembered and propagated later in this function to execution phase, so `attestor_index` seems redundant (also exploitable) in block-signer attestation

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -4391,11 +4395,25 @@ impl AuthorityPerEpochStore {
...
             }) => {
+                let transaction: &Transaction = match &transaction {
+                    SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                        kind: ConsensusTransactionKind::UserTransactionV1(t),
+                        ..
+                    }) => t,
+                    SequencedConsensusTransactionKind::External(ConsensusTransaction {
+                        kind: ConsensusTransactionKind::UserTransactionV2(a),
+                        ..
+                    }) => &a.transaction,
```
roman1e2f5p8s: so attestation data is discarded here? propagation to execution will be a next step?

crates/iota-core/src/authority/authority_per_epoch_store.rs
roman1e2f5p8s: in this PR, the attestation data is dropped, but in the future, it will have to be propagated to execution anyway.
this is exactly where `certificate_author_index` could be propagated to execution, so no `attestor_index` field is needed at all

crates/iota-core/src/post_consensus_validation.rs
roman1e2f5p8s: > Since after consensus the block author is lost

not lost, but explicitly discarded; see https://github.com/iotaledger/iota/pull/11528/changes#r3288664436 and https://github.com/iotaledger/iota/pull/11528/changes#r3288710190

the same mechanism that will propagate attestation data to execution could naturally carry `certificate_author_index` or block author along it

roman1e2f5p8s (commented):
formatting and clippy warning should be fixed

crates/iota-types/src/attestation.rs
vekkiokonio: I don't understand the comment (what to drop?). Also, what's the link about?

roman1e2f5p8s: `#[non_exhaustive]`. updated link

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: I agree - we could use the same mechanism that we use to propagate the attestation but... why if we can directly wrap the attestor index into the attestation itself? Why would this be exploitable?

crates/iota-types/src/messages_consensus.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/356e5e6b53de24672bc7efa8b9930abccbc1bb40

crates/iota-types/src/attestation.rs
vekkiokonio: I actually like this refactoring. Since this will also affect https://github.com/iotaledger/iota/pull/11574 (which is based on the Validator/Explicit dichotomy), I'd address it in a separate issue/PR (see https://github.com/iotaledger/iota-private/issues/411).

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: I'd move the discussion around the `attestor_index` to a separate attestation refactoring issue (https://github.com/iotaledger/iota-private/issues/411)

crates/iota-types/src/attestation.rs
vekkiokonio: I'd move the discussion around attestation refactoring in a [separate issue](https://github.com/iotaledger/iota-private/issues/411)

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: Phase 3 (rewards, slashing) will need attestor identity at multiple points downstream of consensus. With the index in the attestation, a single propagation path delivers everything; without it, we'd have to plumb `certificate_author_index` through a parallel channel and keep the two in sync.

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: This highlights the misalignment: in my vision, the `attestor_index` field is the source of truth, not the block author! The attestation is a self-contained data structure declaring _who_ attested the transaction and _how expensive_ it is.

crates/iota-types/src/error.rs
vekkiokonio: See my [previous comment](https://github.com/iotaledger/iota/pull/11528#discussion_r3302437215)

crates/iota-types/src/error.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/3a518a8022d36d414e7f712b6dd62b4399d63ee2

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/ce36fd4d212935bde9230a56e6e22f7d2b718528

crates/iota-core/src/consensus_validator.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/25e49fac2092762e5860ae6b90a8bd749725ae63

crates/iota-types/src/attestation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/baa08dabf45cceb6ac344e34542d8319f502b87f

crates/iota-types/src/attestation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/baa08dabf45cceb6ac344e34542d8319f502b87f

crates/iota-types/src/attestation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/baa08dabf45cceb6ac344e34542d8319f502b87f

crates/iota-transaction-checks/src/lib.rs
vekkiokonio: Good catch! Fixed in https://github.com/iotaledger/iota/pull/11528/changes/c73008d8471df8d44f8387e650154cdd0667c23f

crates/iota-core/src/authority_server/validator_v2.rs
vekkiokonio: This search is now a hashmap lookup so not worth caching

crates/iota-core/src/authority/authority_per_epoch_store.rs
vekkiokonio: Fixed

crates/iota-core/src/consensus_adapter.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/ae50c3068707a2ac9b7dc29aa41214b06cb8a791

crates/iota-core/src/consensus_adapter.rs
vekkiokonio: Fixed

crates/iota-core/src/consensus_validator.rs
vekkiokonio: Good point, I added in https://github.com/iotaledger/iota/pull/11528/changes/f29c8a120b6e5130ce0d596fd1441d7cc2fe3415

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +985,171 @@ impl AuthorityState {
...
+                .collect::<IotaResult<Vec<_>>>()?;
+
+            let per_authenticator_input_objects_for_exec = per_authenticator_inputs_for_exec
+                .iter()
+                .map(|(objs, _)| objs.clone())
+                .collect();
+
+            let authenticator_gas_budget = protocol_config.max_auth_gas();
+            let (exec_gas_status, per_auth_checked, auth_and_tx_checked) =
+                iota_transaction_checks::check_transaction_and_move_authenticator_input(
```
miker83z: Here you are executing the same checks for the second time (with a small change that I will describe later):
- `iota_transaction_checks::check_transaction_and_move_authenticator_input` here is equivalent to the check perfomed in `self.run_validation_checks -> self.check_transaction_inputs_for_validation -> iota_transaction_checks::check_move_authenticator_input_for_validation + iota_transaction_checks::check_transaction_input`
- the only change is that `iota_transaction_checks::check_transaction_and_move_authenticator_input` calls `check_transaction_input_inner` with the last parameter (`is_execute_transaction_to_effects`) set to true, while `iota_transaction_checks::check_transaction_input` calls the same function but with that param set to false.
- I suggest to remove this `iota_transaction_checks::check_transaction_and_move_authenticator_input` invocation here and modify `self.run_validation_checks` to support a new parameter, i.e., `is_execute_transaction_to_effects`, that is set to true for `attest_transaction` and false for `handle_transaction_validation_checks`, then propagate this parameter down to `iota_transaction_checks::check_transaction_input`

crates/iota-transaction-checks/src/lib.rs
miker83z: `check_transaction_and_move_authenticator_input` can be removed. See this https://github.com/iotaledger/iota/pull/11528#discussion_r3311401705 comment

crates/iota-core/src/authority.rs
miker83z: > The advantage of letting OOG transaction pass is that it elimenates that as an attack vector on attestors i.e. sending transactions with insufficient gas to expend attestor resources without paying.

This is not applicable for AA TXs: 
- I use the gas coin of an AA account, 
- make the authenticator go OOG (it will abort the execution, but charge estimated_computation_cost) 
- and then repeat this to drain any gas coin that I don't own

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +985,171 @@ impl AuthorityState {
...
+                    tx_data_bytes,
+                    &mut None,
+                );
+            effects
+        };
+
+        // Build AttestationData.
+        let estimated_computation_cost = effects.gas_cost_summary().computation_cost;
+
+        // Collect all input object refs seen during execution. Start from the
```
miker83z: Why are you collecting object refs with versions being the pre-execution ones? You will find the same info already in the TxData that is signed (through its digest) together with the attestation data. So it seems a duplication to me. The objects that will appear here and not in the TxData would only be the dynamic fields used during the execution, but I don't see why thisdynamic fields' information might be necessary.

crates/iota-core/src/post_consensus_validation.rs
miker83z: I confirm that the Move bytecode verifier runs check at publish time, instead, the receiving object validity is checked at 2 different time:
- with the current certificate path, when a certificate is created a received object is locked with a dedicated MarkerValue::Received in the DB;  this is currently done in TransactionInputLoader with read_receiving_objects_for_signing; after the changes the same process and must happen sometime in pre-consensus. 
- post-consensus, during the execution of the TX, a received object validation is done by a native function that checks that MarkerValue::Received in the DB, so it has a dependency on the previous step.

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -220,17 +220,42 @@ impl ValidatorService {
...
+            if epoch_store.protocol_config().enable_validator_attestation() {
+                let state_for_attest = state.clone();
+                let epoch_store_for_attest = epoch_store.clone();
+                let (result, verified_tx) = tokio::task::spawn_blocking(move || {
+                    let result =
+                        state_for_attest.attest_transaction(&verified_tx, &epoch_store_for_attest);
+                    (result, verified_tx)
+                })
+                .await
+                .expect("attest_transaction blocking task panicked");
```
piotrm50:
```suggestion
                .expect("attest_transaction should return Err for invalid transactions instead of panicking");
```

piotrm50 (approved):

crates/iota-core/src/authority.rs
vekkiokonio: That's a very good catch! I fixed in https://github.com/iotaledger/iota/pull/11528/changes/8f7746183d0442f66d375346fb1d4632b9b0dd6c

crates/iota-core/src/authority.rs
vekkiokonio: You're right, owned, gas, and receiving objects can be dropped since they're already in the `TransactionData`. What remains is shared objects, Move-authenticator + deny-list objects, and dynamic-field/child objects, which I believe are needed for detecting malicious attestors. I may drop more of these once the detection mechanism is implemented and it's clearer which aren't needed.

BTW, I also changed how these are collected, they now come from the execution's inner temporary store. Fixed in https://github.com/iotaledger/iota/pull/11528/changes/9c90ec5728f8aaf9f94dbf7c7391d3e9f155d892

crates/iota-core/src/authority.rs
```diff
@@ -1020,6 +983,179 @@ impl AuthorityState {
...
+            let authenticators_for_exec = move_authenticators
+                .into_iter()
+                .zip(funcs_for_exec)
+                .zip(per_authenticator_checked_inputs)
+                .map(|((ma, func_ref), (checked_inputs, _))| {
+                    (ma.to_owned(), func_ref, checked_inputs)
+                })
+                .collect::<Vec<_>>();
+
+            let (inner_temp_store, _, effects, _) = epoch_store
```
Thoralf-M: The dry-run's authentication result is discarded here, and `effects.status()` is never checked.`authenticate_then_execute_transaction_to_effects` folds a failed Move authentication into failed effects (its doc: "charges gas for the failed execution of the authentication and produces transaction effects"), so an invalid-auth AA transaction gets attested, sequenced, and executed — charging the abstract account's gas without its authenticator's consent. The legacy path rejects with `MoveAuthenticatorExecutionFailure` pre-consensus; this path should likewise reject when the authentication result is `Err`. Note this doesn't need a Byzantine attestor: any external user can drain an AA's gas this way. Table 2 in https://github.com/iotaledger/iota-private/issues/383#issuecomment-4632823982 defers auth re-run to execution, but the execution re-run charges gas rather than dropping — that case isn't covered by Table 3.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -246,22 +291,40 @@ pub async fn validate_and_resolve_conflicts(
...
+        //   - Coin deny list v1 will be enforced at execution time.
+        //   - Receiving-object validity: the Move runtime fails the `receive()` call
+        //     when the ref doesn't match current state.
+        //   - Move bytecode verifier on publish: the Move VM re-verifies every newly
+        //     published package; the signing-time variant only adds a stricter meter as
+        //     a DoS gate.
+        //   - Gas, ownership, `MoveAuthenticator` execution: re-applied in the
+        //     execution pipeline (`check_certificate_input` and
+        //     `authenticate_then_execute_transaction_to_effects`).
+        if attestation.is_none() {
```
Thoralf-M, line 342: Skipping Check `#6` means V2 transactions acquire persistent owned-object locks with no ownership/version validation at all —`validity_check` is structural only. A Byzantine attestor can self-attest a tx naming a victim's owned object at its current ref: every honest validator locks it here,`prepare_certificate` later deterministically rejects the tx (so the object version never advances), and I can't find a path that releases the lock —`process_certificate` only releases the `tx_guard`. That freezes the victim's object. V1 is immune because Check `#6` runs before lock acquisition. Suggest either a cheap ownership/existence check (multi-get of the owned input refs) for V2 before acquiring locks, or deterministic lock release when execution-side validation rejects an attested tx. The skip-rationale comment covers execution-time re-checks but not lock acquisition — please address that explicitly either way.

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -220,17 +220,42 @@ impl ValidatorService {
...
+            if epoch_store.protocol_config().enable_validator_attestation() {
+                let state_for_attest = state.clone();
+                let epoch_store_for_attest = epoch_store.clone();
+                let (result, verified_tx) = tokio::task::spawn_blocking(move || {
+                    let result =
+                        state_for_attest.attest_transaction(&verified_tx, &epoch_store_for_attest);
+                    (result, verified_tx)
+                })
+                .await
+                .expect("attest_transaction should return Err for invalid transactions instead of panicking");
```
roman1e2f5p8s: the error message is wrong. it should be something like 

```suggestion
                .expect("dry-run attest_transaction task panicked or was cancelled");
```

but do we want to re-panic here if `attest_transaction` panics?
why not processing this `JoinError` and rejecting tx instead of panic here?

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -577,20 +578,16 @@ impl ConsensusAdapter {
...
+            // Soft-bundle batches must be homogeneous — every transaction must be
+            // of the same kind, because submit_and_wait_inner assumes a single
+            // transaction kind across the batch.
+            let first_kind = std::mem::discriminant(&transactions[0].kind);
+            fp_ensure!(
+                transactions
+                    .iter()
+                    .all(|tx| std::mem::discriminant(&tx.kind) == first_kind),
+                IotaError::InvalidTxKindInSoftBundle
+            );
```
roman1e2f5p8s, line 596:
```suggestion
            );
            
            // Only user transactions (white-flag) or certificates (certificate mode)
            // should be soft-bundled. Nothing currently produces batches of other
            // transaction kind, but if something starts to, reject any other kind
            // defensively rather then allowing a silent pass-through into consensus.
            fp_ensure!(
                transactions[0].kind.is_user_transaction()
                    || matches!(
                        transactions[0].kind,
                        ConsensusTransactionKind::CertifiedTransaction(_)
                    ),
                IotaError::InvalidTxKindInSoftBundle
            );

```

crates/iota-core/src/authority.rs
```diff
@@ -5512,6 +5651,90 @@ impl AuthorityState {
...
+        // for each authenticator.
+        let (gas_status, tx_checked_input_objects, per_authenticator_checked_inputs) = self
+            .check_transaction_inputs_for_validation(
+                protocol_config,
+                reference_gas_price,
+                tx_data,
+                tx_input_objects,
+                &tx_receiving_objects,
+                &move_authenticators,
+                per_authenticator_inputs.clone(),
```
roman1e2f5p8s: the clone is only needed if `is_execute_transaction_to_effects == true` (`attest_transaction` path)
otherwise, `handle_transaction_validation_checks` discards `per_authenticator_inputs` anyway

small perf: gate cloning based on `is_execute_transaction_to_effects` before, and move here the value

crates/iota-core/src/authority_server/validator_v2.rs
vekkiokonio: Yes, this has been addressed in the [follow-up PR](https://github.com/iotaledger/iota/pull/11574)

vekkiokonio: Now merged to `protocol-research/feat/transaction-attestation`  branch (current)

crates/iota-core/src/authority.rs
Thoralf-M: Check got added in https://github.com/iotaledger/iota/pull/11528/changes/74f479579a15c199690ee16670a75f44a7ca0f4c

vekkiokonio: This issue is part of a bigger discussion, see, e.g., https://github.com/iotaledger/iota/pull/11528#discussion_r3311500383. To be safe, for now I don't allow attesting transactions with a failing dry-run (e.g. invalid Move auth, OOG). This is an interim fix; it will be addressed more comprehensively in https://github.com/iotaledger/iota-private/issues/410.

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: The rough idea is to give a canceled execution status instead of dropping invalid transactions during execution. With this approach, locks will be released. This task is being treated on a separate [issue](https://github.com/iotaledger/iota-private/issues/438)

Thoralf-M (approved):

crates/iota-core/src/consensus_adapter.rs
vekkiokonio: I added this check as a part of the bundle-homogeneity check. Fixed in https://github.com/iotaledger/iota/pull/11528/changes/946fb07a6104a6d42643c9a2967f64ede738b04c

crates/iota-core/src/authority.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11528/changes/946fb07a6104a6d42643c9a2967f64ede738b04c

iota-execution/latest/iota-adapter/src/execution_engine.rs
```diff
@@ -447,33 +449,45 @@ mod checked {
             },
         );
 
+        // Capture whether authentication failed before the result is moved into the
+        // body execution.
```
miker83z, line 454: Please, add this comment:
```suggestion
        // TODO: enhance the way the authenticator error is propagated https://github.com/iotaledger/iota/issues/11986 
        // Capture whether authentication failed before the result is moved into the
        // body execution.
```

PR #10127, opened 2026-02-06
feat(consensus): Score integration to Starfish: dag state updates

oliviasaa:
## Intent (clean summary, separate from the polluted current diff)

> The current PR diff includes a large amount of unrelated upstream churn because the branch has not been rebased onto current `develop` after the Starfish migration / scorer refactor. The summary below extracts the actual scoring-related intent from Olivia's commits.

Make the Starfish `ScoringMetricsStore` durable across restarts and drive updates from DAG-state eviction. Three pillars:

### 1. Eviction-driven updates (Starfish)
- When `DagState::flush()` runs, compute scoring-metric deltas from blocks about to be evicted and persist them via the `ScoringMetricsStore`.
- Main commit: `e7e713a5` — `+748` in `crates/starfish/core/src/scoring_metrics_store.rs`, `+28` in `dag_state.rs`.

### 2. Recovery from storage on restart (Starfish)
- Initialize the in-memory `ScoringMetricsStore` from persisted storage at authority start, so accumulated metrics survive restart.
- Main commit: `b0c90991` — `+226` in `scoring_metrics_store.rs`, `+14` in `dag_state.rs`.

### 3. Persistent storage layer for metrics (Starfish)
- Add a scoring-metrics column to the starfish storage layer (mem + rocksdb) with versioned encoding so the on-disk format can evolve.
- Commits: `1eb6f6bc` (storage scaffolding +139 across `mem_store`/`rocksdb_store`/`mod`/`store_tests`), `29eadd87` (small follow-up).

### 4. Simtest + comment polish
- `a6c7be16` — wire the new dependency into starfish simtests.
- `9c6be669` — comments.

## Out-of-scope / will drop on rebase

These commits parallel the work in the **consensus** (Mysticeti) crate or in `iota-types` and are obsolete because:
- **Mysticeti is gone from `develop`** — all `consensus/core/*` changes (`2ea2f09a` versioned storage type, `791d4649` versioned update methods, `67707fc9` type relocation, `e1d5d574` `MisbehaviorsV1` alias) drop out on rebase.
- **`iota-types` / `iota-core/scorer.rs` refactor is already absorbed** by PR #10779 (scorer refactoring): commit `97c57229` ("new misbehavior count module") overlaps heavily with the new `MisbehaviorObservations` introduced there. `548ba3f5` (iterator tests) needs to be re-targeted at the post-refactor types.

## Tracking

- Issue: https://github.com/iotaledger/iota-private/issues/278
[...]

PR #11531, opened 2026-05-14
feat(starfish-core): implement misbehavior tracking pipeline in DagState

piotrm50:
# Description of change

Port the runtime misbehavior tracking logic onto the `MisbehaviorStore` types introduced in #10088.

- `MisbehaviorStore` tracks per-authority misbehavior in two buckets (`in_memory` + `persisted`).
- On flush: recompute the in-memory window, accumulate evicted counts, write the persisted snapshot to storage.
- On startup: restore persisted counts from RocksDB and recompute in-memory counts from cached block refs.
- Wire faulty block header detection into `MisbehaviorStore` from every peer receive site (subscriber main + bundle, header synchronizer including the own-last-header fetch, and both commit syncers).
- Source-aware classification: `Subscriber` keeps `UnexpectedAuthority` as Unprovable; commit-chain / fetch-shape errors stay Untracked and belong to a separate commit-sync metric.

Stacks on top of #10088. Supersedes #10127.

fixes https://github.com/iotaledger/iota-private/issues/278
fixes https://github.com/iotaledger/iota-private/issues/280
Part of https://github.com/iotaledger/iota-private/issues/277


crates/starfish/core/src/misbehavior_store.rs
```diff
@@ -0,0 +1,865 @@
...
+        self.gauges[idx].missing_proposals.add(missing as i64);
+        self.gauges[idx].equivocations.add(equiv as i64);
+    }
+
+    /// Stable read clone of one authority's counters.
+    fn snapshot(&self, idx: usize) -> MisbehaviorCountsV1 {
+        self.authorities[idx].lock().unwrap().clone()
+    }
+
+    fn reset(&self) {
```
filipdulic, line 371: does not reset the gauges, which breaks the always synced invariant

filipdulic (approved):

crates/starfish/core/src/misbehavior_store.rs
piotrm50: right, fixed

polinikita (approved):

alexsporn (approved):

PR #11537, opened 2026-05-15
feat(docs): group framework functions by visibility

### LINKED ISSUE #6261: Generated Move documentation improvements
We are currently hosting the generated Move documentation for the system packages on [docs.iota.org](https://docs.iota.org/references/iota-move).

An improvement for usability could be to adapt `move-docgen` to properly handle function visibility.
For example right now `public(package)` functions are showed mixed together with other public and entry functions, which make it difficult for developers to quickly understand the public API of the module.

I think it could be a nice UX improvement if we would group the functions together in groups like `Entry`, `Public` , `Internal` and make the generated specs nicer to use.
Functions taking only one struct argument (i.e. callable with the dot-syntax) could be grouped together with the structs for enhanced visibility.

A possibility also could be to create separate sub-pages for the different structs inside a module and a page holding all the global module functions.

luca-moser:
# Description of change

Restructures the generated Move framework reference docs (rendered at docs.iota.org/references/iota-move) so the public surface of each module is easier to scan.

image

## What changed

**`move-docgen` (`external-crates/move/crates/move-docgen/src/docgen.rs`)**

- Functions are grouped by visibility into Public / Entry / Public(package) / Public(friend) / Private buckets, alphabetized within each bucket. Only the legacy `public(friend)` bucket stays in a collapsed `<details>`.
- Structs and enums are wrapped in top-level `Structs` and `Enums` sections so the page has clear categories alongside `Constants` and the function buckets.
- Dot-syntax methods (functions whose first parameter is an in-module struct or enum) are emitted as headings directly under their datatype, alphabetized. Free functions stay in the visibility buckets.
- The redundant `Function` / `Macro function` heading prefix is dropped (the parent bucket conveys the kind); macros are marked `### \`name\` (macro)`.
- New `DocgenFlags::include_module_toc` (default `true`) lets renderers that build their own heading sidebar (Docusaurus) skip the duplicate in-page TOC.

**`move-package`** — `BuildConfig` gains a `docgen_flags: DocgenFlags` field, replacing a long-standing `TODO this should be configurable` and letting the framework build opt out of the in-page TOC.

**`iota-framework` build (`crates/iota-framework/tests/build-system-packages.rs`)** — sets `docgen_flags.include_module_toc = false` when building the system packages, and adds `toc_max_heading_level: 4` to each module's frontmatter so per-struct method headings surface in Docusaurus's right-nav.

**`iota-genesis-builder`** — adds the new `docgen_flags` field to its exhaustive `MoveBuildConfig` construction (irrelevant in that path; `generate_docs: false`).

**Tests** — `move-docgen-tests` snapshots and the `move-stdlib` / `iota-framework` generated docs are refreshed.

## Motivation

Addresses [#6261](https://github.com/iotaledger/iota/issues/6261): today `public(package)` helpers, `entry` admin functions, and the public consumer API are rendered as one flat list, making it hard to tell a module's actual API surface (`iota::coin`, `iota_system`) from its internals. Grouping by visibility and surfacing dot-syntax methods next to their struct directly mirrors what Alex asked for.

Specifically:

- `cargo test -p move-docgen-tests` — all 12 snapshots pass.
- `cargo clippy -p move-docgen -p iota-framework --all-targets -- -D warnings` — clean.
- `UPDATE=1 cargo test -p iota-framework --test build-system-packages` and `UPDATE=1 cargo test -p move-stdlib check_that_docs_are_updated` — regenerates `docs/generated-docs/framework/` and the in-tree `move-stdlib/docs/std/*.md`; inspected `iota/coin.mdx`, `iota_system/iota_system.mdx`, and `std/vector.md` confirm the new layout (visibility buckets, alphabetized, per-struct method sections).
- Local Docusaurus dev server — regenerated framework pages render without MDX errors and the right-hand TOC shows `Structs` / `Enums` / `Constants` / `Public Functions` / … as top-level categories with method names nested under each struct.

No protocol, node, indexer, RPC, GraphQL, CLI, SDK, or gRPC behaviour changes — purely affects the build-time documentation generator and the markdown it emits, so no Release Notes section.


external-crates/move/crates/move-docgen-tests/tests/move/annotation/a__m.md@collapsed_sections.snap
```diff
@@ -8,6 +8,8 @@ info:
   no_collapsed_sections: true
   include_dep_diagrams: false
   include_call_diagrams: false
+  include_module_toc: true
+snapshot_kind: text
```
alexsporn: to get rid of these `snapshot_kind: text` changes you should update your cargo-insta with `cargo install cargo-insta --force`

lzpap (approved):
tried it out locally, worked as expected.

Question: didn't we have a vercel test deployment before on every pr that changed docs?

alexsporn:
> Question: didn't we have a vercel test depoyment before on every pr that changed docs?

Yes, but needs a change in the docs folder to trigger: https://github.com/iotaledger/iota/blob/11940e255042580f855ca106c067777dfb7a6e4d/.github/workflows/preview_wiki.yml#L16

alexsporn:
I ran the wiki-preview action manually from this branch but it seems to not re-generate the files:
https://iota-docs-6z6kruesh-iota1.vercel.app

alexsporn:
Got the docs to work locally. I would remove this blue dropdown that hides these functions by default, since they are visible in the index to the right anyway:
screenshot_2026-05-18_at_09 17 28_720

alexsporn (approved):
Very nice improvement 🎉

alexsporn (approved):

miker83z (approved):
LGTM!

Issue #11500, opened 2026-05-11
[indexer,graphql]: test backward-history based consistency model


kodemartin:
## Description

We should test the new feature in a production-like environment.

Draft the test plan.

tomxey:
1. Migration test:
 - we run the migrations that create the `checkpointed_objects` table on some network that has already been running before
 - we run a second indexer instance that has `checkpointed_objects` from the start and sync it from genesis
 - compare the to DBs if they have exactly the same `checkpointed_objects` at cp N
 - check if available range of the migration indexer is set properly (it should not have history from genesis)
2. Query parity test
 - run 2 indexers (or one indexer, but 2 readers), one using backward logic and one using forwards logic
 - fill the network with some transactions that create/modify/delete/recreate DFs and also do some other kind of ops like object transfers or move calls, so we have some data to query on
 - issue different kinds of queries using forward and backward graphql readers, assert results are the same
 - try to cover different query types, and also pagination
3. Pruning
 - sync the indexer with pruning enabled
 - check if data is removed, watermarks are updated and if available range reported by graphql is adjusted accordingly
4. Pruning after migration
 - after migration pruning watermark for backward history is set by the migration (doesn't start from 0)
 - we need to check if pruning can resume from such watermark properly and that no rows are missed (potential off by one errors)
5. Migration on big DB
 - check if migration takes reasonable amount of time on big DB like the staging one

tomxey:
# Local test results (tests 1-4)

Tests run against a local `iota-localnet` docker stack (single validator + access fullnode, 120 s/epoch). Scripts and full results live in [`consistent_views_qa_scripts/`](https://github.com/iotaledger/iota/tree/sc-platform/consistent-views-qa/consistent_views_qa_scripts) (branch `sc-platform/consistent-views-qa`); see `localnet_results.md` there for raw numbers.

For tests 1 and 2 we first run `qa-workload`, a small Rust binary that publishes a Move package and creates some test data covering: DFs / DOFs, add→remove→add lifecycles, transfers, wrap/unwrap, wrap+delete tombstones, plain deletions, and `borrow_mut` quirks. Each test then drives the indexers from genesis to the `final_checkpoint` (~1246 cps after workload, to contain all data generated by `qa-workload`).

### 1. Migration parity ✅
OLD-code indexer and NEW-code indexer sync the same cp range into two DBs (`db_a`, `db_b`). NEW migration is then applied on `db_a` and `checkpointed_objects` is diffed.
[...]

PR #11539, opened 2026-05-15
feat(graphql): dedupe DF ids before correlating MAX subquery

### LINKED ISSUE #11535: [indexer, graphql]: dynamic_fields::query Source B re-evaluates MAX subquery per backward-history row
# `Owner.dynamicFields` (paginated, backward-history path) is 3-4 s on staging

## Problem

The paginated DF listing query — `owner(address, rootVersion=V).dynamicFields(...)` or `object(address, version=V).dynamicFields(...)` — takes **3-4 seconds per call** on staging when the parent has any meaningful backward-history activity. Background and full perf table: [issue #11500 comment](https://github.com/iotaledger/iota/issues/11500#issuecomment-4458292943).

Test parameters: parent = `0x405584e5568824afea966c0bed59e65f571d013cc8c3c5e4d616d173717d51b3` (wrapped inner Versioned of Random on staging), 1 distinct DF object_id (the `RandomInner` field), ~225 k historical versions in `objects_backward_history`. Sweep of K=0…200 000 versions back: p50 latency flat at 3.6-3.9 s — no degradation with K, the *baseline* is the problem.

## Root cause — Source B of `dynamic_fields::query`

[`backward_view/dynamic_fields.rs:129-150`](https://github.com/iotaledger/iota/blob/infra/feat/backward-history-consistent-views/crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs#L129-L150):

```sql
SELECT <bh cols>
FROM objects_backward_history
WHERE owner_id = $parent
  AND owner_type = 2
  AND df_kind IS NOT NULL
  AND object_version = (
    SELECT MAX(object_version)
    FROM objects_version ov
    WHERE ov.object_id = objects_backward_history.object_id     -- ← per outer row
      AND ov.object_version <= $parent_version
  );
```

`EXPLAIN (ANALYZE, BUFFERS)` on staging:

```
Index Scan using objects_backward_history_owner ...
   Index Cond: (owner_type = 2 AND owner_id = '\x405584e5…')
   Filter:    (df_kind IS NOT NULL AND object_version = (SubPlan 2))
   Rows Removed by Filter: 306 620
   Buffers: shared hit=1 492 479    (~12 GB of buffer pool traffic)
SubPlan 2 → Limit → Append over objects_version_00..ff
   loops=306 620                     -- ← the killer
   Buffers: shared hit=1 226 484
JIT: Functions: 779, Total 1 310 ms
Execution Time: 4 747 ms
```

The outer scan returns 306 620 rows matching `owner_id = parent AND owner_type = 2`. The parent owns only **1 distinct DF object_id**; those 306 k rows are the same DF repeated across its ~225 k historical versions + parent-side churn. The correlated `MAX(object_version)` subquery is then evaluated 306 620 times — once per outer row, instead of once per distinct DF id. Postgres can't lift the subquery on its own because of the correlation `ov.object_id = objects_backward_history.object_id`.

### 1. Rewrite the SQL to dedupe before correlating

Either form is equivalent (verified by EXPLAIN ANALYZE — identical plan + identical timing to ~2 ms noise):

**LATERAL:**

```sql
SELECT bh.*
FROM (
  SELECT DISTINCT object_id FROM objects_backward_history
  WHERE owner_id = $parent AND owner_type = 2 AND df_kind IS NOT NULL
) df_ids
CROSS JOIN LATERAL (
  SELECT MAX(object_version) AS target_v
  FROM objects_version
  WHERE object_id = df_ids.object_id AND object_version <= $parent_version
) v
JOIN objects_backward_history bh
  ON bh.object_id = df_ids.object_id AND bh.object_version = v.target_v;
```

**Correlated subquery in WHERE on the deduped derived table:**

```sql
SELECT bh.*
FROM (
  SELECT DISTINCT object_id FROM objects_backward_history
  WHERE owner_id = $parent AND owner_type = 2 AND df_kind IS NOT NULL
) df_ids
JOIN objects_backward_history bh ON bh.object_id = df_ids.object_id
WHERE bh.object_version = (
  SELECT MAX(object_version) FROM objects_version
  WHERE object_id = df_ids.object_id AND object_version <= $parent_version
);
```

The correlation is now against the deduped derived table (1 row in our case) instead of against the 306 k-row outer stream — so the subquery is evaluated once per *distinct DF*, not once per backward-history row.

### 2. Add a partial index on `objects_backward_history`

```sql
CREATE INDEX objects_backward_history_owner_df
  ON objects_backward_history (owner_id, object_id)
  WHERE owner_type = 2 AND df_kind IS NOT NULL;
```

The DF listing filter (`owner_id = $parent AND owner_type = 2 AND df_kind IS NOT NULL`) is shared by both DF query paths — `dynamic_fields::query` (parent-version pinned) and `consistent::query` + `apply_filter` (latest). The existing `objects_backward_history_owner` index `(superseded_at_checkpoint, owner_type, owner_id)` doesn't include `object_id`, so even after dedupe, Postgres must scan every matching row to enumerate distinct DF ids. The new partial index is index-only and densely packed → `SELECT DISTINCT object_id` becomes an Index Only Scan.

## Measured impact (staging)

| Variant | Execution time |
|---|---:|
| Original SQL, existing indexes | **4 747 ms** |
| Original SQL + new partial index | ~4 600 ms (no real change) |
| Rewritten SQL, existing indexes | 1 286 ms |
| Rewritten SQL + new partial index | 106 ms |
| Rewritten SQL + new partial index, post-VACUUM | **59 ms** |

Both parts are needed. Index alone doesn't help the existing code; rewrite alone leaves a Sort/Unique over 411 k rows. Together: **~80× faster** than today.

## Notes

- The same filter is used by `consistent::query` callers via `apply_filter` ([`types/dynamic_field.rs:294-303`](https://github.com/iotaledger/iota/blob/infra/feat/backward-history-consistent-views/crates/iota-graphql-rpc/src/types/dynamic_field.rs#L294-L303)) — both DF listing paths benefit from the new index. Need to also verify the rewrite pattern applies there.
- Single-DF-by-name lookups (`Owner.dynamicField(name)`, `Owner.dynamicObjectField(name)`) are unaffected — they derive the DF id and look up by `object_id` directly, hitting the existing `objects_backward_history_id_version` index.

tomxey:
# Description of change

Source B of `dynamic_fields::query` evaluated the correlated `MAX(object_version)` subquery per backward-history row (loops = N rows, not N distinct DFs), pushing latency to 3-4 s on staging for parents whose DFs have many historical versions.

Rewrite to dedupe DF ids first, then correlate. Add partial index `objects_backward_history_owner_df` so the dedupe is an Index Only Scan.

Worst case query time down from **4 747 ms** to **59 ms** on staging (#11500).

fixes #11535

- [x] Patch-specific tests (correctness, functionality coverage) - **built graphql rpc and tested performance on staging, speedup confirmed**

### DIFF
--- a/crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
+++ b/crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
@@ -126,24 +126,43 @@ fn dynamic_fields_from_checkpointed_objects(
 /// records every real version, including tombstone versions). When
 /// `target_v` lands on a tombstone version, no Active state row exists at
 /// that key — the candidate naturally drops out of the result.
+///
+/// The query is shaped so the correlated `MAX(object_version)` subquery
+/// is keyed off a **deduped** set of DF `object_id`s (one row per DF) —
+/// not the un-deduped owner scan of `objects_backward_history` (which for
+/// a parent with deep DF history can have hundreds of thousands of rows
+/// for the same DF). The dedupe inner query relies on the partial index
+/// `objects_backward_history_owner_df` on
+/// `(owner_id, object_id) WHERE owner_type = 2 AND df_kind IS NOT NULL`.
 fn dynamic_fields_from_historical_objects(
     parent_version: i64,
     page: &Page<Cursor>,
     filter_fn: &impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
-    let history_filtered = filter_fn(query!(format!(
-        "SELECT {HISTORY_COLUMNS} FROM objects_backward_history"
-    )));
+    // Distinct DF object_ids owned by the parent. `filter_fn` adds the
+    // `owner_id = $parent AND owner_type = 2 AND df_kind IS NOT NULL`
+    // predicate, which together with the partial index lets Postgres do
+    // an Index Only Scan to find these ids.
+    let df_ids = filter_fn(query!(
+        "SELECT DISTINCT object_id FROM objects_backward_history"
+    ));
 
-    let with_target = filter!(
-        history_filtered,
-        format!(
-            "object_version = (\
-                 SELECT MAX(object_version) FROM objects_version ov \
-                 WHERE ov.object_id = objects_backward_history.object_id \
-                   AND ov.object_version <= {parent_version})"
-        )
+    // For each distinct DF, fetch the backward-history row at target_v
+    // via the correlated MAX subquery.
+    let (df_ids_sql, df_ids_binds) = df_ids.finish();
+    let sql = format!(
+        "SELECT {HISTORY_COLUMNS} FROM ( \
+             SELECT objects_backward_history.* \
+             FROM ({df_ids_sql}) df_ids \
+             JOIN objects_backward_history \
+                 ON objects_backward_history.object_id = df_ids.object_id \
+             WHERE objects_backward_history.object_version = ( \
+                 SELECT MAX(object_version) FROM objects_version \
+                 WHERE object_id = df_ids.object_id \
+                   AND object_version <= {parent_version}) \
+         ) AS objects_backward_history"
     );
+    let with_target = RawQuery::new(sql, df_ids_binds);
 
     let source = query!("SELECT candidates.* FROM ({}) candidates", with_target);
     page.apply::<StoredBackwardObject>(source)
--- a/crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
+++ b/crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
@@ -127,13 +127,17 @@ fn dynamic_fields_from_checkpointed_objects(
 /// `target_v` lands on a tombstone version, no Active state row exists at
 /// that key — the candidate naturally drops out of the result.
 ///
-/// The query is shaped so the correlated `MAX(object_version)` subquery
-/// is keyed off a **deduped** set of DF `object_id`s (one row per DF) —
-/// not the un-deduped owner scan of `objects_backward_history` (which for
-/// a parent with deep DF history can have hundreds of thousands of rows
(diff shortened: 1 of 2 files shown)


sergiupopescu199 (approved):

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
```diff
@@ -126,24 +126,43 @@ fn dynamic_fields_from_checkpointed_objects(
...
 /// `target_v` lands on a tombstone version, no Active state row exists at
 /// that key — the candidate naturally drops out of the result.
+///
+/// The query is shaped so the correlated `MAX(object_version)` subquery
+/// is keyed off a **deduped** set of DF `object_id`s (one row per DF) —
+/// not the un-deduped owner scan of `objects_backward_history` (which for
+/// a parent with deep DF history can have hundreds of thousands of rows
+/// for the same DF). The dedupe inner query relies on the partial index
+/// `objects_backward_history_owner_df` on
+/// `(owner_id, object_id) WHERE owner_type = 2 AND df_kind IS NOT NULL`.
```
kodemartin: This is a bit hard to actually follow, as it includes ad-hoc terminology (e.g. "keyed off") and redundant information (e.g. deduped [...], then not the un-deduped).

Something along this path might have been more clear:  

> "The query is shaped so that DF `object_id`s from `objects_backwards_history` that match \<the predicate\> are first deduplicated. \<Reason for deduplication\>

kodemartin (approved):

crates/iota-graphql-rpc/src/backward_view/dynamic_fields.rs
tomxey: done in [f6f4470](https://github.com/iotaledger/iota/pull/11539/commits/f6f44709ff94efa19b6d38189d330e20869582f0)

PR #11263, opened 2026-04-20
feat(ci): run heavy tests only if PR is labeled with `ready`

muXxer:
# Description of change

To reduce the load on our CI runners, we will run heavy tests like `rust-tests`, `rust-simtests`, `split-cluster`, `split-cluster-sync-check`  or `e2e` (`turbo-hierarchy.yml`) only if a PR is marked with `ready` label.

## What changed                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                             
  - `hierarchy.yml`: `split-cluster` and `split-cluster-sync-check` jobs now require the "ready" label on PRs                                                                                                                                                                      
  - `turbo_hierarchy.yml`: `e2e` job now requires the "ready" label on PRs                                                                                                                                                                                                       
  - `_rust_tests.yml`: `rust-tests`, `rust-simtests` jobs now require the "ready" label on PRs
                                                                                                                                                                                                                                                                             
## How it works
  - PR opened/pushed → cheap jobs run normally, expensive jobs skip (no "ready" label)                                                                                                                                                                                       
  - Developer adds "ready" label → clicks "Re-run all jobs" → expensive jobs now run                                                                                                                                                                                         
  - Subsequent pushes with "ready" present → expensive jobs run automatically       
  - Nightly / push to main branches → everything runs, label check bypassed                                                                                                                                                                                                  
  - No trigger changes, no concurrency group changes, no interference with the labeler bot   

## Links to any relevant issues

#11253 

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #11540, opened 2026-05-15
chore: Running heavy tests in PR on demand by clicking checkbox in PR description

### LINKED ISSUE #11253: ci: only run heavy CI jobs if PR is labeled with "ready"
Currently every commit pushed to a PR triggers all tests, including the heavy ones like rust-tests, simtests, turbo hierarchy e2e tests. This cloggs the CI runners. We should only run the heavy tests after the PR is labeled with "ready".

chriscandless:
# Description of change

## Motivation

The repository runs both light CI (lint, format, typos, license, compile checks) and heavy CI (Rust tests, Rust simtests, split-cluster compatibility checks). Heavy tests are slow and consume the self-hosted runner pool heavily.

Heavy tests currently run automatically on every non-draft PR push. When contributors push several commits in a row, the runner pool gets saturated by heavy tests for work that isn't ready for that level of verification yet - starving other PRs and develop.

A previous proposal (#11263) gated heavy tests behind a ready label. In practice that approach turned out to be very hard to make robust: it accumulated a large number of edge cases - label added before vs. after leaving draft, draft↔ready toggles while heavy tests are mid-run, "remove the label to cancel", re-adding the label re-running everything from scratch, concurrency interactions, and so on - and the fixes for these kept overlapping and breaking one another, so a clean, predictable UX was very difficult to reach. On top of that, a label is persistent state: it is easy to leave on a stale PR, to forget to remove, or to accidentally re-trigger on every subsequent push.

So, instead of the label-based gate, this PR introduces an explicit "button" model: heavy tests on a PR run only when a contributor explicitly asks for them by ticking a checkbox in the PR description. The checkbox auto-unchecks immediately, so every run is a deliberate one-shot opt-in - there is no lingering state to forget. Behavior on long-lived branches (develop, devnet, testnet, mainnet, release branches) and nightly is unchanged: those still get full heavy coverage automatically.

Fixes #11253.

## How it works

1. PR templates contains a checkbox:

- [ ] Run heavy tests

1. The marker is invisible in the rendered description; the user only sees "Run heavy tests (…)".
2. ci_trigger.yml (generic dispatcher) runs on pull_request_target: [opened, edited]. When a box is ticked it:
  - finds every checked box matching the marker pattern;
  - unchecks them all in a single PR-body edit, before dispatching anything (this, plus a bot-sender filter, makes an infinite re-trigger loop impossible);
  - resolves each (workflow, inputs) against the ALLOWED_MARKER_DISPATCHES allowlist - workflows not registered there,  and input keys not listed for an allowed workflow, are dropped with a warning;
  - skips the dispatch if a run of that workflow is already active for the same HEAD SHA (don't restart in-flight progress when the user just re-ticked);
  - dispatches each surviving (workflow, inputs) via workflow_dispatch, passing pr_number, head_sha, and the allowlisted marker inputs.
3. heavy_tests.yml is the heavy-CI entry point. Triggers:
  - workflow_dispatch - from the checkbox dispatcher, or manually from the Actions UI with a head_sha,
  - push to long-lived branches - unchanged "full coverage on every merge" behavior.
It exposes a test_only_changed_crates input (default true) backing the two checkbox variants: changed crates only (fast) vs full workspace (comprehensive). Safety override: if workspace-level Rust config changed (root Cargo.toml/Cargo.lock), it always expands to the full workspace.
4. Concurrency redesign. The heavy reusable workflows previously each defined their own concurrency group derived from  github.event.pull_request.number. That field only exists on pull_request events; when these workflows are reached via workflow_dispatch it collapses to a constant, so every PR's heavy run collided into one global group and serialized cross-PR - defeating the 10-runner pool. A reusable workflow fundamentally cannot compute a correct concurrency key (it depends on the caller's trigger), so concurrency was moved to the entry workflows that own the trigger.

## Behavior

On a PR (the new part):
- PR opened / pushed → light CI (hierarchy.yml) runs as before (existing draft gate kept); heavy CI does not run.
- Contributor ticks a "Run heavy tests" checkbox → heavy_tests.yml is dispatched for the PR's current HEAD; the checkbox auto-unchecks immediately.
- If a heavy run is already active for this PR's HEAD SHA, ticking any "Run heavy tests" checkbox is a no-op and logs a notice — wait for that run to finish before re-triggering. Variant switching mid-run (e.g. "changed" → "full") is not supported on purpose: it would mean killing in-flight progress.
- Heavy runs targeting a different SHA (e.g. a new push) still supersede older in-flight runs via heavy_tests.yml's own concurrency, since testing the latest commit is always what we want.
- Two different PRs requesting heavy at the same time → run in parallel, bounded only by the runner pool (no cross-PR serialization).

Unchanged:
- Push to develop → heavy CI runs for every commit (queued, never cancelled).
- Push to devnet / testnet / mainnet / release branches → heavy CI runs, latest commit only.
- Nightly schedule → unchanged.

### New
- .github/workflows/ci_trigger.yml - generic checkbox→workflow dispatcher (marker parsing, batch uncheck, loop guard, extra-input passthrough).
- .github/workflows/heavy_tests.yml - heavy CI entry point; workflow_dispatch + push to long-lived branches; test_only_changed_crates changed/full modes with workspace-config safety override.

### Modified
- hierarchy.yml - removed split-cluster and split-cluster-sync-check (moved into heavy_tests.yml); trimmed the now-unused inputs passed to _rust.yml. Light jobs and their draft gate are unchanged.
- _rust.yml - removed the heavy chain (rust-tests sub-job calling _rust_tests.yml, and the crates-changes/external-changes filters that only fed it) and the inputs only it used; kept rust-lints + execution-cut.
- _rust_tests.yml, _split_cluster.yml, _split_cluster_sync_check.yml - removed their workflow-level concurrency blocks (now owned by the entry workflows; comment added explaining why a reusable can't own concurrency).
- nightly.yml - added a concurrency block (nightly-${{ github.ref }}, cancel-in-progress: false) to preserve the nightly de-dup it previously inherited from the reusables.
- .github/PULL_REQUEST_TEMPLATE/{default_internal_contributors,default_external_contributors,infra}.md - added the "## CI" section with the two checkboxes.

## Adding a new on-demand workflow later

1. Give the workflow a workflow_dispatch trigger with pr_number (string, optional) and head_sha (string, required) inputs (plus any extra inputs you want marker-driven), and check out ${{ inputs.head_sha || github.sha }}.
2. Register the workflow in ALLOWED_MARKER_DISPATCHES in ci_trigger.yml, listing which of its inputs may be set from a  marker. Workflows not in the map (and inputs not listed for an allowed workflow) are closed by default.
3. Add a checkbox line to the PR sub-template(s): - [ ] <label> .

## Security

ci_trigger.yml uses pull_request_target so it has the permissions needed to edit the PR body and dispatch even on fork  PRs. It never checks out or executes PR-author-controlled code. Two complementary layers limit what a malicious PR description can do via the marker:

* ALLOWED_MARKER_DISPATCHES is a closed-by-default allowlist of (workflow → marker-overridable inputs). A marker pointing at a workflow not in the map is rejected; input keys not listed for an allowed workflow are stripped. This blocks both "dispatch an arbitrary workflow_dispatch workflow on develop" (e.g. release.yml) and "smuggle dispatcher-controlled inputs like head_sha".
* Dispatched workflows are resolved from develop (ref: 'develop'), so a fork can't ship a modified workflow file via the PR — only the version on the default branch runs.

## ⚠️ Before merge

This branch contains a temporary dry-run commit for pre-merge testing (ci_trigger.yml: pull_request_target→pull_request, and createWorkflowDispatch→core.notice log, both flagged # >>> TEMP - REVERT BEFORE MERGE <<<). It exists because pull_request_target always uses the base-branch file and workflow_dispatch requires the workflow to exist on the default branch, so the full end-to-end flow can only be exercised after merge. Revert that commit before merging. The full cross-workflow dispatch + heavy_tests.yml wiring should be verified on a test PR once this is on develop.

## CI

Tick a box below to trigger the corresponding workflow on this PR's current HEAD.
Each box auto-unchecks once the run is dispatched — tick again to re-run.

- [ ] Run heavy tests

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,394 +1,153 @@
-# CI workflows: design and security guide
+# CI workflows: how to edit safely
 
-This directory contains GitHub Actions workflows for `iotaledger/iota`.
-**Read this guide before editing any workflow file.** Several of these
-workflows interact with GitHub's security model in subtle ways — small
-changes (a different trigger, a missing `persist-credentials: false`, an
-unpinned action) can introduce critical vulnerabilities or quietly break
-the intended invariants.
+**Read this before editing any workflow file here.** Some of these
+workflows run PR-author-controlled code with elevated privileges. A small
+change — a different trigger, a missing `persist-credentials: false`, an
+unpinned action — can turn a safe workflow into a remote-code-execution or
+secret-exfiltration hole, or silently make tests run against the wrong code.
 
-The guide is written for both humans and LLMs assisting with edits.
+Written for both humans and LLMs assisting with edits.
 
-## Table of contents
+## The rules (apply all of them)
 
-- [Why this matters](#why-this-matters)
-- [Trigger semantics](#trigger-semantics)
-- [Permissions](#permissions)
-- [Checkout settings](#checkout-settings)
-- [Action pinning](#action-pinning)
-- [Untrusted input handling](#untrusted-input-handling)
-- [Our architecture](#our-architecture)
-- [Dangerous changes — review carefully](#dangerous-changes--review-carefully)
-- [How to add a new on-demand workflow](#how-to-add-a-new-on-demand-workflow)
-- [References](#references)
-
-## Why this matters
-
-GitHub Actions has been the target of multiple high-profile attacks in
-2024–2026:
-
-- **tj-actions/changed-files** (Mar 2025, [CVE-2025-30066](https://github.com/advisories/ghsa-mrrh-fwg8-r2c3))
-  — attacker rewrote version tags to point at malicious commit;
-  ~23,000 repos exposed CI secrets in logs.
-- **Trivy / trivy-action** (Feb–Mar 2026) — `pull_request_target` workflow
-  exploited to exfiltrate PAT; 75 of 76 version tags force-pushed.
-- **Ultralytics YOLO** (Dec 2024) — branch-name injection into `run:`
-  under `pull_request_target` led to RCE with secrets access.
-- **PostHog (Shai Hulud v2)** — `pull_request_target` workflow exploited
-  to leak npm publishing token; compromised SDK packages published.
-- **Spotbugs**, **Gluestack** ([CVE-2025-53104](https://www.sysdig.com/blog/cve-2025-53104-command-injection-via-github-actions-workflow-in-gluestack-ui)),
-  **Langflow** ([GHSA-87cc-65ph-2j4w](https://github.com/langflow-ai/langflow/security/advisories/GHSA-87cc-65ph-2j4w))
-  — various injection patterns through `pull_request_target`.
-
-Most of these exploited patterns that look innocent in isolation but are
-dangerous in combination — especially `pull_request_target` plus
-checkout-of-PR-code, or unpinned actions. The sections below are the
-playbook for avoiding each pattern.
-
-## Trigger semantics
-
-Three event types matter:
-
-| Event                 | When it fires                     | `github.ref` resolves to             | Workflow file loaded from | Fork PR: token / secrets             |
-| --------------------- | --------------------------------- | ------------------------------------ | ------------------------- | ------------------------------------ |
-| `pull_request`        | PR opened / sync / etc.           | merge-ref (`refs/pull/<N>/merge`)    | PR HEAD                   | **Read-only**, no secrets            |
-| `pull_request_target` | same PR events                    | BASE ref (e.g. `refs/heads/develop`) | BASE                      | **Full default**, secrets accessible |
-| `workflow_dispatch`   | API call or "Run workflow" button | The `ref` passed at dispatch         | That ref                  | **Full default**, secrets accessible |
-
-### When to use each
-
-- **`pull_request`** — most workflows. Runs PR code in a safe sandbox.
-  Suitable for "test this PR's code", linting, etc. GitHub auto-restricts
-  permissions/secrets for fork PRs, which is what makes this safe.
-- **`pull_request_target`** — ONLY for bot-like tasks that need write
-  permissions and DO NOT execute PR-author-controlled code. Examples:
-  labeling, commenting, dispatching other workflows. Never `actions/checkout`
-  PR HEAD here without explicit safeguards — that defeats the purpose.
-- **`workflow_dispatch`** — programmatic or manual triggers. Inherits the
-  repo's default GITHUB_TOKEN permissions; NOT fork-restricted. Use when
-  one workflow dispatches another (e.g., `ci_trigger.yml` →
-  `heavy_tests.yml`) or for manual maintainer-driven runs.
-
-### The "pwn request" pattern
-
-A workflow that uses `pull_request_target` AND checks out / executes
-PR-author-controlled code lets an attacker run arbitrary code with full
-repo permissions and secret access. This is the pattern behind most
-recent breaches. To avoid:
-
-- Don't combine `pull_request_target` with PR-code execution.
-- If you must (we do, via `ci_trigger.yml` → `heavy_tests.yml`):
-  - Narrow `permissions:` to the minimum.
-  - Set `persist-credentials: false` on every checkout.
-  - Require human approval before code execution (we use the
-    maintainer-tick gate in `ci-trigger-dispatch.js`).
-
-## Permissions
-
-### Principle of least privilege
-
-Set `permissions:` explicitly on every workflow that touches anything
-sensitive. The repo's default GITHUB_TOKEN permissions are often broader
-than necessary. Declaring `permissions: { ... }` overrides the default;
-unlisted scopes become `none`.
-
-For workflows that run PR-author-controlled code, narrow permissions are
-critical — they limit the blast radius if the token is exfiltrated. Our
-`heavy_tests.yml` is a good example:
-
-```yaml
-permissions:
-  contents: read # for actions/checkout + dorny/paths-filter
-  actions: write # for the rust-tests-cancel / rust-simtests-cancel jobs
-```
-
-Everything else (`pull-requests`, `issues`, `packages`, ...) is `none`.
-
-### How GitHub restricts token / secrets for fork PRs
-
-For `pull_request` events from forks, GitHub automatically:
-
-- Forces GITHUB_TOKEN to **read-only**, regardless of repo settings or
-  `permissions:` declarations.
-- Makes `${{ secrets.X }}` evaluate to empty.
-
-This protection is **bypassed** by `pull_request_target` and
-`workflow_dispatch`. If you use those for fork PRs, the workflow runs with
-full default privileges — narrow `permissions:` explicitly to compensate.
-
-### Token leakage paths
-
-A token can be exposed to running code via:
-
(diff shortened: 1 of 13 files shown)


.github/workflows/heavy_tests.yml
```diff
@@ -0,0 +1,159 @@
+# SECURITY-SENSITIVE: read .github/workflows/README.md before editing.
```
bingyanglin, line 1: The same 7-line warning block is copy-pasted to 8 workflow files, but every point it makes is already covered in the README it tells you to read. The warning mentions `persist-credentials: false`, unpinned actions, `head_sha` plumbing, fork-restricted triggers — all of these have their own sections in the README with full explanations. So the header is just a summary of the README, followed by "read the README." Do you think they might be redundant, adding noise, and wasting context window for LLM/human readers? Maybe just keep a one-liner like `# SECURITY-SENSITIVE: read .github/workflows/README.md before editing` and drop the rest?

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
...
+dispatcher will run for unrelated description edits (extra runner cost,
+not a security issue). If you skip step 4, the checkbox won't appear in
+new PRs.
+
+## References
+
+### GitHub Actions documentation
+
+- [Secure use reference](https://docs.github.com/en/actions/reference/security/secure-use)
+- [Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)
```
bingyanglin: This is the old document. Now it always redirects to https://docs.github.com/en/actions/reference/security/secure-use, and the title is now `Secure use reference`

```suggestion
- [Secure use reference](https://docs.github.com/en/actions/reference/security/secure-use)
```

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
...
+- [tj-actions/changed-files (CVE-2025-30066)](https://github.com/advisories/ghsa-mrrh-fwg8-r2c3)
+- [Wiz — Hardening GitHub Actions: Lessons from Recent Attacks](https://www.wiz.io/blog/github-actions-security-guide)
+- [Snyk — Trivy supply chain compromise](https://snyk.io/articles/trivy-github-actions-supply-chain-compromise/)
+- [Orca — pull_request_nightmare Part 2](https://orca.security/resources/blog/pull-request-nightmare-part-2-exploits/)
+- [Endor Labs — Lessons from Trivy](https://www.endorlabs.com/learn/github-actions-security-lessons-from-trivy)
+- [Unit 42 — tj-actions Supply Chain Attack](https://unit42.paloaltonetworks.com/github-actions-supply-chain-attack/)
+
+### Linting / static analysis tools (consider adding)
+
+- [zizmor](https://github.com/woodruffw/zizmor) — GitHub Actions linter
```
bingyanglin: The `woodruffw` repo was actually removed. It was relocated to `zizmorcore` org
```suggestion
- [zizmor](https://github.com/zizmorcore/zizmor) — GitHub Actions linter
```

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
...
+
+- [Secure use reference](https://docs.github.com/en/actions/reference/security/secure-use)
+- [Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)
+- [Reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)
+- [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows)
+
+### Notable breaches & writeups
+
+- [tj-actions/changed-files (CVE-2025-30066)](https://github.com/advisories/ghsa-mrrh-fwg8-r2c3)
+- [Wiz — Hardening GitHub Actions: Lessons from Recent Attacks](https://www.wiz.io/blog/github-actions-security-guide)
```
bingyanglin: The title is wrong.
```suggestion
- [Wiz — How to Harden GitHub Actions: An Updated Guide](https://www.wiz.io/blog/github-actions-security-guide)
```

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
...
+not a security issue). If you skip step 4, the checkbox won't appear in
+new PRs.
+
+## References
+
+### GitHub Actions documentation
+
+- [Secure use reference](https://docs.github.com/en/actions/reference/security/secure-use)
+- [Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)
+- [Reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)
```
bingyanglin: The url was old
```suggestion
- [Reusing workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)
```

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
...
+- [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows)
```
bingyanglin: The url is the legacy one.
```suggestion
- [Events that trigger workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows)
```

bingyanglin (commented):
You might need to use `dprint fmt` to fix the fmt error

.github/workflows/heavy_tests.yml
chriscandless: LLMs do not follow strict rules - they may not read the file even if they see this. So such severe explanation may increase chances the LLM will follow it. Moreover, the explanation contain specific list of dangerous changes, so again it may pop up, when such changes are made. I see wastage oа context in this case as a minor issue in comparison to the consequences of not following those rules. There is no "overkill" when we deal with security.

Plus additional teeny-tiny advantage is that humans can also search for `persist-credentials: false` and find this comment.

So I personally would prefer to leave it as this.

.github/workflows/README.md
chriscandless: Nice! Applying

bingyanglin (approved):

.github/workflows/README.md
```diff
@@ -0,0 +1,394 @@
+# CI workflows: design and security guide
+
+This directory contains GitHub Actions workflows for `iotaledger/iota`.
+**Read this guide before editing any workflow file.** Several of these
+workflows interact with GitHub's security model in subtle ways — small
+changes (a different trigger, a missing `persist-credentials: false`, an
+unpinned action) can introduce critical vulnerabilities or quietly break
+the intended invariants.
+
+The guide is written for both humans and LLMs assisting with edits.
```
Thoralf-M: Even though it's also for LLMs, I would reduce the content a lot. No need for `## Why this matters` and the references section, also other parts should be simplified. Nobody working on this here cares about all the references CVEs

chriscandless: Agree, README may be simplified, I also had that in mind. Will do

chriscandless: Done - left only pure info about how-to do it properly, what changes to review more cautiously, and help two references to pages about safety.

muXxer:
Superseeded by the new merge-queue feature of github.
#12322

Issue #11412, opened 2026-05-04
node: syncing from genesis causes OOM


muXxer:
Syncing from genesis causes OOM. There seems to be a memleak for long running nodes. We need to investigate this issue.

Test plan:
- sync a node from mainnet genesis (> 100 epochs) and monitor the memory usage

jkrvivian:
## Root causes & fixes

| Cause | Fix |
|---|---|
| `AuthorityPerEpochStore` pruning was time-based; fast genesis sync crossed epoch boundaries faster than the interval, so epochs accumulated. | [eb6ad1c](https://github.com/iotaledger/iota/commit/eb6ad1c469c07a1a446a07baf5141c5ed4216547) — prune on every epoch transition. |
| `PeerHeights` maps grew unbounded across millions of checkpoints — no cleanup path. | [2983501](https://github.com/iotaledger/iota/commit/29835011c1f8257f00739c5a7f8f4ca5404092d5) — periodic cleanup every 10k verified checkpoints. |
| After cleanup, hashbrown bucket array kept doubling because `retain` leaves tombstones that count toward the load factor (>300 MiB monotonic plateau). | [#11551](https://github.com/iotaledger/iota/pull/11551) — `shrink_to_fit()` after `retain`. |

With all three in place, the unbounded growth observed in genesis-sync heap profiles is eliminated. After the fixes,  for a fullnode syncing from genesis, the memory usage is around 23 - 24GB. 🎉 

The main memory leaks in the Rust side are resolved. To improve it further, the operators can tweak the RocksDB parameters to cap RocksDB memory, such as:
* `objects` block cache size in `crates/iota-core/src/authority/authority_store_tables.rs`
* `cache_index_and_filter_blocks`
* `pin_l0_filter_and_index_blocks_in_cache`

PR #11551, opened 2026-05-18
fix(node): Call shrink_to_fit on HashMaps in PeerHeights

jkrvivian:
# Description of change

Call shrink_to_fit() on both PeerHeights maps (unprocessed_checkpoints, sequence_number_to_digest) after retain.

PeerHeights is periodically trimmed by `cleanup_old_checkpoints` every 10,000 verified checkpoints. The trim uses HashMap::retain, which removes entries logically but does not shrink the underlying bucket array — removed slots are turned into tombstones.

So in a long sync, each insert/retain cycle pushes the bucket array up by a doubling, even though the live set stays small. On a fullnode syncing from genesis we observed this on PeerHeights by jemalloc heap profiles taken mid-sync, with the allocation backtraces pointing at the hashbrown rehash path from insert_checkpoint.

shrink_to_fit rebuilds each map to fit only the live entries, dropping all tombstones.

For a node starting from a recent snapshot, the catch-up window to the tip is typically small, so this fires only a handful of times before the node is at the tip. Once it's synced, there is no sync_to_checkpoint task running, so the call doesn't fire at all.

Part of #11412

### DIFF
--- a/crates/iota-network/src/state_sync/mod.rs
+++ b/crates/iota-network/src/state_sync/mod.rs
@@ -256,10 +256,15 @@ impl PeerHeights {
     }
 
     pub fn cleanup_old_checkpoints(&mut self, sequence_number: CheckpointSequenceNumber) {
+        // `retain` leaves tombstones that count toward the load factor; without
+        // the follow-up `shrink_to_fit`, the bucket array keeps doubling on
+        // subsequent inserts during long syncs.
         self.unprocessed_checkpoints
             .retain(|_digest, checkpoint| *checkpoint.sequence_number() > sequence_number);
+        self.unprocessed_checkpoints.shrink_to_fit();
         self.sequence_number_to_digest
             .retain(|&s, _digest| s > sequence_number);
+        self.sequence_number_to_digest.shrink_to_fit();
     }
 
     // TODO: also record who gives this checkpoint info for peer quality


muXxer (approved):

alexsporn (approved):

crates/iota-network/src/state_sync/mod.rs
```diff
@@ -256,10 +256,15 @@ impl PeerHeights {
     }
 
     pub fn cleanup_old_checkpoints(&mut self, sequence_number: CheckpointSequenceNumber) {
+        // `retain` leaves tombstones that count toward the load factor; without
+        // the follow-up `shrink_to_fit`, the bucket array keeps doubling on
+        // subsequent inserts during long syncs.
         self.unprocessed_checkpoints
             .retain(|_digest, checkpoint| *checkpoint.sequence_number() > sequence_number);
+        self.unprocessed_checkpoints.shrink_to_fit();
```
semenov-vladyslav, line 264: If the goal is to minimize the HashMap footprint, maybe we can consider storing boxed values instead of plain `Checkpoint` which is pretty heavy.

semenov-vladyslav (approved):

crates/iota-network/src/state_sync/mod.rs
jkrvivian: Yeah, that's something we can consider.

PR #11559, opened 2026-05-18
refactor(snapshot): insert partition to generic database

### LINKED ISSUE #11308: refactor(iota-snapshot): decouple reading from writing

kodemartin:
# Description of change

This patch refactors `iota-snapshot` so that to extend support for insertion of object partitions to databases other than node database. The motivation was to allow restoring the IOTA Indexer with formal snapshots.

See #11308 for more context.

Fixes #11308 

### DIFF
--- a/crates/iota-snapshot/src/database.rs
+++ b/crates/iota-snapshot/src/database.rs
@@ -1,0 +1,43 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+//! Logic and abstractions for supporting different databases that the snapshot
+//! can restore.
+
+use std::future::Future;
+
+use anyhow::Result;
+use bytes::Bytes;
+use iota_core::authority::{AuthorityStore, authority_store_tables::AuthorityPerpetualTables};
+use iota_storage::SHA3_BYTES;
+
+use crate::{FileMetadata, reader::LiveObjectIter};
+
+/// A trait for databases that can be restored from a formal snapshot.
+pub trait Database {
+    /// Inserts a partition of live objects.
+    ///
+    /// The checksum of the partition can be computed as the SHA3 hash of all
+    /// objects' digests included. Then it can be verified against the given
+    /// expected value.
+    fn insert_partition(
+        &self,
+        file_metadata: FileMetadata,
+        bytes: Bytes,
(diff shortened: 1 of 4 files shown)


crates/iota-snapshot/src/database.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use anyhow::Result;
+use bytes::Bytes;
+use iota_core::authority::{AuthorityStore, authority_store_tables::AuthorityPerpetualTables};
+use iota_storage::SHA3_BYTES;
+
+use crate::{FileMetadata, reader::LiveObjectIter};
+
+/// A trait for databases that can be restored from a formal snapshot.
+pub trait Database {
```
muXxer: Just a small nit, but also fine with me. Should we maybe call it a bit more specific, like `SnapshotRestoreTarget` or `RestoreTarget` instead of generic `Database`. Might be less confusing.

muXxer (approved):

crates/iota-snapshot/src/database.rs
kodemartin: hmm...what do you think about simple `Restore`? we already have the `iota_snapshot` namespace, so a backend `impl iota_snapshot::restore::Restore` would be rather intuitive no?

kodemartin: https://github.com/iotaledger/iota/pull/11559/commits/8d16ab3b9c9ae4b2ede8f9e396f4e53451ad58c5

crates/iota-snapshot/src/restore.rs
```diff
@@ -0,0 +1,43 @@
...
+pub trait Restore {
+    /// Inserts a partition of live objects.
+    ///
+    /// The checksum of the partition can be computed as the SHA3 hash of all
+    /// objects' digests included. Then it can be verified against the given
+    /// expected value.
+    fn insert_partition(
+        &self,
+        file_metadata: FileMetadata,
+        bytes: Bytes,
```
tomxey, line 26: Is there some use for having those arguments separately, instead of passing one `LiveObjectIter`?

kodemartin: yes, [LiveObjectIter](https://iotaledger.github.io/iota/src/iota_snapshot/reader.rs.html#658-660) is not `Send` as it wraps `dyn Read`, and this would be restrictive in contexts where the feature would have to be `Send` (e.g. when called within `tokio::task::spawn`.

tomxey (approved):

PR #11561, opened 2026-05-18
feat(iota-core): persist ReportAggregator state across restarts

piotrm50:
# Description of change

Persists `ReportAggregator` state to RocksDB so per-validator received-reports tallies survive node restarts. Stacks on top of #11534.

Refs https://github.com/iotaledger/iota/pull/10366 (the original implementation by @oliviasaa, which we couldn't reasonably rebase post-Starfish migration).

## What gets persisted

- `DBReceivedReportsStatePerAuthority { received_metrics, invalid_reports_count }` for each authority that has changed state in this epoch.
- New column family on `AuthorityEpochTables`: `received_reports_state: DBMap<u32, DBReceivedReportsStatePerAuthority>`.

## How writes work

When `process_report` mutates the aggregator for authority X, it returns the post-merge snapshot. The consensus handler stashes that snapshot into `ConsensusCommitOutput.report_state_snapshots[X]` (a `BTreeMap`, last-writer-wins). When the quarantine flushes the commit's `DBBatch`, the captured snapshots get written atomically with the rest of the commit's effects.

This means the persisted row written under commit N's batch is **exactly** "aggregator state for that authority after commit N's processing" — independent of when quarantine flushes the batch, and matching what `scoreboard.update_scores` saw at end of N. Cross-validator determinism preserved.

## How restore works

`AuthorityPerEpochStore::new` calls `ReportAggregator::restore_from_tables(&tables)` immediately after constructing the aggregator. The existing consensus-handler replay path then reprocesses any committed-but-unflushed commits; merge-max idempotency reproduces the same end state.

## Explicit non-goals

- **Score-update optimization** (recalc only when ≥1 report seen in commit) — tracked separately.
- **Verify-time invalid-count durability**. `verify_consensus_transaction`-time `increment_invalid_reports_count` bumps stay in-memory only. They land on disk the next time a `process_report` for the same authority captures a fresh snapshot. Counter is observability-only; replay re-runs verify on the recovery path, so correctness is preserved. Documented in code.
- **Separate replay logic for unflushed commits**. Not needed — the existing consensus-handler recovery path replays for free, and merge-max is idempotent.

fixes https://github.com/iotaledger/iota-private/issues/310

Local verification:
- `cargo ci-clippy` — clean.
- `IOTA_SKIP_SIMTESTS=1 cargo nextest run -p iota-core --lib` — 562 tests pass.
- New tests:
  - `test_restore_from_iter_populates_in_memory_state`
  - `test_restore_from_iter_rejects_out_of_range_authority`
  - `test_process_report_returns_post_merge_snapshot`
  - `test_restore_round_trip_through_dbmap` — full persist→restore round trip via `AuthorityEpochTables::open` with a tempdir.

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -1094,6 +1094,9 @@ impl AuthorityPerEpochStore {
         let report_version = MisbehaviorReportVersion::from_protocol(&protocol_config);
         let misbehavior_monitor = MisbehaviorMonitor::new(name, report_version, committee_size);
         let report_aggregator = ReportAggregator::new(report_version, committee_size);
+        report_aggregator
+            .restore_from_tables(&tables)
+            .expect("AuthorityEpochTables should contain valid ReportAggregator state");
         let voting_power = committee.members().map(|(_, v)| *v).collect::<Vec<u64>>();
         let scoreboard = Scoreboard::new(voting_power, &protocol_config);
 
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -836,6 +836,17 @@ pub struct AuthorityEpochTables {
 
     /// Accumulated per-object debts for randomness congestion control.
     congestion_control_randomness_object_debts: DBMap<ObjectID, CongestionPerObjectDebt>,
+
+    /// Per-validator received misbehavior reports state. Keyed by the
+    /// reporter's `AuthorityIndex` (the in-memory `Vec` index used by
+    /// `ReportAggregator`). Each entry is the merged-max observation set plus
+    /// invalid-report counter, snapshotted from the live aggregator when its
+    /// state changes inside a consensus commit. Written atomically with the
+    /// rest of `ConsensusCommitOutput::write_to_batch`. Survives restart;
+    /// `ReportAggregator::restore_from_tables` loads it during epoch-store
+    /// construction.
+    pub(crate) received_reports_state:
+        DBMap<u32, report_aggregator::DBReceivedReportsStatePerAuthority>,
 }
 
 fn signed_transactions_table_default_config() -> DBOptions {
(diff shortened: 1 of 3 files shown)


crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -836,6 +836,17 @@ pub struct AuthorityEpochTables {
...
+    /// Per-validator received misbehavior reports state. Keyed by the
+    /// reporter's `AuthorityIndex` (the in-memory `Vec` index used by
+    /// `ReportAggregator`). Each entry is the merged-max observation set plus
+    /// invalid-report counter, snapshotted from the live aggregator when its
+    /// state changes inside a consensus commit. Written atomically with the
+    /// rest of `ConsensusCommitOutput::write_to_batch`. Survives restart;
+    /// `ReportAggregator::restore_from_tables` loads it during epoch-store
+    /// construction.
+    pub(crate) received_reports_state:
+        DBMap<u32, report_aggregator::DBReceivedReportsStatePerAuthority>,
```
alexsporn: why an `u32` here for `AuthorityIndex`?

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -133,6 +141,15 @@ impl ConsensusCommitOutput {
         self.consensus_messages_processed.insert(key);
     }
 
+    pub(crate) fn record_report_state_snapshot(
+        &mut self,
+        authority_index: u32,
```
alexsporn: Can't we use a shared typealias for `AuthorityIndex` everywhere? This would make the third one, we have u8, u32 and u64 definitions of these

crates/iota-core/src/authority/report_aggregator.rs
```diff
@@ -1,13 +1,21 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+//! `invalid_reports_count` bumps from `verify_consensus_transaction` are not
```
alexsporn: This comment seems to be misplaced here at the top

alexsporn (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: ConsensusOutputApi's `AuthorityIndex` is still u32 and as a result we don't downscale this later in iota-core to u8. I will investigate if we can change `AuthorityIndex` to `u8` completely, to match the committee size type in Starfish.

cyberphysic4l (approved):

VorobyevIlya (approved):

semenov-vladyslav (approved):

PR #383, opened 2024-05-24
feat(sui-genesis-builder): Test alias migration with native tokens

PhilippGackstatter:
# Description of change

Test alias migration with native tokens.

Eventually, the other alias migration tests should also be renamed to remove the `test_` prefix, according to the upcoming guidelines, but I did not want to touch too many things since other PRs will conflict with this one.

I also stole a the foundry creation function from #209 - we should combine those eventually.

I'm extracting the underlying `Executor` out of the `Migration` struct. This gives access to the `native_tokens` map (useful in tests to lookup the coin id) and avoids needing to add the created objects to a newly created `Executor`. Running the migration through `Migration` rather than `Executor` directly also avoids having to effectively duplicate `Migration::migrate_foundries` and `Migration::migrate_outputs`. I'm still of the opinion that testing the migration is best done through the `Migration` struct because of the above-mentioned methods that would only have to be reimplemented for tests when using the `Executor` directly.

(Still getting used to the new commit format, so I did forget it in some cases).

## Links to any relevant issues

fixes #202

## How the change has been tested

`cargo test --package sui-genesis-builder --lib --all-features`

## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

PR #11574, opened 2026-05-19
feat(iota-core, iota-types): wire validator attestation into sequencer and execution

vekkiokonio:
### Description of change

This PR is **Phase 1.3** of the [Validator Attestation](https://github.com/iotaledger/iota-private/issues/383). It wires the attestor's pre-consensus estimate into the sequencer's per-object congestion control, propagates the attestation through to execution for future consumers (attestor reward/penalty accounting, divergence metrics), and adds post-consensus safety checks so malformed or stale attestations cannot affect scheduling.

_This PR targets `protocol-research/feat/validator-attestation`, not `develop`._

### What changed

**New congestion-control mode** (`iota-protocol-config`)

* `PerObjectCongestionControlMode::TotalComputationCost` — uses the attestor's `estimated_computation_cost` from `UserTransactionV2` for shared-object scheduling. Falls back to `gas_budget / gas_price` for unattested transactions.

**Pipeline propagation** (`iota-core` + `iota-types`)

* New wrapper `VerifiedExecutableAttestedTransaction { tx: VerifiedExecutableTransaction, attestation: Option<Attestation> }` in `iota-types/src/executable_transaction.rs`. Fields are encapsulated; access is via `attested_computation_cost()` and `into_parts()`. Implements `Deref<Target = VerifiedExecutableTransaction>` and `From<VerifiedExecutableTransaction>` so existing call sites keep working unchanged.
* `ConsensusTransactionResult::{Scheduled, Cancelled}` now carry the wrapper instead of `VerifiedExecutableTransaction`.
* The V1/V2 dispatch in `process_consensus_transaction` is flattened into one arm per variant via a shared `try_schedule_user_transaction` helper that takes `Option<Attestation>`.
* `TransactionManager::enqueue_attested` added; `PendingCertificate.certificate` is now `VerifiedExecutableAttestedTransaction`, making the attestation available at execution time for future consumers.

**Scheduling integration** (`iota-core/authority/authority_per_epoch_store.rs`, `shared_object_congestion_tracker.rs`, `suggested_gas_price_calculator.rs`)

* `get_estimated_execution_duration`, `try_schedule`, `bump_object_execution_slots`, `calculate_suggested_gas_price`, and `find_clearing_gas_price` now take `&VerifiedExecutableAttestedTransaction`. In the new mode, the duration is `attestation.estimated_computation_cost()`, with `gas_budget / gas_price` as the unattested-fallback.

**Post-consensus safety for V2 attested transactions** (`iota-core/post_consensus_validation.rs`, `authority.rs`)

* **Cost floor** — `estimated_computation_cost < base_tx_cost_fixed()` is rejected as `IotaError::AttestationCostBelowMinimum`. The lowest computation bucket guarantees an honest dry-run reports at least `base_tx_cost_fixed`; anything below is malformed and could otherwise bypass per-object congestion accounting via a duration of zero.
* **Coin deny-list re-check** — for V2 attested transactions, `check_coin_deny_list_for_attested_tx` runs at post-consensus (where rejection is safe) rather than at execution time (where a deny-list-driven abort would crash the validator). Catches deny-list-state drift between the attestor's view and the validator's current view. Input-load failures (e.g., stale-attestation `ObjectNotFound`) and actual deny-list violations are logged distinctly.

**Tests**

* Existing congestion-control test matrices in `shared_object_congestion_tracker.rs`, `suggested_gas_price_calculator.rs`, and `authority_tests.rs` updated to exercise `TotalComputationCost` via shared `TotalGasBudget | TotalComputationCost` match arms (the two modes share scheduling semantics; only the duration source differs).
* New unit tests in `authority_per_epoch_store_tests.rs`:
  * `test_get_estimated_execution_duration_total_computation_cost_mode` — attested cost vs. unattested-fallback.
  * `test_total_computation_cost_attested_vs_unattested_commit_scheduling` — end-to-end commit scheduling under the new mode.
* New unit tests in `post_consensus_validation_tests.rs`:
  * `test_v2_cost_below_minimum` — verifies `AttestationCostBelowMinimum` rejection still surfaces the digest for soft-lock release.

### Links to any relevant issues

Closes [#386](https://github.com/iotaledger/iota-private/issues/386)
Part of Epic [#383](https://github.com/iotaledger/iota-private/issues/383)

### How the change has been tested

### Release Notes

- [x] Protocol: New `PerObjectCongestionControlMode::TotalComputationCost` variant. Requires `enable_white_flag_flow` and `enable_validator_attestation`.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-core/src/authority.rs
```diff
@@ -1807,6 +1813,20 @@ impl AuthorityState {
...
+                let tx_receiving_objects = self
+                    .input_loader
+                    .read_receiving_objects(&tx_data.receiving_objects(), epoch_store.epoch())?;
+                check_coin_deny_list_v1(
+                    tx_data.sender(),
+                    &tx_checked_input_objects,
+                    &tx_receiving_objects,
+                    &vec![],
+                    &self.get_object_store(),
+                )?;
```
roman1e2f5p8s: if deny list changes between attestation and this call site, `check_coin_deny_list_v1` would lead to `fatal!` during execution, so all validators would crash

proposed fix: this check should be removed from the `try_execute_immediately` path and moved to post-consensus validation and cause a tx drop rather than a validator node crash.

crates/iota-core/src/authority.rs
```diff
@@ -1899,6 +1919,20 @@ impl AuthorityState {
...
+                    .input_loader
+                    .read_receiving_objects(&tx_data.receiving_objects(), epoch_store.epoch())?;
+                check_coin_deny_list_v1(
+                    tx_data.sender(),
+                    &authenticator_and_tx_checked_input_objects,
+                    &tx_receiving_objects,
+                    &vec![],
+                    &self.get_object_store(),
+                )?;
+            }
```
roman1e2f5p8s: same concern as https://github.com/iotaledger/iota/pull/11574/changes#r3342606097

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -270,12 +272,15 @@ impl CongestionControlParameters {
...
-        cert: &VerifiedExecutableTransaction,
+        cert: &VerifiedExecutableAttestedTransaction,
     ) -> ExecutionTime {
         match self.per_object_congestion_control_mode {
             PerObjectCongestionControlMode::None => 0,
             PerObjectCongestionControlMode::TotalGasBudget => cert.gas_budget(),
             PerObjectCongestionControlMode::TotalTxCount => 1,
+            PerObjectCongestionControlMode::TotalComputationCost => cert
+                .attested_computation_cost()
+                .unwrap_or_else(|| cert.gas_budget()),
```
roman1e2f5p8s: but `attested_computation_cost()` is compute units, not computation cost in NANOS:

https://github.com/iotaledger/iota/blob/b39a7f0ea5da36d8983b80e89a4304c0386cab22/crates/iota-core/src/authority.rs#L1117-L1127

unattested tx would fall back to NANO here.

fix:

```suggestion
             PerObjectCongestionControlMode::TotalComputationCost => {
                cert.attested_computation_cost().unwrap_or_else(|| {
                    cert.gas_budget()
                        .checked_div(cert.data().transaction_data().gas_price())
                        .unwrap_or(0)
                })
            }
```

crates/iota-core/src/authority.rs
```diff
@@ -1899,6 +1919,20 @@ impl AuthorityState {
...
+                let tx_receiving_objects = self
+                    .input_loader
+                    .read_receiving_objects(&tx_data.receiving_objects(), epoch_store.epoch())?;
+                check_coin_deny_list_v1(
+                    tx_data.sender(),
+                    &authenticator_and_tx_checked_input_objects,
+                    &tx_receiving_objects,
+                    &vec![],
+                    &self.get_object_store(),
+                )?;
```
roman1e2f5p8s: the attestor (`authority.rs:5745`) runs with a non-empty `per_authenticator_checked_input_objects`, but here we pass them empty

```suggestion
                let per_authenticator_checked_input_objects =
                    per_authenticator_checked_input_objects
                        .iter()
                        .collect::<Vec<&_>>();
                check_coin_deny_list_v1(
                    tx_data.sender(),
                    &authenticator_and_tx_checked_input_objects,
                    &tx_receiving_objects,
                    &per_authenticator_checked_input_objects,
                    &self.get_object_store(),
                )?;
```

crates/iota-core/src/authority/authority_per_epoch_store.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11574/changes/e26ac47dd6631016b2971fb8f0313875a57506ed.

Used `cert.transaction_data()` instead of `cert.data().transaction_data()`.

crates/iota-core/src/authority.rs
vekkiokonio: Good catch! I moved the check to post-consensus validation: https://github.com/iotaledger/iota/pull/11574/changes/0d5e4bef4ffba296df0bc67adc76c720be4c775c

crates/iota-core/src/authority.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11574/changes/0d5e4bef4ffba296df0bc67adc76c720be4c775c

crates/iota-core/src/authority.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11574/changes/0d5e4bef4ffba296df0bc67adc76c720be4c775c

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -229,13 +229,26 @@ impl ValidatorService {
...
-                .await
-                .expect("attest_transaction blocking task panicked");
+                .await;
+                let (result, verified_tx) = match join_result {
+                    Ok(pair) => pair,
+                    Err(join_err) => {
+                        // A panic inside `attest_transaction` (e.g., an
+                        // arithmetic panic in Move VM dry-run) surfaces here
+                        // as a `JoinError`. Convert to a rejection rather
+                        // than re-panicking on the per-tx task.
```
roman1e2f5p8s, line 244: how about also logging here rather than silently moving the error to rejected reason?
```suggestion
                        // than re-panicking on the per-tx task.
                        tracing::error!(?tx_digest, "attest_transaction task failed: {join_err}");
```

roman1e2f5p8s (commented):
clippy not fixed

crates/iota-types/src/transaction.rs
```diff
@@ -2998,6 +2998,12 @@ impl CheckedInputObjects {
         Self(input_objects)
     }
 
+    // Only called by post-consensus coin deny-list re-checks for attested
+    // transactions.
+    pub fn new_for_deny_list_check_only(input_objects: InputObjects) -> Self {
```
Thoralf-M: What's the point of the different `new_` methods if they do exactly the same?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -163,17 +164,25 @@ pub async fn validate_and_resolve_conflicts(
...
+            let min_attested_cost = epoch_store.protocol_config().base_tx_cost_fixed();
+            let attested_cost = attestation.estimated_computation_cost();
             let error = match attestation {
                 Attestation::Validator { attestor_index, .. } => {
                     if *attestor_index != block_author {
                         Some(IotaError::AttestationAuthorMismatch {
                             expected: *attestor_index,
                             actual: block_author,
                         })
+                    } else if attested_cost < min_attested_cost {
```
bingyanglin: Is it safe without ceiling check? An attestor can claim anything from the floor up and under-report to skip congestion limits, or over-report to grief other users of a shared object. Is this ok? 
Also the doc on `object_versions` says it's there to tell misbehavior from honest state drift, but `object_versions` is never read anywhere. Is this an known issue and will be fixed (or already fixed?)

crates/iota-core/src/authority.rs
```diff
@@ -1787,6 +1796,32 @@ impl AuthorityState {
         let tx_data = certificate.data().transaction_data();
         tx_data.validity_check(protocol_config)?;
 
+        // V2 attested path: re-verify the user signature here as a Byzantine attestor
```
piotrm50: Actually, herre we need to perform all the checks that the attestor performed.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -325,6 +343,34 @@ pub async fn validate_and_resolve_conflicts(
                 keep[i] = false;
                 continue;
             }
+        } else {
+            let verified_tx = VerifiedTransaction::new_from_verified(transaction.clone());
+            if let Err(e) = authority_state
```
piotrm50, line 379: wait, as far as I understood, we want to run this pre-execution instead of here. In theory, the state can still change between here and the execution, no? 

My general understanding was that the only place where every attested check is re-verified is right before execution.

crates/iota-types/src/executable_transaction.rs
```diff
@@ -80,3 +83,56 @@ impl VerifiedExecutableTransaction {
...
+
+/// Wraps a [`VerifiedExecutableTransaction`] with its pre-consensus
+/// [`Attestation`] (if any). Carrying the full attestation lets downstream code
+/// consult both scheduling metadata and attestor identity / observed object
+/// versions.
+#[derive(Clone, Debug)]
+pub struct VerifiedExecutableAttestedTransaction {
+    tx: VerifiedExecutableTransaction,
+    /// `None` for unattested transactions (e.g., `UserTransactionV1`).
+    attestation: Option<Attestation>,
```
piotrm50: Is a new struct really needed here? Why can't we add an `Option<Attestation>` field to the existing `VerifiedExecutableTransaction`?

Plus, this type is only used in `iota-core` so perhaps this should be moved there.

crates/iota-types/src/transaction.rs
piotrm50: Also, the comment above says 
```
// DO NOT CALL outside of iota-transaction-checks, genesis, or replay.
```

so I'm not sure if adding this method here and calling it in `iota-core` is correct. Not sure why the comment was added there, but I feel like this should be looked into in more detail. In the case that this is justified, the comment should be updated.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -1064,15 +1065,18 @@ mod object_cost_tests {
 
         let (max_execution_duration_per_commit, max_overshoot_per_commit) = match mode {
             PerObjectCongestionControlMode::None => unreachable!(),
-            PerObjectCongestionControlMode::TotalGasBudget => (12, 0),
+            PerObjectCongestionControlMode::TotalGasBudget
+            | PerObjectCongestionControlMode::TotalComputationCost => (12, 0),
```
piotrm50: Should `TotalComputationCost` be added to the list of values this test is run with? 
```
 #[values(
            PerObjectCongestionControlMode::TotalGasBudget,
            PerObjectCongestionControlMode::TotalTxCount
        )]
        mode: PerObjectCongestionControlMode,
```

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
piotrm50, line 1119: `build_transaction` does not have any special logic for `TotalComputationCost` mode.  Is this by design that we do not want to differentiate between `TotalGasBudget` and `TotalComputationCost`?

crates/iota-types/src/attestation.rs
piotrm50, line 63: This value is never read. Is this planned to be used in a future PR?

piotrm50 (changes requested):

crates/iota-types/src/transaction.rs
vekkiokonio: Good points. I had added `new_for_deny_list_check_only` because `check_coin_deny_list_v1` required `&CheckedInputObjects` and the post-consensus path didn't have a way to produce one.

However, since the function only reads `obj.coin_type_opt()` and the type guarantees of `CheckedInputObjects` are anyway enforced at execution-time, I changed the function to take `&InputObjects` directly and deleted `new_for_deny_list_check_only`. Fixed in https://github.com/iotaledger/iota/pull/11574/changes/76b0b6e05182e71cd165eed42f51ac42d0fa2b2e.

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: > Also the doc on object_versions says it's there to tell misbehavior from honest state drift, but object_versions is never read anywhere. Is this an known issue and will be fixed (or already fixed?)

Object versions will be used once we implement misbehavior detection system in Phase 3 (see https://github.com/iotaledger/iota-private/issues/392).

vekkiokonio: > Is it safe without ceiling check? An attestor can claim anything from the floor up and under-report to skip congestion limits, or over-report to grief other users of a shared object. Is this ok?

That's a good point. I've added a ceiling check in https://github.com/iotaledger/iota/pull/11574/changes/21a748a0818a2349dc0aff9e8d6af60b3aec71a0. Attestation computational units mismatches should be captured during execution.

crates/iota-types/src/attestation.rs
vekkiokonio: Yes, object versions will be used once we implement misbehavior detection system in Phase 3 (see https://github.com/iotaledger/iota-private/issues/392).

crates/iota-core/src/authority.rs
```diff
@@ -1114,8 +1117,15 @@ impl AuthorityState {
...
-        let estimated_computation_cost = effects.gas_cost_summary().computation_cost;
+        // Step 7: build AttestationData.
+        // `gas_cost_summary().computation_cost` is in NANO; convert to gas
+        // units (`computation_units = computation_cost / gas_price`) so the
+        // attestation is independent of the gas price the user chose.
+        let estimated_computation_cost = effects
+            .gas_cost_summary()
+            .computation_cost
+            .checked_div(tx_data.gas_price())
+            .unwrap_or(0);
```
roman1e2f5p8s: this part is confusing

`estimated_computation_cost` is derived from `computation_cost` - the only difference in the names is "estimated". but the units of both are different.

I think `estimated_computation_cost` should carry the raw computation cost in `NANOS` from dry-run effects - this is the only ground truth here, not compute units. the only place where we will need compute units is for the sequencer, so let the sequencer divide the raw computation cost by the gas price there - that would be the only place where `checked_div` lives.

crates/iota-types/src/attestation.rs
```diff
@@ -50,7 +50,7 @@ pub enum Attestation {
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub enum AttestationData {
     V1 {
-        /// Expected computation cost, in NANOS, used by the sequencer to
+        /// Expected computation cost, in gas units, used by the sequencer to
         /// improve shared-object scheduling before execution.
         estimated_computation_cost: u64,
```
roman1e2f5p8s: `computation_cost` is widely used in the codebase, and its units are `NANOS`.
so anyone looking at `estimated_computation_cost` might first think its units are also `NANOS`.

if you want the attestation data to carry gas units, then rename this field to `estimated_compute_units`.

but I do think we should keep this field named as `estimated_computation_cost` but carry the raw computation cost in `NANOS` from dry-run effects - this is the only ground truth here, not compute units. the only place where we will need compute units is for the sequencer, so let the sequencer divide the raw computation cost by the gas price - that wuld be the only place where `checked_div` lives.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -524,8 +524,9 @@ impl ConsensusTransactionOrdering {
 pub enum PerObjectCongestionControlMode {
     #[default]
     None, // No congestion control.
-    TotalGasBudget, // Use txn gas budget as execution cost.
-    TotalTxCount,   // Use total txn count as execution cost.
+    TotalGasBudget,       // Use txn gas budget as execution cost.
+    TotalTxCount,         // Use total txn count as execution cost.
+    TotalComputationCost, // Use attested computation cost as execution cost.
```
roman1e2f5p8s: this should be renamed since we use compute units, not computation cost in `NANOS`, in the sequencer

```suggestion
    /// Use total compute (gas) units, derived from attested computation cost,
    as execution cost.
    TotalComputeUnits,
```

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -163,17 +164,35 @@ pub async fn validate_and_resolve_conflicts(
...
                     if *attestor_index != block_author {
                         Some(IotaError::AttestationAuthorMismatch {
                             expected: *attestor_index,
                             actual: block_author,
                         })
+                    } else if attested_cost < min_attested_cost {
+                        Some(IotaError::AttestationCostBelowMinimum {
+                            actual: attested_cost,
+                            minimum: min_attested_cost,
+                        })
```
roman1e2f5p8s: an honest transaction could be wrongly dropped here if `base_tx_cost_fixed > gas_rounding_step` - currently both coincide at 1000, but they are independent knobs, and there is no assertion that `base_tx_cost_fixed == gas_rounding_step`.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -163,17 +164,35 @@ pub async fn validate_and_resolve_conflicts(
 
         // Check #3: Attestor verification (UserTransactionV2 only).
         // The block signature transitively authenticates the attestation;
-        // verify the claimed attestor matches the actual block author.
+        // verify the claimed attestor matches the actual block author and that
+        // the payload is not malformed.
         if let Some(attestation) = attestation {
             let block_author =
                 starfish_config::AuthorityIndex::from(tx.0.certificate_author_index as u8);
+            let min_attested_cost = epoch_store.protocol_config().base_tx_cost_fixed();
```
roman1e2f5p8s: `base_tx_cost_fixed` is the budget floor; `gas_rounding_step` is the computation floor.
this only works because `base_tx_cost_fixed == gas_rounding_step == 1000` today.

but if it ever becomes that `base_tx_cost_fixed > gas_rounding_step` (likely does not make much sense in practice, but possible - there are no guards against it), an honest transaction could be wrongly dropped with `AttestationCostBelowMinimum` 

```suggestion
            let protocol_config = epoch_store.protocol_config();
            let min_attested_cost = protocol_config
                .base_tx_cost_fixed()
                .min(protocol_config.gas_rounding_step());
```

roman1e2f5p8s (changes requested):

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -163,17 +164,35 @@ pub async fn validate_and_resolve_conflicts(
...
+            let attested_cost = attestation.estimated_computation_cost();
+            let tx_data = transaction.data().transaction_data();
+            let max_attested_cost = tx_data
+                .gas_budget()
+                .checked_div(tx_data.gas_price())
+                .unwrap_or(u64::MAX);
```
roman1e2f5p8s, line 185: this would not be needed at all if `estimated_computation_cost` held raw computation cost in `NANOS`

crates/iota-core/src/authority.rs
vekkiokonio: Sure, all checks but deny config and deny list are performed again pre-execution. A detailed list of checks has been added to the appendix of the Transaction Attestation document. A related list can also be found in https://github.com/iotaledger/iota-private/issues/383.

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: This check has been moved to post-consensus to avoid validator crashes. See https://github.com/iotaledger/iota/pull/11574#discussion_r3342606097

crates/iota-types/src/error.rs
```diff
@@ -539,6 +539,28 @@ pub enum IotaError {
...
+         an honest dry-run cannot bucketize below this floor"
+    )]
+    AttestationCostBelowMinimum { actual: u64, minimum: u64 },
+
+    #[error(
+        "Attestation reports estimated_computation_cost = {actual} gas units, \
+         above the user's budget ceiling of {ceiling} (`gas_budget / gas_price`); \
+         an honest dry-run cannot exceed what the tx can pay for"
+    )]
+    AttestationCostAboveBudget { actual: u64, ceiling: u64 },
```
roman1e2f5p8s: either use `minimum/maximum` or `floor/ceiling` for consistency

crates/iota-types/src/error.rs
```diff
@@ -539,6 +539,28 @@ pub enum IotaError {
         actual: starfish_config::AuthorityIndex,
     },
 
+    #[error(
+        "Attestation reports estimated_computation_cost = {actual} gas units, \
+         below the protocol minimum of {minimum} (`base_tx_cost_fixed`); \
+         an honest dry-run cannot bucketize below this floor"
```
roman1e2f5p8s: "bucketize" is related to `gas_rounding_step`, not `base_tx_cost_fixed`

```suggestion
         an honest dry-run cannot report below this floor"
```

crates/iota-types/src/error.rs
```diff
@@ -539,6 +539,28 @@ pub enum IotaError {
         actual: starfish_config::AuthorityIndex,
     },
 
+    #[error(
```
roman1e2f5p8s, line 542: these two errors, `AttestationCostBelowMinimum` and `AttestationCostAboveBudget`, should have their messages changed if/after attested `estimated_computation_cost` is in `NANOS` and is compared against `min(base_tx_cost_fixed, gas_rounding_step) * gas_price` for the minimum

crates/iota-types/src/executable_transaction.rs
vekkiokonio: > Is a new struct really needed here? Why can't we add an Option<Attestation> field to the existing VerifiedExecutableTransaction?

I would prefer this new struct rather than adding `Option<Attestation>` to `VerifiedExecutableTransaction`. That type is a type alias over the generic Envelope, which is shared by every envelope in the system. To make this change, I'd unnecessarily pollute every other envelope to change their serialized form.

> Plus, this type is only used in iota-core so perhaps this should be moved there.

I moved it to `transaction_manager.rs`. Happy to move if you'd prefer elsewhere.

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
vekkiokonio: I'd rather not add `TotalComputationCost`. These tests are agnostic to which mode produced the per-tx cost, so running them under `TotalComputationCost` would just duplicate `TotalGasBudget`. I've set `TotalComputationCost => unreachable!()` in these tests (needed only for match exhaustiveness).

Please note that `TotalComputationCost` mode is covered by `test_get_estimated_execution_duration_total_computation_cost_mode`

crates/iota-core/src/authority/shared_object_congestion_tracker.rs
vekkiokonio: I'd skip `TotalComputationCost` mode here. See my previous reply: https://github.com/iotaledger/iota/pull/11574#discussion_r3389695190

crates/iota-core/src/authority.rs
vekkiokonio: I agree the naming is confusing. I renamed the field to computation_units (https://github.com/iotaledger/iota/pull/11574/changes/8c7baa67a881a0bbdae872ef832b3e267fe566eb). I prefer to keep computation units rather than raw NANOS: the field exists to feed the sequencer's scheduling estimate, which is a work quantity, so computation units makes the attested value self-describing and matches the field's name and purpose.

crates/iota-types/src/attestation.rs
vekkiokonio: I renamed the field to `computation_units` (https://github.com/iotaledger/iota/pull/11574#discussion_r3396357359)

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Correct! I renamed the sequencer mode to `TotalComputationUnits` (https://github.com/iotaledger/iota/pull/11574/changes/8c7baa67a881a0bbdae872ef832b3e267fe566eb)

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11574/changes/28dc7fae2441e58b6b017808c03fa3c574c7b531

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: Fixed in https://github.com/iotaledger/iota/pull/11574/changes/28dc7fae2441e58b6b017808c03fa3c574c7b531

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: All validators would need to perform `checked_div` during sequencing, so no real gain by switching to `NANOS`

crates/iota-types/src/error.rs
vekkiokonio: The message has been modified

crates/iota-types/src/error.rs
vekkiokonio: Fixed. Using `minimum/maximum`.

cyberphysic4l (approved):

crates/iota-core/src/authority.rs
```diff
@@ -1787,6 +1794,32 @@ impl AuthorityState {
...
+                    // Explicit attestations are rejected at post-consensus
+                    // (Check #3); reaching here means a protocol-level bug.
+                    return Err(IotaError::UnsupportedFeature {
+                        error: "Explicit attestation reached prepare_certificate".into(),
+                    });
+                }
+            };
+            if epoch_store
+                .signature_verifier
+                .verify_tx(certificate.data())
```
bingyanglin: This `verify_tx` is on the execution path, so a bad signature returns `Err` and `execution_driver` turns it into `fatal!` — every honest validator crashes once attestation is enabled. The V2 user signature isn't checked anywhere earlier (`consensus_validator` skips it for the `Validator` arm, post-consensus `Check #3` only checks attestor + cost), so a Byzantine block author can sequence a bad-sig `UserTransactionV2` and halt the network. Maybe verify it in post-consensus `Check #3` and drop the tx there, instead of failing in `prepare_certificate` where the error path goes to `fatal!`?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -325,6 +356,34 @@ pub async fn validate_and_resolve_conflicts(
                 keep[i] = false;
                 continue;
             }
+        } else {
+            let verified_tx = VerifiedTransaction::new_from_verified(transaction.clone());
+            if let Err(e) = authority_state
+                .check_coin_deny_list_for_attested_tx(&verified_tx, epoch_store.epoch())
```
bingyanglin, line 380: `check_coin_deny_list_v1` reads the deny list with `cur_epoch=None` (latest value, live store), but execution uses `Some(epoch)`. Post-consensus validation isn't serialized after the previous commit's execution, so if a deny-list update and a regulated transfer land in different commits, validators at different execution progress make different drop decisions here — divergent checkpoints, so a fork. Maybe pass `Some(epoch_store.epoch())` to match execution, or wait for the consistent-view mechanism (the TODO in `authority_per_epoch_store.rs`)?

crates/iota-protocol-config/src/lib.rs
```diff
@@ -524,8 +524,9 @@ impl ConsensusTransactionOrdering {
 pub enum PerObjectCongestionControlMode {
     #[default]
     None, // No congestion control.
-    TotalGasBudget, // Use txn gas budget as execution cost.
-    TotalTxCount,   // Use total txn count as execution cost.
+    TotalGasBudget,        // Use txn gas budget as execution cost.
+    TotalTxCount,          // Use total txn count as execution cost.
+    TotalComputationUnits, // Use attested computation units as execution cost.
```
bingyanglin, line 529: This new variant isn't wired to any protocol version — the only assignment (line 2338) hardcodes `TotalTxCount`, so flipping it to `TotalComputationUnits` later would change `get_estimated_execution_duration` and shared-object scheduling on every node regardless of protocol version. That's consensus-visible. Maybe gate the assignment on a protocol version (and `enable_validator_attestation`) so the whole fleet flips together?

crates/iota-core/src/post_consensus_validation.rs
roman1e2f5p8s: > All validators would need to perform checked_div during sequencing

true, but that integer division would be negligible compared to what each validator already does during sequencing: hashmap lookups, recursion, sorting, etc. note that it would be the only place in the code where `checked_div` happens. with the current CUs design, there are multiple `checked_div` calls, and they inconsistently fall back to either 0 or `u64::MAX` - fine, gas price cannot be 0.

> so no real gain by switching to `NANOS`

the real gain of using `NANOS` is that it is the ground truth value that dry-run effects give you. currently, we are planning to use CUs only in the sequencer, but using `NANOS` would be a better future-proof design, and all of those comparisons would become more straightforward than the current `checked_div` then compare.

nevertheless, I am okay with the current state: proper `computation_units` name + CUs units.

Thoralf-M (approved):
Approving the iota-types changes

crates/iota-core/src/authority.rs
piotrm50: that's a good one! also, if we don't verify signatures right after consensus, a byznatine validator can cancel any transaction by inserting the same transaction earlier in his block but with incorrect signature - if it's earlier in the block then it will win whiteflag conflict resolution and fail right pre-execution.

vekkiokonio: You're right, that's a valid concern. I went with verifying the user signature pre-consensus in the block verifier exactly as `UserTransactionV1` does, rather than post-consensus. That rejects the bad-sig tx with the block before it's ever sequenced, so it can't win conflict resolution and the `verify_tx` in `prepare_certificate` is removed entirely (no more `fatal!` path). Fixed in https://github.com/iotaledger/iota/pull/11574/changes/c1095610bc5308bebe8f215950355298d4653535

BTW this needs a bigger refactor beyond the signature: gas and Move-authentication checks can also fail first at `prepare_certificate` and hit the same `fatal!`. Either we re-run those checks post-consensus, or we keep them pre-execution and mark failing txs as `Cancelled`. I've opened a separate [issue](https://github.com/iotaledger/iota-private/issues/438) to track that.

piotrm50: pre-consensus only works for honest validators. a malicious validator can still cancel any transaction by introducing a conflict with invalid signature of an honest transaction that will win the conflict resolution post-consensus and eventually fail the node pre-execution.

vekkiokonio: Invalid signatures are rejected by consensus, so a transaction with a bad signature cannot be committed. This means the pre-consensus check in `IotaTxValidator` (`validate_transactions`) is enough. This is the same for V1 and V2: both verify the user signature only in `IotaTxValidator`, and neither re-checks it post-consensus.

crates/iota-core/src/post_consensus_validation.rs
vekkiokonio: This function will be revisited, but I currently fixed in https://github.com/iotaledger/iota/pull/11574/changes/47bd3d8fca273e80320effdff0c53a592631bb42

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Yes, you're right! I tied the mode to `enable_validator_attestation`:
```
if self.enable_validator_attestation() {
    return PerObjectCongestionControlMode::TotalComputationUnits;
}
self.feature_flags.per_object_congestion_control_mode
```

Since attestation is protocol-version-gated, the mode flips with it across the whole fleet, with no independent assignment to drift.  Interim until attestation gets its own version arm — added a TODO to set the stored flag there and revert this to a plain getter.

bingyanglin (approved):

roman1e2f5p8s (approved):

PR #11582, opened 2026-05-20
feat(move): Add the view attribute to the language

### LINKED ISSUE #6548: feat: move view functions
Enable developers to create view functions in Move that directly read on-chain state. Writing these in Move should be simpler and more reliable than handling reads on the client side, making them the preferred method for querying app-specific data. However, this doesn't replace the need for off-chain indexing, as not all data relationships exist on-chain. The goal is providing an intuitive interface for developers to write and run these view functions.

miker83z:
# Description of change

Adds first-class **view functions** to Move on IOTA via a new `#[view]` function attribute, enforced end-to-end across the compiler, the bytecode verifier, and package metadata.

A `#[view]` function is a function a package author explicitly marks as a read-only, side-effect-free entry point intended to be called off-chain (e.g. for inspection / dev-inspect style calls). Marking a function `#[view]` constrains its signature so it is safe to call as a view:

- it must be `public`;
- it must return at least one value;
- it must not take mutable references or `mut` parameters;
- neither its parameters nor its return types may contain values that are
  unsafe to pass/return by value as a view (e.g. objects with `key`).

The feature is enforced at three layers so the guarantees survive past the source compiler:

- **Compiler (`move-compiler`, IOTA flavor):** the `#[view]` attribute is   registered (function position only) and the typing pass validates the   signature constraints above, emitting precise diagnostics on violations.
- **Linter:** a new IOTA-mode lint warns when a function already satisfies the   view constraints but is not annotated `#[view]`, and suggests adding the   annotation. The lint shares its signature check with the typing pass.
- **Bytecode verifier (`iota-verifier`):** re-verifies the same constraints on   published bytecode, independent of how the package was compiled, and records   view functions in the module metadata (accepted only while the protocol flag below is enabled; otherwise view-function metadata is rejected at publish).

On-chain, view functions are surfaced through **package metadata**: `package_metadata.move` is split into `package_metadata` and `module_metadata`, and a package's view-function metadata is published as part of its module metadata. This is gated behind a new protocol feature flag, `package_metadata_with_dynamic_module_metadata`, which stores `ModuleMetadata` as a dynamic field and requires `publish_package_metadata`. The same flag also gates the `View` attribute in module bytecode itself (#12050): while it is disabled, builds do not emit view-function metadata (modules carry V1 runtime metadata, authenticator attributes only, with a build warning for any dropped `#[view]`) and publish verification rejects it, so a package carrying view-function metadata cannot land on chain before the feature activates. The flag is enabled at the IIP8 protocol version on non-(testnet/mainnet) chains for now.

View functions also participate in **package upgrade compatibility**, so an upgrade cannot silently drop or break a previously published view function.

fixes #6548

Added coverage includes: compiler typing tests (`move_2024/typing/view_function/{valid,invalid}`), linter tests (`move_2024/linter/view_function/{invalid,suggestions}`), verifier tests (`iota-verifier/tests/view_functions/{valid,invalid}.mvir`), framework `package_metadata` / `module_metadata` tests, and transactional upgrade tests (`view_base` / `view_removed`).

### Release Notes

- [x] Protocol: Adds the `package_metadata_with_dynamic_module_metadata` feature
  flag; package metadata can now store module metadata, including view functions, as a dynamic field.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


Thoralf-M:
Need to consolidate the release notes of the individual PRs in this PR description?

kodemartin (approved):

piotrm50 (approved):

lollobene (approved):

lollobene:
> Need to consolidate the release notes of the individual PRs in this PR description?

Not needed cause inner commits already contain the descriptions

Thoralf-M:
> > Need to consolidate the release notes of the individual PRs in this PR description?
> 
> Not needed cause inner commits already contain the descriptions

Ah right, if it doesn't get squashed it's fine. I was just confused because this PR here also has a release notes entry

Thoralf-M (approved):

bingyanglin (approved):
Approved `iota-core/src/unit_tests`, `iota-protocol-config/src`, `iota-swarm-config/tests/snapshots`, and `iota-types/src`

PR #11591, opened 2026-05-20
feat(graphql): migrate Loader<HistoricalKey> to backward-diff

### LINKED ISSUE #11584: [graphql]: Replace `objects_history` in `Loader<HistoricalKey>`
Subissue of #11580 (R1 in [the audit](https://github.com/iotaledger/iota/issues/11580)).

**Location**: `crates/iota-graphql-rpc/src/types/object.rs:1532-1602`

**What it reads**: a specific `(object_id, object_version)` historical state of an object (full row content).

**Backs GraphQL**: `Object(address, version: V)` lookup. `Loader<OptimisticKey>` (`object.rs:1654`) falls back into this loader for misses, so it migrates transitively.

**Current tables**: `objects_history ⨝ objects_version`.

**Replacement direction**:

- Keep `objects_version` as the existence oracle and source of `cp_sequence_number` (the create-checkpoint of the version, used by the existing in-memory `checkpoint_viewed_at` filter).
- For each `(object_id, object_version)`:
  - `LEFT JOIN checkpointed_objects` for the row's current content (used in preference to `objects` because backward-history entries are written *before* `checkpointed_objects` is bumped — `objects` has no such ordering guarantee, and a query for a just-superseded version could otherwise see "neither live nor in history" mid-write).
  - `LEFT JOIN objects_backward_history` filtered to `object_status = ACTIVE`, keyed on `(object_id, object_version)`. Non-ACTIVE rows (`WrappedOrDeleted`, `NotYetCreated`) are excluded because they use placeholder versions (`lamport - 1`) which we don't want to return.
  - `object_status` taken via `COALESCE(co.object_status, bh.object_status, WrappedOrDeleted)` — whichever source matched, falling back to `WrappedOrDeleted` in case result is only found in `objects_version`
- `WHERE` clause gates the synthetic-`WrappedOrDeleted` fallback by the backward-history watermark (`min_available_cp` from `watermarks`): version below backward-history watermark are dropped so the caller sees `None` (data unavailable, pruned) instead of a false tombstone. Mirrors the established pattern in `backward_view::historical::tombstones_from_objects_version`.

tomxey:
# Description of change

Replace the `objects_history ⨝ objects_version` read in `Loader<HistoricalKey>` with a `checkpointed_objects ∪ objects_backward_history` merge, joined onto `objects_version`.

### Query logic

For each requested `(object_id, object_version)`:

- **`objects_version`** is the existence oracle and source of `cp_sequence_number` (used by the existing `checkpoint_viewed_at` filter applied in Rust).
- **`LEFT JOIN checkpointed_objects`** picks up the row if it is the object's current state (may itself be a `WrappedOrDeleted` tombstone, since `persist_checkpointed_objects` writes deletions too).
- **`LEFT JOIN objects_backward_history WHERE object_status = ACTIVE`** picks it up if it is a superseded ACTIVE version. Non-ACTIVE rows are excluded because they use placeholder versions that we don't want to return to the user.
- `object_status` is `COALESCE(co.object_status, bh.object_status, WrappedOrDeleted)` — the third arm covers the wrapped/deleted objects missing from history (at correct versions), mirroring `backward_view::historical::tombstones_from_objects_version`.
- The `WHERE` clause gates that third arm by the `objects_backward_history` watermark: if neither table matched and the version is older than `min_available_cp`, the row is dropped so the loader returns `None` (data pruned) rather than a false tombstone.

fixes #11584

- `cargo clippy --all-targets --all-features`, `cargo nextest run -p iota-graphql-e2e-tests --features pg_integration`.
- `EXPLAIN (ANALYZE, BUFFERS)` on staging (3.1M `checkpointed_objects`, 1.6M `objects_backward_history`, 119M `objects_version`): all queries execute fast and use indexes

### DIFF
--- a/crates/iota-graphql-rpc/src/types/object.rs
+++ b/crates/iota-graphql-rpc/src/types/object.rs
@@ -1553,7 +1553,7 @@ impl Loader<HistoricalKey> for Db {
         // be a `WrappedOrDeleted` tombstone) or `objects_backward_history`.
         //
         // Only `ACTIVE` rows from `objects_backward_history` are joined
-        // (`bh.object_status = {active}`) because they are the only rows that
+        // (`bh.object_status = $3`) because they are the only rows that
         // carry the real `(object_id, object_version)`. The other two statuses
         // use placeholder versions, so matching on them by `(object_id,
         // object_version)` could by chance return incorrect row (with
@@ -1575,12 +1575,11 @@ impl Loader<HistoricalKey> for Db {
         // `cp_sequence_number` (the checkpoint at which the version was
         // committed), used by the in-memory `checkpoint_viewed_at` filter
         // below.
-        let sql = format!(
-            "SELECT \
+        let sql = "SELECT \
                 v.object_id, \
                 v.object_version, \
                 v.cp_sequence_number AS checkpoint_sequence_number, \
-                COALESCE(co.object_status, bh.object_status, {wrapped_or_deleted}::int2) AS object_status, \
+                COALESCE(co.object_status, bh.object_status, $4::int2) AS object_status, \
                 COALESCE(co.object_digest, bh.object_digest) AS object_digest, \
                 COALESCE(co.owner_type, bh.owner_type) AS owner_type, \
                 COALESCE(co.owner_id, bh.owner_id) AS owner_id, \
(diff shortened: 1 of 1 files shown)


sergiupopescu199 (approved):
lgtm 🦫

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1534,37 +1534,88 @@ impl Loader<HistoricalKey> for Db {
...
+            LEFT JOIN objects_backward_history bh \
+                   ON bh.object_id = v.object_id \
+                  AND bh.object_version = v.object_version \
+                  AND bh.object_status = {active} \
+            WHERE co.object_id IS NOT NULL \
+               OR bh.object_id IS NOT NULL \
+               OR v.cp_sequence_number >= COALESCE(\
+                      (SELECT min_available_cp FROM watermarks \
+                       WHERE entity = '{BACKWARD_HISTORY_WATERMARK_ENTITY}'), 0)"
+        );
```
kodemartin: I suggest binding all values (`wrapped_or_deleted`, `active` etc.) similarly to `ids`, `versions`

tomxey: done in 8d2cbde455

kodemartin (approved):

PR #11598, opened 2026-05-21
upstream(core): Port latest traffic controller

jkrvivian:
## Description of change

- Port commits:
  - [MystenLabs/sui@9006185](https://github.com/MystenLabs/sui/commit/9006185009afa2fe239fec771a7d067437814a42) — [TrafficControl] Enable DOS protection by default (dryRun) (#22143)
  - [MystenLabs/sui@13950175](https://github.com/MystenLabs/sui/commit/13950175c3) — [Traffic Control] Fix blocked metric (#20590)
  - [MystenLabs/sui@ef1bb1e0](https://github.com/MystenLabs/sui/commit/ef1bb1e0d3) — [traffic controller] show dry run status via metric (#22363)
  - [MystenLabs/sui@264d7664](https://github.com/MystenLabs/sui/commit/264d7664b3) — [Traffic Control] Add dynamic config via admin api (#20887)
  - [MystenLabs/sui@6427b4d6](https://github.com/MystenLabs/sui/commit/6427b4d628) — [traffic controller] fix race condition in blocklist metric dec() call (#22771)
  - [MystenLabs/sui@4e30d3b3](https://github.com/MystenLabs/sui/commit/4e30d3b35b) — [traffic controller] downgrade shutdown error case to warn! log (#22782)
  - [MystenLabs/sui@f5d1b6c9](https://github.com/MystenLabs/sui/commit/f5d1b6c935) — [dos protection] increase the default spam threshold (#23301)
- Description:

Ports a series of upstream traffic-control improvements:

- **Enable DoS protection by default (dryRun)** (#22143): Turns the traffic controller on in dry-run mode by default. No traffic is actually blocked — only tallying and metrics collection.
- **Fix blocked metric** (#20590): Corrects the blocked-requests metric so it no longer increments in dry-run mode, removing the false impression that requests were being blocked.
- **Show dry run status via metric** (#22363): Adds a metric so operators can discern whether dry-run is enabled or disabled on a given node.
- **Add dynamic config via admin api** (#20887): Adds an admin API endpoint to hot-reconfigure the traffic control policy without clearing blocklists or the count-min sketch. Only works on nodes already running traffic control with a non-allowlist policy.
- **Fix race condition in blocklist metric dec() call** (#22771): Fixes a race where `check_and_clear_blocklist` could be called in parallel for the same IP, causing the blocklist length metric to be decremented multiple times for a single removal (sometimes producing negative values).
- **Downgrade shutdown error case to warn! log** (#22782): Downgrades a noisy error log to `warn!` when tally processing fails during shutdown, since it isn't actionable.
- **Increase the default spam threshold** (#23301): Raises the default spam threshold in `default_dos_protection_policy()` to better fit observed traffic patterns.

Part of #11643

## Release notes

- [x] Nodes (Validators and Full nodes): 
  - DoS protection is enabled in dryRun mode by default for validators. 
  - A new admin API endpoint allows hot-reconfiguration of the traffic control policy without clearing blocklists. 
  - A new metric exposes whether dry-run is enabled, and the blocked-requests / blocklist-length metrics are now accurate.

### DIFF
--- a/crates/iota-core/src/authority_server.rs
+++ b/crates/iota-core/src/authority_server.rs
@@ -41,8 +41,9 @@ use iota_types::{
 };
 use nonempty::{NonEmpty, nonempty};
 use prometheus::{
-    Histogram, IntCounter, IntCounterVec, Registry, register_histogram_with_registry,
-    register_int_counter_vec_with_registry, register_int_counter_with_registry,
+    Gauge, Histogram, IntCounter, IntCounterVec, Registry, register_gauge_with_registry,
+    register_histogram_with_registry, register_int_counter_vec_with_registry,
+    register_int_counter_with_registry,
 };
 use tap::TapFallible;
 use tonic::{
@@ -191,6 +192,7 @@ pub struct ValidatorServiceMetrics {
     forwarded_header_invalid: IntCounter,
     forwarded_header_not_included: IntCounter,
     client_id_source_config_mismatch: IntCounter,
+    x_forwarded_for_num_hops: Gauge,
 }
 
 impl ValidatorServiceMetrics {
@@ -340,6 +342,12 @@ impl ValidatorServiceMetrics {
             client_id_source_config_mismatch: register_int_counter_with_registry!(
                 "validator_service_client_id_source_config_mismatch",
                 "Number of times detected that client id source config doesn't agree with x-forwarded-for header",
+                registry,
+            )
+            .unwrap(),
+            x_forwarded_for_num_hops: register_gauge_with_registry!(
+                "validator_service_x_forwarded_for_num_hops",
+                "Number of hops in x-forwarded-for header",
                 registry,
             )
             .unwrap(),
@@ -1020,6 +1028,17 @@ impl ValidatorService {
         request: &tonic::Request<T>,
         source: &ClientIdSource,
     ) -> Option<IpAddr> {
+        let forwarded_header = request.metadata().get_all("x-forwarded-for").iter().next();
+
+        if let Some(header) = forwarded_header {
+            let num_hops = header
+                .to_str()
+                .map(|h| h.split(',').count().saturating_sub(1))
+                .unwrap_or(0);
+
+            self.metrics.x_forwarded_for_num_hops.set(num_hops as f64);
+        }
+
         match source {
             ClientIdSource::SocketAddr => {
                 let socket_addr: Option<SocketAddr> = request.remote_addr();
--- a/crates/iota-core/src/authority_server.rs
+++ b/crates/iota-core/src/authority_server.rs
@@ -36,7 +36,7 @@ use iota_types::{
         TransactionInfoResponse,
     },
     multiaddr::Multiaddr,
-    traffic_control::{ClientIdSource, PolicyConfig, RemoteFirewallConfig, Weight},
+    traffic_control::{ClientIdSource, Weight},
     transaction::*,
 };
 use nonempty::{NonEmpty, nonempty};
@@ -59,9 +59,7 @@ use crate::{
         ConnectionMonitorStatusForTests, ConsensusAdapter, ConsensusAdapterMetrics,
     },
     starfish_adapter::LazyStarfishClient,
-    traffic_controller::{
-        TrafficController, metrics::TrafficControllerMetrics, parse_ip, policies::TrafficTally,
-    },
+    traffic_controller::{TrafficController, parse_ip, policies::TrafficTally},
 };
 
 #[cfg(test)]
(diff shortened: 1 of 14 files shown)


crates/iota-types/src/traffic_control.rs
```diff
@@ -278,6 +298,30 @@ impl Default for PolicyConfig {
     }
 }
 
+impl PolicyConfig {
+    pub fn default_dos_protection_policy() -> PolicyConfig {
+        PolicyConfig {
```
Alex6323: Consistency NIT: also use `Self` here like we do for the `Default` impl?

piotrm50 (approved):

crates/iota-types/src/traffic_control.rs
```diff
@@ -278,6 +298,30 @@ impl Default for PolicyConfig {
     }
 }
 
+impl PolicyConfig {
+    pub fn default_dos_protection_policy() -> Self {
+        PolicyConfig {
```
Alex6323: sorry, I meant this as well 🙈

jkrvivian: Oh! I missed that. :trollface:

bingyanglin (approved):

semenov-vladyslav (approved):

kodemartin (approved):

cyberphysic4l (approved):

crates/iota-node/src/lib.rs
```diff
@@ -2339,11 +2338,12 @@ pub async fn build_http_server(
     let mut router = axum::Router::new();
 
     let json_rpc_router = {
+        let traffic_controller = state.traffic_controller.clone();
         let mut server = JsonRpcServerBuilder::new(
             env!("CARGO_PKG_VERSION"),
             prometheus_registry,
+            traffic_controller,
```
muXxer, line 2345: We need to implement the traffic controller also in the gRPC API. I will create an issue.

crates/iota-core/src/traffic_controller/mod.rs
```diff
@@ -84,78 +94,203 @@ impl Debug for TrafficController {
...
+            error_threshold,
+            spam_threshold,
+            dry_run,
+        } = params;
+        if let Some(error_threshold) = error_threshold {
+            self.metrics
+                .error_client_threshold
+                .set(error_threshold as i64);
+            Self::update_policy_threshold(
+                self.error_policy.as_ref().unwrap(),
```
muXxer, line 241: `unwrap()` will panic if the controller is in `Acl::Allowlist` mode (where spam_policy/error_policy are None). `reconfigure_traffic_control` is reachable from the admin HTTP endpoint, so a single curl on an allowlist-configured node crashes the process. Return `IotaError::InvalidAdminRequest` instead.

crates/iota-core/src/traffic_controller/mod.rs
```diff
@@ -84,78 +94,203 @@ impl Debug for TrafficController {
...
+                dry_run,
+            )
+            .await?;
         }
+        if let Some(spam_threshold) = spam_threshold {
+            self.metrics
+                .spam_client_threshold
+                .set(spam_threshold as i64);
+            Self::update_policy_threshold(
+                self.spam_policy.as_ref().unwrap(),
```
muXxer, line 252: `unwrap()` will panic if the controller is in `Acl::Allowlist` mode (where spam_policy/error_policy are None). `reconfigure_traffic_control` is reachable from the admin HTTP endpoint, so a single curl on an allowlist-configured node crashes the process. Return `IotaError::InvalidAdminRequest` instead.

crates/iota-core/src/traffic_controller/mod.rs
```diff
@@ -205,27 +332,34 @@ impl TrafficController {
...
                 Ok(_) => {}
             }
+        } else {
+            warn!("TrafficController not yet accepting tally requests.");
         }
     }
 
     /// Handle check with dry-run mode considered
     pub async fn check(&self, client: &Option<IpAddr>, proxied_client: &Option<IpAddr>) -> bool {
+        let policy_config = { self.policy_config.read().await.clone() };
```
muXxer, line 346: I wonder if we should rather store the `dry_run` as an atomic bool, and write it in `admin_reconfigure`.

crates/iota-e2e-tests/tests/traffic_control_tests.rs
muXxer, line 1: Can we add a small test that tests the allowlist after the `reconfigure_traffic_control` ?

crates/iota-core/src/authority_server.rs
```diff
@@ -343,6 +343,12 @@ impl ValidatorServiceMetrics {
                 registry,
             )
             .unwrap(),
+            x_forwarded_for_num_hops: register_gauge_with_registry!(
```
muXxer, line 346: A gauge is a bit weird, no? Shouldn't it be rather a histogram?

crates/iota-core/src/traffic_controller/mod.rs
```diff
@@ -40,13 +47,13 @@ pub const DEFAULT_DRAIN_TIMEOUT_SECS: u64 = 300;
...
 #[derive(Clone)]
-struct Blocklists {
+pub struct Blocklists {
     clients: Blocklist,
     proxied_clients: Blocklist,
 }
 
 #[derive(Clone)]
-enum Acl {
+pub enum Acl {
```
muXxer, line 56: Was it necessary to make `Acl` and `Blocklists` public?

muXxer (dismissed):

PR #11606, opened 2026-05-21
feat(iota-data-ingestion, iota-kvstore): allow backfilling for `iota-kv-store` worker type.

### LINKED ISSUE #11328: [data-ingestion]: support backfilling in the kv worker
## Description

We want to enable backfilling, after making changes in the schema of BigTable as part of #11327 .

sergiupopescu199:
# Description of change

This PR introduces two primary improvements to the `iota-kvstore` & `iota-data-ingestion`'s `iota-kv-store` worker type ingestion pipeline:

- **Configurable Table Selection**: Added the ability to specify a subset of tables for ingestion. This allows for targeted historical backfilling and reduces unnecessary write operations.

- **Flexible Checkpoint Resumption**: Fixed logic handling to allow starting the ingestion pipeline from a specific checkpoint (e.g., resuming from a mid-stream checkpoint rather than always defaulting to 0 or the latest watermark). This is a critical requirement for historical backfills and lays a clean foundation for other worker types to support custom start/end ranges with minimal refactoring.

fixes #11328 

- Tested manually the correctness of the new patch. Started a bigtable instance, for more information consult the `README.md` in the `iota-kvstore` crate.
- Started locally the `local-kv-storage` by using the `bigtable.yaml` config from the `iota/dev-tool/iota-data-ingestion/kv-store/config`.

### Validation Steps:

- **Checkpoint Resumption**: Verified that setting `start-checkpoint` allows the pipeline to skip genesis and start at a custom sequence.

    - Result: Configured `start-checkpoint: 1` and `end-checkpoint: 1499`. Confirmed the pipeline stopped correctly at 1500 and the watermark updated accordingly.

- **Table Filtering**: Verified that the tables configuration correctly limits the scope of ingested data.

    - Result: Ran ingestion with **objects** excluded from the tables list. Confirmed that while other tables populated correctly, the objects table remained empty.

>[!NOTE]

### DIFF
--- a/crates/iota-data-ingestion/src/main.rs
+++ b/crates/iota-data-ingestion/src/main.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{env, path::PathBuf, time::Duration};
+use std::{collections::HashSet, env, path::PathBuf, time::Duration};
 
 use anyhow::{Result, anyhow};
 use iota_data_ingestion::{
@@ -15,7 +15,7 @@ use iota_data_ingestion_core::{
     reader::v2::{CheckpointReaderConfig, RemoteUrl},
 };
 use iota_grpc_client::Client;
-use iota_kvstore::{BigTableClient, KvWorker};
+use iota_kvstore::{BigTableClient, BigtableTable, KvWorker};
 use iota_types::messages_checkpoint::CheckpointSequenceNumber;
 use prometheus::Registry;
 use serde::{Deserialize, Serialize};
@@ -39,6 +39,8 @@ struct BigTableTaskConfig {
     timeout_secs: usize,
     #[serde(skip_serializing_if = "Option::is_none")]
     emulator_host: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    tables: Option<HashSet<BigtableTable>>,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
@@ -232,6 +234,10 @@ async fn main() -> Result<()> {
                     Default::default(),
                 );
                 executor.register(worker_pool).await?;
+
+                if let Some(end_checkpoint) = config.end_checkpoint {
+                    executor.with_ingestion_limit(IngestionLimit::MaxCheckpoint(end_checkpoint));
+                }
             }
             Task::Kv(kv_config) => {
                 let worker_pool = WorkerPool::new(
@@ -257,10 +263,6 @@ async fn main() -> Result<()> {
                 executor.register(worker_pool).await?;
             }
         };
-    }
-
-    if let Some(end_checkpoint) = config.end_checkpoint {
-        executor.with_ingestion_limit(IngestionLimit::MaxCheckpoint(end_checkpoint));
     }
 
     let reader_options = ReaderOptions {
--- a/crates/iota-data-ingestion/src/main.rs
+++ b/crates/iota-data-ingestion/src/main.rs
@@ -10,11 +10,13 @@ use iota_data_ingestion::{
     HistoricalWriterConfig, KVStoreTaskConfig, KVStoreWorker, RelayWorker, common,
 };
 use iota_data_ingestion_core::{
-    DataIngestionMetrics, FileProgressStore, IndexerExecutor, ReaderOptions, WorkerPool,
+    DataIngestionMetrics, FileProgressStore, IndexerExecutor, IngestionLimit, ReaderOptions,
+    WorkerPool,
     reader::v2::{CheckpointReaderConfig, RemoteUrl},
 };
 use iota_grpc_client::Client;
 use iota_kvstore::{BigTableClient, KvWorker};
+use iota_types::messages_checkpoint::CheckpointSequenceNumber;
 use prometheus::Registry;
 use serde::{Deserialize, Serialize};
 use tokio_util::sync::CancellationToken;
@@ -54,6 +56,12 @@ struct IndexerConfig {
     path: PathBuf,
     tasks: Vec<TaskConfig>,
     progress_store_path: String,
+    /// The sequence number of the checkpoint to start ingestion from.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    start_checkpoint: Option<CheckpointSequenceNumber>,
+    /// The inclusive sequence number of the checkpoint to end ingestion at.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    end_checkpoint: Option<CheckpointSequenceNumber>,
     #[serde(skip_serializing_if = "Option::is_none")]
     remote_store_url: Option<String>,
     #[serde(default = "default_remote_read_batch_size")]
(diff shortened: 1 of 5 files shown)


crates/iota-data-ingestion/src/main.rs
```diff
@@ -245,6 +261,10 @@ async fn main() -> Result<()> {
         };
     }
 
+    if let Some(end_checkpoint) = config.end_checkpoint {
```
tomxey: is it intended that end checkpoint is applied for every worker type, but start checkpoint only for "BigTableKv" ?
it's a bit assymetric

sergiupopescu199: You're right, I’ve addressed this in [this commit](https://github.com/iotaledger/iota/pull/11606/changes/d66c683e188a3b514d6053f0868b147f7f7380ec#diff-fdeac2861711ca299cad11a13a994895310d9d4de5691a9d2fabbf99c7ba93e4R237-R240) to ensure the logic is applied within the scope. 

I've also opened [#11622](https://github.com/iotaledger/iota/issues/11622) to track the global refactor for broader task support in the future.

crates/iota-kvstore/src/bigtable/worker.rs
```diff
@@ -2,19 +2,76 @@
...
 
+/// Represents the BigTable tables used by the KvWorker.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, strum::EnumIter)]
+#[serde(rename_all = "kebab-case")]
+#[non_exhaustive]
+pub enum BigtableTable {
+    Objects,
+    Transactions,
+    TransactionsBySeqNum,
+    Checkpoints,
```
kodemartin, line 50: I think it makes sense to provide full granularity by adding `CheckpointsByDigest` as well here.

crates/iota-kvstore/src/bigtable/worker.rs
```diff
@@ -2,19 +2,76 @@
...
 use crate::{BigTableClient, KeyValueStoreWriter, TransactionData};
 
+/// Represents the BigTable tables used by the KvWorker.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, strum::EnumIter)]
+#[serde(rename_all = "kebab-case")]
+#[non_exhaustive]
+pub enum BigtableTable {
+    Objects,
+    Transactions,
+    TransactionsBySeqNum,
```
kodemartin: I suggest expanding to the full table name: `TransactionsBySequenceNumber`

crates/iota-kvstore/src/bigtable/worker.rs
```diff
@@ -2,19 +2,76 @@
...
+    ///
+    /// # drop(worker);
+    /// # }
+    /// ```
+    pub fn new(client: BigTableClient, tables: impl Into<Option<HashSet<BigtableTable>>>) -> Self {
+        let tables = tables.into();
+        // handle the case where the user provided an empty set of tables.
+        let tables = tables.filter(|s| !s.is_empty());
+        let enabled_tables = BigtableTable::iter()
+            .filter(|t| tables.as_ref().is_none_or(|set| set.contains(t)))
```
kodemartin: This is a bit complicated as if `tables = Some(HashSet)` we already have the collection of the tables to enable. Iterating on all possible entries should be lazily doen if `tables.is_none`. So this could be simplified to something like

```rust
let enabled_tables = tables
    .map(|set| set.into_iter().collect::<Vec<_>>()
    .unwrap_or_else(|| BigtableTable::iter().collect());
```

But then we have to wonder why do we need to hold a `Vec<BigtableTable>` instead of `HashSet` in `KvWorker::enabled_tables`, which would simplify everything to

```rust
let enabled_tables = tables
    .unwrap_or_else(|| BigtableTable::iter().collect());
```

crates/iota-kvstore/src/bigtable/worker.rs
```diff
@@ -2,19 +2,76 @@
...
     pub client: BigTableClient,
+    /// The tables enabled for writing by this worker.
+    pub enabled_tables: Vec<BigtableTable>,
+}
+
+impl KvWorker {
+    /// Creates a new KvWorker with the specified BigTable client and enabled
+    /// tables the user wishes to write to.
+    ///
+    /// If `tables` is `None`, all available tables will be enabled.
```
kodemartin: This is a bit contradicting in the sense that we provide the option to the user to define enabled tables through `tables`.

`None` then would mean no table, and yet we inverse the semantics to enable all.

An idea to make the API clearer, would be to have distinct methods: a `new` method that enables all and a `new_selective(client: BigTableClient, tables: HashSet<Table>)` that enables only the passed ones and returns and panics or returns an error if the set is empty.

crates/iota-kvstore/src/bigtable/worker.rs
```diff
@@ -2,19 +2,76 @@
...
+use serde::{Deserialize, Serialize};
+use strum::IntoEnumIterator;
 
 use crate::{BigTableClient, KeyValueStoreWriter, TransactionData};
 
+/// Represents the BigTable tables used by the KvWorker.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, strum::EnumIter)]
+#[serde(rename_all = "kebab-case")]
+#[non_exhaustive]
+pub enum BigtableTable {
```
kodemartin: Do we need the `Bigtable` prefix? The case is also inconsistent with other names (e.g. `BigTableClient`).

crates/iota-kvstore/src/bigtable/worker.rs
sergiupopescu199: Sure thing, added: https://github.com/iotaledger/iota/pull/11606/commits/15b03128d136be7385b7bca7d1b4c98cecd6b586

crates/iota-kvstore/src/bigtable/worker.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11606/commits/15b03128d136be7385b7bca7d1b4c98cecd6b586

crates/iota-kvstore/src/bigtable/worker.rs
sergiupopescu199: I needed the vector to preserve the order of the enum variants, but I do agree that it's more complicated.

simplified: https://github.com/iotaledger/iota/pull/11606/commits/15b03128d136be7385b7bca7d1b4c98cecd6b586

crates/iota-kvstore/src/bigtable/worker.rs
sergiupopescu199: oh you're right, simplified: https://github.com/iotaledger/iota/pull/11606/commits/15b03128d136be7385b7bca7d1b4c98cecd6b586

crates/iota-kvstore/src/bigtable/worker.rs
sergiupopescu199: `Table` is a good alternative. fixed: https://github.com/iotaledger/iota/pull/11606/commits/15b03128d136be7385b7bca7d1b4c98cecd6b586

kodemartin (approved):
looks good 🎈

tomxey (approved):

kodemartin (approved):

tomxey (approved):

PR #11616, opened 2026-05-21
feat(move): Add package view functions metadata dynamic field

### LINKED ISSUE #11615: [move] View Functions Metadata as dynamic field
Explore how ViewFunctionsMetadata can be implemented as a dynamic field of PackageMetadataV1

lollobene:
# Description of change

Implements **view functions metadata** stored as a dynamic field of `PackageMetadataV1`, introducing a new per-module metadata layout.

### Framework (`iota-framework`)

- **New `iota::module_metadata` module**: a `ModuleMetadata` object acting as a per-module metadata container with inner dynamic fields. Its address is derived from the package storage ID and the module name (via the new public `derived_object::derive_address`), so each module gets a distinct, deterministic metadata object with no cross-package collisions.
- **New dynamic-field layout for `PackageMetadataV1`** (struct layout unchanged for on-chain compatibility). New-layout objects carry two dynamic fields:
  - `PackageMetadataVersionFieldName` → layout version (`2`) (**new**),
  - `ModulesMetadataFieldName` → `VecMap<ModuleName, ModuleMetadata>` (**new**).
- **Per-module metadata stored as a dynamic field of `ModuleMetadata`**:
  - `ModuleMetadataV1FieldName` → the authenticator metadata `ModuleMetadataV1` (**moved** from the `modules_metadata` field of `PackageMetadatav1`),
  - `ViewFunctionMetadataV1FieldName` → the names of the module's `#[view]` functions (**new**).
- New accessors `module_metadata`, `borrow_view_functions_metadata_v1`, `is_view_function_v1`; the legacy inline accessors are `#[deprecated]` and now resolve both layouts (version-aware).
- The metadata object is built by a new private constructor `create_package_metadata_v1_with_dynamic_metadata`, invoked by the adapter (see below), and frozen at creation.

### Execution layer (`iota-adapter`)

- Package-metadata creation moved into a new `programmable_transactions::package_metadata` module with two handlers:
  - **V1 (legacy)**: builds the inline-layout object natively, remains compatible with the old behavior.
  - **V1 with dynamic module metadata**: BCS-encodes the extracted `#[authenticator]`/`#[view]` attribute data and calls the framework constructor through a `System` Move call. New `ExecutionContext` helpers (`add_pure_input` / `num_inputs` / `truncate_inputs`) set up the arguments for the `execute_move_call` and clean them up afterwards.
- The handler is selected by the new protocol feature flag `package_metadata_with_dynamic_module_metadata` (enabled at protocol version 29; requires `publish_package_metadata`).

### Testing infrastructure

- New transactional-test-runner flag `--module-metadata-dynamic` to enable/disable the new layout.

fixes #11615


crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -6,23 +6,31 @@
...
+#[error(code = 2)]
+const EWrongInnerVersion: vector<u8> =
+    b"The provided package metadata has an unsupported inner version.";
 
-// === Structs ===
+// === Dynamic field keys ===
+public struct PackageMetadataV2Key has copy, drop, store {}
 
-/// Key type for deriving the package metadata object address
-public struct PackageMetadataKey has copy, drop, store {}
```
miker83z, line 25: This struct cannot be deleted for compatibility reasons

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -55,6 +79,111 @@ public struct AuthenticatorMetadataV1 has copy, drop, store {
     function_name: ascii::String,
     account_type: TypeName,
 }
+// ===  Constructors ===
+
+public(package) fun create_package_metadata_v1(
+    metadata_id: address,
```
miker83z: We still haven't pulled the derive id feature fully, but, I'd suggest to: 
- just create [such function](https://github.com/MystenLabs/sui/blob/286f8a7454a6af620e963dacefb833f3cd8118cb/crates/sui-framework/packages/sui-framework/sources/derived_object.move#L54) in the derived_object module 
- pass the package_id `ID` here
- then call `let id = derived_object::derive_address(package_id, PackageMetadataKey {})` in the body

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -997,65 +1145,140 @@ mod checked {
...
+                    modules_metadata_map.insert(module.name().to_string(), pending_module_metadata);
                 }
-                // End of PackageMetadataV1 specific
             }
         }
 
         // Only publish package metadata if there is at least one module with
         // relevant metadata
-        if !modules_metadata_map.is_empty() {
-            // Create the package metadata "special" object UID
```
miker83z, line 1039: please keep this version of the logic if the `package_metadata_v2` protocol flag is false. 

This because, we cannot modify the logic of transactions executed in the past. Even if the outcome is the same related to the PackageMetadataV1 object, in the new flow we are introducing gas usage. This difference would make transaction replay and recovering the chain fail.

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -958,20 +967,159 @@ mod checked {
...
+                    modules_metadata,
+                    storage_id,
+                    runtime_id,
+                    package_version,
+                    trace_builder_opt,
+                ),
+            }
+        }
+
+        fn publish_v1(
```
miker83z: This shouldn't be needed after this revert: https://github.com/iotaledger/iota/pull/11616/changes#r3348784363

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -958,20 +967,159 @@ mod checked {
...
+            context: &mut ExecutionContext<'_, '_, '_>,
+            modules_metadata: &BTreeMap<String, PendingModuleMetadata>,
+            storage_id: ObjectID,
+            runtime_id: ObjectID,
+            package_version: u64,
+            trace_builder_opt: &mut Option<MoveTraceBuilder>,
+        ) -> Result<(), ExecutionError> {
+            let (modules, auth_functions, type_names, _) =
+                package_metadata_constructor_args(modules_metadata);
+            let metadata_id = derive_package_metadata_v2_id(storage_id);
```
miker83z: Even though this function might not be needed (see https://github.com/iotaledger/iota/pull/11616/changes#r3348811864), I think it was wrong to derive the V2 here no?

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -958,20 +967,159 @@ mod checked {
         }
     }
 
+    #[derive(Copy, Clone)]
+    enum PackageMetadataHandler {
```
miker83z: Can we try to move these intermediary types into iota-types move_package, so that execution here is not filled with unrelated types?

miker83z: I mean PendingModuleMetadata as well

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -42,6 +50,22 @@ public struct PackageMetadataV1 has key {
     modules_metadata: VecMap<ascii::String, ModuleMetadataV1>,
 }
 
+/// Represents the metadata of a Move package. This includes information
+/// such as the storage ID, runtime ID, version, and metadata for the
+/// functions contained within the package.
+public struct PackageMetadataV2 has key {
```
miker83z: The whole point of this was to not use a new PackageMetadata version type. I don't think we need PackageMetadataV2. Even in create_package_metadata_v2 we can use PackageMetadataV1 and attach to it PackageMetadataV2Inner as df.

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -91,6 +235,16 @@ public fun modules_metadata_v1(
     self.modules_metadata.get(module_name)
```
miker83z: You can modify this body to check the version of the package metadata (you need a df `VersionField` or similar). If the version is not present, then use the code already present here. If the version is present and it's == 2 then use the code implemented in  modules_metadata_v2 below

crates/iota-framework/packages/iota-framework/sources/package_metadata_inner.move
```diff
@@ -0,0 +1,227 @@
...
+#[error(code = 0)]
+const EModuleMetadataNotFound: vector<u8> =
+    b"The requested module metadata was not found in the package metadata.";
+#[error(code = 1)]
+const EAuthenticatorMetadataNotFound: vector<u8> =
+    b"The requested authenticator metadata was not found in the module metadata.";
+// === Structs ===
+
+/// Represents the metadata of a Move package V2.
+public struct PackageMetadataV2Inner has copy, drop, store {
```
miker83z:
```suggestion
public struct PackageMetadataInnerV2 has copy, drop, store {
```

crates/iota-protocol-config/src/lib.rs
```diff
@@ -484,6 +485,11 @@ struct FeatureFlags {
     // If true, only sponsor Move authentication is performed pre-consensus.
     #[serde(skip_serializing_if = "is_false")]
     pre_consensus_sponsor_only_move_authentication: bool,
+
+    // If true, package metadata can be published as PackageMetadataV2 when the
+    // package requires the V2 layout. If false, PackageMetadataV1 is used.
+    #[serde(skip_serializing_if = "is_false")]
+    package_metadata_v2: bool,
```
miker83z:
```suggestion
    // If true, package metadata can be published with a dynamic field containing an inner representation of the package which can be versioned.    
// If false, PackageMetadataV1 is used with no dynamic fields.
    #[serde(skip_serializing_if = "is_false")]
    package_metadata_inner_v2: bool,
```

miker83z (changes requested):

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
```diff
@@ -958,20 +967,159 @@ mod checked {
...
+            context: &mut ExecutionContext<'_, '_, '_>,
+            modules_metadata: &BTreeMap<String, PendingModuleMetadata>,
+            storage_id: ObjectID,
+            runtime_id: ObjectID,
+            package_version: u64,
+            trace_builder_opt: &mut Option<MoveTraceBuilder>,
+        ) -> Result<(), ExecutionError> {
+            let (modules, auth_functions, type_names, view_function_names) =
+                package_metadata_constructor_args(modules_metadata);
+            let metadata_id = derive_package_metadata_id(storage_id);
```
Dkwcs: Is this correct derive here? Meanwhile publish_v1 uses derive_package_metadata_v2_id

crates/iota-framework/packages/iota-framework/sources/package_metadata_inner.move
```diff
@@ -0,0 +1,227 @@
...
+    &self.function_name
+}
+
+/// Return the account type of the authenticator represented by this metadata
+public fun account_type(self: &AuthenticatorMetadataV2): TypeName {
+    self.account_type
+}
+
+public fun view_functions_metadata_v1(self: &ModuleMetadataV2): vector<ViewFunctionMetadataV1> {
+    self.view_functions_metadata
```
Dkwcs: Just for consideration - would it be better to borrow the value, not getting it by value(copy)? As I see we use only for the reading operations

crates/iota-framework/packages/iota-framework/sources/package_metadata_inner.move
```diff
@@ -0,0 +1,227 @@
...
+}
+
+public fun view_function_metadata_v1(
+    self: &ModuleMetadataV2,
+    function_name: &ascii::String,
+): ViewFunctionMetadataV1 {
+    let view_functions_metadata = self.view_functions_metadata_v1();
+    let mut index = view_functions_metadata.find_index!(|m| m.function_name == *function_name);
+    assert!(index.is_some(), EAuthenticatorMetadataNotFound);
```
Dkwcs: Should it be `EViewFunctionMetadataNotFound`?

crates/iota-framework/packages/iota-framework/sources/package_metadata_inner.move
lollobene: Given that we drop `PackageMetadataV2`, what about calling it `PackageMetadataInnerV1` ?

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
lollobene: not really because we either return `ModuleMetadataV1` or `ModuleMetadataV2`

crates/iota-framework/packages/iota-framework/sources/module_metadata_dynamic.move
```diff
@@ -0,0 +1,156 @@
...
+}
+
+public struct ModuleMetadataDynamickey has copy, drop, store {}
+
+/// Represents view function metadata for a package.
+public struct ViewFunctionMetadataV1 has copy, drop, store {
+    function_name: ascii::String,
+}
+
+public struct ViewFunctionMetadaFieldName has copy, drop, store {}
```
miker83z:
```suggestion
public struct ViewFunctionMetadaV1FieldName has copy, drop, store {}
```
I'd then only store the `vector<ascii::String>` of functions as df value and have the functions `borrow_view_function_metadata_v1 -> vector<ascii::String>` and `is_view_function_v1 -> bool`

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -40,6 +52,12 @@ public struct PackageMetadataV1 has key {
     package_version: u64,
     // Handles to internal package modules
     modules_metadata: VecMap<ascii::String, ModuleMetadataV1>,
+    // package_view_function: VecMap<ascii::String, vector<ViewFunctionMetadataV1>>,
+}
+
+public struct ModuleMetadataKey has copy, drop, store {
+    module_name: ascii::String,
+    version: u8,
```
miker83z: remember to remove these lines

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -83,14 +257,72 @@ public fun try_get_modules_metadata_v1(
...
+    }
+}
+
+public fun modules_metadata(
+    self: &PackageMetadataV1,
+    module_name: &ascii::String,
+): &ModuleMetadataDynamic {
+    dynamic_field::borrow<
+        ModulesMetadataDynamicFieldName,
+        VecMap<ModuleName, ModuleMetadataDynamic>,
```
miker83z:
```suggestion
        VecMap<_, _>,
```
I think these should be inferred

crates/iota-framework/packages/iota-framework/sources/module_metadata_dynamic.move
```diff
@@ -0,0 +1,156 @@
...
+
+use iota::derived_object;
+use iota::dynamic_field as field;
+use std::ascii;
+
+// Attempted to destroy a non-empty module_metadata
+const EModuleMetadataNotEmpty: u64 = 0;
+const EViewFunctionMetadataNotFound: u64 = 1;
+
+public struct ModuleMetadataDynamic has key, store {
```
miker83z: Now that I see it written down, I think just using `ModuleMetadata` would be better 😅

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
miker83z, line 363: It should still work as it is.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -83,14 +257,72 @@ public fun try_get_modules_metadata_v1(
...
+    function_name: &ascii::String,
+): &AuthenticatorMetadataV1 {
+    let authenticator_metadata = self.borrow<
+        AuthenticatorMetadataFieldName,
+        vector<AuthenticatorMetadataV1>,
+    >(AuthenticatorMetadataFieldName {});
+    let mut index = authenticator_metadata.find_index!(|m| m.function_name() == *function_name);
+    assert!(index.is_some(), EAuthenticatorMetadataNotFound);
+    authenticator_metadata.borrow(index.extract())
+}
```
miker83z: We should do this only for `AuthenticatorMetadataV2`(once it will be designed). For now I think the best approach is to maintain the legacy `ModuleMetadataV1` as the holder of the `AuthenticatorMetadataV1` instances within `ModuleMetadataDynamic`. So here I'd suggest something like this:
```suggestion
public fun authenticator_function_metadata_v1(
    self: &ModuleMetadataDynamic,
    function_name: &ascii::String,
): &AuthenticatorMetadataV1 {
    let module_metadata_v1 = self.borrow<
        ModuleMetadataV1FieldName,
        ModuleMetadataV1,
    >(ModuleMetadataV1FieldName {});
    module_metadata_v1.authenticator_metadata_v1(function_name)
}
```

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -83,14 +257,72 @@ public fun try_get_modules_metadata_v1(
...
 ): &ModuleMetadataV1 {
-    assert!(self.modules_metadata.contains(module_name), EModuleMetadataNotFound);
-    self.modules_metadata.get(module_name)
+    if (
+        dynamic_field::exists_<ModulesMetadataDynamicFieldName>(
+            &self.id,
+            ModulesMetadataDynamicFieldName {},
+        )
+    ) {
+        abort (EWrongPackageVersion)
```
miker83z: this function should be set as deprecated, but still work. See https://github.com/iotaledger/iota/pull/11616/changes#r3372229431

So at this line you could borrow using the key `ModuleMetadataV1FieldName`

miker83z: Same for try_get_modules_metadata_v1

crates/iota-framework/packages/iota-framework/sources/package_metadata.move
```diff
@@ -83,14 +257,72 @@ public fun try_get_modules_metadata_v1(
...
 /// Aborts if the module is not found.
+/// TO BE DEPRECATED
 public fun modules_metadata_v1(
     self: &PackageMetadataV1,
     module_name: &ascii::String,
 ): &ModuleMetadataV1 {
-    assert!(self.modules_metadata.contains(module_name), EModuleMetadataNotFound);
-    self.modules_metadata.get(module_name)
+    if (
+        dynamic_field::exists_<ModulesMetadataDynamicFieldName>(
```
miker83z: check for the version found at `PackageMetadataVersionFieldName` instead, so that we don't have to assure that `ModulesMetadataDynamicFieldName` always exists for versions > 1.

lollobene: ok, a double check is needed anyway cause we need to check the existence of `PackageMetadataVersionFieldName` and then read its value

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
lollobene: i moved here the legacy V1 publish logic

iota-execution/latest/iota-adapter/src/programmable_transactions/execution.rs
lollobene: I would not mix the data model in `move_package.rs` with the handlers required by `execution.rs`. If the motivation is keeping `execution.rs` tidy,  I'd rather create another submodule called package_metadata.rs in `iota-adapter/programmable_transactions/`

lollobene: https://github.com/iotaledger/iota/pull/11616/changes/e876fd5e9bc11c136d2b810bfb9b7f4c3cd599bb

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,142 @@
...
+public(package) fun new(package_id: ID, module_name: ascii::String): ModuleMetadata {
+    let id_address = derived_object::derive_address(package_id, ModuleMetadatakey(module_name));
+    let id = object::new_uid_from_hash(id_address);
+    ModuleMetadata {
+        id,
+        size: 0,
+    }
+}
+
+public fun add_view_function_metadata_v1(
```
miker83z: all the functions in this module that makes use of `&mut ModuleMetadata` must be public(package) because no user should be allowed to create/modify a ModuleMetadata. Then, move all the public(package) functions below and the public ones above.

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/// Returns true iff the module_metadata is empty (if `length` returns `0`)
+public fun is_empty(module_metadata: &ModuleMetadata): bool {
+    module_metadata.size == 0
+}
+
+/// Destroys an empty module_metadata
+/// Aborts with `EModuleMetadataNotEmpty` if the module_metadata still contains values
+public fun destroy_empty(module_metadata: ModuleMetadata) {
```
miker83z: I think we don't need this

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,142 @@
...
+public fun borrow_view_functions_metadata_v1(
+    module_metadata: &ModuleMetadata,
+): &vector<ascii::String> {
+    borrow<ViewFunctionMetadataV1FieldName, vector<ascii::String>>(
+        module_metadata,
+        ViewFunctionMetadataV1FieldName {},
+    )
+}
+
+public fun borrow_view_function_metadata_v1(
```
miker83z: should be called `is_view_function_v1` and return a boolean.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+/// Represents metadata for an authenticator within the package.
+/// It includes the name of the authenticate function and the TypeName
+/// of the first parameter (i.e., the account object type).
+public struct AuthenticatorMetadataV1 has copy, drop, store {
+    function_name: ascii::String,
+    account_type: TypeName,
+}
+// ===  Constructors ===
+
+public(package) fun create_package_metadata_v1_with_dynamic_metadata(
```
miker83z: Should be private.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+
+    dynamic_field::add(
+        &mut package_metadata.id,
+        ModuleMetadataFieldName {},
+        modules_metadata,
+    );
+    transfer::freeze_object(package_metadata);
+}
+
+public(package) fun create_modules_metadata_v1(
```
miker83z: Should be private.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+        modules_metadata.insert(
+            module_name,
+            create_module_metadata_v1(authenticator_metadata),
+        );
+        i = i + 1;
+    };
+    modules_metadata
+}
+
+public(package) fun create_modules_metadata(
```
miker83z: Should be private.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+        );
+        module_metadata.add_view_function_metadata_v1(view_function_names[i]);
+
+        modules_metadata.insert(ModuleName(module_name), module_metadata);
+        i = i + 1;
+    };
+    modules_metadata
+}
+
+public(package) fun create_module_metadata_v1(
```
miker83z: I think we can avoid an additional function, we can simply pack in place.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+    authenticator_metadata: vector<AuthenticatorMetadataV1>,
+): ModuleMetadataV1 {
+    ModuleMetadataV1 {
+        authenticator_metadata,
+    }
+}
+
+public(package) fun create_authenticator_metadata_v1(
```
miker83z: I think we can avoid an additional function, we can simply pack in place.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+    /// the Storage ID of the first version of a package.
+    runtime_id: ID,
+    /// Version of the package represented by this metadata
+    package_version: u64,
+    // Handles to internal package modules
+    modules_metadata: VecMap<ascii::String, ModuleMetadataV1>,
+}
+
+/// Represents metadata associated with a module in the package.
+/// V1 includes only the authenticator functions information.
```
miker83z:
```suggestion
/// This is now deprecated in favor of iota::module_metadata::ModuleMetadata.
/// It represented the first version of the metadata associated with a module in the 
/// package and included only the authenticator functions information.
```

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+/// Key types for dynamic field keys
+public struct PackageMetadataVersionFieldName has copy, drop, store {}
+public struct ModuleMetadataV1FieldName has copy, drop, store {}
+public struct ModuleMetadataFieldName has copy, drop, store {}
+
+public struct ModuleName(ascii::String) has copy, drop, store;
+
+/// Represents the metadata of a Move package. This includes information
+/// such as the storage ID, runtime ID, version, and metadata for the
+/// functions contained within the package.
```
miker83z:
```suggestion
/// Represents the metadata of a Move package. This includes information
/// such as the storage ID, runtime ID, version. The modules_metadata field
/// is deprecated in favor of a dynamic field attached to this object that
/// stores a iota::module_metadata::ModuleMetadata instance.
```

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
miker83z: Also, move all private and public(package) functions below.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+}
+
+public(package) fun create_modules_metadata_v1(
+    modules: vector<ascii::String>,
+    auth_functions: vector<vector<ascii::String>>,
+    type_names: vector<vector<TypeName>>,
+): VecMap<ascii::String, ModuleMetadataV1> {
+    assert!(modules.length() == auth_functions.length());
+    assert!(modules.length() == type_names.length());
+    let mut modules_metadata = iota::vec_map::empty<ascii::String, ModuleMetadataV1>();
```
miker83z:
```suggestion
    let mut modules_metadata = vec_map::empty<ascii::String, ModuleMetadataV1>();
```
and in the other occurrencies as well. We are already within the framework, so iota seems redundant. Just import iota::vec_map::Self once.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+
+    let id_address = derived_object::derive_address(package_id, PackageMetadataKey {});
+    let id = object::new_uid_from_hash(id_address);
+
+    let mut package_metadata = PackageMetadataV1 {
+        id,
+        storage_id,
+        runtime_id,
+        package_version,
+        modules_metadata: iota::vec_map::empty<ascii::String, ModuleMetadataV1>(),
```
miker83z:
```suggestion
        modules_metadata: vec_map::empty(),
```

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+
+    dynamic_field::add(
+        &mut package_metadata.id,
+        PackageMetadataVersionFieldName {},
+        2,
+    );
+
+    dynamic_field::add(
+        &mut package_metadata.id,
+        ModuleMetadataFieldName {},
```
miker83z:
```suggestion
        ModulesMetadataFieldName {},
```

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+            option::none()
+        }
+    } else {
+        self.modules_metadata.try_get(module_name)
+    }
+}
+
+/// Borrow the module metadata list of the package represented by this metadata.
+/// Aborts if the module is not found.
+/// TO BE DEPRECATED
```
miker83z:
```suggestion
/// Legacy function to borrow the module metadata list of the package represented by this metadata.
/// Aborts if the module is not found.
#[deprecated]
```
you could also move it down the file

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+    metadata.runtime_id
+}
+
+/// Return the version of the package represented by this metadata
+public fun package_version(metadata: &PackageMetadataV1): u64 {
+    metadata.package_version
+}
+
+/// Safely get the module metadata list of the package represented by this metadata
+/// TO BE DEPRECATED
```
miker83z:
```suggestion
/// Legacy function to safely get the module metadata list of the package represented by this metadata
#[deprecated]
```
you could also move it down the file

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+        assert!(package_metadata_version == 2, EWrongPackageVersion);
+        let modules_metadata = self.modules_metadata(module_name);
+        modules_metadata.borrow(ModuleMetadataV1FieldName {})
+    } else {
+        assert!(self.modules_metadata.contains(module_name), EModuleMetadataNotFound);
+        self.modules_metadata.get(module_name)
+    }
+}
+
+public fun modules_metadata(
```
miker83z:
```suggestion
public fun module_metadata(
```
We had `modules_metadata_v1` but I guess it was a mistake...

miker83z: moreover, we also need the try_get_ version which returns an option

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+        ModuleMetadataFieldName {},
+    );
+    let name = ModuleName(*module_name);
+    assert!(modules_metadata.contains(&name), EModuleMetadataNotFound);
+    let idx = modules_metadata.get_idx(&name);
+    let (_, metadata) = modules_metadata.get_entry_by_idx(idx);
+    metadata
+}
+
+public fun view_function_metadata_v1(
```
miker83z: This is redundant

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+    self: &ModuleMetadata,
+    function_name: &ascii::String,
+): &ascii::String {
+    self.borrow_view_function_metadata_v1(function_name)
+}
+
+public fun authenticator_function_metadata_v1(
```
miker83z: we also need the try_get_ version

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+        ModuleMetadataV1FieldName,
+        ModuleMetadataV1,
+    >(ModuleMetadataV1FieldName {});
+    module_metadata_v1.authenticator_metadata_v1(function_name)
+}
+
+/// Safely get the `AuthenticatorMetadataV1` associated with the specified
+/// `function_name` within the module metadata.
+/// TO BE DEPRECATED?
+public fun try_get_authenticator_metadata_v1(
```
miker83z:
```suggestion
/// Legacy function to safely get the `AuthenticatorMetadataV1` associated with the specified
/// `function_name` within the module metadata.
#[deprecated]
public fun try_get_authenticator_metadata_v1(
```

miker83z: you could also move it down the file

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+): Option<AuthenticatorMetadataV1> {
+    self.authenticator_metadata.find_index!(|m| m.function_name == *function_name).and!(|index| {
+        option::some(self.authenticator_metadata[index])
+    })
+}
+
+/// Borrow the `AuthenticatorMetadataV1` associated with the specified
+/// `function_name`.
+/// Aborts if the authenticator metadata is not found for that function.
+/// TO BE DEPRECATED?
```
miker83z:
```suggestion
/// Legacy function to borrow the `AuthenticatorMetadataV1` associated with the specified
/// `function_name`.
/// Aborts if the authenticator metadata is not found for that function.
#[deprecated]
```

miker83z: This will trigger some warnings if you try to compile the framework. Check each occurrency, mostly:
- In `authenticator_function_metadata_v1` you can use `#[allow(deprecated_usage)]`
- In `account_abstraction::create_auth_function_ref_v1` you can modify the body to use `authenticator_function_metadata_v1`

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
miker83z: maybe, just for the authenticator V1 we need a function that takes PackageMetadata, module name and function name and internally calls `modules_metadata_v1()`. In this way, users such as `account_abstraction::create_auth_function_ref_v1` don't need to know that it is needed to call `modules_metadata_v1()` instead of `module_metadata()` and we handle the deprecation internally (in relation to this module).

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,436 @@
...
+
+#[test_only]
+public fun create_package_metadata_v1_with_dynamic_metadata_for_testing(
+    storage_id: ID,
+    modules: vector<ascii::String>,
+    auth_functions: vector<vector<ascii::String>>,
+    type_names: vector<vector<TypeName>>,
+    view_functions: vector<vector<ascii::String>>,
+): PackageMetadataV1 {
+    let modules_metadata = create_modules_metadata(
```
miker83z: now you can just call the `create_package_metadata_v1_with_dynamic_metadata` constructor directly.

iota-execution/latest/iota-adapter/src/programmable_transactions/package_metadata.rs
```diff
@@ -0,0 +1,399 @@
...
+            package_version: u64,
+            trace_builder_opt: &mut Option<MoveTraceBuilder>,
+        ) -> Result<(), ExecutionError> {
+            let (modules, auth_functions, type_names, view_function_names) =
+                package_metadata_constructor_args(modules_metadata);
+            // The Move constructor derives both the package metadata object address
+            // and the per-module metadata object addresses from the package storage
+            // ID, so we pass the storage ID as the package ID.
+            let args = vec![
+                to_bcs_argument(&ID::new(storage_id), "package metadata package ID")?,
```
miker83z: I think it's clearer to use the bcs::to_bytes inline here, instead of wrapping that up with a funciton

iota-execution/latest/iota-adapter/src/programmable_transactions/package_metadata.rs
```diff
@@ -0,0 +1,399 @@
...
+        args: Vec<Vec<u8>>,
+        trace_builder_opt: &mut Option<MoveTraceBuilder>,
+    ) -> Result<(), ExecutionError> {
+        let saved_linkage = context.linkage_view.steal_linkage();
+        let result = (|| {
+            let original_address = context.set_link_context(IOTA_FRAMEWORK_PACKAGE_ID)?;
+            let runtime_id =
+                ModuleId::new(original_address, PACKAGE_METADATA_MODULE_NAME.to_owned());
+            context
+                .execute_function_bypass_visibility(
```
miker83z: Here it is manually going through and forcing `execute_function_bypass_visibility`. I would rather try to reuse `execute_move_call::<Mode>` like the function `init_modules` is doing. 
This would allow us to have only 1 entry point (i.e., through `execute_move_call`) to the `execute_function_bypass_visibility` instead of having to support 2 after these changes.

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
lollobene: This can't be private otherwise we can't call it from a PTB

miker83z: Have you tried calling a private function from a system transaction? 😉

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
lollobene: we can't return an option because we would need `ModuleMetadata` by value, but it is not copiable

crates/iota-adapter-transactional-tests/tests/package_metadata/no_module_dynamic_metadata.move
```diff
@@ -0,0 +1,46 @@
...
+public fun assert_inline_metadata(metadata: &PackageMetadataV1) {
+    let module_name = ascii::string(b"inline_v1");
+    let auth_function_name = ascii::string(b"authenticate");
+    let module_metadata = metadata.modules_metadata_v1(&module_name);
+    let authenticator_metadata = module_metadata.authenticator_metadata_v1(&auth_function_name);
+    assert!(authenticator_metadata.account_type() == type_name::get<Account>(), 0);
+    assert!(*authenticator_metadata.function_name() == auth_function_name, 1);
+}
+
+//# run test::no_module_dynamic_metadata::assert_inline_metadata --sender A --args object(1,1)
```
valeriyr:
```suggestion
//# run test::no_module_dynamic_metadata::assert_inline_metadata --sender A --args object(1,0)
```

crates/iota-adapter-transactional-tests/tests/programmable/private_account_functions.snap
```diff
@@ -8,30 +8,27 @@ A: object(0,0)
...
 
 task 2, lines 18-21:
-//# programmable --inputs object(1,1) "account" "authenticate"
+//# programmable --inputs object(1,2) "account" "authenticate"
 //> 0: test::account::create();
 //> 1: iota::authenticator_function::create_auth_function_ref_v1<test::account::Account>(Input(0), Input(1), Input(2));
 //> 2: iota::account::create_account_v1<test::account::Account>(Result(0), Result(1));
-Error: Transaction Effects Status: Non Entry Function Invoked. Move Call must start with an entry function
-Execution Error: NonEntryFunctionInvoked: Non Entry Function Invoked. Move Call must start with an entry function; caused by: Cannot directly call iota::account::create_account_v1.; at command index: 2
+Error: Error checking transaction input objects: Object object(1,2) is owned by object 0xe10faf06a75a91ed3d7783e4ea58a7a72430655dfa3672559ed3c6d53b742c9c. Objects owned by other objects cannot be used as input arguments
```
valeriyr: This test suite checks that it must be impossible to call protected functions from a PTB, like this:
Cannot directly call iota::account::create_account_v1
It is broken and needs to be fixed.

crates/iota-framework-snapshot/manifest.json
```diff
@@ -738,5 +738,30 @@
         "id": "0x000000000000000000000000000000000000000000000000000000000000107a"
       }
     ]
+  },
+  "29": {
+    "git_revision": "b2a7e207983a97d95bf53b190290939b8341708c-dirty",
```
valeriyr: The revision shouldn't be `dirty`.

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,140 @@
...
+}
+
+public struct ModuleMetadataKey(ascii::String) has copy, drop, store;
+
+public struct ViewFunctionMetadataV1FieldName has copy, drop, store {}
+
+public fun borrow_view_functions_metadata_v1(
+    module_metadata: &ModuleMetadata,
+): &vector<ascii::String> {
+    borrow<ViewFunctionMetadataV1FieldName, vector<ascii::String>>(
```
valeriyr:
```suggestion
    borrow(
```

valeriyr: Looks like the compiler can infer the types in this case.

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,140 @@
...
+    )
+}
+
+public fun is_view_function_v1(
+    module_metadata: &ModuleMetadata,
+    function_name: &ascii::String,
+): bool {
+    let view_functions_metadata = borrow_view_functions_metadata_v1(module_metadata);
+    view_functions_metadata.find_index!(|s| s == *function_name).is_some()
+}
```
valeriyr: It is not super important, but I would suggest renaming the `module_metadata` parameter in this file to `self` and using the dot notation:
```
self.borrow_view_functions_metadata_v1();
```

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
```diff
@@ -0,0 +1,140 @@
...
+
+/// Returns true iff there is an value associated with the key `k: K` in the module_metadata `module_metadata: &ModuleMetadata`
+/// with an assigned value of type `V`
+public fun contains_with_type<K: copy + drop + store, V: store>(
+    module_metadata: &ModuleMetadata,
+    k: K,
+): bool {
+    field::exists_with_type<K, V>(&module_metadata.id, k)
+}
```
valeriyr: Do we expect these functions to be public?

crates/iota-framework/packages/iota-framework/sources/derived_object.move
valeriyr: Can we delete this function since we have `derive_address`?

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
lollobene: good one

crates/iota-framework/packages/iota-framework/sources/derived_object.move
lollobene: I think so, @miker83z ?

miker83z: Yes, that was created because we did not pull the function that we now pulled from upstream!

crates/iota-framework/packages/iota-framework/sources/package_metadata/module_metadata.move
lollobene: Given that `borrow` is used by package_metadata.move, we can make it `public(package)`.
We could do the same for `contains` and `contains_with_type`, even though they are not used anywhere atm, so either we remove them and add later if needed, or make them `public(package)` @miker83z

lollobene: Actually, I added `contains_view_functions_metadata_v1` which uses `contains`, it might be helpful but can be removed

miker83z: keep them as public(package) then

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2922,6 +2940,12 @@ impl ProtocolConfig {
                     // fresh inbound traffic -- e.g. after a validator restart -- instead of
                     // staying pending forever and blocking epoch close.
                     cfg.feature_flags.always_advance_dkg_to_resolution = true;
+                    // Publish package metadata with the module metadata stored as a
+                    // dynamic field.
+                    if cfg.feature_flags.publish_package_metadata {
+                        cfg.feature_flags
+                            .package_metadata_with_dynamic_module_metadata = true;
+                    }
```
miker83z:
```suggestion
                    cfg.feature_flags.package_metadata_with_dynamic_module_metadata = true;
```

miker83z: This is enough

iota-execution/latest/iota-adapter/src/programmable_transactions/package_metadata.rs
```diff
@@ -0,0 +1,423 @@
...
+            Ok(())
+        })();
+        // Drop the pure inputs synthesized for the call above.
+        context.truncate_inputs(restore_inputs_len);
+        context.linkage_view.reset_linkage();
+        context.linkage_view.restore_linkage(saved_linkage)?;
+        result
+    }
+
+    fn to_bcs_argument<T: Serialize>(
```
miker83z: nit: I think we can use bcs::to_bytes inline, instead of this. That's the normal usage in the rest of the codebase.

miker83z (approved):

bingyanglin (approved):
Approved `iota-core`, `iota-protocol-config`, `iota-swarm-config`, and `iota-types`.

kodemartin (approved):

piotrm50 (approved):

crates/iota-framework/packages/iota-framework/sources/package_metadata/package_metadata.move
```diff
@@ -0,0 +1,478 @@
...
+    assert!(modules_metadata.contains(&name), EModuleMetadataNotFound);
+    let idx = modules_metadata.get_idx(&name);
+    let (_, metadata) = modules_metadata.get_entry_by_idx(idx);
+    metadata
+}
+
+/// Borrows the `AuthenticatorMetadataV1` of the function named `function_name`
+/// within the module named `module_name`.
+/// Aborts if the module or the authenticator metadata is not found.
+#[allow(deprecated_usage)]
```
lzpap, line 106: Maybe I'm missing something, but why aren't the new authenticator metadata related getters are in this module, while the view function getters are in `iota::module_metadata`?

lollobene: because `AuthenticatorMetadataV1` is defined inside `iota::package_metadata`, so adding getters in `iota::module_metadata` would require importing `AuthenticatorMetadataV1` type, and such would create a circular dependency.

Issue #11619, opened 2026-05-22
[graphql]: remove WrappedOrDeleted object status


kodemartin:
## Requirements

- [ ] Implement changes
- [ ] Sync with `iota-rust-sdk`

PR #11620, opened 2026-05-22
chore(graphql): deprecate WrappedOrDeleted object status

kodemartin:
# Description of change

Patch that prepares the removal of the `WrappedOrDeleted` object status from the GraphQL API.

Part of #11619

### Release Notes

- [x] GraphQL: :warning: Deprecates `WRAPPED_OR_DELETED` variant of `ObjectKind`. This kind of objects will be treated as non-existent. The variant will be removed with the v1.26 release.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/object.rs
+++ b/crates/iota-graphql-rpc/src/types/object.rs
@@ -113,7 +113,7 @@ pub enum ObjectStatus {
     /// The object is deleted or wrapped and only partial information can be
     /// loaded from the indexer.
     #[graphql(
-        deprecation = "will be removed with v1.26, as such objects can be construed as non-existent"
+        deprecation = "will be removed with v1.26, as such objects can be considered non-existent"
     )]
     WrappedOrDeleted,
 }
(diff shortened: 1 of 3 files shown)


crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -112,6 +112,9 @@ pub enum ObjectStatus {
     Indexed,
     /// The object is deleted or wrapped and only partial information can be
     /// loaded from the indexer.
+    #[graphql(
+        deprecation = "will be removed with v1.26, as such objects can be construed as non-existent"
```
tomxey: nit: "considered non-existent" would be easier to understand probably

tomxey (approved):

sergiupopescu199 (approved):

crates/iota-graphql-rpc/src/types/object.rs
kodemartin: https://github.com/iotaledger/iota/pull/11620/commits/d508b9cfc66dc72dbcd103ba72bce30ff2114f62

PR #11626, opened 2026-05-22
feat: make AuthContext aware of the AuthenticatorFunctionRefV1 instances used during authentication

### LINKED ISSUE #11579: [AA] Make the AuthContext aware of the Authenticator Function used during authentication
- This should be enabled for sender and sponsor authenticators

valeriyr:
# Description of change

Move authenticator contracts can now inspect which authenticator function the sender and sponsor used during transaction execution via two new AuthContext accessors: sender_authenticator_function_info_v1() and sponsor_authenticator_function_info_v1().

Introduced a Move-side struct AuthenticatorFunctionInfoV1, which is a full representation of AuthenticatorFunctionRefV1 but without a generic Account parameter. It is added since the Rust-side AuthenticatorFunctionRefV1 doesn't handle information about the phantom Account type, so we can not resolve it when an accessor is called.

fixes #11579 

### Release Notes

- [x] Protocol: Move authenticator contracts can now inspect which authenticator function the sender and sponsor used during transaction execution via two new AuthContext accessors: sender_authenticator_function_info_v1() and sponsor_authenticator_function_info_v1().
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-framework/packages/iota-framework/sources/auth_context.move
```diff
@@ -91,6 +134,10 @@ native fun native_sender_auth_digest(): &vector<u8>;
 
 native fun native_sponsor_auth_digest(): &Option<vector<u8>>;
 
+native fun native_sender_authenticator_function_info_v1<F>(): &Option<F>;
```
miker83z, line 137:
```suggestion
native fun native_sender_authenticator_function_info<F>(): &Option<F>;
```
Given the use of the generic, we can return a v2, v3,... in the future, or am I wrong?

miker83z (approved):
lgtm!

crates/iota-framework/packages/iota-framework/sources/auth_context.move
valeriyr: I would rename it later if it is necessary. Theoretically, you can return any version, yes, but at the same time, you need to understand which version should be taken from the Rust-side AuthContext and which price to pay.

miker83z: Isn't it enough the public function invoking this native function to do what you said? That is: 
- sender_authenticator_function_info_v1 -> native_sender_authenticator_function_info\<AuthenticatorFunctionInfoV1\>()
- sender_authenticator_function_info_v2 -> native_sender_authenticator_function_info\<AuthenticatorFunctionInfoV2\>()
- ...

valeriyr: Not really, in this case, you have a type and its layout. You need to have the ability to check if this is v1 or v2; also, we don't know how much v2 will be different from v1. It makes sense to refactor this function when we add v2, so we will have clear requirements.

bingyanglin (approved):

miker83z (approved):

kodemartin (approved):

PR #11629, opened 2026-05-23
fix(move-analyzer): skip module-name UseDef from a different file

### LINKED ISSUE #2421: Check suspicious `move-analyzer` behavior
The `external-crates/move/crates/move-analyzer/tests/macros.ide` test changes the result depending on if the iota license header is added in the `external-crates/move/crates/move-stdlib/sources/vector.move` file(line 2).

The following test part is affected:
```
"macros.move": [
        {
          "use_line": 7,
          "use_ndx": 0
        },
        {
          "use_line": 7,
          "use_ndx": 1
        },
        {
          "use_line": 7,
          "use_ndx": 2
        },
```
If remove the second line from the `external-crates/move/crates/move-stdlib/sources/vector.move` file, the first test:
```
{
  "use_line": 7,
  "use_ndx": 0
},
```
has the following result:
```
== macros.move ========================================================
-- test 0 -------------------
use line: 7, use_ndx: 0
Use: 'n foo(', start: 12, end: 18
Def: 'vector', line: 6, def char: 12
TypeDef: no info
On Hover:
module std::vector

A variable-sized container that can hold any type. Indexing is 0-based, and
vectors are growable. This module has many native functions.
```
This looks like an issue because the plugin highlights the vector that is not presented in the checked function definition:
```
macro fun Macros::macros::foo($i: u64, $body: |u64| -> u64): u64
```

The behavior is reproduced in the Sui repository on the `1.32.2` version.

In `Visual Studio Code` it looks like this:

Screenshot 2024-10-03 at 11 36 15

lollobene:
# Description of change

- The `move-analyzer` hover would occasionally point at an unrelated module (e.g., `std::vector`) when the user hovered on a position that doesn't even contain an identifier.
- The Move compiler's typed AST gives resolved module identifiers the  **declaration-site** `Loc`. The analyzer in `add_fun_use_def`, `add_datatype_use_def` and `add_const_use_def` was using `mod_name.loc()` as if it were the use-site position, then inserting a `UseDef` into the *current* module's `UseDefMap`. `UseDef` only stores `col_start`/`col_end` (no file hash), so when the declaration sits in a different file the coordinates got reinterpreted against the wrong file's content — usually harmless, occasionally aligning with a real query position.
- The reported symptom was that `tests/macros.ide` produced a different result depending on whether the IOTA license header was present in `move-stdlib/sources/vector.move` — the header just shifted the bogus entry off the line the test queries, hiding the bug.

## Fix
Gate the module-name `UseDef` insertion on mod_name.loc().file_hash() == use_pos.file_hash()`. If the module name wasn't actually written in the same file as the use (e.g., dot calls like `v.length()` where the receiver type's module is synthesized), skip the insertion instead of poisoning the current module's map.

## Caveats
- Aliased imports (`use std::vector as Vec; Vec::length(...)`) and the second `add_fun_use_def(..., &fun_def_loc)` call in `visit_use_funs` are not addressed here; a complete fix would have `UseDef` carry an explicit use-site `file_hash` and enforce it at construction. Filed as follow-up.

fixes #2421

### DIFF
--- a/external-crates/move/crates/move-analyzer/src/analysis/parsing_analysis.rs
+++ b/external-crates/move/crates/move-analyzer/src/analysis/parsing_analysis.rs
@@ -675,7 +675,6 @@ impl<'a> ParsingAnalysisContext<'a> {
             return;
         };
         self.use_defs.insert(
-            mod_name_start.line,
             UseDef::new(
                 self.references,
                 &BTreeMap::new(),
@@ -739,7 +738,7 @@ impl<'a> ParsingAnalysisContext<'a> {
                     alias_start,
                     alias.loc.file_hash(),
                 );
-                self.use_defs.insert(alias_start.line, ud);
+                self.use_defs.insert(ud);
             }
             return;
         }
@@ -767,7 +766,7 @@ impl<'a> ParsingAnalysisContext<'a> {
                     alias_start,
                     alias.loc.file_hash(),
                 );
-                self.use_defs.insert(alias_start.line, ud);
+                self.use_defs.insert(ud);
             }
         }
     }
--- a/external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
+++ b/external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
@@ -156,21 +156,30 @@ impl TypingAnalysisContext<'_> {
             return;
         };
         // insert use of the const's module
+        //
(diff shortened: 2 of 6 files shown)


miker83z:
Closed in favor of pulling the fix from upstream https://github.com/MystenLabs/sui/commit/6c78e5d54d822dfd9033d0bdb54e070a73d94977

PR #11656, opened 2026-05-27
refactor(iota-indexer): support filtered queries on transactions with a fallback to the archival store

### LINKED ISSUE #9498: feat(indexer): Consider fallback support for case where we don't prune indices
Created from: https://github.com/iotaledger/iota/issues/9474
> We could consider to support case where we prune the transactions/events table, but don't prune the indices.
In such case we could fetch tx/ev ids from indices, then convert that to digests and fetch txs/evs from KV by digest.

sergiupopescu199:
# Description of change

This PR adds support for pruned transactions to be retrieved form the fallback archival store in the `IndexerApi::query_transaction_blocks`.

Included the `signal` feature flag to `tokio` dependency needed for the graceful shutdown.

fixes #9498 

The fix was verified by manually simulating a pruned state and confirming the API successfully fetches missing data from the fallback store.

1. Started a local network using the `iota-localnet` cli tool
```shell
cargo r -- -p iota-localnet -- start --force-regenesis --with-grpc
```
2. Start the kv-store ingestion task to write checkpoints into BigTable (see the README in the `iota-kvstore` crate for details):
```shell
cargo r -p iota-data-ingestion --bin iota-data-ingestion -- bigtable.yaml
```
3. Start the kv-store REST API
```shell
cargo -p iota-rest-kv r -- --config bigtable.yaml
```
4. Start the indexer writer:
```shell
cargo r -p iota-indexer -- --database-url=postgres://postgres:postgrespw@localhost:5432/iota_indexer indexer --remote-store-url=http://localhost:50051 --reset-db
```
5. Start the indexer reader with the fallback KV URL configured:
```shell
 cargo r -p iota-indexer -- --database-url=postgres://postgres:postgrespw@localhost:5432/iota_indexer --metrics-address=0.0.0.0:9185 json-rpc-service --rpc-client-url=http://localhost:50051 --fallback-kv-url=http://localhost:3555 --rpc-address=0.0.0.0:9001
```

Once the database had ingested some transactions, we picked a starting point for the test queries. We listed checkpoints:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "iota_getCheckpoints",
  "params": [
    "10",
    4,
    false
  ]
}
```
and used the first transaction digest (`37JSrDLCumYdftCbyXMCupcjCLEw2hmq1SXtwHVSVdz5`) from checkpoint sequence number `11` as the cursor.

We then ran a filtered query and recorded the response:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "iotax_queryTransactionBlocks",
    "params": [
        {
            "filter": {
                "TransactionKind": "SystemTransaction"
            },
            "options": null
        },
        "37JSrDLCumYdftCbyXMCupcjCLEw2hmq1SXtwHVSVdz5",
        100,
        false
    ]
}
```
From the response we picked several transactions to delete directly from the `transactions` table by their sequence number (we removed `18`–`20` and `23`), simulating pruned rows whose data is still available in the archival store.

We then re-ran the identical query. The response was to the original one, confirming the deleted rows were recovered from the fallback store, and that ordering was preserved.

We repeated the same procedure with `descending_order: true`, deleting transactions from that page and re-querying. The results matched the original and ordering was preserved.

> [!NOTE]
> This patch doe snot affect the normal ingestion operation of the indexer, it only affects the reading part (JSON-RPC), which tests were conducted above.

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -1559,51 +1559,35 @@ impl IndexerReader {
             .filter(|_| stored_txes.len() != tx_sequence_numbers.len())
         {
             Some(fallback_reader) => {
-                // pre-allocate results with None. Found transactions will replace None with
-                // Some(tx).
-                let mut results: Vec<Option<StoredTransaction>> =
-                    vec![None; tx_sequence_numbers.len()];
+                let found: HashSet<i64> =
+                    stored_txes.iter().map(|tx| tx.tx_sequence_number).collect();
+                let missing_seqs: Vec<i64> = tx_sequence_numbers
+                    .iter()
+                    .copied()
+                    .filter(|s| !found.contains(s))
+                    .collect();
 
-                match is_descending {
-                    Some(true) => tx_sequence_numbers.sort_unstable_by_key(|&s| Reverse(s)),
-                    Some(false) => tx_sequence_numbers.sort_unstable(),
-                    None => {}
-                }
-
-                // each entry is a tuple of (key, original_index) so we can merge fetched data
-                // back into the correct position in the results vector.
-                let mut tx_seq_to_index = tx_sequence_numbers
-                    .into_iter()
-                    .enumerate()
-                    .map(|(idx, seq)| (seq, idx))
-                    .collect::<HashMap<i64, usize>>();
-
-                for tx in stored_txes {
-                    if let Some(i) = tx_seq_to_index.remove(&tx.tx_sequence_number) {
-                        results[i] = Some(tx);
-                    }
-                }
-
-                let missing_tx_digests = self
+                let missing_digests = self
                     .db()
-                    .resolve_tx_sequence_numbers_to_digests(
-                        tx_seq_to_index.keys().copied().collect(),
-                    )
+                    .resolve_tx_sequence_numbers_to_digests(missing_seqs)
                     .await?;
 
-                let historical_transactions = fallback_reader
-                    .transactions(&missing_tx_digests)
+                let historical = fallback_reader
+                    .transactions(&missing_digests)
                     .await?
                     .into_iter()
                     .flatten();
 
-                for tx in historical_transactions {
-                    if let Some(i) = tx_seq_to_index.remove(&tx.tx_sequence_number) {
-                        results[i] = Some(tx);
-                    }
+                let mut merged: Vec<StoredTransaction> =
+                    stored_txes.into_iter().chain(historical).collect();
+
+                match is_descending {
+                    Some(true) => merged.sort_unstable_by_key(|tx| Reverse(tx.tx_sequence_number)),
+                    Some(false) => merged.sort_unstable_by_key(|tx| tx.tx_sequence_number),
+                    None => {}
                 }
 
-                results.into_iter().flatten().collect()
+                merged
             }
             None => stored_txes,
         };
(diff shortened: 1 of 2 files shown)


crates/iota-indexer/src/read.rs
```diff
@@ -3073,6 +3131,31 @@ impl<'a> DBReader<'a> {
...
     }
+
+    /// Resolve transaction sequence numbers to their corresponding digests.
+    async fn resolve_tx_sequence_numbers_to_digests(
+        &self,
+        tx_sequence_numbers: Vec<i64>,
+    ) -> Result<Vec<TransactionDigest>, IndexerError> {
+        let pool = self.main_reader.get_pool();
+
+        let rows = run_query_async!(&pool, |conn| {
```
tomxey, line 3126: if we would liketo make sure that tx digests is not pruned more than the tx_* tables used to get the sequence numbers, we could add here a line like:

```rust
 self.ensure_data_not_pruned_for_tx(tx_sequence_numbers.min(), [tx_digests])?;
```

just FYI, not sure if it's desired

tomxey (approved):

crates/iota-indexer/src/read.rs
```diff
@@ -1535,22 +1535,80 @@ impl IndexerReader {
...
+                let mut results: Vec<Option<StoredTransaction>> =
+                    vec![None; tx_sequence_numbers.len()];
+
+                match is_descending {
+                    Some(true) => tx_sequence_numbers.sort_unstable_by_key(|&s| Reverse(s)),
+                    Some(false) => tx_sequence_numbers.sort_unstable(),
+                    None => {}
+                }
+
+                // each entry is a tuple of (key, original_index) so we can merge fetched data
```
kodemartin: `(key, original_index)` is not true anymore because of the sorting

crates/iota-indexer/src/read.rs
```diff
@@ -1535,22 +1535,80 @@ impl IndexerReader {
...
+                    .into_iter()
+                    .flatten();
+
+                for tx in historical_transactions {
+                    if let Some(i) = tx_seq_to_index.remove(&tx.tx_sequence_number) {
+                        results[i] = Some(tx);
+                    }
+                }
+
+                results.into_iter().flatten().collect()
```
kodemartin: Nit: With a slight refactoring I think this can read more cleanly: No extra comments about the pre-allocation, less operations, no double flattening in `historical_transactions` and `results`.

```suggestion
                let found: HashSet<i64> =
                    stored_txes.iter().map(|tx| tx.tx_sequence_number).collect();
                let missing_seqs: Vec<i64> = tx_sequence_numbers
                    .iter()
                    .copied()
                    .filter(|s| !found.contains(s))
                    .collect();

                let missing_digests = self
                    .db()
                    .resolve_tx_sequence_numbers_to_digests(missing_seqs)
                    .await?;

                let historical = fallback_reader
                    .transactions(&missing_digests)
                    .await?
                    .into_iter()
                    .flatten();

                let mut merged: Vec<StoredTransaction> =
                    stored_txes.into_iter().chain(historical).collect();

                match is_descending {
                    Some(true) => merged.sort_unstable_by_key(|tx| Reverse(tx.tx_sequence_number)),
                    Some(false) => merged.sort_unstable_by_key(|tx| tx.tx_sequence_number),
                    None => {}
                }

                merged
```

kodemartin (approved):

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed as part of: https://github.com/iotaledger/iota/pull/11656#discussion_r3311323856

PR #11658, opened 2026-05-27
feat(graphql): migrate Loader<ParentVersionKey> to backward-diff

### LINKED ISSUE #11585: [graphql]: Replace `objects_history` in `Loader<ParentVersionKey>`
Subissue of #11580 (R2 in [the audit](https://github.com/iotaledger/iota/issues/11580)).

**Location**: `crates/iota-graphql-rpc/src/types/object.rs:1681-1778`

**What it reads**: for each id, the latest version `≤ parent_version` (full row content), grouped by `(checkpoint_viewed_at, parent_version)`.

**Backs GraphQL**: `Object::under_parent`, used e.g. by dynamic-object-field lookups bound to a root version.

**Current tables**: `objects_history ⨝ objects_version` with `DISTINCT ON (object_id) … ORDER BY object_version DESC`.

**Replacement direction**: `DISTINCT ON (object_id) … WHERE object_version ≤ parent_version` over `objects_version`, then `LEFT JOIN objects` and `LEFT JOIN objects_backward_history (object_status = ACTIVE)` and `COALESCE` the content. Same general shape as `Loader<LatestAtKey>`.

tomxey:
# Description of change

This PR migrates the `Loader<ParentVersionKey>` to use `objects_backward_history` instead of `objects_history`. 

More precisely it replaces the `objects_history ⨝ objects_version DISTINCT ON` read in `Loader<ParentVersionKey>` with a CTE that uses a `LATERAL` join with `LIMIT 1` over `objects_version` to find the largest version ≤ `parent_version` per id.
With the version found it tries then to load it from merged `checkpointed_objects ∪ objects_backward_history` tables, same as `Loader<HistoricalKey>`.

Picking the version via `LATERAL ... LIMIT 1` lets Postgres use the per-partition `(object_id, object_version)` primary key for a backward index scan; a `DISTINCT ON (object_id) ... WHERE object_id = ANY(...) AND object_version <= V` formulation provoked an external-sort parallel seq scan instead (49 s vs 18 ms on staging).

fixes #11585 

### DIFF
--- a/crates/iota-graphql-rpc/src/types/object.rs
+++ b/crates/iota-graphql-rpc/src/types/object.rs
@@ -12,12 +12,10 @@ use async_graphql::{
     dataloader::Loader,
     *,
 };
-use diesel::{
-    BoolExpressionMethods, ExpressionMethods, JoinOnDsl, QueryDsl, SelectableHelper, sql_types,
-};
+use diesel::{BoolExpressionMethods, ExpressionMethods, QueryDsl, SelectableHelper, sql_types};
 use iota_indexer::{
     models::objects::{StoredHistoryObject, StoredObject},
-    schema::{objects, objects_history, objects_version},
+    schema::objects,
     types::{ObjectStatus as NativeObjectStatus, OwnerType},
 };
 use iota_sdk_types::{StructTag, TypeTag};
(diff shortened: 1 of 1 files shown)


crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1760,29 +1758,90 @@ impl Loader<ParentVersionKey> for Db {
...
 
+        let active = NativeObjectStatus::Active as i16;
+        let wrapped_or_deleted = NativeObjectStatus::WrappedOrDeleted as i16;
+
+        // For each id, pick the largest version `≤ parent_version` from
+        // `objects_version` (the existence oracle for every version ever
+        // committed) via a `LATERAL` join with `LIMIT 1`, then locate the
+        // row content in `checkpointed_objects` (current state of the
+        // object, which may be a `WrappedOrDeleted` tombstone) or
+        // `objects_backward_history`.
```
kodemartin: "existence oracle" is an ambiguous term. If we want to inform about the properties of the table we can be explicit about it.

Also simply mentioning the LATERAL join method here might not be necessary. Readers can clearly see the method in the query. Unless you want to put emphasis on the advantage of the pattern with respect to the old DISTINCT ON. In this case you can add a note at the end of the comment in a dedicated paragraph.

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -1760,29 +1758,90 @@ impl Loader<ParentVersionKey> for Db {
...
+        //
+        // Only `ACTIVE` rows from `objects_backward_history` are joined
+        // (`bh.object_status = $3`) because they are the only rows that
+        // carry the real `(object_id, object_version)`. The other two
+        // statuses use placeholder versions, so matching on them by
+        // `(object_id, object_version)` could by chance return an incorrect
+        // row (with incorrect version).
```
kodemartin: I think that this paragraph is too verbose. It could be simplified to

```suggestion
        // `objects_backward_history` is joined only on rows with `ACTIVE` status. The other
        // two variants (`NOT_YET_CREATED`, `WrappedOrDeleted`) are paired with placeholder versions
        // and matching on them by `(object_id, object_version)` does not guarantee that the version is
        // correct.
```

But I want to make the following observations: If there is a match in `objects_version` we know that it is either `Active` or `WrappedOrDeleted`. `Active` matches must match `Active` rows in backwards history as well, no? That is, every active object is stored with the correct version in `backwars_history`. Conversely we are sure that `WrappedOrDeleted` objects cannot match an `Active` row in `backwards_history`. Can they match the `NOT_YET_CREATED` status? Again no, because the actually wrapped-or-deleted version is guaranteed to be larger than the `lamport - 1` version at the time of creation (we actually store `-1`).

tomxey: simplified in 411d3d735d
I've also simplified the query, seems the active status is not needed after all, objects version guard the version correctness here

crates/iota-graphql-rpc/src/types/object.rs
tomxey: simplified in 411d3d735d

kodemartin (approved):

sergiupopescu199 (approved):

PR #11660, opened 2026-05-27
feat(iota-types): compile iota-types to wasm32

### LINKED ISSUE #11546: Make iota-types wasm32-compatible
Parent: #11545

Make `iota-types` compile to `wasm32-unknown-unknown` while keeping the node build byte-identical.

Part of the **IOTA Move VM WASM** epic (#11545).

### Cargo.toml — gate node-only deps

Move to `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`:

- `anemo`
- `fastcrypto-tbls`
- `iota-network-stack`
- `prometheus`
- `starfish-config`
- `tonic`

### Provide wasm-compatible stubs

- New `src/wasm_multiaddr.rs`: BCS / serde round-trip compatible `Multiaddr(String)` stub, re-exported as `iota_types::multiaddr` on wasm. The on-chain wire format for `Multiaddr` is a UTF-8 string, so the stub is binary-compatible with the real `iota_network_stack::multiaddr::Multiaddr` for purposes of decoding on-chain validator metadata.
- `src/metrics.rs`: split into a native `imp` module backed by `prometheus` and wasm no-op stubs for `IntCounter`, `IntCounterVec`, `Histogram`, `LimitsMetrics`, `BytecodeVerifierMetrics`, etc., with identical public surface so call sites do not need cfg-gating.

### Gate node-only code paths

Add `#[cfg(not(target_arch = "wasm32"))]` to:

- `RandomnessSignature` / `RandomnessPartialSignature` / `RandomnessPrivateKey` type aliases (use `fastcrypto-tbls`) — `src/crypto.rs`.
- `VersionedDkgMessage` / `VersionedDkgConfirmation` and helpers; `ConsensusTransaction::new_randomness_dkg_message` / `_confirmation` — `src/messages_consensus.rs`.
- `tonic::Status` ↔ `IotaError` conversions — `src/error.rs`.
- `CheckpointSummary::report_checkpoint_age` (uses `prometheus::Histogram`) — `src/messages_checkpoint.rs`.
- `EpochStartSystemStateTrait` methods `get_consensus_committee`, `get_validator_as_p2p_peers`, `get_authority_names_to_peer_ids` (pull in `anemo` + `starfish-config`) and their `V1`/`V2` impls — `src/iota_system_state/epoch_start_iota_system_state.rs`.
- `TxContext::random_for_testing_only` — `src/base_types.rs`.

### Correctness fix (not just a wasm gate)

`dynamic_field::derive_field_id` currently uses `key_bytes.len().to_le_bytes()`. `usize::to_le_bytes()` is platform-dependent: 4 bytes on `wasm32`, 8 bytes on `x86_64`. The on-chain Move VM encodes that length as `u64`. Encode it as `u64` explicitly:

```rust
hasher.update(
    u64::try_from(key_bytes.len())
        .expect("key_bytes must fit into a u64")
        .to_le_bytes(),
);
```

This is a real bug — on `wasm32` today the helper would derive a different `Field<K,V>` ID than the chain — and the fix lands regardless of when other wasm work merges.

### CI: introduce the wasm32 check

Add a new step / job to `.github/workflows/_rust_tests.yml`:

```yaml
- name: Install wasm32 target
  run: rustup target add wasm32-unknown-unknown
- name: Check iota-types compiles to wasm32
  run: cargo check -p iota-types --target wasm32-unknown-unknown
```

This is the **only** crate added to the wasm32 CI check in this PR. Subsequent sub-issues each extend this job with their own `cargo check -p <crate> --target wasm32-unknown-unknown` step.

## Acceptance criteria

- [ ] `cargo check -p iota-types --target wasm32-unknown-unknown` passes locally.
- [ ] New wasm32 CI step in `_rust_tests.yml` is green on the PR.
- [ ] All existing non-wasm builds and tests for `iota-types` and downstream crates remain green.
- [ ] No public API regressions on non-wasm targets (additions only, gated by `cfg`).

## Reference implementation

Commit dd31d34e81 on branch [`claude/wasm-stake-inspector-l1rVZ`](https://github.com/iotaledger/iota/tree/claude/wasm-stake-inspector-l1rVZ) contains a working version of all of the above. Treat it as a blueprint, not a copy-paste source — the PR should be re-derived cleanly against current `develop`.

Thoralf-M:
## Summary

Makes the `iota-types` crate compile to `wasm32-unknown-unknown`, a prerequisite for running the Move VM / execution layer in the browser. The guiding rule: **client-visible types stay wasm-compatible; node-only types are gated out entirely.**

Fixes https://github.com/iotaledger/iota/issues/11546

Two mechanisms, applied per module:

- **Module is node-only** → gate the whole module out on wasm. The native file stays byte-identical to `develop`.
  - `messages_consensus` + `messages_grpc` (consensus / gRPC machinery; pull in `fastcrypto-tbls` / `tonic`).
  - `iota_system_state::epoch_start_iota_system_state` (`EpochStartSystemState`; pulls in `anemo` / `starfish-config`). The single trait link that surfaced it — `IotaSystemStateTrait::into_epoch_start_state` — is gated along with its V1/V2 impls.
  - Knock-on edits only: `transaction.rs` now imports `CancelledTransaction` / `ConsensusCommitPrologueV1` / `ConsensusDeterminedVersionAssignments` directly from `iota_sdk_types` (they were re-exports), and the `BcsSignable` impl for `VersionedMisbehaviorReport` is gated.

- **Module exports types the execution/client path needs on wasm** → provide no-op stubs.
  - `metrics`: native items gated in place (no re-indentation); `LimitsMetrics` / `BytecodeVerifierMetrics` stubs appended behind `#[cfg(wasm32)]`.
  - `multiaddr`: `Multiaddr` appears in on-chain validator metadata; backed by a small wasm stub in `wasm_multiaddr.rs` (the only added file).

`Cargo.toml` sorts the wasm-incompatible dependencies (`prometheus`, `anemo`, `tonic`, `fastcrypto-tbls`, `starfish-config`, …) into `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`.

## CI

Adds a `check-wasm32` job to `_rust_tests.yml` that installs the `wasm32-unknown-unknown` target and runs `cargo check -p iota-types --target wasm32-unknown-unknown`, guarding the new constraint going forward.

## Test plan

- [x] `cargo check -p iota-types --target wasm32-unknown-unknown`
- [x] `cargo check -p iota-types` (native)
- [x] `cargo check -p iota-core -p iota-node` (downstream native)
- [x] `cargo clippy -p iota-types -- -D warnings`
- [x] `IOTA_SKIP_SIMTESTS=1 cargo test -p iota-types --lib` (216 passed)


bingyanglin (approved):

crates/iota-types/src/dynamic_field.rs
```diff
@@ -262,7 +262,14 @@ where
     let mut hasher = DefaultHash::default();
     hasher.update([HashingIntentScope::ChildObjectId as u8]);
     hasher.update(parent);
-    hasher.update(key_bytes.len().to_le_bytes());
+    // The length must be encoded as u64 (8 bytes LE) to match the on-chain
+    // derivation used by the Move VM. `usize::to_le_bytes()` is platform-
+    // dependent (4 bytes on wasm32, 8 on x86_64), so cast explicitly.
```
thibault-martinez: Not really a cast. Plus, a cast should actually be enough?

crates/iota-types/src/wasm_multiaddr.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    /// Stub: on the node this validates the multiaddr can be turned into an
+    /// anemo address; on wasm we don't have anemo, so we accept anything that
+    /// deserialized successfully. Returning `Ok(())` is enough because the
+    /// Move verifier on the node already gated invalid metadata at validator
+    /// creation time — we're inspecting already-on-chain state.
+    pub fn to_anemo_address(&self) -> Result<(), ()> {
```
thibault-martinez: Feels like we shouldn't even need this. If we are gating gating anemo out, we should also gate out code that relies on it?

crates/iota-types/src/metrics.rs
```diff
@@ -137,3 +142,114 @@ impl BytecodeVerifierMetrics {
...
+                verifier_runtime_per_ptb_success_latency: StubHistogram,
+                verifier_runtime_per_module_timeout_latency: StubHistogram,
+                verifier_runtime_per_ptb_timeout_latency: StubHistogram,
+            }
+        }
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+pub use wasm_stubs::*;
```
thibault-martinez: I think I would remove this, makes it clearer it's wasm when importing from metrics (since the types don't have wasm in their name).

crates/iota-types/src/metrics.rs
```diff
@@ -137,3 +142,111 @@ impl BytecodeVerifierMetrics {
...
 }
+
+/// wasm32 stubs — `prometheus` pulls in libc and can't compile to
+/// `wasm32-unknown-unknown`. The execution engine reads metric fields by name
+/// and invokes `inc`/`observe`/`start_timer`; we replicate that surface with
+/// no-op types so the call sites compile unchanged. Constructors take no
+/// arguments (instead of a `prometheus::Registry`) — wasm callers use
+/// `LimitsMetrics::new_stub()` / `BytecodeVerifierMetrics::new_stub()`.
+#[cfg(target_arch = "wasm32")]
+pub mod wasm_stubs {
```
Alex6323: Can you not do the same trick as for the `multiaddr`, so imports stay the same for wasm?

thibault-martinez: What trick? The difference is that the original multiaddr module comes from another crate

Alex6323: `#[cfg(target_arch = "wasm32")] pub use wasm_stubs::{LimitsMetrics, BytecodeVerifierMetrics}; ` inside this module?

thibault-martinez: Inside which module? `wasm_stubs`?

Alex6323: mapping `multiaddr` (external and local module) so it becomes the same on imports for either target. For the `stubs` it seems similar. When we import `LimitsMetrics` for the WASM target it's different than for native.

Alex6323: inside `metrics`

thibault-martinez: But it's exactly the same, in one case we imported a module, in the other a type, but symbol-wise, the path are different and I think they should be as it shows intent (I actually made Claude make it even more different in a previous comment.

thibault-martinez: @Thoralf-M can decide but I'd personally keep as is

thibault-martinez: I'm wondering if we even need this at all of if we could/should gate out what is using it.

thibault-martinez (approved):

Alex6323 (approved):

crates/iota-types/src/metrics.rs
Thoralf-M: Just removed the wasm_stubs https://github.com/iotaledger/iota/pull/11660/changes/963cd13133b6fafa45c706c61e6c9aa37e48db6c

Thoralf-M: Just removed the wasm_stubs https://github.com/iotaledger/iota/pull/11660/changes/963cd13133b6fafa45c706c61e6c9aa37e48db6c

Thoralf-M: Yes by feature gating the VerifiedDigestCache we can remove it [`963cd13` (this PR)](https://github.com/iotaledger/iota/pull/11660/changes/963cd13133b6fafa45c706c61e6c9aa37e48db6c)

PR #11662, opened 2026-05-27
fix(iota-core): keep already-executed transactions in white-flag checkpoint roots

### LINKED ISSUE #11649: fork during double spend stress test
While stress testing 4 validators with double spend stress rs in iota-private-network. A fork on validator 3 was found.

- investigate
- recreate via unit or sim test if possible
- fix, confirm via test

[google drive zip of the validator logs](https://drive.google.com/file/d/1wLzZfkLiw50myA_GP283V9FMS7MDnpc8/view?usp=sharing)

filipdulic:
# Description of change

In the white-flag (certificate-less) flow, a transaction that this node had already executed via **state-sync** — before its consensus handler processed the commit that sequenced it — was silently dropped from the locally-built checkpoint by the post-consensus *already-executed* check (`validate_and_resolve_conflicts`, Check #1). Because that pass runs **before** checkpoint `roots` are computed, the transaction was excluded from the local checkpoint while the rest of the committee included it, so the lagging node hit a fatal `Local checkpoint fork detected` and crash-looped.

**Root cause:** an *execution-state* check sat on the *checkpoint-membership* path. Membership must be decided by **consensus processing** (committee-deterministic), not by local execution progress (which depends on state-sync lag). The legacy certificate flow never had this problem — its already-executed filter lives only at the execution layer (`TransactionManager::enqueue`), never on the roots path. Note: the double-spend workload only *created the lag* that triggered this; it is not the cause.

**Fix:** treat an already-executed transaction as a committee-agreed winner that **stays in the sequence** — so it flows through scheduling/deferral into `roots` exactly as on every other validator. It registers its owned-object locks (so double-spend siblings still lose) and skips re-validation; `TransactionManager::enqueue` continues to suppress the actual re-execution. A `fatal!` guards the invariant that such a winner can never be out-locked by a different transaction.

Verified safe for both owned- and shared-object transactions: shared-object version assignment is idempotent (`get_or_init_next_object_versions` is first-touch-wins and guarded across the consensus/executor paths), so consensus assigns the same versions the committee did; owned-object txs run no version assignment at all.

fixes #11649

- [x] Basic tests (linting, compilation, formatting, unit/integration tests) — `cargo clippy -p iota-core --tests` clean, `cargo +nightly fmt` applied

Two unit tests in `post_consensus_validation_tests.rs`:

- `already_executed_tx_must_remain_in_checkpoint_roots` — reproduces the bug (**red before this fix, green after**): a transaction executed via the state-sync path must still appear in checkpoint roots when its consensus commit is later processed.
- `double_spend_loser_excluded_from_checkpoint_roots` — guards against an over-broad fix: the conflict loser must stay **out** of roots, otherwise the checkpoint builder would block forever waiting for effects that never arrive.

> Note: `test_consensus_handler*` and `test_consensus_commit_prologue_generation` exhibit pre-existing parallel-execution flakiness (they fail intermittently on the unmodified base and pass single-threaded; unrelated to this change).

### DIFF
--- a/crates/iota-core/src/post_consensus_validation.rs
+++ b/crates/iota-core/src/post_consensus_validation.rs
@@ -20,7 +20,8 @@
 //!
 //! 1. Non-`UserTransactionV1` — pass through unchanged.
 //! 2. Dedup by `ConsensusTransactionKey` — silent drop.
-//! 3. Already executed — silent drop.
+//! 3. Already executed — **retained** as a committee-agreed winner (registers
+//!    its locks, skips re-validation); not dropped. See issue #11649.
 //! 4. `validity_check()` — drop with error.
 //! 5. Three-tier lock conflict check (local HashMap → quarantine → DB) — drop
 //!    with error. Cheap; performed before expensive checks.
@@ -35,6 +36,7 @@ use std::{
     sync::Arc,
 };
 
+use iota_common::fatal;
 use iota_types::{
     base_types::{ObjectRef, TransactionDigest},
     error::{IotaError, IotaResult},
@@ -58,11 +60,12 @@ use crate::{
 /// conflicts in a single pass.
 ///
 /// For each `UserTransactionV1` in consensus order:
-/// - Runs deduplication, already-executed check, structural validity, lock
-///   conflict check, and deny checks (deny list, gas, ownership, coin deny
-///   list, Move authenticator).
+/// - Runs deduplication, structural validity, lock conflict check, and deny
+///   checks (deny list, gas, ownership, coin deny list, Move authenticator).
 /// - If all checks pass, acquires owned-object locks in a local tracking map.
 /// - Drops the transaction (with an error) on any failure.
+/// - An already-executed transaction is **retained** (not dropped): it registers
+///   its owned-object locks and skips re-validation. See issue #11649.
 ///
 /// Non-`UserTransactionV1` transactions pass through unchanged.
 ///
@@ -77,8 +80,8 @@ use crate::{
 ///
 /// `Ok((dropped, locks))` where:
 /// - `dropped` — `(digest, error)` for every semantically-invalid or
-///   lock-conflicting transaction. Silent drops (duplicates, already-executed)
-///   are **not** included.
+///   lock-conflicting transaction. Silent drops (duplicates) are **not**
+///   included.
 /// - `locks` — Owned-object locks acquired in this commit, to be stored in the
 ///   consensus quarantine so subsequent commits can see them.
 pub async fn validate_and_resolve_conflicts(
@@ -119,12 +122,48 @@ pub async fn validate_and_resolve_conflicts(
 
         let digest = *transaction.digest();
 
-        // Check #1: Already executed — silent dedup.
+        // Check #1: Already executed (typically by state-sync before this node's
+        // consensus handler reached the commit). It is a committee-agreed winner, so
+        // keep it in the sequence to flow into checkpoint roots like on every other
+        // validator (dropping it forks — issue #11649). Register its owned-object
+        // locks so double-spend siblings still lose, then skip re-validation (#2/#5);
+        // `TransactionManager::enqueue` suppresses the re-execution.
         if authority_state
             .get_transaction_cache_reader()
             .try_is_tx_already_executed(&digest)?
         {
-            keep[i] = false;
+            // Byte-based, so safe even though the inputs are already consumed.
+            let owned_inputs = extract_owned_input_objects(tx)?;
+            for obj_ref in &owned_inputs {
+                // A winner cannot be out-locked: an executed tx owns its inputs. A lock
+                // held by a different tx is a consistency violation, not a conflict.
+                let existing_lock = match current_commit_locks
+                    .get(obj_ref)
+                    .copied()
+                    .or_else(|| epoch_store.get_quarantined_owned_object_lock(obj_ref))
+                {
+                    Some(lock) => Some(lock),
+                    None => epoch_store.tables()?.get_locked_transaction(obj_ref)?,
+                };
+                if let Some(other) = existing_lock {
+                    if other != digest {
+                        fatal!(
+                            "already-executed transaction {:?} has owned input {:?} \
+                             locked by a different transaction {:?}",
+                            digest,
+                            obj_ref,
+                            other,
+                        );
+                    }
+                }
+                current_commit_locks.insert(*obj_ref, digest);
+            }
+            debug!(
+                ?digest,
+                num_owned_inputs = owned_inputs.len(),
+                "Transaction already executed; retained as checkpoint root, skipping re-validation"
+            );
+            // keep[i] stays true so the transaction remains in the sequence.
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -77,8 +80,8 @@ use crate::{
 ///
 /// `Ok((dropped, locks))` where:
 /// - `dropped` — `(digest, error)` for every semantically-invalid or
```
piotrm50, line 83: Is it valid? @cyberphysic4l can you verify that such approach is valid in the context of the checkpoint-stall issue you found?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -119,12 +122,48 @@ pub async fn validate_and_resolve_conflicts(
...
         let digest = *transaction.digest();
 
-        // Check #1: Already executed — silent dedup.
+        // Check #1: Already executed (typically by state-sync before this node's
+        // consensus handler reached the commit). It is a committee-agreed winner, so
+        // keep it in the sequence to flow into checkpoint roots like on every other
+        // validator (dropping it forks — issue #11649). Register its owned-object
+        // locks so double-spend siblings still lose, then skip re-validation (#2/#5);
+        // `TransactionManager::enqueue` suppresses the re-execution.
         if authority_state
```
piotrm50, line 132: Can it ever happen that the transaction is removed from the transaction cache (thus does not fall into this branch) and later fails object version validation?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -119,12 +122,48 @@ pub async fn validate_and_resolve_conflicts(
...
             .get_transaction_cache_reader()
             .try_is_tx_already_executed(&digest)?
         {
-            keep[i] = false;
+            // Byte-based, so safe even though the inputs are already consumed.
+            let owned_inputs = extract_owned_input_objects(tx)?;
```
piotrm50, line 137: let's extract this into an own function and call it here and below.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -119,12 +122,48 @@ pub async fn validate_and_resolve_conflicts(
...
+                    }
+                }
+                current_commit_locks.insert(*obj_ref, digest);
+            }
+            debug!(
+                ?digest,
+                num_owned_inputs = owned_inputs.len(),
+                "Transaction already executed; retained as checkpoint root, skipping re-validation"
+            );
+            // keep[i] stays true so the transaction remains in the sequence.
```
piotrm50, line 161: I think we can anyway run part of the checks such as:
* iota_transaction_checks::deny::check_transaction_for_validation
* check_coin_deny_list_v1
* check_transaction_inputs_for_validation - we just check the ownership here, right? If we also check the versions are latest and not yet consumed then we can't do this in a full form. 

MoveAuthenticator checks need to be skipped because their outcome depends on the current on-chain state, which (with state sync execution) is not updated at the pace of consensus commit creation and the state against which we're executinig authentication can be already in the future due to StateSync execution. 

Can we know if a transaction has been executed through StateSync? Are there any other ways in which a transaction is executed and updates the state other than from StateSync and consensus commit output?

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -119,12 +122,48 @@ pub async fn validate_and_resolve_conflicts(
...
+                        fatal!(
+                            "already-executed transaction {:?} has owned input {:?} \
+                             locked by a different transaction {:?}",
+                            digest,
+                            obj_ref,
+                            other,
+                        );
+                    }
+                }
+                current_commit_locks.insert(*obj_ref, digest);
```
cyberphysic4l, line 154: Is there some way that we would expect this consistency violation to actually happen? If the tx is already executed, doesn't that just always mean we are doing state-sync and this stuff has already been checked? Or can an already-executed transaction from a previous commit make it this far under the normal flow in which case it should be removed?

crates/iota-core/src/post_consensus_validation.rs
cyberphysic4l: All seems fine to me and does not interfere with the fix for the checkpoint-stall. If I understand correctly how state sync works, these retained already-executed transactions don't need to have waiter notified because they haven't got to this point via the normal submission to consensus path so there is no semaphore to release etc. Only the dropped (rather than retained) transactions need to notify a waiter as done in my fix.

crates/iota-core/src/post_consensus_validation.rs
piotrm50: We have to acquire locks on this transaction like we normally would. Otherwise a conflicting transaction from the same commit could still go through unnoticed. A transaction that is executed via StateSync needs to go through the same post-consensus flow once it's included in consensus commit (a different, parallel flow that usually happens before state sync, but sometimes can happen after state sync). If an existing lock on an object belongs to a different transaction than the one that has been executed via StateSync, we have a problem and need to panic (hence `fatal!`).

crates/iota-core/src/post_consensus_validation.rs
piotrm50: no, it can't

crates/iota-core/src/post_consensus_validation.rs
piotrm50: done

crates/iota-core/src/post_consensus_validation.rs
piotrm50: no point in doing that. The only way for a transaction to be executed before it's output by consensus is StateSync. If a transaction fails validation post consensus after it's executed and certified by the quorum then we would fork. Doing those checks here does not increase safety. Acquiring locks as is done in this PR is enough.

piotrm50 (approved):

cyberphysic4l (approved):

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
```diff
@@ -996,3 +1004,545 @@ async fn test_dropped_tx_does_not_acquire_locks() {
...
+async fn setup_lock_tier() -> LockTierSetup {
+    let _guard = ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
+        config.set_enable_white_flag_flow_for_testing(true);
+        config
+    });
+    // Hold the guard for the test by leaking it — every test in this section
+    // needs white-flag enabled; the guard is dropped automatically when the
+    // test process exits. (`apply_overrides_for_testing` returns a guard that
+    // reverts on drop; leaking is safe here because each setup re-applies.)
+    std::mem::forget(_guard);
```
semenov-vladyslav: Why do you have to `forget`? Other tests do fine without forgetting it. The part "leaking is safe here because each setup re-applies" is confusing since "re-application" will panic (`assert!(cur.is_none(), "config override already present");`). Note, another test may (unintentionally) reuse the same OS thread and thus the protocol config override.

semenov-vladyslav (changes requested):
Looks good except for forgetting protocol config override in tests.

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
piotrm50: good catch. this needs to be dropped at the end of the test, not leaked. changed now.

semenov-vladyslav (approved):

PR #11672, opened 2026-05-28
fix(iota-core): notify for post-consensus dropped transactions

cyberphysic4l:
# Description of change

A bug was observed during stress testing in the event of any post-consensus drop that ultimately results in checkpoint creation stalling. The problem was that when a transaction is submitted to consensus, a semaphore is taken for the inflight transaction, and if it was dropped rather than completing processing, the waiter was not notified and the semaphore was held until the end of the epoch. This problem only reveals itself when the consensus submissions semaphores run out, preventing any new transactions being submitted, including checkpoint signatures, which results in checkpoint creation stalling if there is no longer a quorum of validators with semaphores available to submit their signatures.

The fix is simply to ensure the waiters are notified in the event of a post-consensus drop.

The problem of checkpoint stalling during stress tests is mentioned in #11649 and is likely a result of this bug.

The problem is no longer observed during stress testing with this patch.

### DIFF
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -1074,15 +1074,9 @@ impl ConsensusAdapter {
         }
 
         let processed_methods = notifications.collect::<Vec<ProcessedMethod>>().await;
-        if processed_methods
-            .iter()
-            .any(|m| *m == ProcessedMethod::Dropped)
-        {
+        if processed_methods.contains(&ProcessedMethod::Dropped) {
             ProcessedMethod::Dropped
-        } else if processed_methods
-            .iter()
-            .any(|m| *m == ProcessedMethod::Checkpoint)
-        {
+        } else if processed_methods.contains(&ProcessedMethod::Checkpoint) {
             ProcessedMethod::Checkpoint
         } else {
             ProcessedMethod::Consensus
--- a/crates/iota-core/src/consensus_adapter.rs
+++ b/crates/iota-core/src/consensus_adapter.rs
@@ -719,7 +719,13 @@ impl ConsensusAdapter {
             }
 
             // If transaction is received by consensus or checkpoint while we wait, we are done.
-            _ = &mut processed_via_consensus_or_checkpoint => {
+            // Capture the resolved `ProcessedMethod` so the latency metric and
+            // `latency_observer` accurately reflect whether the early-fire was a
+            // consensus processing, a checkpoint sync, or a `dropped_tx_status_cache`
+            // hit. Without this, the guard would stay at its default `Consensus`,
+            // mislabeling cache-hit retries.
+            method = &mut processed_via_consensus_or_checkpoint => {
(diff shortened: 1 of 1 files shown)


crates/iota-core/src/post_consensus_validation.rs
piotrm50, line 106: this fix intentionally skips this drop because this is a source of another problem resolved in another PR https://github.com/iotaledger/iota/issues/11649

piotrm50 (approved):

crates/iota-node/src/lib.rs
```diff
@@ -1169,14 +1180,34 @@ impl IotaNode {
             &validator_registry,
         );
 
+        let soft_locks = Arc::new(PreConsensusSoftLocks::new());
```
bingyanglin: Now the ttl is always 12s (default value), do you think we need to make it configurable through `ConsensusConfig` or not necessary?

crates/iota-core/src/authority_server/soft_lock.rs
```diff
@@ -0,0 +1,755 @@
...
+//! | Tx dropped in post-consensus     | Released in `authority_per_epoch_store` after quarantine               |
+//! | Tx forgotten by consensus        | TTL expiry releases locks via background sweep                        |
+//! | Consensus submission fails       | Locks released immediately in error path                              |
+//! | Crash / restart                  | All soft locks lost → clean slate; post-consensus is authoritative    |
+//! | Epoch boundary                   | Same instance is `clear()`-ed; all locks dropped                      |
+//! | Object version mismatch          | Keyed on full `ObjectRef`; different versions don't conflict          |
+//!
+//! # TTL derivation
+//!
+//! Default TTL = `gc_depth(60) × 4 / ~20 rounds/sec = 12 s`.
```
bingyanglin: You're tracking `num_rejected_tx_soft_lock_conflict` which is good for counting conflicts. Do you think it is good to have a histogram of lock hold times would help operators know if 12s is the right TTL for their deployment? Or we are pretty confident 12s is optimal?

bingyanglin (approved):

crates/iota-core/src/consensus_adapter.rs
```diff
@@ -1031,12 +1037,13 @@ impl ConsensusAdapter {
...
-            // aggregate when all transactions are processed, but with this approach can get
-            // notified in a more fine-grained way as transactions can be marked
-            // as processed in different ways. This is mostly a concern for the soft-bundle
-            // transactions.
+            // We wait for each transaction individually to be processed by consensus,
+            // executed in a checkpoint or dropped. We could equally just get
+            // notified in aggregate when all transactions are processed, but
+            // with this approach can get notified in a more fine-grained way as
+            // transactions can be marked as processed in different ways. This
+            // is mostly a concern for the soft-bundle transactions.
```
piotrm50, line 1045: We removed the soft bundles in PCOOL, but I assume this still is relevant for PCOOL. Comment should be updated.

PR #11673, opened 2026-05-28
refactor(release-worklfow): Use gh api for homebrew release checksum

lzpap:
Use gh api for homebrew release checksum instead of no longer available custom action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -277,15 +277,14 @@ jobs:
           echo iota_tag="$(echo ${{ github.ref }} | sed s/'refs\/tags\/'//)" >> $GITHUB_OUTPUT
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Get binary checksums
-        uses: thewh1teagle/checksum@6577963a29f57d780d12a2e5495601298ba739d1 # v2
-        with:
-          tag: ${{ steps.tag.outputs.iota_tag }}
-          # These two parameters are needed to make the resulting checksum file work with shasum256 -c
-          reverse-order: true
-          separator: " "
-
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.tag.outputs.iota_tag }}
+        run: |
+          gh api "repos/${{ github.repository }}/releases/tags/${TAG}" \
+            --jq '.assets[] | select(.name != "checksum.txt") | "\(.digest | sub("^sha256:"; ""))  \(.name)"' \
+            > checksum.txt
+          cat checksum.txt
       - run: ./scripts/homebrew/update_formula.sh ${{ steps.tag.outputs.iota_tag }}
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -277,15 +277,14 @@ jobs:
...
-          # These two parameters are needed to make the resulting checksum file work with shasum256 -c
-          reverse-order: true
-          separator: " "
-
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.tag.outputs.iota_tag }}
+        run: |
+          gh api "repos/${{ github.repository }}/releases/tags/${TAG}" \
+            --jq '.assets[] | select(.name != "checksum.txt") | "\(.digest | sub("^sha256:"; ""))  \(.name)"' \
```
thibault-martinez, line 285: Looks good but I'm just wondering if we can't just download `checksum.txt` (https://github.com/iotaledger/iota/releases/download/v1.23.2/checksum.txt) since it seems to exist and already be the format we want?

thibault-martinez (approved):

.github/workflows/release.yml
thibault-martinez: https://github.com/iotaledger/iota/pull/11673/commits/7c32320de6a08f8b0f2fd18f49d1efd2ec07c48d

DaughterOfMars (approved):

.github/workflows/release.yml
```diff
@@ -277,15 +277,10 @@ jobs:
...
-        with:
-          tag: ${{ steps.tag.outputs.iota_tag }}
-          # These two parameters are needed to make the resulting checksum file work with shasum256 -c
-          reverse-order: true
-          separator: " "
-
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.tag.outputs.iota_tag }}
+        run: gh release download ${TAG} --repo ${{ github.repository }} --pattern checksum.txt
```
Thoralf-M: where is the checksum.txt generated?

thibault-martinez: It's a file available on releases, e.g. https://github.com/iotaledger/iota/releases/download/v1.23.2/checksum.txt

Thoralf-M: Only later in update_formula.sh, no?

thibault-martinez: Wait, it's not?

Thoralf-M: I mean the `run:` below generates it

thibault-martinez: No it does not, the run below consumes it

Thoralf-M: Ah true, but where does it get created then? Isn't that part removed now?

Thoralf-M (approved):

.github/workflows/release.yml
thibault-martinez: Actually the file does not exist, the previous action was uploading it, I wasn't aware.

PR #11685, opened 2026-05-29
feat(docs,examples): Add a new example for an Abstract Account that can sponsor programmatically

### LINKED ISSUE #11667: [AA] Create a doc page with an example for the new sponsor flow

miker83z:
# Description of change

Adds a new AA example related to sponsorship.

### 1. New `whitelist_sponsorship` package demonstrating a **policy-driven sponsor AA**:

- `WhitelistSponsorshipAccount` — dedicated AA type with admin-managed whitelists of (a) accepted sender authenticator function refs (`Bag` keyed by `(package, module, function)` so different sender account types can coexist) and (b) per-user gas allowances (`Table<address, u64>`); plus a cached `package_addr` field set once at `create` via `type_name::get` so the authenticator skips reflection on every call. The example is split into two modules: `whitelist_sponsorship_account` (storage + admin) and `whitelist_sponsorship_authentication` (the `#[authenticator]` and its PTB-scan helper).
- intent-aware sponsor authenticator that checks the tx is sponsored by this account, that the sender uses a whitelisted `MoveAuthenticator` function (looked up via `auth_ctx.sender_authenticator_function_info_v1()`), that `ctx.gas_budget()` fits the sender's allowance, and that the PTB contains a matching `deduct_user_gas_allowance(&mut sponsor)` command — the call takes no user/amount arguments, implicitly targeting `ctx.sender()` and deducting exactly `ctx.gas_budget()`, so the authenticator only needs to confirm the call's existence.
- Walkthrough `README.md` that chains off `onesig/README.md` to demonstrate a OneSig **sender** × `WhitelistSponsorshipAccount` **sponsor** sponsored transaction.
- `public_key_iotaccount::sponsorship_ed25519_authenticator` — a **signature-driven sponsorship primitive** that binds the sponsor's signature to `ctx.digest() || auth_ctx.sender_auth_digest() || bcs(sender_authenticator_function_info_v1)`.

### 2. Add AA sponsorship docs

- How-to: [Sponsor Transactions with a Whitelist and Per-User Gas Allowances](docs/content/developer/move/how-tos/account-abstraction/authenticator/create-whitelist-sponsorship-account.mdx) — walks the authenticator, PTB-scan helper, account creation, admin operations, and the in-PTB `deduct_user_gas_allowance` requirement.
- Security: [Binding Sponsor Approval to Sender Identity](docs/content/developer/move/how-tos/account-abstraction/security/sender-identity-binding.mdx) — motivates the threat model from the standard protocol-created sponsor case (whose signature only covers `ctx.digest()`, leaving no Move-side lever), then shows how an AA sponsor closes the gap via the canonical helper (`authenticate_ed25519_for_sponsorship`) that gets the signed byte layout right.
- Explanation: a new `## Sponsor authenticators authorize someone else's intent` design principle covering (a) the two sponsor kinds (standard protocol-created account vs. Abstract Account), (b) the pre-/post-consensus phase asymmetry (sponsor verification runs at both phases; the sender's Move authenticator only post-consensus, on the sponsor's gas budget), and (c) the contrast between policy-driven and signature-driven sponsor AAs. Plus a `:::tip` callout in the AA introduction pointing readers at the sponsorship story.

Fixes #11667


examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelists.move
```diff
@@ -0,0 +1,199 @@
...
+    package: ID,
+    module_name: ascii::String,
+    function_name: ascii::String,
+}
+
+/// A dynamic field name for the whitelist of accepted sender authenticator functions.
+public struct AuthenticatorFunctionsWhitelistFieldName has copy, drop, store {}
+
+/// A dynamic field name for the per-user sponsored gas allowance table.
+public struct UserGasAllowancesFieldName has copy, drop, store {}
```
valeriyr: nit: not so important for the example, but looks like AuthenticatorFunctionsWhitelist and UserGasAllowances each can have a dedicated module because they look independent.

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
```diff
@@ -0,0 +1,400 @@
...
+) {
+    let account_id = account.borrow_uid();
+
+    // Check if the account was setup.
+    assert!(whitelists::has_whitelists(account_id), EWhitelistsMissing);
+
+    // Check if the transaction is sponsored by this account.
+    let sponsor_opt: Option<address> = ctx.sponsor();
+    assert!(sponsor_opt.is_some(), ENotASponsoredTransaction);
+    assert!(sponsor_opt.destroy_some() == account_id.to_address(), ENotASponsoredTransaction);
```
valeriyr: It could make sense to add a dedicated error, since the transaction is sponsored by another account

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
```diff
@@ -0,0 +1,400 @@
...
+
+/// Deducts `amount` from `user`'s gas allowance on this sponsor account. Intended to be called
+/// from the sender's PTB during a sponsored transaction so the sender's allowance is reduced by
+/// the gas budget. `authenticator` scans the PTB for calls to this exact function.
+///
+/// Only the `user` themselves or the admin can call this.
+public fun deduct_user_gas_allowance(
+    self: &mut WhitelistSponsorshipAccount,
+    user: address,
+    amount: u64,
```
valeriyr: Looks like we can use the gas budget instead of having the `amount` parameter.

valeriyr (approved):
Looks good!

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelists.move
miker83z: I refactored the Account type, now the fields are part of it.

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
miker83z: Now that check was refactored.

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
miker83z: Removed `user: address, amount: u64,` parameters using ctx fields instead.

valeriyr (approved):

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_authentication.move
```diff
@@ -0,0 +1,150 @@
...
+    let key = key_from_info(sender_info_opt.borrow());
+    assert!(
+        account.is_authenticator_function_whitelisted(key),
+        EAuthenticatorFunctionNotWhitelisted,
+    );
+
+    // Check that the transaction gas budget fits within the sender's allowance.
+    // `borrow` itself aborts (with `iota::dynamic_field::EFieldDoesNotExist`) when the sender
+    // has no entry, so we skip an explicit `contains` check to save one `df::exists_`.
+    let allowances = account.borrow_user_gas_allowances();
```
Dkwcs, line 89: Is the allowance intended to be consumed only for successfully executed PTBs? 
Since `deduct_user_gas_allowance` is a common PTB command, any later abort should roll back the allowance update due to PTB atomicity, while the sponsor still pay gas. 
Does this mean a malicious whitelisted sender can repeatedly submit aborting sponsored transactions without reducing allowance?

For example:

1. The sender has an allowance of 100_000_000.
2. The sender creates a sponsored PTB with the following commands:
      1) command 0: deduct_user_gas_allowance(account)
      2) command 1: expensive_operation_that_aborts()
3. The sponsor authenticator checks the transaction and sees that:
      1) the required deduct_user_gas_allowance call is present
      2) the declared gas_budget does not exceed the sender’s allowance.
4. The transaction passes authentication.
5. During execution we encounter abort(the PTB aborts, and all PTB effects are rolled back).
6. As a result, the allowance is not decreased.
7. However, the sponsor may still be charged for the gas consumed by the failed transaction(if that works).
8. The sender can potentially repeat this flow and consume sponsor gas without reducing their allowance.

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
```diff
@@ -0,0 +1,232 @@
...
+/// Deducts the transaction's `gas_budget` from the **sender's** gas allowance on this sponsor
+/// account. Intended to be called from the sender's PTB during a sponsored transaction so the
+/// sender's allowance is reduced by exactly the gas budget the sponsor will pay.
+///
+/// The sponsor authenticator scans the PTB for the presence of this call. Because the function
+/// implicitly targets `ctx.sender()` and always deducts `ctx.gas_budget()`, the authenticator
+/// only needs to confirm such a call exists — it doesn't have to read or compare any arguments
+/// besides the sponsor account itself.
+public fun deduct_user_gas_allowance(self: &mut WhitelistSponsorshipAccount, ctx: &TxContext) {
+    let sender = ctx.sender();
```
Dkwcs, line 104: Should `deduct_user_gas_allowance` assert that sponsor exists to avoid accidental allowance burn in non-sponsored transactions?

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_authentication.move
miker83z: Yes this is a true possibility and I don't see how it could be fixed without a post-execution command/script (i.e., deduct the real gas usage post execution) or with an authenticator that mutates the state (a similar function to deduct_user_gas_allowance would be called within the authenticator execution).

miker83z: The sponsor can block this attack by manually lowering the allowance with a follow up TX, after that the first TX failed.

examples/move/abstract_iota_accounts/whitelist_sponsorship/sources/whitelist_sponsorship_account.move
miker83z: I guess that's fine like this, it's not needed to limit it so much for a remote possibility of an accidental invocation

Dkwcs (approved):

PR #11689, opened 2026-05-29
feat(grpc,core): Apply shared traffic controller to gRPC API and refactor

### LINKED ISSUE #11668: task: add traffic controller to gRPC
1. Add traffic_controller: Option<Arc<TrafficController>> and client_id_source: Option<ClientIdSource> parameters to start_grpc_server; pass state.traffic_controller.clone() and config.policy_config.as_ref().map(|p| p.client_id_source) from build_grpc_server.
2. New TrafficControlLayer in iota-grpc-server/src/ mirroring GrpcMetricsLayer: extract client IP (reuse the helper logic from authority_server.rs:1019-1108), call check, dispatch, then call tally with a tonic::Status-to-Weight mapper analogous to normalize in axum_router.rs:281.
3. Ensure ConnectInfo<SocketAddr> is available on requests — TcpListenerStream in serve_with_incoming_shutdown doesn't add it automatically; tonic's Server::serve_with_incoming_connect_info does. May need a small switch.
4. Decide stream semantics for StreamCheckpoints: tally once at stream open, not per message — easiest done by checking/tallying in the layer, not inside the streaming handler.

jkrvivian:
# Description of change

* Wires the shared `TrafficController` into the gRPC server as a tower
`Layer`, so the same blocklist and tally apply to a given client IP
regardless of whether it hits the JSON-RPC or gRPC surface.
* Client-IP extraction is factored out of `authority_server.rs` and reused by the
new layer.
* Includes follow-up cleanup on the recent traffic-controller port: https://github.com/iotaledger/iota/pull/11598

Closes #11668

### DIFF
--- a/crates/iota-core/src/authority_server.rs
+++ b/crates/iota-core/src/authority_server.rs
@@ -46,10 +46,7 @@ use prometheus::{
     register_int_counter_with_registry,
 };
 use tap::TapFallible;
-use tonic::{
-    metadata::{Ascii, MetadataValue},
-    transport::server::TcpConnectInfo,
-};
+use tonic::transport::server::TcpConnectInfo;
 use tracing::{Instrument, debug, error, error_span, info, trace_span, warn};
 
 use crate::{
@@ -59,7 +56,9 @@ use crate::{
         ConnectionMonitorStatusForTests, ConsensusAdapter, ConsensusAdapterMetrics,
     },
     starfish_adapter::LazyStarfishClient,
-    traffic_controller::{TrafficController, parse_ip, policies::TrafficTally},
+    traffic_controller::{
+        ClientIpStatus, TrafficController, get_client_ip, policies::TrafficTally,
+    },
 };
 
 #[cfg(test)]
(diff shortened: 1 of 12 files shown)


crates/iota-grpc-server/src/traffic_control.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        // Tower contract: the cloned service is the one ready to call, so swap
+        // the clone in and keep the previously-ready inner for this request.
+        let cloned = self.inner.clone();
+        let mut inner = std::mem::replace(&mut self.inner, cloned);
+
+        Box::pin(async move {
+            if !traffic_controller.check(&client, &None).await {
+                let response = Status::resource_exhausted("Too many requests").into_http();
+                tally(&traffic_controller, client, Code::ResourceExhausted);
```
bingyanglin: We tally the request even when it's blocked. Every other place skips the tally on a block (in Sui's JSON-RPC, the validator, and even our own JSON-RPC layer in `axum_router.rs` all return early without tallying). The problem is tallying a blocked client keeps it over the threshold, so it renews its own block and never ages out. And if firewall delegation is on, we make an outbound firewall call for every blocked request. Maybe drop the tally here and only tally on the non-blocked path below. Any reason for adding this code?

crates/iota-grpc-server/src/traffic_control.rs
```diff
@@ -0,0 +1,161 @@
...
+            ClientIpStatus::Ok(ip) => Some(ip),
+            other => {
+                warn!(
+                    "Skipping traffic controller request handling for client {remote_addr:?}: {other:?}"
+                );
+                None
+            }
+        };
+
+        // Tower contract: the cloned service is the one ready to call, so swap
```
bingyanglin: The comment is backwards. It says the clone is the ready one, but it's the original `self.inner` that's ready (poll_ready` set it up), and that's the one we call here. The code is fine though, only the comment is wrong. Maybe just flip it: "`self.inner` is ready, so call it and leave the clone for the next request."

piotrm50 (approved):

crates/iota-grpc-server/src/traffic_control.rs
jkrvivian: You're right, I also double checked how it's done in json rpc, and we should follow the same behavior.

bingyanglin (approved):

crates/iota-grpc-server/src/traffic_control.rs
```diff
@@ -0,0 +1,159 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Tower [`Layer`] that wires the shared [`TrafficController`] into the
+//! gRPC server.
+//!
+//! The fullnode-facing gRPC API exposes the same class of traffic that the
+//! JSON-RPC API protects, including transaction execution endpoints. We share
+//! one controller across both surfaces so the same blocklist and tally apply
+//! to a given client IP regardless of which API it hits.
+//!
```
muXxer:
```suggestion
```

crates/iota-grpc-server/src/traffic_control.rs
```diff
@@ -0,0 +1,159 @@
...
+//! The layer extracts the client IP via [`get_client_ip`] (shared with the
+//! validator), calls [`TrafficController::check`][check] before dispatching,
+//! then calls [`tally`][tally] with a weight derived from the response's
+//! gRPC status code.
```
muXxer: This can also be simplified to something like "The layer extracts the client IP and calls the traffic controller with a weight derived from the responses gRPC status code."

crates/iota-core/src/authority_server.rs
```diff
@@ -1019,101 +1018,68 @@ impl ValidatorService {
...
-                }
+            ClientIpStatus::XForwardedForHeaderMissing => {
+                self.metrics.forwarded_header_not_included.inc();
+                error!(
+                    "x-forwarded-for header not present for request despite node configuring x-forwarded-for tracking type"
+                );
+                None
+            }
+            ClientIpStatus::XForwardedForInvalidUtf8 => {
+                // TODO: once we have confirmed that no legitimate traffic
```
muXxer, line 1058: Please add a tracking issue

muXxer (changes requested):

muXxer (approved):

PR #11691, opened 2026-05-29
feat(consensus): StarfishSpeed — optimistic-commit consensus variant

VorobyevIlya:
# Description of change

Adds **StarfishSpeed**, a new variant of the Starfish DAG consensus
protocol gated behind the `consensus_starfish_speed` protocol flag.
When the flag is active, validators classify each committed leader by
a metastate (optimistic / standard / pending) and, in the optimistic
case, additionally sequence the leader's acknowledgments alongside
its historical sub-DAG — shortening the path from transaction
submission to commit at the cost of a stricter block-creation rule (a
strong-vote quorum requirement). A soft-timeout fallback lets the
validator propose under the base Starfish rule when the strong-vote
quorum cannot be reached in time, preserving liveness.

## What the branch introduces

In dependency order:

- **Protocol & versioned types.** A new protocol flag
  `consensus_starfish_speed`; a `BlockHeaderV2` carrying a
  `strong_vote` payload (the pinned leader authority plus the set of
  authorities whose transactions are not locally available); a
  `CommitV3` carrying an `is_optimistic` flag and the transactions
  sequenced alongside the leader in the optimistic case. Both
  versioned types are activated together via a version-flag check.
- **Data-availability tracking.** `DagState` tracks which transactions
  are locally available per past block, so the `missing` set in the
  strong vote can be computed accurately during proposal.
- **Block proposal.** Strong-vote computation; a strong-vote quorum
  requirement for non-forced block creation; the soft-timeout fallback
  described above.
- **Commit path.** Leader-metastate classification at commit time, and
  Optimistic-metastate ack commit in the linearizer. Strong votes and
  blames are pinned to the leader's authority so a leader-schedule
  rotation does not misattribute them.
- **Recovery & operator controls.** Recovery seeding of the ack
  tracker from stored block headers; adaptive acknowledgment filtering
  that drops acks for authorities persistently blamed by recent
  strong-vote masks (with a local-only enable switch so operators can
  disable it without a protocol change).
- **Observability.** Metrics for strong-vote extra wait,
  missing-authorities distribution, outbound and inbound blames by
  authority, adaptive-ack exclusion size, and dropped acks. A new
  Grafana dashboard (`dev-tools/grafana-local/dashboards/starfish-overview.json`)
  visualizes these metrics alongside the existing Starfish ones.

## Backward compatibility

The `consensus_starfish_speed` protocol flag is not enabled on any
network yet, so this PR is a no-op everywhere.

## Reviewing

Every commit in this branch is a previously-merged, already-reviewed PR (linked from the commit subject); this PR rolls those sub-PRs back into `develop` with no new code on top.

Coverage added on top of the existing Starfish test suite:

- `base_committer_metastate_tests.rs` — a new ~930-line suite exercising
  the optimistic / standard / pending metastate classification across
  DAG shapes (strong-vote quorum reached, missing-authority blames,
  rotated leader schedules, recovery boundaries).
- Phantom-ack regression test guarding the invariant that acks
  committed in the optimistic path cannot reference transactions
  outside the leader's sub-DAG.
- Extensions to the existing `base_committer_tests`,
  `pipelined_committer_tests`, and `universal_committer_tests` to
  cover StarfishSpeed-specific commit shapes.
- Multi-node behaviour is covered by the consensus simtests
  (`#[sim_test]`) that run under `cargo simtest`.

### DIFF
--- a/crates/iota-open-rpc/spec/openrpc.json
+++ b/crates/iota-open-rpc/spec/openrpc.json
@@ -1315,6 +1315,7 @@
                 "consensus_round_prober": false,
                 "consensus_round_prober_probe_accepted_rounds": false,
                 "consensus_smart_ancestor_selection": false,
+                "consensus_starfish_speed": false,
                 "consensus_zstd_compression": false,
                 "convert_type_argument_error": true,
                 "dependency_linkage_error": false,
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -492,6 +492,10 @@ struct FeatureFlags {
     // If true, only sponsor Move authentication is performed pre-consensus.
     #[serde(skip_serializing_if = "is_false")]
     pre_consensus_sponsor_only_move_authentication: bool,
+
+    // If true, enables the optimistic commit rule (StarfishSpeed) in Starfish consensus.
+    #[serde(skip_serializing_if = "is_false")]
+    consensus_starfish_speed: bool,
 }
 
 fn is_true(b: &bool) -> bool {
@@ -1731,6 +1735,10 @@ impl ProtocolConfig {
             );
         }
         pre_consensus_sponsor_only_move_authentication
+    }
+
+    pub fn consensus_starfish_speed(&self) -> bool {
+        self.feature_flags.consensus_starfish_speed
     }
 }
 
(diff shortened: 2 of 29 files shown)


tomxey (approved):

bingyanglin (approved):
Approved `crates/iota-protocol-config/src/lib.rs` and `starfish-overview.json`

polinikita (approved):

PR #11702, opened 2026-06-01
feat(iota-json-rpc-types): own JSON-RPC serialization for primitive types

### LINKED ISSUE #11503: Disconnect JSON-RPC serde definitions from SDK ones
- Add missing types to JSON crates
- Impl JSON serde for all types used in JSON-RPC
- Remove custom JSON impls from SDK (when BCS is standard)

DaughterOfMars:
## Summary

A prior change moved the `schemars::JsonSchema` layer into `iota-json-rpc-types` so the JSON schema definitions need not live in the external `iota-sdk-types` crate. This PR is the matching second half: **the JSON serialization is now defined in `iota-json-rpc-types` itself**, rather than falling through to the SDK types' native `Serialize`/`Deserialize` impls. After this, the crate owns its full JSON-RPC wire format and the external SDK is free to drop those serde impls.

## Changes

- **Adapters own serde** ([`iota_primitives.rs`](crates/iota-json-rpc-types/src/iota_primitives.rs)): added `SerializeAs`/`DeserializeAs` to every primitive marker (`IotaAddress`, `ObjectID`, `Base58`, `Base64`, `GenericSignature`, `Identifier`, `SequenceNumberU64`). String-like types reuse `serde_with::DisplayFromStr` — matching the SDK types' `Display`/`FromStr`, so the wire format is byte-identical; byte payloads reuse the `fastcrypto` encoders.
- **`StructTag`/`TypeTag` self-contained**: they now call the shared, IOTA-specific formatting/parsing helpers in `iota_types` directly (`to_iota_type_tag_string` made `pub`) instead of delegating to the `iota_serde` adapter types. The shared helpers stay in `iota_types` since many other crates depend on them.
- **Wired all field sites** (~145 across 14 files) with `#[serde_as(as = "...")]` alongside the existing `#[schemars(with = ...)]`.
- **Fixed the `next_epoch_*` validator pubkey fields**, which declared a base64 schema but serialized as JSON number-arrays, to serialize as base64 (consistent with their schema and sibling fields).

## Verification

- `cargo check` + `cargo clippy` clean; downstream `iota-json-rpc`, `iota-json-rpc-api`, `iota-indexer` compile.
- `iota-json-rpc-types` unit tests and `iota-types` serde round-trip tests pass.
- **OpenRPC spec example values are byte-for-byte unchanged**, proving the serialization is identical. The only spec delta is 7 benign `"default": null"` annotations (`serde_with` auto-adds `#[serde(default)]` to `Option` serde_as fields; `None` still serializes as `null`). The spec snapshot was refreshed.

Closes https://github.com/iotaledger/iota/issues/11503

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [x] JSON-RPC: The following `IotaValidatorSummary` fields were incorrectly serialized as byte arrays and are now correctly serialized as Base64: `next_epoch_authority_pubkey_bytes`, `next_epoch_proof_of_possession`, `next_epoch_network_pubkey_bytes`, `next_epoch_protocol_pubkey_bytes`
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


piotrm50 (approved):
approving iota-types change

Thoralf-M (approved):
Approving iota-types change

Alex6323 (approved):

crates/iota-open-rpc/spec/openrpc.json
```diff
@@ -3408,7 +3408,7 @@
         }
       ],
       "result": {
-        "name": "Page<Coin,ObjectIDSchema>",
+        "name": "Page<Coin,ObjectIdSchema>",
```
thibault-martinez, line 3411: I didn't do this in my ObjectId PR because I wasn't sure if it was breaking or not. How/when is this result name used?

crates/iota-graphql-rpc/src/types/balance_change.rs
```diff
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use async_graphql::*;
-use iota_json_rpc_types::BalanceChange as StoredBalanceChange;
+use iota_indexer::types::IndexedBalanceChange as StoredBalanceChange;
```
thibault-martinez: This alias was to avoid the clash with `BalanceChange` but there is no clash anymore, maybe up to @iotaledger/infrastructure to decide if they want to drop the `StoredBalanceChange` alias.

crates/iota-open-rpc/spec/openrpc.json
DaughterOfMars: It's not breaking, it's just a name

crates/iota-json-rpc-types/src/object_changes.rs
```diff
@@ -120,17 +133,20 @@ pub enum ObjectChange {
         #[schemars(with = "String")]
         #[serde_as(as = "IotaStructTag")]
```
Thoralf-M: Should this also be updated to `StructTagSchema`?

crates/iota-json-rpc-types/src/iota_system_state_summary.rs
```diff
@@ -687,12 +700,16 @@ pub struct IotaValidatorSummary {
     pub net_address: String,
     pub p2p_address: String,
     pub primary_address: String,
+    #[serde_as(as = "Option<Base64Schema>")]
```
Thoralf-M, line 703: This should probably be mentioned in the release notes as fix

crates/iota-json-rpc-types/src/object_changes.rs
DaughterOfMars: yep, nice catch

crates/iota-json-rpc-types/src/iota_system_state_summary.rs
DaughterOfMars: Good point

sergiupopescu199 (approved):
lgtm ✨

Also ran infra tests locally and all seems to work as expected.

crates/iota-graphql-rpc/src/types/balance_change.rs
sergiupopescu199: @DaughterOfMars @thibault-martinez We can drop it!

crates/iota-types/src/iota_serde.rs
```diff
@@ -137,7 +137,7 @@ pub fn to_iota_struct_tag_string(value: &StructTag) -> Result<String, fmt::Error
     Ok(f)
 }
 
-fn to_iota_type_tag_string(value: &TypeTag) -> Result<String, fmt::Error> {
+pub fn to_iota_type_tag_string(value: &TypeTag) -> Result<String, fmt::Error> {
```
thibault-martinez, line 140: Do we really not have such a thing in the SDK? If not, we should maybe add it?

crates/iota-json-rpc-types/src/object_changes.rs
```diff
@@ -112,25 +123,29 @@ pub enum ObjectChange {
...
     #[serde(rename_all = "camelCase")]
     Unwrapped {
+        #[serde_as(as = "IotaAddressSchema")]
         #[schemars(with = "IotaAddressSchema")]
         sender: IotaAddress,
         #[schemars(with = "OwnerSchema")]
         #[serde_as(as = "OwnerSchema")]
         owner: Owner,
-        #[schemars(with = "String")]
-        #[serde_as(as = "IotaStructTag")]
```
thibault-martinez, line 121: Was it only in this context that it was wrong? Because there are more instances of `#[serde_as(as = "IotaStructTag")]`

crates/iota-types/src/iota_serde.rs
DaughterOfMars: Not that I know, and I'm not sure it belongs there anyway

crates/iota-json-rpc-types/src/object_changes.rs
DaughterOfMars: Hmm, they are probably all wrong

thibault-martinez: So should we change them then?

DaughterOfMars: So the rest are all in iota-indexer, which is fine

thibault-martinez: Alright

crates/iota-json-rpc-types/src/iota_object.rs
```diff
@@ -1104,16 +1112,19 @@ impl From<IotaUpgradeInfo> for UpgradeInfo {
...
+    #[serde_as(as = "SequenceNumberU64Schema")]
     #[schemars(with = "SequenceNumberU64Schema")]
     pub version: SequenceNumber,
     #[schemars(with = "BTreeMap<String, Base64Schema>")]
     #[serde_as(as = "BTreeMap<_, Base64>")]
     pub module_map: BTreeMap<String, Vec<u8>>,
     #[schemars(with = "Vec<IotaTypeOrigin>")]
     pub type_origin_table: Vec<TypeOrigin>,
-    #[schemars(with = "BTreeMap<ObjectIDSchema, IotaUpgradeInfo>")]
+    #[serde_as(as = "BTreeMap<ObjectIdSchema, _>")]
```
thibault-martinez, line 1129: What is serde inferring for `_`? `IotaUpgradeInfo` or `UpgradeInfo`?

crates/iota-json-rpc-types/src/iota_primitives.rs
```diff
@@ -280,7 +424,9 @@ impl SerializeAs<NativeStructTag> for StructTag {
     where
         S: Serializer,
     {
-        IotaStructTag::serialize_as(value, serializer)
+        to_iota_struct_tag_string(value)
+            .map_err(S::Error::custom)?
+            .serialize(serializer)
```
thibault-martinez, line 429: can you explain the difference please?

thibault-martinez (approved):

crates/iota-json-rpc-types/src/iota_object.rs
DaughterOfMars: Good catch. It was inferring the SDK type. I found a few other instances too. In this case, I decided to replace the actual type in the map, because I think it makes more sense.

crates/iota-json-rpc-types/src/iota_primitives.rs
DaughterOfMars: There is none, it's just clearer I think

PR #11714, opened 2026-06-02
feat(move): Add local_implicit_deps for system packages

jkrvivian:
# Description of change

Fix simtest nightly CI errors: https://github.com/iotaledger/network-benchmark/actions/runs/26795993550/job/78992341956
Simtests depends on iota as a git dependency need to build their Move packages against the framework of the **commit under test**, not the moving `develop` branch. Resolving the framework over the network is both non-deterministic and fails in CI. This helper lets those consumers build offline, against the correct sources.

Adds `local_implicit_deps`, a helper that builds the implicit Move dependencies for the system packages (`Iota` framework, `MoveStdlib`, …) as **local-path** dependencies, rooted at this crate's location in the iota source tree (derived from the compile-time `CARGO_MANIFEST_DIR`).

Unlike the existing `implicit_deps`, which produces **git** dependencies pinned to a published framework revision, `local_implicit_deps` points at the framework sources of the *exact checkout this crate was compiled from*.

Also injects these implicit deps into the `#[sim_test]` macro's static-init build (`iota-proc-macros`), so a static-init package without an explicit `Iota` dependency still resolves the framework. For packages that declare one, injection is skipped (no-op).

### DIFF
--- a/crates/iota-move-build/src/lib.rs
+++ b/crates/iota-move-build/src/lib.rs
@@ -15,7 +15,7 @@ use anyhow::bail;
 use fastcrypto::encoding::Base64;
 use iota_package_management::{
     PublishedAtError, resolve_published_id,
-    system_package_versions::{SYSTEM_GIT_REPO, SystemPackagesVersion},
+    system_package_versions::{SYSTEM_GIT_REPO, SystemPackagesVersion, latest_system_packages},
 };
 use iota_sdk_types::ObjectId;
 use iota_types::{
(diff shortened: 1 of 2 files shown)


Thoralf-M (approved):

crates/iota-move-build/src/lib.rs
```diff
@@ -785,6 +785,42 @@ pub fn implicit_deps(packages: &SystemPackagesVersion) -> Dependencies {
...
+        .collect()
+}
+
+/// Useful for callers (e.g. the `#[sim_test]` static initializer) that don't
+/// have a [SystemPackagesVersion] on hand and just want the framework resolved
+/// from the local checkout.
+pub fn local_implicit_deps_latest() -> Dependencies {
+    local_implicit_deps(latest_system_packages())
+}
+
```
miker83z: Could you move these at the bottom of the file?

miker83z (approved):
LGTM!

crates/iota-move-build/src/lib.rs
jkrvivian: Moved and force-pushed!

PR #11718, opened 2026-06-02
feat(iota-types): replace `ObjectInner` with the external SDK `Object` type

### LINKED ISSUE #11631: Use `Object` from SDK

Alex6323:
# Description of change

This PR replaces `Object` / `ObjectInner` (previously defined in `iota-types`)
with the SDK version `iota_sdk_types::Object`. The node keeps its
`Object(Arc<ObjectInner>)` wrapper — the `Arc` is an in-memory cheap-clone
optimization specific to the node and not something SDK consumers need — and
re-exports the SDK type under the original name `ObjectInner`, so existing
struct-literal construction and the `#[serde(from = "ObjectInner")]` wrapper
continue to work unchanged.

Fixes #11631

## Changes

- Remove the local `ObjectInner` struct in [crates/iota-types/src/object.rs](crates/iota-types/src/object.rs) and re-export `iota_sdk_types::Object as ObjectInner`.
- Move the inherent accessor methods (`id`, `version`, `digest`, `is_shared`, `compute_object_reference`, `is_package`, …) from `impl ObjectInner` to `impl Object`. `ObjectInner` is now a foreign type, so it can no longer carry inherent methods; they remain reachable on `Object` exactly as before (previously via `Deref`, now directly on the wrapper), so call sites are unchanged.
- `Object::digest()` now hashes the inner SDK object (`self.as_inner()`), which carries the `BcsSignable`/`Signable` impl.
- Drop the now-redundant SDK→node `TryFrom<iota_sdk_types::Object>` conversion in [crates/iota-types/src/iota_sdk_types_conversions.rs](crates/iota-types/src/iota_sdk_types_conversions.rs): it collided with the `From<ObjectInner>` impl required by `#[serde(from)]`. The one checkpoint conversion that used it switches to the infallible `Object::from`, and a `try_into()?` in [crates/iota-indexer/src/types.rs](crates/iota-indexer/src/types.rs) becomes `Object::from`.
- Bump the `iota-sdk-types` git rev in [Cargo.toml](Cargo.toml) to pick up the SDK side of this change, which adds `Hash` to `Object` (iotaledger/iota-rust-sdk#1186).
- No BCS format-snapshot change: `ObjectInner` already serialized under the name `Object` (`#[serde(rename = "Object")]`), which matches the SDK type, so the registry and on-wire encoding are identical (verified by `test_object_digest_and_serialized_format` and `iota-core`'s `test_format`).

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `iota-types::object::ObjectInner` is now a re-export of
  `iota_sdk_types::Object`. The `Object` wrapper (`Arc<ObjectInner>`) is
  unchanged, and its accessor methods now live on `Object` itself rather than
  on `ObjectInner`. Code that called these methods on a bare `ObjectInner`
  value must go through `Object`, and the redundant
  `TryFrom<iota_sdk_types::Object> for Object` conversion is replaced by the
  infallible `From`. The BCS wire format is unchanged.
- [ ] gRPC:


Cargo.lock
```diff
@@ -11217,7 +11242,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "343d3bd7056eda839b03204e68deff7d1b13aba7af2b2fd16890697274262ee7"
 dependencies = [
  "heck 0.4.1",
- "itertools 0.14.0",
+ "itertools 0.10.5",
```
semenov-vladyslav: Why was it downgraded?

Cargo.lock
```diff
@@ -7411,7 +7411,32 @@ dependencies = [
 [[package]]
 name = "iota-sdk-types"
 version = "0.0.1-alpha.1"
-source = "git+https://github.com/iotaledger/iota-rust-sdk.git?rev=058f623870be7c8fb47d3d52564107eabc2c2f96#058f623870be7c8fb47d3d52564107eabc2c2f96"
+source = "git+https://github.com/iotaledger/iota-rust-sdk.git?rev=058f623870be7c8fb47d3d52564107eabc2c2f96#2bad99e86d173f938cf7c87dec928e5f8e26146c"
```
semenov-vladyslav: Multiple versions of iota-sdk-types doesn't look right.

semenov-vladyslav (changes requested):
Looks good except for `Cargo.lock` and a few clippy lints.

muXxer (approved):

semenov-vladyslav (approved):

crates/iota-types/src/object.rs
```diff
@@ -334,21 +336,6 @@ impl MoveObjectExt for MoveObject {
...
-    pub previous_transaction: TransactionDigest,
-    /// The amount of IOTA we would rebate if this object gets deleted.
-    /// This number is re-calculated each time the object is mutated based on
-    /// the present storage gas price.
-    pub storage_rebate: u64,
-}
-
 #[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash)]
 #[serde(from = "ObjectInner")]
 pub struct Object(Arc<ObjectInner>);
```
thibault-martinez, line 337: Why does this exist?

Alex6323: For performance reasons in the node. More specifically ... to prevent deep copies of large objects.

bingyanglin (approved):

kodemartin (approved):

valeriyr (approved):

Thoralf-M (approved):

DaughterOfMars (approved):

thibault-martinez (changes requested):
Some methods can/should be moved to the SDK

Alex6323:
> Some methods can/should be moved to the SDK

which ones?

Thoralf-M (approved):

thibault-martinez (approved):

Issue #11694, opened 2026-05-29
[data-ingestion{,-core}] store epoch_boundaries in historical checkpoint store


kodemartin:
## Description

Following the analysis https://github.com/iotaledger/iota/issues/11255#issuecomment-4565269248, it became clear that it would be nice for the historical checkpoint store to be a self-sufficient source of all the data available for verification.

As we are missing information on epoch boundaries, the idea would be extend the store with an additional file that maintains the sequence number of the last checkpoint of each epoch. This way the `HistoricalReader` will be able to lookup directly for the last checkpoints and accommodate verification.

To this end the following is proposed
1. A new file `EPOCH_BOUNDARIES` will complement what is stored in the historical checkpoint store: See `iota_data_ingestion_core::history`.
2. The ingestion worker `iota_data_ingestion::workers::historical` should be refactored in order
    * to optionally seed the epoch boundaries file upon restart. This will make the switch to the extended store smooth. 
    * to update the `EPOCH_BOUNDARIES` file on epoch's end

PR #11719, opened 2026-06-03
feat(data-ingestion-core): introduce EPOCH_BOUNDARIES

### LINKED ISSUE #11695: [data-ingestion-core]: store epoch boundaries in historical checkpoint store

kodemartin:
# Description of change

This patch introduces a new file `EPOCH_BOUNDARIES` in the historical checkpoint store.

The file maintains a map between epochs and the sequence number of the respective last checkpoint sequence number. For the underlying context see https://github.com/iotaledger/iota/issues/11694.

Introducing a new file was preferred to integrating this information into the existing `MANIFEST`  primarily to ease the initialization and make traversal of the boundaries more efficient. Conceptually this is warranted, as epoch boundaries is information that lies on a higher level than checkpoint batches.

Fixes #11695 

### DIFF
--- a/crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
+++ b/crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
@@ -6,14 +6,9 @@
 use std::{collections::BTreeMap, ops::RangeBounds};
 
 use bytes::Bytes;
-use fastcrypto::hash::{HashFunction, Sha3_256};
-use iota_storage::{
-    SHA3_BYTES,
-    blob::{Blob, BlobEncoding},
-    object_store::{
-        ObjectStoreGetExt, ObjectStorePutExt,
-        util::{exists, get, put},
-    },
+use iota_storage::object_store::{
+    ObjectStoreGetExt, ObjectStorePutExt,
+    util::{exists, get, put},
 };
 use iota_types::{committee::EpochId, messages_checkpoint::CheckpointSequenceNumber};
 use object_store::path::Path;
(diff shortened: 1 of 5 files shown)


crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
```diff
@@ -0,0 +1,204 @@
...
+///
+/// If the remote file is not found, this returns an empty collection.
+///
+/// # Errors
+///
+/// Fails if the file fails to decode.
+pub async fn read_epoch_boundaries_or_default<S: ObjectStoreGetExt>(
+    remote_store: S,
+) -> Result<EpochBoundaries> {
+    if !exists(&remote_store, &EpochBoundaries::file_path()).await {
```
sergiupopescu199, line 115: Nit: This function effectively downloads the file, and on line 116 we download it again. If possible we could leverage the [head](https://docs.rs/object_store/latest/object_store/trait.ObjectStoreExt.html#tymethod.head) method to check if a file exists without downloading its contents.

crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
```diff
@@ -0,0 +1,204 @@
...
+        return Ok(Default::default());
+    }
+    let bytes = get(&remote_store, &EpochBoundaries::file_path()).await?;
```
sergiupopescu199: The `get` function internally uses an `ExponetialBackoff`, with a max elapsed time of 15 minutes,in addition its internal `get_bytes()` trait implementation does disable all timeouts, inherited from upstream. It would be a good idea if we added a `tokio::timeout` to avoid the problem we encountered [here](https://github.com/iotaledger/iota/pull/8074)

crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
kodemartin: https://github.com/iotaledger/iota/issues/11726

crates/iota-data-ingestion-core/src/history/epoch_boundaries.rs
kodemartin: Good catch! 959b254cbc

sergiupopescu199 (approved):
lgtm 🏎️

tomxey (approved):

PR #11722, opened 2026-06-03
feat(data-ingestion): write epoch boundaries with historical worker

### LINKED ISSUE #11696: [data-ingestion]: store epoch_boundaries in historical checkpoint store

kodemartin:
# Description of change

This patch follows #11719 to implement the writing logic for the new `EPOCH_BOUNDARIES` file.

To ease the initialization in existing instances, the historical worker accepts a new optional configuration option in the form of an `epoch: checkpoint_sequence_number`. If the remote `EPOCH_BOUNDARIES` file is not present it is initialized from this seed. 

Exampl

```yaml
tasks:
  - name: "local-historical-storage"
    concurrency: 1
    historical:
      object-store-config:
        ...
      commit-duration-seconds: 250
      epoch-boundaries:
        0: 1
        1: 100
        2: 2000
```

Fixes #11696 

Tested locally with `localstack` configuration as described in https://github.com/iotaledger/iota/blob/develop/dev-tools/iota-data-ingestion/README.md for the following cases:

- Seeding the file on an existing instance
- Seeding the file from genesis 
- Starting from genesis without seed

### DIFF
--- a/crates/iota-data-ingestion/src/workers/historical.rs
+++ b/crates/iota-data-ingestion/src/workers/historical.rs
@@ -24,7 +24,6 @@ use iota_storage::{
     FileCompression, StorageFormat,
     blob::{Blob, BlobEncoding},
     compress,
-    object_store::util::exists,
 };
 use iota_types::{
     committee::EpochId, full_checkpoint_content::CheckpointData,
@@ -115,10 +114,8 @@ impl HistoricalReducer {
     /// Initializes the epoch boundaries from the provided seed if not already
     /// initialized.
     async fn seed_epoch_boundaries(&self, epoch_boundaries: EpochBoundaries) -> anyhow::Result<()> {
-        if !exists(&self.remote_store, &EpochBoundaries::file_path()).await {
-            write_epoch_boundaries(&epoch_boundaries, self.remote_store.clone()).await?;
-            tracing::info!("Initialized epoch boundaries");
-        }
+        write_epoch_boundaries(&epoch_boundaries, self.remote_store.clone()).await?;
+        tracing::info!("Initialized epoch boundaries");
         Ok(())
     }
 
--- a/crates/iota-data-ingestion/src/workers/historical.rs
+++ b/crates/iota-data-ingestion/src/workers/historical.rs
@@ -12,6 +12,9 @@ use iota_data_ingestion_core::{
     Reducer,
     history::{
         CHECKPOINT_FILE_MAGIC, MAGIC_BYTES,
+        epoch_boundaries::{
+            EpochBoundaries, read_epoch_boundaries_or_default, write_epoch_boundaries,
+        },
         manifest::{
             Manifest, create_file_metadata_from_bytes, finalize_manifest, read_manifest_from_bytes,
         },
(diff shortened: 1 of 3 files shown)


crates/iota-data-ingestion/src/workers/historical.rs
```diff
@@ -96,6 +111,34 @@ impl HistoricalReducer {
             Err(err) => Err(err)?,
         })
     }
+
+    /// Initializes the epoch boundaries from the provided seed if not already
+    /// initialized.
+    async fn seed_epoch_boundaries(&self, epoch_boundaries: EpochBoundaries) -> anyhow::Result<()> {
+        if !exists(&self.remote_store, &EpochBoundaries::file_path()).await {
```
sergiupopescu199: Nit: This function effectively downloads the file, If possible we could leverage the [head](https://docs.rs/object_store/latest/object_store/trait.ObjectStoreExt.html#tymethod.head) method to check if a file exists without downloading its contents.

sergiupopescu199 (approved):

crates/iota-data-ingestion/src/workers/historical.rs
kodemartin: I would say that this is not critical for the current patch, as interacting with `EPOCH_BOUNDARIES` is a rare operation.

Created an issue to refactor the upstream method, though: #11726

tomxey: this here is a bit risky, the "exists" may say "false" in case there are any issues with reading the file, which doesn't have to mean that the file doesn't exist (e.g. connection issue)
in such case we will reupload the seed file and overwrite the remote one

kodemartin: good catch! fixed with 1b19f06e87

we lose the retries, but that's ok. This is a startup operation.

tomxey: don't this change break the "record_epoch_boundary" function? it relies on the ability to overwrite the file with newer contents

kodemartin: true, my bad! fixed with https://github.com/iotaledger/iota/pull/11722/commits/ff8d83e9052cde1d1a82a307c66856a888bc5ab2

tomxey (approved):

crates/iota-data-ingestion/src/workers/historical.rs
```diff
@@ -96,6 +116,53 @@ impl HistoricalReducer {
...
+    /// Fails if there is a gap between the epochs already recorded and the
+    /// current one.
+    async fn record_epoch_boundary(
+        &self,
+        epoch_id: EpochId,
+        checkpoint_sequence_number: CheckpointSequenceNumber,
+    ) -> anyhow::Result<()> {
+        let mut epoch_boundaries =
+            read_epoch_boundaries_or_default(self.remote_store.clone()).await?;
+        epoch_boundaries.insert_next(epoch_id, checkpoint_sequence_number)?;
```
tomxey: if there is no high level retry mechanism for the whole function we may also consider to wrap those 3 lines inside of the retry below, in case of connection issues "read_epoch_boundaries_or_default" will return empty boundaries map which in turn will result in failure of "epoch_boundaries.insert_next"

kodemartin: there is already a retry mechanism inside `read_epoch_boundaries_or_default` after the existence check.

I removed the existence check and delegated defaulting to the callers, so that the renamed method `read_epoch_boundaries` does not return the default on errors other than `NotFound`.

We still interpret the retry timeout as an indication of `NotFound` and this will halt the writer, but this is something we have to live by. If there is a file and there are network issues that cause the timeout, restarting the service is something plausible.

I sought an alternative, but we currently don't have access to the direct `get` method to S3, so we cannot check for `Error::NotFound` definitively.

tomxey (approved):

PR #11728, opened 2026-06-03
feat: default framework-level IOTA account

### LINKED ISSUE #11690: [AA] Create a default IOTA account at the framework level
It must have at least 2 constructors: 
- a simple one 
- one where the object UID is derived through the ClaimRegistry

valeriyr:
# Description of change

Default framework-level IOTA account.

fixes #11690


muXxer (approved):
Approving `crates/iota-swarm-config/tests/snapshots/snapshot_tests__populated_genesis_snapshot_matches-2.snap`

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
```diff
@@ -0,0 +1,367 @@
...
+/// Aborts if the transaction sender is not the account.
+/// Aborts if no public key is currently attached.
+public fun rotate_builtin_auth_public_key(
+    self: &mut SmartAccount,
+    public_key: PublicKey,
+    ctx: &TxContext,
+): PublicKey {
+    ensure_tx_sender_is_smart_account(self, ctx);
+
+    builtin_authenticator_functions::rotate_public_key(&mut self.id, public_key)
```
miker83z, line 321: We already have `attach` and `detach` above if someone wants to break the link with the authenticator function for some reason. 

With this rotate, what do you think about automatically rotate the auth function as well? It could simply call  `account::rotate_auth_function_ref_v1(self, resolve_builtin_authenticator(public_key.scheme()));`

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
```diff
@@ -0,0 +1,367 @@
...
+///
+/// Aborts if the transaction sender is not the account.
+/// Aborts if a public key is already attached.
+public fun attach_builtin_auth_public_key(
+    self: &mut SmartAccount,
+    public_key: PublicKey,
+    ctx: &TxContext,
+) {
+    ensure_tx_sender_is_smart_account(self, ctx);
+
```
miker83z, line 247: Here, in `detach_builtin_auth_public_key` and `rotate_builtin_auth_public_key`, it throws no events. Maybe, instead of adding them here, we could add some in the appropriate functions in `builtin_authenticator_functions`.

valeriyr: Maybe, it is not clear which events we need. I haven't added them at all at the moment.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
valeriyr: It is possible, but inconsistent with the logic.
Currently, if you want to rotate the authenticator, you need to deal with all the required dynamic fields and then rotate; the same flow applies to any authenticator. The only place where it is automated is the builder constructors.
We can consider this option, but it should be consistent with the attach function, for example.
Maybe we can just remove rotating this field.

miker83z: Currently, `rotate_builtin_auth_public_key` is equivalent to `detach_builtin_auth_public_key`+`attach_builtin_auth_public_key`, so, if someone really wants to do this operation, it is allowed. 

Since we "own" the `builtin_authenticator_functions` I suggest we can decide to make the `rotate_builtin_auth_public_key` operation safer and automatically infer the authenticator. Could be renamed `rotate_builtin_authenticator_and_public_key`

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
miker83z: The authenticator rotation is covered by this: https://github.com/iotaledger/iota/blob/e880f16474f4c42aea1a0ba4de004455e8ba16b6/crates/iota-framework/packages/iota-framework/sources/account_abstraction/account.move#L32

We could have one `PublicKeyAttached` in `builtin_authenticator_functions::attach_public_key` and detached.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
valeriyr: This function will be misleading and inconsistent because you can have a public key attached but use a completely different authenticator. If you want to rotate a key, it does not always mean you want to rotate the authenticator as well. Dynamic fields management should be separated from the authenticator.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
valeriyr: I propose clarifying which events we need, creating an issue, and adding them in a dedicated PR.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
```diff
@@ -0,0 +1,367 @@
...
+    } else if (signature_scheme == signature_scheme::secp256r1()) {
+        builtin_authenticator_functions::secp256r1_authenticator_function_ref_v1<SmartAccount>()
+    } else if (signature_scheme == signature_scheme::multisig()) {
+        builtin_authenticator_functions::multisig_authenticator_function_ref_v1<SmartAccount>()
+    } else if (signature_scheme == signature_scheme::passkey()) {
+        builtin_authenticator_functions::passkey_authenticator_function_ref_v1<SmartAccount>()
+    } else {
+        abort EUnsupportedSignatureScheme
+    }
+}
```
valeriyr: I propose moving this to `builtin_authenticator_functions`, since any `AuthenticatorFunctionRefV1` can be attached to `SmartAccount`, so we don't really control the supported signature schemes here.

valeriyr: moved in ea76fc4a15

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
Dkwcs: Let’s imagine we create an account with Ed25519 public key and Ed25519 auth(because of public key schema).
So we could call - `rotate_builtin_auth_public_key(account, secp256k1_public_key, ctx)` then we have:
Stored public key - Secp256k1
Active authenticator - Ed25519 
Will it cause any inconsistency during authentication?

Or even worse if we just call `detach_builtin_auth_public_key(account, ctx)`, but active authenticator still waits public key field to use, what will happened in this case?(I’m not sure that builtin_authenticators use dynamic fields, just to check)

Dkwcs: Just checked that 
```
pub fn load_builtin_public_key<F>(
    account_object_id: ObjectID,
    get_object: F,
) -> IotaResult<(
    ObjectID,
    Option<(MovePublicKey, DynamicallyLoadedObjectMetadata)>,
)>
where
    F: FnOnce(ObjectID) -> IotaResult<Option<Object>>,
{
    let public_key_field_id = dynamic_field::derive_dynamic_field_id(
        account_object_id,
        &PublicKeyFieldName::tag().into(),
        &PublicKeyFieldName::default().to_bcs_bytes(),
    )
    .map_err(|_| UserInputError::UnableToGetAccountPublicKeyId { account_object_id })?;

    match get_object(public_key_field_id)? {
        Some(object) => {
            let metadata = DynamicallyLoadedObjectMetadata::from(&object);
            let field: Field<PublicKeyFieldName, MovePublicKey> = object
                .data
                .try_as_move()
                .expect("dynamic field should never be a package object”)
                .to_rust()
                .ok_or(UserInputError::InvalidAccountPublicKeyField { account_object_id })?;
            Ok((public_key_field_id, Some((field.value, metadata))))
        }
        None => Ok((public_key_field_id, None)),
    }
}
```
In this case we can self-lock the account, I guess

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
```diff
@@ -0,0 +1,346 @@
...
+    registry: &mut ClaimRegistry,
+    public_key: PublicKey,
+    ctx: &mut TxContext,
+): SmartAccountBuilder {
+    let mut account = SmartAccount { id: registry.claim(public_key, ctx) };
+    builtin_authenticator_functions::attach_public_key(&mut account.id, public_key);
+
+    SmartAccountBuilder {
+        account,
+        authenticator: builtin_authenticator_functions::from_signature_scheme<
```
Dkwcs, line 131: Should function `attach_public_key` have at least a basic validation inside? I see it takes domain type `PublicKey` and it should be already validated, however we still validate scheme later -builtin_authenticator_functions::from_signature_scheme<SmartAccount>(public_key.scheme()), 
I assume that semantically it would be better to validate scheme/public_key before attaching it, taking into account it can be aborted later

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
valeriyr: Yes, of course, you can.
Using the admin functions means that you understand what you are doing; it is not for an average user.
Achieving dynamic fields and the authenticator consistency is a developer's responsibility; you can't protect a developer from an issue here because they can always rotate the authenticator to an inconsistent one.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
valeriyr: We can consider adding this check, but there is currently no way to create an unsupported public key.
`from_signature_scheme` doesn't really validate the scheme; the error exists only because there is no `match` expression alternative in Move, and you have to return something in the else branch, but you can't.

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
```diff
@@ -0,0 +1,346 @@
...
+}
+
+/// Returns `true` if and only if `self` has a dynamic field with the specified `name`.
+public fun has_field<Name: copy + drop + store>(self: &SmartAccount, name: Name): bool {
+    dynamic_field::exists_(&self.id, name)
+}
+
+/// Returns `true` if and only if `self` has a built-in authenticator public key attached.
+public fun has_builtin_auth_public_key(self: &SmartAccount): bool {
+    builtin_authenticator_functions::has_public_key(&self.id)
```
Dkwcs, line 189: Just to consider - does this function really confirm that public key related to the builtin auth? It confirms that account has attached public key and it can be any

valeriyr: Not really, it checks that there is a dynamic field attached with the `builtin_authenticator_functions:: PublicKeyFieldName` key.
This key is used only by the built-in authenticators.

Dkwcs (approved):

crates/iota-framework/packages/iota-framework/sources/account_abstraction/smart_account.move
miker83z: https://github.com/iotaledger/iota/issues/11844

miker83z (approved):

sergiupopescu199 (approved):

PR #11734, opened 2026-06-03
upstream(iota-core,iota-protocol-config): persist and recover random beacon DKG failure across restarts

piotrm50:
## Description of change

- Port of upstream PR: [MystenLabs/sui#26856](https://github.com/MystenLabs/sui/pull/26856)
- Description:

At the start of each epoch, validators run a DKG protocol that bootstraps the random beacon. If DKG cannot reach its confirmation threshold before the timeout round it terminates as *failed*, which disables randomness for the epoch so randomness-using transactions are cancelled and the epoch can still close.

Previously this terminal-failure verdict lived only in memory:

- the on-disk `dkg_output` table only stored *successful* DKG outputs, so a restart past the timeout round reloaded an empty cell and reported DKG as `Pending` again; and
- `advance_dkg` only ran on commits carrying new DKG traffic, so after a restart with no fresh inbound messages the state machine never ticked again and the timeout was never re-evaluated.

Together these meant a validator could permanently forget that DKG had failed: randomness-using transactions were deferred forever (never cancelled), and the epoch could not close.

This makes the DKG outcome restart-proof:

1. **Persist the verdict.** A new `dkg_output_v2` table stores `Option<Output>`, so a terminal failure is recorded as `Some(None)`. `RandomnessManager::try_new` reads it first and falls back to the legacy `dkg_output` table for backward compatibility.
2. **Keep advancing while pending.** `advance_dkg` now runs on every commit while DKG is pending, regardless of new traffic, so a validator resolves DKG from already-persisted state — completing, or failing once the timeout round passes — even with no fresh inbound messages.

**Adapted to the fork:**

- The "keep advancing while pending" behavior is consensus-observable (it changes which commit DKG resolves on), so it is gated behind a new `always_advance_dkg_to_resolution` protocol feature flag, enabled in **protocol version 28**, so every validator switches at the same upgrade boundary.
- The advance is driven from `authority_per_epoch_store::process_consensus_transactions`, as this fork has not adopted the upstream commit-handler rewrite that moved DKG driving into `consensus_handler`.
- The upstream mainnet epoch fallback is omitted (not applicable to this fork).
- Tests follow the fork's existing standalone `test_dkg_*` style rather than the upstream `DkgTestSetup` harness.

Part of #11322

Ran locally:
- `cargo +nightly fmt`, `dprint fmt`, `cargo clippy -p iota-protocol-config -p iota-core --all-targets --all-features -- -D warnings`
- `cargo nextest run -p iota-core --lib epoch::randomness::tests::` — `test_dkg_v1`, `test_dkg_expiration_v1` (extended to cover restart recovery of the failure verdict), and `test_dkg_recovers_processed_messages_after_restart`
- `cargo nextest run -p iota-protocol-config` — snapshot tests (version 28 regenerated)
- `cargo llvm-cov` over the DKG tests confirms the persist + recover paths are exercised

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -829,6 +829,12 @@ pub struct AuthorityEpochTables {
     /// Records the final output of DKG after completion, including the public
     /// VSS key and any local private shares.
     pub(crate) dkg_output: DBMap<u64, dkg_v1::Output<PkG, EncG>>,
+    /// Successor of `dkg_output`. Wraps the output in an `Option` so the
+    /// terminal-failure verdict is persistable: `Some(Some(out))` = success,
+    /// `Some(None)` = DKG failed (terminal), table-absent = still pending.
+    /// `RandomnessManager::try_new` loads this table first and falls back to
+    /// `dkg_output` for backward compatibility with existing epoch stores.
+    pub(crate) dkg_output_v2: DBMap<u64, Option<dkg_v1::Output<PkG, EncG>>>,
 
     /// Holds the value of the next RandomnessRound to be generated.
     pub(crate) randomness_next_round: DBMap<u64, RandomnessRound>,
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -3888,7 +3888,16 @@ impl AuthorityPerEpochStore {
             );
         }
 
-        if randomness_state_updated {
+        // Advance the DKG state machine while it is still pending, even when no
+        // new messages or confirmations arrived this commit. Without this, a
+        // validator that restarts past the timeout round will never tick the
+        // state machine again (no new messages arrive after restart), so DKG
+        // can stay Pending forever -- deferring all randomness-using
+        // transactions and blocking epoch close.
+        let dkg_pending = randomness_manager
+            .as_ref()
+            .is_some_and(|rm| rm.dkg_status() == DkgStatus::Pending);
+        if randomness_state_updated || dkg_pending {
             if let Some(randomness_manager) = randomness_manager.as_mut() {
                 randomness_manager
                     .advance_dkg(output, consensus_commit_info.round)
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -3888,15 +3888,23 @@ impl AuthorityPerEpochStore {
             );
         }
 
-        // Advance the DKG state machine while it is still pending, even when no
-        // new messages or confirmations arrived this commit. Without this, a
-        // validator that restarts past the timeout round will never tick the
-        // state machine again (no new messages arrive after restart), so DKG
-        // can stay Pending forever -- deferring all randomness-using
-        // transactions and blocking epoch close.
-        let dkg_pending = randomness_manager
-            .as_ref()
-            .is_some_and(|rm| rm.dkg_status() == DkgStatus::Pending);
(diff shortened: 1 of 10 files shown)


polinikita (approved):

kodemartin (approved):

roman1e2f5p8s (approved):

bingyanglin (approved):

crates/iota-core/src/authority/consensus_quarantine.rs
```diff
@@ -71,7 +71,9 @@ pub(crate) struct ConsensusCommitOutput {
     dkg_confirmations: BTreeMap<PartyId, VersionedDkgConfirmation>,
     dkg_processed_messages: BTreeMap<PartyId, VersionedProcessedMessage>,
     dkg_used_message: Option<VersionedUsedProcessedMessages>,
-    dkg_output: Option<dkg_v1::Output<PkG, EncG>>,
+    // Outer Option = "set this commit"; inner Option = success (`Some(out)`) /
+    // terminal failure (`None`). Persisted to `dkg_output_v2`.
```
semenov-vladyslav:
```suggestion
    // DKG state, the inner Option value persisted to `dkg_output_v2`:
    // - `Some(Some(out))` -- DKG success
    // - `Some(None)` -- terminal DKG failure
    // - `None` -- DKG pending (default value)
```
The comment is hard to understand. I would clarify it as suggested.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -828,6 +828,12 @@ pub struct AuthorityEpochTables {
     /// Records the final output of DKG after completion, including the public
     /// VSS key and any local private shares.
     pub(crate) dkg_output: DBMap<u64, dkg_v1::Output<PkG, EncG>>,
+    /// Successor of `dkg_output`. Wraps the output in an `Option` so the
+    /// terminal-failure verdict is persistable: `Some(Some(out))` = success,
+    /// `Some(None)` = DKG failed (terminal), table-absent = still pending.
```
semenov-vladyslav:
```suggestion
    /// terminal-failure verdict is persistable: `Some(out)` = success,
    /// `None` = DKG failed (terminal), table-absent = still pending.
```
There's only one `Option` layer. The second one is from the `get` operation, which makes the comment confusing.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -828,6 +828,12 @@ pub struct AuthorityEpochTables {
     /// Records the final output of DKG after completion, including the public
     /// VSS key and any local private shares.
     pub(crate) dkg_output: DBMap<u64, dkg_v1::Output<PkG, EncG>>,
```
semenov-vladyslav, line 834: Should it be marked deprecated somehow?

semenov-vladyslav (approved):
Looks good. I would improve comments a bit.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: applied

crates/iota-core/src/authority/consensus_quarantine.rs
piotrm50: applied.

crates/iota-core/src/authority/authority_per_epoch_store.rs
piotrm50: Other deprecated tables such as deferred_transactions are not marked in any way. We will need to make a pass in another PR and clean up those old tables. Also, this one is still read as a fallback. Once we get rid of this fallback and it's not used anywhere we can mark this as deprecated and eventually clean it up.

PR #11757, opened 2026-06-04
feat(data-ingestion-core): add epoch boundary verifier

### LINKED ISSUE #11255: snapshots: extract the verification logic
The snapshot state verification logic (e.g. verification of the `ECMHLiveObjectSetDigest` against the last certified checkpoint) should be extracted so that the indexer can reuse it during restoring from formal snapshots. See `expensive_check_is_consistent_state`

kodemartin:
# Description of change

This patch introduces the logic and necessary types to verify the last checkpoint of an epoch.

Verification requires knowing the active committee of the epoch, which is published in the last checkpoint of the previous epoch. So in order for the active committee to be verified, we need to verify the checkpoints at the boundary of all previous epoch up to Genesis.

We introduce a new type `EpochBoundaryVerifier` that traverses and verifies all such checkpoints. It exposes a way to stream all the verified checkpoints up to the target epoch, and a method that folds this chain of verifications to return the verified checkpoint of the target epoch.

This has been motivated by the need to support verification of formal snapshots using the historical checkpoint store.

Fixes #11255  

### DIFF
--- a/crates/iota-data-ingestion-core/src/history/verifier.rs
+++ b/crates/iota-data-ingestion-core/src/history/verifier.rs
@@ -4,7 +4,7 @@
 //! Verify the last checkpoint of an epoch against the committee of that epoch.
 //!
 //! The committee of an epoch is recorded only in the last checkpoint of the
-//! previous epoch, as [`EndOfEpochData::next_epoch_committee`]. Verifying the
+//! previous epoch, as `EndOfEpochData::next_epoch_committee`. Verifying the
 //! last checkpoint of epoch `N` therefore requires the committee taken from the
 //! last checkpoint of epoch `N - 1`, which must itself be verified the same
 //! way going all the way back to the genesis committee.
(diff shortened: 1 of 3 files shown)


sergiupopescu199 (approved):
lgtm 🚀

crates/iota-data-ingestion-core/src/history/verifier.rs
```diff
@@ -0,0 +1,256 @@
...
+                .next_epoch_committee
+                .iter()
+                .cloned()
+                .collect(),
+        );
+
+        Ok(Some((verified, next_epoch_committee)))
+    }
+
+    /// Fetches a single checkpoint summary into memory.
```
tomxey, line 235: nit: as I understand, to fetch the single checkpoint summary it needs to download the whole blob with full checkpoint data of some group of checkpoints

not sure how big such blob is, but maybe we could note in the comment that this operation is heavier than it looks?

tomxey (approved):

crates/iota-data-ingestion-core/src/history/verifier.rs
kodemartin: good point. Updated docs with https://github.com/iotaledger/iota/pull/11757/commits/def6e62e7cd2cfe8ff5dd21687e451e86cc543e7

PR #11762, opened 2026-06-05
chore: use external iota-sdk crate over individual crates

DaughterOfMars:
## Description

Simplify the external SDK dependencies to use only the single unifying `iota-sdk` crate (as `iota-sdk-ext`), rather than all of the individual crates which must be kept in sync with each other.

- Replaces `iota-sdk-types`, `iota-sdk-crypto`, `iota-sdk-transaction-builder`, `iota-sdk-graphql-client`, `iota-grpc-types` and `iota-grpc-client` with `iota-sdk-ext`, mapping each to the meta-crate's feature names (`types`, `crypto`, `grpc`, `txn-builder`, `graphql`).
- `crates/iota-rust-sdk` now forwards to `iota-sdk-ext` instead of aggregating the individual crates.
- Pins `iota-sdk-ext` at `0d9653f1` — the revision `develop` already uses — and points the two standalone example workspaces at the same revision, so the whole tree resolves to one SDK.

Drive-by fixes: neither example workspace built on `develop`. `docs/examples/rust` failed against its stale `164d6626` pin, and `examples/tic-tac-toe/cli` still imported multisig types that have since moved out of `iota-types`. Both are updated to the current SDK API.


bingyanglin (commented):
Do we need to also change `scripts/update_sdk_types_rev.sh` which still updates the four old dep names?

DaughterOfMars:
> Do we need to also change `scripts/update_sdk_types_rev.sh` which still updates the four old dep names?

I guess, although I think the script is much less needed after this change

alexsporn:
Could/Should we leverage the new `iota-rust-sdk` (https://github.com/iotaledger/iota/pull/11944) crate for this instead of adding a `iota-sdk-ext`?
I think it makes total sense to centralize the access to the external SDK so that we only have the dependency in the `crates/iota-rust-sdk/Cargo.toml` and not distributed over all the repo. Then we also don't need the script to update the revs.

DaughterOfMars:
> Could/Should we leverage the new `iota-rust-sdk` (#11944) crate for this instead of adding a `iota-sdk-ext`? I think it makes total sense to centralize the access to the external SDK so that we only have the dependency in the `crates/iota-rust-sdk/Cargo.toml` and not distributed over all the repo. Then we also don't need the script to update the revs.

I actually think that crate should not exist, as the `iota-sdk` crate in the external repo serves this exact function.

thibault-martinez (commented):
I am not against the idea of this PR but IMO we should not be doing it before we are done with the SDK replacement. We will have to change the `iota_sdk_ext` name again in a few weeks so this is just noise for now.

DaughterOfMars:
> I am not against the idea of this PR but IMO we should not be doing it before we are done with the SDK replacement. We will have to change the `iota_sdk_ext` name again in a few weeks so this is just noise for now.

It's not just noise. This dramatically simplifies the imports. It doesn't matter what we name the package, we can change it easily at any time.

Alex6323 (approved):

Thoralf-M (approved):

filipdulic (approved):

filipdulic:
approved for consensus

muXxer (approved):

thibault-martinez:
Drafting until https://github.com/iotaledger/iota/pull/12762 is merged

PR #11771, opened 2026-06-05
[dev-tools] Move Multisig tests to SDK

divol89:
Moves the multisig unit/integration tests from crates/iota-types/src/unit_tests/multisig_tests.rs to crates/iota-sdk/tests/multisig_tests.rs in accordance with the repo's type-location guidelines.

PR is generated automatically.

Static review only. No local tests or builds were run due to resource-safety policy.


thibault-martinez:
Thanks for your interest but we meant the external SDK, not the internal one. Plus this issue/TODO is not open for external contributors.

PR #11786, opened 2026-06-05
chore(iota-types): move `new_empty_v1` constructor

### LINKED ISSUE #11596: `TransactionEffects::new_empty_v1` should not be part of the Ext trait
We have a Test trait, should probably be there.
Rename the trait if the "API" part is confusing.

Alex6323:
# Description of change

Moves this ctor as discussed in a comment on #10336 , but was deemed to be deferred ... UNTIL NOW!

Fixes #11596 


crates/iota-types/src/effects/mod.rs
```diff
@@ -358,6 +353,16 @@ pub trait TransactionEffectsExt: transaction_effects_ext::Sealed {
     }
 }
 
+/// Test-only counterpart to [`TransactionEffectsExt`]: enum-level extensions
+/// (version-selecting constructors and aggregating helpers). Sealed;
+/// implemented for the enum only.
+pub trait TransactionEffectsExtForTesting: transaction_effects_ext::Sealed {
```
roman1e2f5p8s, line 357: should we add the same note here as we have for `TransactionEffectsAPIForTesting`?

```suggestion
pub trait TransactionEffectsExtForTesting: transaction_effects_ext::Sealed {
    // All of these should be #[cfg(test)], but they are used by tests in other
    // crates, and dependencies don't get built with cfg(test) set as far as I
    // can tell.
```

roman1e2f5p8s (approved):

bingyanglin (approved):

crates/iota-core/src/generate_format.rs
```diff
@@ -579,7 +579,19 @@ fn get_registry() -> Result<Registry> {
     // Trace FullCheckpointContents, CheckpointTransaction and CheckpointData
     // via trace_value (they transitively contain TypeTag).
     let sample_transaction = Transaction::new(sender_data.clone());
-    let sample_effects = TransactionEffects::new_empty_v1(TransactionDigest::default());
```
DaughterOfMars, line 582: Why was this changed?

Alex6323: because that's production code, not for testing, which the ctor is created for solely. I know it may seem weird to not use the testing ctor here because it currently creates the same output, but I wanted to have a clear separation of concerns.

Alex6323: Just imagine we wanted to change something in the tests, then this would leak here and probably be easily missed.

crates/iota-types/src/effects/mod.rs
Alex6323: I think we can add a comment here as well, but I'll change it a bit because I don't like the "as far as I can tell" that much :slightly_smiling_face:

Alex6323: I gave it a closer look, applied your suggestion and made my own improvements to the docs/comments in https://github.com/iotaledger/iota/pull/11786/commits/e1ed425ee1c517b6bd4923084ea7da433e395df9. Fine with it @roman1e2f5p8s ?

crates/iota-core/src/generate_format.rs
DaughterOfMars: That's fair, I suppose!

DaughterOfMars (approved):

piotrm50 (approved):

Thoralf-M (approved):

crates/iota-core/src/generate_format.rs
thibault-martinez: I'm re-opening this to say that this is actually for testing, not production code

thibault-martinez: Not necessarily asking you to revert but if the argument for replacing was production code, it's not, it's only regression testing

muXxer (approved):

kodemartin (approved):

PR #11790, opened 2026-06-05
feat(starfish): allow burst block proposals within a rate window instead of fixed min delay

### LINKED ISSUE #11789: feat(starfish): mitigate block-rate disparity of slow validators caused by the fixed min block delay
Validators on high-latency or bursty links (heavy-tail validators on mainnet/testnet) produce visibly fewer blocks per second than the fast quorum. Their threshold-clock advances arrive bursty; the fixed `min_block_delay` (50ms) permits only one proposal per interval regardless of accumulated waiting, so every short inter-advance gap forces a round skip and the validator's blocks also land late relative to their round.

**Key points:**
- Block production should be limited in sustained rate, not in instantaneous spacing: idle time waiting on the network should accrue budget for burst catch-up.
- Round skips reduce strong-link vote opportunities for committed leaders, lowering reputation scores and costing leader slots — a compounding penalty.
- The limit must remain a local, per-authority concern (no protocol change): no receive-side validation depends on inter-block spacing, and Byzantine validators were never bound by it.
- Rate-limited proposal declines are currently silent (no metric).

polinikita:
# Description of change

Replaces the fixed `min_block_delay` gate between consecutive own blocks with a GCRA (token-bucket) rate limiter: sustained production stays at one block per `min_block_delay`, while time spent waiting on the network accrues budget for bursts of up to `block_rate_window / min_block_delay` back-to-back blocks (40 in production, 5 in msim, 8 in tests). A validator that fell behind — e.g. bursty block arrival on a high-latency link — can then propose on every threshold-clock advance instead of skipping rounds, mitigating the block-rate spread between fast and slow validators observed on mainnet/testnet.

Design notes:
- GCRA state is a single theoretical-arrival timestamp; burst capacity 1 degenerates to the previous fixed-delay rule, and under saturation conforming instants are spaced exactly `min_block_delay` apart (no burst/dead-zone oscillation).
- Forced proposals (leader timeout etc.) still bypass the check but consume budget, with an overdraft cap reproducing today's "forced block delays the next non-forced one by one interval".
- The leader timeout task is unchanged: with the cap, the retry timer armed at `last_own_block + min_block_delay` remains the latest instant at which a proposal can still be rate-blocked, so it fires exactly when conformance is guaranteed.
- The limiter is seeded on startup by replaying own cached block-header timestamps from DagState, so a quick restart does not grant a fresh burst budget; after a long downtime the bucket is full, which is the desired catch-up behavior.
- Purely local node configuration — no protocol change, safe in mixed fleets (no receive-side validation depends on inter-block spacing). `block-rate-window <= min-block-delay` is a config-only rollback to the previous behavior.
- Rate-limited declines, previously silent, are now counted via `core_skipped_proposals{reason="block_rate_limited"}`.
- The block broadcast channel keeps headroom for a full burst (msim capacity equaled the burst size).

Tested on a local network that has the same artefacts as Mainnet/Testnet, where slow validators receive blocks in bursts from relay nodes and produce a noticeably smaller number of blocks. With this change, there is no difference between the block rate of slow and fast validators. As a side effect, it could improve the e2e latency for transactions handled by slow validators

Fixes #11789

### DIFF
--- a/crates/starfish/core/src/block_rate_limiter.rs
+++ b/crates/starfish/core/src/block_rate_limiter.rs
@@ -1,0 +1,163 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::time::Duration;
+
+/// Rate limiter for own block proposals (GCRA / token-bucket): sustained
+/// production of at most one block per `min_block_delay`, with bursts of up to
+/// `block_rate_window / min_block_delay` back-to-back blocks after idle
+/// periods. A burst capacity of 1 degenerates to a fixed minimum delay
+/// between consecutive blocks.
+///
+/// State is a single instant: the time at which an ideal schedule emitting
+/// exactly one block per interval would emit the next block. Idle time lets it
+/// fall back toward `now` (accruing budget); each proposal advances it by one
+/// interval.
+pub(crate) struct BlockRateLimiter {
+    /// UTC ms at which the next block would be emitted under an ideal schedule
+    /// of one block per `interval_ms` (the GCRA theoretical arrival time). It
+    /// runs ahead of `now` after recent proposals and falls back toward `now`
+    /// while idle; the gap below `now` is the unspent burst budget.
+    next_block_ms: u64,
+    /// Sustained spacing between blocks (`min_block_delay`), in ms.
+    interval_ms: u64,
+    /// Maximum number of back-to-back blocks allowed after an idle period.
+    burst: u64,
+}
+
+impl BlockRateLimiter {
+    pub(crate) fn new(min_block_delay: Duration, burst: u64) -> Self {
+        Self {
+            next_block_ms: 0,
+            interval_ms: min_block_delay.as_millis().max(1) as u64,
+            burst: burst.max(1),
+        }
+    }
+
+    /// Whether a proposal at `now_ms` fits the rate envelope.
+    pub(crate) fn is_conforming(&self, now_ms: u64) -> bool {
+        self.next_block_ms.saturating_sub(now_ms) <= (self.burst - 1) * self.interval_ms
+    }
+
+    /// Records a proposal at `now_ms`. Called for every own block, including
+    /// forced proposals that bypass the conformance check; the cap bounds
+    /// their overdraft so the next non-forced proposal waits at most one
+    /// interval after the last block.
+    pub(crate) fn record(&mut self, now_ms: u64) {
+        self.next_block_ms = (self.next_block_ms.max(now_ms) + self.interval_ms)
+            .min(now_ms + self.burst * self.interval_ms);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use starfish_config::Parameters;
+
+    use crate::block_rate_limiter::BlockRateLimiter;
+
+    /// Drive the tests from the default consensus parameters rather than magic
+    /// numbers, so they track the production config (interval and burst).
+    fn interval_ms_and_burst() -> (u64, u64) {
+        let params = Parameters::default();
+        (
+            params.min_block_delay.as_millis() as u64,
+            params.block_rate_burst(),
+        )
+    }
+
+    fn limiter() -> BlockRateLimiter {
+        let params = Parameters::default();
+        BlockRateLimiter::new(params.min_block_delay, params.block_rate_burst())
+    }
+
+    #[test]
+    fn burst_after_idle_then_reject() {
+        let (interval_ms, burst) = interval_ms_and_burst();
+        let mut l = limiter();
+        let now = 1_000_000;
+        // Exactly `burst` back-to-back proposals conform, the next does not.
+        for _ in 0..burst {
+            assert!(l.is_conforming(now));
+            l.record(now);
+        }
+        assert!(!l.is_conforming(now));
+        // Budget for one more block regenerates after one interval.
+        assert!(!l.is_conforming(now + interval_ms - 1));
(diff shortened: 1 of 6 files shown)


crates/starfish/core/src/block_rate_limiter.rs
```diff
@@ -0,0 +1,140 @@
...
+            .min(now_ms + self.burst * self.interval_ms);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const T: u64 = 50;
+    const B: u64 = 40;
```
VorobyevIlya: could you make more descriptive names?

crates/starfish/core/src/core.rs
```diff
@@ -235,6 +244,20 @@ impl Core {
 
         let encoder = create_encoder(&context);
 
+        // Seed the rate limiter from own recent blocks so a quick restart does
+        // not grant a fresh burst budget. Blocks older than the window cannot
+        // affect it, so a bounded round lookback suffices.
+        let burst = context.parameters.block_rate_burst();
+        let mut proposal_rate_limiter =
+            BlockRateLimiter::new(context.parameters.min_block_delay, burst);
+        let lookback_start = last_signaled_round.saturating_sub(2 * burst as Round);
```
VorobyevIlya: why is it 2*burst and not just burst?

crates/starfish/core/src/core.rs
```diff
@@ -1451,10 +1479,12 @@ impl CoreSignals {
...
         // Blocks buffered in broadcast channel should be roughly equal to thosed cached
         // in dag state, since the underlying blocks are ref counted so a lower
-        // buffer here will not reduce memory usage significantly.
-        let (tx_block_broadcast, rx_block_broadcast) = broadcast::channel::<VerifiedBlock>(
-            context.parameters.dag_state_cached_rounds as usize,
-        );
+        // buffer here will not reduce memory usage significantly. Keep headroom
+        // for a full proposal burst so lagging subscribers don't drop blocks.
+        let capacity = (context.parameters.dag_state_cached_rounds as usize)
+            .max(2 * context.parameters.block_rate_burst() as usize);
```
VorobyevIlya: and here, why do we multiply by 2?

crates/starfish/core/src/block_rate_limiter.rs
```diff
@@ -0,0 +1,140 @@
...
+/// periods. A burst capacity of 1 degenerates to a fixed minimum delay
+/// between consecutive blocks.
+///
+/// State is a single theoretical-arrival time: the earliest instant at which
+/// an ideal schedule emitting exactly one block per interval would emit next.
+/// Idle time lowers it relative to `now` (accruing budget); each proposal
+/// advances it by one interval.
+pub(crate) struct BlockRateLimiter {
+    /// Theoretical arrival time, UTC ms.
+    tat_ms: u64,
```
VorobyevIlya: so as far as I understand, this is the moment when the next block should be proposed in idealized conditions. Why is it called like this then?

VorobyevIlya (commented):
Need to rebase after the merge of Starfish Speed

crates/starfish/core/src/block_rate_limiter.rs
polinikita: Renamed `tat_ms` → `next_block_ms` and expanded the doc: it is the UTC ms at which an ideal one-block-per-`interval_ms` schedule would emit the next block (the GCRA theoretical arrival time). It runs ahead of `now` after recent proposals and falls back toward `now` while idle, and the gap below `now` is the unspent burst budget.

crates/starfish/core/src/block_rate_limiter.rs
polinikita: Renamed the test constants `T` → `INTERVAL_MS` and `B` → `BURST` (and updated the comments that referenced them).

polinikita:
Rebased onto current \`develop\` (now includes the merged StarfishSpeed and the dead-config cleanup):
- reconciled the burst limiter with StarfishSpeed (kept `soft_leader_timeout` + `block_rate_window`; the extra-wait metric now keys off the limiter's `is_conforming` instead of the old `>= min_block_delay`);
- dropped `max_forward_time_drift` here too, since the dead-config PR removed it on `develop`.

Also addressed the inline comments: renamed `tat_ms` → `next_block_ms` and the test constants `T`/`B` → `INTERVAL_MS`/`BURST`.

crates/starfish/core/src/block_rate_limiter.rs
polinikita: Follow-up: rather than introduce named constants, the tests now read the interval and burst directly from `Parameters::default()` (via a small `interval_ms_and_burst()` helper), so they track the production config instead of hard-coded numbers. Also dropped the `use super::*` wildcard in favor of explicit imports.

crates/starfish/core/src/core.rs
polinikita: Reduced to a single `burst`-round lookback. That captures the recent own blocks that still affect the limiter, and replaying anything older is harmless (the limiter absorbs it), so the extra margin was not needed.

crates/starfish/core/src/core.rs
polinikita: Reduced to a single `burst`. This floor only matters when it exceeds `dag_state_cached_rounds` — i.e. in msim/tests; in production `dag_state_cached_rounds` (500) dominates the `max`, so one burst of capacity is enough.

VorobyevIlya (approved):

piotrm50 (approved):

PR #11800, opened 2026-06-08
feat: Add state sync available checkpoint content gap test

### LINKED ISSUE #11496: node: add a test for gaps in the checkpoint data during sync
Nodes paniced during syncing from genesis because there was an undocumented invariant in the code, which was violated. We should add a test, so that this doesn't happen in the first place.

Link to the fix:
https://github.com/iotaledger/iota/pull/11485

semenov-vladyslav:
# Description of change

This PR add a reproducer test for a fix in #11485. Before the fix a missing checkpoint would be pushed to the end of retry queue, causing state sync to try and fetch the following checkpoint thus breaking the invariant. The test simulates a gap in the available checkpoint contents and checks that the state sync does not break the invariant. Without the fix (with `checkpoint_contents_tasks.push_back`) the test fails.

Fixes #11496

### DIFF
--- a/crates/iota-network/src/state_sync/tests.rs
+++ b/crates/iota-network/src/state_sync/tests.rs
@@ -1031,6 +1031,13 @@ async fn sync_with_checkpoints_gap() -> anyhow::Result<()> {
     // this window.
     tokio::time::sleep(Duration::from_secs(2)).await;
 
+    assert!(
+        store_2
+            .get_full_checkpoint_contents_by_sequence_number(2)
+            .is_some(),
+        "content loop did not even fetch seq 2 within the window — test is not exercising the gap",
+    );
+
     // REGRESSION CHECK: the synced watermark must not have advanced past
     // genesis (sequence 0) while checkpoint 1's contents are unavailable.
     assert_eq!(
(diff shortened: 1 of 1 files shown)


crates/iota-network/src/state_sync/tests.rs
```diff
@@ -936,3 +937,143 @@ async fn sync_with_checkpoints_watermark() {
...
+        }
+    })
+    .await
+    .expect("node 2 failed to sync all checkpoint summaries");
+
+    // Give the content-sync loop 2 s to run.  With a 50 ms retry interval
+    // ~40 retry cycles fire for checkpoint 1.  If push_back were used
+    // (pre-fix), checkpoints 2–5 would advance the watermark to 5 within
+    // this window.
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
bingyanglin, line 1032: Not sure a fixed 2s here is good especially for a heavy/slow CI box, and it might be even safer to add an assertion here like

```rust
// e.g. before the watermark assert:
assert!(
    store_2.get_full_checkpoint_contents_by_sequence_number(2).is_some(),
    "content loop did not even fetch seq 2 within the window — test is not exercising the gap",
);
```
so we can prove the loop ran and still refused to advance.
How do you think?

muXxer (approved):

crates/iota-network/src/state_sync/tests.rs
semenov-vladyslav: Sounds reasonable. Added assert here https://github.com/iotaledger/iota/pull/11800/commits/451a62fd6be96b6663287e3b682042912d08b5e2

bingyanglin (approved):

PR #11807, opened 2026-06-08
refactor(iota-indexer): add fallback support for `TransactionFilter{V2}::FromOrToAddress` transaction query

### LINKED ISSUE #11678: [`iota-indexer`]: use historical fallback for  `FromOrToAddress` filter.
following https://github.com/iotaledger/iota/issues/9498

Refactor the Indexer `FromOrToAddress` filtered query to resolve to this new BigTable mapping for historical transactions

sergiupopescu199:
# Description of change

This PR adds historical fallback support to the indexer API for `TransactionFilter{V2}::FromOrToAddress`.

### When the fallback is triggered

The fallback is triggered in three explicit cases:

1. **Cursor resolution below watermark**: the DB query raises `DataPruned` after `ensure_data_not_pruned_for_tx` checks the cursor's sequence number against the watermark.

2. **ASC + no cursor while pruning is active**: the DB cannot distinguish between "no history" and "fully pruned history", so it conservatively raises `DataPruned`, triggering the fallback.

3. **DESC page spans the watermark**: a short DESC page (DB returned fewer rows than the limit) with active pruning means the page crossed the pruning boundary. The fallback fetches the missing pre-watermark suffix using the last DB row's digest as the cursor.

A fast path skips the DB entirely when the cursor is in the fallback's in-memory cursor cache and its sequence number is below the watermark. This avoids round-trips during deep pagination through pruned data.

fixes #11678 

- Synced the indexer and the iota-kvstore worker upon a local netowrk
- Issued `iotax_queryTransactionBlocks` with `FromOrToAddress` filter before pruning.
- Deleted records from `tx_senders` & `tx_recipients`.
- Updated `min_available_tx` in `watermarks` table to reflect pruning.
- Issued the same queries to check if historical fallback would be triggered.
- Pagination works as expected, only limitation is when no cursor is provided in ascending order. We hit the limitation described above.

> [!NOTE]

### DIFF
--- a/crates/iota-indexer/src/historical_fallback/client.rs
+++ b/crates/iota-indexer/src/historical_fallback/client.rs
@@ -383,7 +383,7 @@ impl HttpRestKVClient {
 
         let bytes = resp.bytes().await?;
         bcs::from_bytes::<Vec<T>>(&bytes).map_err(|e| {
-            IndexerError::Serde(format!("failed to deserialize multi_get response: {e:?}"))
+            IndexerError::Serde(format!("failed to deserialize paginated response: {e:?}"))
         })
     }
 
--- a/crates/iota-indexer/src/historical_fallback/client.rs
+++ b/crates/iota-indexer/src/historical_fallback/client.rs
@@ -3,7 +3,7 @@
 
 //! Module containing the client for interacting with the REST API KV server.
 
-use std::time::Duration;
+use std::{fmt::Display, num::NonZeroUsize, time::Duration};
 
 use bytes::Bytes;
 use futures::{
@@ -12,7 +12,7 @@ use futures::{
 };
 use iota_storage::http_key_value_store::{ItemType, Key};
 use iota_types::{
-    base_types::{ObjectID, SequenceNumber},
+    base_types::{IotaAddress, ObjectID, SequenceNumber},
     digests::{CheckpointDigest, TransactionDigest},
     effects::{TransactionEffects, TransactionEffectsAPI, TransactionEvents},
     messages_checkpoint::{
@@ -36,7 +36,11 @@ use crate::{
     historical_fallback::metrics::HistoricalFallbackClientMetrics,
 };
 
-const CACHE_TIME_TO_IDLE: Duration = Duration::from_secs(600);
+pub(crate) const CACHE_TIME_TO_IDLE: Duration = Duration::from_secs(600);
+
+/// Represents the sequence number of a transaction.
+pub type TransactionSequenceNumber = u64;
+
 /// Request payload for multi_get containing list of keys.
 #[derive(Serialize, Debug)]
 struct MultiGetRequest {
@@ -126,6 +130,41 @@ pub(crate) trait KeyValueStoreClient {
         object_refs: &[(ObjectID, SequenceNumber)],
         before_version: bool,
     ) -> IndexerResult<Vec<Option<Object>>>;
+}
+
+/// Paginated reads against the historical KV store.
+///
+/// Provides an interface for retrieving ordered subsets of values associated
+/// with a single primary key. Distinct from [`KeyValueStoreClient`], which
+/// is designed for point lookups.
+#[async_trait::async_trait]
+pub(crate) trait PaginatedKeyValueStoreClient {
+    /// Fetches a paginated list of transaction digests that affect a given
+    /// address.
+    ///
+    /// An address is considered "affected" by a transaction if it appears
+    /// as the sender, a recipient, or the gas payer.
+    ///
+    /// # Pagination
+    ///
+    /// * **Cursor:** The `cursor` is an *exclusive* boundary. Pass `None` to
+    ///   fetch the first page. For subsequent pages, provide the
+    ///   [`TransactionSequenceNumber`] from the last item of the previous
+    ///   result.
+    /// * **Limit:** The `limit` is the maximum number of items per page. The
+    ///   actual result may contain fewer items than requested.
+    /// * **Ordering:**
+    ///   - `oldest_first = false` (default): newest to oldest.
+    ///   - `oldest_first = true`: oldest to newest.
+    ///
+    /// The `cursor` semantics remain exclusive regardless of scan direction.
+    async fn transaction_digests_by_address(
+        &self,
+        address: IotaAddress,
+        cursor: Option<TransactionSequenceNumber>,
+        limit: usize,
+        oldest_first: bool,
+    ) -> IndexerResult<Vec<(TransactionSequenceNumber, TransactionDigest)>>;
 }
 
 #[derive(Clone)]
@@ -275,6 +314,75 @@ impl HttpRestKVClient {
 
         let bytes = resp.bytes().await?;
         bcs::from_bytes::<Vec<Option<Bytes>>>(&bytes).map_err(|e| {
+            IndexerError::Serde(format!("failed to deserialize multi_get response: {e:?}"))
+        })
+    }
+
+    /// Fetches a paginated list of items from a range-query endpoint.
+    ///
+    /// This method performs a one-to-many lookup, retrieving a paginated list
+    /// of records associated with the given `key`.
+    ///
+    /// # Pagination Logic
+    ///
+    /// * **Cursor-based:** The `cursor` is an *exclusive* boundary. When
+    ///   requesting the first page, pass `None`. For subsequent pages, use the
+    ///   cursor identifier from the last item of the previous result set.
+    /// * **Limits:** The `limit` enforces an upper bound on items returned.
+    /// * **Reversed:** The `reversed` flag determines the scan direction.
+    ///   - `false` (default): Follows the natural storage order of the `Key`.
+    ///   - `true`: Attempts to reverse the scan direction.
+    async fn paginate<C, T>(
+        &self,
+        key: Key,
+        cursor: Option<C>,
+        limit: impl TryInto<NonZeroUsize>,
+        reversed: bool,
+    ) -> IndexerResult<Vec<T>>
+    where
+        C: Display,
+        T: for<'de> Deserialize<'de>,
+    {
+        let limit = limit.try_into().map_err(|_| {
+            IndexerError::HistoricalFallbackInput("limit must be greater than 0".into())
+        })?;
+
+        let (item_type, encoded_key) = key.to_path_elements();
+        let mut url = self.base_url.join(&format!("{item_type}/{encoded_key}"))?;
+
+        url.query_pairs_mut()
+            .append_pair("limit", &limit.get().to_string());
+
+        if let Some(cursor) = cursor {
+            url.query_pairs_mut()
+                .append_pair("cursor", &cursor.to_string());
+        }
+
+        if reversed {
+            url.query_pairs_mut().append_pair("oldest_first", "true");
+        }
+
+        trace!("fetching from url: {url}");
+
+        let resp = self.client.get(url.clone()).send().await?;
+
+        trace!(
+            "got response {} for url: {url}, len: {:?}",
+            resp.status(),
+            resp.headers()
+                .get(CONTENT_LENGTH)
+                .unwrap_or(&HeaderValue::from_static("0"))
+        );
+
+        if !resp.status().is_success() {
+            return Err(IndexerError::HistoricalFallbackStorageError(format!(
+                "multi_fetch request failed with status: {}",
+                resp.status()
+            )));
+        }
+
+        let bytes = resp.bytes().await?;
+        bcs::from_bytes::<Vec<T>>(&bytes).map_err(|e| {
             IndexerError::Serde(format!("failed to deserialize multi_get response: {e:?}"))
         })
     }
@@ -614,5 +722,25 @@ impl KeyValueStoreClient for HttpRestKVClient {
             .collect();
 
         Ok(objects)
+    }
+}
+
+#[async_trait::async_trait]
+impl PaginatedKeyValueStoreClient for HttpRestKVClient {
+    #[instrument(level = "trace", skip_all)]
+    async fn transaction_digests_by_address(
+        &self,
+        address: IotaAddress,
+        cursor: Option<TransactionSequenceNumber>,
+        limit: usize,
+        oldest_first: bool,
+    ) -> IndexerResult<Vec<(TransactionSequenceNumber, TransactionDigest)>> {
+        self.paginate(
+            Key::TransactionDigestsByAddress(address),
+            cursor,
+            limit,
+            oldest_first,
+        )
+        .await
     }
 }
--- a/crates/iota-indexer/src/historical_fallback/reader.rs
+++ b/crates/iota-indexer/src/historical_fallback/reader.rs
@@ -65,7 +65,7 @@ pub(crate) struct HistoricalFallbackReader {
     /// storage through REST API interface.
     client: HttpRestKVClient,
     package_resolver: PackageResolver,
-    cache_cursor: MokaCache<TransactionDigest, TransactionSequenceNumber>,
+    cursor_cache: MokaCache<TransactionDigest, TransactionSequenceNumber>,
 }
 
 impl HistoricalFallbackReader {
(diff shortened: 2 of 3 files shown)


crates/iota-indexer/src/read.rs
```diff
@@ -3141,6 +3151,121 @@ impl<'a> DBReader<'a> {
...
+        is_descending: bool,
+    ) -> IndexerResult<Vec<StoredTransaction>> {
+        let tx_tables = [
+            CommitterTables::TxSenders,
+            CommitterTables::TxRecipients,
+            CommitterTables::Transactions,
+        ];
+
+        let cursor_tx_seq = if let Some(cursor) = cursor {
+            let tx_seq = self.resolve_cursor_tx_digest_to_seq_num(cursor).await?;
```
kodemartin: This fails with a `PostgresRead` error

https://github.com/iotaledger/iota/blob/17b38ed9a96a84590eaba4fc1989ca135d15722a/crates/iota-indexer/src/read.rs#L2744-L2753

which implies that `is_data_pruned || is_db_response_empty` will return `false` at L1213 and the fallback will not be triggered.

So `tx_digests` yields the lower threshold in pagination. If its retention period is the same as `tx_{sender,recipient}` we effectively don't use the fallback.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -497,4 +508,76 @@ impl HistoricalFallbackReader {
...
+        &self,
+        address: IotaAddress,
+        cursor: Option<TransactionDigest>,
+        limit: usize,
+        oldest_first: bool,
+    ) -> IndexerResult<Vec<TransactionDigest>> {
+        let cursor = match cursor {
+            Some(digest) => match self.cache_cursor.get(&digest) {
+                Some(tx_sequence_number) => Some(tx_sequence_number),
+                None => self.resolve_transaction_sequence_number(digest).await?,
```
kodemartin: If the transaction digest is not found this returns `None` which is the same as not providing a cursor. I reckon we should expect this to resolve into an `Ok(Some`. Otherwise, it is a user-input error no?

crates/iota-indexer/src/historical_fallback/client.rs
```diff
@@ -279,6 +318,75 @@ impl HttpRestKVClient {
...
+        if !resp.status().is_success() {
+            return Err(IndexerError::HistoricalFallbackStorageError(format!(
+                "multi_fetch request failed with status: {}",
+                resp.status()
+            )));
+        }
+
+        let bytes = resp.bytes().await?;
+        bcs::from_bytes::<Vec<T>>(&bytes).map_err(|e| {
+            IndexerError::Serde(format!("failed to deserialize multi_get response: {e:?}"))
```
kodemartin: `multi_fetch` and `multi_get` references are off here. Consider revising.

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1180,50 @@ impl IndexerReader {
             .await
     }
 
+    /// Fetches a paginated list of transactions that affect a given address,
+    /// using the fallback storage if the database is pruned.
+    async fn query_transactions_either_sender_or_recipient_address_with_fallback(
```
kodemartin: Simplify and generalize to
```suggestion
    async fn query_transactions_by_affected_addresses_with_fallback(
```

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -61,6 +65,7 @@ pub(crate) struct HistoricalFallbackReader {
     /// storage through REST API interface.
     client: HttpRestKVClient,
     package_resolver: PackageResolver,
+    cache_cursor: MokaCache<TransactionDigest, TransactionSequenceNumber>,
```
kodemartin: `cursor_cache` would be more intuitive I reckon.

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1180,50 @@ impl IndexerReader {
...
+        &self,
+        address: IotaAddress,
+        cursor: Option<TransactionDigest>,
+        limit: usize,
+        is_descending: bool,
+        options: iota_json_rpc_types::IotaTransactionBlockResponseOptions,
+    ) -> IndexerResult<Vec<IotaTransactionBlockResponse>> {
```
kodemartin, line 1214: Couldn't we check the `fallback.cache_cursor` first to see quickly if we are in the pruned data range?

kodemartin (commented):
Regarding this 
> The indexer keeps no per-address metadata about which transactions have been pruned. When the database serves a FromOrToAddress query without a cursor in ascending order, it has no signal to distinguish what the initial transaction sequence number for this address was.

Can't we make a fallback call to resolve the earliest sequence number for the given address and start pagination there?

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/2c96295b22d790abc4a670b8e12776bf4e0390be

crates/iota-indexer/src/historical_fallback/client.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/2c96295b22d790abc4a670b8e12776bf4e0390be

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/a1df163a92882bea4d0238adc372bd060962b9a0

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/a1df163a92882bea4d0238adc372bd060962b9a0

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/a1df163a92882bea4d0238adc372bd060962b9a0

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -79,12 +84,26 @@ impl HistoricalFallbackReader {
...
         })
     }
 
+    /// Returns the cached sequence number for a transaction digest.
+    pub(crate) fn cached_cursor(
+        &self,
+        cursor: &TransactionDigest,
+    ) -> Option<TransactionSequenceNumber> {
+        self.cursor_cache.get(cursor)
+    }
```
kodemartin: Do we need this helper method?. It just saves a `.get` and introduces two symbols with similar meaning: `cached_cursor`, `cursor_cache`.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -497,4 +516,81 @@ impl HistoricalFallbackReader {
...
+        let checkpoints = self.resolve_checkpoints(&[digest]).await?;
+        let (summary, contents) = checkpoints.get(&digest).cloned().ok_or_else(|| {
+            IndexerError::HistoricalFallbackInput(format!(
+                "checkpoint summary and contents linked to transaction: {digest} not found",
+            ))
+        })?;
+
+        let transaction_sequence_number = contents
+            .enumerate_transactions(&summary)
+            .find_map(|(seq, ed)| (ed.transaction == digest).then_some(seq));
```
kodemartin: I think we should just `.expect` here. FInding the parent checkpoint by the transaction digests, guarantees that we will find the transaction in the checkpoint, no?

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1202,101 @@ impl IndexerReader {
...
+
+        // fast path: cursor is known to KV (cached) and in pruned territory
+        // (below watermark). Serve directly from KV until pagination crosses
+        // into the unpruned DB range.
+        if let Some(kv_reader) = self
+            .fallback_reader()
+            .filter(|kv| matches!(cursor, Some(c) if kv.cached_cursor(&c).is_some_and(|s| (s as i64) < watermark)))
+        {
+            let stored_txs = kv_reader
+                .transactions_by_address(address, cursor, limit, !is_descending)
```
kodemartin: Nit: the filter closure does a lot. We could simplify by moving the `cursor.is_some()` check above

```suggestion
        if let Some((cursor, kv_reader)) = cursor
            .zip(self.fallback_reader())
            .filter(|(c, kv)| kv.cached_cursor(&c).is_some_and(|s| (s as i64) < watermark))
        {
            let stored_txs = kv_reader
                .transactions_by_address(address, Some(cursor), limit, !is_descending)
```

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1202,101 @@ impl IndexerReader {
...
+
+        let stored_txs = match db_res {
+            // top-up: a short DESC page during active pruning indicates we've crossed
+            // the watermark. Fetch the remaining pre-watermark rows from KV.
+            //
+            // note: 'watermark > 0' is essential, without pruning, a short page
+            // simply signals end-of-data.
+            Ok(mut rows) => match self
+                .fallback_reader()
+                .filter(|_| is_descending && rows.len() < limit && watermark > 0)
```
kodemartin: I get that the filter fits the conditions that trigger the logic in one line nicely, but at the same time it lowers them with respect to the existence of the fallback. Hence, we need the comment above to understand what happens.

So I wonder if the logic here could be streamlined. Here's an idea 

```
1. Handle the errors: `if let Err(e) = db_res` or `let Ok(stored_txs) = db_res else {}`
2. Then on Ok, branch from the main condition (rows.len() < limit && is_descending && watermark > 0)
```

crates/iota-indexer/src/read.rs
```diff
@@ -3141,6 +3224,150 @@ impl<'a> DBReader<'a> {
             })
             .collect()
     }
+
+    /// Returns a list of [`StoredTransaction`]s by their corresponding sequence
+    /// numbers.
+    ///
+    /// - `true` or `Some(true)` for DESC.
+    /// - `false` or `Some(false)` for ASC.
+    /// - `None` for undefined order.
```
kodemartin: I suggest mentioning  that this refers to `is_descending`

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1202,101 @@ impl IndexerReader {
...
+                                    tx.transaction_digest
+                                ))
+                            })
+                        })
+                        .transpose()?
+                        .or(cursor);
+
+                    let remaining = limit - rows.len();
+                    let kv_rows = kv_reader
+                        .transactions_by_address(address, cursor, remaining, !is_descending)
```
kodemartin: As `cursor` here we pass the digest of a transaction found in the database. This means we already know the transaction sequence number. 

But with `kv_reader.transactions_by_address` we have to resolve the sequence number again through the cache or through a query.  

If you insert the `(digest, sequence_number)` from the last row into the `cursor_cache` before calling this we would avoid the query to BitTable to resolve the sequence number.

crates/iota-indexer/src/read.rs
```diff
@@ -3141,6 +3224,150 @@ impl<'a> DBReader<'a> {
...
+    ///
+    /// Enforces data retention policies by consulting the pruning watermark for
+    /// the underlying index tables
+    /// [`TRANSACTIONS_BY_ADDRESS_TABLES`](IndexerReader::TRANSACTIONS_BY_ADDRESS_TABLES)
+    ///
+    /// # Errors
+    ///
+    /// * [`IndexerError::DataPruned`] — signals the caller to fall back to the
+    ///   historical store. Two triggers:
+    ///   * `cursor` resolves to a sequence number at or below the pruning
```
kodemartin: The cursor is resolved to a number strictly below through `ensure_data_not_pruned_for_tx`


```suggestion
    ///   * `cursor` resolves to a sequence number below the pruning
```

crates/iota-indexer/src/read.rs
```diff
@@ -1201,6 +1202,101 @@ impl IndexerReader {
...
+
+                    rows.extend(kv_rows);
+                    Ok(rows)
+                }
+                None => Ok(rows),
+            },
+            Err(e) => match self
+                .fallback_reader()
+                .filter(|_| matches!(e, IndexerError::DataPruned(_)))
+            {
```
tomxey: checking conditions in filter that are totally independent of the filtered value (kv) seems a bit awkward to me
have you considered writing it like:
```
match (db_res, self.fallback_reader()) {
...
(Err(DataPruned(_)), Some(kv_reader)) => ...
}
```

Here and also above for the Ok case.

sergiupopescu199: you're right, thank you! fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

crates/iota-indexer/src/read.rs
sergiupopescu199: Good catch thank you! Fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

crates/iota-indexer/src/read.rs
sergiupopescu199: Took into account also Tomasz's [suggestion](https://github.com/iotaledger/iota/pull/11807#discussion_r3387680533), the match still remains but is a bit more transparent now.

fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: Usually I'm a bit afraid when to panic but in this case you're right, better to catch the storage inconsistency.

fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

crates/iota-indexer/src/historical_fallback/reader.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/b28f6fa193774c99767ccc96bfad4135d0baa4d4

tomxey (approved):

crates/iota-indexer/src/read.rs
kodemartin: Nit: That's better. But please indulge me and consider this suggestion:

```rust
        let Some(kv_reader) = self.fallback_reader() else {
            return self
                .stored_transaction_to_transaction_block(db_res?, options)
                .await;
        };

        let (mut rows, db_pruned) = match db_res {
            Err(IndexerError::DataPruned(_)) => (vec![], true),
            res => (res?, false),
        };

        // Fill from KV: either the whole page (`db_pruned`) or a top-up when a
        // short DESC page during active pruning indicates we've crossed the watermark.
        if db_pruned || (is_descending && rows.len() < limit && watermark > 0) {
            // Continue pagination from the last DB row, or from the original
            // cursor if the DB returned nothing. Avoids restarting KV from the
            // top of history and re-fetching rows the user already saw.
            let kv_cursor = if let Some(tx) = rows.last() {
                let digest =
                    TransactionDigest::from_bytes(&tx.transaction_digest).map_err(|e| {
                        IndexerError::PersistentStorageDataCorruption(format!(
                            "failed to decode transaction digest: {:?} with err: {e:?}",
                            tx.transaction_digest
                        ))
                    })?;
                kv_reader
                    .cursor_cache
                    .insert(digest, tx.tx_sequence_number as u64);
                Some(digest)
            } else {
                cursor
            };

            let kv_rows = kv_reader
                .transactions_by_address(address, kv_cursor, limit - rows.len(), !is_descending)
                .await?
                .into_iter()
                .flatten();
            rows.extend(kv_rows);
        }

        self.stored_transaction_to_transaction_block(rows, options)
            .await

```

kodemartin (approved):

crates/iota-indexer/src/read.rs
sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11807/commits/69eed83149c2fe86f0d45cae114f33d824161280

PR #11809, opened 2026-06-08
feat(indexer): Remove 'objects_history'

### LINKED ISSUE #11701: [indexer]: Remove `objects_history` table
In https://github.com/iotaledger/iota/issues/11580 all places that read `objects_history` has been removed.

Following up on that we should remove the logic that writes to `objects_history` and remove the table itself.

tomxey:
# Description of change

Remove 'objects_history' and related code.
Create `objects_version` rows directly from checkpoint data, not through object history models.
Switch objects model from objects history to checkpointed objects table as it's still used.

fixes https://github.com/iotaledger/iota/issues/11701

### Release Notes

- [x] Indexer: ⚠️ Includes a migration that removes the `objects_history` table (expected duration: a few seconds). **Action required for operators**: remove any `objects_history = <epochs>` entry from your pruning configuration TOML before upgrading, from now on indexer will treat it as unknown table.


crates/iota-indexer/src/ingestion/primary/prepare.rs
```diff
@@ -209,17 +207,29 @@ impl PrimaryWorker {
...
 
-    fn derive_object_versions(
-        object_history_changes: &TransactionObjectChangesToCommit,
-    ) -> IndexerResult<Vec<StoredObjectVersion>> {
-        let mut object_versions = vec![];
-        for changed_obj in object_history_changes.changed_objects.iter() {
-            object_versions.push(changed_obj.try_into()?);
-        }
-        for deleted_obj in object_history_changes.deleted_objects.iter() {
-            object_versions.push(deleted_obj.into());
```
kodemartin, line 220: I guess by removing these, we lose the need for the `try_into` and `into` conversions for `StoredObjectVersion`. Unless it makes sense to reuse them below.

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -74,7 +74,6 @@ pub struct Pruner {
 #[serde(rename_all = "snake_case")]
 #[non_exhaustive]
 pub enum PrunableTable {
-    ObjectsHistory,
```
kodemartin, line 77: Does it have implications for operators with a `objects_history = 2` setting in the pruning configuration file? 

If so include in the release notes.

kodemartin (commented):
Looks good generally. A couple minor comments.

Missing the release notes as well, informing about the migration, and any implications for pruning configuration.

crates/iota-indexer/src/ingestion/primary/prepare.rs
tomxey: indeed, removed in 567970d3ad

crates/iota-indexer/src/pruning/pruner.rs
tomxey: added release notes section

kodemartin (approved):

sergiupopescu199 (approved):
lgtm 🐧

PR #11810, opened 2026-06-08
chore: fix clippy warnings in cfg(msim) / simtest-gated code

### LINKED ISSUE #7999: Fix clippy warnings for node-team related crates
We should fix all the clippy warnings in our code, also the simtest related ones.

chriscandless:
# Description of change

`cargo ci-clippy` (the CI clippy gate) builds **without** `--cfg msim`, so clippy never lints the simtest-only (`cfg(msim)`) code paths and warnings had accumulated there unnoticed. Running `scripts/simtest/clippy.sh` surfaces them.

This fixes **one compile error and all clippy warnings** that script reports across the node-team crates, so `scripts/simtest/clippy.sh` now passes with `-D warnings`. The normal `cargo ci-clippy` stays green as well.

Highlights:

- **iota-genesis-builder** — the optional `iota-sdk-crypto` dep lived under `[target.'cfg(not(msim))'.dependencies]`, but the `test_outputs` module that uses it is gated only by the `test-outputs` feature, so `--all-features` under msim failed to compile (E0432). Moved it to `[dependencies]` (still optional).
- **iota-core** — `.map(|p| p)` (an identity map kept only to shorten the `'static` built-in-package refs so they `chain` with the locally-owned extras) tripped `map_identity`; replaced with `collect` + `extend`, which satisfies the lint without the lifetime breakage that simply removing the map causes.
- **starfish-simtests** — gated the test-only `node`/`simtests` modules to `cfg(all(msim, test))` (kills a `dead_code` on `RestartMode` variants that are constructed only in `#[sim_test]` fns); `anyhow::Result` → `eyre::Result` (disallowed-types); plus `option_map_unit_fn` / `unnecessary_map_or` / `unnecessary_cast`.
- **starfish-core** — dropped the unused `context`/params from the orphaned `delete_all_transactions_from_store` helper; rstest `#[values]` → `#[case]` for snake_case generated test names (the msim `#[tokio::test]` wrapper otherwise produces `_params_1__true_true__internal`).
- **iota/tests/shell_tests.rs** — `#[cfg(not(msim))]` on the non-msim test helpers (matching `ptb_files_tests.rs`) to fix `dead_code`.
- **iota-types / iota-config / iota-node / reconfiguration_tests** — assorted `uninlined_format_args` and an unused `use super::*`.

Two decisions worth a reviewer's eye:

1. `RestartMode` `dead_code` is fixed by cfg-gating the test modules rather than `#[expect(dead_code)]`. The crate already uses `#[expect]` elsewhere, so say the word if you'd prefer that for consistency.
2. `delete_all_transactions_from_store` has no callers yet — the in-progress erase-transactions path currently goes through `ConsensusAuthority::stop_and_clear_transactions`. I simplified its signature rather than keep dead params; happy to restore them if they're intended scaffolding.

> Follow-up (out of scope here): these warnings will regress unless `scripts/simtest/clippy.sh` is wired into CI — worth a separate issue.

fixes #7999

Verified locally:
- `scripts/simtest/clippy.sh` (i.e. `cargo ci-clippy` with `--cfg msim`) → **0 warnings**.
- normal `cargo ci-clippy` (`--all-targets --all-features -D warnings`) → **0 warnings**.
- `cargo +nightly fmt --check`, `dprint check`, `scripts/cargo_sort` consolidate → clean.

The change is a behavior-preserving lint/build cleanup: all targets (incl. tests) compile under both configs, but the test **suite was not executed** — no runtime behavior is altered, so no new tests are added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13144,8 +13144,8 @@ dependencies = [
 name = "starfish-simtests"
 version = "0.1.0"
 dependencies = [
- "anyhow",
  "arc-swap",
+ "eyre",
  "iota-config",
  "iota-macros",
  "iota-metrics",
--- a/crates/iota-config/src/local_ip_utils.rs
+++ b/crates/iota-config/src/local_ip_utils.rs
@@ -39,7 +39,7 @@ impl SimAddressManager {
             .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
         // If offset ever goes beyond 255, we could use more bytes in the IP.
         assert!(offset <= 255);
-        format!("10.10.0.{}", offset)
+        format!("10.10.0.{offset}")
     }
 
     pub fn get_next_available_port(&self) -> u16 {
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -4870,7 +4870,11 @@ impl AuthorityState {
         #[cfg(msim)]
         let extra_packages = framework_injection::get_extra_packages(self.name);
         #[cfg(msim)]
-        let system_packages = system_packages.map(|p| p).chain(extra_packages.iter());
+        let system_packages = {
+            let mut packages: Vec<_> = system_packages.collect();
+            packages.extend(extra_packages.iter());
+            packages
+        };
 
         for system_package in system_packages {
             let modules = system_package.modules().to_vec();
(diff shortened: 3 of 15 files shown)


muXxer (approved):

piotrm50 (approved):
Nice! Happy that this is finally resolved!

valeriyr (approved):

Thoralf-M (approved):
Approving the changes in the iota crate

PR #11813, opened 2026-06-08
fix(core-node): bump anemo to recent revision

### LINKED ISSUE #11295: Panic during validator shutdown
## Bug description

Panic during a restart with systemd:

```
2026-04-22T08:56:54.079532+00:00 iota-validator-mainnet-01 systemd[1]: Stopping iota-node.service - Iota Node Client...
....
2026-04-22T08:56:56.049701+00:00 iota-validator-mainnet-01 systemd[1]: Started iota-node.service - Iota Node Client.
2026-04-22T08:56:56.031245+00:00 iota-validator-mainnet-01 systemd[1]: Starting iota-node.service - Iota Node Client...
2026-04-22T08:56:56.011323+00:00 iota-validator-mainnet-01 systemd[1]: iota-node.service: Main process exited, code=killed, status=6/ABRT
2026-04-22T08:56:54.080369+00:00 iota-validator-mainnet-01 iota-node[3858496]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2026-04-22T08:56:54.080355+00:00 iota-validator-mainnet-01 iota-node[3858496]: called `Result::unwrap()` on an `Err` value: JoinError::Cancelled(Id(350423237)) panic.file="/home/runner/.cargo/git/checkouts/anemo-d4d262aa9f0cef1a/c4e7e4c/crates/anemo/src/network/connection_manager.rs" panic.line=151 panic.column=69
2026-04-22T08:56:54.080304+00:00 iota-validator-mainnet-01 iota-node[3858496]: 2026-04-22T08:56:54.079962Z ERROR connection-manager{peer=df0873d9}: telemetry_subscribers: panicked at /home/runner/.cargo/git/checkouts/anemo-d4d262aa9f0cef1a/c4e7e4c/crates/anemo/src/network/connection_manager.rs:151:69:
2026-04-22T08:56:54.080225+00:00 iota-validator-mainnet-01 iota-node[3858496]: called `Result::unwrap()` on an `Err` value: JoinError::Cancelled(Id(350423237))
```

Systemd setup using:
```
StandardOutput=journal
StandardError=journal
KillMode=process
TimeoutSec=180
Restart=always
RestartSec=60
```

## Quick Claude analysis:

```
This is an anemo bug that surfaces during an ungraceful Tokio runtime teardown. Here's the chain:

  The panic site — anemo/src/network/connection_manager.rs:151:
  Some(connecting_output) = self.pending_connections.join_next() => {
      self.handle_connecting_result(connecting_output.unwrap());  // <- col 69
  }
  pending_connections is a JoinSet<ConnectingOutput> that holds in-flight QUIC handshake tasks. join_next() returns
  Option<Result<T, JoinError>>. The code unwraps the inner Result, assuming the task always completes normally — but
  JoinError::Cancelled(Id(350423237)) means a task in the set was aborted, and the unwrap panics. The panic triggers abort()
  (systemd's status=6/ABRT), which is why the node exits and systemd restarts it one second later.

  Why it happens on shutdown (and only then)

  Anemo's graceful shutdown path is fine — ConnectionManagerRequest::Shutdown breaks the event loop before shutdown().await calls
  pending_connections.shutdown().await (which joins aborted handles without unwrapping). The bug fires when tasks in
  pending_connections are cancelled while the event loop is still spinning — i.e. before the Shutdown mailbox message has been
  processed. Ways that happens:

  1. Tokio runtime drop cancels tasks — if iota-node drops the runtime on SIGTERM without first calling NetworkRef::shutdown on
  the anemo network and awaiting the oneshot, then all spawned tasks get cancelled. The connection-manager task itself reports the
   cancel through join_next().
  2. Explicit abort elsewhere on the JoinSet — I don't see this path in anemo, so (1) is the likely cause.

  The peer=df0873d9 field in the span and the panic landing on line 151 (pending connections) rather than line 155 (established
  connection handlers) points to an in-flight outbound dial getting cancelled mid-handshake as the runtime winds down.

  What to do about it

  - Proper fix (upstream): in anemo's select branch, tolerate JoinError::Cancelled — match connecting_output { Ok(out) =>
  self.handle_connecting_result(out), Err(e) if e.is_cancelled() => {} , Err(e) => std::panic::resume_unwind(e.into_panic()) }.
  The mysten fork is what IOTA pins (Cargo.toml:238 at rev c4e7e4c), so a follow-up is either a bump to a revision that has this
  or a vendor patch.
  - Workaround on the iota-node side: ensure shutdown explicitly calls NetworkRef::shutdown on the anemo p2p network
  (crates/iota-node/src/lib.rs:1189-1290) and awaits the oneshot before the outer runtime is dropped. Worth checking whether your
  shutdown orchestration actually does this on SIGTERM or just returns from main and lets the runtime drop tear everything down.
  - Impact assessment: this is benign — it only happens during exit and after the validator has already stopped serving. Systemd
  restarts cleanly one second later (your logs confirm). It does, however, pollute metrics/alerts for "unexpected exits" and
  breaks a clean status=0 exit, which matters for dashboards that treat status=6/ABRT as a crash.
```

semenov-vladyslav:
# Description of change

This PR bumps anemo revision to the most recent revision fixing issue with validator shutdown.

Fixes #11295.

### Release Notes

- [x] Nodes (Validators and Full nodes): Fix a rare panic during validator shutdown.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,7 +152,7 @@ dependencies = [
 [[package]]
 name = "anemo"
 version = "0.0.0"
-source = "git+https://github.com/mystenlabs/anemo.git?rev=c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2#c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2"
+source = "git+https://github.com/mystenlabs/anemo.git?rev=f8b708601d002a75955dafebae4f75628c6f42b4#f8b708601d002a75955dafebae4f75628c6f42b4"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -187,7 +187,7 @@ dependencies = [
 [[package]]
 name = "anemo-build"
 version = "0.0.0"
-source = "git+https://github.com/mystenlabs/anemo.git?rev=c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2#c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2"
+source = "git+https://github.com/mystenlabs/anemo.git?rev=f8b708601d002a75955dafebae4f75628c6f42b4#f8b708601d002a75955dafebae4f75628c6f42b4"
 dependencies = [
  "prettyplease",
  "proc-macro2",
@@ -198,7 +198,7 @@ dependencies = [
 [[package]]
 name = "anemo-cli"
 version = "0.0.0"
-source = "git+https://github.com/mystenlabs/anemo.git?rev=c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2#c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2"
+source = "git+https://github.com/mystenlabs/anemo.git?rev=f8b708601d002a75955dafebae4f75628c6f42b4#f8b708601d002a75955dafebae4f75628c6f42b4"
 dependencies = [
  "anemo",
  "anemo-tower",
@@ -214,7 +214,7 @@ dependencies = [
 [[package]]
 name = "anemo-tower"
 version = "0.0.0"
-source = "git+https://github.com/mystenlabs/anemo.git?rev=c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2#c4e7e4cb4b624d7738c2016d7b7885c6774ff9c2"
+source = "git+https://github.com/mystenlabs/anemo.git?rev=f8b708601d002a75955dafebae4f75628c6f42b4#f8b708601d002a75955dafebae4f75628c6f42b4"
 dependencies = [
  "anemo",
  "bytes",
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -9851,7 +9851,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
```
bingyanglin, line 9854: Is this related to the anemo bump?

Thoralf-M (approved):

Cargo.lock
muXxer: Should be fine. It was resolved by cargo this way.

muXxer (approved):

PR #11236, opened 2026-04-14
feat(docs): PA - Audit Trails wiki

itsyaasir:
# Description of change

This PR introduces the **IOTA Notarization Toolkit** documentation and restructures the existing notarization docs around it. The Notarization Toolkit (part of the IOTA Trust Framework) bundles two components, and the docs are now organized to reflect that:

- **Single Notarization** — the existing notarization pages, moved under `single-notarization/`.
- **Audit Trails** — a brand-new, complete documentation set.

### What changed

**New: Audit Trails component documentation**

A full Diátaxis-structured documentation set under `docs/content/developer/iota-notarization/audit-trails/`:

- **Introduction** (`index.mdx`) — component landing page.
- **Explanations** — `about-audit-trails`, `role-based-access-control`, `locking-configuration`, `privacy-and-off-chain-documents`, `events`.
- **Getting started** — `rust`, `wasm`, `local-network-setup`.
- **How-tos** — creating/deleting audit trails, adding/reading/deleting records, managing access control and record tags, configuring locking, enforcing capability constraints, updating metadata, accessing read-only methods, and utility functions.
- **Real-world examples** — `clinical-trial`, `customs-clearance`, `digital-product-passport`.
- **References** — auto-generated Wasm API docs (download pipeline), external Rust API link.

**Restructure: Notarization Toolkit layout**

- Moved all existing Single Notarization pages from `iota-notarization/{explanations,getting-started,how-tos}/...` into `iota-notarization/single-notarization/...`, and `how-tos/real-world/` into `single-notarization/real-world-examples/`.
- Added a new Toolkit landing page (`iota-notarization/index.mdx`) introducing both components, plus a new `single-notarization/index.mdx` component landing page.
- Unified both components under a single `Notarization` sidebar (`docs/content/sidebars/notarization.js`), ordered by reading path (introduction → getting-started → explanations → how-tos → references).
- New `single-notarization` and `audit-trails` component tags in `tags.yml`.
- New banner assets (`banner_audit_trails.png`, `banner_single_notarization.png`).

**Build pipeline**

- Updated `docs/site/docusaurus.config.js`, `docs/site/package.json`, and `docs/site/scripts/get-iota-notarization-references.sh` to download and place the per-component Wasm reference docs and external Audit Trails Rust references.
- Code examples are referenced from the external `iotaledger/notarization` repo using the `v0.1` tag (Single Notarization) and `trails-v0.1` tag (Audit Trails).

**Naming conventions**

[...]

PR #11814, opened 2026-06-09
fix(docs): correct notarization wasm references copy path

Thoralf-M:
# Description of change

The scheduled **Release Wiki** workflow has been failing in the Docusaurus build ([example run](https://github.com/iotaledger/iota/actions/runs/27177256168)) with `Sidebar category Wasm has neither any subitem nor a link`.

## Root cause

The notarization **Wasm** sidebar category is autogenerated from `developer/iota-notarization/references/wasm`, which is populated at build time by `get-iota-notarization-references.sh`. The notarization `wasm.tar.gz` was repackaged with an extra top-level `notarization-docs/` prefix, so the script's `cp` from `./docs/wasm/*` silently copied nothing, leaving the directory empty and the sidebar category empty (which Docusaurus rejects). Identity/hierarchies use the flat `docs/wasm/` layout, so only notarization broke.

## Changes

1. Copy from `./notarization-docs/docs/wasm/*` to match the new tarball layout.
2. Patch a malformed link in the upstream-generated `LockMetadata.md` (`NotarizationClient.mdupdatemetadata` → `NotarizationClient.md#updatemetadata`), which otherwise fails the build under `onBrokenLinks: throw`. Tracked upstream in iotaledger/notarization#282; remove the `sed` workaround once a fixed tarball is published.

https://claude.ai/code/session_01BJxVLbmQFgJxFajfMRi8Po

### DIFF
--- a/docs/site/scripts/get-iota-notarization-references.sh
+++ b/docs/site/scripts/get-iota-notarization-references.sh
@@ -8,7 +8,7 @@ cd tmp
 curl -sL https://s3.eu-central-1.amazonaws.com/files.iota.org/iota-wiki/iota-notarization/0.1/wasm.tar.gz  | tar xzv
 # Create the target directory structure if it doesn't exist
 mkdir -p ../../content/developer/iota-notarization/references/wasm
-cp -Rv ./docs/wasm/* ../../content/developer/iota-notarization/references/wasm/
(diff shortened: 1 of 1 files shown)


Thoralf-M:
https://github.com/iotaledger/iota/pull/11236 will replace fix it

PR #11818, opened 2026-06-09
feat: historical epoch boundaries

### LINKED ISSUE #11694: [data-ingestion{,-core}] store epoch_boundaries in historical checkpoint store

kodemartin:
# Description of change

This replays already approved patches that support writing epoch boundaries in the historical checkpoint store.

Closes #11694

### DIFF
--- a/crates/iota-data-ingestion-core/src/errors.rs
+++ b/crates/iota-data-ingestion-core/src/errors.rs
@@ -55,6 +55,9 @@ pub enum IngestionError {
     #[error("reading historical data failed: `{0}`")]
     HistoryRead(String),
 
+    #[error("invalid epoch boundary update: `{0}`")]
+    EpochBoundary(String),
+
     #[error("max downloaded checkpoints limit reached")]
     MaxCheckpointsCapacityReached,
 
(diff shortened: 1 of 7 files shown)


sergiupopescu199 (approved):
lgtm 🎸

tomxey (approved):

PR #11826, opened 2026-06-09
feat(iota-localnet): configure indexer prune policy via --pruning-config-path

### LINKED ISSUE #11804: [iota-localnet] Allow configuration of the indexer prune policy
## Description

In order to limit the volume of data used by the indexer on a localnet, I would like to to configure the prune policy when the indexer is started by the `iota-localnet start ... --with-indexer ...` command.

## Open questions (optional)

Perhaps add the parameter "--pruning-config-path", like the one that exists on the `iota-indexer indexer ...` command?

No.

Thoralf-M:
# Description of change

`iota-localnet start --with-indexer` started the indexer writer with no retention policy, so the indexer database grew without bound. This flattens the indexer's `PruningOptions` into the localnet indexer args, exposing `--pruning-config-path` (and `--epochs-to-keep`) and wiring it through to the writer so retention can be configured.

As part of this, `IndexerTypeConfig::writer_mode` now builds its retention config via `PruningOptions::load_from_file()` so it honors `--pruning-config-path` (it previously read only `epochs_to_keep` and ignored the config file). That removed the only caller of `RetentionConfig::new_with_default_retention_only_for_testing`, so the now-dead helper was deleted — it was also a no-op-with-extra-steps (it built an `objects_history` override map and then discarded it, making it equivalent to `RetentionConfig::new(epochs_to_keep, HashMap::new())`).

[run-ci]

fixes #11804

Ran a localnet against Postgres: with `--pruning-config-path` the pruner starts and the indexer DB reaches a steady state (old checkpoints deleted); without it, no pruner runs and the DB grows unbounded.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EcFDLESTbAyuetw2PosNEa)_

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `iota-localnet start --with-indexer` now accepts `--pruning-config-path` to configure indexer data retention (the DB no longer grows without bound).
- [ ] Rust SDK:
- [ ] gRPC:


docs/content/developer/references/cli/localnet.mdx
```diff
@@ -70,10 +70,25 @@ When built with the `indexer` feature, additional options are available:
 | `--pg-db-name <NAME>` | Postgres database name (default: `iota_indexer`). |
 | `--pg-user <USER>` | Postgres username (default: `postgres`). |
 | `--pg-password <PASS>` | Postgres password (default: `postgrespw`). |
+| `--pruning-config-path <PATH>` | Path to a TOML retention config for the indexer writer (see [Indexer pruning config file](#indexer-pruning-config-file) below). Without it the indexer keeps all data and its database grows without bound. |
```
thibault-martinez, line 73: If `PruningOptions` is flattened, we're also getting `--epochs-to-keep` and `--optimistic-pruner-batch-size`? How would that clash with the same options within the file if they exist?

crates/iota-indexer/src/test_utils.rs
```diff
@@ -77,10 +77,10 @@ impl IndexerTypeConfig {
 
     pub fn writer_mode(pruning_options: Option<PruningOptions>) -> Self {
         Self::Writer {
-            retention_config: pruning_options.as_ref().and_then(|pruning_options| {
+            retention_config: pruning_options.and_then(|pruning_options| {
                 pruning_options
-                    .epochs_to_keep
-                    .map(RetentionConfig::new_with_default_retention_only_for_testing)
+                    .load_from_file()
```
thibault-martinez: Kind of the same question as in the docs. This load options from file but what if `pruning_options` already carried options?

docs/content/developer/references/cli/localnet.mdx
Thoralf-M: Yes — flattening also brings in `--epochs-to-keep` and `--optimistic-pruner-batch-size`, but neither clashes with the file:

- `--epochs-to-keep` vs the file's `epochs_to_keep`: resolved in `PruningOptions::load_from_file()` — when `--pruning-config-path` is set the file wins and `--epochs-to-keep` is ignored (with a warning); with no path, the flag value is used.
- `--optimistic-pruner-batch-size`: not part of the TOML format and a deprecated no-op, so there's nothing for it to clash with.

Both already render as `DEPRECATED` in `--help` (`--epochs-to-keep` is slated for removal in v1.28.0), and since localnet flattens the same `PruningOptions` struct, they'll drop off automatically when removed upstream.

---

thibault-martinez: I see, thanks

thibault-martinez (approved):

tomxey (approved):

PR #11835, opened 2026-06-10
fix(graphql-rpc): resolve child parent relations consistently

### LINKED ISSUE #11834: [graphql-rpc]: object owner is not resolved consistently
## Description

Currently when resolving object owners we do so by querying for the latest object at the `checkpoint_viewed_at`

https://github.com/iotaledger/iota/blob/cb23c42bf8dd9cd68c6d411e8efeb014e8f0f46a/crates/iota-graphql-rpc/src/types/object.rs#L653-L659

This fetches the maximum version of the object at the `checkpoint_viewed_at`. 

But in the context of dynamic object fields, that resolve into an `Object` owner, the version of each child is constrained by the root parent version by the invariant `child_version <= root_parent_version`. This invariant is not satisfied necessarily through the `latest_at` lookup. 

It is reminded that when adding a new dynamic-object field, the `Child` object is owned by the `Field` object that represents dynamic fields, which in turn is owned by the root `Parent` object. 

Suppose that in the same checkpoint 

1. We add a new `Child` as a dynamic-object field in `Parent`. This binds the versions of `Parent`, `Field`, `Child` to the same value, say `N`.
2. `Parent` is mutated: Bumps to version `N+1`
3. `Child` is removed from the dynamic object fields of `Parent`, so the `Field` object is deleted at version `N+2`.

If we now lookup for the owner of `Child` at version `N` through the `latest_at` operation , we would expect to get the `Field` at version `N`. But with the `latest_at` lookup we will get the latest deleted version of `Field` which is the max version for the checkpoint.

To fix this we should use instead the `under_parent` lookup for resolving this relations, which gets the maximum version of the object that satisfies `<= root_parent_version`.

This emerged while working on #11619 where the `immutable_dof.move` consistency test in `iota-graphql-e2e-tests` was fetching the incorrect deleted version of the owner.

kodemartin:
# Description of change

This patch fixes the resolution of consistent object owners in GraphQL. To illustrate the problem (and the solution) a dedicated test `owner.move` is hereby added, that was returning incorrect results with the original setting. See 3ebd14341a1d14c70bf2dd4955221dfdb6fb9267

Fixes #11834

### Release Notes

- [x] GraphQL: :screwdriver: Fixes a bug in the consistent resolution of object owners

### DIFF
--- a/crates/iota-graphql-e2e-tests/tests/consistency/dynamic_fields/owner.move
+++ b/crates/iota-graphql-e2e-tests/tests/consistency/dynamic_fields/owner.move
@@ -1,0 +1,203 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// 1. create Parent (version 2)
+// 2. create Child (version 3)
+// 3. add Child to Parent -> Parent (v4) - Field(v4) - Child (v4)
+// 4. mutate Parent -> Parent (v5)
+// 5. reclaim Child -> Parent(v6), Child (v6), Field deleted
+
+// We check consistency by querying the dynamic fields rooted on parent
+// throughout the parent's evolution.
+
+// The stored values should resolve the owner to the correct version of the Field.
+
+// Parent - dynamic fields resolve the owner correctly to Field1(v4)
+// Child1 (v4) should have the Field as an owner
+// Child1 (v5) does not exist
+// Child1 (v6) does not have an owner
+
+//# init --protocol-version 29 --addresses Test=0x0 --accounts A --simulator
+
+//# publish
+module Test::M1 {
+    use iota::dynamic_object_field as ofield;
+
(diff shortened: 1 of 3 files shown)


sergiupopescu199 (approved):
lgtm 🚀

crates/iota-graphql-e2e-tests/tests/consistency/dynamic_fields/owner.move
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) Mysten Labs, Inc.
```
tomxey, line 1: nit: do we consider this file authored by mysten?

tomxey (approved):

crates/iota-graphql-e2e-tests/tests/consistency/dynamic_fields/owner.move
kodemartin: it reuses existing test move contracts, so attribution seems proper

PR #11841, opened 2026-06-10
feat: allow mutable shared objects in Move authenticators (POC)

### LINKED ISSUE #11806: [AA] Investigate on the possibility of making an authenticator function stateful

valeriyr:
# Description of change

Enables using mutable shared objects in authenticators.

> [!WARNING]  
> It is a POC, not an implementation that can be considered as ready for being merged.

fixes #11806


miker83z:
It seems feasible to make the authenticator stateful, so we started a design for the implementation of it.

PR #11845, opened 2026-06-11
refactor(graphql)!: remove WRAPPED_OR_DELETED status

### LINKED ISSUE #11619: [graphql]: remove WrappedOrDeleted object status

kodemartin:
# Description of change

This patch removes the deprecated `WRAPPED_OR_DELETED` status for objects. Such objects now resolve as non-existent.

In this context the following additional changes have been implemented:

- Refactored the related SQL queries to filter out non-active objects in the databse. This affects the backward-view source queries (`consistent.rs`, `historical.rs`) and the `HistoricalKey` and   `ParentVersionKey` loaders. The queries now return only rows whose `object_status` is  `Active`, applied before pagination so pages fill correctly. 
- Remove the obsolete API: `resolve_tombstone_versions`, the synthetic  `tombstones_from_objects_version` source, the `from_backward_history` column and field, `has_type_or_owner`, and the watermark-based synthesis in the loaders.
- Refactored `ObjectKind` from an enum into a struct `ActiveObject` holding `native`, `status`, and optional `bcs`. This removes the semantic overlapping with `ObjectStatus`, and the redundancy of storing  `StoredHistoryObject`.
- Implemented `TryFrom<StoredHistoryObject> for ObjectKind` which succeeds only for active non-corrupted objects.

Resolves #11619

- [x] Patch-specific tests (correctness, functionality coverage): `graphql-e2e-tests` have been updated to account for the change.

### Release Notes

- [x] GraphQL: :warning: Removes the  `WRAPPED_OR_DELETED` variant of `ObjectKind`. This kind of objects are now treated as non-existent.


crates/iota-graphql-e2e-tests/tests/consistency/object_at_version.move
```diff
@@ -123,18 +123,6 @@ module Test::M1 {
       }
     }
   }
-  previous_version: object(
```
tomxey, line 126: could we keep it to test that it returns empty result?

crates/iota-graphql-e2e-tests/tests/prune/backward_history_pruning.move
```diff
@@ -69,12 +69,6 @@ module P0::m {
 
 //# run-graphql
 {
-  real_tombstone_v3: objects(filter: {objectKeys: [{objectId: "@{obj_2_0}", version: 3}]}) {
-    nodes { status version }
-  }
-  lamport_v6: objects(filter: {objectKeys: [{objectId: "@{obj_2_0}", version: 6}]}) {
-    nodes { status version }
-  }
```
tomxey, line 77: maybe we could leave them here to ensure object doesn't pop up in response for neither real or synthetic tombstone?

crates/iota-graphql-rpc/src/backward_view/mod.rs
```diff
@@ -45,40 +45,37 @@ impl TryFrom<ObjectFilter> for HistoricalFilter {
...
-/// from backward diff results.
-pub(super) const NOT_YET_CREATED: i16 = BackwardHistoryObjectStatus::NotYetCreated as i16;
+/// Active status in `checkpointed_objects` (`ObjectStatus`). Backward-view
+/// sources keep only these rows; wrapped or deleted tombstones and
+/// `NotYetCreated` markers are excluded.
+pub(super) const CHECKPOINTED_ACTIVE: i16 = NativeObjectStatus::Active as i16;
+
+/// Active status in `objects_backward_history` (`BackwardHistoryObjectStatus`).
+/// Counterpart of [`CHECKPOINTED_ACTIVE`] for the backward-history table.
+pub(super) const HISTORY_ACTIVE: i16 = BackwardHistoryObjectStatus::Active as i16;
```
tomxey, line 60: I think we could merge those two constants into one. I think many queries read from both tables and merge the results treating the columns in the same way, no matter what table they came from, if the enums in those 2 tables would have different numbers assigned we would have correctness issues in other places in code.

Maybe we could have one constant only plus some compile time assert that those are actually the same? If it's possible in rust.

crates/iota-graphql-rpc/src/backward_view/mod.rs
```diff
@@ -45,40 +45,37 @@ impl TryFrom<ObjectFilter> for HistoricalFilter {
...
-/// Column list for `objects_backward_history` rows, tagged with
-/// `from_backward_history = TRUE`.
+/// Column list selected from `objects_backward_history` for backward-view
+/// queries.
 pub(super) const HISTORY_COLUMNS: &str = "\
     object_id, object_version, object_status, \
     object_digest, owner_type, owner_id, object_type, object_type_package, object_type_module, \
-    object_type_name, serialized_object, coin_type, coin_balance, df_kind, \
-    TRUE AS from_backward_history";
+    object_type_name, serialized_object, coin_type, coin_balance, df_kind";
```
tomxey: Those are the same now, did you consider merging them?

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -716,16 +698,12 @@ impl ObjectImpl<'_> {
...
-            K::Indexed(_, stored) => stored.serialized_object.as_ref().map(Base64::from),
-
-            K::NotIndexed(native) => {
-                let bytes = bcs::to_bytes(native)
+        let kind = &self.0.kind;
+        Ok(match &kind.bcs {
+            Some(serialized) => Some(Base64::from(serialized)),
+
+            None => {
+                let bytes = bcs::to_bytes(&kind.native)
```
tomxey: Is `bcs` field only an optimization? Since seems like we can always compute it from `bcs::to_bytes(&kind.native)`

crates/iota-graphql-rpc/src/backward_view/consistent.rs
```diff
@@ -42,9 +43,12 @@ fn consistent_checkpointed_objects(
...
     filter_fn: &impl Fn(RawQuery) -> RawQuery,
 ) -> RawQuery {
-    let checkpointed_filtered = filter_fn(query!(format!(
-        "SELECT {CHECKPOINTED_COLUMNS} FROM checkpointed_objects"
-    )));
+    let checkpointed_filtered = filter!(
+        filter_fn(query!(format!(
+            "SELECT {CHECKPOINTED_COLUMNS} FROM checkpointed_objects"
+        ))),
+        format!("object_status = {CHECKPOINTED_ACTIVE}")
```
tomxey, line 47: Do we have some issue created to remove non-active rows from `checkpointed_objects` table in the future?

crates/iota-graphql-rpc/src/backward_view/mod.rs
kodemartin: A compile time assert would by to somehow consolidate the types. And I think making it type-safe would be the way to go, instead of handling it here.

I will give it a thought, and either refactor or open an new issue to deal with this.

crates/iota-graphql-rpc/src/backward_view/mod.rs
kodemartin: No, but it makes sense.

crates/iota-graphql-rpc/src/types/object.rs
kodemartin: yes that exactly what it is. We skip the computation when we deal with Indexed objects, which is the majority of the cases.

crates/iota-graphql-rpc/src/backward_view/consistent.rs
kodemartin: No, but it makes sense that we simplify the schema and the Rust API in the context of removing the statue from JSON-RPC. Will create an issue in due time.

crates/iota-graphql-rpc/src/types/object.rs
tomxey: could we mention it shortly somewhere, e.g. next to the place where field is declared in the struct?

crates/iota-graphql-rpc/src/types/object.rs
```diff
@@ -85,16 +85,14 @@ pub(crate) struct Object {
...
-#[expect(clippy::large_enum_variant)]
-pub(crate) enum ObjectKind {
-    /// An object loaded from serialized data, such as the contents of a
-    /// transaction that hasn't been indexed yet.
-    NotIndexed(NativeObject),
-    /// An object fetched from the index.
-    Indexed(NativeObject, StoredHistoryObject),
-    /// The object is wrapped or deleted and only partial information can be
-    /// loaded from the indexer. The `u64` is the version of the object.
-    WrappedOrDeleted(u64),
```
tomxey, line 97: were there any issues with just removing this enum variant? (as opposed to refactoring the enum into a struct)

kodemartin: no, the refactoring seemed simply like a warranted simplification of the API.

crates/iota-graphql-e2e-tests/tests/consistency/object_at_version.move
kodemartin: 1ddb327adc it is not empty though, and rightly so

crates/iota-graphql-e2e-tests/tests/prune/backward_history_pruning.move
kodemartin: 1ddb327adc

crates/iota-graphql-rpc/src/backward_view/mod.rs
kodemartin: https://github.com/iotaledger/iota/issues/11861

crates/iota-graphql-rpc/src/backward_view/mod.rs
kodemartin: 191240db25

crates/iota-graphql-rpc/src/types/object.rs
kodemartin: 91f9975e54

sergiupopescu199 (approved):
lgtm 🔬

tomxey (approved):

PR #11846, opened 2026-06-11
feat(indexer): simplify pruner logic and add pruning options

### LINKED ISSUE #11381: [indexer]: simplify pruner logic
## Description

`iota_indexer::pruner` defines two enums, `PruningStrategy` and `PruningChunk`, with a strict 1-to-1 mapping. `PruningStrategy` is consulted only to choose `range_end` and the per-chunk batch size in `create_pruning_chunks`. `PruningChunk` then carries the resulting `(start, end)` range to `prune_by_chunk` for execution.

This split forces every new strategy to touch six places (both enums, `next_chunk_start`, `pruning_strategy`, `create_pruning_chunks`, `prune_by_chunk`) and produces duplicate code in `create_pruning_chunks`:

- `ByCheckpoint` and `ByCheckpointWithLimit` build chunks identically (same `range_end = min_available_cp`, same batch size).
- `ByTransaction` and `ByGlobalSeq` likewise (same `range_end = min_available_tx`, same batch size).
- All five `PruningChunk::next_chunk_start` arms return `end + 1`.

The variant differences only matter at prune time, not at chunk-creation time.

It seems thus that we can simplify this and make future contributions minimal and more intuitive by keeping `PruningStrategy` as the single source of truth and deleting `PruningChunk`.

## Proposal
1. Move the per-strategy parameters onto `PruningStrategy` as two helpers:
   - `batch_size(&self) -> u64` 
   - `range_end(&self, &StoredWatermark) -> u64` 

2. Simplify `create_pruning_chunks` to one batching loop that returns `Iterator<Item = (u64, u64)>` 

3. Refactor `TablePruner::prune_by_chunk` to match on `self.table.pruning_strategy()` and use the explicit range `prune_by_chunk(&self, start: u64, end: u64)`.

4. Inline `next_chunk_start` at its single call site as `end + 1`.

tomxey:
# Description of change

This PR simplifies pruner logic and also cleans up a few things:

- replaces the PruningChunk enum with `(start, end)` tuples, so create_pruning_chunks has a single batching loop instead of one per strategy.
- removes hardcoded and feature dependent constants, replaced with `--pruning-delay-ms` and `--pruning-batch-size` (and matching env vars), passed from CLI down to the Pruner.
- sets v1.29.0 as the removal release for the deprecated `--optimistic-pruner-batch-size` argument. (removal issue https://github.com/iotaledger/iota/issues/11847)

fixes #11381

### Release Notes

- [x] Indexer: added `--pruning-delay-ms` and `--pruning-batch-size` arguments (also configurable via `PRUNING_DELAY_MS` and `PRUNING_BATCH_SIZE` env vars). The already deprecated `--optimistic-pruner-batch-size` argument will be removed in v1.29.0.

### DIFF
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -319,8 +319,13 @@ pub struct PruningOptions {
     pub pruning_delay_ms: u64,
     /// Upper bound on units (checkpoints, transactions, or global sequence
     /// numbers) pruned per chunk, and on rows deleted per statement for the
-    /// `WithLimit` strategies.
-    #[arg(long, env = "PRUNING_BATCH_SIZE", default_value_t = DEFAULT_PRUNING_BATCH_SIZE)]
+    /// `WithLimit` strategies. Must be > 0.
+    #[arg(
+        long,
+        env = "PRUNING_BATCH_SIZE",
+        default_value_t = DEFAULT_PRUNING_BATCH_SIZE,
+        value_parser = clap::value_parser!(u64).range(1..),
+    )]
     pub pruning_batch_size: u64,
     /// DEPRECATED: will be removed in v1.29.0. This parameter is no longer
     /// used. Optimistic transactions are now pruned by the unified pruner.
--- a/crates/iota-indexer/src/config.rs
+++ b/crates/iota-indexer/src/config.rs
@@ -300,7 +300,10 @@ pub enum Command {
     },
 }
 
-#[derive(Args, Default, Debug, Clone)]
+pub const DEFAULT_PRUNING_DELAY_MS: u64 = 2 * 60 * 60 * 1000; // 2 hours
+pub const DEFAULT_PRUNING_BATCH_SIZE: u64 = 1000;
+
+#[derive(Args, Debug, Clone)]
 pub struct PruningOptions {
     /// DEPRECATED: will be removed in v1.28.0. Use `--pruning-config-path`
     /// pointing at a TOML retention config instead.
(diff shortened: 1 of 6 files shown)


crates/iota-indexer/src/config.rs
```diff
@@ -309,12 +312,34 @@ pub struct PruningOptions {
...
+    /// Delay in milliseconds between a watermark's lower bound being advanced
+    /// and the pruner acting on it. Lets in-flight reads complete or timeout
+    /// before their data is pruned.
+    #[arg(long, env = "PRUNING_DELAY_MS", default_value_t = DEFAULT_PRUNING_DELAY_MS)]
+    pub pruning_delay_ms: u64,
+    /// Upper bound on units (checkpoints, transactions, or global sequence
+    /// numbers) pruned per chunk, and on rows deleted per statement for the
+    /// `WithLimit` strategies.
+    #[arg(long, env = "PRUNING_BATCH_SIZE", default_value_t = DEFAULT_PRUNING_BATCH_SIZE)]
+    pub pruning_batch_size: u64,
```
kodemartin, line 329: I think it would be better to keep only one source of configuration settings with the file at `pruning_config_path`. So I suggest extending the pruning config as opposed to extending the CLI options.

What do you think?

sergiupopescu199 (approved):
lgtm 🐳

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -314,87 +300,30 @@ impl<'a> TablePruner<'a> {
...
+            self.pruning_batch_size
+        } else {
+            1
+        };
+        info!(
+            "pruning table {} ({strategy:?}) in range: [{lowest_unpruned_key}..{range_end})",
+            self.table.as_ref(),
+        );
+        (lowest_unpruned_key..range_end)
+            .step_by(batch_size as usize)
```
kodemartin, line 322: A caveat here is that `step_by` panics when `batch_size = 0`.  So if the user submits `--pruning-batch-size 0` we spawn pruning tasks for each table that panic silently.

crates/iota-indexer/src/pruning/pruner.rs
```diff
@@ -314,87 +300,30 @@ impl<'a> TablePruner<'a> {
...
+            .map(move |start| {
+                let end = (start + batch_size).min(range_end) - 1;
```
kodemartin: Underflow possibility here if `batch_size = 0`. `saturating_sub` might be plausible here, otherwise `checked_sub` can be used.

crates/iota-indexer/src/pruning/pruner.rs
kodemartin: Why do we swallow the error here? If affects watermark updating in `check_and_prune`.

kodemartin: Same here, is it ok to swallow the error?

crates/iota-indexer/src/config.rs
tomxey: The pruning config is currently only about retention policies, and that's how it's treated in the code after it's loaded. Mixing it with more technical parameters looks a bit unclean so I would prefer to leave it as is.

For other parts of the code if we would need to configure some timeout or batch size we would do so by cli or env, so that seems consistent.

crates/iota-indexer/src/pruning/pruner.rs
tomxey: rejecting 0 batch sizes: d8ca5f84a2

crates/iota-indexer/src/pruning/pruner.rs
tomxey: rejecting 0 batch sizes in: d8ca5f84a2

crates/iota-indexer/src/pruning/pruner.rs
tomxey: fixed: d8ca5f84a2

tomxey: I think it's not ok, fixed: d8ca5f84a2

kodemartin (approved):

PR #11848, opened 2026-06-11
feat(iota-{indexer,kvstore,rest-kv,data-ingestion}): support address-transaction relation

### LINKED ISSUE #11632: [kvstore,rest-kv,data-ingestion]: support address-transaction relation
## Description

We want store the relationship modeled by `tx_{senders,recipients}` in the archival store (key-value store).

* `tx_senders`: Stores the transaction senders
* `tx_recipients`: Stores the sender, the addresses that own output objects
* We could also add the transaction payers in the mix, similar to what Sui did with the `affected_addresses`.

The goal would be to support queries of the type, fetch all transactions touched by a particular address (either as a sender, payer, or recipient) in order of execution.

To do so, it seems that a key of the `<address>_<transaction_sequence_number>` would allows us to get ordered results for a given address.

The relation will be a mapping between this key and the transaction digest

```
KEY | VALUE
------|----------
<address>_<tx_sequence_number> | <digest>
```

This can reuse the changes implemented for #11594

> [!NOTE]
>
> Assuming that most recent transactions are more valuable than order we can adopt the following optimization:
>
> Instead of storing `tx_sequence_number` which implies an ascending order of transactions (from older to newer), we store its `u64` complement in BigTable, to promote descending order (from newer to older) queries as forward scans.
>
> From [BigTable docs](https://docs.cloud.google.com/bigtable/docs/reads?authuser=6):
> > Reverse scans are less efficient than forward scans. In general, design your row keys so that most scans are forward. Use reverse scans for short scans, such as 50 rows or less, to maintain low-latency response time.

sergiupopescu199:
# Description of change

This PR adds support for efficient address-to-transaction lookups across the data pipeline. Clients can now query all transactions involving an `IotaAddress` (as `sender`, `recipient`, or `gas payer`), including historical data that has been pruned from the indexer's relational store.

- **`iota-kvstore`**: Refactored the `KvWorker` to write the address-transaction relation during checkpoint ingestion. The worker now accepts an optional list of tables at startup, so a backfill instance can populate only a subset.

- **`iota-data-ingestion`**: The `BigTableKV` task type now accepts an optional `tables` list from `config.yaml`, enabling targeted backfills. To bound a backfill to a checkpoint range, the `IndexerConfig` accepts two new optional parameters: `start-checkpoint` and `end-checkpoint` (both inclusive).

- **`iota-rest-kv`**: Adds a `GET /txa/{address}` endpoint. Unlike existing endpoints, which are 1:1 key-value lookups, this one exposes a one-to-many relation. It returns an ordered list of transaction digests with cursor-based pagination (defaulting to descending order) to handle addresses with long histories.

- **`iota-indexer`**: Adds historical fallback support to the JSON-RPC API for the `TransactionFilter{V2}::FromOrToAddress` filter. When the indexer's relational store has pruned the requested data, the query is served from the historical KV store.

fixes #11632

Each commit corresponds to a previously-reviewed sub-PR. Test coverage (unit-tests or manual) is documented in the individual commit messages.

> [!NOTE]

### DIFF
--- a/crates/iota-data-ingestion/src/main.rs
+++ b/crates/iota-data-ingestion/src/main.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{env, path::PathBuf, time::Duration};
+use std::{collections::BTreeSet, env, path::PathBuf, time::Duration};
 
 use anyhow::{Result, anyhow};
 use iota_data_ingestion::{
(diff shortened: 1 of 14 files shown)


kodemartin (approved):
🎈

tomxey (approved):

jkrvivian (approved):

PR #11858, opened 2026-06-11
fix(starfish): bound peer-controlled consensus inputs

### LINKED ISSUE #11857: Harden Starfish peer-controlled request and response bounds
## Context

A Starfish security review identified several related paths where an authenticated consensus peer can make a node deserialize, accumulate, encode, or verify more data than the protocol operation requires.

The affected findings are:

- **M2 / L4:** transaction bundles can reach Reed-Solomon encoding before the serialized transaction payload is bounded, and rejected transaction payloads are not consistently recorded as peer or author misbehavior. The same unbounded-encode shape exists on commit-sync and transaction-sync fetch responses.
- **M4:** commit sync limits the number of commits but not the number of certifier block headers verified for the response.
- **L1:** fetch-header block references are deserialized and validated before the request is truncated to the configured service limit.
- **L3:** fetch-commits accepts inverted commit ranges, underflowing the subsequent range arithmetic in debug builds.
- **I3:** latest-header synchronization can accumulate and later verify more headers than were requested.
- Commit certifier votes are read across all commit digests at an index, so a fabricated vote digest counts toward the server-side certifiability decision; combined with minimal-subset serving it can displace real voters from the served certifier set.

## Required changes

- Bound the serialized transaction envelope using protocol transaction byte and count limits before Reed-Solomon encoding, on the block-bundle path and on commit-sync / transaction-sync fetch responses.
- Record malformed, oversized, commitment-mismatched, and invalid transaction payloads through the existing invalid-bundle metrics and misbehavior store.
- Reject commit-sync responses containing excessive certifier headers before header parsing and signature verification, allowing bounded per-authority multiplicity while un-upgraded servers serve undeduplicated votes.
- Filter certifier votes by the digest of the locally stored commit, return at most one certifier header per authority, and stop once quorum is represented.
- Reject inverted commit ranges before range arithmetic.
- Truncate fetch-header requests before block-reference deserialization and validation.
- Reject latest-header streams once the cumulative response exceeds the number of requested authorities.
- Add focused regression coverage for each boundary.

## Non-goals

This issue does not address broader server admission-control findings such as concurrent HTTP/2 stream limits, request deadlines, or TLS handshake limits.

polinikita:
# Description of change

Bounds peer-controlled inputs on the Starfish consensus request/response paths before they drive heavy work, and fixes commit certification to count only votes for the stored commit digest.

**Transaction payloads**

- Adds `BlockVerifier::check_and_parse_transactions`, which bounds a serialized transaction payload and parses it before any expensive processing. The limit is derived from the protocol transaction byte/count limits plus BCS length-prefix overhead and lives next to `check_transactions`, whose limits define it; a test pins the BCS wire layout the bound assumes. A zero protocol limit disables the bound, matching the block verifier convention.
- Applies the bound on the block-bundle path before Reed-Solomon encoding and Merkle construction, and on commit-sync / transaction-sync fetch responses before erasure-encoding the fetched payload.
- Records malformed, oversized, and commitment-mismatched bundle payloads in `bundles_with_invalid_parts` and the misbehavior store.

**Commit sync**

- Serves a minimal certifier set: votes are filtered by the digest of the locally stored commit, deduplicated by author, and selection stops at quorum stake. `Store::read_commit_votes` now reads a single `(index, digest)` keyspace, so fabricated vote digests can neither inflate the server-side quorum decision nor displace real voters from the served set.
- Caps fetched commit vote headers at twice the committee size before parsing and signature verification; the multiplicity accepts responses from servers that do not dedup votes by author, and can be tightened to one per authority once all networks serve deduplicated votes.
- Rejects inverted commit ranges in `fetch_commits` before range arithmetic.
- Validates every `AuthorityIndex` carried by a fetched commit (leader, header refs, transaction refs, reputation scores) against the committee in `verify_commits`, with peer attribution, before the content is indexed into per-authority state downstream.

**Header sync**

- Truncates fetch-header requests to the configured fetch size before block-reference deserialization and validation, with the size selection shared via `Parameters::max_headers_per_fetch`.
- Rejects latest-header responses that return more headers than the number of requested authorities.

**Parameter renames (operator-facing)**

`max_headers_per_regular_sync_fetch` → `max_headers_per_header_sync_fetch` and `max_transactions_per_regular_sync_fetch` → `max_transactions_per_transaction_sync_fetch`: the old names suggested regular *commit* sync, but these bound the header and transaction synchronizers. The commit-sync parameters keep their unqualified names since both regular and fast commit sync use them. Node configs setting the old keys fall back silently to the defaults.

No protocol-version or wire-format change. During a rolling upgrade, responses from un-upgraded servers stay within the new client-side caps for honest vote multiplicities.

fixes #11857

### DIFF
--- a/crates/starfish/core/src/network/tonic_network.rs
+++ b/crates/starfish/core/src/network/tonic_network.rs
@@ -57,6 +57,21 @@ const MAX_FETCH_RESPONSE_BYTES: usize = 4 * 1024 * 1024;
 // the responses.
 const MAX_TOTAL_FETCHED_BYTES: usize = 128 * 1024 * 1024;
 
+fn validate_header_response_count(
+    peer: AuthorityIndex,
+    requested: usize,
+    received: usize,
+) -> ConsensusResult<()> {
+    if received > requested {
+        return Err(ConsensusError::UnexpectedNumberOfHeadersFetched {
+            authority: peer,
+            requested,
+            received_headers: received,
+        });
+    }
+    Ok(())
+}
+
 // Implements Tonic RPC client for Consensus.
 pub(crate) struct TonicClient {
     context: Arc<Context>,
@@ -257,10 +272,15 @@ impl NetworkClient for TonicClient {
             .into_inner();
         let mut blocks = vec![];
         let mut total_fetched_bytes = 0;
+        let max_headers = authorities.len();
         loop {
             match stream.message().await {
                 Ok(Some(response)) => {
                     let vec_serialized_block_headers = response.vec_serialized_block_header;
+                    let received_headers = blocks
+                        .len()
+                        .saturating_add(vec_serialized_block_headers.len());
+                    validate_header_response_count(peer, max_headers, received_headers)?;
                     for b in &vec_serialized_block_headers {
                         total_fetched_bytes += b.len();
                     }
@@ -646,9 +666,24 @@ impl<S: NetworkService> ConsensusService for TonicServiceProxy<S> {
             return Err(tonic::Status::internal("PeerInfo not found"));
         };
         let inner = request.into_inner();
+        let highest_accepted_rounds = inner.highest_accepted_rounds;
+        let max_fetch_size = if highest_accepted_rounds.is_empty() {
+            self.context.parameters.max_headers_per_commit_sync_fetch
+        } else {
+            self.context.parameters.max_headers_per_regular_sync_fetch
+        };
+        if inner.block_refs.len() > max_fetch_size {
+            warn!(
+                "Truncated fetch headers request from {} to {} blocks for peer {}",
+                inner.block_refs.len(),
+                max_fetch_size,
+                peer_index
+            );
+        }
         let block_refs = inner
             .block_refs
             .into_iter()
+            .take(max_fetch_size)
             .filter_map(|serialized| match bcs::from_bytes(&serialized) {
                 Ok(r) => Some(r),
                 Err(e) => {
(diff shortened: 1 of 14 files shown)


piotrm50 (approved):

crates/starfish/core/src/network/tonic_network.rs
```diff
@@ -646,9 +657,23 @@ impl<S: NetworkService> ConsensusService for TonicServiceProxy<S> {
             return Err(tonic::Status::internal("PeerInfo not found"));
         };
         let inner = request.into_inner();
+        let highest_accepted_rounds = inner.highest_accepted_rounds;
+        let max_fetch_size = self
+            .context
+            .parameters
+            .max_headers_per_fetch(highest_accepted_rounds.is_empty());
```
VorobyevIlya, line 664: Do we again use this check to understand what type of request we have here? Maybe we should make it clearer

crates/starfish/core/src/network/tonic_network.rs
```diff
@@ -646,9 +657,23 @@ impl<S: NetworkService> ConsensusService for TonicServiceProxy<S> {
...
+        if inner.block_refs.len() > max_fetch_size {
+            warn!(
+                "Truncated fetch headers request from {} to {} blocks for peer {}",
+                inner.block_refs.len(),
+                max_fetch_size,
+                peer_index
+            );
```
VorobyevIlya, line 671: And then we can output what type of request it was

VorobyevIlya (approved):

PR #11865, opened 2026-06-11
fix(typed-store): make iterator bounds consistent and add prefix scans

### LINKED ISSUE #10929: typed-store: issues with typed-store iterator
typed-store implements a wrapper for the rocksdb backend iterator. This wrapper has a number of issues as outlined [here](https://github.com/iotaledger/iota/pull/10780#pullrequestreview-3999778557):
- safe_iter_with_bounds and reversed_safe_iter_with_bounds are not consistent with the key ranges they produce when both bounds are provided;
- in rocksdb the upper bound is always excluded, but the wrapping logic incorrectly handles some edge cases (eg. iterator_bounds_with_range, big_endian_saturating_add_one);
- rocksdb keys are just [u8] ordered lexicographically (big-endian), to convert to [u8] user code serializes the key values; however, the user code needs to use fixed-sized (eg. u64, [u8; 32]) and variable-sized (eg. String) keys and combine those keys into tuples to construct composite keys. Settings upper bounds for such composite keys is not trivial.

chriscandless:
# Description of change

Make the typed-store iterator wrapper produce consistent key ranges in both directions and remove the need for hand-built maximum keys in composite-key scans.

`safe_iter_with_bounds(lo, hi)` and `reversed_safe_iter_with_bounds(lo, hi)` computed their raw rocksdb bounds independently: the forward iterator covered `[lo, hi)` while the reversed one covered `[lo, hi]`, so reversing a scan did not yield the same key set. Separately, scanning "all entries under a leading key" of a composite key required constructing artificial maximum keys (`ObjectKey::max_for_id`, `usize::MAX` tuple tails, etc.).

Those issues are solved with two new sets of methods:
- Added `safe_range_iter_reversed` - reversed version of pre-existing `safe_range_iter`. It is a better version of `reversed_safe_iter_with_bounds`, because it **explicitly** defines type of bounds (included, excluded). Also `safe_range_iter` and `safe_range_iter_reversed` are now using same underlying logic for bound calculation, so it is impossible that they ever become asymmetric. 
- Added `safe_iter_with_prefix` / `safe_iter_with_prefix_reversed` - enable clear iteration by prefix, which is what magical `MAX` values were mostly used for. This removes most current and future uses of `MAX` values. A `safe_iter_with_prefix_from` variant covers cursor-paginated prefix scans (start from a cursor, no `MAX` upper bound).

The change is split into five self-contained commits (the latter four are independently revertable):

1. **`fix(typed-store): unify iterator bound handling and add prefix scans`** — route all bound computation through one range engine (`iterator_bounds_with_range`) and add `safe_range_iter_reversed`, which by construction yields exactly the keys of `safe_range_iter` in descending order. `SafeRevIter` now takes an exclusive upper boundary and steps off the boundary key when the initial seek lands on it, so exclusive upper bounds work in reverse as well. Add `safe_iter_with_prefix` / `safe_iter_with_prefix_reversed`, which derive the upper bound by incrementing the serialized prefix bytes, and migrate `AuthorityStore::get_latest_marker` as the first consumer. The Option-based wrappers keep their existing documented semantics as thin presets over the range engine.

2. **`refactor(typed-store): remove reversed_safe_iter_with_bounds in favor of safe_range_iter_reversed`** — the Option-based reverse iterator hard-coded inclusive bounds, invisible at the call site. Migrate all callers to the range form, where inclusivity is spelled out by the range syntax, and drop the wrapper together with its vestigial `Result`. The translation is semantics-preserving:

   | old | new |
   |---|---|
   | `(None, None)` | `..` |
   | `(None, Some(hi))` | `..=hi` |
   | `(Some(lo), Some(hi))` | `lo..=hi` |

   Note for upstream ports: Sui still has `reversed_safe_iter_with_bounds`; ported code touching these call sites should apply the mapping above.

3. **`refactor(iota-core): use prefix iterators instead of artificial MAX keys`** — replace the pure prefix scans (`get_events_by_events_digest`, `have_deleted_owned_object_at_version_or_after`, `get_latest_live_version_for_object_id`, `get_latest_object_ref_or_tombstone`, `get_latest_object_or_tombstone`, `try_get_object`, and the `#[cfg(msim)]` `count_object_versions`) with prefix iterators. Sites where a bound carries a real value (cursors, version cutoffs, `range_iter_live_object_set` spanning multiple ids) keep the explicit range form. Two behavior deltas, both safe: (a) a prefix scan also tightens the *lower* bound (from unbounded / `min_for_id` to the prefix start) — safe because every migrated reader re-checks the leading key (epoch / object id) of the first row, so a narrower lower bound cannot change the result; (b) the replaced `..=max_for_id` upper bounds capped the version component at `MAX_VALID_EXCL` while a prefix scan covers the full `u64` space — equivalent because sentinel versions (`CANCELLED_READ`, `CONGESTED_*`, `MAX_VALID_EXCL`) are scheduling-time artifacts, never persisted as object or marker table keys (a write-path invariant; verified by inspecting the write paths, not unit-tested, since a read-side test could only re-demonstrate the inclusion rather than prove the invariant).

4. **`refactor(typed-store,iota-core): add safe_iter_with_prefix_from and drop MAX bounds in cursor scans`** — add `safe_iter_with_prefix_from(&prefix, &cursor)` (the rocksdb `prefix_iterator_from` idiom: lower bound `prefix ++ cursor`, upper bound derived by incrementing the serialized prefix bytes). `cursor` is the remainder of the key after the prefix, **not** a full key, so the prefix is never duplicated and the lower bound is a prefix of the key by construction. Migrates the cursor-paginated prefix scans (`get_dynamic_fields_iterator`, `dynamic_field_iter`, `package_versions_iter`) off their `(prefix, MAX)` upper bounds; those callers keep their `.skip(cursor.is_some())` (the cursor stays the inclusive lower bound). `get_newer_object_keys` is migrated too but has no cursor — its lower bound is `object.1.next()` (version + 1, an exclusive-by-construction start) and it uses no `.skip`. The prefix scan covers the whole prefix, including a key sitting exactly at the type's maximum tail that the old exclusive `(prefix, MAX)` bound dropped — never a real version/marker, a corrected edge for dynamic-field ids, equivalent for all real data. Reversed-pagination scans and the hash-bounded `owner_iter` keep explicit ranges — their bounds are not pure prefixes.

5. **`fix(typed-store): surface deserialization errors in SafeIter instead of ending iteration`** — `SafeIter::next` mapped a failed key/value deserialization to `None` via `.ok()`, indistinguishable from end-of-data, so one undeserializable row silently truncated the rest of the scan (the item type is `Result<(K, V), _>` precisely so the error can be reported; upstream returns it, the IOTA port had regressed to swallowing it with no test on the path). Now returns `Some(Err(TypedStoreError::Serialization(..)))`, matching how the in-memory backend (`memstore`) and `DBMap::get`/`multi_get` already handle deserialization failures, with a regression test that reopens a column family with a wider value type and asserts iteration yields `Some(Err(..))`. Found in review of this PR.

The forward `safe_iter_with_bounds` is intentionally kept: its `[lo, hi)` contract matches the conventional half-open semantics, it is part of the `Map` trait shared with upstream, and it is now implemented as a thin preset over `safe_range_iter` with the contract documented on the trait.

Intentional runtime-behavior changes (no consumer relies on the old behavior):
- Prefix scans include an entry at the type's maximum tail that the old exclusive `(prefix, MAX)` bound excluded (see commits 3-4); only matters for a key exactly at that tail, which the migrated tables never hold.
- `SafeIter::next` now surfaces a deserialization failure as `Some(Err(..))` instead of mapping it to `None` (silently truncating the scan); every iterator consumer is affected — `?`/collect propagate the error, `.next()`-style readers return it. No consumer was relying on silent skip-on-bad-row.

Edge-case note (issue point 2): the inclusive upper-bound arm of `iterator_bounds_with_range` returned `None` (unbounded) when the bound serialized to all-`0xFF`. That is fine for fixed-width keys, but a longer key extending the maximum sorts after it and was wrongly included in a `..=max` range. This PR returns `ser(max) ++ [0]` instead (the first key past the maximum), mirroring the excluded-lower arm, with a regression test. On top of that it adds reverse-direction and prefix regression coverage (forward/reverse symmetry across all bound kinds, prefix scans at the `0xFF` serialization limit, boundary keys present in the DB).

### Remaining `MAX` usages (deliberately not removed)

A few iterator bounds still build `MAX`/`MIN` keys by hand. They are left as-is because they are not pure forward prefix scans, so the new prefix APIs do not fit them cleanly; each would be a separate, higher-risk change:

- **Reversed pagination without a prefix** (`get_transactions` reversed over `transaction_order`, keyed by a bare `TxSequenceNumber`): `..=cursor.unwrap_or(MAX)` — there is no composite prefix here, `MAX` just means "no cursor → start from the end of the table". Removing it means branching on `cursor` (`..=c` vs `..`), pure cosmetics over a non-fragile sentinel.
- **Reversed prefix pagination** (`get_transactions_from_index`, `get_event_from_index`, `events_by_transaction`, `all_events`): these scan a leading key downwards from a cursor. They would fit a *reversed* `safe_iter_with_prefix_from`, but that is a new method with a different shape (in reverse the cursor is the upper bound and the prefix is the lower), and these sites currently stop via a `take_while`/predicate rather than a DB-level lower bound — switching the cutoff mechanism is a behavioral change that warrants its own PR with tests.
- **`owner_iter`** (gRPC): the `OwnerIndexKey` prefix length is filter-dependent (`owner`, or `(owner, id_hash)`, or `(owner, id_hash, params_hash)`), and the `MAX` values span the remaining unfixed fields. Expressible via prefix scans, but only by reworking `owner_bounds` per filter branch — not a mechanical swap.
- **`range_iter_live_object_set`**: `min_for_id(id_A)..=max_for_id(id_B)` spans a range of *different* object ids (table sharding for background scans). The heads differ, so there is no common prefix; the alternative `..(next(id_B), 0)` just moves the synthetic key elsewhere.
- **Range delete / compaction** (`schedule_delete_range` for events, `compact_range` for objects) and config sentinels (`num_epochs_to_retain == u64::MAX` meaning "retain forever"): these are not iterators at all — a "delete/compact by prefix" helper would be a separate API surface, and the config value is not a key.

Net: this PR removes the `MAX` keys that were standing in for a missing forward prefix API. The reversed-prefix cases are the natural next step (a reversed `prefix_from`) and are left for a follow-up.

fixes #10929

`cargo check` / `cargo clippy --all-targets --all-features -- -D warnings` are green for `typed-store`, `iota-core`, `iota-transactional-test-runner`, and `starfish-core`; the `typed-store` unit suite passes (48/48, including the new symmetry/prefix regression tests). Each commit builds and passes the typed-store suite on its own. Note: running the typed-store suite with parallel `cargo test` hits a pre-existing Prometheus `AlreadyReg` registry race (also fails on `develop`); the suite is green under `nextest` (process per test) and with `--test-threads=1`.

### DIFF
--- a/crates/typed-store/src/database.rs
+++ b/crates/typed-store/src/database.rs
@@ -772,6 +772,31 @@ impl<K, V> DBMap<K, V> {
         self.iter_forward_raw(lower_bound, upper_bound)
     }
 
+    /// Forward iterator over entries whose key begins with `prefix`, starting
+    /// at `prefix ++ cursor` rather than at the first key under the prefix.
+    ///
+    /// `cursor` is **not** a full key — it is the remainder of the key that
+    /// follows `prefix` (typically the trailing field(s) of a tuple key). Both
+    /// `prefix` and `cursor` are serialized with `be_fix_int_ser` and
+    /// concatenated, mirroring how a composite key is encoded; the scan then
+    /// covers `[prefix ++ cursor, end-of-prefix)`. This lets a paginated scan
+    /// resume from a cursor without an artificial maximum key for the upper
+    /// bound.
+    pub fn safe_iter_with_prefix_from<P, C>(&self, prefix: &P, cursor: &C) -> DbIterator<'_, (K, V)>
+    where
+        P: ?Sized + Serialize,
+        C: ?Sized + Serialize,
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let (lower_bound, upper_bound) = prefix_iterator_bounds(prefix);
+        let lower_bound = lower_bound.map(|mut lower| {
+            lower.extend_from_slice(&be_fix_int_ser(cursor));
+            lower
+        });
+        self.iter_forward_raw(lower_bound, upper_bound)
+    }
+
     /// Reverse counterpart of [`Self::safe_iter_with_prefix`]: the matching
     /// entries in descending key order (e.g. for "latest entry under a
     /// prefix").
--- a/crates/typed-store/src/database.rs
+++ b/crates/typed-store/src/database.rs
@@ -784,28 +784,6 @@ impl<K, V> DBMap<K, V> {
         let (lower_bound, upper_bound) = prefix_iterator_bounds(prefix);
         self.iter_reversed_raw(lower_bound, upper_bound)
     }
-
-    /// Creates a reversed iterator with optional inclusive bounds.
-    ///
-    /// Both bounds are inclusive: `(Some(lo), Some(hi))` yields `[lo, hi]` in
-    /// descending order. This is a thin wrapper over
-    /// [`Self::safe_range_iter_reversed`] with `lo..=hi`; use the range method
-    /// directly when you need exclusive bounds.
-    pub fn reversed_safe_iter_with_bounds(
-        &self,
-        lower_bound: Option<K>,
-        upper_bound: Option<K>,
-    ) -> Result<DbIterator<'_, (K, V)>, TypedStoreError>
-    where
-        K: Serialize + DeserializeOwned,
-        V: Serialize + DeserializeOwned,
-    {
-        let range = (
-            lower_bound.map(Bound::Included).unwrap_or(Bound::Unbounded),
-            upper_bound.map(Bound::Included).unwrap_or(Bound::Unbounded),
-        );
-        Ok(self.safe_range_iter_reversed(range))
-    }
 }
 
 /// Provides a mutable struct to form a collection of database write operations,
@@ -1218,7 +1196,7 @@ where
     fn schedule_delete_all(&self) -> Result<(), TypedStoreError> {
         let first_key = self.safe_iter().next().transpose()?.map(|(k, _v)| k);
         let last_key = self
-            .reversed_safe_iter_with_bounds(None, None)?
+            .safe_range_iter_reversed(..)
             .next()
             .transpose()?
             .map(|(k, _v)| k);
--- a/crates/typed-store/src/database.rs
+++ b/crates/typed-store/src/database.rs
@@ -32,7 +32,7 @@ use crate::{
         safe_iter::{SafeIter, SafeRevIter},
     },
     traits::{Map, TableSummary},
-    util::{be_fix_int_ser, iterator_bounds, iterator_bounds_with_range},
+    util::{be_fix_int_ser, iterator_bounds_with_range, prefix_iterator_bounds},
 };
 
 #[derive(Clone)]
@@ -665,9 +665,132 @@ impl<K, V> DBMap<K, V> {
         )
     }
 
-    /// Creates a safe reversed iterator with optional bounds.
-    /// Both upper bound and lower bound are included.
-    #[allow(clippy::complexity)]
+    /// Shared rocksdb `SafeIter` construction (raw iterator + scan metrics)
+    /// used by both the forward and reverse raw iterators.
+    fn rocks_safe_iter<'a>(&self, db: &'a RocksDB, readopts: ReadOptions) -> SafeIter<'a, K, V>
+    where
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let db_iter = db
+            .underlying
+            .raw_iterator_cf_opt(&rocks_cf(db, self.column_family.name()), readopts);
+        let (_timer, bytes_scanned, keys_scanned, _perf_ctx) = self.create_iter_context();
+        SafeIter::new(
+            self.cf_name().to_string(),
+            db_iter,
+            _timer,
+            _perf_ctx,
+            bytes_scanned,
+            keys_scanned,
+            Some(self.db_metrics.clone()),
+        )
+    }
+
+    /// Forward iterator over the raw byte bounds `[lower_bound, upper_bound)`;
+    /// both bounds (already serialized, `upper_bound` exclusive) are applied to
+    /// the read options.
+    fn iter_forward_raw(
+        &self,
+        lower_bound: Option<Vec<u8>>,
+        upper_bound: Option<Vec<u8>>,
+    ) -> DbIterator<'_, (K, V)>
+    where
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        match &self.db.storage {
+            Storage::Rocks(db) => {
+                let readopts =
+                    rocks_util::apply_range_bounds(self.opts.readopts(), lower_bound, upper_bound);
+                Box::new(self.rocks_safe_iter(db, readopts))
+            }
+            Storage::InMemory(db) => {
+                db.iterator(self.column_family.name(), lower_bound, upper_bound, false)
+            }
+        }
+    }
+
+    /// Reverse counterpart of [`Self::iter_forward_raw`], yielding the same
+    /// keys in descending order. Only the lower bound is applied to the
+    /// read options; the (exclusive) `upper_bound` is enforced by
+    /// [`SafeRevIter`]'s initial seek, because setting an
+    /// iterate-upper-bound would stop `seek_for_prev` from landing on the
+    /// boundary key.
+    fn iter_reversed_raw(
+        &self,
+        lower_bound: Option<Vec<u8>>,
+        upper_bound: Option<Vec<u8>>,
+    ) -> DbIterator<'_, (K, V)>
+    where
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        match &self.db.storage {
+            Storage::Rocks(db) => {
+                let readopts =
+                    rocks_util::apply_range_bounds(self.opts.readopts(), lower_bound, None);
+                Box::new(SafeRevIter::new(
+                    self.rocks_safe_iter(db, readopts),
+                    upper_bound,
+                ))
+            }
+            Storage::InMemory(db) => {
+                db.iterator(self.column_family.name(), lower_bound, upper_bound, true)
+            }
+        }
+    }
+
+    /// Reverse counterpart of [`Map::safe_range_iter`]: yields exactly the keys
+    /// of `safe_range_iter(range)` in descending order.
+    ///
+    /// Both directions derive their bounds from the same
+    /// `iterator_bounds_with_range`, so they are guaranteed to cover the
+    /// identical set of keys regardless of the bound inclusivity.
+    pub fn safe_range_iter_reversed(&self, range: impl RangeBounds<K>) -> DbIterator<'_, (K, V)>
+    where
+        K: Serialize + DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let (lower_bound, upper_bound) = iterator_bounds_with_range(range);
+        self.iter_reversed_raw(lower_bound, upper_bound)
+    }
+
+    /// Forward iterator over every entry whose key begins with `prefix`.
+    ///
+    /// `prefix` is serialized with `be_fix_int_ser` and must form a prefix of
+    /// the column family's key encoding — typically the leading field(s) of a
+    /// tuple key. This avoids constructing artificial maximum keys to bound a
+    /// composite-key scan.
+    pub fn safe_iter_with_prefix<P>(&self, prefix: &P) -> DbIterator<'_, (K, V)>
+    where
+        P: ?Sized + Serialize,
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let (lower_bound, upper_bound) = prefix_iterator_bounds(prefix);
+        self.iter_forward_raw(lower_bound, upper_bound)
+    }
+
+    /// Reverse counterpart of [`Self::safe_iter_with_prefix`]: the matching
+    /// entries in descending key order (e.g. for "latest entry under a
+    /// prefix").
+    pub fn safe_iter_with_prefix_reversed<P>(&self, prefix: &P) -> DbIterator<'_, (K, V)>
+    where
+        P: ?Sized + Serialize,
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        let (lower_bound, upper_bound) = prefix_iterator_bounds(prefix);
+        self.iter_reversed_raw(lower_bound, upper_bound)
+    }
+
+    /// Creates a reversed iterator with optional inclusive bounds.
+    ///
+    /// Both bounds are inclusive: `(Some(lo), Some(hi))` yields `[lo, hi]` in
+    /// descending order. This is a thin wrapper over
+    /// [`Self::safe_range_iter_reversed`] with `lo..=hi`; use the range method
+    /// directly when you need exclusive bounds.
     pub fn reversed_safe_iter_with_bounds(
         &self,
         lower_bound: Option<K>,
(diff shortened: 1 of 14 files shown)


crates/typed-store/src/database.rs
```diff
@@ -708,16 +735,57 @@ impl<K, V> DBMap<K, V> {
...
+        self.iter_reversed_raw(lower_bound, upper_bound)
+    }
+
+    /// Forward iterator over every entry whose key begins with `prefix`.
+    ///
+    /// `prefix` is serialized with `be_fix_int_ser` and must form a prefix of
+    /// the column family's key encoding — typically the leading field(s) of a
+    /// tuple key. This avoids constructing artificial maximum keys to bound a
+    /// composite-key scan.
+    pub fn safe_iter_with_prefix<P>(&self, prefix: &P) -> DbIterator<'_, (K, V)>
```
bingyanglin, line 765: `safe_iter_with_prefix` only works when the prefix serializes to a clean byte-prefix of the key, and a variable-length leading field like a `String` would silently return wrong results. The doc comment warns about it, but nothing stops it. Maybe add a `debug_assert` or a marker trait so a bad prefix type fails loudly instead of quietly scanning the wrong range?

crates/typed-store/src/rocks/safe_iter.rs
bingyanglin: If a key or value fails to deserialize here, `next()` returns `None` and the iterator just looks finished, so one bad row silently cuts off the rest of the scan. Not new in this PR, but this path now carries more of the iteration. Maybe return `Some(Err(..))` instead of swallowing it?

crates/typed-store/src/rocks/tests.rs
```diff
@@ -603,6 +630,145 @@ async fn test_safe_range_iter_exclusive_lower_bound_at_max_with_inclusive_upper(
...
+    check((Unbounded, Included(20))); // ..=20
+    check((Included(50), Unbounded)); // 50..
+    check((Excluded(50), Unbounded)); // (50, ..)
+    check((Unbounded, Unbounded)); // full table
+    check((Included(40), Excluded(40))); // empty
+    check((Included(50), Included(50))); // single element
+}
+
+#[tokio::test]
+async fn test_safe_iter_with_prefix() {
```
bingyanglin, line 642: The prefix tests all use single-field keys, which is the easy case. Maybe add one with a 3-tuple key where the prefix is the first two fields, and assert `safe_iter_with_prefix` matches `safe_range_iter(lo..hi)`. That's the case most likely to break if the byte-prefix assumption ever slips.

crates/typed-store/src/rocks/safe_iter.rs
chriscandless: That's a great find! 100% worth fixing. Fixed

crates/typed-store/src/database.rs
chriscandless: Yeah, I see the issue. But I don't think this PR is good place for this change:

- All pre-existing functions have this same flaw, and for now it is not even a flaw yet, because existing cases are already accounting for this issue: either just ignore it because they query an exact leading key, or use a fixed-length representation (hashes as keys in gRPC owner/coin).
- Adding trait would require significant amount of changes, and all of them unrelated to the original issue.
- A simple `debug_assert` won't catch the variable-length case: from the serialized bytes of a single value you can't distinguish a length prefix from a fixed-width field. The only cheap runtime check is comparing the prefix against an actual key, which nobody has.
- Detecting it properly would need a custom serializer that inspects the type's structure - doable, but IMHO that's too much for this PR.

If you want this improvement for future use cases, I believe we better do it in a separate PR.

piotrm50 (approved):

chriscandless:
I found important bug - I've accidentally changed values of cursors from "dynamic remainder" to "entire key". I now fixing it.

crates/typed-store/src/rocks/tests.rs
chriscandless: Good point. Added (and even more).

crates/typed-store/src/database.rs
```diff
@@ -665,36 +665,63 @@ impl<K, V> DBMap<K, V> {
...
+    }
+
+    /// Reverse iterator over the raw byte bounds `[lower_bound, upper_bound)`,
+    /// yielding keys in descending order.
+    ///
+    /// `upper_bound` is the exclusive upper boundary (first key outside the
+    /// range). For rocksdb only the lower bound is applied to the read options;
+    /// the upper boundary is enforced by the iterator's initial seek (see
+    /// [`SafeRevIter::new`]).
+    fn iter_reversed_raw(
```
semenov-vladyslav, line 720: I would factor this function and `iter_forward_raw` into a common one since they have a lot in common. That way it would be easier to see the difference (and asymmetry).

crates/typed-store/src/database.rs
```diff
@@ -665,36 +665,63 @@ impl<K, V> DBMap<K, V> {
...
+        upper_bound: Option<Vec<u8>>,
+    ) -> DbIterator<'_, (K, V)>
+    where
+        K: DeserializeOwned,
+        V: DeserializeOwned,
+    {
+        match &self.db.storage {
+            Storage::Rocks(db) => {
+                let readopts =
+                    rocks_util::apply_range_bounds(self.opts.readopts(), lower_bound, None);
```
semenov-vladyslav, line 732: I don't quite understand why not set the upper bound here too (it is set in `iter_forward_raw`)? That way rocksdb would know that the range is bounded from both sides.

crates/typed-store/src/util.rs
semenov-vladyslav: Why is this value `None`? Shouldn't there be `key_buf.push(0);` similar to excluded lower branch case? With `None` there is no upper bound, thus iterator would incorrectly include keys that are out of bounds. It seems that this branch should have not been triggered with the standard rust types that we usually use as keys in the repo. However, I was able to construct a rust type and a value of that type that would trigger this branch.

semenov-vladyslav:
```suggestion
```
Let's remove this check. In all the code paths `is_max` is being checked already, doing it here is useless.

crates/typed-store/src/rocks/safe_iter.rs
```diff
@@ -120,12 +124,22 @@ pub struct SafeRevIter<'a, K, V> {
...
+        match exclusive_upper_bound {
             None => iter.db_iter.seek_to_last(),
-            Some(key) => iter.db_iter.seek_for_prev(&key),
+            Some(bound) => {
+                iter.db_iter.seek_for_prev(&bound);
+                // The bound is exclusive: move to the previous key if the
+                // excluded key actually exists in the db.
+                if iter.db_iter.valid() && iter.db_iter.key() == Some(bound.as_slice()) {
+                    iter.db_iter.prev();
+                }
```
semenov-vladyslav, line 141: Shouldn't `seek_for_prev` already return an entry with the key strictly less that `bound`?

crates/typed-store/src/database.rs
chriscandless: Valid point. I have read before, that in Rust people usually prefer separate functions over flag arguments. But I see in this case it is not important because it not an interface of the package, but internal helper. 
So I tried what you propose and created single function. It is fine, but IMHO it reads worse than two separate functions, because it required a lot of branching.
And as readability is everything, I settled on a compromise 😄 - two functions with shared code moved into third function. Please check it out and tell me if that look good to you.

Below is that "single" function I was talking about (which I dropped):
```
fn iter_raw(
    &self,
    lower_bound: Option<Vec<u8>>,
    upper_bound: Option<Vec<u8>>,
    direction: Direction,
) -> DbIterator<'_, (K, V)>
where
    K: DeserializeOwned,
    V: DeserializeOwned,
{
    let reverse = matches!(direction, Direction::Reverse);
    match &self.db.storage {
        Storage::Rocks(db) => {
            // Forward: the upper bound goes to the read options. Reverse: it
            // is the seek target for SafeRevIter instead.
            let (readopts_upper, seek_upper) = if reverse {
                (None, upper_bound)
            } else {
                (upper_bound, None)
            };
            let readopts = rocks_util::apply_range_bounds(
                self.opts.readopts(),
                lower_bound,
                readopts_upper,
            );
            let db_iter = db
                .underlying
                .raw_iterator_cf_opt(&rocks_cf(db, self.column_family.name()), readopts);
                keys_scanned,
                Some(self.db_metrics.clone()),
            );
            if reverse {
                Box::new(SafeRevIter::new(iter, seek_upper))
            } else {
                Box::new(iter)
            }
        }
        Storage::InMemory(db) => {
            db.iterator(self.column_family.name(), lower_bound, upper_bound, reverse)
        }
    }
}
```

crates/typed-store/src/database.rs
chriscandless: Both ways are equivalent from the point of view of logic:
- current variant uses `seek_for_prev + `prev` to land on the first included element from the top
- the same effect can be achieved by using readopts (`iterate_upper_bound`) + `seek_to_last`

There are two reasons it is done the current way:
- semantical reason: readops is a regular way to achieve getting "invalid" element when reaching that position. There is no need in setting it in reverse func if we only descend from it.
- historical reason: `iterate_upper_bound` is designed for forward iteration, and its correctness for reverse seek (`seek_for_prev`) wasn't guaranteed - the seek could still land on a key at/above the bound. The original reverse iterator avoided relying on it too.

chriscandless: Oh, I see this ticket was even closed as `not planned` so maybe that "historical" bug even still there and actually not "historical" at all lol: https://github.com/facebook/rocksdb/issues/9904)

crates/typed-store/src/rocks/safe_iter.rs
chriscandless: Nah.
The name is really confusing, I also was googling it. `seek_for_prev` may land on a given target if that element is present. So it's like "inclusive upper bound". That's the reason why that conditional `prev` is there.

(see the doc - it has this example with `Prev`: https://github.com/facebook/rocksdb/wiki/SeekForPrev)

crates/typed-store/src/util.rs
chriscandless: Damn it...!
Fixed.

chriscandless: This one is tricky.

In Rust when making addition it is expected to specify in the name of the function what would happen in case of overflow: `saturating_add` (checks for max/min), `wrapping_add` (MAX goes to MIN, kind-a C++), `checked_add` / `overflowing_add` (return `Option`/`Result`/`bool`).
Such a plain name as `big_endian_add_one` does not specify anything, which can be treated as "caller checks for max". Practically this means "undefined behaviour in case of overflow", and that is kind-a not Rust-style. So we need to do at least anything about it.

In theory we can super easily check in the loop that max value was reached and return error. But that is waaay too intrusive for this code - nobody actually cares about it, everyone already checks for `is_max`.

So what I did: I put `debug_assert!(!is_max(v))` at the start. And yeah, I know that is ineffective - it can be done more effectively just my modifying the addition loop. But I wanted to exclude that code in prod, so I decided to go with that ineffective double-iteration over value to be able to use lean `debug_assert`.

bingyanglin (approved):

semenov-vladyslav (approved):

crates/typed-store/src/database.rs
semenov-vladyslav: I see that forward and reverse iterators have slightly different initialization ways, and None for upper bound works here. Thought it would just be nicer and symmetric with upper bound set to the provided value, especially that we already have one.

crates/typed-store/src/util.rs
semenov-vladyslav: `debug_assert!` is one way to do that. Another way would be to put `unreachable!` after the loop and instead of breaking the loop in the `else` branch -- just `return;`.

crates/typed-store/src/database.rs
chriscandless: If we add it there just for symmetric logic, it may confuse someone, because the upper bound is already implemented using `seek_for_prev`. So additionally setting the upper bound via readopts may look strange and confusing. To make it less confusing we'd have to remove the `seek_for_prev`logic - but its necessity was explained in the previous comment.

crates/typed-store/src/util.rs
chriscandless: You know what? That is actually more justified than what I did. My code has a chance to return wrong key in prod, which is not good. Panicking is to be on a safer side, which is good, especially considering it is super easy to do here. So switched to `unreachable!`

muXxer (approved):

roman1e2f5p8s (approved):

miker83z (approved):

PR #11866, opened 2026-06-11
feat(iota-types+light-client+ingestion-core): deduplicate committee-chain verification into `CommitteeChainVerifier`

muXxer:
# Description of change

The "walk over end-of-epoch checkpoints, verify each epoch's certified closing checkpoint against the current committee, then advance to the committee its end_of_epoch_data elects" was implemented independently in `iota-light-client` and in `iota-data-ingestion-core::EpochBoundaryVerifier`. Duplicating this logic risks the security-relevant checks drifting apart.

This PR extracts the per-epoch step into a single shared type, `committee::CommitteeChainVerifier` in `iota-types`:
- verify_epoch_close(summary) verifies the summary as the certified close of the current epoch and advances to the next committee; errors leave the verifier unchanged.
- Deliberately a sync, transport-agnostic state machine: consumers keep their own fetch loops (files/RPC, remote-store stream, in-memory list), while all verification logic (epoch match, signatures, close-of-epoch requirement, next-committee construction) is single-sourced.
- Cheap structural checks run before the expensive BLS verification; they can only reject, nothing is trusted until signatures verify.
- iota-light-client and EpochBoundaryVerifier are refactored onto it (net code removal, behavior unchanged).

### DIFF
--- a/crates/iota-types/src/committee.rs
+++ b/crates/iota-types/src/committee.rs
@@ -529,7 +529,12 @@ impl CommitteeChainVerifier {
             .expect("checked before verification");
 
         self.committee = Committee::new(
-            self.committee.epoch + 1,
+            self.committee
+                .epoch
+                .checked_add(1)
+                .ok_or(IotaError::AdvanceEpoch {
+                    error: "epoch number overflow".to_string(),
+                })?,
             end_of_epoch_data
                 .next_epoch_committee
                 .iter()
--- a/crates/iota-types/src/committee.rs
+++ b/crates/iota-types/src/committee.rs
@@ -25,6 +25,7 @@ use crate::{
         AuthorityKeyPair, AuthorityPublicKey, NetworkPublicKey, random_committee_key_pairs_of_size,
     },
     error::{IotaError, IotaResult},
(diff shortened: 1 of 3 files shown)


semenov-vladyslav (approved):

crates/iota-types/src/committee.rs
```diff
@@ -451,12 +452,109 @@ impl Display for CommitteeWithNetworkMetadata {
...
+        }
+
+        let verified = summary.try_into_verified(&self.committee)?;
+        let end_of_epoch_data = verified
+            .end_of_epoch_data
+            .as_ref()
+            .expect("checked before verification");
+
+        self.committee = Committee::new(
+            self.committee.epoch + 1,
```
Thoralf-M: This was .checked_add(1)
            .expect("epoch number overflow") before, might want to keep that, even if it should never be relevant in production

bingyanglin (approved):

Alex6323 (approved):

alexsporn (approved):

crates/iota-types/src/committee.rs
muXxer: good catch!

kodemartin (approved):
lgtm 🩰

PR #11875, opened 2026-06-12
feat(consensus): add sliding-window leader scorer

### LINKED ISSUE #11886: Add sliding window logic for leader schedule

VorobyevIlya:
# Description of change

Adds a sliding-window reputation scorer as an alternative score source for
the existing `LeaderSwapTable`, in place of the `ScoringSubdag` snapshot.
Instead of recomputing reputation from a fixed snapshot every N commits, it
maintains a running aggregate over a configurable sliding window and updates
it incrementally on every commit.

The new code is dead in this PR — a follow-up will dispatch to it behind a
protocol flag and rebuild the swap table from its scores — but it lands here
as a reviewable, testable unit.

Two pieces:

- `leader_scoring::compute_per_commit_contribution` — pure function computing
  per-authority score deltas for a new commit, given the three previous
  pending committed subdags. Distributed-vote scoring with the propagation
  lookahead bounded at r+2.
- `sliding_window_schedule::SlidingWindowSchedule` — stateful scorer that
  maintains a running per-authority aggregate over a configurable window
  (adding each commit's contribution, evicting the oldest when the window is
  full) and exposes it via `reputation_scores()` as `ReputationScores` — the
  same type the swap table is built from — tagged with the window's commit
  range. Restart-replay via `from_committed_subdags`.

Leader selection is unchanged: the swap table still picks leaders. This PR
only changes how the scores feeding it can be produced. Protocol-config
integration, the update cadence, and restart recovery are deferred to the
wiring follow-up.

### DIFF
--- a/crates/starfish/core/src/leader_scoring.rs
+++ b/crates/starfish/core/src/leader_scoring.rs
@@ -244,6 +244,9 @@ impl ScoringSubdag {
 
 /// Walks `[c_minus_2, c_minus_1, c]` in order, returning every commit up to and
 /// **including** the first whose leader round is strictly greater than `upper`.
+/// If none exceed `upper`, returns all three — this only happens if the caller
+/// supplied commits that violate the strictly-increasing-leader-rounds
+/// invariant, in which case the contribution will degrade gracefully to zeros.
 fn scan_until_leader_round_above<'a>(
     c_minus_2: &'a SubDagBase,
     c_minus_1: &'a SubDagBase,
@@ -254,13 +257,10 @@ fn scan_until_leader_round_above<'a>(
     for cmt in [c_minus_2, c_minus_1, c] {
         out.push(cmt);
         if cmt.leader.round > upper {
-            return out;
+            break;
         }
     }
-    panic!(
-        "scan must end on a commit with leader.round > {upper}; \
-         strictly-increasing-leader-rounds invariant violated upstream",
-    );
+    out
 }
 
 /// Compute the per-commit score contribution for the new commit `c`, scoring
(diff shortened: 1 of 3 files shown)


crates/starfish/core/src/leader_scoring.rs
```diff
@@ -242,6 +242,105 @@ impl ScoringSubdag {
...
+    for (author, voting_refs) in &voting_blocks_by_author {
+        // Equivocation in the lookback window → zero contribution.
+        if voting_refs.len() != 1 {
+            continue;
+        }
+        let voting_ref = voting_refs[0];
+        let mut certifying_stake: u64 = 0;
+        for (cert_author, cert_ancestors) in &certifying_blocks {
+            if cert_ancestors.contains(&voting_ref) {
+                certifying_stake += committee.stake(*cert_author);
```
polinikita: Certifying stake is double-counted on equivocation. Why not using `add` method that avoids that?

crates/starfish/core/src/sliding_window_schedule.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// tagged with the index of the commit it scored. Retained so eviction can
+    /// subtract them from the aggregate and so the window's commit range can be
+    /// reported alongside the scores.
+    scores_entries: VecDeque<(CommitIndex, Vec<u64>)>,
+}
+
+impl SlidingWindowSchedule {
+    /// Creates a fresh scorer. `window_size` is clamped to a minimum of 1.
+    pub(crate) fn new(context: Arc<Context>, window_size: u32) -> Self {
+        let window_size = window_size.max(1);
```
polinikita: assert might be better here as handling incorrect values should be treated the same way everywhere

crates/starfish/core/src/sliding_window_schedule.rs
```diff
@@ -0,0 +1,225 @@
...
+//! recent `window_size` scored commits. The aggregate updates on every commit
+//! (O(1) amortized over committee size): each commit's contribution is added,
+//! and once the window is full the oldest is evicted and subtracted.
+//!
+//! [`SlidingWindowSchedule::reputation_scores`] exposes the aggregate as
+//! [`ReputationScores`] — the same type the V2 `LeaderSwapTable` is built from.
+//! The sliding window is an alternative *score source* for the existing swap
+//! table; it does not select leaders itself.
+
+#![cfg_attr(
```
polinikita, line 32: The module seems not to be implemented yet. I would include more details, methods here to see the context of usage. E.g. how replay_start is used.

crates/starfish/core/src/leader_scoring.rs
VorobyevIlya: Good idea, fixed

crates/starfish/core/src/sliding_window_schedule.rs
VorobyevIlya: fixed

polinikita (approved):

crates/starfish/core/src/sliding_window_schedule.rs
VorobyevIlya: modified comments

PR #11888, opened 2026-06-12
refactor(iota-data-ingestion): add hybrid historical checkpoint store for all workers.

### LINKED ISSUE #10791: [iota-data-ingestion]: all workers add hybrid storage checkpoint support
With the ability of fullnodes to prune their internal database, it becomes problematic to process historical checkpoints form a fullnode.

In this regard we need to refactor the iota-data-ingestion to accept historical & live URLs in order to sync older checkpoints which might been pruned on the fullnode.

sergiupopescu199:
# Description of change

This PR adds the ability to either provide the fullnode gRPC URL or the hybrid historical store as a remote source for checkpoint ingestion.

> [!NOTE]
> The `Blob` worker only works with the fullnode gRPC URL, this is because by design its purpose is to only ingest current epoch checkpoint only, making the historical ingestion unnecessary.

fixes #10791 

- Ran the ingestion workers by providing the hybrid historical urls. 

> [!NOTE]

### DIFF
--- a/dev-tools/iota-data-ingestion/live/config.yaml
+++ b/dev-tools/iota-data-ingestion/live/config.yaml
@@ -5,22 +5,34 @@
 # Using the same path as the Node allows direct file access instead of HTTP requests,
 # significantly improving sync speed.
 path: "./test-checkpoints"
-# IOTA Node gRPC URL
+# Remote checkpoint store
 #
-# The IOTA Node's gRPC URL acts as a fallback mechanism for fetching checkpoints
-# when they are not available in the local ingestion path. This provides redundancy
-# and ensures data availability even when local files are not accessible.
+# Acts as a fallback mechanism for fetching checkpoints that are not available in the local
+# ingestion path. Provides redundancy and keeps ingestion running even when local
+# files are not accessible.
 #
-# Note: This URL is optional when:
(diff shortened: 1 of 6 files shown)


dev-tools/iota-data-ingestion/live/config.yaml
```diff
@@ -5,22 +5,34 @@
...
-# and ensures data availability even when local files are not accessible.
+# Acts as a fallback mechanism for fetching checkpoints that are not available in the local
+# ingestion path. Provides redundancy and keeps ingestion running even when local
+# files are not accessible.
 #
-# Note: This URL is optional when:
-# 1. The ingestion path matches the IOTA Node's path on same host machine)
-# 2. You don't need a fallback mechanism for data retrieval
-# 3. Local filesystem access is sufficient for your use case
+# Two source types are supported (pick one):
```
tomxey: This is actually not true for the live/blob worker, right?
isn't this comment misleading?

sergiupopescu199: fixed: https://github.com/iotaledger/iota/pull/11888/commits/eb3766c2de56fa8a5bc3f9ad848db9ca2bd0fe49

tomxey (approved):

kodemartin (approved):
lgtm 🍒

PR #11892, opened 2026-06-12
fix(starfish): add local validation and arithmetic guard rails

### LINKED ISSUE #11891: fix(starfish): harden local validation and arithmetic in consensus
Several intra-node validation gaps in the starfish consensus crates. None are peer-protocol changes; all are local guard rails.

**Key points:**
- `ReputationScores::from_scores_desc` indexes the per-authority score vector with an `AuthorityIndex` taken from fetched commit data without a bounds check (potential panic from peer-influenced data).
- `Store::read_fast_sync_ongoing` maps a RocksDB read error to "not ongoing", silently mis-routing recovery instead of surfacing the storage failure.
- `Committee::new` accepts zero-stake authorities and computes total stake with an unchecked sum that wraps in release builds.
- `Parameters` has no validation: zero values for fetch/bundle limits stall synchronization and dissemination, and a zero tonic keepalive interval (also used as the keepalive timeout) tears down every connection.
- `last_received + 1` in the block-bundle subscription handler can overflow on a peer-supplied `Round::MAX`.
- `round + 1` in the threshold clock can overflow on a far-future block round.

polinikita:
# Description of change

Hardens intra-node assumptions in the starfish consensus crates with local guard rails. No peer-protocol or wire-format changes. The change:

- Rejects zero-stake authorities in `Committee::new`.
- Detects total-stake overflow instead of wrapping the sum.
- Computes quorum thresholds safely for large stake values (widened to `u128` for the doubling).
- Rejects invalid zero-valued consensus parameters at startup (`Parameters::validate`).
- Prevents round overflow at `Round::MAX` (saturating arithmetic in the threshold clock and the peer-driven block-bundle subscription).
- Rejects out-of-range authority indices in fetched reputation scores instead of panicking.
- Stops treating storage-read failures as a valid `false` (`read_fast_sync_ongoing`).
- Propagates recovery storage and output-channel errors consistently through the commit-observer recovery chain.
- Reports missing persisted block headers as structured errors.
- Improves validation coverage with focused unit tests.

Fixes #11891

### DIFF
--- a/crates/starfish/config/src/committee.rs
+++ b/crates/starfish/config/src/committee.rs
@@ -68,7 +68,6 @@ impl Committee {
             .map(|a| a.stake)
             .try_fold(0u64, u64::checked_add)
             .expect("Total stake must not overflow u64!");
-        assert_ne!(total_stake, 0, "Total stake cannot be zero!");
         // Widen to u128 for the doubling; the result is at most total_stake.
         let quorum_threshold = (2 * total_stake as u128 / 3 + 1) as u64;
         let validity_threshold = total_stake.div_ceil(3);
--- a/crates/starfish/config/src/committee.rs
+++ b/crates/starfish/config/src/committee.rs
@@ -43,6 +43,10 @@ pub struct Committee {
 }
 
 impl Committee {
+    /// Panics on invalid input: empty or oversized committee, or zero or
+    /// overflowing stake. The committee is computed on chain, so violations
+    /// indicate a corrupted or misconstructed configuration rather than a
+    /// recoverable condition.
     pub fn new(epoch: Epoch, authorities: Vec<Authority>) -> Self {
         assert!(!authorities.is_empty(), "Committee cannot be empty!");
         assert!(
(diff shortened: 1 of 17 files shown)


crates/starfish/config/src/committee.rs
```diff
@@ -51,9 +55,22 @@ impl Committee {
...
+                authority.hostname
+            );
+        }
+
+        let total_stake = authorities
+            .iter()
+            .map(|a| a.stake)
+            .try_fold(0u64, u64::checked_add)
+            .expect("Total stake must not overflow u64!");
         assert_ne!(total_stake, 0, "Total stake cannot be zero!");
```
piotrm50: Isn't this assert dead now? With the per-authority `stake > 0` check above plus a non-empty committee, the sum can't be zero anymore. Probably safe to drop (or leave a one-liner saying it's covered above).

crates/starfish/config/src/parameters.rs
```diff
@@ -216,6 +216,73 @@ impl Parameters {
...
+            ),
+            (
+                "tonic.message_size_limit",
+                self.tonic.message_size_limit as u128,
+            ),
+            (
+                "tonic.keepalive_interval",
+                self.tonic.keepalive_interval.as_nanos(),
+            ),
+        ];
```
piotrm50, line 288: `tonic.excessive_message_size` isn't in this list. Unlike the other tonic fields it has no documented "0 disables it" meaning, so a stray 0 would just flag every message as excessive. Worth adding here?

piotrm50 (approved):

VorobyevIlya (approved):

PR #11894, opened 2026-06-14
feat(iota-tool): add backfill-checkpoint-summaries tool

muXxer:
# Description of change

Formal-snapshot restore no longer touches the checkpoint archive. The default (and now only) path seeds the end-of-epoch summaries and committees from the snapshot's chain-verified EPOCH_INFO; the `--all-checkpoints flag`, the `archive_store_config` plumbing, and the `start_summary_sync_from_archive` machinery are removed from download_formal_snapshot.

The full-summary-history download, previously bolted onto restore as `--all-checkpoints`, becomes a standalone command, iota-tool `backfill-checkpoint-summaries`, runnable on any stopped node. It downloads every intermediate checkpoint summary from the archive into the node's checkpoint store up to `min(highest_synced, archive_latest)`, optionally verifying the chain pairwise from genesis. It only adds historical summaries below the node's existing watermarks. No watermark is moved, so it's safe to re-run, and it's decoupled from restore (a node restored from a formal snapshot can become a full-header source for peers, or serve historical checkpoint queries).

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Added `backfill-checkpoint-summaries` command to `iota-tool` to download the full checkpoint summary history for a stopped node
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-archival/src/reader.rs
+++ b/crates/iota-archival/src/reader.rs
@@ -287,12 +287,11 @@ impl ArchiveReader {
     /// A summary already present in `store` is left untouched; a new one is
     /// inserted only after [`verify_checkpoint`] checks it against the
     /// previous summary (its committee signature and `previous_digest`
-    /// linkage) when `verify` is set. Unlike a bulk unverified load, this
-    /// never overwrites an existing summary and never persists an unverified
-    /// one — so it is safe to run against a live node's store. Summaries are
-    /// processed in order (via [`get_or_insert_verified_checkpoint`]'s
-    /// dependency on the previous checkpoint), so the genesis checkpoint must
-    /// already be present.
+    /// linkage). Unlike a bulk unverified load, this never overwrites an
+    /// existing summary and never persists an unverified one — so it is safe
+    /// to run against a live node's store. Summaries are processed in order
+    /// (via [`get_or_insert_verified_checkpoint`]'s dependency on the previous
+    /// checkpoint), so the genesis checkpoint must already be present.
     ///
     /// [`get_or_insert_verified_checkpoint`]: Self::get_or_insert_verified_checkpoint
     pub async fn read_summaries_for_range<S>(
@@ -300,7 +299,6 @@ impl ArchiveReader {
         store: S,
         checkpoint_range: Range<CheckpointSequenceNumber>,
         checkpoint_counter: Arc<AtomicU64>,
-        verify: bool,
     ) -> Result<()>
     where
         S: WriteStore + Clone,
@@ -338,7 +336,7 @@ impl ArchiveReader {
                                     && s.sequence_number < checkpoint_range.end
                             })
                             .try_for_each(|summary| {
-                                Self::get_or_insert_verified_checkpoint(&store, summary, verify)?;
+                                Self::get_or_insert_verified_checkpoint(&store, summary, true)?;
                                 checkpoint_counter.fetch_add(1, Ordering::Relaxed);
                                 Ok::<(), anyhow::Error>(())
                             })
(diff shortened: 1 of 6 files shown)


crates/iota-tool/src/lib.rs
```diff
@@ -586,199 +585,215 @@ pub async fn restore_from_db_checkpoint(
...
+        &genesis_committee,
+        None,
+    ));
+    let checkpoint_store = CheckpointStore::new(&node_db_path.join("checkpoints"));
+    let store = AuthorityStore::open_no_genesis(perpetual_db, false, &Registry::default())?;
+    let cache_traits = build_execution_cache_from_env(&Registry::default(), &store);
+    let state_sync_store =
+        RocksDbStore::new(cache_traits, committee_store, checkpoint_store.clone());
+
+    insert_genesis_checkpoint(&checkpoint_store, &genesis)?;
```
bingyanglin: There's no chain-id check on `--genesis`. If an operator passes the wrong network's (or a stale) genesis, `insert_genesis_checkpoint`'s guard keys on that foreign digest so it isn't skipped, and `update_highest_synced_checkpoint` has no monotonic guard, so it overwrites checkpoint 0 and resets highest_synced down to 0, which breaks the "never moves a watermark" promise. Maybe bail before touching anything if `get_checkpoint_by_sequence_number(0)` exists and its digest doesn't match the passed genesis, like `download_formal_snapshot` already does?

crates/iota-tool/src/lib.rs
```diff
@@ -586,199 +585,215 @@ pub async fn restore_from_db_checkpoint(
...
+        ProgressStyle::with_template("[{elapsed_precise}] {wide_bar} {pos}/{len} ({msg})").unwrap(),
+    ));
+    let download_counter = Arc::new(AtomicU64::new(0));
+    spawn_rate_ticker(
+        download_bar.clone(),
+        download_counter.clone(),
+        "checkpoints synced per sec",
+    );
+    archive_reader
+        .read_summaries_for_range_no_verify(
```
bingyanglin: This writes the whole `[1, target]` range into the live DB before `verify_checkpoint_range runs`, and verify panics on a bad summary instead of returning Err, so a wrong-network or tampered archive overwrites the restore's verified end-of-epoch summaries before anything checks them. The old `start_summary_sync` wrote into a throwaway `staging/` dir; this one writes in place. Maybe download into a staging store and merge only after verify passes, or use the `get_or_insert_verified_checkpoint` so existing summaries aren't overwritten and each is checked before insert?

crates/iota-tool/src/lib.rs
```diff
@@ -586,199 +585,215 @@ pub async fn restore_from_db_checkpoint(
...
-                let total_checkpoints_per_sec =
-                    num_summaries as f64 / s_instant.elapsed().as_secs_f64();
-                cloned_progress_bar.set_position(s_start + num_summaries);
-                cloned_progress_bar.set_message(format!(
-                    "checkpoints synced per sec: {total_checkpoints_per_sec}"
-                ));
-                tokio::time::sleep(Duration::from_secs(1)).await;
-            }
-        });
+    println!("Backfilled checkpoint summaries up to checkpoint {target}");
```
bingyanglin, line 710: When the archive lags the node (`archive_latest < highest_synced`), `target` caps at the archive and `(archive_latest, highest_synced]` stays end-of-epoch-only, but we still print `Backfilled ... up to {target}` with no warning. Maybe we need more precise warning/logs here?

crates/iota-archival/src/reader.rs
```diff
@@ -281,7 +281,20 @@ impl ArchiveReader {
     /// Load checkpoints from archive into the input store `S` for the given
     /// checkpoint range. Summaries are downloaded out of order and inserted
     /// without verification
```
bingyanglin: Can remove these three lines which are already deprecated.
```suggestion
```

bingyanglin (approved):

PR #11913, opened 2026-06-15
feat: add on-chain Attestor registry in the system state

alexsporn:
# Description of change

On-chain registry for third-party ("explicit") transaction attestors, the foundation for opening attestation beyond the validator set.

**Registry (`attestor_registry.move`, dynamic field on `IotaSystemState`):**

- Permissionless registration with a bonded escrow and a dedicated, rotatable signing key (proof-of-possession verified by a native); the registering address is the stable identity.
- Bond levels are rates (basis points) applied to `min_validator_joining_stake` — 10% to join, 5% before eviction — so they stay consistent as the stake is tuned.
- Escrow is held as an at-stake bond capped at the joining bond plus the excess above it; top-ups land in the excess and fold in only at the boundary, so they cannot rescue an attestor slashed below the threshold mid-epoch. Eviction burns the entire escrow.
- All membership changes settle at epoch boundaries: activations (re-checked against the current joining bond), staged key rotations, voluntary removals (refund), low-bond eviction (burn), inactivity drop (penalty + refund). Exits and events are batched to stay within epoch-change transaction limits.
- Display metadata lives as per-attestor dynamic fields; cleanup is enforced by a must-consume departure list.

**Safety properties:**

- Disabling the feature flag cannot freeze escrow: deregistration and boundary refunds process whenever the registry exists; only parameter-reading work (eviction, inactivity, rebalance, activations) suspends.
- The epoch-start state upgrades to the attestor-carrying version only while the flag is on, so chains with the feature off keep storing a state older binaries can decode.
- The epoch-change transaction cannot abort on registry processing: parameter reads are guarded by the same invariants that populate the registry, and metadata removal tolerates absence.
- Feature getters use effective semantics (flag AND prerequisites) on both the Rust and Move side, so an incomplete flag combination reads as disabled everywhere instead of panicking or diverging.

**Off-chain:**

- `iota attestor` CLI: register / deregister / deposit-bond / rotate-key / update-{name,description,url,logo} / display, with dedicated `attestor.key` handling and the default bond derived from the network's protocol config.
- Rust mirrors of the registry types, the epoch-start attestor snapshot (`EpochStartSystemStateV3`) exposed to the epoch store for future explicit-attestation verification, and snapshot-restore support.

**Not in this change:** rewards, penalties and per-attestor statistics (see the incentive design); the activity feed — until it is threaded in, every attestor is reported active, so the inactivity drop cannot fire; acceptance of `Explicit` attestations at consensus admission (still rejected as unsupported).

- Design: attestation incentive mechanism design document (bond model, eviction economics).
- fixes https://github.com/iotaledger/iota-private/issues/388

Move unit and scenario suites for the registry (registration, bond derivation, escrow split, rotation, eviction, inactivity, activation re-check, flag-off boundary processing), CLI unit tests, protocol-config snapshot tests, and `cargo check --workspace --all-targets` clean.

### Release Notes

- [x] Protocol: Adds the on-chain attestor registry to the system package and the attestor parameters (bond rates, count and inactivity limits), all inert unless the `enable_external_attestation` feature flag is enabled.
- [x] Nodes (Validators and Full nodes): When `enable_external_attestation` is enabled, the epoch-start state carries the active attestor set (new `EpochStartSystemStateV3` variant, also handled on formal snapshot restore); with the flag off, nodes keep storing the previous state version.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: New `iota attestor` command group (register, deregister, deposit-bond, rotate-key, update-name/description/url/logo, display) for managing a third-party attestor; requires a network with `enable_external_attestation` enabled.
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-framework/packages/iota-system/sources/iota_system.move
```diff
@@ -215,6 +214,160 @@ public entry fun set_candidate_validator_commission_rate(
...
+    if (!dynamic_field::exists_(&self.id, attestor_registry::registry_key())) {
+        dynamic_field::add(
+            &mut self.id,
+            attestor_registry::registry_key(),
+            attestor_registry::new(),
+        );
+    };
+    dynamic_field::borrow_mut(&mut self.id, attestor_registry::registry_key())
+}
+
```
lzpap, line 231: Is there a strong reason why the registry needs to live as a dynamic field of `IotaSystemState` and not as an entry in the `extra_fields` bag of the `IotaSystemStateInnerX`?
 - The public facing API would anyway be implemented on the wrapper, implementation passed down to the inner struct, but
 - advance_epoch would become cleaner and the actual logic would happen where it already is happening - within the inner struct's implementation.

cc @valeriyr

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+        let mut exit_reasons = vector<u8>[];
+        if (feature_enabled) {
+            let low_bond_threshold = low_bond_threshold();
+            let max_inactivity_epochs: u64 = protocol_config::get_attr(
+                ATTESTOR_MAX_INACTIVITY_EPOCHS_PARAM,
+            );
+            inactivity_penalty = protocol_config::get_attr(ATTESTOR_INACTIVITY_PENALTY_PARAM);
+            self.active_attestors.length().do!(|i| {
+                let entry = &self.active_attestors[i];
+                if (entry.bond.value() < low_bond_threshold) {
```
bingyanglin: This checks only `bond` before `excess_bond` is merged. A parameter increase can evict active attestors and burn all funds, even after a top-up or removal request. Is this intended? Also do we need to add a parameter-change test?

crates/iota-types/src/iota_system_state/attestor_registry.rs
```diff
@@ -0,0 +1,579 @@
...
+/// plain schemes (ed25519 / secp256k1 / secp256r1). Backs the
+/// `attestor_registry::validate_attestor_pubkey` native.
+pub fn verify_attestor_pubkey(pubkey: &[u8]) -> Result<(), u64> {
+    let Some((&flag, key_bytes)) = pubkey.split_first() else {
+        return Err(E_INVALID_ATTESTOR_PUBKEY);
+    };
+    let scheme = SignatureScheme::from_byte(flag).map_err(|_| E_INVALID_ATTESTOR_PUBKEY)?;
+    match scheme {
+        SignatureScheme::Ed25519 | SignatureScheme::Secp256k1 | SignatureScheme::Secp256r1 => {
+            PublicKey::try_from_bytes(scheme, key_bytes)
```
bingyanglin: This accepts both 33-byte and 65-byte secp keys. The same key can register twice because Move compares raw bytes. Maybe we require canonical encoding?

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+    ctx: &mut TxContext,
+): (Balance<IOTA>, DepartedAttestors) {
+    let mut evicted_bonds = balance::zero<IOTA>();
+    let mut departed = vector<address>[];
+    let mut exited = vector<AttestorExitInfo>[];
+
+    // --- 1. Combined exits ---
+    // Only active attestors can ever exit, and only active attestors can
+    // populate `pending_removals` (deregister requires an active entry), so
+    // a chain with the feature flag on but the exit-threshold params unset
```
bingyanglin: `register` reads only the joining rate and the max count, not the three exit params read at L488-492. A config with the flag on and one exit param `None` registers fine, then aborts the change-epoch tx at the first boundary with an active attestor. Should we read all five in `register`?

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,733 @@
...
+        )
+    })?;
+    Ok(config_dir.join("attestor.key"))
+}
+
+fn write_attestor_key(path: &Path, keypair: &IotaKeyPair, allow_overwrite: bool) -> Result<()> {
+    if !allow_overwrite && path.exists() {
+        bail!("an attestor key already exists at {path:?}; move it away if you are re-registering");
+    }
+    write_keypair_to_file(keypair, path)?;
```
bingyanglin: The key file gets mode `0600` only after it is written. Maybe create it with mode `0600` to remove the permission window?

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+                    attestor_address,
+                    reason: EXIT_REMOVAL,
+                    refunded_amount: bond.value(),
+                    burned_amount: 0,
+                });
+                transfer::public_transfer(coin::from_balance(bond, ctx), attestor_address);
+            } else {
+                rebalance(&mut entry, min_joining_bond);
+                activated.push_back(entry.attestor_address);
+                self.active_attestors.push_back(entry);
```
bingyanglin, line 662: `activation_epoch`/`last_active_epoch` are set at registration. If activation is delayed (feature off, safe mode), the entry starts with inactivity debt once a real feed replaces the "mark all" placeholder. Do we need to set both to `new_epoch` here or we don't handle this edge case?

crates/iota-node/src/lib.rs
```diff
@@ -2201,6 +2201,20 @@ impl IotaNode {
...
+        // Attach the new epoch's active attestor snapshot so the epoch store
+        // exposes it via `attestor_set()` for explicit-attestation
+        // verification. Feature-gated: with the flag off the epoch-start
+        // state keeps its previous version, decodable by older binaries.
+        let next_epoch_start_system_state =
+            match iota_types::iota_system_state::attestor_registry::epoch_start_attestors(
+                state.get_object_store().as_ref(),
+                next_epoch_start_system_state.protocol_version(),
+                state.get_chain_identifier(),
+            )? {
```
bingyanglin: An `Err` here ends `monitor_reconfiguration` with a `warn!`. The node keeps serving and never reconfigures again. The two calls below use `.expect(..)`. Do we need to fail the node explicitly, and validate the registry type before BCS decoding?

crates/iota-framework/packages/iota-system/sources/iota_system.move
```diff
@@ -536,9 +690,36 @@ fun advance_epoch(
...
+    let has_registry = dynamic_field::exists_(&wrapper.id, attestor_registry::registry_key());
+    let attestor_evicted_bonds = if (feature_enabled || has_registry) {
+        let registry = load_attestor_registry_mut(wrapper);
+        if (feature_enabled) {
+            // The per-attestor activity feed is not threaded into this
+            // transaction yet; until it is, report every attestor as active
+            // so none are dropped for inactivity.
+            let mut all_indices = vector[];
+            registry.active_count().do!(|i| all_indices.push_back(i));
+            registry.refresh_activity(all_indices, new_epoch - 1);
```
bingyanglin, line 717: Not sure if we need to consider safe mode. But`refresh_activity` runs only when the feature is on, and in safe mode this block does not run at all, but `new_epoch` keeps advancing. So once the real feed replaces the mark-all placeholder, the first boundary after a feature-off or safe-mode window longer than `attestor_max_inactivity_epochs` drops every active attestor with a penalty. If the fix is a `last_processed_epoch` on the registry, that field must be in `AttestorRegistryV1` now.

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+    });
+    option::none()
+}
+
+// === Bond top-up ===
+
+/// Add `additional` to the sender's excess (active or pending entry). It
+/// folds into the at-stake bond only at the boundary rebalance, so the
+/// eviction check at the very next boundary does not see it.
+public(package) fun deposit(
```
bingyanglin, line 395: `deposit` does not check `pending_removals`; `rotate_key` does. After a deregister the deposit is refunded at best and burned at worst (eviction wins over removal). Is this intended?

crates/iota-framework/packages/iota-system/sources/iota_system.move
miker83z: It's the first field that will take part to a later "decomposition" of the IotaSystemState. After the attestor changes, there will (hopefully) be an upgrade to V3 in which we will start to make also the rest of fields dynamic. This will make upgrades smoother.

crates/iota-types/src/iota_system_state/attestor_registry.rs
miker83z: Correct, it should only allow the compressed form otherwise one can register attestor A with a key in a compressed form and then attestor B with the same key but uncompressed.

crates/iota-framework/packages/iota-system/sources/iota_system.move
miker83z: If a safe-mode window is longer than attestor_max_inactivity_epochs (which is currently 7) that is a bigger trouble 😅

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+            self.active_attestors[idx].last_active_epoch = ending_epoch;
+        }
+    });
+}
+
+// === Key rotation ===
+
+/// Stage a replacement signing key for the sender's active entry; the key
+/// is swapped in place at the next epoch boundary. Staging again before the
+/// boundary overwrites the previously staged key.
```
miker83z:
```suggestion
/// boundary overwrites the previously staged key.
/// Re-staging the same key aborts with `EDuplicatePubkey`.
```

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+            } else if (reason == EXIT_INACTIVITY) {
+                let penalty_amount = inactivity_penalty.min(bond.value());
+                evicted_bonds.join(bond.split(penalty_amount));
+                exited.push_back(AttestorExitInfo {
+                    attestor_address,
+                    reason,
+                    refunded_amount: bond.value(),
+                    burned_amount: penalty_amount,
+                });
+                transfer::public_transfer(coin::from_balance(bond, ctx), attestor_address);
```
miker83z:
```rust
if (bond.value() > 0) {  
    transfer::public_transfer
} else  {
    bond.destroy_zero();
};
```

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+            } else {
+                exited.push_back(AttestorExitInfo {
+                    attestor_address,
+                    reason,
+                    refunded_amount: bond.value(),
+                    burned_amount: 0,
+                });
+                transfer::public_transfer(coin::from_balance(bond, ctx), attestor_address);
```
miker83z: Same as https://github.com/iotaledger/iota/pull/11913/changes#r3932826765

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,868 @@
...
+                next_epoch_attestor_pubkey.destroy!(|_| ());
+                bond.join(excess_bond);
+                departed.push_back(attestor_address);
+                exited.push_back(AttestorExitInfo {
+                    attestor_address,
+                    reason: EXIT_REMOVAL,
+                    refunded_amount: bond.value(),
+                    burned_amount: 0,
+                });
+                transfer::public_transfer(coin::from_balance(bond, ctx), attestor_address);
```
miker83z: same as https://github.com/iotaledger/iota/pull/11913/changes#r3932826765

iota-execution/latest/iota-move-natives/src/attestor_registry.rs
```diff
@@ -0,0 +1,77 @@
...
+    debug_assert!(ty_args.is_empty());
+    debug_assert!(args.len() == 3);
+
+    let cost_params = context
+        .extensions_mut()
+        .get::<NativesCostTable>()?
+        .validator_validate_metadata_bcs_cost_params
+        .clone();
+
+    native_charge_gas_early_exit!(context, cost_params.validator_validate_metadata_cost_base);
```
miker83z: validator_validate_metadata_cost_base is 20000. I'd suggest to add a dedicated cost base for the attestor pubkey, similar to the pub key verification native base costs. The higher one is `ecdsa_r1_secp256r1_verify_keccak256_cost_base` (4225) so I'd use that.

roman1e2f5p8s (commented):
Two parts can now be removed from the PR description, as they were fixed during the latest rebase:

> **Known failure, pre-existing and out of scope:** `test_attestor_cli_lifecycle` fails with `Function Not Found` because test-cluster genesis prefers the protocol v32 framework bytecode snapshot (landed on develop 2026-07-23) over the live framework, so the branch's new system-package entry points don't exist on-chain in clusters. Resolved by the planned feature-branch rebase onto develop and re-gating the attestation work at protocol version 33.

and

> `test_attestor_cli_lifecycle` red for the snapshot/genesis reason above.

crates/iota-e2e-tests/tests/attestor_registry_tests.rs
```diff
@@ -0,0 +1,202 @@
...
+            "IOTA_PROTOCOL_CONFIG_FEATURE_FLAGS_OVERRIDE_ENABLE_EXTERNAL_ATTESTATION",
+            "true",
+        ),
+    ]);
+
+    let test_cluster = TestClusterBuilder::new().build().await;
+    let sender = test_cluster.get_address_0();
+
+    // One gas coin for gas and a separate whole coin as the bond (each default
+    // test coin far exceeds MIN_ATTESTOR_JOINING_BOND).
```
roman1e2f5p8s: `MIN_ATTESTOR_JOINING_BOND` is not defined anywhere

crates/iota-framework/packages/iota-system/sources/iota_system.move
```diff
@@ -215,6 +214,161 @@ public entry fun set_candidate_validator_commission_rate(
...
+            attestor_registry::registry_key(),
+            attestor_registry::new(),
+        );
+    };
+    dynamic_field::borrow_mut(&mut self.id, attestor_registry::registry_key())
+}
+
+/// Register the sender as an attestor with a dedicated signing key
+/// (`flag || raw pubkey`, plain schemes only) and its proof of possession,
+/// locking `bond` (>= MIN_ATTESTOR_JOINING_BOND). Takes effect at the next
```
roman1e2f5p8s: `MIN_ATTESTOR_JOINING_BOND` seems a dead constant

crates/iota-types/src/iota_system_state/attestor_registry.rs
vekkiokonio: Right! Fixed in cbd248e0aacdd71ecfd18769687f9bbe271b7785

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: I added a helper function to assert all params - see 6bf8d836dd68e11285f3bf70c0eddfcc7eeca86b

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: I'd leave it as it is. To be honest, it is a bit awkward to have attestors topping up *after* requesting to leave the attestor set. However, even though they do it, the deposit is refunded with the rest of the escrow at the boundary unless the attestors misbehaved in the same epoch.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -3380,6 +3421,18 @@ impl ProtocolConfig {
                     // is off.
                     cfg.feature_flags.pcool_skip_immutable_object_locks = true;
                 }
+                35 => {
+                    if chain != Chain::Testnet && chain != Chain::Mainnet {
+                        // Attestor registry parameters (devnet only for now),
+                        // read from Move via get_attr.
+                        // 10% and 5% of `min_validator_joining_stake`.
+                        cfg.attestor_joining_bond_rate = Some(1_000);
```
piotrm50, line 3445: Do we want to add all those parameters directly to the config now? I think it would be easier to have some default values for them in a constant or in the getters. That way this won't need to be rebased constantly every 2 weeks after each release before this feature is merged.

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,736 @@
...
+                        "an attestor key already exists at {key_path:?}; it remains valid \
+                         on-chain until the next epoch boundary, so move it away first if you \
+                         want to keep it, then re-run rotate-key"
+                    );
+                }
+
+                // Written before submitting so a failed or ambiguous submit never loses the
+                // freshly generated key.
+                let keypair = generate_attestor_keypair(key_scheme)?;
+                write_attestor_key(&key_path, &keypair, true)?;
```
piotrm50: why do we allow overwrite here if above we bail if the key path exists?

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Yes, you're right. Fixed in 9df4d22079f221bfc6f6418deb3a920d1e9b3dee

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Good catch, it wasn't done on purpose. I fixed it so only slashing can trigger a burn (ac06c79c40834f320c2512319d46e3cf2e5219a2). Instead, a parameter raise is judged against the whole escrow. I've also added a test.

crates/iota-node/src/lib.rs
vekkiokonio: :+1:

crates/iota-protocol-config/src/lib.rs
```diff
@@ -3380,6 +3421,18 @@ impl ProtocolConfig {
...
+                        cfg.attestor_low_bond_threshold_rate = Some(500);
```
roman1e2f5p8s, line 3446: `attestor_low_bond_threshold_rate` must be below `attestor_joining_bond_rate`, right?
but there appears to be no invariant to enforce that

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,910 @@
...
+    assert_inactivity_params_configured();
+    let min_joining_bond = min_joining_bond();
+    let low_bond_threshold = low_bond_threshold();
+    assert!(bond.value() >= min_joining_bond, EBondTooLow);
+    assert!(
+        self.active_attestors.length() + self.pending_active.length()
+            < protocol_config::get_attr(MAX_ATTESTOR_COUNT_PARAM),
+        ETooManyAttestors,
+    );
+    validate_attestor_pubkey(attestor_pubkey, proof_of_possession, sender);
```
piotrm50: move this below the remaining assertions

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,910 @@
...
+    validate_attestor_pubkey(new_pubkey, proof_of_possession, sender);
+    assert!(!pubkey_in_use(self, &new_pubkey), EDuplicatePubkey);
+    let entry = &mut self.active_attestors[idx];
+    entry.next_epoch_attestor_pubkey = option::some(new_pubkey);
+}
+
+// === Epoch boundary processing ===
+
+/// Process the epoch boundary for the registry. Order:
+/// 0. (Reserved) slashing executes before exits — see the design doc.
```
piotrm50: which design doc? it might be clear now but in a year?

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
```diff
@@ -0,0 +1,910 @@
...
+    // needs, so an incomplete config can never populate the set.
+    //
+    // With the feature disabled only voluntary removals are processed: they
+    // read no params, so escrowed bonds can always exit, while eviction and
+    // inactivity stay suspended.
+    if (!self.active_attestors.is_empty()) {
+        let mut inactivity_penalty: u64 = 0;
+        let mut exit_indices = vector<u64>[];
+        let mut exit_reasons = vector<u8>[];
+        if (feature_enabled) {
```
piotrm50: what feature?

crates/iota-framework/packages/iota-system/sources/iota_system.move
```diff
@@ -536,9 +690,36 @@ fun advance_epoch(
...
     assert!(ctx.sender() == @0x0, ENotSystemAddress);
+
+    // Registry lives on the wrapper UID, so process it before borrowing the
+    // inner state; evicted bonds are burned inside the inner advance_epoch.
+    // The registry outlives the feature flag: whenever it exists, voluntary
+    // removals and refunds are still processed, so disabling the feature
+    // cannot freeze escrowed bonds. The flag gates everything else
+    // (eviction, inactivity, rebalance, activations) and creates the empty
+    // registry on the first boundary once enabled.
+    let feature_enabled = attestor_registry::is_feature_enabled();
```
piotrm50: here it's clear that it's attestor feature, but a few lines below, or in a different file it's not as obvious. Rename to something more explicit e.g.`attestor_registry_enabled`

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed in 2d5168681024619b726b105c1af8277052e9b15c

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Makes sense, I introduced a `refund` helper function.

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Fixed with a `refund` helper function.

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Fixed with a `refund` helper function.

crates/iota-e2e-tests/tests/attestor_registry_tests.rs
vekkiokonio: Good catch, the comment wasn't updated after the latest changes

crates/iota-framework/packages/iota-system/sources/iota_system.move
vekkiokonio: I updated the comment.

crates/iota-protocol-config/src/lib.rs
vekkiokonio: Yep, correct. I added an assertion

iota-execution/latest/iota-move-natives/src/attestor_registry.rs
vekkiokonio: Sounds good. In 54bb689e21e1a4b9624f05699b20713e183487ef, I added the `attestor_validate_pubkey_cost_base` and set it to 4225.

crates/iota-protocol-config/src/lib.rs
vekkiokonio: What I could do is to define a `apply_attestor_params()` next to the config holding all the attestor values. In this case, the rebase conflict shrinks to one line. Would that be OK, or are you thinking of something else?

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed in abbd43f45974d4126f7f9431cf6ad5d12e269264

crates/iota-framework/packages/iota-system/sources/iota_system.move
vekkiokonio: Done!

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: attestor registry. I renamed it to `attestor_registry_enabled` as you suggested.

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: I improved the comment

crates/iota-framework/packages/iota-system/sources/attestor_registry.move
vekkiokonio: Done

vekkiokonio (approved):

bingyanglin (approved):

crates/iota-types/src/iota_system_state/attestor_registry.rs
```diff
@@ -0,0 +1,585 @@
...
+        SignatureScheme::Secp256r1 => Secp256r1PublicKey::LENGTH,
+        _ => return Err(E_INVALID_ATTESTOR_PUBKEY),
+    };
+    // The secp parsers also accept the 65-byte uncompressed form, and Move
+    // compares keys as raw bytes, so without this check one key could be
+    // registered twice under two encodings.
+    if key_bytes.len() != canonical_len {
+        return Err(E_INVALID_ATTESTOR_PUBKEY);
+    }
+    PublicKey::try_from_bytes(scheme, key_bytes)
```
roman1e2f5p8s, line 110: The length check above fixes the secp case, but `ed25519` has a separate problem on this line. `PublicKey::try_from_bytes` for `ed25519` ends up in `ed25519_consensus::VerificationKey::try_from`, which only checks that the bytes decompress to a curve point. Small-order points pass, and they are 32 bytes, so the length check does not catch them. For those keys, the proof of possession proves nothing.

So someone can register an attestor whose key nobody owns. Once explicit attestations are checked, anything signed by that attestor could have come from anyone, so it cannot be slashed for one and can deny whatever it did sign.

Fix here: for `ed25519`, decompress and reject if `is_small_order()` or `!is_torsion_free()`. Both are needed; the identity passes the second. Also compare the `point.compress()` against the input, since noncanonical encodings would otherwise let one key register twice.

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,742 @@
...
+        gas_budget: Option<u64>,
+    },
+    /// Deregister the sender as an attestor. A pending attestor is refunded
+    /// immediately; an active one is removed and refunded at the next epoch
+    /// boundary.
+    Deregister {
+        #[arg(long)]
+        gas_budget: Option<u64>,
+    },
+    /// Add to the sender's attestor bond (nanos), effective immediately.
```
roman1e2f5p8s: "effective immediately" seems to contradict with what the Move side does

roman1e2f5p8s (approved):

crates/iota-framework/packages_compiled/iota-system
miker83z, line 1: You also need to create the snapshots with `cargo run --bin iota-framework-snapshot`, so that the tests use this binary.

miker83z (approved):

piotrm50 (approved):

crates/iota-framework/packages_compiled/iota-system
vekkiokonio: Will do directly in the feature branch :)

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed the comment

crates/iota-types/src/iota_system_state/attestor_registry.rs
vekkiokonio: Fixed in 845dbd56373125be422acc31833278bc2a8e0513

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,744 @@
...
+
+/// Same encoding as `write_keypair_to_file`, but the file is always newly
+/// created with mode 0600, so the private key is never readable by others
+/// and an existing file is never overwritten.
+fn write_attestor_key(path: &Path, keypair: &SimpleKeypair) -> Result<()> {
+    let contents = keypair.to_bech32().map_err(|e| anyhow!(e))?;
+    let mut file = match fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .mode(0o600)
```
Thoralf-M, line 363: This needs some `#[cfg(unix)]` gating to not break with windows

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,744 @@
...
+            IotaTransactionBlockResponseOptions::new()
+                .with_input()
+                .with_effects(),
+            Some(iota_types::quorum_driver_types::ExecuteTransactionRequestType::WaitForLocalExecution),
+        )
+        .await
+        .map_err(|err| anyhow::anyhow!(err.to_string()))
+}
+
+async fn get_gas_obj_ref(
```
Thoralf-M, line 543: There is some duplication between this and validator_commands.rs, reuse the fns instead of duplicating
`get_gas_obj_ref`, `construct_unsigned_0x5_txn` and the body of `call_0x5` are copies

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed in 23501af5a7c47ec369633ed2c646c13d7633af04

crates/iota/src/attestor_commands.rs
vekkiokonio: Added the helpers to a new `system_txn` module

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,665 @@
...
+            SimpleKeypair::from(get_key_pair_from_rng::<Secp256k1PrivateKey, _>(&mut rng).1)
+        }
+        SignatureScheme::Secp256r1 => {
+            SimpleKeypair::from(get_key_pair_from_rng::<Secp256r1PrivateKey, _>(&mut rng).1)
+        }
+        other => bail!(
+            "unsupported attestor key scheme: {other}, expected ed25519, secp256k1 or secp256r1"
+        ),
+    })
+}
```
Thoralf-M, line 336: Should be able to clean this up like this
```Rust
fn generate_attestor_keypair(scheme: SignatureScheme) -> Result<SimpleKeypair> {
    Ok(match scheme {
        SignatureScheme::Ed25519 => SimpleKeypair::from(Ed25519PrivateKey::random()),
        SignatureScheme::Secp256k1 => SimpleKeypair::from(Secp256k1PrivateKey::random()),
        SignatureScheme::Secp256r1 => SimpleKeypair::from(Secp256r1PrivateKey::random()),
        other => bail!(
            "unsupported attestor key scheme: {other}, expected ed25519, secp256k1 or secp256r1"
        ),
    })
}
```

crates/iota/src/attestor_commands.rs
```diff
@@ -0,0 +1,665 @@
...
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn attestor_key_file_is_owner_read_write_only() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("attestor.key");
+        let kp = SimpleKeypair::from(get_key_pair_from_rng::<Ed25519PrivateKey, _>(&mut OsRng).1);
```
Thoralf-M: Also here can use `Ed25519PrivateKey::random()` instead

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed in b12c60c608ac97348c1da3d5c3e38f9199121f20

crates/iota/src/attestor_commands.rs
vekkiokonio: Fixed in b12c60c608ac97348c1da3d5c3e38f9199121f20

PR #11916, opened 2026-06-15
upstream(core): Wrap notify_read_effects with within_alive_epoch

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@0471c68](https://github.com/MystenLabs/sui/commit/0471c68e336c863f595a6f2da5589eb82b41319f)
- Description:

Wrap the effects notify-read in `submit` with `epoch_store.within_alive_epoch(...)`
so the future is cancelled at the epoch boundary instead of hanging across a
reconfiguration. This avoids the `fatal!("digests must exist")` panic in
`execution_cache.rs` that occurs when the digests become available but their
effects are no longer present after an epoch change.

To enable this, `epoch_store` is threaded into `submit` (now taking an owned
`Arc<AuthorityPerEpochStore>`), and `execute_transaction_impl` takes
`&Arc<AuthorityPerEpochStore>` so it can clone it down.

Upstream fixes the following test failure:

```
thread '<unnamed>' panicked at crates/sui-core/src/execution_cache.rs:540:46:
digests must exist
```

Part of #11885

### DIFF
--- a/crates/iota-core/src/transaction_orchestrator.rs
+++ b/crates/iota-core/src/transaction_orchestrator.rs
@@ -393,7 +393,7 @@ where
     #[instrument(level = "trace", skip_all, fields(tx_digest = ?request.transaction.digest()))]
     pub async fn execute_transaction_impl(
         &self,
-        epoch_store: &AuthorityPerEpochStore,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
         request: ExecuteTransactionRequestV1,
         client_addr: Option<SocketAddr>,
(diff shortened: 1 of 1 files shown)


crates/iota-core/src/transaction_orchestrator.rs
bingyanglin, line 514: Note that the log now is stall, just it will be correct again after https://github.com/MystenLabs/sui/pull/24824 is ported.

bingyanglin (approved):

VorobyevIlya (approved):

PR #11920, opened 2026-06-15
fix(starfish): bound far-future block rounds to cap suspender memory

### LINKED ISSUE #11919: starfish(core): far-future block rounds grow the block suspender without bound
A single Byzantine validator can exhaust a node's memory by streaming self-authored blocks at monotonically increasing far-future rounds.

**Key points:**
- `SignedBlockVerifier::verify` enforces no upper bound on `block.round()`; ancestor rounds are validated only *relative* to the block's own round, so a block at round `local + 1e6` citing 2f+1 fabricated round-(R-1) ancestor refs passes verification.
- Such blocks are suspended (their ancestors never arrive), and the only reclamation of the suspender (`suspended_headers`, `missing_ancestors`, `headers_to_fetch`, `suspended_transactions`) is below the GC floor — which never advances to far-future rounds.
- Sustained streaming therefore grows the suspender (headers + transaction payloads + fetch entries) without bound → OOM. One Byzantine committee member can remotely crash the target.

polinikita:
# Description of change

## Intuition

A Byzantine validator can self-author blocks at far-future rounds (`local + 1e6`, `+1`, …), each citing 2f+1 *fabricated* round-(R-1) ancestor refs. They pass `verify` — which only bounds ancestor rounds *relative* to the block's own round and has no absolute upper bound — but can never connect, so they accumulate per-block state without bound (suspender entries, shard-reconstructor transactions, DagState shards). GC is keyed off `gc_round_for_last_commit`, a *low* watermark, so it never reclaims rounds *above* the frontier. The fix drops a peer-disseminated block once its round is implausibly far above the node's accepted frontier; a node that legitimately trails that far catches up via commit sync, the proper catch-up path.

## What it does

Drop peer-disseminated blocks/headers whose round exceeds
`highest_accepted_round + dag_state_cached_rounds + peer_round_ahead_margin`, enforced at each peer ingress path and again in the block manager, all sharing one formula (`Parameters::peer_disseminated_round_ceiling`) and one metric (`dropped_far_future_blocks_total`, labeled by `source`):

- **`handle_subscribed_block_bundle`** (streamed bundles) — `source="block_bundle"`. Placed after signature/header verification and commit-vote observation, but *before* shard extraction, transaction messages, digest filtering, and `add_shards`. Required here because the streaming path populates `ShardReconstructor::processed_transactions` and `DagState` shard state *before* `BlockManager` ever sees the block. This ingress point checks only the bundle's *primary* block; far-future *additional* headers carried in the same bundle are forwarded to core and bounded by the `BlockManager` layer below.
- **`process_fetched_headers_from_authority`** (header synchronizer) — `source="header_synchronizer"`. The fetch path has no other round bound: `verify_block_headers` accepts any validly-signed header a peer returns (not just the refs we requested), and the only other limit is a header *count* truncate. Placed after verification and vote observation, before headers are forwarded to core — rejecting a round earlier than the block manager would.
- **`BlockManager` accept path** (`drop_far_future` in `try_accept_block_headers` / `try_accept_blocks`, gated on `DataSource::is_peer_disseminated()`) — `source="block_manager"`. Bounds the suspender (`suspended_headers` / `missing_ancestors` / `headers_to_fetch` / `suspended_transactions`). It is the *sole* far-future bound for additional headers carried in a bundle (the bundle ingress checks only the primary block), so this counter legitimately counts those drops; the two dedicated ingress points cover everything else, so any count beyond additional-header drops flags an otherwise-unguarded peer path.

Certified/local sources (`CommitSyncer`, `FastCommitSyncer`, `Recover`, `OwnBlock`) and the transaction/shard synchronizers are exempt — commit sync is the catch-up path and bypasses these ingress paths.

## Why this horizon

- **Anchor = local frontier.** `highest_accepted_round` is Byzantine-unforgeable: a far-future block is never *accepted*, only suspended/dropped, so it cannot inflate the frontier. The quorum-committed round is only a *lower* bound on where the network is (a 2/3 set with a laggard can drag it down), so it is unsafe as an *upper* reject bound.
- **`peer_round_ahead_margin` (new local parameter, default 1000), above `cached_rounds`.** A reconnecting or restarted node re-pushes roughly `dag_state_cached_rounds` of cached history plus its own freshly produced blocks; the horizon must sit above the cache window so those honest catch-up blocks are not rejected. The default is generous enough that the bound never fires in honest operation — no cache-stress test overrides or msim defaults had to change.
- **It is a local consensus parameter, not protocol state.** Divergent values across nodes do not affect safety — they only change how aggressively each node drops blocks that could never connect — so an operator can tune the horizon without a protocol upgrade. Centralized in `peer_disseminated_round_ceiling` with saturating arithmetic so it cannot overflow.
- **This bounds the round *horizon*, not memory.** Actual memory still scales with committee size, ancestors per header, per-slot equivocations (audit M5), and payload bytes — out of scope here.

## Testing

- `BlockManager`: drop of a far-future header, the exact-ceiling boundary (retained at ceiling, dropped at +1), per-disseminated-source coverage (`BlockBundleStream` + `HeaderSynchronizer`), and certified-source exemption (`CommitSyncer` suspended, not dropped).
- `AuthorityService`: end-to-end signed far-future bundle dropped after verification — counted under `source="block_bundle"`, never forwarded to core, nothing sent to the shard reconstructor (both fast-sync variants).
- `HeaderSynchronizer`: a far-future fetched header is dropped (counted under `source="header_synchronizer"`) while an in-window header is forwarded to core.
- `Parameters` default snapshot updated for the new field.
- Full `starfish-core` / `starfish-config` lib suites pass; clippy clean.

Fixes #11919

### DIFF
--- a/crates/starfish/core/src/authority_service.rs
+++ b/crates/starfish/core/src/authority_service.rs
@@ -787,30 +787,42 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         }
         self.commit_vote_monitor.observe_block(&verified_block);
 
-        // 5. Drop a far-future primary block before shard/transaction
-        // processing, so it cannot grow the shard reconstructor or DagState
-        // shard state.
-        let far_future_ceiling = self.dag_state.read().peer_disseminated_round_ceiling();
-        if block_ref.round > far_future_ceiling {
-            self.context
-                .metrics
-                .node_metrics
-                .dropped_far_future_blocks_total
-                .with_label_values(&["block_bundle"])
-                .inc();
+        // 5. Bound the far-future parts of the bundle, reusing the block-manager
+        // helper as the synchronizer does. Additional headers above the connect
+        // ceiling are dropped; a far-future primary block is dropped here too, so
+        // it is neither processed for shards (the reconstructor bypasses the
+        // block manager) nor forwarded to the core.
+        additional_block_headers = crate::block_manager::drop_far_future(
+            &self.context,
+            &self.dag_state,
+            additional_block_headers,
+            DataSource::BlockBundleStream,
+            |header| header.round(),
+        );
+        let verified_blocks = crate::block_manager::drop_far_future(
+            &self.context,
+            &self.dag_state,
+            vec![verified_block],
+            DataSource::BlockStreaming,
+            |block| block.round(),
+        );
+        let primary_block_far_future = verified_blocks.is_empty();
+        if primary_block_far_future {
             debug!(
-                "Dropping far-future streamed block {block_ref} from peer {peer}; round {} exceeds ceiling {far_future_ceiling}",
+                "Dropped far-future streamed block {block_ref} from peer {peer}; round {} exceeds the connect ceiling",
                 block_ref.round
             );
-            return Ok(());
         }
 
-        // 6. Collect shards from a bundle and check their proofs.
-
-        let serialized_shards =
-            std::mem::take(&mut serialized_block_bundle_parts.serialized_shards);
-        let verified_shards =
-            self.extract_shards_from_bundle(peer, peer_hostname, serialized_shards, block_ref)?;
+        // 6. Collect shards from a bundle and check their proofs. Skipped for a
+        // far-future primary block so its shards never reach the reconstructor.
+        let verified_shards = if primary_block_far_future {
+            Vec::new()
+        } else {
+            let serialized_shards =
+                std::mem::take(&mut serialized_block_bundle_parts.serialized_shards);
+            self.extract_shards_from_bundle(peer, peer_hostname, serialized_shards, block_ref)?
+        };
 
         // 7. Reject blocks when local commit index is lagging too far from quorum
         //    commit index.
@@ -831,18 +843,21 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
         self.add_digests_to_filter(peer_hostname, &mut additional_block_headers, block_ref)
             .await;
 
-        // 9. Prepare transaction messages for shard reconstructor and send them
-        let transaction_messages = TransactionMessage::create_transaction_messages(
-            &verified_block,
-            &verified_shards,
-            peer.value(),
-        );
-        if let Err(e) = self
-            .transaction_message_sender
-            .send(transaction_messages)
-            .await
-        {
-            warn!("Failed to send transaction messages to shard reconstructor: {e}");
+        // 9. Prepare transaction messages for shard reconstructor and send them.
+        // Skipped for a far-future primary block (no shards were collected).
+        if !primary_block_far_future {
+            let transaction_messages = TransactionMessage::create_transaction_messages(
+                &verified_blocks[0],
+                &verified_shards,
+                peer.value(),
+            );
+            if let Err(e) = self
+                .transaction_message_sender
+                .send(transaction_messages)
+                .await
+            {
+                warn!("Failed to send transaction messages to shard reconstructor: {e}");
+            }
         }
 
         // 10. Add additional headers from bundle to dag, receive missing ancestors for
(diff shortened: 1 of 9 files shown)


piotrm50 (approved):

bingyanglin (approved):

jkrvivian (approved):

crates/starfish/core/src/authority_service.rs
```diff
@@ -778,21 +778,41 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        for block_header in additional_block_headers.iter() {
+            self.commit_vote_monitor.observe_block(block_header);
+        }
+        self.commit_vote_monitor.observe_block(&verified_block);
+
+        // 5. Drop a far-future primary block before shard/transaction
+        // processing, so it cannot grow the shard reconstructor or DagState
+        // shard state.
+        let far_future_ceiling = self.dag_state.read().peer_disseminated_round_ceiling();
+        if block_ref.round > far_future_ceiling {
```
VorobyevIlya: so we drop everything if the block round is too far ahead, even the additional block headers and shards, whose rounds may be ok? It is also never logged, so the corresponding metric may be misleading

crates/starfish/core/src/block_manager/mod.rs
```diff
@@ -168,6 +168,36 @@ impl BlockManager {
         unsuspended_headers
     }
 
+    /// Drops peer-disseminated blocks/headers whose round is too far above the
+    /// accepted frontier to ever connect, bounding the round horizon retained
+    /// in the suspender. Non-peer-disseminated sources pass through unchanged.
```
VorobyevIlya: I think it is redundant since blocks/headers already go through some check before reaching block_manager, but we can leave it as additional guard

crates/starfish/core/src/metrics.rs
```diff
@@ -990,6 +991,12 @@ impl NodeMetrics {
                 "Total number of incoming block headers dropped because their round is at or below the GC floor",
                 registry,
             ).unwrap(),
+            dropped_far_future_blocks_total: register_int_counter_vec_with_registry!(
+                "dropped_far_future_blocks_total",
+                "Number of peer-disseminated blocks dropped because their round is too far above the accepted frontier to ever connect, by enforcement point",
```
VorobyevIlya: blocks or headers, since headers from header synchronizer also go there.

crates/starfish/core/src/header_synchronizer.rs
```diff
@@ -766,6 +768,22 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> HeaderSynchron
                 .join(", "),
         );
 
+        // Drop headers too far above the accepted frontier to ever connect. A
```
VorobyevIlya: can we reuse the function from block manager here?

crates/starfish/core/src/dag_state.rs
```diff
@@ -95,6 +95,24 @@ pub(crate) enum DataSource {
 }
 
 impl DataSource {
+    /// Whether the data was pushed or served by a peer, as opposed to certified
+    /// or locally produced sources.
+    pub(crate) fn is_peer_disseminated(&self) -> bool {
```
VorobyevIlya: misleading name. For example, transactionSynchronizer is also peer disseminated

crates/starfish/core/src/header_synchronizer.rs
polinikita: Yes, let us reuse that one

crates/starfish/core/src/metrics.rs
polinikita: renamed to headers

crates/starfish/core/src/block_manager/mod.rs
polinikita: I wanted to leave that as a guard in case we change something on a path to it

crates/starfish/core/src/authority_service.rs
polinikita: Other things are also dropped. Keep in mind that only headers/shards within cached_round could be sent. This means that the gap between local rounds is at least `peer_round_ahead_margin`, which is set well beyond when the commit syncer kicks in.

PR #11922, opened 2026-06-16
feat(filtered-prometheus): Add prometheus filtering layer

### LINKED ISSUE #8576: Prometheus metrics endpoint exposes millions of high-cardinality time series
## Bug description
On **IOTA TESTNET** when running a validator node, the `/metrics` endpoint generates an extremely large number of metrics (~2.35M lines).  
Most of these are unique due to high-cardinality labels such as `authority`, `peer`, or `path`, combined with histogram buckets (`*_bucket`).  

This causes Prometheus to ingest millions of time series, which quickly leads to:
- High memory usage
- Large on-disk database size
- Slow or failing queries
- Potential Prometheus crashes (OOM)

There doesn’t seem to be any configuration option in `validator.yaml` or the documentation to disable, blacklist, or limit metrics at the node level.

## Rust version
Which version of Rust are you running?

- Rust version: docker version so don't know

## Version
Which version of the IOTA validator node are you running?

- Version number, commit, or branch: iotaledger/iota-node:v1.6.1-rc on docker

## Hardware specification
What hardware are you using?

- Operating system: Ubuntu 24.04 LTS
- RAM: 64 GO
- Cores: 16
- Device: VPS

## Steps To reproduce the bug
1. Run an IOTA validator node with `metrics-address` enabled
2. Query the `/metrics` endpoint
3. Save the output and count the number of lines:
```bash
   curl http://localhost:9184/metrics > metrics.txt
   wc -l metrics.txt
````

4. Observe \~2.35M unique metrics

## Expected behaviour

The node should provide a way to:

* Limit high-cardinality metrics
* Disable or filter certain metrics (e.g. histograms by path/peer/authority)
* Configure whitelists/blacklists in `validator.yaml`

## Actual behaviour

All metrics are always exposed, with no documented way to reduce or filter them.
Prometheus ingestion becomes impractical due to excessive cardinality.

## Errors

No direct errors from the node, but Prometheus is affected:

* Very high memory/disk usage
* Risk of OOM
* Queries become unresponsive

### PR DESCRIPTION
**Pull request by semenov-vladyslav**
# Description of change

This PR provides a drop-in replacement crate - `prometheus-filtered` - for `prometheus` adding metrics filtering capability; it provides thin `Option` wrappers for common prometheus metric types: `Gauge`, `CounterVec`, `Histogram`, etc.. Metrics can be disabled (filtered-out) at run-time during metric registration. Filter can be created explicitly or from env var `METRICS_FILTER`; format: `[on|off][,<prefix>={on|off}]*` where `prefix` is metrics name prefix or rust module name where metric is declared. Disabled metrics' operations are no-ops.

Fixes #8576.

### Release Notes

- [x] Nodes (Validators and Full nodes): Prometheus metrics can now be filtered with `METRICS_FILTER` environment variable.

### CODE REVIEW

File: crates/prometheus-filtered/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,914 @@
...
+    ) -> prometheus::Result<Self> {
+        Ok(Self {
+            inner: prometheus::Registry::new_custom(prefix, labels)?,
+            filter: Arc::new(Filter::from_env()),
+        })
+    }
+
+    /// Creates a registry combining the env-var filter (appended last, highest
+    /// priority) with the supplied config string.
+    pub fn with_filter(filter_str: &str) -> Self {
```
bingyanglin, line 570: The doc says with_filter combines "the env-var filter" with filter_str, but the body only calls Filter::parse(filter_str) and never reads any env var?

thibault-martinez (approved):
Approving dev-tools changes

crates/prometheus-filtered/src/lib.rs
semenov-vladyslav: Updated the comment.

jkrvivian (approved):

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// 3. `module.starts_with(pattern)` — module path prefix.
+    /// 4. `module` contains `"::{pattern}"` — exact module component.
+    #[inline]
+    pub fn is_enabled(&self, name: &str, module: &str) -> bool {
+        let mut result = true;
+        for dir in &self.directives {
+            if dir.pattern.is_empty()
+                || name.starts_with(dir.pattern.as_str())
+                || module.starts_with(dir.pattern.as_str())
+                || module.contains(&format!("::{}", dir.pattern))
```
piotrm50, line 526: This matches any component that just starts with the pattern, so `authority=off` also disables `authority_aggregator`. We should match on full `::` boundaries instead.

Can we add a test for this?

```rust
#[test]
fn module_pattern_respects_component_boundary() {
    let filter = Filter::parse("authority=off");
    assert!(!filter.is_enabled("certs_total", "iota_core::authority"));
    assert!(filter.is_enabled("certs_total", "iota_core::authority_aggregator"));
}
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Drop-in replacement for the `prometheus` crate with optional per-metric
+//! filtering.
+//!
+//! Replace `use prometheus::*` with `use prometheus_filtered::*` and set
+//! `METRICS_FILTER` (or call `Registry::with_config`) to control which metrics
```
piotrm50: `with_config` doesn't exist, this should be `with_filter`.

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+
+    impl_generic_metric_traits!(GenericCounterVec);
+    impl_generic_metric_vec!(GenericCounterVec, GenericCounter);
+
+    pub struct GenericGaugeVec<P: Atomic>(pub(super) Option<prometheus::core::GenericGaugeVec<P>>);
+
+    impl_generic_metric_traits!(GenericGaugeVec);
+    impl_generic_metric_vec!(GenericGaugeVec, GenericGauge);
+
+    pub struct Histogram(pub Option<prometheus::Histogram>);
```
piotrm50: This field is `pub`, but the other wrappers use `pub(super)`. Can we make it `pub(super)` too so we don't leak the `Option`? Same for `HistogramVec`.

piotrm50 (changes requested):

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+
+//! Drop-in replacement for the `prometheus` crate with optional per-metric
+//! filtering.
+//!
+//! Replace `use prometheus::*` with `use prometheus_filtered::*` and set
+//! `METRICS_FILTER` (or call `Registry::with_config`) to control which metrics
+//! are registered.
+//!
+//! Filter syntax: comma-separated `pattern=on|off` directives, last-match
+//! wins.  A bare `off` or `on` sets the global default.  A pattern matches if
```
muXxer:
```suggestion
//! wins. A bare `off` or `on` sets the global default. A pattern matches if
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+struct FilterDirective {
+    /// Empty string means global catch-all.
+    pattern: String,
+    enabled: bool,
+}
+
+/// Parses and evaluates `METRICS_FILTER`-style directives.
+///
+/// Filter string: comma-separated `pattern=on|off`. Bare `on`/`off` is a
+/// global default.  A pattern matches if it is a prefix of the metric name OR
```
muXxer:
```suggestion
/// global default. A pattern matches if it is a prefix of the metric name OR
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+/// [`default_registry`]) so that metrics on the default registry honour the
+/// same filtering as those on explicit registries.
+pub fn default_filter() -> &'static Arc<Filter> {
+    use std::sync::OnceLock;
+    static INSTANCE: OnceLock<Arc<Filter>> = OnceLock::new();
+    INSTANCE.get_or_init(|| Arc::new(Filter::from_env()))
+}
+
+/// Returns a reference to the global default `Registry`, wrapping the
+/// underlying `prometheus::default_registry()`.  Metrics registered here
```
muXxer:
```suggestion
/// underlying `prometheus::default_registry()`. Metrics registered here
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Wrapper macros
+// ---------------------------------------------------------------------------
+//
+// Each macro captures `module_path!()` at the call site so the filter can
+// match by subsystem in addition to metric name.
+//
+// The `$registry` must be a `prometheus_filtered::Registry`.  On success the
```
muXxer:
```suggestion
// The `$registry` must be a `prometheus_filtered::Registry`. On success the
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+        if $crate::default_filter().is_enabled(name, module) {
+            $crate::prometheus::register_counter!(name, $help)
+                .map($crate::core::GenericCounter::new_some)
+        } else {
+            ::std::result::Result::Ok($crate::core::GenericCounter::new_none())
+        }
+    }};
+}
+
+/// register_counter_vec!(name, help, labels)  — global registry, filtered.
```
muXxer:
```suggestion
/// register_counter_vec!(name, help, labels) — global registry, filtered.
```

crates/prometheus-filtered/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+        if ($registry).is_enabled(name, module) {
+            $crate::prometheus::register_gauge_with_registry!(name, $help, ($registry).inner())
+                .map($crate::core::GenericGauge::new_some)
+        } else {
+            ::std::result::Result::Ok($crate::core::GenericGauge::new_none())
+        }
+    }};
+}
+
+/// register_counter!(name, help)  — global prometheus registry, filtered.
```
muXxer:
```suggestion
/// register_counter!(name, help) — global prometheus registry, filtered.
```

muXxer (changes requested):

roman1e2f5p8s (changes requested):
`examples/custom-indexer/rust` is broken now, it does not compile:

```
expected `prometheus_filtered::Registry`, found `prometheus::Registry`
```

crates/prometheus-filtered/src/lib.rs
semenov-vladyslav: This behavior was intentional: `cargo test -- <prefix>` and `RUST_LOG=<prefix>=<level> cargo run` allow to filter by matching not the whole module name but rather the prefix. Just thought similar behavior would be less confusing. To workaround it is possible to specify the filter as follows:
```rs
    #[test]
    fn module_pattern_respects_component_boundary() {
        let filter = super::Filter::parse("authority=off,authority_aggregator=on");
        assert!(!filter.is_enabled("certs_total", "iota_core::authority"));
        assert!(filter.is_enabled("certs_total", "iota_core::authority_aggregator"));
    }
```

crates/prometheus-filtered/src/lib.rs
semenov-vladyslav: Not sure `pub(super)` is even needed there.

roman1e2f5p8s (approved):

crates/prometheus-filtered/src/lib.rs
piotrm50: fair enough.

piotrm50 (approved):

muXxer (approved):

miker83z (approved):

kodemartin (approved):

PR #11923, opened 2026-06-16
feat(snapshot): anchor EPOCH_INFO to signed checkpoint summaries

bingyanglin:
# Description of change

Hardens the formal-snapshot `EPOCH_INFO` trust model. Each entry previously carried `start_system_state` as raw BCS bytes taken on faith from the (unsigned) snapshot transport. Now every field is proven against its certified `last_checkpoint_summary`, and each epoch's start state is derived from the *next* epoch's verified boundary objects (epoch 0 from genesis). A restoring node trusts only data reachable from a signed summary.

- On-disk entry `EpochInfoV1Entry` (moved to `iota-types`): drop `start_system_state`; carry `last_checkpoint_summary`, `last_checkpoint_contents`, `end_of_epoch_tx_effects`, `end_of_epoch_tx_events`, and raw `next_epoch_start_system_state_objects` (`0x5` + inner state). `epoch`/`start_checkpoint` are derived from the signed summaries, not stored.
- Add `verify_epoch_boundary_proof`: hash-chains contents → effects → events → start-state objects back to the signed summary; rejects on any mismatch.
- `verify_epoch_info_chain` now also takes `genesis_system_state` (epoch 0's start, which no entry proves); derives each entry's `epoch`/`start_checkpoint` from the signed summaries and cross-checks the start committee against the certified chain.
- Decode each epoch's `system_state` from its boundary's digest-verified object bytes instead of trusting `start_system_state` (which can't round-trip byte-identically against the effects' object digest).
- `EpochInfoV2` (the `epochs_v2` row, `iota-types`): hold the on-disk entry directly as `epoch_info_entry: Option<EpochInfoV1Entry>` (finalized ⟺ `Some`); store only the non-derivable start-of-epoch identity (`epoch`, `start_checkpoint`, `start_timestamp_ms`, `system_state`) and expose `protocol_version`/`reference_gas_price`/`end_timestamp_ms`/`end_checkpoint` as derived getters.
- `iota-core/grpc_indexes.rs`: at each boundary capture contents, epoch-change effects, and system-state objects into the entry; load events during sparse checkpoint assembly so the entry is complete.
- New helpers (`iota-types`): `CheckpointData::end_of_epoch_transaction` (runtime-checked, not `debug_assert`), `CheckpointContents::end_of_epoch_execution_digests`, `get_iota_system_state_objects`.
- `iota-snapshot/writer.rs`: write the row's `epoch_info_entry` straight into `EPOCH_INFO` (no projection step).
- `iota-grpc-server` `get_epoch`: serve the derived getters instead of the removed stored fields.
- `iota-node` / `iota-tool`: pass `genesis.iota_system_object()` to the verifier.
- `simulacrum`: set `epoch_info_entry` to `None`.

### DIFF
--- a/crates/iota-snapshot/src/lib.rs
+++ b/crates/iota-snapshot/src/lib.rs
@@ -319,7 +319,8 @@ impl EpochInfo {
 /// Chain-verified `EPOCH_INFO`: the `chain_id` matched and every entry was
 /// anchored to its certified `last_checkpoint_summary` — the committee chain
 /// walked from the genesis committee, every byte hash-checked back to that
-/// signed summary. The only constructor is `verify_epoch_info_chain`.
+/// signed summary. The only constructor is `verify_epoch_info_chain`, so holding
+/// one is proof the data is anchored to the operator-provided genesis.
 #[derive(Debug)]
 pub struct VerifiedEpochInfo {
     epoch_info: EpochInfo,
--- a/crates/iota-snapshot/src/lib.rs
+++ b/crates/iota-snapshot/src/lib.rs
@@ -38,24 +38,23 @@ use iota_sdk_types::ObjectId;
 use iota_storage::{
     FileCompression, SHA3_BYTES, compute_sha3_checksum, object_store::util::path_to_filesystem,
 };
+// Re-exported so `iota_snapshot::EpochInfoV1Entry` (the snapshot file entry,
+// defined in `iota-types` because `EpochInfoV2` embeds it) stays a stable path.
+pub use iota_types::storage::EpochInfoV1Entry;
 use iota_types::{
     IOTA_SYSTEM_STATE_OBJECT_ID,
     base_types::ObjectRef,
-    committee::{Committee, CommitteeChainVerifier, EpochId},
+    committee::{Committee, CommitteeChainVerifier},
     digests::ChainIdentifier,
-    effects::{
-        TransactionEffects, TransactionEffectsAPI, TransactionEffectsExt, TransactionEvents,
-    },
+    effects::{TransactionEffectsAPI, TransactionEffectsExt},
     global_state_hash::GlobalStateHash,
     iota_system_state::{
         IotaSystemState, IotaSystemStateTrait,
         epoch_start_iota_system_state::EpochStartSystemStateTrait, get_iota_system_state,
     },
-    messages_checkpoint::{
-        CertifiedCheckpointSummary, CheckpointContents, ECMHLiveObjectSetDigest,
-    },
+    messages_checkpoint::{CheckpointSequenceNumber, ECMHLiveObjectSetDigest},
     object::Object,
-    storage::{CloseOfEpochProof, EpochInfoV2},
+    storage::EpochInfoV2,
 };
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use object_store::path::Path;
@@ -290,71 +289,6 @@ impl Manifest {
     }
 }
 
-/// Per-epoch entry of the `EPOCH_INFO` file. Every field is anchored to the
-/// certified `last_checkpoint_summary` and checked by
-/// `verify_epoch_boundary_proof`, so a restoring node trusts only data
-/// reachable from the signed summary, not the (unsigned) transport.
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct EpochInfoV1Entry {
-    /// Epoch this entry describes.
-    pub epoch: EpochId,
-
-    /// First checkpoint of this epoch (`0` for genesis; otherwise the prior
-    /// epoch's `last_checkpoint_summary.sequence_number + 1`).
-    pub start_checkpoint: iota_types::messages_checkpoint::CheckpointSequenceNumber,
-
-    /// Certified summary of this epoch's closing checkpoint — the signed
-    /// anchor every other field is proven against.
-    pub last_checkpoint_summary: CertifiedCheckpointSummary,
-
-    /// Contents of that closing checkpoint; `hash == last_checkpoint_summary`'s
-    /// `content_digest`.
-    pub last_checkpoint_contents: CheckpointContents,
-
-    /// Effects of the epoch-change tx (the last tx of the closing checkpoint);
-    /// its `(transaction, effects)` digest pair is the last entry of
-    /// `last_checkpoint_contents`.
-    pub end_of_epoch_tx_effects: TransactionEffects,
-
-    /// Events from the epoch-change tx (carries `SystemEpochInfoEvent`); `hash
-    /// == end_of_epoch_tx_effects.events_digest`, or empty on safe-mode
-    /// boundaries where that digest is `None`.
-    pub end_of_epoch_tx_events: TransactionEvents,
-
-    /// Raw serialized bytes of object `0x5` and its inner system-state object
-    /// as written by this boundary — epoch `epoch + 1`'s start state. Each
-    /// object's digest matches a written-object entry in
-    /// `end_of_epoch_tx_effects`.
-    pub next_epoch_start_system_state_objects: Vec<Vec<u8>>,
-}
-
-impl TryFrom<EpochInfoV2> for EpochInfoV1Entry {
-    /// `"proof"` when the row isn't finalized.
-    type Error = &'static str;
-
-    /// Project a finalized index row into its on-disk entry; `Err` if not
-    /// finalized. Destructuring `CloseOfEpochProof` keeps this exhaustive with
-    /// the bundle.
-    fn try_from(row: EpochInfoV2) -> Result<Self, Self::Error> {
-        let CloseOfEpochProof {
-            last_checkpoint_summary,
-            last_checkpoint_contents,
-            end_of_epoch_tx_effects,
-            end_of_epoch_tx_events,
-            next_epoch_start_system_state_objects,
-        } = row.proof.ok_or("proof")?;
-        Ok(EpochInfoV1Entry {
-            epoch: row.epoch,
-            start_checkpoint: row.start_checkpoint,
-            last_checkpoint_summary,
-            last_checkpoint_contents,
-            end_of_epoch_tx_effects,
-            end_of_epoch_tx_events,
-            next_epoch_start_system_state_objects,
-        })
-    }
-}
-
 /// On-disk schema for the per-snapshot `EPOCH_INFO` file. Versioned for
 /// future schema evolution. `entries[i]` is the entry for epoch `i`;
 /// length is `snapshot_epoch + 1`.
@@ -419,11 +353,19 @@ impl VerifiedEpochInfo {
         } = self;
         // `start_system_states[i]` is epoch `i`'s start state; `zip` drops the
         // trailing one (the state the last boundary proves, which has no row).
+        // Each epoch's start checkpoint is the previous epoch's last + 1 (0 for
+        // epoch 0), derived inline from the signed summaries.
+        let mut previous_end_checkpoint: Option<u64> = None;
         epoch_info
             .into_entries()
             .into_iter()
             .zip(start_system_states)
-            .map(|(entry, start_system_state)| epoch_info_v2_row(entry, start_system_state))
+            .map(|(entry, start_system_state)| {
+                let start_checkpoint = previous_end_checkpoint.map_or(0, |seq| seq + 1);
+                previous_end_checkpoint =
+                    Some(*entry.last_checkpoint_summary.data().sequence_number());
+                epoch_info_v2_row(entry, start_system_state, start_checkpoint)
+            })
             .collect()
     }
 
@@ -466,26 +408,14 @@ pub fn verify_epoch_info_chain(
     // `start_system_states[i]` is epoch `i`'s start state; epoch 0's is the
     // genesis root, every later one is derived from the previous boundary.
     let mut start_system_states = vec![genesis_system_state];
-    let mut previous_end_checkpoint: Option<u64> = None;
     for (index, entry) in epoch_info.entries().iter().enumerate() {
         // EPOCH_INFO-specific: entries must be the contiguous epochs from 0.
-        // Everything else (summary epoch, signatures, close-of-epoch,
-        // committee handover) is the chain walk itself.
+        // The epoch comes from the signed summary, not a stored field, so it
+        // can't disagree with the data it anchors.
+        let epoch = entry.last_checkpoint_summary.epoch();
         anyhow::ensure!(
-            entry.epoch == index as u64,
-            "EPOCH_INFO entry at index {index} declares epoch {}",
-            entry.epoch,
-        );
-
-        // The start checkpoint must be the previous epoch's last + 1 (0 for
-        // epoch 0). Cross-checks the unsigned `start_checkpoint` against the
-        // signed sequence numbers.
-        let expected_start = previous_end_checkpoint.map_or(0, |seq| seq + 1);
-        anyhow::ensure!(
-            entry.start_checkpoint == expected_start,
-            "EPOCH_INFO entry for epoch {index} declares start_checkpoint {} but the \
-             previous epoch's signed close implies {expected_start}",
-            entry.start_checkpoint,
+            epoch == index as u64,
+            "EPOCH_INFO entry at index {index} carries a summary for epoch {epoch}",
         );
 
         // Defense in depth: the committee in epoch `index`'s start state must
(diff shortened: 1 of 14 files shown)


crates/iota-types/src/full_checkpoint_content.rs
```diff
@@ -66,6 +66,23 @@ impl CheckpointData {
             .collect()
     }
 
+    /// The transaction that closes this checkpoint's epoch — the `AdvanceEpoch`
+    /// / `advance_epoch_safe_mode` transaction. `None` if this isn't an
+    /// epoch-boundary checkpoint (genesis included), or its last transaction
+    /// unexpectedly isn't an end-of-epoch transaction.
+    pub fn end_of_epoch_transaction(&self) -> Option<&CheckpointTransaction> {
+        self.checkpoint_summary.end_of_epoch_data.as_ref()?;
```
muXxer, line 75: Maybe add a comment on this line. I understand what it does, but it was a bit weird at the beginning :D

crates/iota-types/src/messages_checkpoint.rs
```diff
@@ -529,6 +529,19 @@ impl CheckpointContents {
...
+    /// The `(transaction, effects)` digest pair of the epoch-change
+    /// transaction in an epoch-boundary checkpoint: the last entry, since the
+    /// epoch-change transaction is always ordered last, after every user
+    /// transaction. `None` if `summary` is not an epoch boundary (no
+    /// `end_of_epoch_data`) or the contents are empty.
+    pub fn end_of_epoch_execution_digests(
+        &self,
+        summary: &CheckpointSummary,
+    ) -> Option<&ExecutionDigests> {
+        summary.end_of_epoch_data.as_ref()?;
```
muXxer: same here.

muXxer: Isn't it weird that we pass this summary in here, and use it as a guard? What if someone passes the wrong summary, then this doesn't prevent any mistake. I think the whole helper function is weird. No one really makes sure that those are "end_of_epoch_execution_digests". Why not put this logic on the caller site instead, where we might better know that this is a end of epoch checkpoint?

crates/iota-types/src/iota_system_state/mod.rs
```diff
@@ -295,23 +295,32 @@ impl IotaSystemState {
     }
 }
 
-pub fn get_iota_system_state_wrapper(
+/// The `0x5` object together with the `IotaSystemStateWrapper` decoded from
+/// it. `get_iota_system_state_wrapper` discards the object; callers that need
+/// the raw bytes (e.g. to persist it) keep it.
```
muXxer:
```suggestion
/// The raw system state wrapper object together with the `IotaSystemStateWrapper` decoded from it's contents.
```

crates/iota-types/src/iota_system_state/mod.rs
```diff
@@ -388,6 +397,22 @@ pub fn get_iota_system_state(object_store: &dyn ObjectStore) -> Result<IotaSyste
...
 }
 
+/// The two objects `get_iota_system_state` reads to decode the system state:
+/// the `0x5` wrapper and its inner system-state object. These two fully
+/// determine the state, so none of the per-validator objects the epoch-change
+/// tx also writes are needed. Returned as raw `Object`s so a caller can
+/// persist the exact bytes their `ObjectDigest`s commit to.
+pub fn get_iota_system_state_objects(
+    object_store: &dyn ObjectStore,
+) -> Result<Vec<Object>, IotaError> {
```
muXxer: Would it be cleaner to return a tuple of 2 Objects here instead of a vector?

crates/iota-types/src/iota_system_state/mod.rs
```diff
@@ -388,6 +397,22 @@ pub fn get_iota_system_state(object_store: &dyn ObjectStore) -> Result<IotaSyste
     }
 }
 
+/// The two objects `get_iota_system_state` reads to decode the system state:
+/// the `0x5` wrapper and its inner system-state object. These two fully
+/// determine the state, so none of the per-validator objects the epoch-change
+/// tx also writes are needed. Returned as raw `Object`s so a caller can
+/// persist the exact bytes their `ObjectDigest`s commit to.
```
muXxer:
```suggestion
/// The two objects `get_iota_system_state` reads to decode the system state:
/// the raw system state wrapper object and its inner system-state object. These two fully
/// determine the state, so none of the per-validator objects the epoch-change
/// tx also writes are needed. Returned as raw `Object`s so a caller can
/// persist the exact bytes their `ObjectDigest`s commit to.
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -855,26 +855,80 @@ pub struct EpochInfo {
...
+    pub start_timestamp_ms: u64,
+    /// `IotaSystemState` of object `0x5` at this epoch's start.
     pub system_state: IotaSystemState,
-    /// Certified summary of this epoch's last checkpoint.
-    pub last_checkpoint_summary: Option<CertifiedCheckpointSummary>,
-    /// Events from the AdvanceEpoch tx that closed this epoch.
-    pub end_of_epoch_tx_events: Option<TransactionEvents>,
+    /// Close-of-epoch entry (the same shape written to the snapshot
+    /// `EPOCH_INFO` file); `None` until this epoch's boundary is indexed. The
+    /// row is finalized exactly when this is `Some`.
```
muXxer:
```suggestion
    /// Close-of-epoch entry; `None` until this epoch's boundary is indexed. The
    /// row is finalized exactly when this is `Some`.
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -855,26 +855,80 @@ pub struct EpochInfo {
...
+            .as_ref()
+            .map(|entry| *entry.last_checkpoint_summary.data().sequence_number())
+    }
+}
+
+/// Per-epoch entry of the snapshot `EPOCH_INFO` file, also held by
+/// [`EpochInfoV2`] as its `epoch_info_entry`. Every field is anchored to the
+/// certified `last_checkpoint_summary`. `epoch` and `start_checkpoint` are not
+/// stored — they are derived from the signed summaries.
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
muXxer:
```suggestion
/// Per-epoch entry of the snapshot `EPOCH_INFO` file. Every field is anchored to the
/// certified `last_checkpoint_summary`.
#[derive(Clone, Debug, Serialize, Deserialize)]
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -855,26 +855,80 @@ pub struct EpochInfo {
...
+    /// `last_checkpoint_contents`.
+    pub end_of_epoch_tx_effects: TransactionEffects,
+    /// Events from the epoch-change tx (carries `SystemEpochInfoEvent`); `hash
+    /// == end_of_epoch_tx_effects.events_digest`, or empty on safe-mode
+    /// boundaries where that digest is `None`.
+    pub end_of_epoch_tx_events: TransactionEvents,
+    /// Raw serialized bytes of object `0x5` and its inner system-state object
+    /// as written by this boundary — the next epoch's start state. Each
+    /// object's digest matches a written-object entry in
+    /// `end_of_epoch_tx_effects`.
```
muXxer:
```suggestion
    /// Raw serialized bytes of system-state wrapper object and its inner system-state object
    /// as written by this boundary — the next epoch's start state. Each
    /// object's digest matches a written-object entry in
    /// `end_of_epoch_tx_effects`.
```

crates/iota-types/src/storage/read_store.rs
```diff
@@ -855,26 +855,80 @@ pub struct EpochInfo {
...
+    /// `IotaSystemState` of object `0x5` at this epoch's start.
     pub system_state: IotaSystemState,
-    /// Certified summary of this epoch's last checkpoint.
-    pub last_checkpoint_summary: Option<CertifiedCheckpointSummary>,
-    /// Events from the AdvanceEpoch tx that closed this epoch.
-    pub end_of_epoch_tx_events: Option<TransactionEvents>,
+    /// Close-of-epoch entry (the same shape written to the snapshot
+    /// `EPOCH_INFO` file); `None` until this epoch's boundary is indexed. The
+    /// row is finalized exactly when this is `Some`.
+    pub epoch_info_entry: Option<EpochInfoV1Entry>,
```
muXxer: By reading the code, I have the feeling we should give this field another name, to make sure that this is the info that is closing the epoch and also the proof. The type can stay the same.

crates/iota-snapshot/src/writer.rs
```diff
@@ -562,41 +561,28 @@ impl StateSnapshotWriterV1 {
                          watermark covering it — watermark/row inconsistency"
                     )
                 });
-            let last_checkpoint_summary = epoch_info.last_checkpoint_summary.unwrap_or_else(|| {
+            // The boundary's whole entry is committed in one atomic batch, so a
+            // row the watermark covers must be finalized; a missing entry means
+            // the watermark advanced over an unfinalized row. Panics for the
+            // same blast-radius reason as above.
```
muXxer:
```suggestion
            // the watermark advanced over an unfinalized row. Panics for the
            // same reason as above.
```

crates/iota-snapshot/src/lib.rs
```diff
@@ -37,15 +38,22 @@ use iota_sdk_types::ObjectId;
 use iota_storage::{
     FileCompression, SHA3_BYTES, compute_sha3_checksum, object_store::util::path_to_filesystem,
 };
+// Re-exported so `iota_snapshot::EpochInfoV1Entry` (the snapshot file entry,
+// defined in `iota-types` because `EpochInfoV2` embeds it) stays a stable path.
+pub use iota_types::storage::EpochInfoV1Entry;
```
muXxer: Is this needed? we introduced that type in our feature branch, didn't we?

crates/iota-snapshot/src/lib.rs
```diff
@@ -323,36 +309,25 @@ impl EpochInfo {
...
-/// Chain-verified `EPOCH_INFO`: the snapshot's `chain_id` matched, and every
-/// entry's certified `last_checkpoint_summary` was verified against its
-/// epoch's committee, walking the committee handoffs from the genesis
-/// committee. The only constructor is [`verify_epoch_info_chain`], so holding
-/// one is proof the data is anchored to the operator-provided genesis — the
-/// same trust model as the checkpoint-archive sync.
+/// Chain-verified `EPOCH_INFO`: the `chain_id` matched and every entry was
+/// anchored to its certified `last_checkpoint_summary` — the committee chain
+/// walked from the genesis committee, every byte hash-checked back to that
+/// signed summary. The only constructor is `verify_epoch_info_chain`.
```
muXxer:
```suggestion
/// signed summary. The only constructor is `verify_epoch_info_chain`, so holding
/// one is proof the data is anchored to the operator-provided genesis.
```

crates/iota-snapshot/src/lib.rs
```diff
@@ -323,36 +309,25 @@ impl EpochInfo {
...
 #[derive(Debug)]
 pub struct VerifiedEpochInfo {
     epoch_info: EpochInfo,
     committees: Vec<Committee>,
+    /// Digest-verified start state per epoch: `[i]` is epoch `i`'s start state
+    /// — the genesis root for `i == 0`, else derived from epoch `i - 1`'s
+    /// boundary. Length `snapshot_epoch + 2`; the trailing entry has no row.
+    start_system_states: Vec<IotaSystemState>,
```
muXxer, line 329:
```suggestion
    /// Digest-verified start state per epoch: `[i]` is epoch `i`'s start state
    /// — the genesis root for `i == 0`, else derived from epoch `i - 1`'s
    /// boundary. Length `snapshot_epoch + 2`; the trailing entry has no epoch info row.
    start_system_states: Vec<IotaSystemState>,
```

crates/iota-snapshot/src/lib.rs
```diff
@@ -400,60 +405,132 @@ pub fn verify_epoch_info_chain(
...
+    let summary = entry.last_checkpoint_summary.data();
+
+    // 1. Contents hash to the signed summary.
+    anyhow::ensure!(
+        *entry.last_checkpoint_contents.digest() == summary.content_digest,
+        "last_checkpoint_contents does not hash to the signed content_digest",
+    );
+
+    // 2. The epoch-change effects are the last tx of the verified contents (by
+    //    position, so a same-block tx that also touches `0x5` can't stand in).
```
muXxer:
```suggestion
    // 2. The epoch-change effects are the last tx of the verified contents.
```

crates/iota-snapshot/src/lib.rs
```diff
@@ -400,60 +405,132 @@ pub fn verify_epoch_info_chain(
...
+    let expected = entry
```
muXxer:
```suggestion
    let expected_execution_digest = entry
```

crates/iota-snapshot/src/lib.rs
```diff
@@ -400,60 +405,132 @@ pub fn verify_epoch_info_chain(
...
+        .last_checkpoint_contents
+        .end_of_epoch_execution_digests(summary)
+        .ok_or_else(|| anyhow::anyhow!("the closing checkpoint has no transactions"))?;
+    let effects = &entry.end_of_epoch_tx_effects;
+    anyhow::ensure!(
+        effects.execution_digests() == *expected,
+        "end_of_epoch_tx_effects is not the last transaction of the closing checkpoint",
```
muXxer: Isn't it more that the hash doesn't match?

crates/iota-types/src/full_checkpoint_content.rs
bingyanglin: Added a comment on that line in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d).

crates/iota-types/src/messages_checkpoint.rs
bingyanglin: I removed the whole helper in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d), so this line is gone now.

bingyanglin: Removed the helper and moved the logic to the caller in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d).

crates/iota-types/src/iota_system_state/mod.rs
bingyanglin: Changed it to return `[Object; 2]` instead of a `Vec` in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d). It shows there are always two objects, and the caller can still loop with `.iter()`. A tuple would make the caller unpack and rebuild a list, so I guess array fits better here?

crates/iota-types/src/storage/read_store.rs
bingyanglin: Renamed the field to `epoch_close_proof` in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d).

crates/iota-snapshot/src/lib.rs
bingyanglin: You are right. I removed the re-export in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d). Thanks!

crates/iota-snapshot/src/lib.rs
bingyanglin: Reworded it in [a1e732e](https://github.com/iotaledger/iota/pull/11923/commits/a1e732e30f4808f8e31d2a1dc3b8016680d5049d).

crates/iota-snapshot/src/lib.rs
```diff
@@ -468,12 +465,13 @@ fn verify_epoch_boundary_proof(entry: &EpochInfoV1Entry) -> anyhow::Result<IotaS
     // 2. The epoch-change effects are the last tx of the verified contents.
     let expected_execution_digest = entry
         .last_checkpoint_contents
-        .end_of_epoch_execution_digests(summary)
+        .inner()
+        .last()
```
muXxer, line 469: Is the guard `end_of_epoch_data` now completely gone? Or is it still somewhere called in this call graph? It's hard to see on github UI.

bingyanglin: It is still there but moved one step up, into `verify_epoch_close`. For each entry we call `verify_epoch_close(summary)` just before `verify_epoch_boundary_proof`. And `verify_epoch_close` returns an error if `end_of_epoch_data` is `None`.

PR #11924, opened 2026-06-16
feat(starfish-core): derive fetch response byte limits from per-fetch caps

piotrm50:
# Description of change

Replaces the fixed per-call fetched-bytes cap on the consensus fetch paths with per-endpoint budgets computed from each fetch's count parameters and the maximum serialized size of the items it returns, so the client-side stream-termination limit scales with committee size and the configured per-fetch caps instead of a single hard-coded constant.

Adds two size estimators, each pinned by a test against the BCS wire layout: `max_signed_block_header_bytes(committee_size)` and `max_commit_bytes(committee_size, gc_depth)`. The per-fetch budgets in `tonic_network.rs` combine these with the per-fetch count parameters (`max_headers_per_fetch`, `max_transactions_per_*_sync_fetch`) and the transaction-payload limit.

Builds on the parameters and limits introduced in #11858 (now merged), reusing its `max_headers_per_fetch` selector and `serialized_transactions_size_limit`.

Relates to #11857

### DIFF
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -43,6 +43,58 @@ pub(crate) const GENESIS_ROUND: Round = 0;
 /// Block proposal as epoch UNIX timestamp in milliseconds.
 pub type BlockTimestampMs = u64;
 
+/// BCS-serialized size of a [`BlockRef`]: `round` (u32, 4) + `author`
+/// (`AuthorityIndex`, 1) + `digest` (32). Pinned by
+/// `max_signed_block_header_bytes_bounds_maximal_header`.
+pub(crate) const SERIALIZED_BLOCK_REF_BYTES: usize = 37;
+
+/// ULEB128 byte length of `value`, matching how BCS frames sequence and
+/// variant lengths.
+pub(crate) const fn uleb128_len(mut value: usize) -> usize {
+    let mut len = 1;
+    while value >= 0x80 {
+        value >>= 7;
+        len += 1;
+    }
+    len
+}
+
+/// Upper bound on the BCS-serialized size of a [`SignedBlockHeader`] for a
+/// committee of `committee_size` authorities.
+///
+/// `BlockHeaderV2` is the largest header variant; its size is dominated by two
+/// committee-sized vectors. Layout:
+/// - 55 fixed bytes: `epoch` (8) + `round` (4) + `author` (1) + `timestamp_ms`
+///   (8) + overlap indices (1 + 1) + `transactions_commitment` (32).
+/// - `references`: at most `3 * committee_size` [`BlockRef`]s (37 bytes each)
+///   plus the sequence-length prefix.
+/// - `commit_votes`: at most `committee_size` `CommitVote`s (36 bytes each:
+///   index (4) + digest (32)) plus the sequence-length prefix.
+/// - 34 bytes for `Option<StrongVote>`: `Some` tag (1) + `leader_authority` (1)
+///   + `AuthoritySet` bitmask (32).
+///
+/// The [`SignedBlockHeader`] wrapper adds the `BlockHeader` enum tag (1), the
+/// signature length prefix (1), and the 64-byte signature.
+pub(crate) fn max_signed_block_header_bytes(committee_size: usize) -> usize {
+    const FIXED_HEADER_BYTES: usize = 55;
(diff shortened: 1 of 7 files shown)


filipdulic (approved):

crates/starfish/core/src/block_header.rs
```diff
@@ -43,6 +43,58 @@ pub(crate) const GENESIS_ROUND: Round = 0;
...
+
+/// Upper bound on the BCS-serialized size of a [`SignedBlockHeader`] for a
+/// committee of `committee_size` authorities.
+///
+/// `BlockHeaderV2` is the largest header variant; its size is dominated by two
+/// committee-sized vectors. Layout:
+/// - 55 fixed bytes: `epoch` (8) + `round` (4) + `author` (1) + `timestamp_ms`
+///   (8) + overlap indices (1 + 1) + `transactions_commitment` (32).
+/// - `references`: at most `3 * committee_size` [`BlockRef`]s (37 bytes each)
+///   plus the sequence-length prefix.
```
VorobyevIlya, line 70: So it assumes that block_restrictions flag is on, right? Should we gate the checks then, or is it fine, since we are allowing block restrictions on all networks?

piotrm50: yeah, we should have that assumption, otherwise we could receive and start processing arbitrarily large header from the response.

crates/starfish/core/src/network/fetch_limits.rs
```diff
@@ -0,0 +1,282 @@
...
+pub(crate) const MAX_FETCH_RESPONSE_BUFFER_BYTES: usize = 512 * 1024 * 1024;
+
+/// Streaming budget for a fetch whose response is a single kind of item.
+///
+/// While receiving, reject any element larger than `max_item_bytes` and
+/// terminate the stream once the accumulated bytes exceed `max_total_bytes`.
+pub(crate) struct FetchResponseBudget {
+    /// Upper bound on any single serialized item; a larger item means the peer
+    /// is misbehaving.
+    pub(crate) max_item_bytes: usize,
```
polinikita: This is a local protocol config. Not necessarily misbehaving

crates/starfish/core/src/commit.rs
```diff
@@ -43,6 +43,43 @@ pub(crate) const WAVE_LENGTH: Round = 3;
...
+    if gc_depth == 0 || committee_size == 0 {
+        return usize::MAX;
+    }
+    // `TransactionRef` is round (4) + author (1) + transactions_commitment (32).
+    const SERIALIZED_TRANSACTION_REF_BYTES: usize = 37;
+    // enum tag (1) + index (4) + previous_digest (32) + timestamp_ms (8) +
+    // leader BlockRef (37) + is_optimistic (1).
+    const FIXED_COMMIT_BYTES: usize = 1 + 4 + 32 + 8 + SERIALIZED_BLOCK_REF_BYTES + 1;
+
+    let max_subdag_blocks = gc_depth.saturating_mul(committee_size);
```
polinikita, line 68: I would add .saturating_add(1) - the leader can also be committed.

crates/starfish/core/src/network/tonic_network.rs
```diff
@@ -172,13 +173,20 @@ impl NetworkClient for TonicClient {
             match stream.message().await {
                 Ok(Some(response)) => {
                     for b in &response.vec_serialized_block_header {
```
polinikita, line 205: I would also add a bound for items we receive in the stream. A possible attack. Vector of empty Bytes. Here and in other paths.

polinikita (changes requested):

VorobyevIlya (approved):

crates/starfish/core/src/network/fetch_limits.rs
piotrm50: reworded

crates/starfish/core/src/network/tonic_network.rs
piotrm50: added a limit on number of headers here and in for transactions in fetch_commits_and_transactions. Remaining handlers have proper checks.

crates/starfish/core/src/commit.rs
piotrm50: Good point.

polinikita (approved):

PR #11929, opened 2026-06-16
feat: remove `AsRef<[u8]>` for GenericSignature

thibault-martinez:
# Description of change

Removes the `AsRef<[u8]>` impls — and the backing `OnceCell<Vec<u8>>` "cached bytes" fields — from `GenericSignature`, `PasskeyAuthenticator`, and `MoveAuthenticator`, replacing them with explicit `to_bytes()` / `from_bytes()` methods (and an explicit `EncodeDecodeBase64` impl on `GenericSignature` to replace the fastcrypto blanket impl). Companion change in `iota-rust-sdk` does the same for `MultisigAggregatedSignature` (SDK rev bumped `550fb7b` → `be41523`).

**Why:** the cached-bytes machinery added real complexity — lazily-initialized `OnceCell` fields, hand-written `PartialEq`/`Eq`/`Hash` that delegated to the serialized bytes, and `ToFromBytes` impls — for a cache that, in practice, buys almost nothing.

## Analysis: is dropping the cache safe?

- **Wire-compatible.** Every new `to_bytes()` produces byte-for-byte the same output as the old `as_ref()` (`flag || bcs(..)`). `from_bytes`, the human-readable/binary `Serialize`/`Deserialize` paths, and base64 encoding are unchanged. The transaction wire format does not change.
- **Performance impact is negligible.** The cache only ever helped the rare schemes (Passkey, Move, MultiSig). Plain ed25519/secp256k1/secp256r1 — i.e. essentially all traffic — never had a cache: `as_ref()` was a zero-copy borrow into the stored `[u8; 97/98]` array. For those, the change is one extra ~97-byte alloc+memcpy per serialize (tens of ns, no BCS). For Passkey/Move/MultiSig, each serialize now re-runs a BCS encode (single-digit µs) instead of returning cached bytes. For context, one signature verify is ~30–50 µs (ed25519) to several hundred µs (secp256r1), and the node's dominant cost is BLS authority-signature batch verification (a different type that never had this cache). The lost cache is well under 1% of per-tx work.
- **Where signatures actually get serialized/hashed on receipt.** The `TransactionDigest` (the tx identity, used as the key everywhere) is computed over `TransactionData` only — signatures are never serialized for it, and no `HashMap`/`HashSet` keys on a signature, so the `Hash` impls are effectively never hit on the hot path. A signature is BCS-serialized ~2× for an owned-object tx (verify-cache-key digest + RocksDB write) and ~4× across a shared-object cert's lifecycle (cert digest, full-message digest, consensus submission, checkpoint contents) — each bounded per-node by the existing verification caches. Verification crypto itself reads raw sig/pk slices and does not serialize the `GenericSignature`.

- [ ] I have checked that new and existing unit tests pass locally with my changes — ⚠️ `as_ref_is_cached` currently fails (see Outstanding above)


crates/iota-types/src/move_authenticator.rs
```diff
@@ -146,76 +158,27 @@ impl AuthenticatorTrait for MoveAuthenticator {
 /// MoveAuthenticator level.
 impl Hash for MoveAuthenticator {
```
DaughterOfMars, line 154: I think we should use the derived hash, shouldn't be any difference

DaughterOfMars: I'm not 100% sure though

crates/iota-types/src/move_authenticator.rs
```diff
@@ -146,76 +158,27 @@ impl AuthenticatorTrait for MoveAuthenticator {
...
 }
 
-/// Necessary trait for
-/// [SenderSignerData](crate::transaction::SenderSignedData). This trait is
-/// implemented only for MoveAuthenticator and not for specific versions of
-/// MoveAuthenticator (e.g., MoveAuthenticatorV1) because the custom
-/// serialization/deserialization signature logic is defined at the
-/// MoveAuthenticator level.
-impl Eq for MoveAuthenticator {}
+// TODO actually different than derived?
```
DaughterOfMars: Should be identical in function

crates/iota-types/src/move_authenticator.rs
thibault-martinez: to_bytes is different than what Hash would otherwise use, I think

DaughterOfMars: Yes, it is, but I think that the result would be the same, no?

thibault-martinez: No, to_bytes pushes a SignatureScheme, the derived trait has no way to know what

DaughterOfMars: Yes but the signature scheme is identical for all instances of this type

thibault-martinez: Unless someone expert tells us how this hash ends up being used, I wouldn't touch it. I don't know if anything stores it, uses it later, etc

piotrm50 (approved):

crates/iota-types/src/move_authenticator.rs
DaughterOfMars: Well the purpose of the hash is to create unique hash values for each unique instance, and that happens with or without using the bytes

thibault-martinez: Deriving it would lead to a different hash for the exact same instance

DaughterOfMars: yes but that's not important (afaik) because this hash is only used for rust stuff, not for protocol hashing and such

thibault-martinez: Yeah that's what I'm saying, I don't know if we're knowledgeable enough to take this decision (I'm definitely not)

DaughterOfMars: definitely good to get an expert opinion

crates/iota-types/src/move_authenticator.rs
```diff
@@ -120,6 +111,27 @@ impl MoveAuthenticator {
     pub fn validity_check(&self, config: &ProtocolConfig) -> UserInputResult {
         self.inner.validity_check(config)
     }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let as_bytes = bcs::to_bytes(&self.inner).expect("BCS serialization should not fail");
+        let mut bytes = Vec::with_capacity(1 + as_bytes.len());
+        bytes.push(SignatureScheme::MoveAuthenticator.flag());
+        bytes.extend_from_slice(as_bytes.as_slice());
```
semenov-vladyslav:
```suggestion
        let mut bytes = vec![];
        bytes.push(SignatureScheme::MoveAuthenticator.flag());
        bcs::serialize_into(&mut bytes, &self.inner).expect("BCS serialization should not fail");
```

crates/iota-types/src/move_authenticator.rs
```diff
@@ -120,6 +111,27 @@ impl MoveAuthenticator {
...
+        // The first byte matches the flag of MoveAuthenticator.
+        if bytes.first().ok_or(FastCryptoError::InvalidInput)?
+            != &SignatureScheme::MoveAuthenticator.flag()
+        {
+            return Err(FastCryptoError::InvalidInput);
+        }
+
+        let inner: MoveAuthenticatorInner =
+            bcs::from_bytes(&bytes[1..]).map_err(|_| FastCryptoError::InvalidSignature)?;
+        Ok(Self { inner })
```
semenov-vladyslav:
```suggestion
        match bytes.split_first() {
            Some((first, tail)) if first == &SignatureScheme::MoveAuthenticator.flag() => {
                let inner: MoveAuthenticatorInner =
                    bcs::from_bytes(tail).map_err(|_| FastCryptoError::InvalidSignature)?;
                Ok(Self { inner })
            },
            _ => Err(FastCryptoError::InvalidInput),
        }
```

crates/iota-types/src/passkey_authenticator.rs
```diff
@@ -203,22 +198,40 @@ impl PasskeyAuthenticator {
...
+        // The first byte matches the flag of PasskeyAuthenticator.
+        if bytes.first().ok_or(FastCryptoError::InvalidInput)?
+            != &SignatureScheme::PasskeyAuthenticator.flag()
+        {
+            return Err(FastCryptoError::InvalidInput);
+        }
+        let passkey: PasskeyAuthenticator =
+            bcs::from_bytes(&bytes[1..]).map_err(|_| FastCryptoError::InvalidSignature)?;
+        Ok(passkey)
     }
```
semenov-vladyslav: Same could be applied here.

semenov-vladyslav (approved):

crates/iota-types/src/move_authenticator.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/11929/commits/43d5d0787783c724efb82e04909d67d6a181619c

crates/iota-types/src/move_authenticator.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/11929/commits/43d5d0787783c724efb82e04909d67d6a181619c

crates/iota-types/src/passkey_authenticator.rs
thibault-martinez: This was moved to the SDK so I did it there https://github.com/iotaledger/iota-rust-sdk/pull/1227/changes/9157dfd8ae222bf9b1aad05f38c0f965ed3e1a07

crates/iota-types/src/move_authenticator.rs
thibault-martinez: Will chat with some people about it later https://github.com/iotaledger/iota/issues/11957

crates/iota-types/src/move_authenticator.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/11929/commits/0046dadee42823526954c359b3d1ccc9b563b0d6

DaughterOfMars (approved):
LGTM once the commented stuff is removed

Thoralf-M (approved):

bingyanglin (approved):

tomxey (approved):

PR #11944, opened 2026-06-16
feat(iota-rust-sdk): add `iota-rust-sdk` re-exports

muXxer:
# Description of change

Re-export the `iota-rust-sdk` crates this repo pins (`iota-sdk-types`, `iota-sdk-crypto`, the gRPC client/types, the GraphQL client, and the transaction builder) from `iota-sdk`, behind feature flags that mirror the upstream `iota-rust-sdk` meta-crate.

A downstream repo that links against this repo and `iota-rust-sdk` directly risks ending up with two incompatible copies of e.g. `iota-sdk-types`. Cargo only unifies deps resolving to the same git rev, so a rev mismatch yields two distinct `ObjectId` types that won't interoperate.

By re-exporting, consumers can reach these types through `iota_rust_sdk::*` and drop their own `iota-rust-sdk` git dependency entirely, getting exactly the rev `iota-rust-sdk` pins. No more manual rev-matching.

##  Details
  
- Features mirror upstream (crypto/ed25519/…, types/serde/rand/…, grpc, graphql, txn-builder) so consumers can enable any underlying feature through iota-sdk alone.
- default keeps the build light: types + crypto on; grpc/graphql/txn-builder/bls12381/proptest are opt-in (the first three pull extra git deps).
- transaction_builder re-exports iota-rust-sdk's iota-sdk-transaction-builder, distinct from the local iota-transaction-builder the client uses internally.
- types is the one divergence from upstream: iota-sdk-types stays non-optional (the client uses it internally), so the feature gates only the re-export.

  ▎ Note for consumers: to actually get version-matching, reach these crates only via iota_rust_sdk::* — don't also add a direct iota-rust-sdk git dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1651,6 +1651,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
+name = "bech32"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
+
+[[package]]
 name = "beef"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1780,8 +1786,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90dbd31c98227229239363921e60fcf5e558e43ec69094d46fc4996f08d1d5bc"
 dependencies = [
  "bitcoin_hashes",
- "rand 0.7.3",
- "rand_core 0.5.1",
+ "rand 0.8.6",
+ "rand_core 0.6.4",
  "serde",
  "unicode-normalization",
 ]
@@ -3892,7 +3898,7 @@ dependencies = [
  "ark-serialize",
  "auto_ops",
  "base64ct",
- "bech32",
+ "bech32 0.9.1",
  "bincode",
  "blake2",
  "blst",
@@ -7389,6 +7395,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "iota-rust-sdk"
+version = "1.26.0-alpha"
+dependencies = [
+ "iota-sdk-crypto",
+ "iota-sdk-graphql-client",
+ "iota-sdk-grpc-client",
+ "iota-sdk-grpc-types",
+ "iota-sdk-transaction-builder",
+ "iota-sdk-types",
+]
+
+[[package]]
 name = "iota-sdk"
 version = "1.26.0-alpha"
 dependencies = [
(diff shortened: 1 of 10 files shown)


crates/iota-rust-sdk/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2025 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-rust-sdk/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+authors = ["IOTA Foundation <info@iota.org>"]
+edition = "2021"
+license = "Apache-2.0"
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+iota-graphql-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", package = "iota-sdk-graphql-client", optional = true }
```
Thoralf-M: `, package = "iota-sdk-graphql-client"`
really needed? You don't have it for the iota-sdk-transaction-builder

crates/iota-rust-sdk/Cargo.toml
Thoralf-M, line 1: Can you also add this here https://github.com/iotaledger/iota/blob/3af22008a14ac32bcaaf6c3c156810a538ecd70f/scripts/update_sdk_types_rev.sh#L71-L74 ?

crates/iota-rust-sdk/Cargo.toml
Thoralf-M: I think like this it should also work
```suggestion
iota-sdk-graphql-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", optional = true }
```

scripts/update_sdk_types_rev.sh
```diff
@@ -56,20 +56,23 @@ update_rev() {
...
 
-# Update both iota-sdk-types and iota-sdk-crypto in the given Cargo.toml.
+# Update every pinned iota-rust-sdk crate in the given Cargo.toml.
 update_file() {
     local file="$1"
     update_rev iota-sdk-types "$file"
     update_rev iota-sdk-crypto "$file"
     update_rev iota-sdk-grpc-types "$file"
     update_rev iota-sdk-grpc-client "$file"
+    update_rev iota-graphql-client "$file"
```
Thoralf-M:
```suggestion
    update_rev iota-sdk-graphql-client "$file"
```
?

crates/iota-rust-sdk/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+[lints]
+workspace = true
+
+[dependencies]
+iota-grpc-client = { workspace = true, optional = true }
+iota-grpc-types = { workspace = true, optional = true }
+iota-sdk-crypto = { workspace = true, optional = true, features = ["ed25519"] }
+iota-sdk-graphql-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", optional = true }
+iota-sdk-transaction-builder = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", default-features = false, optional = true }
+iota-sdk-types = { workspace = true, optional = true, features = ["rand"] }
```
Thoralf-M: Should be only added via the feature flags below, otherwise they can't be disabled
```suggestion
iota-sdk-crypto = { workspace = true, optional = true }
iota-sdk-graphql-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", optional = true }
iota-sdk-transaction-builder = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "550fb7b80a15fa2316ec09254884ef243f8688ea", default-features = false, optional = true }
iota-sdk-types = { workspace = true, optional = true }
```

Thoralf-M (approved):

bingyanglin (approved):

scripts/cargo_sort/cargo_sort.py
```diff
@@ -1699,9 +1699,11 @@ def run_consolidate_mode_with_loop(args, target_dir: str, root_cargo_toml: str,
 
     # add special cases
     internal_crates_dict["iota-flamegraph-svg"] = None
+    internal_crates_dict["iota-graphql-client"] = None
```
bingyanglin: Do we have this workspace?

jkrvivian (approved):

Cargo.lock
```diff
@@ -11373,7 +11433,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "27c6023962132f4b30eb4c172c91ce92d933da334c59c23cddee82358ddafb0b"
 dependencies = [
  "anyhow",
- "itertools 0.14.0",
+ "itertools 0.10.5",
```
semenov-vladyslav, line 11436: Weird downgrade

Cargo.lock
```diff
@@ -15206,7 +15266,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
semenov-vladyslav, line 15269: Another weird downgrade.

semenov-vladyslav (approved):

piotrm50 (approved):

Cargo.lock
muXxer: I checked it again, it always happens with a "cargo check" automatically.

Cargo.lock
muXxer: I checked it again, it always happens with a "cargo check" automatically.

miker83z (approved):

PR #11951, opened 2026-06-17
feat(iota-move-natives): add protocol_config::get_attr<T> native for reading protocol parameters in Move.

### LINKED ISSUE #11884: [move][framework] Extend protocol_config native to read protocol config parameters in Move
## Description

PR https://github.com/iotaledger/iota/pull/11443 introduced `iota::protocol_config::is_feature_enabled(feature_flag_name: vector<u8>): bool`, a `public(package)` native function that lets framework Move code query protocol **feature flags** at runtime.

We also need access to protocol config **parameters** (the typed config values such as `max_arguments`, `max_move_object_size`, gas-schedule values, etc.).

## Current state

- The Rust side already exposes everything we need: the `ProtocolConfigAccessors` derive macro (`iota-protocol-config-macros`) generates `ProtocolConfig::lookup_attr(name: String) -> Option<ProtocolConfigValue>`, where `ProtocolConfigValue` is an enum encoding the field's type.
- Current parameter types in `ProtocolConfig`: `u64` (majority), `u32`, `u16`, and `bool`. No floating-point fields exist today.
- The existing native gets the `ProtocolConfig` from the `ObjectRuntime` extension, so the same access path can be reused.
- There is precedent for a generic native returning `Option<T>`: `iota::config::read_setting_impl` dispatches on its type arguments in the native and constructs the `Option` value in Rust (crates/iota-framework/packages/iota-framework/sources/config.move#L264-L273).

## Proposed solution

Add a single generic native function to `iota::protocol_config`:

```move
/// Returns the value of a protocol config parameter.
/// Returns `none` if the parameter is not defined in the current
/// protocol version, or if `T` does not match the parameter's type.
public(package) native fun get_attr<T: copy + drop + store>(param_name: vector<u8>): Option<T>;
```

The native implementation:
1. Looks up the parameter via `ProtocolConfig::lookup_attr`.
2. Matches the requested type argument (`Type::U16` / `Type::U32` / `Type::U64` / `Type::Bool`) against the returned `ProtocolConfigValue` variant.
3. On match, returns `some(value)`; on missing parameter or type mismatch, returns `none` (with a `debug_assert!`, consistent with `is_feature_enabled`'s forward-compatibility convention — a framework upgrade may ship a parameter read before the binary defines the parameter).
4. Any non-primitive or unsupported `T` returns `none` (or aborts — see open questions).
5. **Unsupported `T`.** A struct or unsupported primitive as `T` cannot be expressed as a config value — decide between returning `none` and aborting with a dedicated error code.
6. **Gas cost.** `is_feature_enabled` is zero-cost since it's framework-internal; presumably the same applies here.

Dkwcs:
# Description of change

Adds a generic `get_attr<T>` native function to the `iota::protocol_config` framework module, extending the existing is_feature_enabled API to cover numeric and boolean protocol parameters (not just feature flags, but config protocol params).

This PR fills the remaining gap: reading u16, u32, u64, and bool config parameters by name. The native delegates to the auto-generated ProtocolConfig::lookup_attr accessor and wraps the result in T, abort when:
- the parameter is absent at the current protocol version (deprecated or not yet introduced),
- the requested type T does not match the parameter's actual type.

Fixes #11884.

Six new Move unit tests in protocol_config_tests.move cover:

- Some(u64) read (max_tx_size_bytes)
- Some(u32) read (max_arguments)
- Some(u16) read (binary_module_handles)

### Release Notes

- [x] Protocol: Added get_attr<T: copy + drop + store>(param_name: vector<u8>): T to the iota::protocol_config framework module (protocol version 30).
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-framework/packages/iota-framework/tests/protocol_config_tests.move
```diff
@@ -19,3 +19,49 @@ fun test_is_feature_enabled_false() {
...
+    let val: Option<u32> = protocol_config::get_attr(b"max_arguments");
+    assert_eq(val, option::some(512u32));
+}
+
+#[test]
+// binary_module_handles is a u16 set to 100 since protocol v1.
+fun test_get_attr_u16() {
+    let val: Option<u16> = protocol_config::get_attr(b"binary_module_handles");
+    assert_eq(val, option::some(100u16));
+}
```
valeriyr, line 44: to cover all the possible return types, it would be nice to add one more test: test_get_attr_bool

Dkwcs: as I see we don’t have relevant bool param in config

lollobene: have you considered `bridge_should_try_to_finalize_committee : Option<bool>` ?

iota-execution/latest/iota-move-natives/src/protocol_config.rs
```diff
@@ -68,3 +69,82 @@ pub fn is_feature_enabled(
...
+                        false,
+                        "get_attr: type mismatch for protocol config parameter"
+                    );
+                    option_none(ty)?
+                }
+            }
+        }
+
+        // Invalid UTF-8 parameter names are treated as missing.
+        Err(_) => option_none(ty)?,
```
lollobene: Maybe return an error here instead of returning None?

crates/iota-framework/packages/iota-framework/tests/protocol_config_tests.move
Dkwcs: @lollobene this param is deprecated 
```
    /// Deprecated because of bridge removal.
    /// Whether to try to form bridge committee
    // Note: this is not a feature flag because we want to distinguish between
    // `None` and `Some(false)`, as committee was already finalized on Testnet.
    bridge_should_try_to_finalize_committee: Option<bool>,
```
I even have the tests that uses this param
```
#[test]
// bridge_should_try_to_finalize_committee was deprecated to None at protocol v9.
// At max protocol version it is not defined, so the native must return none.
fun test_get_attr_deprecated_to_none() {
    let val: Option<bool> = protocol_config::get_attr(b"bridge_should_try_to_finalize_committee");
    assert_eq(val, option::none());
}
```

iota-execution/latest/iota-move-natives/src/protocol_config.rs
Dkwcs: It’s discussable, it was an open question in the issue.
I was trying to follow the existing logic, such as `is_feature_enabled`, which does not return the error. I assume this also helps with backward compatibility

Imagine the framework at v30 releases this Move code:
`
let limit: u64 = get_attr(b"new_tx_limit"); // this param only exists in v30 binary
`
If `get_attr` aborted on missing, any validator running a v29 binary that re-executes a v30 checkpoint would abort/fail the transaction. That is a consensus breaking piece - the v30 binary says the transaction succeeded, the v29 binary says it aborted, however with option, the v29 binary returns none everything should work.

lollobene: makes totally sense

valeriyr: `is_feature_enabled` contains `debug_assert!(false);` in this case.
The question is related not to an absent parameter but to the `String::from_utf8` error branch, right?

valeriyr: Actually, `debug_assert!(false)` is present for a not-found parameter as well in `is_feature_enabled`. 
Validator on version 29 can't use a parameter that was added in version 30 since it uses the framework version 29.

muXxer (approved):

muXxer:
If it doesn't get merged before todays cut, we need to bump the protocol version.

iota-execution/latest/iota-move-natives/src/protocol_config.rs
```diff
@@ -68,3 +69,79 @@ pub fn is_feature_enabled(
...
 }
+
+/// Abort code returned when the parameter name is not valid UTF-8.
+const E_INVALID_UTF8_PARAM_NAME: u64 = 0;
+/// Abort code returned when the parameter is absent in the current protocol
+/// version.
+const E_PARAM_NOT_FOUND: u64 = 1;
+/// Abort code returned when the requested Move type does not match the actual
+/// parameter type stored in the protocol config.
+const E_TYPE_MISMATCH: u64 = 2;
```
miker83z: Move these at the top of the file

miker83z (approved):

crates/iota-framework/packages/iota-framework/tests/protocol_config_tests.move
```diff
@@ -19,3 +19,26 @@ fun test_is_feature_enabled_false() {
     let is_enabled = protocol_config::is_feature_enabled(b"consensus_smart_ancestor_selection");
     assert_eq(is_enabled, false);
 }
+
+// --- get_attr tests ---
```
valeriyr, line 23: Why have you removed these tests?

#[test]
#[expected_failure(abort_code = 0, location = iota::protocol_config)]
// A non-UTF-8 byte sequence as the parameter name is a programming error and must abort.
fun test_get_attr_invalid_utf8() {
    let _: u64 = protocol_config::get_attr(x"ff");
}

#[test]
#[expected_failure(abort_code = 1, location = iota::protocol_config)]
// An unknown parameter name is a programming error and must abort.
fun test_get_attr_unknown_param() {
    let _: u64 = protocol_config::get_attr(b"nonexistent_parameter_name");
}

#[test]
#[expected_failure(abort_code = 2, location = iota::protocol_config)]
// max_arguments is a u32; requesting it as u64 is a programming error that must abort.
fun test_get_attr_type_mismatch() {
    let _: u64 = protocol_config::get_attr(b"max_arguments");
}

#[test]
#[expected_failure(abort_code = 1, location = iota::protocol_config)]
// bridge_should_try_to_finalize_committee was deprecated to None at protocol v9;
// requesting it is a programming error and must abort.
fun test_get_attr_deprecated_to_none() {
    let _: bool = protocol_config::get_attr(b"bridge_should_try_to_finalize_committee");
}

Dkwcs: Currently we have debug_assert for all possible negative branches.
Anything that could fire debug_assert, causes tests failing in the CI and must not be part of the tests

iota-execution/latest/iota-move-natives/src/protocol_config.rs
Dkwcs: Done

crates/iota-framework/packages/iota-framework/tests/protocol_config_tests.move
valeriyr: If we abort the execution, I would consider removing debug asserts

Dkwcs: As long as it’s only us users, I’m not sure it’s critical @miker83z

valeriyr: The problem is that we have two different error handling options in debug and release.
If we suppose the aborts never fire, we can use expects or `assert` which work in the release configuration as well.
If we suppose the function can abort, we can use this during implementation instead asserts and test the expected behavior.

miker83z: Let's follow @valeriyr's idea

Dkwcs: Done

valeriyr (approved):

kodemartin (approved):

PR #11956, opened 2026-06-17
upstream(core): port ExecutionScheduler alongside TransactionManager

### LINKED ISSUE #11901: Node: add transaction scheduler
We should port the nice refactor that upstream did. We need to take care that it doesn't conflict with our p-cool efforts.

* https://github.com/MystenLabs/sui/commit/0b711bb247e3fed4304cfa9de8dd6631062c2813

chriscandless:
# Description of change

Port the upstream `ExecutionScheduler` (MystenLabs/sui#22009) alongside the existing `TransactionManager`, both behind an `enum_dispatch` `ExecutionSchedulerWrapper`. `TransactionManager` is moved into a new `execution_scheduler` module with unchanged behavior and **remains the default**; `ExecutionScheduler` is opt-in via the `ENABLE_EXECUTION_SCHEDULER` env var. With the default, production behavior is unchanged.

Why: keep pace with upstream's post-consensus execution-scheduling refactor while ensuring it does not conflict with IOTA's P-COOL (post-consensus owned-object locking) flow. The two are cleanly layered — the scheduler is a downstream consumer of post-consensus survivors — and this is verified end-to-end against both schedulers.

## Intentional divergences from upstream (SUI #22009)

This is a faithful port, but a few places deliberately depart from the upstream diff. Each is called out so reviewers comparing against #22009 know the difference is intentional, not an oversight.

1. **`ExecutingGuard` is held through execution (overload accounting).** Upstream drops the executing guard at dispatch, so its pending count excludes in-flight execution. Here the guard is moved into the `PendingTransaction` and held by the execution-driver task until execution completes (or the epoch is cancelled), so `num_pending_transactions()` counts pending **and** executing. IOTA's legacy `TransactionManager` counted executing transactions (`inflight_queue_len = pending + executing`), and the overload-admission threshold is tuned against that semantics; matching it keeps the scheduler swap behavior-preserving for load shedding. A pending-only count would silently admit more load for the same config value. Pinned by `execution_scheduler_counts_executing_transaction` and `execution_scheduler_drops_executing_guard_on_epoch_termination`.

2. **Cache availability reconciled onto IOTA's production check, not upstream's rewrite.** Upstream rewrites `try_multi_input_objects_available` with a `partition_map`/`Either` split over `VersionedObject`/`Package`. Here we keep IOTA's production body (cancelled-sentinel filtering, the receiving + deleted-owned fallback, and deleted-shared marker handling using IOTA marker names) and only rewire `notify_read_input_objects` to delegate to it; the earlier drifted `notify_read_input_objects_impl` free function is deleted. This preserves IOTA-specific availability correctness and avoids a functional rewrite (with its deref-level and marker-name hazards) for no behavioral gain. It also closes a latent hang: the drifted impl's receiving branch lacked the `have_deleted_owned_object_at_version_or_after` fallback, so a received-then-deleted owned input — a case P-COOL can produce — could wait forever. Pinned by the cache-only-contract, received-then-deleted, and wait-for-all-inputs tests.

3. **Scheduler selector: a named default constant plus env overrides, honored in every build.** Upstream selects via a `cfg(test)` coin-flip (`gen_bool(0.5)`) plus env. Here a single `DEFAULT_USE_EXECUTION_SCHEDULER` constant (currently `false` = `TransactionManager`) is combined with `ENABLE_TRANSACTION_MANAGER` (opt-out, takes precedence) and `ENABLE_EXECUTION_SCHEDULER` (opt-in), honored in all builds so the whole suite can pin either scheduler deterministically. A proper node-config selector is future work.

**Behavioral note (a `TransactionManager`-vs-`ExecutionScheduler` difference, not a divergence from upstream):** `ExecutionScheduler` registers per-object pending state asynchronously inside the per-transaction task (via `PendingGuard`) where `TransactionManager` does it synchronously at enqueue; one overload test was made scheduler-agnostic (polling) accordingly.

**Further `TransactionManager`-vs-`ExecutionScheduler` differences, all shared with upstream and left as-is:**

- **Per-object overload counts a different population.** `TransactionManager` registers a waiting transaction under every *missing* input's object id, including read-only shared inputs; `OverloadTracker` registers under *mutable shared* object ids only, whichever input was actually missing. So the scheduler under-counts transactions waiting read-only on a hot object and over-counts transactions holding it mutably while stalled on an unrelated owned input. This feeds `max_transaction_manager_per_object_queue_length`, which is tuned against the `TransactionManager` semantics.
- **`transaction_manager_num_missing_objects` is not populated** under the scheduler, which keeps no central registry of missing inputs. Unlike the ready-rate and dispatch-queue metrics above, nothing reads it to make a decision and nothing decrements it, so it reads flat rather than wrong. Upstream deleted the metric along with the `TransactionManager`.
- **A duplicate enqueue is counted twice while it settles.** The scheduler spawns one task per enqueued transaction with no dedup, so the same digest briefly holds two `ExecutingGuard`s; `TransactionManager` dedups at enqueue. It converges to zero (covered by `duplicate_enqueue_executes_once`), but `num_pending_transactions()` double-counts in the window.

**Deliberate lint exception (a deviation from the repo convention, and it matches upstream):** `ExecutionSchedulerWrapper` carries `#[allow(clippy::large_enum_variant)]`. We are aware of the repo guideline against silencing lints with `#[allow(...)]` and are making a conscious, owned exception: there is exactly one `ExecutionSchedulerWrapper` per node and it lives behind an `Arc`, so the variant-size difference has no practical cost, and keeping `TransactionManager` unboxed matches upstream exactly (upstream does not box `Inner` either). Boxing solely to satisfy the lint would diverge from upstream for no real benefit. Happy to switch to boxing if reviewers prefer.

## A scheduler bug found and fixed during testing

Running the full suite under `ENABLE_EXECUTION_SCHEDULER=1` surfaced one real bug, fixed in this PR. The scheduler computed a transaction's input dependencies from `transaction_data().input_objects()`, which omits the objects read by the transaction's `MoveAuthenticator`s (abstract-account authenticators). Execution reads the full set via `collect_all_input_object_kind_for_reading()`, so the scheduler could report an abstract-account transaction ready and dispatch it before an authenticator's shared input (e.g. the `Clock`) was available, panicking in the input loader. The scheduler now schedules on `collect_all_input_object_kind_for_reading()` so the readiness check covers exactly what execution reads. The legacy `TransactionManager` is not affected: it computes readiness from the envelope's `SenderSignedData::input_objects()`, which already includes every authenticator's inputs.

## A defensive notify fix in the input-object wait path

Review of the notify path surfaced a related gap in `notify_marker_written`: it woke input-object waiters only for `SharedDeleted` markers, not `OwnedDeleted`. An `OwnedDeleted` marker also satisfies availability for a received-then-deleted owned object used as a *receiving* input (via `multi_input_objects_available`), and a deleted object is never written at that version — so no object-write notification ever arrives. A waiter that registered before such a marker was written would therefore never be woken, and the `ExecutionScheduler` — which relies solely on this notify path — could wait forever on that input. A hung transaction is a checkpoint root, so this would stall checkpoint creation. (`TransactionManager` does not use this path; it tracks missing inputs and is woken via `notify_commit`.)

This is a **defensive correctness fix, not a reachable bug**. A live transaction cannot reach the state: a receiving reference is pinned at signing to a currently-live object version whose object-write already notifies, and consensus ordering places that write before any later deletion marker — so the marker is never the sole wakeup for a registered waiter, and the scheduler is off by default regardless. Upstream SUI `main` has the identical gap. Taken as hardening before the scheduler is ever enabled. Pinned by `test_writeback_wait_for_owned_deleted`, which writes the marker after the wait begins and times out without the one-line change.

## Ready-transaction accounting restored under the scheduler

Review also found that `ExecutionScheduler::send_transaction_for_execution` omitted the metric accounting `TransactionManager::transaction_ready` performs. With the scheduler active:

- `txn_ready_rate_tracker` stayed at zero. The overload monitor derives latency-based load shedding from that rate and returns 0% when it is zero, so that defense was silently off (queue-length shedding was unaffected). This is the same overload behavior the `ExecutingGuard` divergence above exists to preserve.
- `execution_driver_dispatch_queue` counted into negative numbers — the execution driver decrements it for every transaction it receives regardless of scheduler, and only the `TransactionManager` incremented it — and `transaction_manager_num_ready` never moved.

Fixed by mirroring the three lines in `send_transaction_for_execution`. `schedulers_record_ready_transaction_accounting` runs **both** schedulers through the same assertions and fails without the change. The gap slipped through because no test asserted on these metrics, and the overload-monitor tests exercise the shedding math with synthetic rates rather than the wiring that produces them.

Inherited from upstream SUI, where it is live rather than dormant: `TransactionManager` was deleted there, nothing replaced the accounting, and `ExecutionScheduler` is the only scheduler.

fixes #11901

Validated under **both** schedulers (default `TransactionManager` and `ENABLE_EXECUTION_SCHEDULER=1`). Standalone-scheduler unit tests cover input readiness (wait for **all** missing inputs, fan-out release), effects-digest propagation on both the fast and wait dispatch paths, wrong-epoch filtering, and reconfiguration clearing the pending/overload state. Cache-level tests cover the cache-only availability contract (no store reads) and the received-then-deleted owned-input resolution with a negative control. Reconfiguration cleanup is covered at the unit level, where the scheduler-specific mechanism lives (`within_alive_epoch` cancellation dropping the pending/executing guards and clearing the overload tracker). Authority- and e2e-level tests exercise the post-consensus P-COOL seam, congestion-cancellation, a reverse-order dependency chain, duplicate-enqueue idempotency, fork-detection panics, and fullnode cold-sync — each against both schedulers where applicable. `cargo ci-clippy` (incl. msim clippy) and `cargo +nightly fmt` are clean on the touched crates; the heavier simtests/e2e run in CI.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Add an opt-in `ExecutionScheduler` (selected via the `ENABLE_EXECUTION_SCHEDULER` env var); off by default, default behavior unchanged.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


jkrvivian (approved):

filipdulic (approved):

crates/iota-core/src/authority_server/validator_v2.rs
```diff
@@ -610,8 +611,8 @@ impl ValidatorService {
             ValidatorHealthResponse {
                 num_inflight_execution_transactions: self
                     .state
-                    .transaction_manager()
-                    .inflight_queue_len()
+                    .execution_scheduler()
+                    .num_pending_certificates()
```
piotrm50: I don't think that pending certificates is a good measure for validator_v2 API which serves PCOOL flow without certificates.

crates/iota-core/src/execution_scheduler/transaction_manager.rs
```diff
@@ -954,6 +866,31 @@ impl TransactionManager {
     }
 }
 
+impl ExecutionSchedulerAPI for TransactionManager {
+    fn enqueue_impl(
+        &self,
+        certs: Vec<(
```
piotrm50: it's not necessarily a certificate, so I'd rename the variable.

crates/iota-core/src/execution_driver.rs
```diff
@@ -111,6 +115,12 @@ pub async fn execution_process(
         spawn_monitored_task!(epoch_store.within_alive_epoch(async move {
             let _scope = monitored_scope("ExecutionDriver::task");
             let _guard = permit;
+            // Held for the whole execution (not dropped right after dispatch as
+            // upstream SUI does): IOTA's `num_pending_certificates` counts
```
piotrm50: do we want to mention upstream here?

crates/iota-core/src/execution_scheduler/execution_scheduler_impl.rs
```diff
@@ -0,0 +1,290 @@
...
+            )),
+        };
+        let _ = self.tx_ready_transactions.send(pending_cert);
+    }
+}
+
+impl ExecutionSchedulerAPI for ExecutionScheduler {
+    fn enqueue_impl(
+        &self,
+        certs: Vec<(
```
piotrm50: it's not always certs

crates/iota-core/src/execution_scheduler/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Debug)]
+pub struct ExecutingGuard {
+    num_executing_certificates: IntGauge,
+}
+
+#[enum_dispatch]
+pub(crate) trait ExecutionSchedulerAPI {
+    fn enqueue_impl(
+        &self,
+        certs: Vec<(
```
piotrm50: not always certs

crates/iota-core/src/execution_driver.rs
```diff
@@ -40,12 +40,16 @@ pub async fn execution_process(
...
         tokio::select! {
             result = rx_ready_transactions.recv() => {
                 if let Some(pending_tx) = result {
                     transaction = pending_tx.transaction;
                     expected_effects_digest = pending_tx.expected_effects_digest;
                     txn_ready_time = pending_tx.stats.ready_time.unwrap();
+                    executing_guard = pending_tx.executing_guard;
                 } else {
                     // Should only happen after the AuthorityState has shut down and
                     // tx_ready_transaction has been dropped by TransactionManager.
```
piotrm50: Still references TransactionManager

crates/iota-core/src/authority.rs
```diff
@@ -3488,8 +3492,14 @@ impl AuthorityState {
...
+        &self.execution_scheduler
+    }
+
+    /// Whether this authority runs the new `ExecutionScheduler` (vs the
+    /// `TransactionManager`). Used by tests to assert which is active.
+    pub fn uses_execution_scheduler(&self) -> bool {
+        self.execution_scheduler.uses_execution_scheduler()
     }
 
     /// Adds transactions / certificates to transaction manager for ordered
```
piotrm50: refs transcation manager despite using execution scheduler

crates/iota-core/src/authority.rs
```diff
@@ -3499,7 +3509,7 @@ impl AuthorityState {
         transactions: Vec<VerifiedExecutableTransaction>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) {
-        self.transaction_manager.enqueue(transactions, epoch_store)
+        self.execution_scheduler.enqueue(transactions, epoch_store)
     }
 
     /// Adds certificates to transaction manager for ordered execution.
```
piotrm50: refs transcation manager despite using execution scheduler here and in some other places across this file

crates/iota-core/src/authority_server/validator_v2.rs
chriscandless: Fixed

crates/iota-core/src/execution_scheduler/transaction_manager.rs
chriscandless: Fixed

crates/iota-core/src/execution_driver.rs
chriscandless: Fixed

crates/iota-core/src/execution_scheduler/execution_scheduler_impl.rs
chriscandless: Fixed

crates/iota-core/src/execution_scheduler/mod.rs
chriscandless: Fixed

crates/iota-core/src/execution_driver.rs
chriscandless: Fixed

crates/iota-core/src/authority.rs
chriscandless: Fixed

crates/iota-core/src/authority.rs
chriscandless: Fixed

piotrm50 (approved):

muXxer (approved):

PR #11967, opened 2026-06-18
fix(iota-indexer): scope the watermark and the cursor pruning check to only the tables provided filter reads

### LINKED ISSUE #11963: [iota-indexer]: JSON-RPC fallback not triggered for filtered transaction queries
## Description 

Filtered queries (e.g. `FromAddress`) currently calculate the `min_available_tx` watermark by aggregating across the entire set of transaction-related tables. Because this includes every table used by any possible filter, regardless of whether the specific query touches them, the indexer is forced to use a global watermark dominated by the most heavily pruned table.

When unrelated tables are pruned further than the specific table required for the current filter, this creates an artificially high lower bound. Consequently, valid rows present in the relevant table are silently dropped from the result set, and the indexer fails to trigger the necessary KV-store fallback, leading to incomplete data responses and incorrect pagination.

```rust
let tx_tables = [
    CommitterTables::Transactions,
    CommitterTables::TxCallsFun,
    CommitterTables::TxCallsMod,
    CommitterTables::TxCallsPkg,
    CommitterTables::TxInputObjects,
    CommitterTables::TxChangedObjects,
    CommitterTables::TxWrappedOrDeletedObjects,
    CommitterTables::TxSenders,
    CommitterTables::TxRecipients,
    CommitterTables::TxKinds,
];

// Currently uses the full set of tables, causing the bound to be
// dominated by the most-pruned table rather than the queried table.
let min_available_tx = self
    .watermark_cache
    .get_lowest_available_tx_for_tables(&tx_tables)
    .unwrap_or(0);
```

For a `FromAddress` query, `tx_senders` might be available from 0, yet `min_available_tx` incorrectly resolves to `677084555` because an unrelated table is pruned to that point. This produces queries like:

```sql
SELECT tx_sequence_number FROM tx_senders
WHERE sender = ... AND tx_sequence_number >= 677084555
ORDER BY tx_sequence_number ASC LIMIT 51;
```
which excludes valid historical data. For one concrete account, this query returns only 30 rows and reports `hasNextPage: false`, while the same query without the lower bound returns 213 rows. The watermark is silently dropping the 183 rows that still exist in `tx_senders`. Because the 30 returned rows all resolve to valid transactions, the KV-store fallback never triggers, resulting in silent data loss.

> [!NOTE]
> This issue remained latent on unpruned databases because the lowest-available tx was 0 across all tables, rendering the over-broad set harmless. The bug only surfaces on pruned mainnet or testnet deployments where table prune-points diverge.

## Proposed Fix

Compute `min_available_tx` using only the tables explicitly required by the specific filter (e.g. `FromAddress -> [TxSenders]`, `ToAddress -> [TxRecipients]`). Map each filter variant to its minimal required table set so the watermark accurately reflects data availability.

### Challenge

While triggering the fallback for descending order is straightforward, ascending order queries previously lacked the necessary first-transaction sequence number for the Address. Version 1.26 provides this information, making the implementation trivial.

sergiupopescu199:
# Description of change

This PR fixes a bug in the iota-indexer where filtered queries were incorrectly dropping valid rows and failing to trigger the KV-store fallback because they calculate the `min_available_tx` watermark using a global set of tables, causing the filter to be unnecessarily restricted by the most heavily pruned table.

fixes #11963

Ran the `iota-indexer` JSON-RPC locally against mainnet, with both the Postgres database and the KV fallback store pointed at mainnet.
- Re-ran the `queryTransactionBlocks` request FromAddress filter on the address previously failed to paginate.
- Confirmed:
  - `min_available_tx` is now scoped to `tx_senders`, which is unpruned, so it resolves to `0`. The emitted SQL no longer carries the inflated lower bound; it reads `tx_sequence_number >= 0`.
  - Pagination walks the full history correctly across pages (descending cursor advancing `< 321263599`, then `< 3618188`, down to the final page), with no rows silently dropped.
  - Transaction content for pruned ranges is served from the fallback store via `multi_get_transaction_block_response_by_sequence_numbers_with_fallback`.
  - The final page returns 13 rows (fewer than the limit of 51), correctly terminating pagination.

> [!NOTE]

### Release Notes

- [x] Indexer: Fixed resolution of the pruning threshold in the JSON-RPC filtered queries.

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -1574,8 +1574,10 @@ impl IndexerReader {
         &self,
         filter: Option<&TransactionFilterKind>,
     ) -> IndexerResult<&'static [CommitterTables]> {
+        use CommitterTables::*;
+
         let Some(filter) = filter else {
-            return Ok(&[CommitterTables::Transactions]);
+            return Ok(&[Transactions]);
         };
         let tables: &'static [CommitterTables] = match filter {
             TransactionFilterKind::V1(TransactionFilter::MoveFunction {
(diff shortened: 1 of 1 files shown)


crates/iota-indexer/src/read.rs
```diff
@@ -1573,6 +1563,89 @@ impl IndexerReader {
...
+    /// used to scope both the `min_available_tx` watermark and the cursor
+    /// pruning check.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the filter is not supported.
+    fn tx_tables_for_filter(
+        &self,
+        filter: Option<&TransactionFilterKind>,
+    ) -> IndexerResult<&'static [CommitterTables]> {
```
tomxey, line 1576: I think for all of the filters we also need tx_digests in case fallback would be needed? To do the sequence number -> digest conversion.

crates/iota-indexer/src/read.rs
tomxey, line 1542: not the scope of this PR, but I wonder if maybe we should replace "transactions" here with "tx_digests" ? this would allow pagination of transactions even if transactions table itself is pruned

we could open a new improvement issue for this if this is something that we want

cc: @kodemartin

crates/iota-indexer/src/read.rs
```diff
@@ -1573,6 +1563,89 @@ impl IndexerReader {
...
+            TransactionFilterKind::V1(TransactionFilter::Checkpoint(_))
+            | TransactionFilterKind::V2(TransactionFilterV2::Checkpoint(_)) => &[
+                CommitterTables::Transactions,
+                CommitterTables::PrunerCpWatermark,
+            ],
+            TransactionFilterKind::V1(TransactionFilter::FromOrToAddress { .. })
+            | TransactionFilterKind::V2(TransactionFilterV2::FromOrToAddress { .. }) => &[
+                CommitterTables::TxSenders,
+                CommitterTables::TxRecipients,
+                CommitterTables::Transactions,
```
kodemartin: We might want to prune `Transactions` more aggressively. In this case we can still use the `min_available_tx` from `Senders,Recipients` and resolve the digests to call the fallback.

```suggestion
```

crates/iota-indexer/src/read.rs
kodemartin: that's true, feel free to open an issue @tomxey

crates/iota-indexer/src/read.rs
sergiupopescu199: That's true! This change should be relevant only for 1.25 release! In 1.26 we handle this filter in the dedicated `query_transactions_by_affected_addresses_with_fallback` and the checks of watermark boundary is a bit different in conjunction with fallback.

So ideally this match branch will not be invoked same goes for `Checkpoint` filter variant.

crates/iota-indexer/src/read.rs
sergiupopescu199: Good catch! 

fixed: https://github.com/iotaledger/iota/pull/11967/commits/46e9aa729c4871837bfedde7de096524820e3f83

kodemartin (approved):
lgtm 🪛

tomxey (approved):

crates/iota-indexer/src/read.rs
tomxey: https://github.com/iotaledger/iota/issues/11974

PR #11975, opened 2026-06-18
feat(consensus): wire the sliding-window leader schedule behind a protocol flag

### LINKED ISSUE #11886: Add sliding window logic for leader schedule

VorobyevIlya:
# Description of change

Wires the sliding-window leader scorer so it can drive the leader schedule,
behind a default-off protocol flag (`consensus_enable_sliding_window_leader_schedule`).
With the flag off — the default everywhere — behavior is byte-for-byte the
existing V2 path, so nothing changes on the network until a future protocol
version enables it.

When the flag is on:

- `CONSENSUS_COMMITS_PER_SCHEDULE` (how often the schedule rotates) becomes a
  version-gated `consensus_commits_per_schedule` param (default 300), and a new
  `consensus_leader_schedule_window_size` param (default 600) sets the scoring
  depth — decoupling how often the schedule rotates from how many commits the
  reputation scores aggregate over. The flag getter asserts the invariant
  `window_size >= commits_per_schedule`.
- `LeaderSchedule` feeds each committed subdag to the scorer and, every
  `commits_per_schedule` commits, rebuilds the `LeaderSwapTable` from the
  scorer's `reputation_scores()` (the running-window aggregate) instead of the
  `ScoringSubdag` snapshot. The rebuilt scores are tagged with the committed
  interval ending at the rotation boundary — the same shape V2 produces — so
  both paths now go through the *same* `range_validation` and persist the
  scores to `CommitInfo`.
- The per-round base election is uniform instead of stake-weighted.
- Restart and fast-sync recover the scorer by replaying the last window from
  storage (fast-sync fetches enough commits to cover that range); the in-effect
  swap table is recovered from `CommitInfo` as before.

Selection still uses the swap table (good/bad, swap bad→good) — only the score
source and the base draw change.

## Restart-invariance

The scorer's scored frontier lags the rotation boundary by `MAX_PENDING_COMMITS`
(3) commits. Persisting that lagging frontier as the `CommitInfo` range end made
a restarted node rebuild extra scoring subdags and rotate the schedule early,
diverging from a node that ran continuously. We persist the rotation boundary
(the last committed index) as the range end — matching V2 — so a restarted node
resumes its scoring subdag from the same commit a never-restarted node would.
`test_leader_schedule_restart_invariant_rotation` drives a real 4-core network
past a rotation to a mid-interval commit, restarts authority 0 through the real
recovery path, and asserts — for both flag states — that the
commits-until-next-rotation and the recovered swap table's good/bad nodes are
identical live-vs-restarted (the boundary is also the swap-table seed, so
recovery reproduces the same selection, not just the same count).

## Incidental cleanups

- Renamed `last_decided_leader` → `last_finalized_leader` and clarified the
  decided / resolved / finalized leader terminology across `Core` and the
  committer.
- Dropped the round-keyed `scan_until_leader_round_above` from per-commit
  scoring; the r+1 / r+2 blocks are collected from the three commits following
  the scored commit.
- Fixed the pre-existing `CoreTextFixture` → `CoreTestFixture` typo in the
  consensus test harness.

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -528,6 +528,13 @@ struct FeatureFlags {
     // Conflicts are resolved deterministically post-consensus using persistent locks.
     #[serde(skip_serializing_if = "is_false")]
     enable_white_flag_flow: bool,
+
+    // If true, the Starfish leader schedule scores reputation over a sliding
+    // window and rebuilds the swap table every `consensus_commits_per_schedule`
+    // commits with a uniform base election; when false, V2 snapshot scoring +
+    // stake-weighted base election is used.
+    #[serde(skip_serializing_if = "is_false")]
+    consensus_enable_sliding_window_leader_schedule: bool,
 }
 
 fn is_true(b: &bool) -> bool {
@@ -1395,6 +1402,16 @@ pub struct ProtocolConfig {
     // `fun native_sender_authenticator_function_info_v1<F>(): &Option<F>`
     // `fun native_sponsor_authenticator_function_info_v1<F>(): &Option<F>`
     auth_context_authenticator_function_info_v1_cost_base: Option<u64>,
+
+    /// Number of committed subdags between leader-schedule recomputations — the
+    /// rotation cadence (formerly the `CONSENSUS_COMMITS_PER_SCHEDULE` const).
+    /// When unset, defaults to 300.
+    consensus_commits_per_schedule: Option<u32>,
+
+    /// Number of committed subdags the sliding-window leader scorer aggregates
+    /// over (the scoring depth). When unset, defaults to 600. Consulted only
+    /// when `consensus_enable_sliding_window_leader_schedule` is set.
+    consensus_leader_schedule_window_size: Option<u32>,
 }
 
 // feature flags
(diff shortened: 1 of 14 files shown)


bingyanglin (approved):
Approve `crates/iota-protocol-config/src/lib.rs`.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -1395,6 +1402,15 @@ pub struct ProtocolConfig {
     // `fun native_sender_authenticator_function_info_v1<F>(): &Option<F>`
     // `fun native_sponsor_authenticator_function_info_v1<F>(): &Option<F>`
     auth_context_authenticator_function_info_v1_cost_base: Option<u64>,
+
+    /// Number of committed subdags between leader-schedule recomputations.
+    /// When unset, defaults to 300.
+    consensus_commits_per_schedule: Option<u32>,
```
polinikita, line 1408: It might make sense to add validation that consensus_commits_per_schedule > 0. Right now a zero override/config value is accepted. That makes LeaderSchedule think an update is due immediately, before any scoring subdag exists, and can panic in the schedule update path. Also validate leader_schedule_window_size > 0 when the sliding-window flag is enabled.

crates/starfish/core/src/core.rs
```diff
@@ -1281,7 +1283,7 @@ impl Core {
                 .metrics
                 .node_metrics
                 .last_decided_leader_round
```
polinikita, line 1285: maybe as a follow up rename the metric and the dashboard

polinikita (approved):

crates/iota-protocol-config/src/lib.rs
VorobyevIlya: done

crates/starfish/core/src/core.rs
VorobyevIlya: will do in a separate PR

PR #11976, opened 2026-06-19
refactor(iota-types): replace node CheckpointCommitment with SDK type

DaughterOfMars:
## Why

Continues the effort to remove duplicated checkpoint-family types from `iota-types` and adopt their canonical `iota_sdk_types` counterparts (cf. `Object` #11718). This is **PR 1 of the checkpoint-family track** — `CheckpointCommitment` is a leaf type, depended on by `EndOfEpochData` and `CheckpointSummary`.

## What

- Drop the node `enum CheckpointCommitment` and re-export `iota_sdk_types::checkpoint::CheckpointCommitment` in its place. It is BCS-identical (enum variant 0 + 32-byte digest), so the wire format does not change.
- Delete the now-redundant `CheckpointCommitment` `From`/`TryFrom` impls from `iota_sdk_types_conversions.rs`, and simplify the `EndOfEpochData`/`CheckpointSummary` conversions whose `commitments` fields are now the same type on both sides (the `.map(Into::into)` became a no-op).
- Keep `ECMHLiveObjectSetDigest` node-side — it's a standalone digest wrapper used by the snapshot writer and global-state hasher, and its `Default` pulls in the node-only `GlobalStateHash`. No conversion helpers are added to it: because `iota-types`' `Digest` is already a re-export of `iota_sdk_types::Digest`, its `.digest` field plugs straight into the SDK `EcmhLiveObjectSet { digest }` variant.
- Update the construction/match sites in `iota-core`, `iota-tool`, `iota-snapshot`, `iota-graphql-rpc`, and the `iota-json-rpc-types` schema wrappers to build/destructure the SDK variant directly. Since the SDK enum is `#[non_exhaustive]`, each match keeps an explicit catch-all arm that panics if a future variant is added. The JSON-RPC schema wrappers are unchanged on the wire — they still (de)serialize through `ECMHLiveObjectSetDigest`.

## Test plan

- `cargo check` across `iota-types`, `iota-core`, `iota-indexer`, `iota-json-rpc-types`, `iota-graphql-rpc`, `iota-grpc-server`, `iota-snapshot`, `iota-tool`.
- `cargo clippy --all-targets` on the touched crates — clean.
- `cargo nextest run -p iota-types -p iota-json-rpc-types` — 243 passed.
- BCS format snapshot test passes; the snapshot diff reflects only the Rust-level variant rename (`ECMHLiveObjectSetDigest` newtype → `EcmhLiveObjectSet { digest }`), not a wire change.

### DIFF
--- a/crates/iota-snapshot/src/uploader.rs
+++ b/crates/iota-snapshot/src/uploader.rs
@@ -12,6 +12,7 @@ use iota_core::{
     checkpoints::CheckpointStore,
     db_checkpoint_handler::{STATE_SNAPSHOT_COMPLETED_MARKER, SUCCESS_MARKER},
 };
+use iota_sdk_types::CheckpointCommitment;
 use iota_storage::{
     FileCompression,
     object_store::util::{
@@ -148,13 +149,16 @@ impl StateSnapshotUploader {
                     .get_epoch_state_commitments(*epoch)
                     .expect("Expected last checkpoint of epoch to have end of epoch data")
                     .expect("Expected end of epoch data to be present");
-                let state_hash_commitment: ECMHLiveObjectSetDigest = commitments
+                let CheckpointCommitment::EcmhLiveObjectSet { digest } = *commitments
                     .last()
                     .expect("Expected at least one commitment")
-                    .clone()
-                    .into();
+                else {
+                    unimplemented!(
+                        "A new CheckpointCommitment variant was added and must be handled"
+                    )
+                };
                 state_snapshot_writer
-                    .write(*epoch, db, state_hash_commitment)
+                    .write(*epoch, db, ECMHLiveObjectSetDigest { digest })
                     .await?;
                 info!("State snapshot creation successful for epoch: {}", *epoch);
                 // Records the on-chain start timestamp of this epoch (= timestamp of the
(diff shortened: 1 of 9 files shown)


crates/iota-snapshot/src/uploader.rs
```diff
@@ -148,12 +149,16 @@ impl StateSnapshotUploader {
...
                     .expect("Expected last checkpoint of epoch to have end of epoch data")
                     .expect("Expected end of epoch data to be present");
-                let ECMHLiveObjectSetDigest(state_hash_commitment) = commitments
+                let CheckpointCommitment::EcmhLiveObjectSet { digest } = *commitments
                     .last()
                     .expect("Expected at least one commitment")
-                    .clone();
+                else {
+                    unimplemented!(
+                        "A new CheckpointCommitment variant was added and must be handled"
```
bingyanglin:
```suggestion
                        "a new CheckpointCommitment variant was added and must be handled"
```

bingyanglin (approved):

Thoralf-M (approved):

piotrm50 (approved):

Alex6323 (approved):

tomxey (approved):

PR #11985, opened 2026-06-19
upstream(iota-json-rpc): pair checkpoint summaries with their own contents

piotrm50:
## Description of change

- Port of upstream PR: [MystenLabs/sui#26964](https://github.com/MystenLabs/sui/pull/26964)
- Description:

`ReadApi::get_checkpoints_internal` fetched checkpoint summaries and contents into two `Option` vectors aligned by sequence number, then flattened each side independently before zipping the dense remainders back together. Checkpoint pruning deletes a checkpoint's contents while leaving its sequence-addressable summary in `certified_checkpoints`, so `multi_get_checkpoints` returns `Some(summary)` with `None` contents for a pruned sequence. Independently dropping the `None`s shifted every later content one slot to the left, so a summary could be paired with a different checkpoint's transaction list. Because `Checkpoint::from` copies those transactions verbatim, `iota_getCheckpoints` could return rows whose sequence number, digest, and signature described one checkpoint while the transaction list came from another.

This zips the two `Option` vectors index-by-index and only constructs a `Checkpoint` when both the summary and its contents are present; any sequence whose contents are unavailable (e.g. pruned) is skipped rather than paired with a neighbor's data. This preserves the existing "return available checkpoints, skip missing ones" behavior while keeping every response row bound to its own sequence number.

N/A

Added a regression test (`test_get_checkpoints_internal_preserves_alignment_across_pruned_contents`) that prunes the contents of an interior sequence and asserts the surviving checkpoints are returned aligned, each carrying its own transaction digests. Verified the test fails against the pre-fix flatten-then-zip logic and passes with the fix. Run with `cargo nextest run -p iota-json-rpc test_get_checkpoints_internal_preserves_alignment_across_pruned_contents`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6527,6 +6527,7 @@ dependencies = [
  "move-core-types",
  "move-package",
  "prometheus",
+ "roaring",
  "serde",
  "serde_json",
  "signature 1.6.4",
--- a/crates/iota-json-rpc/Cargo.toml
+++ b/crates/iota-json-rpc/Cargo.toml
@@ -66,6 +66,7 @@ typed-store.workspace = true
 # external dependencies
 expect-test.workspace = true
 mockall.workspace = true
+roaring.workspace = true
 
 # internal dependencies
 telemetry-subscribers.workspace = true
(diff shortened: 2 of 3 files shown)


kodemartin (approved):
lgtm

sergiupopescu199 (approved):

PR #11989, opened 2026-06-22
feat(graphql): add archival fallback for checkpoint queries

### LINKED ISSUE #11934: [graphql]: add archival fallback for checkpoint queries
Make GraphQL fall back to the archival store when reading checkpoints whose rows have been pruned from Postgres.

### GraphQL paths covered
- `Query.checkpoint(id)`
- `Query.checkpoints(...)`

### Implementation
Route the checkpoint resolvers through `IndexerReader::get_checkpoint_with_fallback` and `get_checkpoints_with_fallback`. JSON-RPC `iota_getCheckpoint` / `iota_getCheckpoints` already use these methods.

Depends on: #11932
Parent: #11855

tomxey:
# Description of change

Makes the GraphQL `Query.checkpoint(id)` and `Query.checkpoints(...)` resolvers fall back to the historical fallback storage when reading checkpoints whose rows have been pruned from Postgres.

### New loaders

- **`IndexerReader::get_stored_checkpoints_by_seqs_with_fallback`** — reads checkpoints by their sequence numbers from Postgres, falls back to historical fallback storage if needed. 
- **`IndexerReader::get_stored_checkpoints_by_digests_with_fallback`** — same as above but with "digest" key.

-------

- **`DBReader::multi_get_checkpoints_by_seqs`** — internal helper; runs one Diesel query to load checkpoint rows for a list of sequence numbers.
- **`DBReader::multi_get_checkpoints_by_digests`** — internal helper; runs one Diesel query to load checkpoint rows for a list of digests.

----------

- **`HistoricalFallbackReader::checkpoints_by_digests`** — reads a batch of checkpoints from the historical fallback storage by their digests (analogous to already existing `checkpoints` function)

fixes https://github.com/iotaledger/iota/issues/11934

**Setup:** 10 checkpoints across 4 epochs (epoch 0: cps 0–2, epoch 1: cps 3–4, epoch 2: cps 5–6, epoch 3: cps 7–9). With `--epochs-to-keep 1` and `--wait-for-checkpoint-pruned 6`, cps 0–6 are pruned and cps 7–9 remain. No historical fallback configured.

| # | Input | Expected |
  |---|---|---|
  | **A** | `checkpoints { ... }` (defaults) | nodes `[7, 8, 9]`, `hasPreviousPage = false`, `hasNextPage = false` — the reachable range is `[7, 9]` and fits in one page |
  | **B** | `checkpoints(first: 2) { ... }` | nodes `[7, 8]`, `hasPreviousPage = false`, `hasNextPage = true` — at the lower bound, more above |
  | **C** | `checkpoints(first: 2, after: cursor{c: 9, s: 2})` | error `DATA_PRUNED` — cursor's seq is below the pruning watermark |
  | **C2** | `checkpoints(first: 2, after: cursor{c: 9, s: 42})` | error `BAD_USER_INPUT` — cursor's seq is above its own `checkpoint_viewed_at` (malformed) |
  | **D** | `checkpoints(first: 2, after: cursor{c: 9, s: 7})` | nodes `[8, 9]`, `hasPreviousPage = true`, `hasNextPage = false` — cursor at the lower boundary is in-range |
  | **E** | `checkpoint(id: { sequenceNumber: 0 })` | `null` — pruned, no fallback to consult |
  | **F** | `checkpoint(id: { sequenceNumber: 7 })` | `{ sequenceNumber: 7 }` — unpruned, resolves |
  | **G** | `epoch(id: 0) { epochId, checkpoints { ... } }` | `epochId: 0`; `checkpoints` errors `DATA_PRUNED` — epoch 0's range is entirely below the watermark |
  | **H** | `epoch(id: 3) { epochId, checkpoints { ... } }` | `epochId: 3`; nodes `[7, 8, 9]`, both flags `false` — unpruned epoch paginates normally |

### Release Notes

- [x] GraphQL: checkpoint queries (`Query.checkpoint`, `Query.checkpoints`, `Epoch.checkpoints`) can now read pruned checkpoints from the fallback KV store when it is configured.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/checkpoint.rs
+++ b/crates/iota-graphql-rpc/src/types/checkpoint.rs
@@ -2,7 +2,10 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap},
+    ops::RangeInclusive,
+};
 
 use async_graphql::{
     connection::{Connection, CursorType, Edge},
@@ -417,36 +420,9 @@ impl Checkpoint {
             ));
         }
 
-        // Reject cursors outside the available range. Below the lower bound means data
-        // was pruned in the middle of pagination; above the upper bound means the
-        // cursor is malformed.
-        let validate_cursor = |name: &str, cursor: Option<&Cursor>| -> Result<Option<u64>, Error> {
-            let Some(cursor) = cursor else {
-                return Ok(None);
-            };
-            let seq = cursor.sequence_number;
-            if seq < absolute_lo_incl {
-                return Err(Error::DataPruned(format!(
-                    "`{name}` cursor (seq {seq}) is below the available range {available_range:?}"
-                )));
-            }
-            if seq > absolute_hi_incl {
-                return Err(Error::Client(format!(
-                    "`{name}` cursor (seq {seq}) is above the available range {available_range:?}"
-                )));
-            }
-            Ok(Some(seq))
-        };
-        // Narrow the range using cursors, keeping in mind cursors are exclusive.
-        let page_lo_incl = validate_cursor("after", page.after())?
-            .map_or(absolute_lo_incl, |s| s.saturating_add(1));
-        let page_hi_incl = validate_cursor("before", page.before())?
-            .map_or(absolute_hi_incl, |s| s.saturating_sub(1));
-
-        let page_range = page_lo_incl..=page_hi_incl;
-        if page_range.is_empty() {
+        let Some(page_range) = page.narrow_to_available_range(&available_range)? else {
             return Ok(Connection::new(false, false));
-        }
+        };
 
         // Take `limit` sequence numbers from the appropriate end of the page range.
         let limit = page.limit();
@@ -515,6 +491,63 @@ impl Checkpointed for Cursor {
 }
 
 impl ScanLimited for Cursor {}
+
+impl Page<Cursor> {
+    /// Narrow page range to the `available` range and return the inclusive
+    /// range this page should cover, or `None` if the range is empty.
+    ///
+    /// A cursor below `available.start()` returns `Error::DataPruned` (the
+    /// data was pruned since the cursor was issued). A cursor above
+    /// `available.end()` returns `Error::Client` (the cursor is malformed).
+    /// When `after > before` the range is empty and no error is returned.
+    fn narrow_to_available_range(
+        &self,
+        available: &RangeInclusive<u64>,
+    ) -> Result<Option<RangeInclusive<u64>>, Error> {
+        let lo = *available.start();
+        let hi = *available.end();
+
+        let after = self.after().map(|c| c.sequence_number);
+        let before = self.before().map(|c| c.sequence_number);
+
+        // If `after > before`, the range is empty; skip cursor validation.
+        if let (Some(after), Some(before)) = (after, before) {
+            if after > before {
+                return Ok(None);
+            }
+        }
+
+        let ensure_in_available_range = |name: &str, seq: u64| -> Result<(), Error> {
+            if seq < lo {
+                return Err(Error::DataPruned(format!(
+                    "`{name}` cursor (seq {seq}) is below the available range {available:?}"
+                )));
+            }
+            if seq > hi {
+                return Err(Error::Client(format!(
+                    "`{name}` cursor (seq {seq}) is above the available range {available:?}"
+                )));
+            }
+            Ok(())
+        };
+        if let Some(seq) = after {
+            ensure_in_available_range("after", seq)?;
+        }
+        if let Some(seq) = before {
+            ensure_in_available_range("before", seq)?;
+        }
+
+        // Cursors are exclusive.
+        let page_lo = after.map(|seq| seq.saturating_add(1)).unwrap_or(lo);
+        let page_hi = before.map(|seq| seq.saturating_sub(1)).unwrap_or(hi);
+        let range = page_lo..=page_hi;
+        if range.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(range))
+        }
+    }
+}
 
 impl Loader<SeqNumKey> for Db {
     type Value = Checkpoint;
--- a/crates/iota-graphql-rpc/src/types/checkpoint.rs
+++ b/crates/iota-graphql-rpc/src/types/checkpoint.rs
@@ -2,7 +2,7 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use async_graphql::{
     connection::{Connection, CursorType, Edge},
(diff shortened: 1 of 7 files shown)


crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -342,31 +379,68 @@ impl Checkpoint {
...
+            .map_err(|err| Error::Internal(format!("Failed to fetch checkpoints: {err}")))?
+            .into_iter()
+            .flatten()
+            .collect();
+        all_rows.sort_by_key(|s| s.sequence_number);
+
+        if all_rows.is_empty() {
+            return Err(Error::Internal(
+                "checkpoints in the requested page have been pruned and are not available in fallback storage".into(),
+            ));
```
kodemartin: Should we perhaps return a `DataPruned` variant instead of an internal server error?

Pruned data should be an accepted operational state after all. How do we handle this in the JSON-RPC?

crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -342,31 +379,68 @@ impl Checkpoint {
...
-        // queries.
-        let mut conn = Connection::new(prev, next);
-        for stored in results {
+        let fetched_lo = all_rows.first().expect("checked non-empty").sequence_number as u64;
+        let fetched_hi = all_rows.last().expect("checked non-empty").sequence_number as u64;
+
+        // Pruning progresses from the `lo` side. Too few rows returned -> no usable
+        // previous page.
+        let fully_unpruned_page = all_rows.len() == expected_count;
+        let has_prev = fully_unpruned_page && fetched_lo > absolute_lo_incl;
```
kodemartin: what happens if `fully_unpruned_page = true` but the previous page is pruned?

crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -475,22 +521,24 @@ impl Loader<DigestKey> for Db {
...
-        let checkpoints: Vec<StoredCheckpoint> = self
-            .execute(move |conn| {
-                conn.results(move || {
-                    dsl::checkpoints.filter(dsl::checkpoint_digest.eq_any(digests.iter().cloned()))
-                })
+        let digests: Vec<CheckpointDigest> = keys
+            .iter()
+            .map(|key| {
+                CheckpointDigest::from_bytes(key.digest.to_vec())
+                    .map_err(|e| Error::Internal(format!("Invalid CheckpointDigest: {e}")))
```
kodemartin: Why not use the `CheckpointDigest::new` instead?

kodemartin (commented):
Looks good, some minor comments/suggestions.

crates/iota-graphql-rpc/src/types/checkpoint.rs
tomxey: 68eb931373

crates/iota-graphql-rpc/src/types/checkpoint.rs
tomxey: 68eb931373

crates/iota-graphql-rpc/src/types/checkpoint.rs
tomxey: 68eb931373

crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -336,37 +375,116 @@ impl Checkpoint {
...
+            }
+            Ok(Some(seq))
+        };
+        // Narrow the range using cursors, keeping in mind cursors are exclusive.
+        let page_lo_incl = validate_cursor("after", page.after())?
+            .map_or(absolute_lo_incl, |s| s.saturating_add(1));
+        let page_hi_incl = validate_cursor("before", page.before())?
+            .map_or(absolute_hi_incl, |s| s.saturating_sub(1));
+
+        let page_range = page_lo_incl..=page_hi_incl;
```
kodemartin: I think this is a bit hard to follow. The inline closure is one thing, so elevating into a function would be one way to go, but what's more important is that it doesn't build on top of the `Page` invariants.

It treats `page.after`, `page.before` as independent values and performs all possible checks. 

If we instead assume (or require) that `page.after <= page.before` the checks won't have to be that verbose. That is we can implement a simpler logical flow as follows:

```
  low_cursor  = after  if given, else before            
  high_cursor = before if given, else after                                                                                                                   
                                                                                                                                                              
  # --- two checks, each against its near edge ---                             
  if low_cursor exists  and low_cursor  < lo:   ERROR "data pruned"                                                                                  
  if high_cursor exists and high_cursor > hi:   ERROR "bad input: malformed cursor"
                                                                                                                                                              
  # --- build the page slice (cursors are exclusive) ---                       
  page_lo = after + 1   if after given,  else lo                                                                                                              
  page_hi = before - 1  if before given, else hi      
```

And this can be done in an `impl Page<Cursor>` method.

kodemartin (commented):
Looks good, with one final suggestion.

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -268,6 +268,60 @@ impl HistoricalFallbackReader {
         Ok(checkpoints)
     }
 
+    /// Fetches multiple checkpoints by their digests from the historical
+    /// fallback storage. The returned vector has the same length as `digests`
+    /// and preserves its order; missing entries become `None`.
+    ///
+    /// # NOTE
+    /// As with [`Self::checkpoints`], `StoredCheckpoint.successful_tx_num` is
```
sergiupopescu199: nit: instead of `Self::` could we use the actual struct name to be more descriptive. Or as an alternative it could be like this ` [`checkpoint`](Self::checkpoint)`.

sergiupopescu199 (approved):

crates/iota-graphql-rpc/src/types/checkpoint.rs
tomxey: done in effd893cd2

i refrained from using the   `low_cursor  = after  if given, else before` kind of logic since I think it's also a bit confusing, it could suggest to the user that `before` can be treated as lower bound, and vice versa

I think the code refactored into a separate method increases clarity nevertheless, let me know what you think

crates/iota-indexer/src/historical_fallback/reader.rs
tomxey: done in 44047f2db9, I think short name is enough

crates/iota-graphql-rpc/src/types/checkpoint.rs
```diff
@@ -419,35 +492,85 @@ impl Checkpointed for Cursor {
...
+                )));
+            }
+            Ok(())
+        };
+        if let Some(seq) = after {
+            ensure_in_available_range("after", seq)?;
+        }
+        if let Some(seq) = before {
+            ensure_in_available_range("before", seq)?;
+        }
```
kodemartin: The refactoring is far clearer, and the method name is straigh-to-the-point.

A nit is the following:

Given that `after <= before`, and if

```
let after = after.map(|c| ("after", c.sequence_number));
let before = ...
``` 

you can collapse this into

```rust
if let Some((name, seq)) = after.or(before) {
       if seq < lo {
             DataPruned
       }
}
if let Some((name, seq)) = before.or(after) {
       if seq > hi {
              Client
       }
}
```

kodemartin (approved):

crates/iota-graphql-rpc/src/types/checkpoint.rs
tomxey: [d697ab3](https://github.com/iotaledger/iota/pull/11989/commits/d697ab3f31c4ac6bed4a302fffccb538129857f3)

PR #11994, opened 2026-06-22
refactor(consensus): remove dead v29 flag-off branches from the consensus runtime

### LINKED ISSUE #11965: refactor(consensus): remove dead flag-off branches after v29 consensus flags activate on mainnet
Several consensus protocol flags are first enabled for mainnet at protocol version 29 (current `MAX_PROTOCOL_VERSION`, enabled globally in the v29 arm; v29 is not yet live on mainnet). Once v29 ships and is stable on mainnet, the legacy flag-off branches that support the pre-flag behavior become dead code and should be removed to simplify the consensus runtime.

Follow-up cleanup, blocked on v29 being permanently active on mainnet.

**Flags enabled for mainnet in v29:**
- `consensus_fast_commit_sync` (#11854) — TransactionRef-based commits and sync. Largest cleanup (~20 flag-off branches): BlockRef vs TransactionRef handling, CommitV1 vs V2/V3, ShardWithProofV1 / SerializedTransactionsV1 (de)serialization, BlockRef-keyed storage paths, `verify_transactions_with_headers`. Across `core.rs`, `dag_state.rs`, `linearizer.rs`, `commit.rs`, `storage/{mod,mem_store,rocksdb_store}.rs`, `authority_service.rs`, `transactions_synchronizer.rs`, `commit_syncer/regular.rs`.
- `consensus_block_restrictions` (#11854) — bounds block-header size and enables block-manager GC. Flag-off no-op guards in `dag_state::evict_pending_commit_votes` and `block_manager::maybe_evict_below_gc_floor`.
- `consensus_median_timestamp_with_checkpoint_enforcement` (#11842) — median commit timestamp + checkpoint monotonicity. The non-median commit-timestamp branch (linearizer) and the enforcement toggle.
- `always_advance_dkg_to_resolution` (#11734) — random-beacon DKG advancement (consensus-adjacent, in iota-core). Minor.

**Scope / constraints:**
- Scope is consensus *runtime* code (`crates/starfish/**` and consensus-adjacent `iota-core`). Per-version arms in `iota-protocol-config` are immutable history and must stay — they reproduce config for historical protocol versions.
- Gate removal on v29 being permanently active on mainnet; before deleting legacy read/deserialization paths, confirm no live epoch or on-disk data still depends on the old formats.
- `consensus_starfish_speed` is defined but not enabled in any version — out of scope.

polinikita:
# Description of change

The four consensus protocol flags first enabled for mainnet at protocol version 29 are now treated as permanently active. Their flag-off runtime branches are dead and have been removed. The flag definitions and historical per-version values remain in `iota-protocol-config`.

**`consensus_fast_commit_sync`**

Uses `TransactionRef` unconditionally across block proposal, commits, `DagState`, storage, cordial knowledge, authority RPCs, and the header, transaction, and commit synchronizers. This removes the legacy header-based commit-sync verification path, pre-versioned shard/transaction deserializers, and endpoint guards.

**`consensus_block_restrictions`**

Makes acknowledgment and commit-vote limits, round bounds, GC eviction, and the quorum-commit-round proposal skip unconditional.

**`consensus_median_timestamp_with_checkpoint_enforcement`**

Makes checkpoint timestamp monotonicity enforcement unconditional.

**`always_advance_dkg_to_resolution`**

Always advances pending random-beacon DKG state so persisted state can resolve without new inbound traffic.

The cleanup also removes obsolete test matrices and setup, `verify_transactions_with_headers`, `SerializedTransactionsV1`, the `FastCommitSyncNotEnabled` error, `GenericTransactionRef::expect_block_ref`, vestigial `Context` storage parameters, and the legacy BlockRef-keyed transaction maps and RocksDB column family.

## Compatibility

Commit digests include the BCS enum tag. The legacy `CommitV1` and `ShardWithProofV1` variant slots therefore remain as deserialize and format anchors, while their construction is removed. `GenericTransactionRef` also retains both variants. Legacy V1 commits and shards still deserialize, and peer-supplied V1 data is rejected by the existing version checks.

The consensus RocksDB path is epoch-scoped as `storage_base_path/{epoch}`. A v29 epoch creates a fresh TransactionRef-keyed store and never needs the pre-v29 BlockRef-keyed transaction column family, so removing it requires no migration of an existing epoch store.

No protocol-version change is required.

> [!IMPORTANT]
> **Deployment precondition.** The cleaned runtime unconditionally runs the flag-on behavior. It is safe to ship only after every network is permanently at a protocol version where all four flags are active.

Fixes #11965

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -4341,15 +4341,9 @@ impl AuthorityPerEpochStore {
         // sees no new inbound traffic (e.g. one that restarted) can stay Pending
         // forever -- deferring all randomness-using transactions and blocking
         // epoch close.
-        //
-        // This changes when `advance_dkg` runs relative to consensus, so it is
-        // gated behind a protocol flag: every validator must flip the behavior
-        // at the same protocol-upgrade boundary, otherwise a mixed-version
-        // network could resolve DKG on different commits and diverge.
-        let dkg_pending = self.protocol_config.always_advance_dkg_to_resolution()
-            && randomness_manager
-                .as_ref()
-                .is_some_and(|rm| rm.dkg_status() == DkgStatus::Pending);
+        let dkg_pending = randomness_manager
+            .as_ref()
+            .is_some_and(|rm| rm.dkg_status() == DkgStatus::Pending);
         if randomness_state_updated || dkg_pending {
             if let Some(randomness_manager) = randomness_manager.as_mut() {
                 randomness_manager
--- a/crates/iota-core/src/checkpoints/mod.rs
+++ b/crates/iota-core/src/checkpoints/mod.rs
@@ -1608,13 +1608,7 @@ impl CheckpointBuilder {
                         "Decrease of checkpoint timestamp, possibly due to epoch change. Sequence: {}, previous: {}, current: {}",
                         sequence_number, last_checkpoint.timestamp_ms, timestamp_ms,
                     );
-                    if self
-                        .epoch_store
-                        .protocol_config()
-                        .consensus_median_timestamp_with_checkpoint_enforcement()
-                    {
-                        timestamp_ms = last_checkpoint.timestamp_ms;
-                    }
+                    timestamp_ms = last_checkpoint.timestamp_ms;
                 }
             }
 
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -1557,19 +1557,11 @@ impl ProtocolConfig {
     }
 
     pub fn max_acknowledgments_per_block(&self, committee_size: usize) -> usize {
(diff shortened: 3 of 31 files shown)


bingyanglin (approved):
Approved `crates/iota-core/src/authority/authority_per_epoch_store.rs` and `crates/iota-core/src/checkpoints/mod.rs`,

roman1e2f5p8s (approved):
approving owned files

piotrm50 (approved):

VorobyevIlya (commented):
So it will be shipped only with version 30 or later?
Also I think that GenericTransactionRef can also be removed, maybe not in this PR.

All tests that tested the case when fast sync is off were removed, but it is still possible to disable fast sync locally. Should this path be still tested?

polinikita:
> All tests that tested the case when fast sync is off were removed, but it is still possible to disable fast sync locally. Should this path be still tested?

I will add a test for that

PR #11996, opened 2026-06-23
feat(core): move the epoch_info chain into the CheckpointStore

bingyanglin:
# Description of change

Moves the per-epoch verified chain (`epoch_info`) out of the gRPC index store into the `CheckpointStore`, so every node (validators and full nodes) holds it. Unblocks the future summary pruning, lets snapshots publish without `enable_grpc_api`, and decouples the chain from `--skip-grpc-indexes`.

- New `epoch_info` + `epoch_info_watermark` CFs in `CheckpointStore`, never pruned; logic in new `checkpoints/epoch_info.rs`.
- Checkpoint executor populates the chain live at every boundary (validators included).
- gRPC index store drops its `epochs_v2` CF and `index_epoch`; keeps only transaction + live-state indexes.
- `GetEpoch` now reads the `CheckpointStore` (`get_epoch_info` moved to `GrpcStateReader`).
- Snapshot writer/uploader read `EPOCH_INFO` from the `CheckpointStore`; `enable_grpc_api` requirement removed.
- Node startup `seed_epoch_info` fills any historical gap before serving: a **recognized chain** (mainnet/testnet/current devnet) restores from that chain's formal-snapshot `EPOCH_INFO` first, then replays local checkpoints for the residual tail; an **unrecognized network** rebuilds from local checkpoints only (no snapshot). A residual gap is **fatal on a recognized chain** (every node must hold the verified chain since genesis) and a warning otherwise (left unfilled — such a node won't produce snapshots or serve the epoch gRPC API for those epochs).
- The formal-snapshot source is **hardcoded per chain** (no operator config); fetched `EPOCH_INFO` is verified against the genesis committee. One-time migration aid for pruned upgrading nodes, removed one release later (#12028).
- Local rebuild reconstructs the chain from genesis using only each epoch's **boundary (change-epoch) transaction** — cheap, and succeeds even when non-boundary transactions or old object versions have been pruned.
- Restore tool seeds the chain unconditionally; `--skip-grpc-indexes` now governs only live-state indexes.
- Docs: `snapshots.mdx` made node-generic; the per-epoch-metadata backfill section removed (the source is now built-in). No new node config.


muXxer (approved):

crates/iota-node/src/lib.rs
```diff
@@ -880,63 +848,121 @@ impl IotaNode {
...
+        // entirely when the chain is already complete (the common restart).
+        if checkpoint_store
+            .epoch_info_gap()
+            .map_err(|e| anyhow::anyhow!("checking epoch_info completeness: {e}"))?
+            .is_some()
+        {
+            // A recognized network pulls the formal snapshot's `EPOCH_INFO`
+            // first — cheap and bulk. Unrecognized networks have no published
+            // snapshot and skip straight to local replay.
+            if let Some(remote_store_config) = &recognized_source {
```
jkrvivian: I think the comments repeat itself, also the code here is quite clear, we can either make it short or just remove it.
```suggestion
            // snapshot pull (recognized chains only)
```

crates/iota-node/src/lib.rs
```diff
@@ -880,63 +848,121 @@ impl IotaNode {
...
+                    genesis.committee()?,
+                    genesis.iota_system_object(),
+                    expected_chain_id,
+                )
+                .await?;
+            }
+
+            // Replay local checkpoints for whatever the snapshot didn't cover
+            // (its published epoch can lag this node's history), or for the
+            // whole chain on an unrecognized network.
```
jkrvivian: Or you can removed the comments entirely
```suggestion
            // local replay
```

jkrvivian (approved):

PR #11999, opened 2026-06-23
fix(scripts): poll for cluster state in split-cluster CI checks

### LINKED ISSUE #11706: flaky: split-cluster CI test fails non-deterministically
## Summary

The `split-cluster / validate-testnet` and `split-cluster / validate-mainnet` CI jobs are flaky. They fail non-deterministically by timing out waiting for the fullnode to log `'Node State has been reconfigured'` within a hard-coded 60-second window.

## Observed failure

Both jobs failed on PR #11672 (first CI run, commit `d14a53e7562b2833f535e2d683b3abdfef737813`) with:

```
sleeping for 60 seconds
shutting down nodes
Could not find 'Node State has been reconfigured' in fullnode log
##[error]Process completed with exit code 1.
```

- Failed run: https://github.com/iotaledger/iota/actions/runs/26567236437
  - `split-cluster / validate-testnet` (job 78271695764)
  - `split-cluster / validate-mainnet` (job 78271695777)

A subsequent push with only unrelated clippy lint fixes (commit `94c41c18a2cc5f0f59af86d8005a61ad89d9ced1`) ran the same test and it passed — confirming the failure is non-deterministic and not a code logic regression.

## Root cause

The split-cluster check script:
1. Builds the new binary, starts a mixed-version cluster
2. Sleeps for exactly 60 seconds
3. Shuts down and greps for `'Node State has been reconfigured'` in the fullnode log

60 seconds is occasionally not enough time for the fullnode to complete reconfiguration and emit the log line, especially under CI runner load. There is no retry or poll logic — it is a single hard-coded sleep followed by a grep.

## Expected behavior

The check should either:
- Poll for the log message with a timeout (e.g. check every 5s for up to 120s), rather than a single fixed sleep, or
- Increase the sleep to a more conservative value that accounts for CI runner variance

## Impact

Causes legitimate PRs to fail CI and require extra pushes to re-trigger, adding friction and noise to the review process.

piotrm50:
# Description of change

The split-cluster compatibility checks (`scripts/compatibility/split-cluster-check.sh` and `split-cluster-sync-check.sh`) waited a fixed `sleep` and then grepped/asserted once. Under CI runner load that fixed window is simultaneously wasteful on a fast runner and too short on a slow one, so the checks fail non-deterministically — and on failure they printed a single generic line with no diagnostics, so triage requires downloading log artifacts.

`split-cluster-check.sh`:
- Replace the fixed `sleep 60` + one-shot grep with a poll loop that waits up to `MAX_WAIT` (default 180s) for the fullnode `Node State has been reconfigured` marker, checking every `POLL_INTERVAL` (5s).
- Capture each node PID and fail fast (with diagnostics) if any node exits early, instead of burning the whole timeout.
- On timeout or early death, dump per-node liveness, a node x marker matrix (reconfiguration + random-beacon DKG markers), and the last 50 lines of every node log. Each node is labelled with the binary version it runs (release vs candidate).
- Add a `trap`-based cleanup so the cluster is always torn down. The DKG checks still verify all four markers (node-0 and node-2, each for `random beacon: created` and `random beacon: DKG complete`).

`split-cluster-sync-check.sh`:
- Replace the fixed `sleep 60` catch-up wait with a poll (up to `CATCHUP_MAX_WAIT`, default 180s) until the late-joining node-3 passes the baseline commit index.
- Keep the Phase-1 warmup a fixed duration (`WARMUP_SECS`, default 180s) on purpose: it builds a large commit backlog so node-3 is far enough behind to actually exercise the Synchronizer/CommitSyncer (a small backlog is caught up through live consensus and exercises neither). It now polls node liveness during warmup to fail fast on an early crash.
- Fail fast on node death and dump version-labelled log tails on failure.

fixes #11706

`bash -n` passes on both scripts. Verified locally against a real 5-node cluster:
- `split-cluster-check.sh`: reconfigures (the poll reported 25s on an idle machine — well within the window, vs the >60s seen on loaded CI runners), and a forced short timeout exercises the diagnostics path (version-labelled liveness table, node x marker matrix, and log tails).
- `split-cluster-sync-check.sh`: the 180s warmup built a 1864-commit backlog, node-3 caught up via the Synchronizer (300 block headers) and CommitSyncer (1533 commits, ~843 KB of transactions), and all four sync checks passed.

Run locally before marking ready to avoid spending CI worker time on the (heavy) split-cluster jobs.

### DIFF
--- a/scripts/compatibility/split-cluster-sync-check.sh
+++ b/scripts/compatibility/split-cluster-sync-check.sh
@@ -85,8 +85,23 @@ done < <(find "$IOTA_CONFIG_DIR" -name "127.0.0.1*.yaml" -print0)
 
 export RUST_LOG=iota=debug,info
 
-# Track PIDs for process management
+# Track child PIDs (and which binary version each runs) so we can report
+# liveness and tear the cluster down cleanly. node-3 is appended only once it
+# starts (late), so the warmup liveness check naturally covers just the initial
+# quorum.
 NODE_PIDS=()
+NODE_NAMES=()
+NODE_VERSIONS=()
+
+start_node() {
+  local name=$1 version=$2 binary=$3 config=$4
+  "$binary" --config-path "$config" > "$LOG_DIR/$name.log" 2>&1 &
+  local pid=$!
+  NODE_PIDS+=("$pid")
+  NODE_NAMES+=("$name")
+  NODE_VERSIONS+=("$version")
+  echo "Started $name ($version) with PID $pid"
+}
 
 # Cleanup function to kill child processes on exit
 cleanup() {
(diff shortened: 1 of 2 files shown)


scripts/compatibility/split-cluster-sync-check.sh
```diff
@@ -143,51 +158,98 @@ sum_metric_values() {
...
-# Start first 3 nodes with release binary
-"$WORKING_DIR/iota-node-release" --config-path "${CONFIGS[0]}" > "$LOG_DIR/node-0.log" 2>&1 &
-NODE_PIDS[0]=$!
-echo "Started node-0 (release) with PID ${NODE_PIDS[0]}"
-
-"$WORKING_DIR/iota-node-release" --config-path "${CONFIGS[1]}" > "$LOG_DIR/node-1.log" 2>&1 &
-NODE_PIDS[1]=$!
-echo "Started node-1 (release) with PID ${NODE_PIDS[1]}"
+POLL_INTERVAL=${POLL_INTERVAL:-5}
+# Fixed warmup so node-0..2 build a large enough commit backlog that the
```
Thoralf-M: Bit weird to have comments that explain a fix instead of just what the code is doing

Thoralf-M (approved):

scripts/compatibility/split-cluster-sync-check.sh
```diff
@@ -143,51 +158,93 @@ sum_metric_values() {
...
+  sleep "$POLL_INTERVAL"
+done
 
-# Capture initial metrics from node-0
+# Capture the baseline commit index that node-3 will have to catch up past.
 get_metrics "${CONFIGS[0]}" "$METRICS_DIR/node-0-before-node3.txt"
 INITIAL_COMMIT_INDEX=$(get_metric_value "$METRICS_DIR/node-0-before-node3.txt" "consensus_last_commit_index")
-echo "Initial commit index on node-0: $INITIAL_COMMIT_INDEX"
+echo "Initial commit index on node-0: $INITIAL_COMMIT_INDEX (after ${WARMUP_SECS}s)"
+if [ "$INITIAL_COMMIT_INDEX" -le 0 ]; then
```
bingyanglin, line 215: No commits in warmup just records a failure but keeps going, so catch-up runs against a zero baseline. Maybe exit 1 here instead?

bingyanglin (approved):

PR #11817, opened 2026-06-09
feat(iota-vm-sdk): wasm-bindgen browser surface

Thoralf-M:
## Summary

Adds an optional, off-by-default `wasm-bindgen` browser surface to `iota-vm-sdk`: the crate compiles to `wasm32` and exports `init` + `simulate`, so a transaction can be run through the local Move VM from JavaScript. Split out of #11755 (merged) for independent review; native builds pull in none of it.

image

## What & why

- **wasm surface** — exports `simulate`, which resolves the objects a transaction touches on demand via a JS callback, so the browser never has to pre-fetch its dynamic-field reads.
- **Target-gated** — the wasm code, its deps, and the stubbed metrics/gas-profile paths all sit behind `cfg(target_arch = "wasm32")` + the feature, so native builds are unchanged.
- **Browser example** (`examples/wasm/`) — simulates a live staking transaction or arbitrary BCS bytes, across networks and execution modes, with signature verification and ready-made success/failure presets.

Build: `cargo build -p iota-vm-sdk --target wasm32-unknown-unknown --features wasm-bindgen`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12000, opened 2026-06-23
chore: validate improved split-cluster diagnostics on #11817 stack (do not merge)

piotrm50:
[run-ci]

Throwaway CI-validation PR — **do not merge, do not review**.

This applies the improved `scripts/compatibility/split-cluster-check.sh` (poll-with-timeout + diagnostics, from #11999) on top of #11817's head, to exercise the new script in a scenario where the old fixed-`sleep 60` + one-shot grep fails consistently (see #11817's split-cluster run). Expected outcome is one of:

- the poll (up to 180s) now finds the reconfiguration marker and the check passes — confirming the consistent failure was just the 60s window being too short; or
- it times out and prints the new diagnostics (per-node liveness, node x marker matrix, log tails) — making the real cause actionable.

Will be closed once the split-cluster job output has been captured.


piotrm50:
Throwaway CI-validation PR — closing now that it has served its purpose.

It ran the improved \`split-cluster-check.sh\` (from #11999) against #11817's candidate. Result: the split-cluster job failed in ~5s (vs the old 60s blind sleep), and the new diagnostics pinpointed the cause — both **candidate** nodes panic on startup at \`external-crates/move/crates/move-vm-profiler/src/lib.rs:399\` (\`ensure_move_vm_profiler_disabled()\` → "Cannot run with Move VM tracing feature enabled"), while the release nodes stay alive. So #11817's split-cluster failure is a real candidate-binary crash (the node was built with the \`move-vm-profiler/tracing\` feature enabled), not a timing flake. Diagnostics validated end-to-end.

PR #12001, opened 2026-06-23
upstream(core): Add generic bulk-ingestion write options for store initialization

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@8ea29284](https://github.com/MystenLabs/sui/commit/8ea29284b3d18d20e6753b5ba595bb396172619c)
- Description:

We have diverged a lot in rpc and grpc. So we only ported the reusable parts of upstream's RPC index init optimization as a generic typed-store facility for bulk-loading a fresh store (initial index build, genesis/formal snapshot restore):
- `write_opt`/`flush` plumbing on `Database`/`DBMap`/`DBBatch` (`flush` covers all column families).
- `bulk_ingestion_write_options()` — WAL-disabled write options.
- `bulk_ingestion_options()` — hardware-sized RocksDB open + per-CF options (disabled compaction, large memtables, batch-size limit).

Wired into the gRPC index initialization. `iota-metrics` moved onto the new workspace `sysinfo` dep.

Part of #11885


crates/iota-core/src/grpc_indexes.rs
```diff
@@ -960,6 +962,14 @@ impl GrpcIndexesStore {
                     .init(&authority_store, checkpoint_store)
                     .expect("unable to initialize gRPC index");
 
+                // Flush all data to disk before dropping tables. This is critical because
+                // WAL is disabled for the bulk writes during initialization. We only need
+                // to flush one table because all tables share the same underlying database.
+                tables
```
bingyanglin, line 988: `meta.flush()` only flushes the default CF, not the named CFs where the bulk data lives. It survives because drop flushes all CFs on shutdown, not this call. Maybe flush all CFs, or fix the comment.

crates/typed-store/src/database.rs
```diff
@@ -311,14 +311,35 @@ impl Database {
         }
     }
 
+    /// Flush all memtables to SST files on disk.
```
bingyanglin: Doc says "flush all memtables" but underlying.flush() is default-CF only. Maybe reword, or iterate the CFs.

crates/iota-core/src/grpc_indexes.rs
muXxer: but the intention is to flush all, no? should we add a flush_all instead or something?

muXxer (commented):
I don't understand why we skipped the "hardware auto-tuning block". Isn't that the big improvement? In my opinion this doesn't only apply to the rpc indexes, but to all init operations where we write lot's of data to the stores, so in my opinion this hardware auto tuning block should live maybe closer to the rocksdb implementation, and then we can reuse it also during genesis loading or loading from formal snapshot maybe?

jkrvivian:
> I don't understand why we skipped the "hardware auto-tuning block". Isn't that the big improvement? In my opinion this doesn't only apply to the rpc indexes, but to all init operations where we write lot's of data to the stores, so in my opinion this hardware auto tuning block should live maybe closer to the rocksdb implementation, and then we can reuse it also during genesis loading or loading from formal snapshot maybe?

Yes, I wasn't sure if we need these tuning options. But since it's only for the initial stage, and switches to the default options on reopen, I ported and added it to the upstream commit.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -95,6 +98,119 @@ pub fn unmark_db_corruption(path: &Path) -> Result<(), Error> {
...
+    // failure.
+    db_options.set_unordered_write(true);
+
+    // Allow CPU-intensive flushing to use all cores.
+    db_options.set_max_background_jobs(num_cpus as i32);
+
+    // We are disabling compaction for all column families below. This means we can
+    // also disable the backpressure that slows down writes when the number of L0
+    // files builds up since we will never compact them anyway.
+    db_options.set_level_zero_file_num_compaction_trigger(0);
```
piotrm50: `set_level_zero_file_num_compaction_trigger(0)` means "trigger compaction once there are >= 0 L0 files" (i.e. always), not "disable". 

The disable sentinel is `-1`, as correctly used on the slowdown trigger on the next line. 

This anyway does not have any effect here because `disable_auto_compactions=true` is set per-CF, but the value (and the comment about disabling backpressure) is misleading.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -95,6 +98,119 @@ pub fn unmark_db_corruption(path: &Path) -> Result<(), Error> {
...
+/// RocksDB options tuned for one-shot bulk ingestion into a freshly created
+/// store, e.g. initial index building, inserting the genesis snapshot, or
+/// restoring a formal snapshot.
+pub struct BulkIngestionOptions {
+    pub db_options: rocksdb::Options,
+    pub column_family_options: DBOptions,
+    pub batch_size_limit: usize,
+}
+
+pub fn bulk_ingestion_options() -> BulkIngestionOptions {
```
piotrm50: Other DB option structs (`ReadWriteOptions`, `DBOptions`, `DBMapTableConfigMap`) live in `rocks/options.rs`. I'd suggest moving `BulkIngestionOptions` and related methods to `options.rs` for consistency.

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -943,23 +951,45 @@ impl GrpcIndexesStore {
...
                     .expect("unable to initialize gRPC index");
 
+                // Flush all data to disk before dropping tables. This is critical because
+                // WAL is disabled for the bulk writes during initialization. Flushing any
+                // table flushes every column family of the shared underlying database, so
+                // one call covers all tables.
+                tables
+                    .meta
+                    .flush()
+                    .expect("failed to flush gRPC index tables to disk");
```
piotrm50: `.expect()` message does is not following the convention. Rather, describe the precondition expected to hold, e.g. `.expect("gRPC index DB should be flushable after bulk ingestion")`.

crates/iota-core/src/grpc_indexes.rs
```diff
@@ -975,7 +1005,22 @@ impl GrpcIndexesStore {
...
-                IndexStoreTables::open(&path)
+                let reopened_tables = IndexStoreTables::open(&path);
+
+                // Sanity check: verify the database version was persisted correctly, i.e.
+                // the WAL-disabled bulk writes were flushed before the reopen.
+                let stored_version = reopened_tables
+                    .meta
+                    .get(&())
+                    .expect("failed to read metadata from reopened database")
+                    .expect("metadata not found in reopened database");
```
piotrm50: Same here. I'd rephrase it to something like  `"reopened gRPC index DB should expose readable metadata"` / `"metadata should have been written before flush and reopen"`.

crates/typed-store/src/rocks/options.rs
```diff
@@ -58,6 +59,120 @@ pub struct DBOptions {
...
+    db_options.set_unordered_write(true);
+
+    // Allow CPU-intensive flushing to use all cores.
+    db_options.set_max_background_jobs(num_cpus as i32);
+
+    // Disable the write backpressure that kicks in as L0 files build up: `-1`
+    // turns off the compaction and slowdown triggers, and the stop trigger is
+    // pushed out of reach. Compaction itself is disabled per column family via
+    // `disable_auto_compactions` below, so no L0 files are ever compacted away.
+    db_options.set_level_zero_file_num_compaction_trigger(-1);
```
bingyanglin: These three L0 triggers are set on `db_options`, but `level0_*_trigger` are per-column-family options, so they never reach the index CFs, which open from `column_family_options`. Maybe move them onto `column_family_options.options`, next to `set_disable_auto_compactions`?

bingyanglin (approved):

piotrm50 (approved):

muXxer (approved):

PR #12006, opened 2026-06-23
feat(core): add the check for smart account claim transactions and pin implicit accounts

### LINKED ISSUE #11900: [AA] Implicit account authentication is non-deterministic
The implicit built-in account feature (#11276) reads the account object at the **latest** version
during sequenced execution, with no consensus version assignment and no scheduling dependency. When
a `claim` / `rotate_builtin_auth_public_key` / `detach` transaction races an in-flight plain-signed
transaction from the same address, validators can read different account states and produce divergent effects digests.

### Where

`TransactionInputLoader::read_built_in_account_objects` (`transaction_input_loader.rs:308`) reads
each account object with `try_get_object` (latest version). Unlike an explicit `MoveAuthenticator`,
whose `object_to_authenticate` gets consensus version assignment, the implicit account object
(ID = signer address) is read out-of-band:
- not in `shared_input_objects()` → `assign_versions_from_consensus` never assigns it a version;
- not in `cert.input_objects()` → `TransactionManager` builds no dependency edge to a concurrent
  claim/rotation.

### Does P-COOL (#11242) fix it?

No. P-COOL removes the owned-only fast path, closing the "cert arrives at different times" window,
but post-consensus execution is not serial in consensus order. `TransactionManager` schedules
on *declared* dependencies and runs ready txs in parallel. The account is undeclared, so a rotation
and a plain-signed tx still have no ordering edge and can execute in either order on different
validators. P-COOL shrinks the window; it does not close it.

### Proposed fix (white-flag / post-P-COOL scope)

Pin the account-object read to a consensus-assigned version by treating each built-in account as
a read-only shared input that participates in version assignment but stays out of
the executor's `InputObjects`.

Because read-only inputs never advance `next_shared_object_versions`, this reduces to: assign the
account a version only when the object exists on chain (→ explicit); otherwise leave it
implicit (signature-derived).

miker83z:
# Description of change

- A transaction that calls `0x2::smart_account::build_v1` turns an EOA account (the TX sender) into an **explicit** Smart Account. 
- When such a "claim" transaction lands in the same consensus commit as other EOA-plain-signed transactions from that same address, we enter into a race: 
  - depending on whether the account is treated as already-explicit, they execute differently across validators, which is non-deterministic and can fork the network.
  - **this PR resolves that race deterministically in the post-consensus white-flag validation pass**. 

## Solution 1: whiteflag check for smart account claim)
1. **Claimers first.** The commit is split into transactions that call `smart_account::build_v1` and the rest. The claimers are validated first; each survivor records its sender as a claimed address.
    1. (The way a TX is recognized as being a "claim" transaction uses a temporary patch; the proper handling of such transaction should be dealt with in a following PR.)
3. **Drop conflicting siblings.** Every other transaction whose **sender or sponsor** matches a claimed address is dropped with a new `IotaError::AccountClaimConflict` before its own validation.

#### Additional key properties:

- Two `build_v1` transactions sharing a sender both survive: the duplicate account creation fails deterministically at execution.
- Already-executed transactions are exempt.
- `AccountClaimConflict` is **retryable** dropped transaction that can be resubmitted in a later commit once the account state has settled, and surfaces to clients as `Rejected` via the existing dropped-tx path.

## [update 1]: Implicit-account version pinning (`9edc77ec51`)

The earlier commits in this PR drop a transaction `T` with a plain signature only when it shares a consensus commit with a conflicting claim `C` (the `claim_senders` / `AccountClaimConflict` check). That handles the **same-commit** ordering, but `C` and `T` on an account `A` can also be ordered across **different** commits, and those two cases need separate, consensus-deterministic handling.

The reason is the determinism boundary: a transaction's place in the total order is fixed at its **consensus commit**, but execution happens later and the lag is *local to each node*. So an execution-time question like *"does `A` have an on-chain object right now?"* is fork-prone, as a faster node may have executed `A`'s claim while a slower node has not. Every auth/version input must come from consensus-handling state, never from a lag-dependent latest read.

This commit adds the two cross-commit complements:

| Ordering of (claim `C`, transfer `T`) on `A` | Handler |
|---|---|
| `C` and `T` in the **same** commit | same-commit `claim_senders` check (earlier in this PR) → drop `T` (`AccountClaimConflict`) |
| `C` **before** `T` | **the anchor** → reject `T` (`PlainSignatureOnClaimedAccount`) |
| `T` **before** `C` | **the pin** → `T` reads `A` as implicit on every node (A @ v1) |

- **The pin (transfer-before-claim).** During version assignment, every implicit account referenced by `T` is pinned at `OBJECT_START_VERSION` (v1) as a **read-only** input, per-transaction, never written back. A claimed account is created at its claim's Lamport version (always ≥ 2), so resolving `A @ v1` always *misses* for a claimed account → `T` deterministically treats `A` as implicit (backed by a synthetic `SmartAccount` built on the fly), regardless of whether the later claim `C` has executed. Read-only keeps `A` out of the Lamport bump, and `A @ v1` is recorded in the effects so state-sync reconstructs the same read.
- **The anchor (claim-before-transfer).** A `build_v1` claim records `ObjectId::from(sender)` in `next_shared_object_versions`, giving post-consensus validation a consensus-deterministic "A was claimed earlier this epoch" signal — so `T` is rejected without depending on whether `C`'s effects have executed yet. (A claim settled in a prior epoch is safely visible in the store thanks to the epoch-boundary barrier.)

Fixes #11900

### Release Notes

- [x] Protocol: Enabled an account claim conflict invalidation check that stops transactions having a sender/sponsor being claimed (smart account) in another transaction in the same commit.
- [x] Nodes (Validators and Full nodes): Validators apply the new conflict rule during post-consensus validation; affected transactions are rejected with a retryable AccountClaimConflict error and can be resubmitted.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-protocol-config/src/lib.rs
```diff
@@ -545,6 +547,15 @@ struct FeatureFlags {
     // Used to gate genesis creation and epoch-change creation for existing networks.
     #[serde(skip_serializing_if = "is_false")]
     enable_claim_registry: bool,
+
+    // If true, the white-flag pass invalidates any transaction whose sender or
```
piotrm50, line 551: Do we need a separate feature flag? Whiteflag is only manually (not through protocol upgrade) on alphanet, so we can still change the logic freely. Unless you plan to enable this much later than the pcool flow, then it makes sense to have a separate gate for it.

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
```diff
@@ -1655,3 +1688,583 @@ async fn already_executed_tx_locked_by_different_digest_is_fatal() {
...
+        verified_plain.clone(),
+        epoch_store.epoch(),
+        1,
+    );
+    authority
+        .try_execute_immediately(&executable, None, &epoch_store)
+        .unwrap();
+
+    let mut transactions = vec![
+        make_user_tx_v1_verified(verified_claim),
```
piotrm50, line 2312: What happens if the plain transaction is earlier in consensus output? Should it also be dropped? We don't really have a guarantee that they will be sequenced in the same order for execution. We definitely should have a test covering this case as I see that in all tests so far the claim transaction is the first one.

crates/iota-core/src/post_consensus_validation.rs
```diff
@@ -95,216 +132,314 @@ pub async fn validate_and_resolve_conflicts(
...
+        debug!(
+            ?digest,
+            num_owned_inputs = owned_inputs.len(),
+            "Transaction already executed; retained as checkpoint root, skipping re-validation"
+        );
+        // keep[i] stays true so the transaction remains in the sequence.
+        return Ok(());
+    }
+
+    // Check Account-claim: a non-claiming transaction whose sender or sponsor
```
piotrm50, line 274: Should this be marked as `Check #2` for easy reference in the future?

crates/iota-core/src/unit_tests/post_consensus_validation_tests.rs
piotrm50: Another scenario that I think would be worth a test:
Plain consuming gas B
Claim X consuming gas A
Claim X consuming gas B
Plain sponsored by X consuming gas C
Plain consuming gas B (optionally).

miker83z: > What happens if the plain transaction is earlier in consensus output? Should it also be dropped? We don't really have a guarantee that they will be sequenced in the same order for execution. We definitely should have a test covering this case as I see that in all tests so far the claim transaction is the first one.

This opened a pandora vase I was working in these days. Please check the new commit: [9edc77e](https://github.com/iotaledger/iota/pull/12006/commits/9edc77ec51efde84f01b1d90d9ac144c2c5d3b73) and the updated PR description.

crates/iota-protocol-config/src/lib.rs
miker83z: I guess no then, but I wanted to be sure first.

PR #12009, opened 2026-06-23
fix(indexer): use fallback for empty descending pages of checkpoints

### LINKED ISSUE #12007: [indexer]: fallback is not always used on descending pages of pruned checkpoints
## Description

When resolving whether to use the fallback for descending pages of checkpoints we require incorrectly that the page returned from the database is non-empty.

https://github.com/iotaledger/iota/blob/53aa9d1f8f6644ec6a5c6909a41909a89a6eaa82/crates/iota-indexer/src/read.rs#L717-L720

But we should instead use the fallback also for empty pages returned from the database, if the cursor is greater than the genesis checkpoint.

kodemartin:
# Description of change

This patch refactors the logic that infers whether we should use the fallback when querying pages of checkpoints.

Fixes #12007 

Plus manual tests.

### Release Notes

- [x] Indexer: :screwdriver: Fixes a bug where the queries on checkpoint data with descending order would not used the fallback in completely pruned ranges of data.

### DIFF
--- a/crates/iota-indexer/src/read.rs
+++ b/crates/iota-indexer/src/read.rs
@@ -3466,3 +3466,175 @@ enum TransactionFilterKind {
     V1(TransactionFilter),
     V2(TransactionFilterV2),
 }
+
+#[cfg(test)]
+mod tests {
+    use std::num::NonZeroUsize;
+
+    use iota_json_rpc_types::Checkpoint;
+
+    use super::IndexerReader;
+
+    fn dummy_checkpoint(sequence_number: u64) -> Checkpoint {
+        Checkpoint {
+            epoch: 0,
+            sequence_number,
+            digest: Default::default(),
+            network_total_transactions: 0,
+            previous_digest: None,
+            epoch_rolling_gas_cost_summary: Default::default(),
+            timestamp_ms: 0,
+            end_of_epoch_data: None,
+            transactions: vec![],
+            checkpoint_commitments: vec![],
+            validator_signature: Default::default(),
+        }
+    }
+
+    /// Wraps `IndexerReader::should_fetch_checkpoints_from_fallback` by passing
+    /// dummy checkpoints with the given sequence numbers.
+    fn should_fetch_checkpoints_from_fallback(
+        cursor: Option<u64>,
+        descending_order: bool,
+        limit: usize,
+        sequence_numbers: &[u64],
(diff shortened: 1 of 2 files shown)


crates/iota-indexer/src/read.rs
```diff
@@ -697,32 +698,33 @@ impl IndexerReader {
...
-            // if first checkpoint matches cursor + 1, data is contiguous from that point.
-            (Some(c), false) => db_response
-                .first()
-                .is_none_or(|chk| chk.sequence_number != c + 1),
+        if db_checkpoints.len() == limit.get() && descending_order {
+            return false;
+        }
+        match db_checkpoints {
+            [] => descending_order && cursor.is_none_or(|seq| seq > 0),
+            [.., lowest_found] if descending_order => lowest_found.sequence_number > 0,
```
tomxey, line 731: could we add to the function doc the assumption that `db_checkpoints` are expected to be sorted according to `descending_order`?

crates/iota-indexer/src/read.rs
```diff
@@ -697,32 +698,33 @@ impl IndexerReader {
...
+    ///
+    /// Depending on the sort type (`descending_order`) we identify the
+    /// following cases:
+    ///
+    /// * Ascending: Should use fallback only when the lowest found checkpoint
+    ///   in a partial or full page does not match the cursor + 1, or genesis.
+    ///   We always expect checkpoint data to be available in the database, so
+    ///   an empty page indicates that the given cursor is too high.
+    /// * Descending: Should use fallback when a partial page does not reach
+    ///   genesis, or if the cursor is not the genesis checkpoint.
```
tomxey: "or if the cursor is not the genesis checkpoint." this looks like fallback should be triggered every time cursor is not genesis

Consider:

> Should use fallback only for a partial page that doesn't reach genesis, or for an empty page where the cursor is not genesis.

sergiupopescu199 (approved):

crates/iota-indexer/src/read.rs
kodemartin: 371189c154

crates/iota-indexer/src/read.rs
kodemartin: 371189c154

tomxey (approved):

PR #12021, opened 2026-06-24
upstream(core): Add labeled in-flight gauge guard to iota-metrics

jkrvivian:
## Description of change

- Port commit:
  - [MystenLabs/sui@b6a2a2a](https://github.com/MystenLabs/sui/commit/b6a2a2a2758e23eac19a24dcb3e74b16a56c4f4e)
- Description:

Only the `iota-metrics` changes are applicable. The rest of the changes we do not have (e.g., no fastpath, no per-transaction certifier)

Ported `IntGaugeVecGuard` and `GaugeGuardFutureExt::count_in_flight_with_labels`, which generalize the existing unlabeled gauge guard to labeled `IntGaugeVec` metrics.
A future is wrapped so the labeled gauge is incremented on construction and
decremented exactly once when the future completes or is dropped.

Part of #11885

### DIFF
--- a/crates/iota-metrics/src/guards.rs
+++ b/crates/iota-metrics/src/guards.rs
@@ -8,7 +8,7 @@ use std::{
     task::{Context, Poll},
 };
 
-use prometheus::IntGauge;
+use prometheus::{IntGauge, IntGaugeVec};
 
 /// Increments gauge when acquired, decrements when guard drops
 pub struct GaugeGuard<'a>(&'a IntGauge);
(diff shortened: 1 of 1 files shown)


bingyanglin (approved):
Just a reminder that `IntGaugeVecGuard` intorduced in this PR was then replaced by `InflightGuard` in https://github.com/MystenLabs/sui/pull/22647 later.

crates/iota-metrics/src/guards.rs
```diff
@@ -30,9 +30,42 @@ impl Drop for GaugeGuard<'_> {
     }
 }
 
+/// Increments an `IntGaugeVec` for a set of label values when acquired,
+/// decrements the same labeled gauge when the guard drops.
+pub struct IntGaugeVecGuard<'a> {
+    gauge: &'a IntGaugeVec,
+    labels: Vec<String>,
```
semenov-vladyslav, line 37: Wouldn't `labels: &'a [&'a str],` work?

crates/iota-metrics/src/guards.rs
```diff
@@ -30,9 +30,42 @@ impl Drop for GaugeGuard<'_> {
...
+            .dec();
+    }
+}
+
 pub trait GaugeGuardFutureExt: Future + Sized {
     /// Count number of in flight futures running
     fn count_in_flight(self, g: &IntGauge) -> GaugeGuardFuture<'_, Self>;
+
+    /// Count number of in flight futures running, tracked on a labeled gauge.
+    fn count_in_flight_with_labels<'a>(
```
semenov-vladyslav, line 64: New method added but not used. Is it for the future?

semenov-vladyslav (approved):

crates/iota-metrics/src/guards.rs
jkrvivian: It's used in the benchmark, I also opened a PR there and requested your review. 🙌

crates/iota-metrics/src/guards.rs
jkrvivian: It works, but the caller needs to keep the label array alive for the guard's whole lifetim. So inline arrays labels no longer work. `Vec<String>` costs one small alloc but frees callers from any lifetime burden, and keeps us aligned with upstream.

semenov-vladyslav: Ok, I see now, how it is used.

semenov-vladyslav (approved):

PR #12023, opened 2026-06-24
fix(iota-genesis-builder): make native-token lookup deterministic

### LINKED ISSUE #11867: Flaky test: `basic_migration_with_native_tokens` fails non-deterministically due to HashMap iteration order
## Summary

The test `iota_genesis_builder::stardust::migration::tests::basic::basic_migration_with_native_tokens` failed intermittently in CI on PR #11697 (attempt 1 of the "Run rust tests" job, job ID `80802386929`). The run was rerun and passed on attempt 2. The failure was a Rust `unwrap()` panic at `crates/iota-genesis-builder/src/stardust/migration/tests/basic.rs:352`.

## Failure evidence

- **PR**: #11697 (merged)
- **Workflow run**: https://github.com/iotaledger/iota/actions/runs/27347670975
- **Failed job** (attempt 1): https://github.com/iotaledger/iota/actions/runs/27347670975/job/80802386929
- **Backtrace**:
  ```
  4: iota_genesis_builder::stardust::migration::tests::basic::basic_migration_with_native_tokens
             at ./src/stardust/migration/tests/basic.rs:352:6
  5: iota_genesis_builder::stardust::migration::tests::basic::basic_migration_with_native_tokens::{{closure}}
             at ./src/stardust/migration/tests/basic.rs:308:40
  ```

## Root cause analysis

Two sources of non-determinism interact in this test:

1. **`random_output_header()`** is called for each of the 10 foundry outputs created in the test loop (lines 320–337). This gives each native token a random Object ID, making the order of objects in the execution result's `written` map unpredictable across runs.

2. **`written.values().find(...)`** inside `extract_native_tokens_from_bag()` (in `tests/mod.rs`) iterates over a `HashMap`'s values in an arbitrary order to locate the expected coin object. When combined with the randomised Object IDs above, the lookup can fail to match the correct coin object and returns `Err("missing coin object")` — which then propagates to the `.unwrap()` at `basic.rs:352`.

## Expected behaviour

The test should pass deterministically regardless of Object ID assignment or HashMap iteration order. The lookup logic should not rely on HashMap iteration order to correlate native-token coins with their expected types.

## Suggested fix

Replace the `HashMap::values().find(...)` approach with a lookup keyed on a stable identifier (e.g. the struct tag or token ID) so that the result does not depend on iteration order. Alternatively, use a deterministic data structure (`BTreeMap`) for the `written` map if iteration order matters to the test assertions.

minato32:
Closes #11867.

`extract_native_tokens_from_bag` searched the written objects via `HashMap::values().find(...)`, so the result depended on iteration order. Combined with the randomised foundry output IDs in `basic_migration_with_native_tokens`, this made the lookup non-deterministic and the test flaky in CI.

Indexed the written objects by their `StructTag` in a `BTreeMap` and look the coin up by key, so the result no longer depends on iteration order.


Thoralf-M:
Thanks for digging into this, @minato32 🙏

Tracing the actual CI failure from #11867, the panic doesn't come from the coin lookup this PR rewrites. The backtrace bottoms out in `run_migration` (`tests/mod.rs:253`) → `migrate_foundries` → `generate_package` → `build_and_compile`, where the Move build of the generated foundry package failed while resolving its `Iota` git dependency:

```
Caused by:
    0: Fetching 'Iota'
    1: Failed to reset to latest Git state 'framework/genesis/mainnet' for package 'Iota' | Exit status: 128
```

That happens *before* `extract_native_tokens_from_bag` ever reaches the `written` lookup (`tests/mod.rs:365+`), so swapping `find(...)` for a `BTreeMap` doesn't change the outcome of the failing run. The "HashMap iteration order" diagnosis came from the issue itself (a misread of the backtrace on my side) — not on you.

The refactor is harmless, but since it doesn't address the real (transient, git-dependency) failure, I'm going to close this. Really appreciate the effort, and sorry the issue pointed you the wrong way — hope to see you on other ones! More detail in #11867.

---

PR #12032, opened 2026-06-24
fix(move-analyzer): fix dangling process on LSP server shutdown

### LINKED ISSUE #10651: [move] Fix dangling move-analyzer process
Last time we had the same issue it was fixed with this -> #8685

Dkwcs:
[run-ci]

# Description of change

Fixes three independent scenarios where iota analyzer (the Move LSP server) leaves a dangling process after VS Code disconnects.

1. io_threads.join() deadlock (normal VS Code close)
When VS Code closes normally (shutdown + exit), the main loop exits via break and calls io_threads.join(). The LspServerWriter thread is blocked in into_iter() waiting for connection.sender to be dropped - but sender lives inside Context, which only drops after join() returns. Circular dependency -> permanent deadlock.
**Fix:** Declare Context inside a dedicated block scope { } that wraps the entire event loop. When the loop exits via break, the block ends and Context (including connection.sender) is dropped before io_threads.join() is called, unblocking the writer thread. The key constraint is that context must be owned by the block - declaring it outside and merely borrowing it inside does not trigger the drop at block exit.

2. Infinite Err spin (VS Code crash / abrupt disconnect)
When VS Code crashes and stdin closes without shutdown/exit, LspServerReader exits and drops its sender. recv() then returns Err(RecvError) permanently - the original Err arm only logged the error, causing the loop to spin at 100% CPU forever.
**Fix:** process::exit(-1) in the Err arm - the channel is permanently disconnected, no recovery is possible.

3. Orphaned process (pipe stays open): When VS Code is killed via Reload Window or SIGKILL, the stdin pipe may not receive EOF if another process in the chain holds the write-end. The analyzer hangs silently with 0% CPU, invisible without external monitoring.
**Fix**: SymbolicatorRunner receives initialize_params.process_id from the LSP handshake and polls sysinfo every 10 seconds. On detecting the parent is gone it triggers process::exit(-1).

Note: In addition PR contains upstream pull changes, the upstream commit (MystenLabs/sui@f5eb259b) addresses only Bugs 2 and 3. Bug 1 remains unresolved in Sui - verified by running our test suite against their binary.

Fixes #10651 .

Tested locally by building the binary with and without the patch and running all hang scenarios (graceful Shutdown+Exit, abrupt stdin close, orphaned processes) against the live process via a Python script. Both scenarios timed out (10s) without the fix and exited cleanly with it.
All test scenarios are described here - https://iotafoundation.atlassian.net/wiki/spaces/LSC/pages/3260284930/Move+Analyzer+Process+Lifecycle+-+Bugs+Analysis+Report.


lollobene (dismissed):

external-crates/move/crates/move-analyzer/src/analyzer.rs
```diff
@@ -208,6 +209,27 @@ pub fn run(implicit_deps: Dependencies) {
...
+/// shown in `lsp-server`'s own `examples/goto_def.rs`.
+fn main_loop(
+    context: Context,
+    ide_files_root: VfsPath,
+    pkg_deps: Arc<Mutex<BTreeMap<PathBuf, symbols::PrecomputedPkgInfo>>>,
+    diag_receiver: crossbeam::channel::Receiver<Result<BTreeMap<PathBuf, Vec<Diagnostic>>>>,
+    symbolicator_runner: &symbols::SymbolicatorRunner,
+    implicit_deps: Dependencies,
+) {
     let mut shutdown_req_received = false;
```
miker83z:
```suggestion
 // main loop
    {
        context
            .connection
            .initialize_finish(
                id,
                serde_json::json!({
                    "capabilities": capabilities,
                }),
            )
            .expect("could not finish connection initialization");

        let mut shutdown_req_received = false;
```

then close the bracket at the bottom:

```rust
                       Err(error) => {
                            eprintln!("IDE message error: {:?}", error);
                            // `error` is of type `RecvError`, which has only one meaning: the channel
                            // is empty and disconnected. Exit the process in such case.
                            std::process::exit(-1);
                        }
                    }
                }
            };
        }
    }

    // context is owned by the block above and dropped when it exits, before join()
    // is called. This unblocks the LspServerWriter thread which waits for
    // the sender to disconnect.
    io_threads.join().expect("I/O threads could not finish");
    symbolicator_runner.quit();
    eprintln!("Shut down language server '{}'.", exe);
    std::process::exit(0);
```

I think this is what @valeriyr was suggesting during the meeting and should make it easier to work with for future upstream pulls.

Dkwcs: Fixed

lollobene (approved):

miker83z (approved):
LGTM!

PR #11968, opened 2026-06-18
feat: replace `MoveAuthenticator` with SDK

thibault-martinez:
# Description of change

`MoveAuthenticator`, `MoveAuthenticatorInner`, and `MoveAuthenticatorV1` were defined locally in `iota-types`, duplicating client-visible types that belong in the `iota-rust-sdk`. Maintaining a second copy in the node is a drift hazard — the BCS/signature representation had to be kept byte-compatible with the SDK by hand. This PR makes the SDK the single source of truth: `iota-types` now re-exports `MoveAuthenticator` / `MoveAuthenticatorV1` from `iota_sdk_types` instead of defining them.

Node-internal behaviour that can't live in the SDK (input/receiving/shared object extraction, `validity_check`, `object_to_authenticate_components`) is moved into a sealed `MoveAuthenticatorExt` trait implemented for the SDK types. Callers are updated to the SDK API:

- `address()` now returns `Address` directly (was `IotaResult<Address>`), simplifying the matching logic in `authority.rs` and `transaction.rs`.
- `type_arguments()` → `type_args()`.
- Construction goes through `MoveAuthenticatorV1::new_shared(...).into()` instead of `MoveAuthenticator::new_v1(...)`.

The `iota-rust-sdk` dependency rev is bumped to `8820ff0…` to pick up the relocated types.

## How the change has been tested

- [ ] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

### Release Notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [x] Rust SDK: `MoveAuthenticator` and `MoveAuthenticatorV1` are now sourced from `iota-rust-sdk` (`iota_sdk_types`); node-internal helpers move to the `MoveAuthenticatorExt` trait.
- [ ] gRPC:

PR #11609, opened 2026-05-21
feat: replace `PasskeyAuthenticator` with SDK version

thibault-martinez:
# Description of change

Replaces the `PasskeyAuthenticator` that was maintained locally in `iota-types` with the canonical implementation from the `iota-rust-sdk` (`iota_sdk_types::crypto::PasskeyAuthenticator`), in line with the repo's direction of pushing client-visible types out into the SDK and importing them here rather than duplicating them.

`iota-types` now re-exports the SDK type and keeps only the node-side glue: the `AuthenticatorTrait` impl, which verifies directly against the SDK authenticator via `PasskeyVerifier` instead of round-tripping through `from_bytes` first. That round-trip was a temporary workaround (the `TODO` referencing #11607), so it is removed here. The local `RawPasskeyAuthenticator`, the hand-written `Serialize`/`Deserialize`/`ToFromBytes`/`AsRef` impls, `new_for_testing`, and the `to_signing_message` helper all go away — serialization and the signing digest (`IntentMessage::signing_digest`) now come from the SDK.

Call sites are updated accordingly: `get_pk()` becomes `public_key()` (with the secp256r1 conversion done where a `PublicKey` is still needed), `new_for_testing` becomes `new`, and test construction uses `SimpleSignature::from_bytes` instead of the old `Signature` type. The `iota-rust-sdk` rev is bumped (`35a2748` → `52cf443`) across the workspace and the example manifests to bring in the SDK type.

## Links to any relevant issues

fixes #11607

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

The existing passkey unit tests (`passkey_authenticator_test.rs`) and the passkey e2e/multisig tests were ported to the SDK API and continue to exercise serde round-trips, challenge/client-data validation, and signature verification (including the mismatched-challenge, wrong-author, and invalid-signature failure paths).

Issue #11590, opened 2026-05-20
Replacing SDK signatures aftermath


thibault-martinez:
Couple of things to do when signatures will have been replaced with SDK ones.
- [x] Remove aliases `use iota_sdk_crypto::Signer as SdkSigner;` → #12321
- [x] Remove `SimpleKeypair::from_bytes(&IotaKeyPair::Ed25519(kp).to_bytes()).unwrap()` conversions → #12321
- [x] Remove ` IotaKeyPair::from_bytes(&kp.to_bytes()).unwrap();` conversion → #12321
- [x] Remove ` IotaKeyPair::from(s_kp)` conversion → #12321
- [x] Replace the node `SignatureScheme` enum with the SDK one → #12342
- [x] Migrate the node keypairs to the SDK signing types and remove the `impl Into` signing funnel — its by-reference `From<&…KeyPair>` impls copy the private key on every conversion, including once per signature in the keystore ([discussion](https://github.com/iotaledger/iota/pull/12034#discussion_r3559374230)) → #12369
- [x] Can replace `impl Into` params with `&impl iota_sdk_crypto::Signer` then? → #12369
  - [x] then also switch `utils::make_transaction` back to `Transaction::from_data_and_signer` instead of hashing and signing manually ([#12321 discussion](https://github.com/iotaledger/iota/pull/12321#discussion_r3612849804)) → #12439
- [x] Remove `multisig_keys()` util? → #12649
- [x] Delete the `IotaSignature` trait once the #10724 `verify_secure`/`AuthenticatorTrait` work lands — it removes the remaining pieces, after which call sites use the SDK directly ([discussion](https://github.com/iotaledger/iota/pull/12034#discussion_r3560272036)) → #12696
- [ ] Migrate the network stacks (p2p/anemo, TLS, consensus-manager boundary) to the SDK key types — deletes `KeyPairWithPath`'s cached fastcrypto ed25519 view and the `network_to_simple_keypair`/`simple_to_network_keypair` conversions; starfish can keep its own key wrappers (conversion stays at the consensus-manager boundary) → #12808
- [x] Replace `IotaKeyPair` with the SDK `SimpleKeypair` — it is the same enum of the three simple-scheme private keys ([discussion](https://github.com/iotaledger/iota/pull/12369#discussion_r3656071351)) → #12439
[...]

Issue #10724, opened 2026-03-16
Replace `SenderSignedData` with SDK version


thibault-martinez:

Thoralf-M:
Prereq mapping from auditing the current state:

- `TransactionData` is already the SDK type (`iota_sdk_types::Transaction`), so the transaction half is done.
- The blocker is the signature half: node `SenderSignedTransaction.tx_signatures` is `Vec<GenericSignature>`, while the SDK's `SignedTransaction` / `SenderSignedTransaction` carry `Vec<UserSignature>`.
- Every `GenericSignature` variant except the simple `Signature` is now SDK-sourced (multisig, passkey #11609, move-auth #11968). The remaining `Signature → SimpleSignature` step is tracked in #12033.

**zkLogin handled here, not as its own migration.** The SDK's `UserSignature` has no zkLogin variant (removed upstream as unneeded). So rather than "migrating" it, the deprecated `GenericSignature::ZkLoginAuthenticatorDeprecated` (plus `PublicKey::ZkLoginDeprecated` and the `SignatureScheme::ZkLoginAuthenticatorDeprecated` flag `0x05`) is simply dropped as part of adopting `UserSignature` — i.e. treated as if the variant never existed. It was never constructible and deserialization already rejects the flag, so removal is safe; the release note can frame it as "there was never such a signature/transaction."

Resulting path:

1. #12033 — `Signature → SimpleSignature` (last non-SDK variant).
2. here — `GenericSignature → UserSignature` (drops the deprecated zkLogin variant) + `SenderSignedData → SignedTransaction` / `SenderSignedTransaction`.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #12034, opened 2026-06-24
feat: replace `Signature` with SDK `SimpleSignature`

### LINKED ISSUE #12033: Replace `Signature` with SDK `SimpleSignature`
Follow-up to the per-variant signature migration (#11609 passkey, #11968 move-auth). `Signature` is the last `GenericSignature` variant still defined locally in `iota-types` instead of being sourced from `iota-rust-sdk`.

### Why this is one unit (not three PRs)

`crates/iota-types/src/crypto.rs` defines `Signature` as an enum of three local newtypes — `Ed25519IotaSignature`, `Secp256k1IotaSignature`, `Secp256r1IotaSignature` — each an opaque `[flag || sig || pubkey]` byte array implementing `IotaSignatureInner`/`IotaSignature`.

The SDK collapses all three schemes into a single `SimpleSignature` enum (`iota-sdk-types`), whose variants are inline structured fields (`Ed25519 { signature, public_key }`, …), **not** standalone importable types. So unlike passkey / multisig / move-auth — each of which was one self-contained SDK type and got its own PR — there is no per-scheme SDK type to migrate to individually. The wrapper `Signature` enum and its three inner structs must be replaced together by the one `SimpleSignature` enum, in a single PR.

### Scope

- [ ] Replace `Signature` + `Ed25519IotaSignature` / `Secp256k1IotaSignature` / `Secp256r1IotaSignature` with `iota_sdk_types::SimpleSignature` (re-export, mirroring passkey/multisig/move-auth).
- [ ] Move the node-only verification surface onto the SDK type: the `IotaSignature` / `IotaSignatureInner` blanket impl + `verify_secure`, and `AuthenticatorTrait for Signature` (`signature.rs`). Route through the SDK's `Verifier` / `SimpleVerifier`, or a sealed `…Ext` trait as done for `MoveAuthenticator` in #11968.
- [ ] Reconcile `PublicKey` (`crypto.rs`) with `iota_sdk_types::PublicKey`, and the signing side (`IotaKeyPair: Signer<Signature>`) with `iota_sdk_crypto::SimpleKeypair` (`Signer<SimpleSignature>` / `Signer<UserSignature>`). The residual keypair-conversion cleanup is tracked in #11590.
- [ ] Follow through on `CompressedSignature` and `SignatureScheme`, which both reference the simple-signature representation.

Byte / BCS layout is identical on both sides (`flag || sig || pubkey`), so this is representation-preserving — verify BCS + JSON round-trips.

### Notes

Large and cross-cutting (~47 files reference `Signature`, ~35 reference `IotaKeyPair`), so noticeably bigger than the single-variant passkey/move-auth PRs.

Blocks #10724 (`SenderSignedData`). The wrapper swap `GenericSignature → UserSignature` — and with it the removal of the deprecated zkLogin variant, which `UserSignature` does not have — land with #10724, not here.

Thoralf-M:
> **SDK dependency:** iotaledger/iota-rust-sdk#1243 (the `SimpleSignature` accessors) has merged; the SDK rev is pinned to `a4c320f3` — the SDK `develop` tip containing that merge (and the versioned-`CheckpointContents` change #1256 that develop's #12191 adopted).

# Description of change

Part of #12033. Replaces the node-local `Signature` (the simple ed25519 / secp256k1 / secp256r1 signature) with a re-export of `iota_sdk_types::SimpleSignature` — the last `GenericSignature` variant not yet sourced from the SDK. Mirrors #11609 (passkey) and #11968 (move-authenticator).

Unlike those leaf types, the three per-scheme inner newtypes (`Ed25519IotaSignature`, `Secp256k1IotaSignature`, `Secp256r1IotaSignature`) collapse into the SDK's single `SimpleSignature` enum, so the wrapper and its variants are replaced **as one unit** — there is no standalone per-scheme SDK type to migrate each to.

Node-only behaviour moves into the `IotaSignature` extension trait implemented directly for `SimpleSignature`: the node `SignatureScheme` accessor, `new_secure` / `new_hashed` construction, and intent-message `verify_secure` (author/address binding + the SDK `SimpleVerifier`). A small `crypto::empty_signature()` helper replaces the old all-zero `Ed25519IotaSignature::default()` placeholder used for system transactions and tests.

Removing the node `Signature` type also removed the `Signer` impls on the per-scheme fastcrypto keypairs (the orphan rule forbids re-adding them for the SDK type). Signing is therefore funnelled through `IotaKeyPair` (partly addressing #11590): the signing entry points are now generic over `impl Into<IotaKeyPair>`, and by-reference `From<&…KeyPair>` impls let the ~230 keypair call sites stay unchanged. `IotaSignature::scheme()` is renamed to `signature_scheme()` so it is not shadowed by the SDK's inherent `SimpleSignature::scheme()`.

## Adopting the SDK accessors (iota-rust-sdk#1243)

Byte/public-key access uses the SDK's inherent `SimpleSignature::{signature_bytes, to_public_key}` instead of re-implementing per-variant matching here: `signature_bytes`/`public_key_bytes` are dropped from the `IotaSignature` trait, and `verify_secure` derives the author address via `to_public_key().into()`. This removes the `unimplemented!()` arms the `#[non_exhaustive]` enum otherwise forces on downstream matches. (Public-key bytes come from `to_public_key().as_ref()`; the raw signature keeps a `signature_bytes()` accessor because the SDK has no unified signature type.)

## Serialization / wire compatibility

Unchanged. `SimpleSignature::to_bytes()` yields the same `flag ‖ sig ‖ pubkey` layout, so BCS is byte-identical and `GenericSignature` (the only path user signatures travel) is unaffected. The SDK type's *direct* JSON form differs, but nothing serializes a bare `Signature` to JSON. Confirmed by the unchanged `iota-core/tests/staged/iota.yaml` format snapshot.

The manual `Serialize`/`Deserialize` impls on the node `Signature` are deleted with the type; #12055's symmetric byte-blob fix is inherent in the SDK impl (both directions use the byte-blob view).

Verified locally at the pre-rebase base against the pinned SDK rev (rebased onto `develop` after #12093 merged; CI re-validates on the new base):

- [x] `cargo check --workspace --all-targets` and `cargo clippy -p iota-types -p iota -p iota-json-rpc-types --all-targets` — both clean.
- [x] Unit tests: `iota-types` `--lib` (198), `iota` keytool (17), `iota-json-rpc-types` `--lib` (16) — includes the crypto serde round-trip and `verify_sender_signature_correctly_with_flag`.
- [x] BCS / `GenericSignature` compatibility: `iota-core` `test_format` snapshot passes against the unchanged `iota.yaml`.

### Release Notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: signatures in transaction output are shown as base64 (e.g. `Ed25519Signature("...")`) rather than raw byte arrays.
- [ ] Rust SDK:
- [ ] gRPC:


DaughterOfMars (approved):

tomxey (approved):

crates/iota-core/src/generate_format.rs
```diff
@@ -178,7 +178,7 @@ fn get_registry() -> Result<Registry> {
     tracer.trace_value(&mut samples, &sig).unwrap();
     // ... and the user signature which does
 
-    let sig: Signature = Signer::sign(&s_kp, b"hello world");
+    let sig: Signature = IotaKeyPair::from(s_kp.copy()).sign(b"hello world");
```
semenov-vladyslav: Copying private keys is considered an antipattern. Here `copy()` is not required, `s_kp` is not used afterwards.

crates/iota-indexer/tests/common/mod.rs
```diff
@@ -398,7 +394,7 @@ pub async fn execute_tx_and_wait_for_indexer_checkpoint(
     indexer_client: &HttpClient,
     store: &PgIndexerStore,
     tx_bytes: TransactionBlockBytes,
-    keypair: &dyn Signer<Signature>,
+    keypair: impl Into<IotaKeyPair>,
```
semenov-vladyslav, line 391: What is the reason for this pattern? It looks weird.

crates/iota-types/src/crypto.rs
```diff
@@ -186,11 +187,52 @@ impl Clone for IotaKeyPair {
...
+        Signature::from_bytes(&bytes).expect("Serialized signature did not have expected size")
+    }
+}
+
+// By-reference conversions into [`IotaKeyPair`], so the per-scheme keypairs
+// (and `IotaKeyPair` itself) can be passed to the signing helpers, which are
+// generic over `impl Into<IotaKeyPair>`.
+impl From<&Ed25519KeyPair> for IotaKeyPair {
+    fn from(kp: &Ed25519KeyPair) -> Self {
+        IotaKeyPair::Ed25519(kp.copy())
```
semenov-vladyslav, line 217: Right, copying private keys is possible, but not great.

semenov-vladyslav (approved):
The pattern `impl Into<IotaKeyPair>` leads to private keys copying which is an antipattern. Approving anyway since it does not introduce practical weaknesses.

crates/iota-core/src/generate_format.rs
Thoralf-M: Right — this call is the keypair's last use, so it is now moved instead of copied: d8fd7b0d.

---

crates/iota-indexer/tests/common/mod.rs
Thoralf-M: These helpers took `&dyn Signer<Signature>` before. With `Signature` now being the SDK `SimpleSignature`, the orphan rule forbids implementing `Signer<SimpleSignature>` for the fastcrypto keypairs (both types are foreign to this crate), so the per-scheme keypairs can no longer be passed as signers directly. Funnelling through `impl Into<IotaKeyPair>` — with by-reference `From<&…KeyPair>` conversions — keeps the many existing `&keypair` call sites compiling unchanged, while `IotaKeyPair` (a local type, so it *can* implement `Signer<Signature>`) does the signing.

---

crates/iota-types/src/crypto.rs
Thoralf-M: Agreed. Two mitigations in the current state: the copies are short-lived and fastcrypto's private keys are `ZeroizeOnDrop`, so each dropped copy is wiped; and the signing itself (`Signer<Signature> for IotaKeyPair`) works by reference — the copy happens only in the `Into` conversion. The copies disappear entirely once the node keypairs move to the SDK signing types (#11590), which implement `Signer<SimpleSignature>` directly and make the funnel unnecessary.

---

Thoralf-M: Or what else do you recommend here?

crates/iota-types/src/crypto.rs
```diff
@@ -777,178 +731,54 @@ impl IotaPublicKey for BLS12381PublicKey {
...
+/// Node-only behaviour layered on top of the SDK [`Signature`]
+/// (`iota_sdk_types::SimpleSignature`): the node [`SignatureScheme`] accessor,
+/// construction from a signer, and intent-message verification.
+pub trait IotaSignature: Sized {
+    /// The node [`SignatureScheme`] of this signature.
+    ///
+    /// Named distinctly from the SDK's inherent `SimpleSignature::scheme()`
+    /// (which returns `iota_sdk_types::SignatureScheme`) to avoid the inherent
+    /// method shadowing this trait method.
+    fn signature_scheme(&self) -> SignatureScheme;
```
thibault-martinez, line 759: We should investigate if it's possible to use the SDK one and remove the node one

crates/iota-types/src/crypto.rs
```diff
@@ -777,178 +731,54 @@ impl IotaPublicKey for BLS12381PublicKey {
...
-        // deserialize the signature
-        let signature = Self::Sig::from_bytes(self.signature_bytes()).map_err(|_| {
-            IotaError::InvalidSignature {
-                error: "Fail to get pubkey and sig".to_string(),
-            }
-        })?;
+/// Node-only behaviour layered on top of the SDK [`Signature`]
+/// (`iota_sdk_types::SimpleSignature`): the node [`SignatureScheme`] accessor,
+/// construction from a signer, and intent-message verification.
+pub trait IotaSignature: Sized {
```
thibault-martinez, line 753: I'm not sure we should have this trait tbh, doesn't really seem to be node-only

Thoralf-M: Agreed — none of this is inherently node-only. The trait exists so this PR doesn't have to do three more migrations at once:

- `new_secure` / `new_hashed`: the node still signs with the fastcrypto keypairs through `IotaKeyPair`. #11590 moves the keypairs to the SDK signing types, which already provide `sign_transaction` / `sign_personal_message` — these methods then go away.
- `signature_scheme()`: only exists because the node still has its own `SignatureScheme` enum (your other comment) — goes away with that enum.
- `verify_secure`: plugs into the node's `AuthenticatorTrait` / `IotaError` verification machinery, which is on the #10724 track.

Each follow-up removes its part; once all three land the trait is empty and gets deleted, and call sites use the SDK directly. I'd keep it for this PR and can reword the doc comment so it says "not yet migrated" rather than "node-only".

---

crates/iota-types/src/crypto.rs
Thoralf-M: Yes — checked in detail, it works as a follow-up:

- **Already compatible:** the flag bytes are identical, and both enums derive `strum::Display` with `serialize_all = "lowercase"`, so the differing variant names (`ED25519`/`Ed25519`, `MultiSig`/`Multisig`, …) produce the same strings — keytool output is unchanged.
- **Small additive SDK changes:** the SDK enum needs `strum::EnumString` (clap parses `key_scheme` CLI args via `FromStr`), and serde impls if the node enum's serialized form turns out to be persisted anywhere (needs one check — the node derives `Serialize/Deserialize` with variant-name strings). `update_hasher_with_flag` is address-derivation logic and stays node-side as a free function.
- **The one blocker:** `ZkLoginAuthenticatorDeprecated` (flag 5) — the SDK intentionally has no flag-5 variant. The node only uses it in rejection/deprecation paths (`key_derive`, `keytool`, `transaction.rs`), so either wait for #10724 (which deletes all zkLogin remnants) or rewrite those few sites to match the raw `0x05` byte now.
- Plus the mechanical `ED25519` → `Ed25519` renames across the repo, same flavor as the `SequenceNumber` → `Version` PR.

So: one small additive SDK PR + one node swap PR, sequenced after #10724 (or immediately with the flag-byte rewrite). Out of scope here — it would triple this diff — but it also removes the `signature_scheme()`/`scheme()` naming split this PR had to introduce. I'd track it as its own issue.

---

Thoralf-M: Added a comment to https://github.com/iotaledger/iota/issues/11590

crates/iota-types/src/crypto.rs
```diff
@@ -666,108 +708,20 @@ where
...
-                    Err(FastCryptoError::InvalidInput)
-                }
-            }
-            _ => Err(FastCryptoError::InvalidInput),
-        }
-    }
+/// An all-zero ed25519 [`Signature`] placeholder, used for system transactions
+/// (which are not signed) and in tests where the signature content is
+/// irrelevant.
+pub fn empty_signature() -> Signature {
```
thibault-martinez: Two things:
- I would mention `ed25519` in the name
- not sure empty is the correct word

crates/iota-types/src/unit_tests/utils.rs
```diff
@@ -313,12 +313,12 @@ mod passkey {
     /// WebAuthn round-trip.
     pub fn make_passkey_authenticator_sig() -> GenericSignature {
         let (_, r1_kp): (_, Secp256r1KeyPair) = get_key_pair();
-        let user_sig: Signature = r1_kp.sign(&[0u8; 32]);
+        let user_sig: SimpleSignature = IotaKeyPair::from(&r1_kp).sign(&[0u8; 32]);
```
thibault-martinez: My problem with doing things in different steps is that we would not necessarily revert to the previous state, which was slightly better

crates/iota-types/src/crypto.rs
Thoralf-M: Renamed https://github.com/iotaledger/iota/pull/12034/commits/5655a00724388e8c1b1e098bd3e3ccaa15c2a2ae

crates/iota-types/src/unit_tests/utils.rs
Thoralf-M: You mean replacement of the different types or what?

thibault-martinez: We could have signed with Secp256r1KeyPair already, we should not have to create an IotaKeyPair

Thoralf-M: https://github.com/iotaledger/iota/pull/12034/changes/333df7c99d18f19d9eadb32071895eb511197486

crates/iota-ledger-signer/examples/ledger_signer.rs
```diff
@@ -77,7 +74,13 @@ async fn main() -> Result<(), anyhow::Error> {
     println!("Signer address: {}", &address);
 
     let signed_tx = signer.sign_transaction(&transaction, &address).await?;
-    println!("Signature: {}", signed_tx.signature.encode_base64());
+    println!(
+        "Signature: {}",
+        base64::Engine::encode(
+            &base64::engine::general_purpose::STANDARD,
+            signed_tx.signature.to_bytes()
+        )
```
thibault-martinez: Don't use the engine

crates/iota-ledger/examples/ledger_sign_tx.rs
```diff
@@ -92,7 +92,13 @@ pub fn main() -> Result<()> {
         &transaction,
         objects,
     )?;
-    println!("Signature: {}", &signature.signature.encode_base64());
+    println!(
+        "Signature: {}",
+        base64::Engine::encode(
+            &base64::engine::general_purpose::STANDARD,
+            signature.signature.to_bytes()
+        )
```
thibault-martinez: Don't use the engine

crates/iota-ledger-signer/examples/ledger_signer.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/12034/changes/f5a655fe460844b617210b69742b78a441886dc5

crates/iota-ledger/examples/ledger_sign_tx.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/12034/changes/f5a655fe460844b617210b69742b78a441886dc5

crates/iota-ledger-signer/examples/ledger_signer.rs
thibault-martinez: Still using the engine

crates/iota-ledger/examples/ledger_sign_tx.rs
thibault-martinez: Still using the engine

crates/iota-ledger-signer/examples/ledger_signer.rs
Thoralf-M: Well but cleaner and otherwise there is no way with this crate, changed to fastcrypto now

crates/iota-ledger/examples/ledger_sign_tx.rs
Thoralf-M: https://github.com/iotaledger/iota/pull/12034/changes/617357a18ea2aa1221697e6796f4e458b4e30a81

crates/iota-ledger-signer/examples/ledger_signer.rs
```diff
@@ -77,7 +75,10 @@ async fn main() -> Result<(), anyhow::Error> {
     println!("Signer address: {}", &address);
 
     let signed_tx = signer.sign_transaction(&transaction, &address).await?;
-    println!("Signature: {}", signed_tx.signature.encode_base64());
```
thibault-martinez, line 80: Maybe we should have `to_base64`/`from_base64` in the SDK tho...

Thoralf-M: Sure, adding it

crates/iota-core/src/unit_tests/auth_unit_test_utils.rs
```diff
@@ -69,7 +69,7 @@ pub fn build_test_modules_with_dep_addr(
 pub async fn publish_package_on_single_authority(
     path: &Path,
     sender: Address,
-    sender_key: &dyn Signer<Signature>,
+    sender_key: impl Into<IotaKeyPair>,
```
thibault-martinez, line 72: Just to be clear, when everything is done, at the end of the process, what type do you think this would be?

Thoralf-M: Near term it should stay `IotaKeyPair`
Long term we can think about moving the crypto stuff from the iota repo also to the SDK, then it might change, not sure

Thoralf-M: Could also become `sender_key: &impl iota_sdk_crypto::Signer<UserSignature>` after `Remove  IotaKeyPair::from_bytes(&kp.to_bytes()).unwrap(); conversion`https://github.com/iotaledger/iota/issues/11590 is done

thibault-martinez: My question was hinting at your second option. Can you take a note somewhere please?

Thoralf-M: Edited the entry in https://github.com/iotaledger/iota/issues/11590

crates/iota/src/keytool.rs
```diff
@@ -834,7 +836,7 @@ impl KeyToolCommand {
                     intent,
                     raw_intent_msg,
                     digest: Base64::encode(digest),
-                    iota_signature: iota_signature.encode_base64(),
+                    iota_signature: Base64::encode(iota_signature.to_bytes()),
```
thibault-martinez: Use the new method, also while you're at it, let's rename iota_signature to signature

Thoralf-M: Why rename it? Can do, but it's a user facing change in the CLI

crates/iota-ledger-signer/examples/ledger_signer.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1263

crates/iota-types/src/crypto.rs
Thoralf-M: @semenov-vladyslav can this be resolved now or do you want a change?

thibault-martinez (commented):
There is some docs that mention `Ed25519IotaSignature`

crates/iota/src/keytool.rs
thibault-martinez: Ah, forget about it then. Idk, iota_signature is just weird.

Thoralf-M: I think the reason for the iota_ prefix was that it adds the type byte so if you use a third party tool and put the base64 string as signature in it will be invalid since the iota specific flag byte needs to be removed first

crates/iota-types/src/unit_tests/utils.rs
```diff
@@ -312,15 +308,11 @@ mod passkey {
...
+        let user_sig: SimpleSignature = r1_kp.sign(&[0u8; 32]);
         let client_data_json = r#"{"type":"webauthn.get","challenge":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","origin":"https://test.iota.org"}"#;
-        let passkey = PasskeyAuthenticator::new(
-            vec![],
-            client_data_json.to_string(),
-            SimpleSignature::from_bytes(user_sig.as_bytes()).unwrap(),
-        )
-        .unwrap();
+        let passkey =
+            PasskeyAuthenticator::new(vec![], client_data_json.to_string(), user_sig).unwrap();
```
thibault-martinez, line 315: Maybe take a note for later: at some point `get_key_pair` would return an SDK key and we could revert to the previous state, I think it's slightly better for tests than a random one

thibault-martinez: Ah but it also returns a random one, I thought it was like Multisig and these were static

thibault-martinez: Don't mind me

lollobene (approved):

valeriyr (approved):

crates/iota-types/src/crypto.rs
semenov-vladyslav: Looks good.

thibault-martinez (approved):

PR #12037, opened 2026-06-24
fix(starfish): process commit observer recovery in bounded batches

### LINKED ISSUE #12036: Process commit observer recovery in bounded batches
During recovery, the commit observer's `recover_linearizer_and_solidifier_state` scans the entire `recovery_start..=last_commit_index` range into a single `Vec` and accumulates one `PendingSubDag` per commit before seeding the commit solidifier. When the unprocessed range is large — e.g. transaction data lagged behind commits for an extended period so commits stay unprocessed — the whole range is materialized in memory at once at startup.

`resend_unprocessed_solid_commits` already processes its range in bounded `COMMIT_RECOVERY_BATCH_SIZE` batches; the linearizer/solidifier recovery phase should do the same so recovery memory does not scale with the size of the unprocessed backlog.

**Key points:**
- The linearizer rebuild window is already bounded (`gc_depth * 2`); the unbounded part is the solidifier-seeding range, `last_sent_commit_index + 1 ..= last_commit_index`.
- Processing the range in bounded batches keeps the transient scan and per-batch pending buffer at `O(batch)` rather than `O(range)`.

polinikita:
# Description of change

During recovery, the commit observer's `recover_linearizer_and_solidifier_state` scanned the entire `recovery_start..=last_commit_index` range into a single `Vec<TrustedCommit>` and accumulated one `PendingSubDag` per commit before seeding the commit solidifier. When the unprocessed range is large (e.g. transaction data lagged behind commits for a long period so commits stay unprocessed), the whole range was materialized in memory at once at startup.

This processes the range in bounded batches, mirroring `resend_unprocessed_solid_commits`: each batch scans one chunk, rebuilds linearizer state for commits in the `gc_depth * 2` window, seeds the solidifier, and finalizes solid sub-dags. The solidifier cursor is set once before the loop; disjoint, ascending, contiguous batches feed the solidifier exactly as a single pass would, so the recovered linearizer/solidifier state and the set and order of sub-dags sent to the consumer are unchanged — only peak memory during recovery is reduced (the transient scan and the per-batch pending buffer become `O(batch)` instead of `O(range)`).

The batch size is a `commit_recovery_batch_size` local parameter (default 250; 3 under the simulator), shared by both recovery phases. `Parameters::validate` rejects a zero value, which would otherwise panic `step_by(0)`.

The commit solidifier's internal pending-sub-dag buffer still holds the unprocessed range while the front of that range is missing transactions; that is inherent to seeding (identical to the previous single-pass behaviour) and is a separate, tail-side concern.

Fixes #12036

### DIFF
--- a/crates/starfish/config/src/parameters.rs
+++ b/crates/starfish/config/src/parameters.rs
@@ -106,6 +106,12 @@ pub struct Parameters {
     #[serde(default = "Parameters::default_commit_sync_batches_ahead")]
     pub commit_sync_batches_ahead: usize,
 
+    /// Maximum number of commits scanned and replayed per batch during
+    /// recovery, bounding peak memory when a large unprocessed range is
+    /// replayed at startup.
+    #[serde(default = "Parameters::default_commit_recovery_batch_size")]
+    pub commit_recovery_batch_size: u32,
+
     /// Maximum number of headers to be included in a bundle. Headers exceeding
     /// the max allowed limit will be truncated.
     #[serde(default = "Parameters::default_max_headers_per_bundle")]
@@ -257,6 +263,10 @@ impl Parameters {
                 self.commit_sync_batch_size as u128,
             ),
             (
+                "commit_recovery_batch_size",
+                self.commit_recovery_batch_size as u128,
+            ),
+            (
                 "commit_sync_batches_ahead",
                 self.commit_sync_batches_ahead as u128,
             ),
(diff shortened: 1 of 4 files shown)


filipdulic (approved):

piotrm50 (approved):
I was thinking about it when I implemented the first part of batching. This is only problematic if `last_sent_commit_index` is old relative to the `last_commit_index` which can only happen when we have missing transaction data on an old commit - this situation would be problematic for all the nodes, not just recovering ones. Anyway, good to have this covered as well.

```
let linearizer_recovery_start = last_commit_index
            .saturating_sub(self.context.protocol_config.gc_depth() * 2)
            .max(1);
        let solidifier_recovery_start = self.last_sent_commit_index.saturating_add(1);
        let recovery_start = linearizer_recovery_start.min(solidifier_recovery_start);
```

Issue #12030, opened 2026-06-24
feat(starfish): shared peer-responsiveness ranking for synchronizer peer selection


polinikita:
> **Epic.** This work is split into sub-issues so the transactions-synchronizer ranking can be tuned in isolation before extending it:
>
> - #12112 — transactions synchronizer (implemented by #12038)
> - #12113 — commit syncer + header synchronizer (follow-up)

---

Consensus synchronizers currently select peers without any positive responsiveness signal. The transactions synchronizer shuffles candidate acknowledgers uniformly and only excludes recently-failed peers; the per-fetch latency it already measures is observed but never fed back into selection. The commit syncer likewise does not rank peers by responsiveness. On a node with slow or asymmetric inbound links this regularly draws slow-but-not-failing peers, adding avoidable latency to payload/commit retrieval — which on such a node propagates into delayed local execution and checkpoint signing.

Introduce a shared consensus component that tracks per-peer responsiveness and exposes a fastest-peer ranking, and use it as a prioritization signal when selecting peers in both the transactions synchronizer and the commit syncer.

**Key points:**
- One reusable peer-responsiveness/ranking component owned by consensus, fed by the latency signals the synchronizers already produce.
- Used by the transactions synchronizer to prefer responsive peers among the eligible acknowledgers when fetching missing committed payloads.
- Used by the commit syncer to prefer responsive peers when fetching commits.
- Ranking is a preference within the already-eligible candidate set, not a relaxation of it — existing safety/liveness bounds (quorum/acknowledger constraints, the f+1-stake failure exclusion) stay intact.
- Resistant to gaming: a peer that appears fast must not be able to monopolize selection or be trusted to deliver — retain randomized tie-breaking and fallback/rotation.

PR #12038, opened 2026-06-24
feat(starfish): rank transaction-synchronizer peers by responsiveness

### LINKED ISSUE #12112: feat(starfish): rank transaction-synchronizer peers by responsiveness
Sub-issue of #12030.

Track per-peer responsiveness and use it to prioritize peer selection in the **transactions synchronizer** only. Scope is deliberately limited to the transactions synchronizer so the ranking can be tuned in isolation before extending it to the other synchronizers.

### Scope
- Shared per-epoch `PeerResponsiveness` tracker fed by the fetch latency/goodput signals the transactions synchronizer already produces.
- Prefer responsive peers among the eligible acknowledgers when fetching missing committed payloads.
- Preference **within the already-eligible candidate set only** — the f+1-by-stake failure exclusion, acknowledger/quorum bounds, and full response verification stay intact, so it cannot affect consensus safety.
- Resistant to gaming: weighted-random (not deterministic) selection, delivery-shortfall scaling so a peer cannot look fast by returning only part of what was requested, and a uniform-exploration floor to avoid monopolization/starvation.
- Node-local `enable_peer_responsiveness_ranking` gate; off = prior behavior exactly.
- Observability: `peer_responsiveness_expected_latency_ms{kind,selection}` gauge exposing expected uniform-vs-weighted fetch latency.

### Out of scope (tracked separately)
- Commit syncer and header synchronizer — see the sibling sub-issue.

Implemented by #12038.

polinikita:
# Description of change

The transactions synchronizer shuffled its eligible fetch peers uniformly and threw away the per-fetch latency it already measures, so on nodes with slow or asymmetric links it kept drawing slow-but-not-failing peers, adding avoidable latency to committed-payload retrieval.

This PR adds a per-epoch `PeerResponsiveness` tracker (`crates/starfish/core/src/peer_responsiveness.rs`) that turns those existing latency/goodput signals into a smoothed per-peer "effective latency" and reorders the **transactions synchronizer's** already-eligible candidate set so responsive peers are tried first.

This is the first slice of the epic (#12030); the commit syncer and header synchronizer are intentionally left untouched here and tracked separately in #12113, so the ranking can be tuned on the transactions synchronizer in isolation before being extended.

Key points:

- **Ranking:** EWMA over successes; asymmetric demotion (a failure jumps to a penalty/timeout latency, recovery is slow). Selection sorts healthy peers by effective latency, permutes the 5 best by weighted sampling (weight `1/effective latency`) - the same selection shape as the checkpoint state-sync's `PeerBalancer`, with a weighted draw inside the window - and keeps the remaining healthy peers in latency order; a peer whose most recent fetch failed is ordered behind every healthy candidate until it succeeds again. A 5% uniform-shuffle exploration floor keeps untried/recovering peers sampled; it also bounds the wait for a transaction whose only acknowledger ranks outside the window (a latency-only tradeoff, exploration guarantees eventual selection).
- **Goodput:** a fetch's latency is scaled by its delivery shortfall (`requested/matched`), so a peer can't look fast by returning only part of what was requested.
- **Safety:** preference-only — the output is always a permutation of the already-eligible input, so the acknowledger/quorum bounds and full response verification are unchanged; it cannot affect consensus. When ranking is on, the previous hard exclusion of recently failed acknowledgers (`LastFailureByPeer`, up to <f+1 by stake) is replaced by a stricter in-ranking rule: a failed peer is deterministically ordered last while remaining in the set, so it is still verified and still sampled by exploration.
- **Gating:** node-local `enable_peer_responsiveness_ranking`, default `true`. Off restores the previous selection: a uniform random shuffle that excludes the most recently failed peers (up to <f+1 by stake); this legacy fallback carries a TODO for removal once the ranking has proven itself in production. Not a protocol-config flag, so it cannot diverge consensus.
- **Observability:** `peer_responsiveness_expected_latency_ms{kind,selection}` gauge (expected uniform-vs-weighted fetch latency), plotted in a new Peer Responsiveness dashboard section.

Fixes #12112
Part of #12030


muXxer (approved):
Approving `dev-tools/grafana-local/dashboards/starfish-overview.json`. Maybe we should add consensus as codeowner for that dashboard?

crates/starfish/core/src/transactions_synchronizer.rs
```diff
@@ -711,25 +655,34 @@ impl<C: NetworkClient, D: CoreThreadDispatcher> TransactionsSynchronizer<C, D> {
...
+        > = if context.parameters.enable_peer_responsiveness_ranking {
+            let mut vec: Vec<_> = blocks_by_authority.into_iter().collect();
+            let mut order: Vec<AuthorityIndex> =
+                vec.iter().map(|(authority, _)| *authority).collect();
+            context.peer_responsiveness.prioritize(
+                DataSource::TransactionSynchronizer,
+                &mut order,
+                &mut rng,
+            );
+            let position: HashMap<AuthorityIndex, usize> =
```
VorobyevIlya: Creating a HashMap for sorting looks like overkill

VorobyevIlya (commented):
I am not sure about the deterministic ordering among first-tier peers. Does it mean that if we have some validators that are super fast, everyone will request transactions from them? Would it put a burden on them? Maybe some randomization here is ok?

Also I expected flag off to restore the previous logic, but it is not, it is something simpler (worse)? Why is that?

crates/starfish/core/src/transactions_synchronizer.rs
polinikita: Right!

polinikita:
> I am not sure about the deterministic ordering among first-tier peers. Does it mean that if we have some validators that are super fast, everyone will request transactions from them? Would it put a burden on them? Maybe some randomization here is ok?

Added a randomization (inverse to latency) among top 5 authorities

> Also I expected flag off to restore the previous logic, but it is not, it is something simpler (worse)? Why is that?

Now restore the behaviour when the flas is off.

VorobyevIlya (approved):

PR #12039, opened 2026-06-24
feat(snapshot): extend VerifiedEpochInfo API

kodemartin:
# Description of change

This patch introduces three changes motivated by the parallel work on using the V2 api in the indexer.

1. It updates the progress bar on regular ticks, so it doesn't made for download completion to appear while restoring
2. Exposes a consuming `VerifiedEpochInfo::into_parts` method that extracts all inner values, so that custom indexing logic on epoch can be implemented efficiently.
3. Expose a getter to the `start_system_states` as a slice, to complement the existing API and make documentation more concise.

Part of #11023 

### DIFF
--- a/crates/iota-snapshot/src/lib.rs
+++ b/crates/iota-snapshot/src/lib.rs
@@ -321,10 +321,6 @@ impl EpochInfo {
 pub struct VerifiedEpochInfo {
     epoch_info: EpochInfo,
     committees: Vec<Committee>,
-    /// Digest-verified start state per epoch: `[i]` is epoch `i`'s start state
-    /// — the genesis root for `i == 0`, else derived from epoch `i - 1`'s
-    /// boundary. Length `snapshot_epoch + 2`; the trailing entry has no epoch
-    /// info row.
     start_system_states: Vec<IotaSystemState>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/iota-snapshot/src/lib.rs
```diff
@@ -340,6 +336,23 @@ impl VerifiedEpochInfo {
         &self.committees
     }
 
+    /// Digest-verified start system state per epochs `[0, snapshot_epoch + 1].
+    ///
+    /// Contains the genesis start system state, plus the ones derived by the
+    /// epoch boundaries represented by [`Self::entries`].
```
muXxer, line 350: The comment is missing the fact that it will be one more entry than the "epoch infos". Something the user should be aware of?

muXxer (approved):

crates/iota-snapshot/src/lib.rs
kodemartin: This applies to `committees` as well. I removed the particular mention for the start-system states, relying on the reference to the `[0, snapshot_epoch + 1]` for both the `committees` and the `start_system_states`.

`entries` are explicitly associated with the `[0, snapshot_epoch]` range, and they are now linked to `start_system_states` doc. 

So the fact about one more entry with respect to `entries` can be inferred, no?

kodemartin: Refined with b4d7dc99d3 nevertheless.

PR #12046, opened 2026-06-25
feat(core+node): add separate tokio runtime for serving API requests

muXxer:
# Description of change

## Problem

A validator runs everything on a single shared multi-threaded tokio runtime (iota_node). When the node signs/votes on a flood of incoming user transactions (creating certificates), those request-handler tasks saturate the shared worker threads, starving latency-critical node-core work — consensus commit processing and checkpoint signature creation — because all tasks compete for the same threads (tokio::spawn inherits Handle::current(), so request handlers and core tasks land on the same runtime).

## Approach

Carve all externally-triggered, client-facing server work onto a dedicated serving runtime, leaving the node core — consensus, execution, state sync, checkpoints, reconfiguration, and all peer-to-peer networking — alone on iota_node. Each client-facing server in the process has a single entry task (one tokio::spawn); routing that entry task onto the serving runtime moves the whole accept loop and every per-request handler with it, via runtime inheritance.

## The three commits:

1. add separate tokio runtime for serving API requests — Adds a serving runtime to IotaRuntimes, plumbs its Handle from main.rs through IotaNode, and re-homes the three client-facing servers onto it:
  - Validator client gRPC (transaction / handle_certificate) via SpawnOnce.
  - JSON-RPC HTTP server (fullnode).
  - gRPC read API server (fullnode).

The node core and p2p networking are untouched. The validator-to-validator consensus hot path runs on the separate Starfish/anemo network and is unaffected; the validator-peer gRPC carried on the moved listener is only a read-only GetCheckpoint plus rare control-plane RPCs.

2. add tokio runtime metrics — Adds per-runtime observability (iota-metrics/src/runtime_metrics.rs), exporting, labelled per runtime:
  - tokio_runtime_workers, tokio_runtime_alive_tasks, tokio_runtime_global_queue_depth (stable tokio RuntimeMetrics)
  - tokio_runtime_scheduler_lag_seconds — a heartbeat histogram measuring wake-up delay, the direct signal of worker-thread starvation.

Started for both iota_node and serving. Gated #[cfg(not(msim))] because the simulator's tokio fork lacks Handle::metrics() (mirrors thread_stall_monitor).

3. optimize runtime worker threads per node role — The core runtime always gets one worker per available core, so it can use the whole machine while serving is idle. The serving workers are added on top (validators: N/4, fullnodes: N/2), deliberately oversubscribing the machine: workers blocked in RocksDB free their core for the other runtime instead of leaving it idle, and when both runtimes are saturated the OS schedules fairly per runnable thread, so the serving pool's size relative to the core pool bounds the CPU share serving can take — at most ~20% of the machine on a validator (N vs N/4 threads), ~33% on a fullnode. This both protects the core under request floods and avoids throttling it when there is no request load at all.

Behavioral / config notes

- Thread counts change. The core runtime keeps one worker per core (as before the split); the serving runtime's workers come on top of that, so the process runs more threads than cores. Both counts are derived from available_parallelism.
- Request-triggered follow-ups (consensus submission, tx finalizer, owned-object cert execution) now run on the serving runtime — intentional backpressure: a flooded node serves/submits slightly slower while the core engine stays unaffected. Consensus-ordered and checkpoint execution run on the core and are protected.
- Thread isolation guarantees the core always has a worker ready to poll; it does not partition raw CPU. Under simultaneous saturation of both runtimes the core is no longer starved and holds ~80% of the CPU (validators), at the cost of some timeslicing between threads.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Client-facing request handling (validator gRPC, JSON-RPC, gRPC read API) now runs on a dedicated tokio runtime, isolated from node-core work (consensus, execution, checkpoints) so a flood of incoming transactions can no longer starve consensus commits or checkpoint signing. The core runtime keeps one worker per CPU core; serving threads are sized per node role and added on top. New tokio_runtime_* metrics expose per-runtime saturation.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -683,14 +683,6 @@ pub fn default_grpc_api_config() -> Option<GrpcApiConfig> {
     Some(GrpcApiConfig::default())
 }
 
-/// Number of CPU cores available to the process, falling back to 8 when it
-/// cannot be determined.
-pub fn available_cpu_cores() -> usize {
-    std::thread::available_parallelism()
-        .map(|n| n.get())
-        .unwrap_or(8)
-}
-
 pub fn default_concurrency_limit() -> Option<usize> {
     Some(DEFAULT_GRPC_CONCURRENCY_LIMIT)
 }
@@ -1522,11 +1514,6 @@ mod tests {
 
     use super::Genesis;
     use crate::NodeConfig;
-
-    #[test]
-    fn available_cpu_cores_is_never_zero() {
-        assert!(super::available_cpu_cores() >= 1);
-    }
 
     #[test]
     fn serialize_genesis_from_file() {
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -37,8 +37,16 @@ use crate::{
     transaction_deny_config::TransactionDenyConfig, verifier_signing_config::VerifierSigningConfig,
 };
 
-// Default max number of concurrent requests served
+// Effectively-unlimited concurrency, used by tests that must not be throttled.
 pub const DEFAULT_GRPC_CONCURRENCY_LIMIT: usize = 20000000000;
+
+// Per-core ceiling on concurrent in-flight validator gRPC requests. Most
+// request time is spent awaiting locks, I/O or consensus rather than on-CPU, so
+// the ceiling is generous; its purpose is to bound total in-flight work so a
+// request flood cannot grow queues and memory without limit, not to throttle
+// normal load. Operators wanting hard load-shedding can lower
+// `grpc_concurrency_limit` and set `grpc_load_shed`.
+const GRPC_CONCURRENCY_LIMIT_PER_CORE: usize = 1000;
 
 /// Default gas price of 1000 Nanos
 pub const DEFAULT_VALIDATOR_GAS_PRICE: u64 = iota_types::transaction::DEFAULT_VALIDATOR_GAS_PRICE;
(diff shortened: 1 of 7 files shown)


crates/iota-core/src/runtime.rs
```diff
@@ -6,15 +6,45 @@ use iota_config::NodeConfig;
...
 pub struct IotaRuntimes {
-    // Order in this struct is the order in which runtimes are stopped
+    // Order in this struct is the order in which runtimes are stopped.
+    /// Node core: consensus, execution, state sync, checkpoints and p2p
+    /// networking.
     pub iota_node: Runtime,
+    /// Client-facing servers (validator gRPC, JSON-RPC, gRPC read API) and the
+    /// per-request handlers they spawn. Isolated from `iota_node` so that a
+    /// flood of external requests cannot starve the node core.
+    pub serving: Runtime,
```
alexsporn, line 54: shouldn't we stop the serving runtime before the node?

crates/iota-metrics/src/runtime_metrics.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2024 IOTA Stiftung
```
alexsporn: 2026?

crates/iota-core/src/runtime.rs
muXxer: Fixed in d8d2e1eaee — `serving` is now the first field, so it is dropped before `iota_node` and client requests are cut off before the core they call into is torn down.

crates/iota-metrics/src/runtime_metrics.rs
muXxer: Fixed in d8d2e1eaee.

crates/iota-config/src/node.rs
```diff
@@ -673,6 +673,14 @@ pub fn default_grpc_api_config() -> Option<GrpcApiConfig> {
     Some(GrpcApiConfig::default())
 }
 
+/// Number of CPU cores available to the process, falling back to 8 when it
+/// cannot be determined.
+pub fn available_cpu_cores() -> usize {
```
jkrvivian: This should be in `crates/iota-core/src/runtime.rs`?  It's not configurable, and the only caller is in runtime.

jkrvivian (approved):

chriscandless (approved):

PR #12050, opened 2026-06-26
feat(move): Gate View attribute publication behind protocol version

### LINKED ISSUE #12048: `IotaAttribute::View` BCS variant is not protocol-gated
`IotaAttribute` is `enum { Authenticator(..) = 0, View = 1 }` and the new `View` variant is written into module bytecode metadata unconditionally by `fill_metadata` ([lib.rs:387](https://github.com/iotaledger/iota/tree/vm-lang/6548-move-view-functions/crates/iota-move-build/src/lib.rs#L387)) and accepted by the verifier ([runtime_module_metadata.rs:103](https://github.com/iotaledger/iota/tree/vm-lang/6548-move-view-functions/iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs#L103)). 

 A `#[view]`-bearing package becomes publishable as soon as the new binary is live, that is, before protocol v30. During the normal mixed-binary rollout window (binaries upgraded, protocol still v28/v29), a not-yet-upgraded validator running the prior binary (whose `IotaAttribute` enum has only the Authenticator variant) hits a hard `bcs::from_bytes` "unexpected variant" error whenever it deserializes such a package, while new-binary validators succeed.
This happens at publish verification, at upgrade-time `normalize_modules_with_metadata`
([move_package.rs:655](https://github.com/iotaledger/iota/tree/vm-lang/6548-move-view-functions/crates/iota-types/src/move_package.rs#L655)), or during state-sync/checkpoint replay. 

### Possible fix
- Use the protocol flag: gate acceptance in `runtime_module_metadata::verify_runtime_metadata` (push `ProtocolConfig` in) and emission in `fill_metadata` behind a protocol flag (reuse the existing flag) in order to reject `View` before when this flag is false so old and new binaries agree.
- Add a mixed-binary test: add a test that rejects `View` with a protocol version where the flag is false.

Additionally, perform a manual test by creating a move package with a view function using the new version of the sw (binaries built out of this feature branch), then try to publish it:
- in a local network using a previous version of the sw (previous release)
- in a local network using the new version of the sw
The publish should fail in both.

lollobene:
# Description of change

`IotaAttribute::View` was written into module bytecode by `fill_metadata` and accepted by the verifier **unconditionally**, so a `#[view]`-bearing package became publishable as soon as the new binary was live, that is, even before the `package_metadata_with_dynamic_module_metadata` protocol feature is enabled. 

During a mixed-binary rollout (binaries upgraded, protocol still on the previous version), a not-yet-upgraded validator cannot handle such a package, so upgraded and not-yet-upgraded validators disagree at publish verification, at upgrade-time `normalize_modules_with_metadata`, or during state-sync/checkpoint replay.

This PR gates the `View` attribute behind `package_metadata_with_dynamic_module_metadata` at both **emission** and **acceptance**, so every binary agrees to reject it until the protocol enables it.

## What changed

- **Versioned, old-binary-safe metadata.** Runtime module metadata is carried in `RuntimeModuleMetadataWrapper { version, inner: bytes }`, and `View` lives only in `IotaAttributeV2` / `RuntimeModuleMetadataV2` (`version: 2`). A binary that predates the feature reads the wrapper as `{version, opaque-bytes}` and rejects `version != 1` with a clean error instead of a hard `bcs` "unexpected variant" failure.
- **Acceptance gated (consensus-critical).** All metadata read paths go through `RuntimeModuleMetadataWrapper::try_into_runtime_module_metadata(config)`, which only admits `version == 2` when the feature is enabled.
- **Emission gated.** `fill_metadata` emits V1 metadata (no `View`) when the feature is off, and prints a build warning when a `#[view]` function is dropped for this reason.
- **`ProtocolBuildConfig`.** The low-level build/verify routines take a small `ProtocolBuildConfig { allow_view_function }` derived from `ProtocolConfig`.
- **CLI is protocol-aware.** `iota client publish`, `iota client verify-source`, and `iota client verify-bytecode-meter` derive the flag from the target network's protocol config. `iota move build` derives it from the resolved network (only when present) and exposes a `--allow-view-function <true|false>` override.

Fixes #12048

# Release Notes

- [ ] Protocol: 
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: The `iota move build` command is now aware of the flag related to enabling the view function in the protocol config of the target network. Same for `iota client publish`, `iota client verify-bytecode-meter` and `iota client verify-source`. The check at build time for view functions can be override by users  in `iota move build` using `--allow-view-function`
- [ ] Rust SDK:
- [ ] gRPC:


crates/iota-types/src/move_package.rs
```diff
@@ -909,25 +983,60 @@ pub struct AuthenticatorAttribute {
...
 
 /// V1 of IOTA specific metadata.
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct RuntimeModuleMetadataV1 {
     /// Attributes attached to functions, by definition index.
-    pub fun_attributes: BTreeMap<String, Vec<IotaAttribute>>,
+    pub fun_attributes: BTreeMap<String, Vec<IotaAttributeV1>>,
+}
+
+/// V1 of IOTA specific metadata.
```
bingyanglin: V2?

miker83z: Yes thx 🙏🏻  [2652498](https://github.com/iotaledger/iota/pull/12050/commits/26524988cdb707102c8e103114ded9387d333949)

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
```diff
@@ -31,7 +34,16 @@ use crate::{
...
-pub fn verify_module(module: &CompiledModule) -> Result<(), ExecutionError> {
+///
+/// `protocol_build_config.allow_view_function` reflects the
+/// `package_metadata_with_dynamic_module_metadata` protocol feature. While it
+/// is disabled, a package carrying the `View` attribute is rejected: the
+/// variant did not exist in older binaries, so accepting it would let a package
+/// onto the chain that a not-yet-upgraded validator cannot even deserialize.
+pub fn verify_module(
+    module: &CompiledModule,
+    protocol_build_config: ProtocolBuildConfig,
```
valeriyr: I would suggest considering passing ProtocolBuildConfig by immutable reference.
Since it is not supposed to be changed, but at the same time, the struct fields are public.

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
```diff
@@ -52,13 +64,13 @@ pub fn verify_module(module: &CompiledModule) -> Result<(), ExecutionError> {
                     "Failed to read bcs bytes for IOTA module metadata: {err}",
                 ))
             })?
-            .try_into()
+            .try_into_runtime_module_metadata(protocol_build_config)
```
Dkwcs, line 67: It would be nice to have a mechanism (probably a dedicated test) that helps us prevent silent breakages in the node. 
Currently, we just need to keep in mind that we should support that ProtocoBuildConfig structure up to date and pass the required data to the verifier.

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
miker83z: Fixed here [383687a](https://github.com/iotaledger/iota/pull/12050/commits/383687a5f36e61e9ff30ee3c1803607c59c9d0ca)

crates/iota-types/src/move_package.rs
```diff
@@ -810,6 +820,42 @@ fn build_upgraded_type_origin_table(
...
 }
 
+/// Protocol-dependent switches that the low-level package build and
+/// verification routines need.
+///
+/// Derived from the network's [`ProtocolConfig`], it lets those routines depend
+/// on a small, explicit set of protocol-gated flags rather than the full
+/// [`ProtocolConfig`].
+#[derive(Debug, Clone, Copy, Default)]
+pub struct ProtocolBuildConfig {
```
Thoralf-M, line 830: Just wondering why did you create this new config and not just pass the ProtocolConfig directly where we need it?

miker83z: It makes it easier to modify the parameters during the build (currently only allow_view_function but it will be populated later) and to create an instance when you don't know which network you are targeting the build for.

crates/iota-move/src/build.rs
```diff
@@ -15,6 +15,52 @@ use crate::manage_package::resolve_lock_file_path;
...
+        } = *defaults;
+        self.allow_view_function.get_or_insert(allow_view_function);
+    }
+}
+
+/// Resolves the CLI overrides into a [`ProtocolBuildConfig`], falling back to
+/// [`ProtocolBuildConfig::default()`] for any option left unset.
+impl From<ProtocolBuildConfigArgs> for ProtocolBuildConfig {
+    fn from(args: ProtocolBuildConfigArgs) -> Self {
+        let defaults = ProtocolBuildConfig::default();
```
Thoralf-M, line 52: The description says "iota move build derives it from the resolved network (or the max protocol version when offline)". The code does not do the offline-max part. A plain iota move build (no --client.env, not --dump-bytecode-as-base64) never contacts the network — iota_commands.rs:305 only fetches the protocol config when client_config.env.is_some(). Otherwise protocol_build_config_args stays default() (None), and build.rs:52 resolves ProtocolBuildConfigArgs::default().into() → ProtocolBuildConfig::default() → allow_view_function = false — not the max version.

Effect: building a #[view] package offline silently drops the attribute unless the user passes --allow-view-function true, even when the target/latest network would accept it.

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
```diff
@@ -67,23 +79,86 @@ pub fn verify_module(module: &CompiledModule) -> Result<(), ExecutionError> {
...
+                                _ => {
+                                    return Err(verification_failure(format!(
+                                        "Unsupported authenticator attribute version {} for function {}",
+                                        attr.version, fn_name
+                                    )));
+                                }
+                            }
+                        }
+                        IotaAttributeV2::View => {
+                            if !protocol_build_config.allow_view_function {
```
bingyanglin, line 156: This is a dead branch?

piotrm50 (approved):
approving `iota-core` and `iota-types` changes

crates/iota-move/src/build.rs
miker83z: We can invert this once mainnet is ready: https://github.com/iotaledger/iota/issues/12206

Do you suggest to keep `--allow-view-function` or rename it to `--disallow-view-function`?

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
miker83z: No, it would be alive between the release of the new sw version including this code and the epoch where the protocol enabling this flag will be starting.

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
miker83z: Cannot find a way to generalize the silent break unfortunately. I guess we need to be more careful 😉

crates/iota-move/src/build.rs
Thoralf-M: Guess it's fine to keep `--allow-view-function` now, when it's live on mainnet and default we never need to configure it anymore is there some value to disallow it later?

miker83z: It could be an option for devs to decide to not publish the package metadata and keep the source code intact (keeping the #[view] attribute), e.g., for localnet/devnet testing deployment

Thoralf-M: So in https://github.com/iotaledger/iota/issues/12206 it should then not just change the default but also switch to `--disallow-view-function` as otherwise there is no way to disable it?

miker83z: At that time we could make it either `--allow-view-function` default true or `--disallow-view-function` default false. But if you think "disallow" would be better, then I'd rather modify it in this PR (currently setting the default as true).

Thoralf-M: What do you mean default as true? Right now it's false

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
Dkwcs: Okay I see, but can we leave the appropriate comment somewhere near ProtocolConfig like - "do not forget about ProtocolBuildConfig mapping ...”?

crates/iota-move/src/build.rs
miker83z: I mean that if we take the direction of renaming `--disallow-view-function` then the default must be `true` (before https://github.com/iotaledger/iota/issues/12206)

miker83z: Ideally we would also need to add `--disallow-authenticator` with a different PR, in order to enable a similar behavior

Thoralf-M: I'm confused, you opened https://github.com/iotaledger/iota/issues/12206 to make the default `true`?

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
miker83z: I would avoid this. We can't know where the next feature that will touch ProtocolBuildConfig will be implemented, and leaving warning comments in ProtocolConfig would be redundant because we know that every feature (even independent from vm/execution/move) would pass from there (so, following this logic, anyone should leave warning comments in protocol config for any possibility).

crates/iota-move/src/build.rs
Thoralf-M: Anyways, not a blocker for this PR here

Thoralf-M (approved):

iota-execution/latest/iota-verifier/src/runtime_module_metadata.rs
Dkwcs: We can’t know for sure, but every time we add a new feature, we might affect fragile parts like ProtocolBuildConfig. There is no need to mention any exact feature names.
Anyway it’s up to you.

Dkwcs (approved):

sergiupopescu199 (approved):

valeriyr (approved):

PR #12054, opened 2026-06-26
fix(consensus): deflake test_fast_sync_voting_blocks_served_to_peer

### LINKED ISSUE #11877: flaky test: `test_fast_sync_voting_blocks_served_to_peer` fails with non-deterministic metric assertion
## Summary

The test `commit_syncer::fast::tests::test_fast_sync_voting_blocks_served_to_peer` in `crates/starfish/core/src/commit_syncer/fast.rs` is non-deterministic and failed during CI for PR #11790 (`starfish/feat/block-rate-burst`), requiring the "Hierarchy" workflow to be rerun **3 times total** before it passed.

## Observed failure

```
thread 'commit_syncer::fast::tests::test_fast_sync_voting_blocks_served_to_peer' panicked at crates/starfish/core/src/commit_syncer/fast.rs:1253:9:
Expected uncertified commits skipped > 0 for fast sync, got 0
```

CI run (attempt 1, job `Run rust tests`): https://github.com/iotaledger/iota/actions/runs/27143627964/job/80119818390

## Root cause

The assertion at [`fast.rs:1253`](https://github.com/iotaledger/iota/blob/develop/crates/starfish/core/src/commit_syncer/fast.rs#L1253) checks that the `commit_sync_fetch_commits_handler_uncertified_skipped{reason="fast_commit_sync"}` metric is `> 0` across all running authorities after the test completes:

```rust
assert!(
    commit_sync_fetch_commits_handler_uncertified_skipped > 0,
    "Expected uncertified commits skipped > 0 for fast sync, got {commit_sync_fetch_commits_handler_uncertified_skipped}"
);
```

This metric is incremented when a commit-sync request is served and the requested commits were obtained via fast sync (i.e., stored as uncertified/voting-block headers). The assertion is timing-sensitive:

- Validator A fast-syncs commits N2–N3 from the remaining 5 validators and stores them in voting storage.
- Validator B restarts and requests those same commits from A.
- If the peer network has progressed enough that commits N2–N3 are now fully certified by the time B asks A for them, A may serve them through the regular certified path instead of the voting-block path — meaning the `uncertified_skipped` metric stays at 0.

The test comment acknowledges that "peer selection is deterministic (not shuffled), so B will request from A first", but this does not prevent the race between commit certification and B's first sync request.

## Why this is not a fluke / infrastructure issue

- The test ran for ~34 seconds (TRY 1 FAIL) before asserting — plenty of time for commits to become certified.
- The assertion is purely about a metric value that depends on relative timing between commit certification and sync requests.
- Passed on a subsequent run without any code change.
- Unrelated to network errors, runner instability, or OOM.

## Suggested fix

Either:
1. **Make the assertion conditional / remove it** if the primary correctness signal is covered by the `total_hits > 0` assertion just above (line 1236–1239), which checks `commit_sync_voting_block_headers_hits`. The `uncertified_skipped` assertion is an additional observability check that can spuriously fail.
2. **Tighten the timing** by measuring the metric *before* the remaining validators continue committing in Phase 5 / before B restarts, so the metric is sampled while commits are still guaranteed to be uncertified on A.
3. **Use `#[sim_test]`** with the deterministic simulator so timing is under test control, preventing the race entirely.

## Affected file

`crates/starfish/core/src/commit_syncer/fast.rs` — assertion at line 1253, test starting at line 912.

VorobyevIlya:
# Description of change

`test_fast_sync_voting_blocks_served_to_peer` was non-deterministic and failed intermittently in CI. There were two independent causes:

- **Gap built from wall-clock time.** The work phases created the commit gap with fixed 10s sleeps. On a slow or loaded host they produce fewer commits, so the gap can fall below the fast-sync threshold; the restarted validator then catches up via regular sync, never stores voting block headers, and the metric assertions fail. Phases now drive a target commit count (twice the threshold) with a generous timeout, so the gap clears the threshold regardless of host speed.

- **Recovery window dwarfed the gap.** The leader-schedule rotation window — which also bounds the fast-sync reinitialization fetch — was a hardcoded constant (300), far larger than the synced gap. The recovering node therefore repopulated its regular block store across the whole gap, and the voting-block store under test was never consulted, leaving `commit_sync_voting_block_headers_hits` at the mercy of incidental alignment. It is now a `consensus_commits_per_schedule` protocol-config parameter (default unchanged — 300, 10 under msim — so no version bump and no snapshot changes), and the test shrinks it below the gap so the voting-serve path is actually exercised.

fixes #11877

### DIFF
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -1798,12 +1798,12 @@ impl ProtocolConfig {
         self.feature_flags.enable_pcool_flow
     }
 
-    pub fn commits_per_schedule(&self) -> u64 {
+    pub fn commits_per_schedule(&self) -> u32 {
         if cfg!(msim) {
             // Exercise faster leader-schedule rotation in simtests.
-            min(10, self.consensus_commits_per_schedule.unwrap_or(300)) as u64
+            min(10, self.consensus_commits_per_schedule.unwrap_or(300))
         } else {
-            self.consensus_commits_per_schedule.unwrap_or(300) as u64
+            self.consensus_commits_per_schedule.unwrap_or(300)
         }
     }
 }
--- a/crates/iota-protocol-config/src/lib.rs
+++ b/crates/iota-protocol-config/src/lib.rs
@@ -1403,6 +1403,10 @@ pub struct ProtocolConfig {
     // `fun native_sender_authenticator_function_info_v1<F>(): &Option<F>`
     // `fun native_sponsor_authenticator_function_info_v1<F>(): &Option<F>`
     auth_context_authenticator_function_info_v1_cost_base: Option<u64>,
+
+    /// Number of committed subdags between leader-schedule recomputations.
+    /// When unset, defaults to 300.
+    consensus_commits_per_schedule: Option<u32>,
 }
 
 // feature flags
@@ -1792,6 +1796,15 @@ impl ProtocolConfig {
 
     pub fn enable_pcool_flow(&self) -> bool {
         self.feature_flags.enable_pcool_flow
+    }
+
+    pub fn commits_per_schedule(&self) -> u64 {
+        if cfg!(msim) {
+            // Exercise faster leader-schedule rotation in simtests.
+            min(10, self.consensus_commits_per_schedule.unwrap_or(300)) as u64
+        } else {
+            self.consensus_commits_per_schedule.unwrap_or(300) as u64
+        }
     }
 }
 
@@ -2393,6 +2406,7 @@ impl ProtocolConfig {
             auth_context_replace_cost_base: None,
             auth_context_replace_cost_per_byte: None,
             auth_context_authenticator_function_info_v1_cost_base: None,
+            consensus_commits_per_schedule: None,
             // When adding a new constant, set it to None in the earliest version, like this:
             // new_constant: None,
         };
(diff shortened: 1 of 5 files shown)


crates/iota-protocol-config/src/lib.rs
```diff
@@ -1797,6 +1797,15 @@ impl ProtocolConfig {
     pub fn enable_pcool_flow(&self) -> bool {
         self.feature_flags.enable_pcool_flow
     }
+
+    pub fn commits_per_schedule(&self) -> u64 {
```
piotrm50: Is this going to be reused with the new leader schedule for Starfish-S or is this deprecated then? It it's the latter then I think we should avoid adding a protocol param and try to add this conditional default value in the logic.

bingyanglin (approved):

jkrvivian (approved):
Approve files owned by the node team.

piotrm50 (approved):

crates/iota-protocol-config/src/lib.rs
VorobyevIlya: It will be reused in Starfish-S leader schedule

Thoralf-M: Why u64 instead of u32? Since you're casting inside to u64 and where it's called you cast back to u32
`let leader_schedule_window = inner.context.protocol_config.commits_per_schedule() as u32;`

VorobyevIlya: you are right, u32 is better. Changed it

Thoralf-M (approved):
Approving the iota-protocol-config change

kodemartin (approved):

PR #12057, opened 2026-06-26
upstream(starfish-core): reduce consensus log verbosity and avoid hot-path block clone

piotrm50:
# Description of change

This bundles three small, low-risk upstream consensus polish changes.

- Port of upstream PRs:
  - [MystenLabs/sui#24746](https://github.com/MystenLabs/sui/pull/24746) — avoid cloning the subscribed block before handling
  - [MystenLabs/sui#22487](https://github.com/MystenLabs/sui/pull/22487) — reduce logging level on hot paths
  - [MystenLabs/sui#22837](https://github.com/MystenLabs/sui/pull/22837) — minor logging updates (compact block-ref display)
- Description:

Three independent changes to the starfish consensus core. First, the subscribed block bundle is owned and unused after being handed to `handle_subscribed_block_bundle`, so it is moved directly instead of cloned, removing one allocation per received block. Second, the chattiest steady-state log lines (block acceptance, leader-decision loop, stake-quorum checks) are downgraded from DEBUG to TRACE, committer context is added to the remaining direct/indirect-rule DEBUG lines, and a couple of list-printing messages are guarded behind `is_empty()` checks. Third, raw `Debug` formatting of ancestor/acknowledgment slices and digest loops is replaced with the compact `BlockRef` `Display` string.

Adapted to the fork:
- starfish's `block_manager.rs` is a directory, so the log changes were applied in `block_manager/mod.rs`.
- the `block.rs` logging hunks from upstream were mapped onto starfish's `block_header.rs`.
- the starfish block-acceptance path is structured differently from upstream, so the per-call-site log levels diverge slightly (see PR discussion).

### DIFF
--- a/crates/starfish/core/src/block_header.rs
+++ b/crates/starfish/core/src/block_header.rs
@@ -665,6 +665,12 @@ impl fmt::Debug for BlockRef {
     }
 }
 
+/// Formats a slice of block references as a comma-separated list of their
+/// short `Display` form, for debug/log output.
+pub(crate) fn format_block_digests(blocks: &[BlockRef]) -> String {
+    blocks.iter().map(|b| b.to_string()).join(", ")
+}
+
 impl Hash for BlockRef {
     fn hash<H: Hasher>(&self, state: &mut H) {
         state.write(&self.digest.0[..8]);
@@ -1293,14 +1299,11 @@ impl fmt::Debug for VerifiedBlockHeader {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         write!(
             f,
-            "{:?}([{}];{}ms;[{}]a;{}c)",
+            "{:?}({}ms;[{}]r;[{}]a;{}c)",
             self.reference(),
-            self.ancestors().iter().map(|a| a.to_string()).join(", "),
             self.timestamp_ms(),
-            self.acknowledgments()
-                .iter()
-                .map(|a| a.to_string())
-                .join(", "),
+            format_block_digests(self.ancestors()),
+            format_block_digests(self.acknowledgments()),
             self.commit_votes().len(),
         )
     }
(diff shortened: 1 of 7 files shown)


crates/starfish/core/src/block_header.rs
```diff
@@ -1292,11 +1293,14 @@ impl fmt::Debug for VerifiedBlockHeader {
...
         write!(
             f,
-            "{:?}({}ms;{:?}r;{:?}a;{}c)",
+            "{:?}([{}];{}ms;[{}]a;{}c)",
             self.reference(),
+            self.ancestors().iter().map(|a| a.to_string()).join(", "),
             self.timestamp_ms(),
-            self.ancestors(),
-            self.acknowledgments(),
+            self.acknowledgments()
```
polinikita: Should acknowledgments be separated this way from ancestors?

crates/starfish/core/src/commit.rs
```diff
@@ -941,16 +942,7 @@ fn format_transaction_ref_digests(transaction_refs: &[GenericTransactionRef]) ->
...
-    for (idx, block) in blocks.iter().enumerate() {
-        if idx > 0 {
-            result.push_str(", ");
-        }
-        result.push_str(&block.digest.to_string());
-        result.push('@');
-        result.push_str(&block.round.to_string());
-    }
-    result
+    blocks.iter().map(|b| b.to_string()).join(", ")
```
polinikita: Does this make it more readable?

piotrm50: its similar, but less code here. also makes block ref format uniform if we use the display trait here instead of having custom one.
```
"B{}({},{})", self.round, self.author, self.digest
```

crates/starfish/core/src/block_header.rs
piotrm50: cleaned it up a bit.

VorobyevIlya (approved):

polinikita (approved):

Issue #12052, opened 2026-06-26
fix(starfish): correct dashboard latency percentiles and per-validator readability


polinikita:
An audit of the Starfish Overview Grafana dashboard (`dev-tools/grafana-local/dashboards/starfish-overview.json`) found a set of metric-correctness and readability issues. Realized impact is misleading latency figures on the Highlights row and slower per-validator debugging; upstream metrics are unaffected.

**Key points:**

- The Highlights `Block Latency` and `Tx Latency` panels label a cross-validator quantile of each validator's *average* commit latency as p5/p50/p95. The backing metrics are histograms, so these are not latency percentiles; a `< 2s` filter additionally hides high-latency samples as gaps.
- Most per-validator timeseries use single-series hover tooltips, so on hover it is not possible to tell which validator a line corresponds to.
- The `Reconstructed, but unknown to DAG state` panel references its metric without the `consensus_` namespace prefix and is permanently empty.
- Two panels render blank legend entries (legend label references absent from the series).
- Two latency panels render seconds as unitless numbers; a seconds table uses percentage-mode thresholds; a few titles carry stray whitespace and one title is duplicated; the `Starfish RAM` row shows count gauges rather than memory.

The audit also surfaced coverage gaps for Byzantine-fault and peer-responsiveness metrics (e.g. equivocations by authority, faulty blocks, peer responsiveness effective latency) that have no panel today; these are tracked for a follow-up.

PR #12053, opened 2026-06-26
fix(starfish): correct dashboard latency percentiles and per-validator readability

polinikita:
# Description of change

Corrects metric-accuracy and per-validator readability issues in the Starfish Overview Grafana dashboard, found during an audit, plus a related obsolete-metric cleanup.

**Dashboard (`starfish-overview.json`):**

- **Latency percentiles** — `Block Latency` and `Tx Latency` (Highlights) computed a cross-validator quantile of each validator's *average* commit latency and labeled it p5/p50/p95. Since the backing metrics are histograms, these are now pooled latency percentiles via `histogram_quantile(φ, sum by (le) (rate(_bucket[..])))`, and the `< 2s` filter that silently hid latency spikes is removed. Also dropped the `or … block_header_commit_latency` fallback from the Tx Latency panels (it mixed block-header latency into a transactions panel).
- **Commit Syncer — regular vs fast** — each `commit_sync_*` metric carries a `source` label (`commit_sync` = regular syncer, `fast_commit_sync` = fast syncer). The Commit Syncer panels are now split into left (regular) / right (fast) pairs — Inflight fetches, Fetched commits rate, Gap Retries, Fetch Once Latency, Pending fetches, Fetch Errors — so the two syncers can be compared side by side. "Commit Sync Difference" stays full-width (its quorum/local index gauges carry no `source` label).
- **Tooltips & legends** — tooltip mode is chosen by series cardinality: per-validator panels (distinguished by host/authority/peer/leader/voter, ~committee size) use single-series tooltips so hover shows only the plot under the cursor; low-cardinality panels (distinguished by a small label set such as source/reason/type/route or a few percentile lines) use multi-series tooltips sorted descending. Legends are enabled on every panel.
- **Empty panel** — `Reconstructed, but unknown to DAG state` referenced its metric without the `consensus_` namespace prefix and never rendered; fixed.
- **Hygiene** — two blank legends fixed; seconds units added to two latency panels; a seconds table switched from percentage to absolute thresholds; a duplicated panel title disambiguated into (live)/(periodic); title whitespace trimmed; row `Starfish RAM` renamed to `Starfish In-Memory State`.

**Obsolete metric cleanup (private-network tooling):** `consensus_block_commit_latency` was renamed to `consensus_block_header_commit_latency` in #9120 and no longer exists. Removed the dead legacy-name references from `run-migration-test.py`, `experiment_common.py`, and the fuzzer's `metrics.py`.

[...]

PR #12070, opened 2026-06-29
chore(starfish): drop obsolete consensus_block_commit_latency references

polinikita:
# Description of change

The consensus block-commit latency histogram was renamed `consensus_block_commit_latency` → `consensus_block_header_commit_latency` in #9120 and the old name no longer exists. This removes the now-dead legacy-name references from the private-network tooling, keeping only the current metric:

- `experiments/run-migration-test.py` — drop the `block_commit_latency or …` fallback from the Block commit p50/p95 queries.
- `experiments/experiment_common.py` — same for `blk_p50`/`blk_p95`; update the docstring (transaction queries still intentionally carry no block fallback).
- `fuzzer/src/net_fuzz/metrics.py` — read `consensus_block_header_commit_latency_sum/count` directly instead of trying the legacy name first.

No behavior change against current nodes (queries already fell back to the live name); this just removes dead references.

Relates to #12052


polinikita:
Folding these changes into #12053 per review (same branch).

PR #12071, opened 2026-06-29
upstream(starfish-core): add block receive delay metric and tighten quorum receive latency

piotrm50:
# Description of change

- Port of upstream PR: [MystenLabs/sui#22457](https://github.com/MystenLabs/sui/pull/22457)
- Description:

Two consensus observability changes, plus dashboard panels for both:

- **New `block_receive_delay` metric** (per authority): for an equal-round block that arrived before the round's 2f+1 quorum formed, record how long after the round started that peer's block arrived. A high value flags a validator whose blocks consistently arrive late within the round.
- **Tightened `quorum_receive_latency`**: measure it from the local node's *own* proposal timestamp for the round to when that round reaches quorum, and only sample rounds the local node actually proposed in.

### Why the previous `quorum_receive_latency` was worse

Previously the metric was recorded in `ThresholdClock::try_advance_round` as the monotonic time between *consecutive round advances*, sampled on **every** advance regardless of whether the local node participated. That skews the metric in exactly the cases you most want it to be trustworthy:

- **Idle / low-activity periods.** The sample for the first advance after a quiet stretch includes the entire idle gap. If the network is quiet for 30s and then a round gathers quorum in ~200ms, the old metric records ~30s — inflating avg/p95 and making the dashboard show false high latency.
- **Rounds the local node didn't drive.** Peer-driven advances (the local node didn't propose, is behind, or is catching up after a restart) still produced samples, so the metric mixed "how fast did a quorum form once I proposed" with "how long since this process last advanced any round."

The new version anchors to the local node's own block timestamp and is gated on the node having proposed in that round, so it reflects genuine quorum-formation latency for rounds the node took part in, and stays quiet during idle/peer-only rounds.

Adapted to the fork:
- Kept develop's overflow-safe `round.saturating_add(1)` and its `quorum_ts` field (reused as the round-start reference for `block_receive_delay`).
- Reconciled with develop's `dag-visualizer` instrumentation: the quorum observation is inlined on `add_block_header`'s new `bool` return, while the visualizer keeps its own `get_round()` before/after check (it must still fire on round *jumps*, which the quorum-advance bool does not cover).

`threshold_clock` unit tests pass (round advancement unchanged by the `bool` return). Compiles clean with and without the `dag-visualizer` feature; fmt + clippy clean.


polinikita (approved):

crates/starfish/core/src/dag_state.rs
```diff
@@ -855,7 +855,23 @@ impl DagState {
         );
         #[cfg(feature = "dag-visualizer")]
         let clock_before = self.threshold_clock.get_round();
-        self.threshold_clock.add_block_header(block_ref);
+        if self.threshold_clock.add_block_header(block_ref) {
```
bingyanglin, line 871: These two latency metrics also fire when we're replaying stored headers on recovery and fast-sync, so after a restart the quorum one records `now - old_proposal_ts`. Maybe skip the observe when `source` is `Recover`/`FastCommitSyncer`?

crates/starfish/core/src/threshold_clock.rs
```diff
@@ -58,23 +51,45 @@ impl ThresholdClock {
...
             Ordering::Equal => {
                 self.aggregator
                     .add(block_header.author, &self.context.committee);
             }
         }
-        self.try_advance_round(block_header.round.saturating_add(1));
+        let advanced = self.try_advance_round(block_header.round.saturating_add(1));
+        // For an equal-round block that did not complete the round's quorum,
+        // record how long after the round started this peer's block arrived.
+        if is_equal_round && !advanced {
```
bingyanglin, line 78: A peer sending two blocks for the same round (different digests) bumps this twice while the quorum counts it once. Maybe skip if we've already recorded this author this round?

piotrm50: if a validator equivocates then this metric is the least concern, especially if both arrive before round advances.

crates/starfish/core/src/dag_state.rs
piotrm50: now it's gated

bingyanglin (approved):

PR #12073, opened 2026-06-29
upstream(iota-core): remove unnecessary async from AuthorityState methods

piotrm50:
# Description of change

- Port of upstream PR: [MystenLabs/sui#26889](https://github.com/MystenLabs/sui/pull/26889)
- Description:

Several `AuthorityState` object-access methods were declared `async` but never awaited anything — they just call synchronously into the object store/cache. This makes them synchronous and drops the now-unnecessary `.await` at every call site. The de-async'd methods: `try_get_object`, `get_object`, `get_iota_system_package_object_ref`, `get_move_objects`, `insert_genesis_object`, `insert_genesis_objects`, `get_transaction_lock`, `try_get_objects`, `get_objects`, `try_get_object_or_tombstone`, `get_object_or_tombstone`. `insert_genesis_objects` also drops a pointless `join_all` over now-synchronous calls in favour of a plain loop.

No behavioural change — the methods did no real async work; this only removes the async/await overhead and noise.

Adapted to the fork:
- Applied to develop's current code rather than cherry-picked: develop had diverged (e.g. `get_move_objects` takes `Address`, not `IotaAddress`) and added call sites the upstream diff didn't touch. The de-async was re-derived against develop and every call site was found with the compiler, so this covers a few more files than the upstream PR (e.g. `post_consensus_validation_tests`, `server_tests`).

### Follow-up commit (starfish-core, not part of upstream #26889)

A second commit de-async'es `ShardReconstructor::get_transactions_with_headers_in_dag_state` (starfish-core), which never awaited anything in either cfg branch (found via `clippy::unused_async`). Its single call site in `send_to_core` drops the `.await`; `send_to_core` itself stays async (it still awaits `add_transactions`). This is fork-only code with no upstream counterpart.

(Note: `AuthorityState::get_system_package_bytes` also becomes awaitless after this change, but is intentionally left `async` to stay aligned with upstream, which keeps it async too.)

`iota-core`, `iota-json-rpc`, `test-cluster`, `transaction-fuzzer`, and `iota-e2e-tests` all build (lib + tests); `iota-core` also builds under the simulator; clippy clean on `iota-core`, `starfish-core`, and the dependent crates; fmt clean. The change is mechanically behaviour-preserving (sync vs async on methods that never awaited).

### DIFF
--- a/crates/test-cluster/src/lib.rs
+++ b/crates/test-cluster/src/lib.rs
@@ -267,8 +267,7 @@ impl TestCluster {
     pub async fn get_object_from_fullnode_store(&self, object_id: &ObjectId) -> Option<Object> {
         self.fullnode_handle
             .iota_node
-            .with_async(|node| async { node.state().get_object(object_id) })
-            .await
+            .with(|node| node.state().get_object(object_id))
     }
 
     pub async fn get_latest_object_ref(&self, object_id: &ObjectId) -> ObjectRef {
--- a/crates/test-cluster/src/lib.rs
+++ b/crates/test-cluster/src/lib.rs
@@ -267,7 +267,7 @@ impl TestCluster {
     pub async fn get_object_from_fullnode_store(&self, object_id: &ObjectId) -> Option<Object> {
         self.fullnode_handle
             .iota_node
-            .with_async(|node| async { node.state().get_object(object_id).await })
+            .with_async(|node| async { node.state().get_object(object_id) })
             .await
     }
 
--- a/crates/iota-core/src/authority.rs
+++ b/crates/iota-core/src/authority.rs
@@ -3039,8 +3039,7 @@ impl AuthorityState {
 
         let requested_object_seq = match request.request_kind {
             ObjectInfoRequestKind::LatestObjectInfo => {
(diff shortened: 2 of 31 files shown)


crates/test-cluster/src/lib.rs
```diff
@@ -267,7 +267,7 @@ impl TestCluster {
     pub async fn get_object_from_fullnode_store(&self, object_id: &ObjectId) -> Option<Object> {
         self.fullnode_handle
             .iota_node
-            .with_async(|node| async { node.state().get_object(object_id).await })
+            .with_async(|node| async { node.state().get_object(object_id) })
```
tomxey: I guess this could become just `with` now, given that function inside doesn't have to be async.

tomxey (approved):

muXxer (approved):

crates/test-cluster/src/lib.rs
piotrm50: done here and in a few other places

VorobyevIlya (approved):

miker83z (approved):

roman1e2f5p8s (approved):

PR #12075, opened 2026-06-29
feat(json-rpc): shutdown JSON-RPC gracefully

### LINKED ISSUE #12072: [indexer]: json-rpc service does not shutdown gracefully
## Description

The JSON-RPC service launched with the `iota-indexer` binary does not shut down upon a termination signal. 

Termination signals in the IOTA Indexer trigger cancellation through a [CancellationToken](tokio_util::sync::CancellationToken), but the JSON-RPC service uses it only in case of a server internal error to communicate with other tasks:

https://github.com/iotaledger/iota/blob/1d2757a291f127df0ead0fdeacc3528a27322227/crates/iota-json-rpc/src/lib.rs#L235-L246

So the JSON-RPC currently stays up even if the case a termination signal is emitted on the application.

kodemartin:
# Description of change

This patch adds allows shutting down the JSON-RPC server when a cancellation token is cancelled.

Fixes #12072

Manual tests, starting the Indexer JSON-RPC service, and interrupting with Ctrl-C.

### DIFF
--- a/crates/iota-json-rpc/src/lib.rs
+++ b/crates/iota-json-rpc/src/lib.rs
@@ -232,23 +232,31 @@ impl JsonRpcServerBuilder {
             Error::Unexpected(format!("invalid listen address {listen_address}: {e}"))
         })?;
 
-        let fut = async move {
-            axum::serve(
-                listener,
-                app.into_make_service_with_connect_info::<SocketAddr>(),
-            )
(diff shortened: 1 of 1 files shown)


tomxey (approved):

sergiupopescu199 (approved):
lgtm!

PR #12076, opened 2026-06-29
refactor(consensus): rename metric to last_finalized_leader_round

VorobyevIlya:
# Description of change

Renames the consensus leader-round gauge from `last_decided_leader_round` to
`last_finalized_leader_round`. The gauge is set from the last finalized leader,
so the old name disagreed with the value it reports — "decided", "committed",
and "finalized" are distinct leader states, and the other leader-round metrics
already match the state they track.

The emitted Prometheus metric changes from `consensus_last_decided_leader_round`
to `consensus_last_finalized_leader_round`; the corresponding Grafana dashboard
panel needs the same rename.

### DIFF
--- a/crates/starfish/core/src/core.rs
+++ b/crates/starfish/core/src/core.rs
@@ -1282,7 +1282,7 @@ impl Core {
             self.context
                 .metrics
                 .node_metrics
-                .last_decided_leader_round
+                .last_finalized_leader_round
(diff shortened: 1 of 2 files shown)


polinikita (approved):

piotrm50 (approved):
we should have a followup PR to update the Grafana dashboard if it uses this metric.

PR #12078, opened 2026-06-30
feat(indexer): restore IOTA Indexer from formal snapshot

### LINKED ISSUE #12058: [indexer]: implement restore logic based on formal snapshot v2

kodemartin:
# Description of change

This patch introduces the logic and CLI to allow users bootstrap the database from a formal snapshot of the target network. 

It builds on top of the `iota_snapshot::reader::StateSnapshotReaderV1`, to verify the snapshot based on a user-provided `genesis.blob`, insert partitions of live objects, and populate all necessary tables in the IOTA Indexer database to kickoff synchronization from the formal-snapshot epoch and onward.

The implementation initializes the following tables: 

- `objects`
- `epochs`
- `chain_identifier`
- `protocol_configs`
- `feature_flags`
- `checkpoints` with the last checkpoint of the formal-snapshot epoch. This is used as watermark for the synchronization.
- `watermarks` with the `first_{tx,chk}_sequence_number` of the epoch following the snapshot as lower bounds, and by setting the `lowest_unpruned_key` to these lower bounds according to each table's pruning strategy. 

> [!NOTE]
> The `displays` table should be also populated, but the associated logic will be patched with a follow-up PR, as it required schema changes.

The new feature is exposed through a new CLI command 

```sh
$ ./iota-indexer help restore
Bootstrap the Indexer database from a formal snapshot

Usage: iota-indexer restore --network <NETWORK> <COMMAND>

Commands:
  run               Start restoring from a formal snapshot of the target network
  available-epochs  Print the epochs for which there is an available formal snapshot
  help              Print this message or the help of the given subcommand(s)

Options:
      --network <NETWORK>  Target network [possible values: mainnet, testnet, devnet]
  -h, --help               Print help

```

To start restoring the `run` subcommand is used that requires a local copy of the `genesis.blob` at `--genesis-path` and a local directory to download the reference files used for checksum verification of the live-object partitions.

```sh
$ ./iota-indexer restore help run
Start restoring from a formal snapshot of the target network

Usage: iota-indexer restore --network <NETWORK> run [OPTIONS] --staging-path <STAGING_PATH> --genesis-path <GENESIS_PATH>

Options:
      --staging-path <STAGING_PATH>
          Local directory used to stage the downloaded MANIFEST and `.ref` files

      --genesis-path <GENESIS_PATH>
          Path to the genesis blob, required to verify the snapshot against the committee chain

      --epoch <EPOCH>
          Epoch to download. Defaults to the latest available epoch

      --num-parallel-downloads <NUM_PARALLEL_DOWNLOADS>
          Number of parallel downloads. Defaults to the available parallelism.
          
          Must be strictly positive.

  -h, --help
          Print help (see a summary with '-h')

```

### Other minor changes

In the process of the feature implementation the following changes were deemed helpful to make the API simpler and more intuitive: 

- Simplification of the `iota_indexer::common::prepare::LiveObject` struct, to derive the transaction digest from the inner object as opposed to storing it explicitly.
- Add a utility function to extract epoch information from the respective end-of-epoch events.
- `PgIndexerStore` methods were refactored in order to reuse code in the context of snapshot restore.

Closes #12058

`./iota-indexer restore --network mainnet available-epochs`

The `run` subcommand is only tested for the previous version of snapshots. Dedicated testing for the V2 snapshot version is planned through #12061 which depends on #11254

QA will be part of #12061

### Release Notes

- [x] Indexer: Introduces a new `restore` CLI command that allows restoring the Indexer database from formal snapshots.


crates/iota-indexer/src/restore/persist.rs
```diff
@@ -0,0 +1,185 @@
...
+                let checkpoint_sequence_number =
+                    previous_transaction_checkpoint.unwrap_or_default();
+                Some(LiveObject::new(checkpoint_sequence_number, object))
+            },
+        );
+        let chunks = chunk!(partition, self.config.parallel_objects_chunk_size);
+        let sha3_digest = hasher.finalize().digest;
+        if *expected_checksum != sha3_digest {
+            tracing::error!(
+                "Sha does not match! expected: {expected_checksum:?}, actual: {sha3_digest:?}",
```
tomxey: nit: error message should start from lowercase I believe, also "sha" would look nicer

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -1058,36 +1062,42 @@ impl PgIndexerStore {
...
+    }
+
+    fn persist_epoch(&self, epoch: EpochToCommit) -> Result<(), IndexerError> {
+        let guard = self
+            .metrics
+            .checkpoint_db_commit_latency_epoch
+            .start_timer();
+        let epoch_id = epoch.new_epoch.epoch;
+
+        self.persist_epochs(vec![epoch])
```
tomxey, line 1093: have you considered having the actual impl in `persist_epoch` and then for `persist_epochs` do: `for epoch in epochs: persist_epoch(epoch)` ?

that would seem more intuitive for me, but maybe it's not possible for some reason?

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -269,7 +269,10 @@ impl PgIndexerStore {
         Ok(())
     }
 
-    fn persist_changed_objects(&self, objects: Vec<LiveObject>) -> Result<(), IndexerError> {
+    pub(crate) fn persist_changed_objects(
```
tomxey: I think `persist_live_objects` would be more accurate now, and would nicely accompany the `persist_removed_objects` method.

In the context of snapshot restore "changed objects" looks strange, as nothing really got changed, we are bootstrapping the initial live versions.

crates/iota-indexer/src/restore/orchestration.rs
```diff
@@ -0,0 +1,76 @@
...
+    let epoch_info = reader
+        .read_epoch_info()
+        .await
+        .map_err(|e| IndexerError::Restore(format!("failed to read epoch info: {e}")))?;
+    let snapshot_chain_id = reader.chain_id();
+    let genesis = Genesis::load(genesis_path)?;
+
+    // It's ok to ignore the handle. Cancellation is effected by dropping the
+    // `read_to_db` future below, so we don't need to call `abort` explicitly.
+    let (_abort_handle, abort_registration) = AbortHandle::new_pair();
```
tomxey, line 56: do we have graceful shutdown in such case?

crates/iota-indexer/src/restore/orchestration.rs
```diff
@@ -0,0 +1,76 @@
...
+        mpsc::channel::<(GlobalStateHash, u64)>(num_parallel_downloads.get());
+
+    let verified_epoch_info = verify_epoch_info(epoch_info, genesis, snapshot_chain_id).await?;
+    let ((), num_objects) = tokio::try_join!(
+        reader
+            .read_to_db(&pg_indexer_store, abort_registration, Some(state_hash_tx))
+            .map_err(IndexerError::from),
+        verify_state_hash(state_hash_rx, &verified_epoch_info),
+    )?;
+    // TODO: persist epochs, checkpoint, chain_identifier. This can be done in
```
tomxey: is this TODO still valid? seems it is handled by the function below

crates/iota-indexer/src/restore/persist.rs
```diff
@@ -0,0 +1,185 @@
...
+        .await
+}
+
+/// We populate the remaining tables after the objects.
+///
+/// This includes `epochs`, `chain_identifier` which we populate in parallel
+/// with setting the checkpoint watermark to the last checkpoint of the snapshot
+/// epoch.
+///
+/// Finally we populate `protocol_configs` and `feature_flags` up to the
```
tomxey: we should also populate the `watermarks` table and set the lower bound of available data to current epoch/checkpoint, so that read queries know that this indexer is pruned and cannot serve data from genesis

crates/iota-indexer/src/restore/persist.rs
```diff
@@ -0,0 +1,185 @@
...
+             }| {
+                hasher.update(object.object_ref().digest.inner());
+                let checkpoint_sequence_number =
+                    previous_transaction_checkpoint.unwrap_or_default();
+                Some(LiveObject::new(checkpoint_sequence_number, object))
+            },
+        );
+        let chunks = chunk!(partition, self.config.parallel_objects_chunk_size);
+        let sha3_digest = hasher.finalize().digest;
+        if *expected_checksum != sha3_digest {
```
tomxey, line 49: this is for my understanding, is this verification needed here if we already have `verify_state_hash` ? or does it have different scope?

crates/iota-indexer/src/restore/setup.rs
```diff
@@ -0,0 +1,191 @@
...
+/// # Errors
+///
+/// Returns an error if the read client cannot be built, or the root MANIFEST
+/// cannot be fetched or parsed.
+pub async fn available_epochs(network: Network) -> IndexerResult<Vec<u64>> {
+    let remote_store = FormalSnapshotStore::new(network)?;
+    let mut epochs = remote_store.available_epochs().await?;
+    epochs.sort_unstable_by_key(|&epoch| Reverse(epoch));
+    Ok(epochs)
+}
```
tomxey: could we do the sort inside of `remote_store.available_epochs()` and then replace all the callers of this function with `FormalSnapshotStore::new(network).available_epochs()`?

This looks a bit redundant to have 2 variants, one external that sorts, and one internal that doesn't.

crates/iota-indexer/src/restore/setup.rs
```diff
@@ -0,0 +1,191 @@
...
+            .iter()
+            .map(|(epoch, _)| *epoch)
+            .collect())
+    }
+
+    /// Verifies that the formal snapshot upload for the given epoch has
+    /// completed.
+    async fn verify_completed_snapshot(&self, epoch: u64) -> IndexerResult<()> {
+        let success_marker = format!("epoch_{epoch}/{SUCCESS_MARKER}");
+        // TODO: sort out failure modes of exists
```
tomxey: will this TODO be cleaned up before merge to develop? should we have some issue for it?

crates/iota-indexer/src/restore/setup.rs
```diff
@@ -0,0 +1,191 @@
...
+    config: ObjectStoreConfig,
+    store: Arc<dyn ObjectStoreGetExt>,
+}
+
+impl FormalSnapshotStore {
+    /// Builds the read client for the network's public formal snapshot store.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the network is not `mainnet` or `testnet`, or if the
```
tomxey: Do we have to make devnet even an accepted value in such case?
Also, I don't see any explicit fail for devnet in the code, would the failure come indirectly from the fact that we don't have devnet snapshot available maybe? If that's the case, it could be clarified in the doc.

crates/iota-indexer/src/restore/persist.rs
kodemartin: these are error logs, not rust errors where the convention applies. I see we use primarily lower case for error logs, but not exclusively.

Will convert to lower case nevertheless.

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: this ensures that all epochs are persisted in one transaction. To have the same effect I would otherwise have to impl just the query in `persist_epoch` and wrap this in the transactional macro.

crates/iota-indexer/src/restore/orchestration.rs
kodemartin: no, this is managed upstream in `iota-snapshot` and abort is on the spawned task level. 

Generally we can accept a corrupted state upon unsuccessful completion of the restore. Re-running resets the database.

crates/iota-indexer/src/restore/persist.rs
kodemartin: yes, its a different scope: this verifies the integrity of each partition based on the checksum found on the `ref` files.

crates/iota-indexer/src/restore/setup.rs
kodemartin: Having two variants, limits the public surface of the API. We only expose what we need, keeping the struct details and API private.

But since the indexer is not public library, I can certainly simplify :+1:

crates/iota-indexer/src/restore/setup.rs
kodemartin: forgot about it. will try to sort it out in this patch.

crates/iota-indexer/src/restore/setup.rs
kodemartin: That was a stale text from a previous iteration that I was using `Chain` as input. Will update.

crates/iota-indexer/src/restore/orchestration.rs
kodemartin: No, it was stale, will remove, thanks!

crates/iota-indexer/src/restore/persist.rs
kodemartin: 33a61bd079

crates/iota-indexer/src/restore/setup.rs
kodemartin: d6bb81bdf7

crates/iota-indexer/src/restore/setup.rs
kodemartin: 3f6c489449

crates/iota-indexer/src/restore/orchestration.rs
kodemartin: 60e24123db

crates/iota-indexer/src/restore/setup.rs
kodemartin: 1f0b4c8292

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: 7e18717447

crates/iota-indexer/src/restore/persist.rs
kodemartin: 5cfc9c85c3

crates/iota-indexer/src/restore/setup.rs
kodemartin: also opened an issue to resolve once and for all upstream: https://github.com/iotaledger/iota/issues/12108

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: Is it actually important to have it in single transaction? what are the benefits?

crates/iota-indexer/src/restore/orchestration.rs
tomxey: would it be worth mentioning in the comment here?

crates/iota-indexer/src/restore/persist.rs
tomxey: I think this will not do, it updates only the upper bound, so the lower bounds will be inserted as default in such case I believe (default = from genesis).

I think we would need some custom code for this to set lower bounds accordingly.

crates/iota-indexer/src/restore/persist.rs
tomxey: but is it possible that `verify_state_hash` passes, and integrity check of a partition fails?

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: Had to be clearer: using a single connection is more important here. The transactional macro implicitly allows that.

crates/iota-indexer/src/restore/persist.rs
kodemartin: is there a similar existing method that you could recommend using, since you are more familiar with this API?

crates/iota-indexer/src/restore/persist.rs
kodemartin: Probably not. I think it's about faster fail. This replicates the logic in `iota-tool`.

crates/iota-indexer/src/restore/orchestration.rs
kodemartin: mentioning what exactly?

crates/iota-indexer/src/restore/persist.rs
kodemartin: I see we have `update_watermark_lower_bound` similarly.

crates/iota-indexer/src/restore/setup.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Builds the config for an unsigned (anonymous) HTTPS object store served at
+/// the given public endpoint.
+fn unsigned_http_store_config(endpoint: &str) -> ObjectStoreConfig {
+    ObjectStoreConfig {
+        object_store: Some(ObjectStoreType::S3),
+        aws_endpoint: Some(endpoint.to_string()),
+        aws_virtual_hosted_style_request: true,
+        object_store_connection_limit: 200,
+        no_sign_request: true,
```
sergiupopescu199: nit: These values seem not to be used by the [make_http](https://iotaledger.github.io/iota/src/iota_storage/object_store/http/mod.rs.html#48-65) method.

crates/iota-indexer/src/restore/orchestration.rs
tomxey: That we can accept a corrupted state, and therefore we don't need graceful shutdown.

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: Why is a single connection needed? Just to understand.

crates/iota-indexer/src/restore/persist.rs
kodemartin: will this do? 7ae657acf4

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: To persist the epochs in one round trip.

crates/iota-indexer/src/restore/orchestration.rs
kodemartin: but the context is different here. We comment on why we ignore the abort handle, which clarifies that we have no control on the abort.

Graceful shutdown could be discussed on a higher level, that is on the restore operation as a whole. I can add this in the `start` method doc instead.

kodemartin: https://github.com/iotaledger/iota/pull/12078/commits/49022e37f2df2c85bfbf732753803c4c50f03276

crates/iota-indexer/src/main.rs
```diff
@@ -135,6 +153,41 @@ async fn main() -> Result<(), IndexerError> {
...
+            let num_parallel_downloads = num_parallel_downloads.unwrap_or_else(|| {
+                std::thread::available_parallelism().unwrap_or(NonZeroUsize::MIN)
+            });
+            {
+                let mut pool_conn = get_pool_connection(&connection_pool)?;
+                reset_database(&mut pool_conn)?;
+            }
+            // resetting the database ensures we start from a clean state.
+            // thus we can accept an unsuccessful completion of the restore, as subsequent
+            // runs rely on the latter predicate.
```
tomxey:
```suggestion
            // resetting the database ensures we start from a clean state.
            // thus we can accept an unsuccessful completion of the restore, as subsequent
            // runs reset the database again.
```

crates/iota-indexer/src/main.rs
```diff
@@ -135,6 +153,41 @@ async fn main() -> Result<(), IndexerError> {
...
+            let store = PgIndexerStore::new(connection_pool.clone(), indexer_metrics.clone());
+            let restore = start(
```
tomxey: nit: `start` is a very unclear name in the context of `main` file. what would you say about using it like `restore::start`?

similarly above, `restore::available_epochs`

crates/iota-indexer/src/restore/persist.rs
tomxey: we could extend it with writing the 'lowest unpruned key'

sergiupopescu199 (approved):
lgtm ✨

crates/iota-indexer/src/main.rs
kodemartin: `available_epochs` has been replaced by the `FormalSnapshotStore` method.

https://github.com/iotaledger/iota/pull/12078/commits/9585bd7523a4e9c80bbc2c6062de3253d6c03a11

crates/iota-indexer/src/restore/setup.rs
kodemartin: Used the default for `object_store_connection_limit`, but kept `no_sign_request: true` as the default would not be consistent semantically.

https://github.com/iotaledger/iota/pull/12078/commits/9ce7cfb4292676ad6da67cb5492b49206d31fe17

tomxey (approved):

PR #12082, opened 2026-06-30
feat: implement ClaimAccount transaction kind

### LINKED ISSUE #12011: [AA] Create a new TransactionKind used for generating Account objects
- must only allow SmartAccount object types right now
    - however, in the future it could allow to claim and build other types; this means, for instance, that the transaction kind variant should have a name like `BuildAccount` instead of `BuildSmartAccount` and that it should be implemented in a way that it could be expanded in the future (e.g., with an enum).
- should allow to have some degree of composability
    - take as reference the current methods of the SmartAccount; it would be good to have a way to implement a "small-PTB" logic such that a user is not limited to only call `build_v1` but use the other SmartAccount builder accessories.

valeriyr:
# Description of change

> [!IMPORTANT]  
> The feature branch is noticeably outdated; it needs to be rebased on develop to port the recent changes from there. It will be rebased once https://github.com/iotaledger/iota/issues/12829 is done.

Adds the `ClaimAccount` user transaction kind: the holder of a keypair claims the address
derived from their public key as an on-chain account object whose id *is* that address.

The transaction carries an `AccountClaimKind`, currently only `SmartAccount(SmartAccountClaim)`.
A claim names the key as a signature-scheme flag plus raw bytes — not a `PublicKey` — so that
every scheme the framework accepts can be expressed, including `MultiSig`, and a
`SmartAccountBuildKind` selecting whether the created object is `Mutable` (shared) or
`Immutable` (frozen). The types live in the SDK (iotaledger/iota-rust-sdk#1239); this PR bumps
the pinned revision and implements the node side.

**Execution.** `setup_claim_account` builds a one-call PTB and runs it in
`execution_mode::System`, because the framework's `smart_account::claim_account_v1` and
`claim_immutable_account_v1` are private on purpose: the object they create has an id equal to
a signature-derivable address, so the `ClaimAccount` transaction kind must stay the only way
such an object can come into existence. The PTB is built by the node, never by the sender —
the sender only chooses the key and the build kind.

**Move.** `claim::claim_address` returns the deterministic `UID` for the sender's
address without touching the `ClaimRegistry` shared object, which keeps the transaction free of
shared inputs and therefore off the consensus path. `smart_account::claim_builder` wires that
`UID` to the built-in authenticator for the key's scheme. Both new claim functions are private
and reachable only from the node's system-mode PTB; `claim_address` is `public(package)`.

**Gating.** A new `enable_claim_account_transaction` protocol feature flag gates the
transaction kind, enabled at protocol version 30 for everything except testnet and mainnet. It
is checked in `TransactionKind::validity_check`, so the transaction is rejected
before execution. Smart-account claims additionally require
`enable_builtin_move_authenticators`, since the claimed account is authenticated by the
built-in authenticator for its key's scheme: claiming without them would create an account
that can never authenticate a transaction, at an address that can only be claimed once.

**Validation.** The claim carries the key as raw bytes and execution builds the Move
`PublicKey` from them without going through `public_key::create`, so `validity_check` applies
that function's rules up front: a scheme valid for account keys (`Ed25519`, `Secp256k1`,
`Secp256r1`, `MultiSig`, `Passkey`), non-empty bytes, a valid curve point, and for `MultiSig` a
canonically encoded committee whose members are themselves valid. A failure at execution time
would be a node bug and is reported as an invariant violation.

Also in this PR: `claim::claim_address` added to the id-leak fresh-id function lists
(bytecode verifier and Move compiler); the multisig member curve check deduplicated into
`iota_types::account_abstraction::public_key`; the kind surfaced through JSON-RPC
(`IotaClaimAccountTransaction`), GraphQL (`ClaimAccountTransaction`) and the BCS format
snapshot; a `ClaimAccount` member added to the gRPC server's transaction-kind filter, which
otherwise hits `unimplemented!()` on the first claim transaction it filters — with
`panic = 'abort'` that takes the node down; and sponsorship permitted for this kind, so a third
party can pay for someone else's claim.

## Known gaps

- **Double-claiming is not prevented yet.** `claim_address` leaves that to its caller and
  `claim_builder` does not implement it, so a second claim for the same address succeeds and
  re-creates the account object under the same id at a higher version — for `Immutable` this
  overwrites a frozen object, for `Mutable` it re-shares the object with a new initial shared
  version. `test_claim_account_twice_is_not_yet_prevented` pins that behaviour so the fix has a
  test to flip; it is being addressed in a dedicated PR.
- **gRPC clients cannot select claim transactions specifically.** The server now recognises the
  kind, but `iota.grpc.v1.filter.TransactionKind` in the SDK proto has no `CLAIM_ACCOUNT`
  member, so no client can encode it in a `TransactionKindsFilter`. Filtering by `SYSTEM`
  correctly excludes claims and `PROGRAMMABLE` correctly does not match them; adding the proto
  member is a follow-up in the SDK.
- The SDK's `TransactionBuilder` only emits programmable kinds, so a claim has to be assembled
  by hand for now.

fixes #12011


crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -70,6 +75,28 @@ public(package) fun claim(
...
+///
+/// Double-claim prevention is not enforced here: the caller is responsible for
+/// ensuring an address is claimed at most once.
+///
+/// Aborts with `EClaimAccountTransactionNotEnabled` if the
+/// `enable_claim_account_transaction` feature flag is disabled, or with
+/// `EAddressMismatch` if `public_key` does not derive to the sender.
+///
+/// `public(package)` — only callable from within the iota-framework package.
+public(package) fun claim_address(public_key: PublicKey, ctx: &TxContext): UID {
```
miker83z: I'd place it already in a dedicated module, so the registry removal will be easier later. Maybe a `claim.move` under the `account_abstraction` folder.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
```diff
@@ -70,6 +75,28 @@ public(package) fun claim(
...
+    assert!(
+        protocol_config::is_feature_enabled(b"enable_claim_account_transaction"),
+        EClaimAccountTransactionNotEnabled,
+    );
```
miker83z: I think, if we place this at the execution level, we would make our life harder in the future. Instead of making it fail with an assert here, if the feature will be ever disabled, it will be more convenient to work at the tx validation level. Having this function fail at execution is something we need to avoid.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -3022,6 +3032,9 @@ impl ProtocolConfig {
                         cfg.builtin_move_authenticator_cost_base = Some(0);
                         // Enable claim registry in devnet only.
                         cfg.feature_flags.enable_claim_registry = true;
+                        // Enable claiming an account for the sender's address in
+                        // devnet only.
+                        cfg.feature_flags.enable_claim_account_transaction = true;
```
bingyanglin, line 3037: Is version bump needed?

valeriyr: No, it is not necessary here; we will move the new parameters to a correct version once the feature branch is rebased on develop. It is noticeably outdated today.

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
valeriyr: moved in 8645610102

crates/iota-framework/packages/iota-framework/sources/claim_registry.move
valeriyr: removed in 8645610102

miker83z (approved):

crates/iota-indexer/src/read.rs
```diff
@@ -1502,10 +1509,14 @@ impl IndexerReader {
...
+                        format!(
+                            "tx_kind NOT IN ({}, {})",
+                            IotaTransactionKind::ProgrammableTransaction as u8,
+                            IotaTransactionKind::ClaimAccount as u8,
+                        )
                     } else {
-                        // No filter applied if both exist
-                        "1 = 1".to_string()
+                        // Both cover every kind but the other user kinds
+                        format!("tx_kind != {}", IotaTransactionKind::ClaimAccount as u8)
```
valeriyr: @iotaledger/infrastructure, please pay attention to this change.
I made this change to preserve consistency, since `ClaimAccount` is neither a system nor a ProgrammableTransaction.
This code requires refactoring, which is out of scope for this PR.

Thoralf-M (approved):

thibault-martinez (commented):
Can't comment on it since it's not changed but this would probably fail ?
https://github.com/iotaledger/iota/blob/46c8c327ef69b190d5e42469714485881a1fac6c/crates/iota-analytics-indexer/src/handlers/transaction_handler.rs#L134-L137

crates/iota-indexer/src/read.rs
kodemartin: refactored and handled all possible combinations with 089b0dcac4

kodemartin (approved):
lgtm for the infra changes

VorobyevIlya (approved):
Approve consensus-owned files

kodemartin (approved):

PR #12083, opened 2026-06-30
feat(indexer): populate displays during formal-snapshot restore

### LINKED ISSUE #12059: [indexer]: restore displays from formal snapshot

kodemartin:
# Description of change

This patch enables the population of the `displays` table while restoring from a formal snapshot. 

During restore the indexer only has access to live `0x2::display::Display<T>`  objects, as opposed to the on-chain `DisplayVersionUpdatedEvent` that live ingestion encodes as `bcs` into the `displays` table. Rendering however only ever needs the display fields stored in both the object and the event.

With these changes we generalize the table schema to allow encoding the fields directly as well, which becomes the default from this version and on. The kind of bcs encoding is distinguished by a new discriminator, that keeps existing databases valid with no backfill.

The changes can be summarized as follows:

  - Added a new `bcs_kind` column in `displays` to distinguish between the two types of bcs encoding stored in `display::bcs`: These are the legacy event, and the new display-fields type.
  - Added conversion methods for `StoredDisplay`  from a live `Display<T>` object, used by the restore path to populate displays as it streams objects.
  - Both live ingestion and restore now persist the fields-only encoding.
  - Generalized display persistence to chunk large batches and write them in parallel, matching how objects are persisted.

> [!NOTE]
> Since the display fields are stored in the display object as well, the idea of redirecting reads to the `objects` table and dropping `displays` was also investigated. This essentially put into question the necessary for a dedicated index for display fields. The `object` table does have an index that supports queries of objects by full-qualified type. However the sizes of the two tables differ by orders of magnitude. In `testnet` for example, `displays` count to hundreds whereas `objects` count to hundreds of millions. This implies that even index lookups on the `objects` table would be less efficient than on the dedicated index.

Resolves #12059

QA is planned as part of #12061 

### Release Notes

- [x] Indexer: Introduces a new migration to add a `bcs_kind` column in `displays` to allow encoding different values in the `bcs` column. :warning: Operators should take all necessary measures to manage any downtime that the migration might incur. Nevertheless, minimal downtime is expected in this case.

### DIFF
--- a/crates/iota-indexer/src/restore/persist.rs
+++ b/crates/iota-indexer/src/restore/persist.rs
@@ -82,7 +82,6 @@ impl Restore for PgIndexerStore {
                     "failed to persist all formal snapshot object chunks: {e:?}",
                 ))
             })?;
-        // TODO: enable chunking
         self.persist_displays(displays.into_values().collect())
             .await?;
         Ok(())
--- a/crates/iota-indexer/src/restore/persist.rs
+++ b/crates/iota-indexer/src/restore/persist.rs
@@ -1,6 +1,8 @@
 // Copyright (c) 2026 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::BTreeMap;
+
 use bytes::Bytes;
 use fastcrypto::hash::{HashFunction, Sha3_256};
 use iota_core::authority::authority_store_tables::LiveObject as SnapshotObject;
@@ -16,6 +18,7 @@ use crate::{
     ingestion::{common::prepare::LiveObject, primary::persist::EpochToCommit},
     models::{
         checkpoints::StoredChainIdentifier,
+        display::StoredDisplay,
         epoch::{EndOfEpochUpdate, StartOfEpochUpdate, extract_epoch_info_event},
     },
     pruning::pruner::PrunableTable,
@@ -31,6 +34,7 @@ impl Restore for PgIndexerStore {
         expected_checksum: &[u8; SHA3_BYTES],
     ) -> anyhow::Result<()> {
         let mut hasher = Sha3_256::default();
+        let mut display_updates = BTreeMap::new();
         let partition = LiveObjectIter::new(&file_metadata, bytes)?.scan(
             &mut hasher,
             |hasher,
@@ -39,6 +43,9 @@ impl Restore for PgIndexerStore {
                  previous_transaction_checkpoint,
              }| {
                 hasher.update(object.object_ref().digest.inner());
+                if let Some(display) = StoredDisplay::try_from_object(&object) {
+                    display_updates.insert(display.object_type.clone(), display);
+                }
                 let checkpoint_sequence_number =
                     previous_transaction_checkpoint.unwrap_or_default();
                 Some(LiveObject::new(checkpoint_sequence_number, object))
@@ -75,6 +82,8 @@ impl Restore for PgIndexerStore {
                     "failed to persist all formal snapshot object chunks: {e:?}",
                 ))
             })?;
+        // TODO: enable chunking
+        self.persist_displays(display_updates).await?;
         Ok(())
     }
 }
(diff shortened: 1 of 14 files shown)


crates/iota-indexer/src/restore/persist.rs
```diff
@@ -37,6 +41,9 @@ impl Restore for PgIndexerStore {
                  previous_transaction_checkpoint,
              }| {
                 hasher.update(object.object_ref().digest.inner());
+                if let Some(display) = StoredDisplay::try_from_object(&object) {
+                    displays.insert(display.object_type.clone(), display);
```
tomxey, line 47: if we would have more than one version of display for the same object type, will it overwrite here?

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -1950,35 +1945,54 @@ impl IndexerStore for PgIndexerStore {
...
         }
 
-        self.spawn_blocking_task(move |this| this.persist_display_updates(display_updates))
-            .await?
+        if displays.len() < self.config.parallel_objects_chunk_size {
+            self.spawn_blocking_task(move |this| this.persist_displays_chunk(displays))
+                .await??;
+            return Ok(());
+        }
+        let chunks = chunk!(displays, self.config.parallel_objects_chunk_size);
```
tomxey, line 1954: Would such case be a problem for the general purpose chunking logic?

I'm wondering if it's worth to optimize for such case.

crates/iota-indexer/src/store/pg_indexer_store.rs
```diff
@@ -253,15 +249,14 @@ impl PgIndexerStore {
...
-    ) -> Result<(), IndexerError> {
+    fn persist_displays_chunk(&self, mut displays: Vec<StoredDisplay>) -> Result<(), IndexerError> {
         transactional_blocking_with_retry!(
             &self.blocking_cp,
             {
-                let value = display_updates.values().collect::<Vec<_>>();
-                |conn| self.persist_displays_in_existing_transaction(conn, value)
+                // The parent uses an FnMut closure that mutably borrows the `displays`
+                // So we can't move the value. To avoid cloning we use `std::mem::take`.
+                let displays = std::mem::take(&mut displays);
```
tomxey: this would zero out the original `displays` vector and in case of a retry it would be empty, or am I missing something?

crates/iota-indexer/src/restore/persist.rs
kodemartin: certainly. how can we have more than one version of display though?

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: Could you elaborate on this? What kind of problem?

crates/iota-indexer/src/store/pg_indexer_store.rs
kodemartin: you're absolutely right, good catch!

kodemartin: 2724e3153e

crates/iota-indexer/src/restore/persist.rs
tomxey: ah, okay, indeed it seems to not be possible, nvm

crates/iota-indexer/src/store/pg_indexer_store.rs
tomxey: I mean, we are handling the special case `if displays.len() < self.config.parallel_objects_chunk_size {`, but wouldn't it be handled automatically by the general purpose chunking logic below?

Basically, can we remove the 
```rust
        if displays.len() < self.config.parallel_objects_chunk_size {
            self.spawn_blocking_task(move |this| this.persist_displays_chunk(displays))
                .await??;
            return Ok(());
        }
```

block?

kodemartin: oh, I see. Well a single `chunk` is not as trivial as persisting directly the batch, because the `persist_chunks` pattern uses `try_join` internally, which adds overhead. So we avoid this by the control statement here.

tomxey: ok, got it

tomxey (approved):

sergiupopescu199 (approved):
lgtm 🐲

PR #12091, opened 2026-07-01
refactor(iota-types): replace node CheckpointSummary with SDK type

DaughterOfMars:
## Why

Continues the effort to remove duplicated checkpoint-family types from `iota-types` and adopt their canonical `iota_sdk_types` counterparts (cf. `Object` #11718, `CheckpointCommitment` #11976, `EndOfEpochData` #12027). This is **PR 3 of the checkpoint-family track** — `CheckpointSummary` is the lynchpin, embedded in `Envelope<CheckpointSummary, S>` (`CertifiedCheckpointSummary` / `SignedCheckpointSummary`), which stay node-side.

## What

- Drop the node `struct CheckpointSummary` and re-export `iota_sdk_types::checkpoint::CheckpointSummary`. It is BCS-identical: the node digest aliases (`CheckpointContentsDigest` / `CheckpointDigest`) are already `= iota_sdk_types::Digest`, so every field type matches.
- Keep `impl Message for CheckpointSummary` (local trait, foreign type) so `Envelope` keeps working unchanged. The SDK type carries an inherent `digest()` that is byte-identical to `Message::digest()` (both Blake2b256 over `"CheckpointSummary::"` + BCS), so direct `.digest()` calls resolving to the inherent method preserve behavior.
- Move the node-only helpers to a `CheckpointSummaryExt` trait: `new_with_protocol_config` (the protocol-config-aware constructor, named to make the `ProtocolConfig` dependency explicit and avoid masquerading as an inherent `new`), `verify_epoch`, `sequence_number`, `timestamp`, `next_epoch_committee`, `report_checkpoint_age`, `is_last_checkpoint_of_epoch`, `version_specific_data`.
- Delete the now-redundant `CheckpointSummary` `TryFrom` impls from `iota_sdk_types_conversions.rs` (they would collide with the reflexive `TryFrom<T> for T`), and simplify the `CertifiedCheckpointSummary` ↔ `SignedCheckpointSummary` conversions whose `checkpoint` field is now the same type on both sides.
- Drop the node `Display` impl (cannot be implemented on a foreign type); the lone consumer is a transactional-test-runner debug command, switched to `{:?}`.

## Format

The BCS wire format is unchanged. The serde-reflection registry snapshot moves `version_specific_data` from `SEQ: U8` to `BYTES` because the SDK field serializes via `serialize_bytes` — BCS encodes both identically (verified empirically across empty/single/>127-length inputs).

## Test plan

- `cargo check --workspace --all-targets`, `cargo ci-clippy` — clean.
- `iota-core` format snapshot test passes after regeneration (inspected; representational-only).
- 821 `iota-types` / `iota-core` lib + format tests pass.

## Follow-ups (separate)

`sequence_number`, `next_epoch_committee`, and `is_last_checkpoint_of_epoch` are pure accessors that belong in `iota_sdk_types`; a companion SDK PR adds them (plus per-field getters). Once the shared rev is bumped, those are removed from `CheckpointSummaryExt` and call sites bind to the SDK methods.


crates/iota-types/src/messages_checkpoint.rs
```diff
@@ -168,7 +134,40 @@ impl CheckpointSummary {
         end_of_epoch_data: Option<EndOfEpochData>,
         timestamp_ms: CheckpointTimestamp,
         randomness_rounds: Vec<RandomnessRound>,
-    ) -> CheckpointSummary {
+    ) -> Self;
+
+    fn verify_epoch(&self, epoch: EpochId) -> IotaResult;
+
+    fn sequence_number(&self) -> &CheckpointSequenceNumber;
```
Thoralf-M: Can be removed here since you added it to the SDK?

DaughterOfMars: I haven't updated the rev because there are other changes that need to be made first

Thoralf-M: What other changes? Is there an issue for it or are you working on it already?

DaughterOfMars: Other changes in the SDK that haven't been added to the core repo. No there is no issue and I am not working on it yet

jkrvivian (approved):

crates/iota-types/src/messages_checkpoint.rs
```diff
@@ -156,8 +119,11 @@ impl Message for CheckpointSummary {
     }
 }
 
-impl CheckpointSummary {
-    pub fn new(
+/// Node-only helpers for [`CheckpointSummary`], which is defined in
+/// `iota_sdk_types`. These live on an extension trait because inherent methods
+/// cannot be added to a type that is foreign to this crate.
+pub trait CheckpointSummaryExt: Sized {
```
thibault-martinez: Seal the trait like the other Ext traits

crates/iota-types/src/messages_checkpoint.rs
thibault-martinez: Ok but shouldn't we deal with the changes so that we can properly do this PR?

Thoralf-M: I think it's only the digest changes that are done in https://github.com/iotaledger/iota/pull/12093 so after it's merged this PR should be updated with the SDK rev

roman1e2f5p8s (approved):

crates/iota-types/src/messages_checkpoint.rs
Thoralf-M: Can be done now

piotrm50 (approved):

crates/iota-types/src/messages_checkpoint.rs
thibault-martinez: https://github.com/iotaledger/iota/pull/12091/commits/ea1e78f0a685d219cb84d2e2a116e9216373a551

Thoralf-M (approved):

sergiupopescu199 (approved):

Dkwcs (approved):

valeriyr (approved):

PR #12107, opened 2026-07-01
feat(dev-tools): private-network experiment tooling

VorobyevIlya:
# Description of change

Tooling for running Starfish/consensus experiments on the local private network. All changes live under `dev-tools/iota-private-network/`:

- **Experiment scripts**: a half-network rolling-upgrade test, a post-run canary that filters shutdown-window noise, plus a StarfishOverview dashboard link and an `iota-spammer` precheck wired into the benchmark/fuzz runners.
- **Build self-healing**: verify built images match `HEAD` and rebuild if stale, and recover from a stale cargo cache on build failure (staleness auto-prune narrowed to cache mounts only).
- **Experiment lock**: `cleanup`/`bootstrap`/`run` refuse to start while another experiment holds the lock, avoiding concurrent runs stepping on each other.

### DIFF
--- a/dev-tools/iota-private-network/experiments/experiment_common.py
+++ b/dev-tools/iota-private-network/experiments/experiment_common.py
@@ -797,6 +797,7 @@ def start_grafana(grafana_dir: Path, override_file: str | None = None) -> None:
     run_timed(cmd, "Starting monitoring stack", cwd=grafana_dir)
     print()
     log(f"  Grafana: {_C.CYAN}http://localhost:3000/dashboards{_C.RESET}")
+    log(f"  StarfishOverview: {_C.CYAN}http://localhost:3000/d/StarfishOverview{_C.RESET}")
     log(f"  Prometheus: {_C.CYAN}http://localhost:9090/targets{_C.RESET}")
 
 
@@ -941,6 +942,23 @@ def _update_or_clone_benchmark_repo() -> None:
     run_timed(
         ["git", "clone", NETWORK_BENCHMARK_REPO, str(NETWORK_BENCHMARK_DIR)],
         "Cloning network-benchmark", check=False,
+    )
+
+
+def ensure_iota_spammer_script() -> None:
+    """Verify the iota-spammer fuzz script is reachable BEFORE the network
+    comes up. Mirrors `ensure_stress_image` for the iota-spammer backend:
+    skipping the check leaves the run to fail in `start_spammer` ~5 minutes
+    in, after bootstrap + validator startup. The script lives in a separate
+    private repo, so a missing checkout is a likely-enough misconfiguration
+    to be worth surfacing up front."""
+    script = Path.home() / "iota-spammer" / "scripts" / "spamming_fuzz_test.sh"
+    if script.is_file():
+        return
+    raise RuntimeError(
+        f"iota-spammer spammer requested but script not found at {script}; "
+        "clone github.com/iotaledger/iota-spammer to ~/iota-spammer "
+        "or select the stress backend (-C stress)"
     )
 
 
--- a/dev-tools/iota-private-network/experiments/experiment_common.py
+++ b/dev-tools/iota-private-network/experiments/experiment_common.py
@@ -338,6 +338,12 @@ def acquire_single_run_lock(runner: str) -> None:
     )
     fh.flush()
     _run_lock_fh = fh  # keep the fd open: the flock dies with the process
+    # Signal child processes (bootstrap.sh -> cleanup.sh chain, plus any
+    # direct invocations of cleanup.sh / run.sh from within an orchestrator)
+    # that the shared lock is already held by their parent, so their own
+    # lock-checks should not fire. Default subprocess.run inherits env, so
+    # setting this in os.environ propagates cleanly.
+    os.environ["IOTA_EXPERIMENT_LOCK_HELD"] = "1"
 
 
 def _lock_holder_pid(holder: str) -> int | None:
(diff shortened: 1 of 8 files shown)


dev-tools/iota-private-network/experiments/experiment_common.py
```diff
@@ -1200,6 +1224,123 @@ def run_loop(cfg, prefix: str, final_prefix: str) -> None:
...
+    def _in_shutdown_window(idx: int, ts: datetime) -> bool:
+        for lo, hi in stop_windows.get(idx, ()):
+            if lo <= ts <= hi:
+                return True
+        return False
+
+    filtered = 0
+    unknown: list[str] = []
+    for archive in archives:
+        idx = int(archive.stem.rsplit("-validator-", 1)[1])
```
bingyanglin, line 1312: If a marker line's timestamp fails to parse, it's skipped instead of flagged. For a safety canary that means a real panic with an odd timestamp silently passes. Maybe treat an unparseable timestamp as `unknown` (fail open) instead of `continue`?

piotrm50 (approved):

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
```diff
@@ -0,0 +1,595 @@
...
+        type=lambda v: v.lower() in ("true", "1", "yes"),
+        help="Whether to build the local upgrade image (default: true)",
+    )
+    parser.add_argument(
+        "-n", "--num-validators",
+        default=19,
+        type=int,
+        choices=range(2, 101),
+        metavar="N",
+        help="Number of validators to run (2-100, default: 19)",
```
polinikita: validate_num_validators() rejects anything above 30. Align that

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
```diff
@@ -0,0 +1,595 @@
...
+
+def phase11_second_half(cfg: Config, second_half: list[int]) -> None:
+    phase_start = time.time()
+    log(_phase_banner(f"Rolling upgrade — second half ({len(second_half)} validators)", "PHASE 11"))
+    log(f"  Upgrading: {second_half}")
+    rolling_upgrade(cfg, second_half, label="second half")
+    log(_phase_complete("Phase 11", time.time() - phase_start))
+
+
+def phase9_observe_mixed(cfg: Config, epoch_0_start: float, epoch_0: int) -> tuple[int, float]:
```
polinikita: It reads each binary’s maximum supported protocol version and only logs expectations. A wrong on-chain protocol transition still ends successfully. It might make sense to assert if something goes wrong

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
```diff
@@ -0,0 +1,595 @@
...
+
+# Copyright (c) 2026 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+"""Half-network upgrade test.
+
+Brings up all validators on a released image, then mid-epoch-0 rolls the
+locally-built upgrade image out to a deterministic random half of the
+network. The mixed-binary network holds the lower of the two halves'
+MAX_PROTOCOL_VERSION values (the new-binary half is below 2f+1). Mid-epoch-1,
```
polinikita: Is not that starting mid epoch-2?

polinikita (commented):
Left a few comments

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
VorobyevIlya: yes, fixed

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
VorobyevIlya: added version check and raise an error if it doesn't coincide with expected

dev-tools/iota-private-network/experiments/run-half-upgrade-test.py
VorobyevIlya: fixed

dev-tools/iota-private-network/experiments/experiment_common.py
VorobyevIlya: yes, that is a better approach; done

dev-tools/iota-private-network/run.sh
```diff
@@ -6,14 +6,32 @@
...
+# Refuse to bring up validators over another active run's network — same
+# `iota-network` and container names, so recreates / port conflicts ensue.
+# Bypass via IOTA_EXPERIMENT_LOCK_HELD or an explicit -f.
+LOCK_PATH="/tmp/iota-experiments.lock"
+# Read-only FD + shared flock avoid a false "locked" under fs.protected_regular
+# when a dead orchestrator left the lock file behind.
+if [ "$FORCE" != "true" ] \
+   && [ "${IOTA_EXPERIMENT_LOCK_HELD:-0}" != "1" ] \
+   && [ -f "$LOCK_PATH" ] \
+   && ! (flock -n -s 9) 9<"$LOCK_PATH" 2>/dev/null; then
```
bingyanglin: `flock` isn't on stock macOS, so `(flock -n -s 9)` returns 127 and `! 127` is true. Once the lock file exists, these scripts always report "another experiment run is active" on a Mac. Maybe gate the block on `command -v flock >/dev/null` so a missing `flock` doesn't read as locked? (same pattern in `cleanup.sh` and `bootstrap.sh`)

VorobyevIlya: good catch; done — flock(1) when present, else a python3 fcntl probe (macOS has no flock), and an explicit "cannot be verified" error + -f hint when neither tool can answer

bingyanglin (approved):

polinikita (approved):

PR #12114, opened 2026-07-01
fix(core): Distinguish NotFound and other errors in object-store exists

### LINKED ISSUE #11726: refactor(iota-storage): avoid downloading file in object_store::util::exists
Nit: This function effectively downloads the file, If possible we could leverage the [head](https://docs.rs/object_store/latest/object_store/trait.ObjectStoreExt.html#tymethod.head) method to check if a file exists without downloading its contents.

_Originally posted by @sergiupopescu199 in https://github.com/iotaledger/iota/pull/11722#discussion_r3347672954_

### LINKED ISSUE #12108: [iota_storage]: utils::exists returns false negatives
## Description

`iota_storage::object_store::exists` is defined as:

https://github.com/iotaledger/iota/blob/ec5134bbb273d00aa9fe9d82d86857334e65f771/crates/iota-storage/src/object_store/util.rs#L101-L103

It returns `false` both when the object genuinely does not exist and when the store is unreachable (network/DNS/5xx). Thus a transient error gets reported as a definitive absence.

## Expected behaviour

`exists` should return `false` only if the object is absent, and an error in case of transient client errors.

jkrvivian:
# Description of change

Add `ObjectStoreGetExt::exists` returning `Result<bool>`: `Ok(false)` for a
genuine absence, `Err` for other failures. Each backend detects absence
natively using a metadata-only primitive rather than downloading the body:

- `Arc<dyn ObjectStore>` uses `head` and matches `Error::NotFound`
- S3/GCS issue an HTTP `HEAD` and classify by status (2xx / 404 / other)
- `LocalStorage` uses `Path::try_exists`

Fixes #12108 
Fixes #11726

### DIFF
--- a/crates/iota-storage/src/lib.rs
+++ b/crates/iota-storage/src/lib.rs
@@ -9,27 +9,19 @@ use std::{
     fs::File,
     io,
     io::{BufReader, Read, Write},
-    ops::Range,
     path::PathBuf,
-    sync::{
-        Arc,
-        atomic::{AtomicU64, Ordering},
-    },
+    sync::Arc,
 };
 
 use anyhow::{Result, anyhow};
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use bytes::{Buf, Bytes};
 use fastcrypto::hash::{HashFunction, Sha3_256};
-use futures::StreamExt;
 use iota_types::{
     committee::Committee,
-    messages_checkpoint::{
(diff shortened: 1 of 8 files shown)


bingyanglin (approved):

kodemartin (commented):
looks good! ⚡ 

By refactoring to using `Method::HEAD` I think this also resolves #11726

kodemartin (approved):

muXxer (approved):

PR #12118, opened 2026-07-01
fix(core): amortize checkpoint interval over a sliding window

### LINKED ISSUE #12106: perf(starfish): sliding-window block creation raises block rate but lowers checkpoint rate
Sliding-window block creation improves block creation rate across all network sizes, but checkpoint creation rate has regressed consistently — most notably on large networks.

**Observed:**

- Before (fixed 50ms minimum block interval): block rate **18.9–19.6 b/s**, checkpoint rate **4.85 c/s**
- After (sliding window): block rate **20.0 b/s** (p5 ~19.9), checkpoint rate **~4.22–4.27 c/s**

A higher block rate now yields a *lower* checkpoint rate. The regression scales with committee size — large networks are hit hardest.

**Key points:**

- Regression is consistent across networks and stable over time, not a one-off run.
- Block rate and checkpoint rate have decoupled: the block-creation win does not translate into checkpoint throughput.
- The checkpoint-rate drop coincides exactly with the block-rate step up to 20 b/s (see chart), and persists afterward.

Image

polinikita:
# Description of change

Checkpoint creation rate regressed after sliding-window block creation (#11790) raised the block/commit rate: block rate rose to the 20 b/s ceiling but checkpoint rate fell (~4.85 → ~4.25 c/s, worst on large committees).

`min_checkpoint_interval_ms = 200` caps the checkpoint rate to bound the storage and bandwidth cost of creating, certifying, and syncing checkpoints; the rate also drives end-to-end latency for consumers that wait on verified checkpoints. This PR keeps the same ~5 c/s asymptotic cap while restoring the rate to it under the relaxed block-creation rules.

## Root cause

`maybe_build_checkpoints` gated each checkpoint on the previous one, then advanced the anchor to the *actual* triggering commit timestamp:

```
build when  current_commit_ts >= last_checkpoint_ts + min_checkpoint_interval_ms
then        last_checkpoint_ts = current_commit_ts
```

Commit timestamps are discrete, so the trigger overshoots the deadline every time. Anchoring on the actual timestamp makes every checkpoint pay that overshoot, pushing sustained spacing above `interval` — worse as higher block rates coarsen the timestamp grid.

## Fix

Keep the strict adjacent check and add a windowed one; checkpoint `k` builds when **either** holds:

```
adjacent:  current_commit_ts >= timestamp(checkpoint k−1) + interval
windowed:  current_commit_ts >= timestamp(checkpoint k−N) + N * interval
```

The windowed arm amortizes the overshoot over `N` checkpoints (paid once per window, not per checkpoint), restoring the rate to the ceiling; the adjacent arm keeps every gap — hence every coalesced checkpoint — within one interval, avoiding the burst-then-jumbo clustering a window-only rule drifts into. Both arms consume ≥ one interval of commit time per checkpoint, so the sustained ~5 c/s cap is preserved. `N` is the new protocol parameter `checkpoint_rate_window_size`.

## Determinism

Checkpoint boundaries must be identical on every validator or checkpoints cannot be certified. The gate reads only deterministic data: commit timestamps and this epoch's already-built checkpoint timestamps, from the per-epoch builder summaries (`get_built_checkpoint_summary`). Every validator builds every checkpoint of its epoch, so the look-back resolves to the same summary everywhere — or to none, exactly when it would cross the epoch start (windowed arm inert). No wall-clock, no per-node or cross-epoch state.

## Rollout

`checkpoint_rate_window_size` is set to `20` at **protocol version 31, devnet only** (non-Mainnet/Testnet); unset at versions ≤ 30 and on Mainnet/Testnet, preserving the strict adjacent behavior. It takes effect at the next epoch boundary that upgrades a network to v31.

## Observability

New `commits_per_checkpoint` histogram (commits coalesced per checkpoint, buckets 1–10 plus 10+) surfaced as a "Commits per checkpoint" heatmap on the Starfish dashboard; under load a checkpoint typically holds 3–5 commits (average ~4).

## Validation

Tested on local multi-validator deployments under varying network conditions: the sustained checkpoint rate rose from ~4.2 to ~4.95 c/s, matching the predicted ceiling.
image

Fixes #12106

### Release Notes

- [x] Protocol: Adds a new protocol parameter `checkpoint_rate_window_size`, enabled on devnet, which amortizes the minimum checkpoint interval over a sliding window of checkpoints so the sustained checkpoint rate holds at its ceiling instead of sagging.

### DIFF
--- a/crates/iota-core/src/checkpoints/mod.rs
+++ b/crates/iota-core/src/checkpoints/mod.rs
@@ -1193,6 +1193,7 @@ impl CheckpointBuilder {
             // Min interval has elapsed, we can now coalesce and build a checkpoint.
             last_height = Some(height);
             last_timestamp = Some(current_timestamp);
+            let commits_in_checkpoint = grouped_pending_checkpoints.len();
             debug!(
                 checkpoint_commit_height_from = grouped_pending_checkpoints
                     .first()
@@ -1208,6 +1209,11 @@ impl CheckpointBuilder {
                 .await
             {
                 Ok(seq) => {
+                    // Count only on success; a failed build retries the same
+                    // group and would otherwise double-count it.
+                    self.metrics
+                        .commits_per_checkpoint
+                        .observe(commits_in_checkpoint as f64);
                     // Advance the window anchor to the highest checkpoint just
                     // built (a single call may emit several when chunked).
                     last_seq = Some(seq);
--- a/crates/iota-core/src/checkpoints/mod.rs
+++ b/crates/iota-core/src/checkpoints/mod.rs
@@ -1114,14 +1114,21 @@ impl CheckpointBuilder {
             .epoch_store
             .last_built_checkpoint_builder_summary()
             .expect("epoch should not have ended");
-        let mut last_height = summary.clone().and_then(|s| s.checkpoint_height);
-        let mut last_timestamp = summary.map(|s| s.summary.timestamp_ms);
+        let mut last_height = summary.as_ref().and_then(|s| s.checkpoint_height);
+        let mut last_timestamp = summary.as_ref().map(|s| s.summary.timestamp_ms);
+        let mut last_seq = summary.map(|s| s.summary.sequence_number);
 
         let min_checkpoint_interval_ms = self
             .epoch_store
             .protocol_config()
             .min_checkpoint_interval_ms_as_option()
             .unwrap_or_default();
+        // When set, the interval may additionally be amortized over this many
+        // recent checkpoints of the current epoch.
+        let checkpoint_rate_window_size = self
+            .epoch_store
+            .protocol_config()
+            .checkpoint_rate_window_size_as_option();
         let mut grouped_pending_checkpoints = Vec::new();
         let mut checkpoints_iter = self
             .epoch_store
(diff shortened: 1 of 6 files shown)


VorobyevIlya (commented):
So you decided to not include minimum interval? Do we allow "burst" of checkpoints?

polinikita:
> So you decided not to include a minimum interval? Do we allow a "burst" of checkpoints?

Yes, a lower minimum interval (say of 100-150ms) can make it further more balanced. I left this idea for a while just to not introduce two parameters if this is not much needed in practice. 

With the current design and in an extreme scenario, we do allow a burst of checkpoints, each of which includes only 1 commit. Note that this is not harmful since the total load over a few seconds for checkpoint signing/signiture verification is not huge and each operation is not on a critical path. The extreme scenario when this could happen is when we had a commit before that, which was produced after a very long delay, say 200 + X milliseconds. This is generally bypassed by our leader scheduler. The commits are produced still on average pace of at most 1 commit per 50ms, so a resulting burst will be of an average length at most X/50ms.

crates/iota-core/src/checkpoints/mod.rs
```diff
@@ -1120,6 +1154,9 @@ impl CheckpointBuilder {
             // Min interval has elapsed, we can now coalesce and build a checkpoint.
             last_height = Some(height);
             last_timestamp = Some(current_timestamp);
+            self.metrics
```
bingyanglin: `commits_per_checkpoint.observe(...)` runs before the build can fail, so on the error path it re-counts the same group again on the next retry. Maybe move the observe into the `Ok(seq)` arm so only real checkpoints get counted?

polinikita: changed accrodingly, thanks

bingyanglin (approved):
Approved `iota-core/src/checkpoints`, `iota-protocol-config`, `iota-swarm-config/tests/snapshots`, and `dev-tools/grafana-local/dashboards`

VorobyevIlya (approved):

bingyanglin (approved):

kodemartin (approved):

PR #12119, opened 2026-07-02
fix(starfish-core): process the covered commit prefix of truncated fast sync fetches

### LINKED ISSUE #11984: Fast commit sync: process fetch responses incrementally instead of buffering the whole response
Follow-up to #11924.

## Problem

The fast commit-sync path (`fetch_commits_and_transactions`) buffers the entire response — commits, certifier headers, and all transactions referenced by the committed range — before processing it. The transaction payload has no per-element count cap on this path (the server returns every transaction the range references), so the honest worst-case payload for a full batch is effectively unbounded (hundreds of GiB at large committees/batch sizes). #11924 added a fixed buffer clamp (`MAX_FETCH_RESPONSE_BUFFER_BYTES`, 512 MiB) plus per-element and count checks as an interim safeguard, but a clamped response is truncated, so the path currently relies on retrying the same range to make progress.

## Goal

Avoid buffering the whole response before processing it. Options discussed:

1. **Smaller batches** — decrease `fast_commit_sync_batch_size` and fetch more batches, so each response is small enough to buffer safely.
2. **Split commit/transaction fetching** — fetch the commits first, then fetch a known number of transactions with custom batching, bounding the transaction volume per fetch by an explicit count.
3. **Stream processing** — process the response incrementally as chunks arrive, bounding peak memory to roughly one chunk regardless of batch size. This includes resolving the existing `// TODO: create subdags for prefix of commits` so a truncated/partial response makes forward progress and the remainder is fetched with a later range.

## Notes

- Measured testnet average transaction size is ~557 B and realistic batch payloads are tens of MiB, so the clamp rarely trips today — but it can at peak load, where prefix-recovery / incremental processing becomes necessary.
- See the #11924 discussion for the full memory analysis.

piotrm50:
# Description of change

The fast commit sync fetch path buffers a bounded amount of the streamed response and terminates the stream once the byte limit is reached. A response cut off this way (or by a mid-stream network error) carries all commits of the range but transactions for only some of them. Previously `fetch_once` rejected such a response with `FetchedTransactionsMismatch`, discarding everything and retrying the same range — with no forward progress if the range's transaction payload never fits the limit.

This PR resolves the existing `TODO: create subdags for prefix of commits`: when transactions are missing, the fetched batch is truncated to the longest prefix of commits whose transactions were all received, subdags are built for that prefix, and the scheduler requeues the remainder of the range via the existing `requeue_partial_range` mechanism. A prefix of a verified batch remains trusted on its own because `verify_commits` certifies the last commit with a vote quorum and all earlier commits transitively through the digest chain. The response then no longer needs to fit a full batch in memory for the syncer to make progress: peak buffering stays bounded by the response byte limit and larger ranges complete across multiple partial fetches.

The error-per-response behavior is preserved when not even the first commit's transactions were received, since such a response allows no progress. Verified voting block headers are kept in full even when the commit batch is truncated, so they remain available for serving future fetch requests. A new `commit_sync_truncated_fetches` metric counts truncated responses to make byte-limit trips observable in production.

fixes #11984

Added unit tests for the prefix-truncation helper (full coverage, partial coverage, no coverage, commits without transactions) and a behavioral test driving `fetch_once` with a response whose transaction payload covers only the first of two commits. All starfish-core lib tests (336) and starfish simtests (30) pass locally.

### DIFF
--- a/crates/starfish/core/src/commit_syncer/fast.rs
+++ b/crates/starfish/core/src/commit_syncer/fast.rs
@@ -11,6 +11,7 @@ use std::{
     time::Duration,
 };
 
+use bytes::Bytes;
 use iota_metrics::spawn_logged_monitored_task;
 use parking_lot::RwLock;
 #[cfg(not(test))]
@@ -544,7 +545,9 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
         shared_fetch_loop(inner, commit_range, 2, Self::fetch_once).await
     }
 
-    // Fetches commits and transactions from a single authority.
+    // Fetches commits and transactions from a single authority. When the
+    // response covers only part of the range, returns the prefix of commits
+    // whose transactions were all fetched.
     async fn fetch_once(
         inner: Arc<Inner<C>>,
         target_authority: AuthorityIndex,
@@ -571,7 +574,7 @@ impl<C: NetworkClient> FastCommitSyncer<C> {
         //    returned commit, and the returned commits are chained by digest,
         // so earlier commits are certified as well.
         let max_commits = inner.sync_type.max_commits_per_response(&inner.context);
-        let (commits, voting_block_headers) = Handle::current()
+        let (mut commits, voting_block_headers) = Handle::current()
             .spawn_blocking({
                 let inner = inner.clone();
                 move || {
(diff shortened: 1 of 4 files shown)


piotrm50:
Added `839c30e899`: malformed `SerializedTransactionsV2` entries in a fast-sync fetch now fail the fetch with `MalformedTransactions` (mirroring regular sync) instead of being silently skipped. Without this, a garbage entry was indistinguishable from a benign byte cutoff, so the new truncated-response fallback would treat it as a missing suffix and keep the peer in rotation while capping each fetch at the covered prefix. The deserialize loop is extracted into `process_serialized_transactions` with unit tests alongside the truncation helper's.

polinikita (approved):

VorobyevIlya (approved):

PR #12131, opened 2026-07-02
fix(iota-grpc-server): feed embedded batch errors into the traffic-control error policy

muXxer:
# Description of change

The e2e migration of the traffic-control tests off the node JSON-RPC (#12130) surfaced gaps in the fullnode gRPC traffic control around **batch APIs**:

1. **Embedded per-item errors never fed the error policy.** Batch APIs report per-item failures as `google.rpc.Status` values embedded inside an otherwise *successful* gRPC response. The `TrafficControlLayer` derives its tally from the transport-level status only, so a client spamming e.g. invalid transactions produced `Code::Ok` tallies and could never be error-blocked.
2. **A batch counted as a single request under the spam policy**, regardless of how many items it carried — so a client could dilute its request rate simply by batching.

(The first gap was originally misdiagnosed as a headers-vs-trailers issue; the `handler_errors_feed_the_error_policy` test shows unary handler errors *are* visible to the layer — tonic returns them as a trailers-only response, so `grpc-status` is in the HTTP headers. The real blind spot is the batch response body.)

## Fix

- `TrafficControlLayer` injects a `TallyHandle` (client IP + traffic controller) into the request extensions.
- Batch handlers report **each item** through the handle: every item counts as one request for the spam policy and, on a client error, feeds the error policy. The layer skips its default per-request tally when a handler has accounted for the items (no double counting).
- Endpoints covered:
  - `ExecuteTransactions`, `SimulateTransactions` — unary; tallied per response item.
  - `GetObjects`, `GetTransactions` — server-streaming; tallied per item as the response stream is consumed. Per-item read failures ride inside `Ok` responses and are tallied as `Code::Ok` (a client cannot know an object or transaction was pruned), so they feed the spam policy only. A stream-level error (e.g. a single item exceeding the request's message size limit) terminates the stream and is tallied by its status code, like a unary handler error; an empty batch produces no per-item tallies and is charged as one request up front — either would otherwise escape traffic control entirely.
- `GetObjects` / `GetTransactions` batches are now size-capped (`max_get_objects_batch_size` / `max_get_transactions_batch_size`, default 1000) and oversized batches are rejected — previously they were unbounded, which meant unbounded storage lookups and an unbounded per-item tally that could flood the tally channel. Empty batches remain allowed.
- Pagination (`ListOwnedObjects`, …), the checkpoint subscription, and single-item reads are unchanged — they are not client-submitted batches of independent lookups.

Not changed (considered): the metrics layer still derives response-code metrics from the transport-level status; counting embedded application errors as transport `OK` is arguably correct for transport metrics.

## Tests

`crates/iota-grpc-server/tests/traffic_control.rs`:

- `handler_errors_feed_the_error_policy` — unary `InvalidArgument` blocks the client.
- `embedded_batch_errors_feed_the_error_policy` — per-item errors in OK `ExecuteTransactions` responses feed the error policy (**fails without the fix**).
- `batched_requests_accrue_spam_per_item` — a single `ExecuteTransactions` batch is charged per item under the spam policy (**fails without the fix**).
- `batched_reads_accrue_spam_per_item` — a single `GetObjects` batch is charged per item under the spam policy (**fails without the fix**).
- `stream_errors_feed_the_error_policy` — a streaming read whose only stream element is an error is still tallied and feeds the error policy (**fails without the fix**).
- `empty_read_batches_accrue_spam` — empty read batches are charged as one request under the spam policy (**fails without the fix**).
- `oversized_read_batch_is_rejected` — a `GetObjects` batch over the configured limit is rejected with `InvalidArgument`.
- `successful_requests_do_not_feed_the_error_policy` — valid requests never trigger the error policy.
- `read_errors_do_not_feed_the_error_policy` — per-item read misses never trigger the error policy.

Gap found during #12130 (Phase 2 of the node JSON-RPC removal).

Full `iota-grpc-server` test suite green and clippy clean; the per-item, stream-error, and empty-batch tests verified to fail on the unpatched code.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Full node gRPC traffic control now accounts for batch APIs per item — each item of `ExecuteTransactions`, `SimulateTransactions`, `GetObjects`, and `GetTransactions` counts as one request for the spam policy, and per-item client errors feed the error policy — so batching can no longer dilute a client's request rate or hide its errors. `GetObjects` / `GetTransactions` batches are now size-capped via the new `max_get_objects_batch_size` / `max_get_transactions_batch_size` config (default 1000).
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] gRPC: Batch APIs are now tallied per item by the full node traffic control (spam rate and error policy), so batching cannot bypass rate limiting.

### DIFF
--- a/crates/iota-grpc-server/src/ledger_service/mod.rs
+++ b/crates/iota-grpc-server/src/ledger_service/mod.rs
@@ -110,9 +110,7 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
         // transport-level traffic control: charge one request per requested
         // item so batching cannot dilute the spam rate.
         if let Some(tally_handle) = &tally_handle {
-            for _ in 0..item_count {
-                tally_handle.tally_item(tonic::Code::Ok);
-            }
+            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
         }
         Ok(append_info_headers!(response, self.reader.clone()))
     }
@@ -136,9 +134,7 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
         // transport-level traffic control: charge one request per requested
         // item so batching cannot dilute the spam rate.
         if let Some(tally_handle) = &tally_handle {
-            for _ in 0..item_count {
-                tally_handle.tally_item(tonic::Code::Ok);
-            }
+            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
         }
         Ok(append_info_headers!(response, self.reader.clone()))
     }
--- a/crates/iota-grpc-server/src/ledger_service/mod.rs
+++ b/crates/iota-grpc-server/src/ledger_service/mod.rs
@@ -10,12 +10,13 @@ mod get_transactions;
 
 use std::sync::Arc;
 
+use futures::StreamExt;
 use iota_config::node::GrpcApiConfig;
 use iota_grpc_types::v1::ledger_service::{self as grpc_ledger_service};
 use iota_protocol_config::Chain;
 use iota_types::digests::ChainIdentifier;
 use tokio_util::sync::CancellationToken;
-use tonic::{Request, Response, Status};
+use tonic::{Code, Request, Response, Status};
 
 use crate::{traffic_control::TallyHandle, types::*};
 
@@ -103,15 +104,25 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
             .as_ref()
             .map_or(0, |batch| batch.requests.len());
         validate_read_batch_size(item_count, self.config.max_get_objects_batch_size)?;
-        let response = get_objects::get_objects(self.reader.clone(), request)
-            .map(|stream| Response::new(Box::pin(stream) as Self::GetObjectsStream))
-            .map_err(tonic::Status::from)?;
+        let stream =
+            get_objects::get_objects(self.reader.clone(), request).map_err(tonic::Status::from)?;
         // The batch's items are streamed lazily and invisible to the
-        // transport-level traffic control: charge one request per requested
-        // item so batching cannot dilute the spam rate.
+        // transport-level traffic control. Mark the request as accounted so the
+        // layer skips its default per-request tally, then charge one request per
+        // item as it is produced, so batching cannot dilute the spam rate. Reads
+        // feed the spam policy only: a client cannot know an object or
+        // transaction was pruned, so per-item read failures must not count
+        // towards the error policy.
         if let Some(tally_handle) = &tally_handle {
-            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
+            tally_handle.mark_accounted();
         }
+        let stream = stream.map(move |result| {
+            if let (Some(tally_handle), Ok(response)) = (&tally_handle, &result) {
+                tally_handle.tally_items(response.objects.iter().map(|_| Code::Ok));
+            }
+            result
+        });
+        let response = Response::new(Box::pin(stream) as Self::GetObjectsStream);
         Ok(append_info_headers!(response, self.reader.clone()))
     }
 
@@ -126,16 +137,26 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
             .as_ref()
             .map_or(0, |batch| batch.requests.len());
         validate_read_batch_size(item_count, self.config.max_get_transactions_batch_size)?;
-        let response =
+        let stream =
             get_transactions::get_transactions(self.reader.clone(), self.config.clone(), request)
-                .map(|stream| Response::new(Box::pin(stream) as Self::GetTransactionsStream))
                 .map_err(tonic::Status::from)?;
         // The batch's items are streamed lazily and invisible to the
-        // transport-level traffic control: charge one request per requested
-        // item so batching cannot dilute the spam rate.
+        // transport-level traffic control. Mark the request as accounted so the
+        // layer skips its default per-request tally, then charge one request per
+        // item as it is produced, so batching cannot dilute the spam rate. Reads
+        // feed the spam policy only: a client cannot know an object or
+        // transaction was pruned, so per-item read failures must not count
+        // towards the error policy.
         if let Some(tally_handle) = &tally_handle {
-            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
+            tally_handle.mark_accounted();
         }
+        let stream = stream.map(move |result| {
+            if let (Some(tally_handle), Ok(response)) = (&tally_handle, &result) {
+                tally_handle.tally_items(response.transaction_results.iter().map(|_| Code::Ok));
+            }
+            result
+        });
+        let response = Response::new(Box::pin(stream) as Self::GetTransactionsStream);
         Ok(append_info_headers!(response, self.reader.clone()))
     }
 
--- a/crates/iota-grpc-server/src/ledger_service/mod.rs
+++ b/crates/iota-grpc-server/src/ledger_service/mod.rs
@@ -10,7 +10,7 @@ mod get_transactions;
 
 use std::sync::Arc;
 
-use futures::StreamExt;
+use futures::{Stream, StreamExt};
 use iota_config::node::GrpcApiConfig;
 use iota_grpc_types::v1::ledger_service::{self as grpc_ledger_service};
 use iota_protocol_config::Chain;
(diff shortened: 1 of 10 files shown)


crates/iota-grpc-server/tests/traffic_control.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+/// Successful requests must not count towards the error policy or get
+/// blocked.
+#[tokio::test]
+async fn successful_requests_do_not_feed_the_error_policy() {
+    let policy_config = PolicyConfig {
+        connection_blocklist_ttl_sec: 120,
+        // Panics inside the traffic controller if any error is ever tallied.
+        error_policy_type: PolicyType::TestPanicOnInvocation,
```
bingyanglin: `TestPanicOnInvocation` panics inside the tally loop task, and a panic in a spawned task doesn't fail a `#[tokio::test]`, so this test still passes even if successful requests do feed the error policy. Maybe use `TestNConnIP(1)` instead?

crates/iota-grpc-server/tests/traffic_control.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests for the traffic-control layer of the gRPC server,
+//! in particular that handler errors — which tonic delivers in the response
+//! *trailers*, not the response headers — feed the error policy.
```
bingyanglin: Opposite? Tonic sends unary handler errors as a trailers-only response, so `grpc-status` ends up in the HTTP headers.

crates/iota-grpc-server/tests/common/mod.rs
```diff
@@ -512,3 +512,37 @@ pub async fn start_test_server(
 
     (server_handle, grpc_reader)
 }
+
+/// Like [`start_test_server`], but with the given traffic controller wired
+/// into the server's `TrafficControlLayer` and an optional transaction
+/// executor (required for the `TransactionExecutionService` to be
+/// registered).
+pub async fn start_test_server_with_traffic_controller(
```
bingyanglin, line 559: Very similar to `start_test_server`. I guess some logic can be shared so it can be a little bit cleaner?

crates/iota-grpc-server/tests/traffic_control.rs
muXxer: Switched the successful-requests test to a `TestNConnIP(1)` error policy. A wrongly error-tallied successful request now blocks the client and fails the test's `expect(...)`, instead of panicking in the detached tally-loop task where a `#[tokio::test]` cannot observe it.

crates/iota-grpc-server/tests/traffic_control.rs
muXxer: Corrected the module doc — right, tonic returns a unary handler error as a trailers-only response, so `grpc-status` lands in the HTTP headers where the layer reads it.

crates/iota-grpc-server/tests/common/mod.rs
muXxer: Extracted a shared `start_test_server_with(...)` that both `start_test_server` and `start_test_server_with_traffic_controller` delegate to.

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +83,48 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
+            .as_ref()
+            .map_or(0, |batch| batch.requests.len());
+        let response = get_objects::get_objects(self.reader.clone(), request)
             .map(|stream| Response::new(Box::pin(stream) as Self::GetObjectsStream))
             .map_err(tonic::Status::from)?;
+        // The batch's items are streamed lazily and invisible to the
+        // transport-level traffic control: charge one request per requested
+        // item so batching cannot dilute the spam rate.
+        if let Some(tally_handle) = &tally_handle {
+            for _ in 0..item_count {
```
bingyanglin: Nothing caps `item_count` here. `execute_transactions` calls `validate_batch_size` first, this doesn't. A big batch floods the log once the tally channel fills. Maybe add the same cap?

crates/iota-grpc-server/src/traffic_control.rs
```diff
@@ -4,8 +4,15 @@
...
 //!
-//! The layer extracts the client IP via and calls the traffic controller
-//! with a weight derived from the response's gRPC status code.
+//! The layer extracts the client IP and tallies each request against the
+//! traffic controller, deriving the error weight from the response's gRPC
+//! status code. A batch API's items are invisible at this level: their results
+//! ride in the response body — embedded in a unary response or produced lazily
+//! by a stream — while the layer only inspects the response status, so a batch
+//! would count as a single request regardless of how many items it carries.
+//! Handlers report each item through the [`TallyHandle`] request extension so
```
bingyanglin: This says items feed the error policy, but the ledger callers tally eagerly with a hardcoded `Ok`, so they only do spam accounting. Maybe spell out that eager callers are spam-only?

crates/iota-grpc-server/src/ledger_service/mod.rs
muXxer: Added `max_get_objects_batch_size` / `max_get_transactions_batch_size` config (default 50) and reject oversized batches, so the per-item tally loop is bounded and cannot flood the tally channel. Empty batches are still allowed. Covered by `oversized_read_batch_is_rejected`.

crates/iota-grpc-server/src/traffic_control.rs
muXxer: Updated the `TallyHandle` docs to spell out that the eager streaming-read callers pass `Code::Ok` and are therefore spam-only; only the unary execute/simulate callers feed the error policy.

bingyanglin (approved):

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +96,46 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
-        let response = get_objects::get_objects(self.reader.clone(), request.into_inner())
+        let tally_handle = request.extensions().get::<TallyHandle>().cloned();
+        let request = request.into_inner();
+        let item_count = request
+            .requests
+            .as_ref()
+            .map_or(0, |batch| batch.requests.len());
+        validate_read_batch_size(item_count, self.config.max_get_objects_batch_size)?;
+        let response = get_objects::get_objects(self.reader.clone(), request)
             .map(|stream| Response::new(Box::pin(stream) as Self::GetObjectsStream))
```
semenov-vladyslav:
```suggestion
            .map(|stream| {
                Response::new(Box::pin(stream.map(move |r| {
                    if let Some(tally_handle) = &tally_handle {
                        tally_handle.tally_item(
                            r.as_ref()
                                .err()
                                .map_or(tonic::Code::Ok, |status| status.code()),
                        );
                    }
                    r
                })) as Self::GetObjectsStream)
            })
```
Tallying client per all object requests at once seems weird, especially since the items are streamed lazily.

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +96,46 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
             .map_err(tonic::Status::from)?;
+        // The batch's items are streamed lazily and invisible to the
+        // transport-level traffic control: charge one request per requested
+        // item so batching cannot dilute the spam rate.
+        if let Some(tally_handle) = &tally_handle {
+            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
+        }
```
semenov-vladyslav:
```suggestion
```

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +96,46 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
+        let tally_handle = request.extensions().get::<TallyHandle>().cloned();
+        let request = request.into_inner();
+        let item_count = request
+            .requests
+            .as_ref()
+            .map_or(0, |batch| batch.requests.len());
+        validate_read_batch_size(item_count, self.config.max_get_transactions_batch_size)?;
+        let response =
+            get_transactions::get_transactions(self.reader.clone(), self.config.clone(), request)
+                .map(|stream| Response::new(Box::pin(stream) as Self::GetTransactionsStream))
```
semenov-vladyslav:
```suggestion
                .map(|stream| {
                    Response::new(Box::pin(stream.map(move |r| {
                        if let Some(tally_handle) = &tally_handle {
                            tally_handle.tally_item(
                                r.as_ref()
                                    .err()
                                    .map_or(tonic::Code::Ok, |status| status.code()),
                            );
                        }
                        r
                    })) as Self::GetTransactionsStream)
                })
```

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +96,46 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
+                .map_err(tonic::Status::from)?;
+        // The batch's items are streamed lazily and invisible to the
+        // transport-level traffic control: charge one request per requested
+        // item so batching cannot dilute the spam rate.
+        if let Some(tally_handle) = &tally_handle {
+            tally_handle.tally_items((0..item_count).map(|_| tonic::Code::Ok));
+        }
```
semenov-vladyslav:
```suggestion
```

semenov-vladyslav (approved):
Overall looks good. Tallying streaming requests could be done more precisely, if I am not wrong.

jkrvivian (approved):

Thoralf-M (approved):
Approving the .config/nextest.toml change

crates/iota-grpc-server/src/ledger_service/mod.rs
```diff
@@ -83,23 +97,66 @@ impl grpc_ledger_service::ledger_service_server::LedgerService for LedgerGrpcSer
...
+        // layer skips its default per-request tally, then charge one request per
+        // item as it is produced, so batching cannot dilute the spam rate. Reads
+        // feed the spam policy only: a client cannot know an object or
+        // transaction was pruned, so per-item read failures must not count
+        // towards the error policy.
+        if let Some(tally_handle) = &tally_handle {
+            tally_handle.mark_accounted();
+        }
+        let stream = stream.map(move |result| {
+            if let (Some(tally_handle), Ok(response)) = (&tally_handle, &result) {
```
semenov-vladyslav: Shouldn't Err result also be tallied? Otherwise a stream consisting of only Err results would not be tallied at all.

muXxer: Nice catch! fixed

PR #12133, opened 2026-07-03
chore: source fastcrypto from crates.io (=0.1.11)

Thoralf-M:
Source the main `fastcrypto` from crates.io, pinned to an exact `=0.1.11`, instead of a floating git revision.

- `-tbls`/`-vdf`/`-zkp` aren't on crates.io yet, so they're pinned by git revision (the `fastcrypto-v0.1.11` tag commit — a tag can be moved) and a `[patch]` unifies everything onto the crates.io release: one fastcrypto, no duplicate copy.
- Adds a CI guard that fails if the workspace ever resolves more than one fastcrypto, keeping this repo and the SDK (`iota-sdk-crypto`) in sync.
- Bumps the `iota-rust-sdk` git dependencies to the matching SDK commit (ed25519/secp256r1/secp256k1 moved onto fastcrypto).

**Behavior change:** fastcrypto 0.1.11 lowers the `poseidon_bn254` input limit from 32 to 16 — inputs above 16 are now rejected by the native. zklogin, the main consumer of large poseidon inputs, is disabled (and stays disabled), so on-chain impact is nil. Framework tests updated accordingly.

Companion SDK change: iotaledger/iota-rust-sdk#1248

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -402,9 +402,9 @@ iota-graphql-config = { path = "crates/iota-graphql-config" }
 iota-graphql-rpc = { path = "crates/iota-graphql-rpc" }
 iota-graphql-rpc-client = { path = "crates/iota-graphql-rpc-client" }
 iota-graphql-rpc-headers = { path = "crates/iota-graphql-rpc-headers" }
-iota-grpc-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", package = "iota-sdk-grpc-client", rev = "559083b9e901e53008f15d85d50df7f007c681ba" }
+iota-grpc-client = { git = "https://github.com/iotaledger/iota-rust-sdk.git", package = "iota-sdk-grpc-client", rev = "0135f905e0479728ef8c87cea324a882e50e1b5b" }
 iota-grpc-server = { path = "crates/iota-grpc-server" }
-iota-grpc-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", package = "iota-sdk-grpc-types", rev = "559083b9e901e53008f15d85d50df7f007c681ba" }
+iota-grpc-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", package = "iota-sdk-grpc-types", rev = "0135f905e0479728ef8c87cea324a882e50e1b5b" }
 iota-http = { path = "crates/iota-http" }
 iota-indexer = { path = "crates/iota-indexer" }
 iota-indexer-builder = { path = "crates/iota-indexer-builder" }
@@ -443,9 +443,9 @@ iota-replay = { path = "crates/iota-replay" }
 iota-rest-kv = { path = "crates/iota-rest-kv" }
 iota-rpc-loadgen = { path = "crates/iota-rpc-loadgen" }
 iota-sdk = { path = "crates/iota-sdk" }
-iota-sdk-crypto = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "559083b9e901e53008f15d85d50df7f007c681ba", default-features = false }
-iota-sdk-transaction-builder = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "559083b9e901e53008f15d85d50df7f007c681ba", default-features = false }
-iota-sdk-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "559083b9e901e53008f15d85d50df7f007c681ba", default-features = false }
+iota-sdk-crypto = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "0135f905e0479728ef8c87cea324a882e50e1b5b", default-features = false }
+iota-sdk-transaction-builder = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "0135f905e0479728ef8c87cea324a882e50e1b5b", default-features = false }
+iota-sdk-types = { git = "https://github.com/iotaledger/iota-rust-sdk.git", rev = "0135f905e0479728ef8c87cea324a882e50e1b5b", default-features = false }
 iota-simulator = { path = "crates/iota-simulator" }
 iota-snapshot = { path = "crates/iota-snapshot" }
 iota-source-validation = { path = "crates/iota-source-validation" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -273,10 +273,11 @@ either = "1.8.0"
 enum_dispatch = "^0.3"
 expect-test = "1.4.0"
 eyre = "0.6.8"
-fastcrypto = "0.1.10"
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", tag = "fastcrypto-v0.1.10" }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -273,10 +273,11 @@ either = "1.8.0"
 enum_dispatch = "^0.3"
 expect-test = "1.4.0"
 eyre = "0.6.8"
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9", features = ["experimental"] }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9", package = "fastcrypto-zkp" }
+fastcrypto = "=0.1.11"
+# rev "47f8f6b..." is the fastcrypto-v0.1.11 tag, pinned by revision because a tag can be moved.
```
thibault-martinez: Don't put the actual rev in the comment, annoying to update

Thoralf-M: Done — dropped the rev from the comment; it now just refers to the `fastcrypto-v0.1.11` tag commit.

---

scripts/check_fastcrypto_single_version.sh
thibault-martinez, line 1: Is this really necessary? Would it not fail to compile?

thibault-martinez (approved):

Cargo.toml
```diff
@@ -273,10 +273,11 @@ either = "1.8.0"
 enum_dispatch = "^0.3"
 expect-test = "1.4.0"
 eyre = "0.6.8"
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9", features = ["experimental"] }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9", package = "fastcrypto-zkp" }
+fastcrypto = "=0.1.11"
+# Pinned by revision (the fastcrypto-v0.1.11 tag commit) because a tag can be moved.
```
thibault-martinez: I would go as far as not even mentioning the exact version because I feel like we would forget to update it but just resolve if you want to keep it

Thoralf-M: Removed the version from the comment entirely — it now just says pinned by revision, keep in sync with the fastcrypto version above.

---

scripts/check_fastcrypto_single_version.sh
Thoralf-M: It wouldn't reliably fail to compile. Cargo permits multiple semver-incompatible copies of a crate to coexist, so two fastcrypto versions only break the build where incompatible types actually cross the node↔SDK (`iota-sdk-crypto`) boundary — otherwise you silently ship two copies (duplicate code, separate global state). The guard catches that silent case.

The second check can't be caught by compilation at all: it verifies (via the crate's `.cargo_vcs_info.json`) that the crates.io `fastcrypto` we patch to was published from the same commit the git-pinned `-tbls`/`-vdf`/`-zkp` reference. A matching version number can otherwise resolve from a different source and compile fine.

Happy to drop it if you'd still prefer — let me know.

---

Thoralf-M: Vlad asked for the script to be extra safe since it's easy to diverge and get problems this way

muXxer (approved):

crates/iota-framework/packages/iota-framework/tests/crypto/poseidon_tests.move
```diff
@@ -40,45 +40,14 @@ fun test_poseidon_bn254_hash() {
     let expected = 4203130618016961831408770638653325366880478848856764494148034853759773445968u256;
     let actual = poseidon_bn254(&msg);
     assert!(actual == expected);
+}
 
+#[test]
+// poseidon_bn254 accepts at most 16 inputs; more are rejected by the native and
+// surface as ENonCanonicalInput.
+#[expected_failure(abort_code = iota::poseidon::ENonCanonicalInput)]
```
thibault-martinez:
```suggestion
// poseidon_bn254 accepts at most 16 inputs; more are rejected by the native and
// surface as ENonCanonicalInput.
#[test]
#[expected_failure(abort_code = iota::poseidon::ENonCanonicalInput)]
```

crates/iota-framework/packages/iota-framework/tests/crypto/poseidon_tests.move
```diff
@@ -40,45 +40,14 @@ fun test_poseidon_bn254_hash() {
...
-        24u256,
-        25u256,
-        26u256,
-        27u256,
-        28u256,
-        29u256,
-    ];
-    let expected = 4123755143677678663754455867798672266093104048057302051129414708339780424023u256;
-    let actual = poseidon_bn254(&msg);
-    assert!(actual == expected);
```
thibault-martinez, line 78: Why are we removing checks?

thibault-martinez: I mean, I see the comment but what happened?

Thoralf-M: The two removed cases hashed **30 inputs** (`0..=29`, expected `4123755…`) and **32 inputs** (expected `15368023…`). fastcrypto 0.1.11 lowers the `poseidon_bn254` cap from 32 to 16, so those inputs are now rejected by the native (abort `ENonCanonicalInput`) — their exact-hash assertions can no longer run. Rather than delete coverage, they were replaced by `test_poseidon_bn254_too_many_inputs`, which asserts that >16 inputs abort.

All the ≤16-input hash checks are kept (1, 2, and 15 inputs), plus the canonical/non-canonical/empty-input checks — so nothing within the supported range lost coverage.

If you'd rather preserve the 32-input behavior instead of accepting the cap, that's possible but needs an upstream fastcrypto addition (a bytes-level `poseidon_merkle_tree` wrapper — the byte↔field helpers are private); happy to go that route if you prefer.

---

Thoralf-M: `fastcrypto 0.1.11 lowers the poseidon_bn254 input limit from 32 to 16 — inputs above 16 are now rejected by the native`

thibault-martinez: Alright

scripts/check_fastcrypto_single_version.sh
thibault-martinez: Alright

miker83z (approved):

PR #12137, opened 2026-07-03
feat(config+node): per-service concurrency limits for the validator gRPC server

muXxer:
# Description of change

## Problem

Two related issues with admission control on the validator gRPC server:

1. **The default concurrency limit is effectively unlimited.** `grpc_concurrency_limit` defaults to `20000000000`, so a flood of incoming requests can grow queues and memory without bound. Runtime isolation (#12046) prevents a flood from starving the node core of worker threads, but without a finite admission bound the failure mode just moves from starvation to OOM.

2. **The limit is global per listener, not per service.** The single `GlobalConcurrencyLimitLayer` wraps the whole server, which hosts three services on one listener: `Validator` and `ValidatorV2` (client transaction submission / certificate processing — the flood source) and `ValidatorPeer` (read-only `GetCheckpoint` plus rare control-plane RPCs between validators). Once the limit is finite, a client flood that exhausts it also crowds validator-peer RPCs out of admission slots.

## Approach

- **`iota-network-stack`**: new `ServiceConcurrencyLimit` middleware (`concurrency.rs`) — a semaphore-based concurrency bound around a single gRPC service that preserves `NamedService` so tonic routing still works, plus `ServerBuilder::add_service_with_concurrency_limit`. With `load_shed` enabled, requests over the limit are rejected immediately with `RESOURCE_EXHAUSTED` (previously, global load shedding surfaced as a transport-level error rather than a proper gRPC status); otherwise they wait for a slot, matching `tower::limit::ConcurrencyLimit` permit semantics. Covered by unit tests (shed over limit, queue without shed, independence of limits).

- **`iota-config`**: `grpc_concurrency_limit` (absolute) is replaced by **`grpc_concurrency_limit_per_core`** (default: 1000). The effective per-service limit is the configured value multiplied by the machine's CPU cores at server startup, so the same config file scales with the hardware and never bakes in a machine-specific number. The per-core default is generous by design (most request time is spent awaiting locks, I/O or consensus, not on-CPU): it bounds worst-case in-flight work, it does not throttle normal load. Operators wanting hard load-shedding can lower it and set `grpc_load_shed`. The `DEFAULT_GRPC_CONCURRENCY_LIMIT` (20e9) constant is removed entirely; test/swarm validator configs set an effectively-unlimited per-core value instead so tests and benchmarks are never throttled.

- **`iota-node`**: the validator gRPC server no longer sets `global_concurrency_limit` / `load_shed` on the server config; instead each of the three services is added with its own limit computed from `grpc_concurrency_limit_per_core` x CPU cores (clamped to tokio's `Semaphore::MAX_PERMITS`) and `grpc_load_shed`.

## Behavioral / config notes

- The default changes from effectively unlimited (`20000000000` global) to `1000 x cores` **per service**. Worst-case total in-flight across the three services is 3x the per-service limit.
- **The old `grpc_concurrency_limit` config field is removed, not aliased.** An absolute value silently reinterpreted as per-core would be a large jump, so configs still setting the old key have it ignored and get the new default; operators who tuned it must set `grpc_concurrency_limit_per_core` instead.
- With `grpc_load_shed` enabled, rejected requests now get a clean gRPC `RESOURCE_EXHAUSTED` status per service instead of a connection-level error from the global load-shed layer.
- `grpc_concurrency_limit_per_core` is `NonZeroUsize`: a config setting it to `0` is rejected at config load, since zero would not disable the limit, it would block (or shed) every request.
- Swarm/test config snapshots change accordingly (renamed field, explicit unlimited value for validators).

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): The validator gRPC server now enforces a concurrency limit per service (Validator, ValidatorV2, ValidatorPeer) instead of one global limit, so a flood of client transaction submissions cannot crowd out validator-peer RPCs. The `grpc_concurrency_limit` config field is replaced by `grpc_concurrency_limit_per_core` (default 1000); the effective per-service limit is that value times the machine's CPU cores (previously effectively unlimited). With `grpc_load_shed`, excess requests are rejected with `RESOURCE_EXHAUSTED`.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-network-stack/src/concurrency.rs
+++ b/crates/iota-network-stack/src/concurrency.rs
@@ -7,9 +7,17 @@
 //! server, [`ServiceConcurrencyLimit`] bounds the in-flight requests of a
 //! single gRPC service, so services sharing one listener cannot crowd each
 //! other out of admission slots.
+//!
+//! Tower's per-service [`tower::limit::ConcurrencyLimitLayer`] cannot be used
+//! here: its `ConcurrencyLimit` wrapper does not implement tonic's
+//! [`NamedService`], which `Routes::add_service` requires for routing, and
+//! shedding through tower's `LoadShed` surfaces as a `BoxError`, incompatible
+//! with the router's `Error = Infallible` bound — over-limit requests must be
+//! answered in-band with a gRPC `RESOURCE_EXHAUSTED` response instead.
 
 use std::{
     convert::Infallible,
+    num::NonZeroUsize,
     sync::Arc,
     task::{Context, Poll},
 };
@@ -37,13 +45,13 @@ pub struct ServiceConcurrencyLimit<S> {
 }
 
 impl<S> ServiceConcurrencyLimit<S> {
-    pub fn new(inner: S, limit: usize, load_shed: bool) -> Self {
+    pub fn new(inner: S, limit: NonZeroUsize, load_shed: bool) -> Self {
         Self {
             inner,
             // Clamp: `Semaphore::new` panics above `MAX_PERMITS`, and
             // effectively-unlimited configs multiply large values by the CPU
             // core count.
-            semaphore: Arc::new(Semaphore::new(limit.min(Semaphore::MAX_PERMITS))),
+            semaphore: Arc::new(Semaphore::new(limit.get().min(Semaphore::MAX_PERMITS))),
             load_shed,
         }
     }
@@ -66,6 +74,15 @@ where
     }
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
+        // Admission happens here rather than in `poll_ready` (where
+        // `tower::limit::ConcurrencyLimit` acquires its permit): a shed
+        // request must be answered with a gRPC response, and `poll_ready`
+        // can only signal Ready or Pending — converting Pending into an
+        // error via an outer load-shed layer is ruled out by the router's
+        // `Error = Infallible` bound. Readiness-based acquisition would
+        // also buy no upstream backpressure: the axum router dispatches
+        // every request on a fresh clone of this service.
+        //
         // Take the instance that was driven to readiness and leave the clone
         // for later calls, as `poll_ready` readiness does not transfer to
         // clones.
@@ -146,7 +163,7 @@ mod tests {
             BlockingService {
                 release: release.clone(),
             },
-            1,
+            NonZeroUsize::MIN,
             true,
         );
 
@@ -171,7 +188,7 @@ mod tests {
             BlockingService {
                 release: release.clone(),
             },
-            1,
+            NonZeroUsize::MIN,
             false,
         );
 
@@ -194,8 +211,8 @@ mod tests {
         let blocking = BlockingService {
             release: release.clone(),
         };
-        let saturated = ServiceConcurrencyLimit::new(blocking.clone(), 1, true);
-        let other = ServiceConcurrencyLimit::new(blocking, 1, true);
+        let saturated = ServiceConcurrencyLimit::new(blocking.clone(), NonZeroUsize::MIN, true);
+        let other = ServiceConcurrencyLimit::new(blocking, NonZeroUsize::MIN, true);
 
         let in_flight = tokio::spawn(saturated.clone().oneshot(request()));
         tokio::task::yield_now().await;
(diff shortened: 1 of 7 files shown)


crates/iota-network-stack/src/concurrency.rs
```diff
@@ -0,0 +1,227 @@
...
+/// Clones share the same limit.
+#[derive(Clone)]
+pub struct ServiceConcurrencyLimit<S> {
+    inner: S,
+    semaphore: Arc<Semaphore>,
+    load_shed: bool,
+}
+
+impl<S> ServiceConcurrencyLimit<S> {
+    pub fn new(inner: S, limit: usize, load_shed: bool) -> Self {
```
semenov-vladyslav: Should `limit` be `NonZeroUsize`?

crates/iota-network-stack/src/concurrency.rs
```diff
@@ -0,0 +1,227 @@
...
+        let clone = self.inner.clone();
+        let mut inner = std::mem::replace(&mut self.inner, clone);
+        let semaphore = self.semaphore.clone();
+        let load_shed = self.load_shed;
+
+        Box::pin(async move {
+            // The permit is held until the response future resolves, mirroring
+            // `tower::limit::ConcurrencyLimit`.
+            let _permit = if load_shed {
+                match semaphore.try_acquire_owned() {
```
semenov-vladyslav, line 98: Why is semaphore acquired in `call` but not in `poll_ready`? `poll_ready` should report when `call` cannot be dispatched immediately, this is exactly the case with the semaphore.

crates/iota-network-stack/src/concurrency.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-service concurrency limiting for gRPC servers.
+//!
+//! Unlike [`tower::limit::GlobalConcurrencyLimitLayer`] applied around a whole
```
semenov-vladyslav, line 6: What about `ConcurrencyLimitLayer`?

semenov-vladyslav (commented):
Why not re-use `tower::limit::ConcurrencyLimitLayer`?

piotrm50 (approved):

crates/iota-network-stack/src/concurrency.rs
muXxer: Agreed — changed to `NonZeroUsize` end-to-end (config field, builder API, middleware). A zero limit would not disable limiting, it would block or shed every request, so serde now rejects it at config load with a clear error. Lands with the next push.

crates/iota-network-stack/src/concurrency.rs
muXxer: Deliberate, for two reasons (now also documented in the code):

1. The shed path must answer with a gRPC `RESOURCE_EXHAUSTED` **response**. `poll_ready` can only signal Ready/Pending — the tower-idiomatic pattern (outer `LoadShed` converting Pending into an error) produces `BoxError`, which the router's `Error = Infallible` bound (tonic `Routes::add_service`) rules out. So the admission decision has to be made where a response can be produced, i.e. in `call`.
2. Readiness-based acquisition wouldn't propagate backpressure here anyway: the axum router is always-ready and drives each request through a fresh clone of the route service (`oneshot`), so there is no shared readiness state upstream could observe. An over-limit request waits inside its own response future either way, at the same point in the pipeline.

It also avoids the poll_ready-acquired-permit subtlety (readiness polled, `call` never made) that `tower::limit::ConcurrencyLimit` needs drop-handling for.

crates/iota-network-stack/src/concurrency.rs
muXxer: It doesn't fit for two reasons (added to the module doc now): tower's `ConcurrencyLimit<S>` doesn't implement tonic's `NamedService` (tower has no tonic dependency), which `Routes::add_service` requires for routing — so a name-forwarding wrapper is needed regardless. And it has no shedding: over-limit requests only queue in `poll_ready`; adding tower's `LoadShed` on top surfaces `BoxError`, incompatible with the router's `Error = Infallible` bound. This type is the minimal thing satisfying both constraints.

semenov-vladyslav (approved):

PR #12140, opened 2026-07-03
fix(iota-core): expose `ValidatorClientMetrics` for `TransactionDriver:new` API

roman1e2f5p8s:
# Description of change

`TransactionDriver::new` is public, but its `client_metrics: Arc<ValidatorClientMetrics>` parameter names a type that became unreachable outside the crate when #11007 changed the `validator_client_monitor` module to `pub(crate) mod`. The `ValidatorClientMetrics` type is declared `pub`, but its only path ran through the now crate-private module, so external callers (e.g., clients implementing their own transaction drivers) could not name the argument to call the `TransactionDriver::new` constructor.

This PR re-exports `ValidatorClientMetrics` from `transaction_driver`, giving it a public path (`iota_core::transaction_driver::ValidatorClientMetrics`) co-located with the constructor that requires it. This mirrors the existing `quorum_driver` pattern, where `QuorumDriverMetrics` is re-exported publicly next to `QuorumDriverHandlerBuilder::new` while the module internals stay private.

The genuinely internal types (`ValidatorClientMonitor`, `OperationType`, `OperationFeedback`) in the `validator_client_monitor` module stay crate-private. 

### DIFF
--- a/crates/iota-core/src/transaction_driver/mod.rs
+++ b/crates/iota-core/src/transaction_driver/mod.rs
@@ -31,11 +31,7 @@ use tokio::{
 use tracing::instrument;
 use transaction_submitter::*;
 
-// `validator_client_monitor` is crate-internal, but `ValidatorClientMetrics`
-// appears in the public `TransactionDriver::new` signature. Re-export it here so
-// external callers (e.g., clients implementing custom transaction drivers) have
-// a nameable path to the type; without it the public constructor would take an
-// unnameable argument.
+// `ValidatorClientMetrics` is used in public `TransactionDriver::new`
 pub use crate::validator_client_monitor::ValidatorClientMetrics;
 use crate::{
     authority_aggregator::AuthorityAggregator,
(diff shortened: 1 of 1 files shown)


crates/iota-core/src/transaction_driver/mod.rs
```diff
@@ -31,12 +31,16 @@ use tokio::{
 use tracing::instrument;
 use transaction_submitter::*;
 
+// `validator_client_monitor` is crate-internal, but `ValidatorClientMetrics`
+// appears in the public `TransactionDriver::new` signature. Re-export it here so
+// external callers (e.g., clients implementing custom transaction drivers) have
+// a nameable path to the type; without it the public constructor would take an
+// unnameable argument.
```
muXxer: I wonder if the explanation really needs to be that long. We have the same situation in several places of the code. I guess a `ValidatorClientMetrics is used in public TransactionDriver::new` is sufficient.

muXxer (approved):

crates/iota-core/src/transaction_driver/mod.rs
roman1e2f5p8s: shortened in https://github.com/iotaledger/iota/pull/12140/commits/06114ccc3868ee76bf4662a76472fe800ddcbec2

piotrm50 (approved):

PR #12144, opened 2026-07-03
feat(core): live/historic storage split with per-epoch history buckets

muXxer:
# Description of change

Splits node storage into a hot live set and per-epoch historic buckets, making pruning a constant-time bucket drop instead of per-key deletes plus compaction churn, and letting RPC fullnodes serve long history (target: ~100 epochs) without the hot tables paying for it.

When the `historic-store` pruning config is enabled (fullnode-only, off by default), the pruner **relocates instead of deletes**:

- **Superseded object versions** move out of the perpetual `objects` table into per-epoch column families of a new always-open `history` RocksDB, bucketed by supersession epoch. The live table becomes heads-only (including `Deleted`/`Wrapped` tombstone heads, which every latest-version read depends on); tombstone heads are recorded in a per-bucket expiry list and point-deleted from the live table only when the bucket expires.
- **Checkpoint-keyed history** (transactions, effects, executed-effects links, events, checkpoint contents and certified summaries) moves into the same epoch buckets, bucketed by the epoch of their checkpoint.

Expiring an epoch of history is a `drop_cf` per family — RocksDB unlinks the column-family SSTs outright, with no tombstones and no compaction pass. Crash consistency follows one contract everywhere: write history → flush (durability barrier) → delete source rows + advance watermark in one atomic batch; replay from the watermark is idempotent.

Reads: consensus and execution paths never touch history — a live-table miss there remains a loud failure. Old data is served through fallbacks that trigger only after a live miss: gRPC exact-version object lookups (`GrpcReadStore`), and gRPC + state-sync reads of old transactions/effects/events/checkpoints (`RocksDbStore`), with the advertised `lowest_available_checkpoint` extending back to the earliest relocated checkpoint. DB checkpoints include the history DB (snapshotted last — source-first ordering closes the torn-window between the two stores), so restored nodes keep serving their history; formal snapshots intentionally remain live-set-only.

Supporting changes: typed-store gains runtime `Database::create_cf` (with `cf_names` kept in sync for `flush_all`); the compaction-filter pruner is mutually exclusive with the historic store (enforced before the perpetual DB opens); nodes that previously ran with objects pruning disabled get a one-time watermark fast-forward past already-pruned checkpoint data.

Layout rationale: one DB with per-epoch column families (rather than one DB per epoch) shares a single block cache and file-descriptor budget, and RocksDB's native per-SST bloom filters answer the epoch-routing problem — exact-key lookups probe buckets newest-to-oldest, where a miss in a sealed, compacted bucket costs ~µs from in-memory filters.

Follow-ups tracked in `plans/` on this branch: e2e simtests with the feature enabled and the benchmark matrix (which gate enabling this on production fullnodes), plus a behavior-preserving `PrunerContext` refactor of the pruner's parameter threading.

-

Details: `cargo ci-clippy` (workspace, all targets, `-D warnings`), `cargo +nightly fmt`, and `IOTA_SKIP_SIMTESTS=1 cargo nextest run -p typed-store -p iota-core --lib` (751 passed) run locally. ~25 new tests cover: heads-only invariant with tombstones as live heads, supersession-epoch bucketing, crash-replay idempotence for both object and checkpoint relocation, retention expiry with unwrap-resurrected lineages, live-object-set invariance under relocation and expiry, legacy V1-row migration at read, bucket rediscovery/seal state across restarts, availability-watermark tracking, DB-checkpoint copies serving all families, and the source-first snapshot-ordering torn-window test. Not yet run: e2e simtests with the feature enabled and performance benchmarks (see `plans/historic-store-testing.md`); the feature is off by default, so existing e2e coverage exercises the unchanged path.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Adds an opt-in `historic-store` pruning mode for RPC fullnodes: the pruner relocates superseded object versions and checkpoint-keyed history into per-epoch buckets instead of deleting them, keeping the hot tables small while serving configurable epochs of history; expired epochs are dropped in constant time. Off by default; validators are unaffected (the setting is ignored with a warning). Incompatible with `enable-compaction-filter`. DB checkpoints of nodes with the feature enabled include the history DB.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] gRPC: On fullnodes with the historic store enabled, exact-version `GetObjects` and transaction/effects/events/checkpoint lookups keep working for relocated (pruned) data, and the advertised lowest available checkpoint extends back to the earliest retained history instead of the pruning watermark.
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/iota-config/data/fullnode-template-with-path.yaml
+++ b/crates/iota-config/data/fullnode-template-with-path.yaml
@@ -19,7 +19,9 @@ migration-tx-data-path: "migration.blob"
 
 authority-store-pruning-config:
   num-latest-epoch-dbs-to-retain: 3
-  num-epochs-to-retain: 1
+  # Number of epochs of superseded object versions and pruned checkpoint
+  # data to keep readable; whole epoch buckets are dropped past this window.
+  historic-epochs-to-retain: 2
 
 authority-key-pair:
   path: "authority.key"
--- a/crates/iota-config/data/fullnode-template.yaml
+++ b/crates/iota-config/data/fullnode-template.yaml
@@ -19,4 +19,6 @@ migration-tx-data-path: "/opt/iota/config/migration.blob"
 
 authority-store-pruning-config:
   num-latest-epoch-dbs-to-retain: 3
-  num-epochs-to-retain: 1
+  # Number of epochs of superseded object versions and pruned checkpoint
+  # data to keep readable; whole epoch buckets are dropped past this window.
+  historic-epochs-to-retain: 2
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1065,12 +1065,12 @@ pub struct AuthorityStorePruningConfig {
     /// once out of retention. Fullnode-only; incompatible with
     /// `enable_compaction_filter`.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub historic_object_store: Option<HistoricObjectStoreConfig>,
+    pub historic_store: Option<HistoricStoreConfig>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
-pub struct HistoricObjectStoreConfig {
+pub struct HistoricStoreConfig {
     /// Number of epochs of superseded object versions to retain, bucketed by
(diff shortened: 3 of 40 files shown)


muXxer:
super-seeded by #12695, which uses the new `TaggedDBMap`

PR #12145, opened 2026-07-03
upstream(iota-core): pipeline BCS deserialization in consensus commit handler

piotrm50:
# Description of change

- Port of upstream PR: [MystenLabs/sui#27052](https://github.com/MystenLabs/sui/pull/27052)
- Description: Splits Starfish consensus output processing into two stages connected by a small bounded channel (capacity 2): a dedicated deserialize worker BCS-parses each commit's transactions while the handler processes the previous one. A single worker per stage plus a FIFO channel preserve commit ordering by construction, and the small capacity applies backpressure when the handler is the bottleneck instead of buffering parsed commits unboundedly.

Adapted to the fork:

- Ported only the deserialization-pipelining half. `handle_consensus_output` now takes the pre-parsed transactions as a `ConsensusOutputTransactions` parameter; a `#[cfg(test)]` `handle_consensus_output_for_test` wrapper performs the inline deserialization so the existing single-arg unit-test callsites are unchanged. `StarfishConsensusHandler::new` spawns the deserialize worker and the handler via a `tokio::task::JoinSet` (replacing the single `JoinHandle`, with `abort_all` on drop), and the bounded channel uses `monitored_mpsc::channel("consensus_deserialized_commits", 2)`.
- Dropped the upstream `finalized_transactions_cache` / `is_recently_finalized` replacement: it serves the FPC voting path, which IOTA does not have (there is no `is_recently_finalized` reader), so that piece is N/A here.

### DIFF
--- a/crates/iota-core/src/consensus_handler.rs
+++ b/crates/iota-core/src/consensus_handler.rs
@@ -12,7 +12,7 @@ use std::{
 use arc_swap::ArcSwap;
 use iota_common::random_util::randomize_cache_capacity_in_tests;
 use iota_macros::{fail_point, fail_point_if};
-use iota_metrics::{monitored_mpsc::UnboundedReceiver, monitored_scope, spawn_monitored_task};
+use iota_metrics::{monitored_mpsc, monitored_scope, spawn_monitored_task};
 use iota_sdk_types::CancelledTransaction;
 use iota_types::{
     base_types::{AuthorityName, TransactionDigest},
@@ -25,6 +25,7 @@ use iota_types::{
 use lru::LruCache;
 use serde::{Deserialize, Serialize};
 use starfish_config::Committee as ConsensusCommittee;
+use tokio::task::JoinSet;
 use tracing::{debug, info, instrument, trace_span, warn};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


polinikita (approved):

crates/iota-core/src/consensus_handler.rs
```diff
@@ -483,53 +505,97 @@ impl AsyncTransactionScheduler {
...
+        // is the bottleneck. Single-threaded; ordering is preserved (one
+        // worker, FIFO channel, one handler).
+        let (parsed_sender, mut parsed_receiver) = monitored_mpsc::channel(
+            "consensus_deserialized_commits",
+            CONSENSUS_HANDLER_DESERIALIZE_CHANNEL_CAPACITY,
+        );
+        tasks.spawn(async move {
+            while let Some(consensus_output) = receiver.recv().await {
+                let block_transactions: ConsensusOutputTransactions = {
+                    let _scope = monitored_scope("StarfishConsensusHandler::deserialize_worker");
```
bingyanglin, line 553: During startup replay the deserialize worker still BCS-parses commits at or below `last_processed_commit_at_startup`, but stage 2 just hands those to `handle_prior_consensus_output` and drops the parsed result. Maybe skip `.transactions()` for those indices since the worker already knows the startup watermark.

piotrm50: Currently commits below `last_processed_commit_at_startup` are not even emitted because this parameter is set to 0. This will only be changed once we start doing something useful with those early commits rather than just skipping them, so stage 1 parsing will need to be re-evaluated. Currently no work is wasted because the early commits are not emitted.

bingyanglin (approved):

PR #12146, opened 2026-07-03
feat(core): replace `full_checkpoint_content` table with a size-bounded in-memory cache

muXxer:
## Description

Every checkpoint synced via p2p state sync had its full contents (transactions + effects + signatures) written to disk twice: once into the perpetual store (transactions/effects) and once as a serialized blob into the `full_checkpoint_content` RocksDB table — only for the blob to be deleted again ~5,000 checkpoints later by the checkpoint executor. This is pure write amplification plus tombstone/compaction churn for data that lives minutes to hours.

The table was purely a read optimization: every consumer already has a complete fallback that reconstructs `FullCheckpointContents` from the permanent `checkpoint_content` table plus the transaction and effects stores. Validators never populated the table at all (they don't sync contents via p2p), so the reconstruction path has always been the production serving path on validators.

This PR replaces the table with a size-bounded in-memory cache:

- **New `FullCheckpointContentsCache`** (`iota-core/src/checkpoints/full_checkpoint_contents_cache.rs`): keyed by sequence number with a contents-digest index, byte-accounted via `bcs::serialized_size`, evicting lowest sequence numbers first when over budget (keeps the newest window — what tip-following peers request). It serves the checkpoint executor's bulk transaction loads, peer requests via state sync, and the state-sync re-check.
- **New config knob** `full-checkpoint-contents-cache-size-mb` in `NodeConfig` (default **1024 MiB**, `0` disables the cache; consumers then always use the reconstruction fallback).
- **Dropped column families**: `full_checkpoint_content` and `checkpoint_sequence_by_contents_digest` are marked `#[deprecated_db_map]` and auto-dropped on the first open after upgrade. No migration needed — the data was a short-lived cache and is reconstructable.
- **Removed** the hardcoded `NUM_SAVED_FULL_CHECKPOINT_CONTENTS = 5_000` pruning block in the checkpoint executor and the digest-mapping delete in the authority store pruner.
- **Validators populate the cache from the checkpoint executor**: nodes that execute without syncing contents via state sync assemble transactions and effects when loading a checkpoint anyway; the executor now caches the assembled `FullCheckpointContents` (memory-only, nothing written to disk), so a validator serves the freshest checkpoints to its SSFNs straight from memory instead of reconstructing them — speeding up checkpoint propagation and end-to-end confirmation.
- **No wasted work when the cache wouldn't keep the entry**: the executor's contents assembly (clones every transaction/effect + a BCS size pass) is gated on a `should_cache()` pre-check — skipped when the cache is disabled (budget 0) and during deep catch-up, where the cache window rides the state-sync frontier far ahead of the executor and lowest-seq eviction would drop the entry immediately. `insert()` itself also rejects entries older than the window of a full cache.
- **New metrics**: `full_checkpoint_contents_cache_{hits,misses,evictions,entries,bytes}`. Hits/misses carry a `lookup` label: `by_seq` (executor bulk loads and sequence-number reads) vs `by_digest` (peer serving). Note that on validators, `by_seq` misses accrue once per checkpoint by design — the executor checks the cache before inserting the entry it then produces — so `by_digest` is the serving-health signal there.

### Behavior notes

- After a restart the cache is empty, so the executor's catch-up window and the first peer requests use the per-item / reconstruction fallback until the cache refills (and re-populates it as checkpoints execute). Transactions and effects are always durably written before the cache insert, so no data can be lost.
- During deep catch-up, if execution lags sync by more than the cache budget, the executor degrades to per-item loads of data state sync just wrote (hot in RocksDB caches). Hit/miss metrics make this observable; raising the knob is the mitigation.
- The peer-facing `lowest_available_checkpoint` watermark is governed by the epoch-based pruner and is unaffected.

### Ops notes

- The two CFs are dropped automatically on first start after the upgrade (one-time disk/compaction win, no operator action). Downgrade is safe: older binaries recreate the CFs empty, which is the normal validator state.
- Per-CF RocksDB `DBMetrics` for the two dropped tables disappear; new cache metrics replace them.
- The budget is accounted in serialized (BCS) bytes, so a full cache's resident memory runs somewhat above the configured value (default 1024 MiB) due to in-memory representation overhead. Memory-constrained deployments can lower the knob or set it to `0` — with the cache disabled, all cache-related work (including the executor's contents assembly) is skipped, not just the storage.

## Test plan

- New unit tests for the cache (insert/lookup by seq and digest, byte accounting, lowest-seq eviction, oversized entries, re-insert accounting, zero-budget disable).
- New `CheckpointStore` test: insert populates cache + durable `checkpoint_content` row; reopening yields a cache miss while the digest-form contents survive for reconstruction.
- New checkpoint-executor tests pinning the produce-path behavior: the fallback load populates the contents cache (retrievable by sequence number and contents digest), skips it when the cache is disabled, and skips it when the cache window rides the state-sync frontier ahead of the executing sequence (deep catch-up).
- `cargo nextest run -p iota-core --lib` (689 passed), `-p typed-store -p iota-network -p iota-archival --lib` (109 passed, incl. state-sync + archive-sync tests).
- `cargo simtest -p iota-e2e-tests`: `basic_checkpoints_integration_test`, `test_checkpoint_timestamps_non_decreasing`, `test_full_node_cold_sync`, `test_full_node_sync_flood` — all pass.
- `cargo ci-clippy` clean; full workspace `--all-targets` compiles.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Replace the on-disk `full_checkpoint_content` table with a size-bounded in-memory cache for serving checkpoint contents to state-sync peers; validators populate it from the checkpoint builder. Adds an optional node config `full-checkpoint-contents-cache-size-mb` (default `1024`; set `0` to disable). The `full_checkpoint_content` and `checkpoint_sequence_by_contents_digest` RocksDB column families are dropped automatically on first start.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

#### Breaking Changes Rollout

No breaking changes. The new config parameter is additive with a default, and the dropped column families are removed automatically on first start (downgrade-safe: older binaries recreate them empty).

### DIFF
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -724,7 +724,6 @@ impl CheckpointExecutor {
             .get_full_checkpoint_contents_by_sequence_number(seq)
             .tap_some(|_| debug!("loaded full checkpoint contents in bulk for sequence {seq}"))
         {
-            let full_contents = full_contents.as_ref().clone();
             let num_txns = full_contents.size();
             let mut tx_digests = Vec::with_capacity(num_txns);
             let mut transactions = Vec::with_capacity(num_txns);
@@ -732,7 +731,7 @@ impl CheckpointExecutor {
             let mut fx_digests = Vec::with_capacity(num_txns);
 
             full_contents
-                .into_iter()
+                .iter()
                 .zip(checkpoint_contents.iter())
                 .for_each(|(execution_data, digests)| {
                     let tx_digest = digests.transaction;
@@ -742,11 +741,11 @@ impl CheckpointExecutor {
 
                     tx_digests.push(tx_digest);
                     transactions.push(VerifiedExecutableTransaction::new_from_checkpoint(
-                        VerifiedTransaction::new_unchecked(execution_data.transaction),
+                        VerifiedTransaction::new_unchecked(execution_data.transaction.clone()),
                         epoch,
                         seq,
                     ));
-                    effects.push(execution_data.effects);
+                    effects.push(execution_data.effects.clone());
                     fx_digests.push(fx_digest);
                 });
 
--- a/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -29,14 +29,14 @@ use iota_config::node::{CheckpointExecutorConfig, RunWithRange};
 use iota_macros::fail_point;
 use iota_sdk_types::{RandomnessRound, TransactionKind};
 use iota_types::{
-    base_types::{TransactionDigest, TransactionEffectsDigest},
+    base_types::{ExecutionData, TransactionDigest, TransactionEffectsDigest},
     effects::{TransactionEffects, TransactionEffectsAPI},
     executable_transaction::VerifiedExecutableTransaction,
     full_checkpoint_content::CheckpointData,
     global_state_hash::GlobalStateHash,
     messages_checkpoint::{
         CheckpointContents, CheckpointContentsExt, CheckpointSequenceNumber, CheckpointSummaryExt,
-        VerifiedCheckpoint,
+        FullCheckpointContents, VerifiedCheckpoint,
     },
     transaction::{TransactionDataAPI, TransactionKey, VerifiedTransaction},
 };
@@ -774,7 +774,7 @@ impl CheckpointExecutor {
 
             let (tx_digests, fx_digests): (Vec<_>, Vec<_>) =
                 digests.iter().map(|d| (d.transaction, d.effects)).unzip();
-            let transactions = self
+            let verified_transactions: Vec<VerifiedTransaction> = self
                 .transaction_cache_reader
                 .multi_get_transaction_blocks(&tx_digests)
                 .into_iter()
@@ -782,11 +782,10 @@ impl CheckpointExecutor {
                 .map(|(i, tx)| {
                     let tx = tx
                         .unwrap_or_else(|| fatal!("transaction not found for {:?}", tx_digests[i]));
-                    let tx = Arc::try_unwrap(tx).unwrap_or_else(|tx| (*tx).clone());
-                    VerifiedExecutableTransaction::new_from_checkpoint(tx, epoch, seq)
+                    Arc::try_unwrap(tx).unwrap_or_else(|tx| (*tx).clone())
                 })
                 .collect();
-            let effects = self
+            let effects: Vec<TransactionEffects> = self
                 .transaction_cache_reader
                 .multi_get_effects(&fx_digests)
                 .into_iter()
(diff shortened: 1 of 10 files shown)


roman1e2f5p8s (approved):

crates/iota-core/src/checkpoints/full_checkpoint_contents_cache.rs
```diff
@@ -0,0 +1,483 @@
...
+    /// Whether contents for `seq` are worth assembling and inserting: false
+    /// when the cache is disabled, and false when the cache is at budget and
+    /// `seq` is older than everything cached — lowest-seq eviction would
+    /// remove the entry immediately.
+    ///
+    /// Lets callers skip the expensive contents assembly that precedes
+    /// [`Self::insert`], e.g. the checkpoint executor during deep catch-up,
+    /// where the cache window rides the state-sync frontier far ahead of the
+    /// executor.
+    pub fn should_cache(&self, seq: CheckpointSequenceNumber) -> bool {
```
bingyanglin, line 181: `should_cache` returns true whenever `total_bytes < max_bytes`, but eviction always leaves the cache just under budget, so that's almost always true? 
So in deep catch-up (the exact case the doc names) the executor still clones every `tx`/`effect` and serializes the whole thing, and only then does insert throw it away. Maybe skip when the cache is non-empty and `seq < lowest`?

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -797,6 +796,38 @@ impl CheckpointExecutor {
                 })
                 .collect();
 
+            // Landing here means the contents were not synced via state sync
+            // but produced locally (or the cache entry was evicted). Cache the
+            // assembled contents so state-sync peers — e.g. a validator's
+            // SSFNs — are served the freshest checkpoints without
+            // reconstruction, speeding up checkpoint propagation.
```
bingyanglin: Might not necessary to mention the SSFN examples?

crates/iota-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -797,6 +796,38 @@ impl CheckpointExecutor {
...
+            //
+            // The assembly clones every transaction and effect, so skip it
+            // when the cache wouldn't retain the entry: cache disabled, or
+            // deep catch-up, where the cache window rides the state-sync
+            // frontier far ahead of the executor and would evict this entry
+            // immediately.
```
bingyanglin: Might be too much? I might just remove the comments after ":".

crates/iota-core/src/checkpoints/full_checkpoint_contents_cache.rs
```diff
@@ -0,0 +1,483 @@
...
+            FullCheckpointContentsCacheMetrics::new_for_tests(),
+        );
+        let (digest_a, contents_a) = entry(1);
+        let (digest_b, contents_b) = entry(2);
+        cache.insert(10, digest_a, contents_a, 100);
+        cache.insert(11, digest_b, contents_b, 100);
+
+        // At budget: entries older than the cached window would be evicted
+        // immediately, newer ones displace the window.
+        assert!(!cache.should_cache(5));
```
bingyanglin, line 465: This test only checks `should_cache` when `total_bytes == max_bytes` exactly, but after a real eviction it always lands under budget, and there `should_cache` returns true, so the skip never fires. Maybe add a case that evicts first, then assert a below-window seq is still skipped?

bingyanglin (approved):

PR #12152, opened 2026-07-03
upstream(iota-core,iota-types): pre-consensus dedup for duplicate transaction resubmissions

piotrm50:
# Description of change

- Port of upstream PR: [MystenLabs/sui#27069](https://github.com/MystenLabs/sui/pull/27069)
- Description: Adds a digest-keyed pre-consensus dedup layer that short-circuits duplicate transaction resubmissions within a short window (default 1s), so amplified resubmission storms are dropped before expensive signature verification.

Originally developed on the `consensus/feat/pcool-feature` stack (IOTA #11151); now that P-COOL has merged, this is rebased onto `develop` as a single commit. The only rebase conflict was an adjacency collision in `crates/iota-types/src/error.rs` (the new `RecentlyResubmitted` variant next to a reworded `TransactionNotFound`), resolved by keeping both.

Adapted to the fork:

- New `RecentSubmissions` sibling to `PreConsensusSoftLocks`: module `crates/iota-core/src/authority_server/recent_submission.rs` — a moka `Cache<TransactionDigest, Instant>` (TTL = window), `try_record` returning `Ok(())` on a miss and `Err(elapsed)` on a duplicate within the window. 4 unit tests cover first-insert, duplicate-within-window, expiry-after-window, and distinct-digests.
- Wired into `ValidatorService` (default 1s window) and into `submit_single_tx` (`authority_server/validator_v2.rs`) after the executed-effects short-circuit and before signature verification. This is orthogonal to the soft locks, which stay idempotent on the same digest by design: soft locks are object-keyed, whereas this dedup is digest-keyed and also covers shared-object and gasless transactions.
- New retryable error `IotaError::RecentlyResubmitted { digest }`, added to the retryable list so clients back off / wait for effects instead of treating it as a hard failure.
- Three Prometheus metrics: `validator_service_num_rejected_tx_recently_resubmitted`, `validator_service_recently_submitted_cache_size`, `validator_service_recently_submitted_resubmission_interval_seconds`.
- `DEFAULT_RECENT_SUBMISSION_WINDOW = 1s` is hard-coded to match the upstream default and stay consistent with `soft_lock.rs`'s hard-coded TTL; can be wired through `NodeConfig` later if per-deployment tuning is wanted.

### DIFF
--- a/crates/iota-core/src/authority_server/metrics.rs
+++ b/crates/iota-core/src/authority_server/metrics.rs
@@ -37,6 +37,9 @@ pub struct ValidatorServiceMetrics {
     pub x_forwarded_for_num_hops: Gauge,
     pub num_rejected_tx_soft_lock_conflict: IntCounter,
     pub soft_lock_table_size: IntGauge,
+    pub num_rejected_tx_recently_resubmitted: IntCounter,
+    pub recently_submitted_cache_size: IntGauge,
+    pub recently_submitted_resubmission_interval: Histogram,
 }
 
 impl ValidatorServiceMetrics {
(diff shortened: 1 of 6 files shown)


piotrm50:
Superseded by #12407, which suppresses the same duplicate resubmissions by reusing the existing pre-consensus soft locks (with a digest-keyed in-flight check) instead of adding a separate cache. Closing in favor of that approach.

Issue #10749, opened 2026-03-17
pcool: Deny governance — implement DenyRuleProposal consensus message


piotrm50:
> **Epic** — break into sub-tasks when work begins.

Implement the deny governance protocol designed in #10745.

Depends on: #10745

---

### Suggested sub-tasks (create when starting)
- Implement `DenyRuleProposal` consensus message type with BCS serialization
- Add verification of `DenyRuleProposal` in `consensus_validator.rs`
- Implement vote tally accumulation in `authority_per_epoch_store.rs`
- Implement epoch boundary rule computation (quorum check) and persistence
- Update `post_consensus_validation.rs` to apply voted deny rules instead of local config
- Coexistence path: local config as emergency override on top of governance rules
- Unit tests for voting and tally logic
- Integration tests across epoch boundary

filipdulic:
previous sugested sub-tasks, revised today.
Suggested sub-tasks (create when starting)

    Implement DenyRuleProposal consensus message type with BCS serialization
    Add verification of DenyRuleProposal in consensus_validator.rs
    Implement vote tally accumulation in authority_per_epoch_store.rs
    Implement epoch boundary rule computation (quorum check) and persistence
    Update post_consensus_validation.rs to apply voted deny rules instead of local config
    Coexistence path: local config as emergency override on top of governance rules
    Unit tests for voting and tally logic
    Integration tests across epoch boundary

PR #12153, opened 2026-07-03
feat(iota-core): add deny rule proposal persistence, stake aggregation, and active-set cache (#12151)

### LINKED ISSUE #12151: Add deny rule proposal persistence, stake aggregation, and active-set cache
Part of #10749. Adds the data layer for deny governance: per-authority proposal storage, stake-weighted aggregation into the active rule set, and a crash-safe in-memory cache. No consensus wiring yet — unit-tested in isolation (mirrors the `OverloadNotificationV1` pattern).

filipdulic:
# Description of change

Adds the data layer for deny governance (#10749), mirroring the `OverloadNotificationV1` pattern: per-authority proposal storage, stake-weighted aggregation into the active rule set, and a crash-safe in-memory cache. No consensus wiring yet — nothing records proposals in production (the recording entry point is `#[cfg(test)]` until the consensus handler lands), so the change is behavior-neutral and unit-tested in isolation.

- **Persistence:** epoch-scoped `deny_rule_proposals: DBMap<AuthorityName, DenyRuleProposal>`; proposals recorded into `ConsensusCommitOutput` and flushed atomically with `last_consensus_stats`; quarantine cache seeded at construction, advanced in lockstep at flush, overlaid with queued commits by `current_deny_rule_proposals()`.
- **Generation dedup:** a proposal supersedes only with a strictly newer generation, both across commits (`should_record_deny_rule_proposal`, backed by a no-clone `deny_rule_proposal_generation` lookup) and within one commit (`ConsensusCommitOutput::record_deny_rule_proposal`).
- **Aggregation:** `compute_active_deny_rules(proposals, committee)` — pure function; deny-list entries activate at f+1 (`validity_threshold`), kill switches at 2f+1 (`quorum_threshold`); non-members weigh 0. Each switch is tallied through a named field accessor (no positional coupling).
- **Active set:** `active_deny_rules: ArcSwap<DenyRuleSet>` + `get_active_deny_rules()`; seeded in `new()` by recomputing from the persisted table, so a mid-epoch restart resumes with the rules from all flushed commits (unflushed commits are re-recorded by consensus replay). `update_active_deny_rules()` reports whether the set changed — the seam a future on-chain materialization attaches to.
- `EpochStartConfiguration` (+V1/V2) gains `Clone`, enabling same-epoch store reconstruction in tests.

Fixes #12151.

### DIFF
--- a/crates/iota-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/iota-core/src/authority/authority_per_epoch_store.rs
@@ -9,7 +9,7 @@ use std::{
     sync::Arc,
 };
 
-use arc_swap::ArcSwapOption;
+use arc_swap::{ArcSwap, ArcSwapOption};
 use dashmap::DashMap;
 use enum_dispatch::enum_dispatch;
 use fastcrypto::{groups::bls12381, traits::ToFromBytes};
@@ -31,7 +31,7 @@ use iota_protocol_config::{
     Chain, PerObjectCongestionControlMode, ProtocolConfig, ProtocolVersion,
 };
 use iota_sdk_types::{
-    CancelledTransaction, CheckpointTimestamp, ObjectId, ObjectReference, RandomnessRound,
+    Address, CancelledTransaction, CheckpointTimestamp, ObjectId, ObjectReference, RandomnessRound,
     TransactionDigest, TransactionEffectsDigest, TransactionKind, Version, VersionAssignment,
 };
 use iota_storage::mutex_table::{MutexGuard, MutexTable};
@@ -39,6 +39,7 @@ use iota_types::{
     base_types::{AuthorityName, CommitRound, ConciseableName, EpochId},
     committee::{Committee, CommitteeTrait, StakeUnit},
     crypto::{AuthoritySignInfo, AuthorityStrongQuorumSignInfo},
+    deny_rule_governance::{DenyRuleProposal, DenyRuleSet},
     digests::ChainIdentifier,
     effects::TransactionEffects,
     error::{IotaError, IotaResult},
@@ -698,6 +699,12 @@ pub struct AuthorityPerEpochStore {
     /// `Rejected` response instead of waiting for the gRPC deadline.
     dropped_tx_status_cache: super::dropped_tx_status_cache::DroppedTxStatusCache,
 
+    /// The active deny rule set: the stake-weighted aggregate of the recorded
+    /// deny rule proposals. Recomputed via
+    /// `update_active_transaction_deny_rules` and seeded from the
+    /// `deny_rule_proposals` table on construction.
+    active_transaction_deny_rules: ArcSwap<DenyRuleSet>,
+
     /// Pre-consensus soft locks for owned objects (P-COOL flow).
     ///
     /// Set via `OnceCell` rather than passed through the constructor because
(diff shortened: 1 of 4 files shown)


bingyanglin (approved):

roman1e2f5p8s (approved):

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2943,6 +2970,113 @@ impl AuthorityPerEpochStore {
...
+        info!(?rules, "active deny rules changed");
+        self.active_deny_rules.store(Arc::new(rules));
+        Ok(true)
+    }
+
+    /// Computes the stake-weighted aggregate of the given deny rule proposals:
+    /// a deny list entry is active with f+1 supporting stake, a kill switch
+    /// with 2f+1. Pure function of the proposals and committee, so every
+    /// validator derives the same set from the same consensus state.
+    pub(crate) fn compute_active_deny_rules(
```
piotrm50: This function is key to the logic but it's only covered by a single test with out hitting edge cases for stake thresholds. Also, not all types of rules are covered by the test.

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2943,6 +2970,113 @@ impl AuthorityPerEpochStore {
...
+
+    /// Whether `proposal` is newer than the recorded proposal (if any) from
+    /// the same authority and should therefore be recorded.
+    pub fn should_record_deny_rule_proposal(&self, proposal: &DenyRuleProposal) -> bool {
+        let stale = self
+            .consensus_quarantine
+            .read()
+            .deny_rule_proposal_generation(&proposal.authority)
+            .is_some_and(|generation| generation >= proposal.generation);
+        if stale {
```
piotrm50: does this log belong here or in the caller?

crates/iota-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -698,6 +699,11 @@ pub struct AuthorityPerEpochStore {
     /// `Rejected` response instead of waiting for the gRPC deadline.
     dropped_tx_status_cache: super::dropped_tx_status_cache::DroppedTxStatusCache,
 
+    /// The active deny rule set: the stake-weighted aggregate of the recorded
+    /// deny rule proposals. Recomputed via `update_active_deny_rules` and
+    /// seeded from the `deny_rule_proposals` table on construction.
+    active_deny_rules: ArcSwap<DenyRuleSet>,
```
piotrm50: shoul we call it transaction_deny_rules to contrast and avoid confusion with coin_deny_rules?

crates/iota-core/src/authority/epoch_start_configuration.rs
```diff
@@ -98,7 +98,7 @@ impl fmt::Display for EpochFlag {
 }
 
 /// Parameters of the epoch fixed at epoch start.
-#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+#[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq)]
```
piotrm50, line 101: Why do we need clone here?

piotrm50 (approved):
left some nit comments, approve overall

crates/iota-core/src/authority/epoch_start_configuration.rs
filipdulic: clone is needed by the new persistence test, which reopens the epoch store to check that recorded proposals survive a restart: `AuthorityPerEpochStore::new` takes the config by value, while the original store holds it in an `Arc`. The enum is plain data, so the derive is cheap and safe.

crates/iota-core/src/authority/authority_per_epoch_store.rs
filipdulic: Extended the test to cover all three deny lists and all six switches at 3/2/1/0 supporting members plus a non-member. Note the lists share one generic activated() helper and the switches share switch_active, so code-path coverage was already complete — this makes field coverage explicit.

crates/iota-core/src/authority/authority_per_epoch_store.rs
filipdulic: moved to an else branch at the call site.

crates/iota-core/src/authority/authority_per_epoch_store.rs
filipdulic: Renamed across the stack: active_transaction_deny_rules + matching accessors.

PR #12162, opened 2026-07-06
fix(iota-framework): handle empty eligible validators in committee selection

piotrm50:
# Description of change

Committee selection remaps the set of eligible validators each epoch based on pending removals and low-stake departures. When that remapping leaves the eligible set empty, the downstream stake calculations have no members to work with and the epoch transition aborts.

This change makes committee selection resilient to an empty eligible set: when the remapped set is empty, it falls back to all current active validators so a non-empty committee is always produced. It also extracts a small `all_validator_indices` helper used by the fallback paths and consolidates the duplicated epoch-advance test helpers into one parameterized helper.

### Release Notes

- [x] Protocol: Fixed an issue where committee selection could produce an empty committee when validator removals or low-stake departures emptied the eligible-validator set, aborting the epoch transition. Thanks to [@Najnomics](https://github.com/Najnomics) for reporting this issue.


crates/iota-protocol-config/src/lib.rs
```diff
@@ -2964,6 +2966,8 @@ impl ProtocolConfig {
                     // iota_system via a new iota_system::protocol_config
                     // module.
                 }
+                // version 31 is a new framework version but with no config changes
+                31 => {}
```
bingyanglin, line 2970: #12163 also bumps to protocol 31 and adds a different `bytecode_snapshot/31` build.

valeriyr (approved):

tomxey (approved):

bingyanglin (approved):

PR #12163, opened 2026-07-06
refactor: validator metadata validation cleanup

valeriyr:
# Description of change

1. Make metadata checks stricter.
2. Code reorganisation and cleanup.
3. Improves testing of metadata.


crates/iota-protocol-config/src/lib.rs
```diff
@@ -2963,6 +2971,8 @@ impl ProtocolConfig {
                     // Also expose `is_feature_enabled` and `get_attr<T>` to
                     // iota_system via a new iota_system::protocol_config
                     // module.
+
+                    cfg.feature_flags.validator_metadata_verify_v2 = true;
```
thibault-martinez, line 2979: Protocol Version 30 has already been cut

valeriyr: Moved to 31

crates/iota-framework/packages/iota-system/tests/validator_metadata_tests.move
```diff
@@ -0,0 +1,454 @@
+// Copyright (c) Mysten Labs, Inc.
```
Dkwcs, line 1: Consider porting `duplicate_metadata_resubmission_after_inactive` as well.
This is the one metadata test from the upstream PR not covered here.
The short scenario -  a validator's addresses/keys must become reusable once it is neither active nor pending, so a later candidate can register with the same metadata.

crates/iota-protocol-config/src/lib.rs
```diff
@@ -2964,6 +2973,9 @@ impl ProtocolConfig {
                     // iota_system via a new iota_system::protocol_config
                     // module.
                 }
+                31 => {
+                    cfg.feature_flags.validator_metadata_verify_v2 = true;
+                }
```
bingyanglin, line 2980: #12162 also bumps to protocol 31 and adds its own bytecode_snapshot/31 with a different framework build.

crates/iota-framework/packages/iota-system/tests/validator_metadata_tests.move
valeriyr: If we consider the `validator_metadata_tests.move` module, `duplicate_metadata_resubmission_after_inactive` is not the only test that is not ported.
I ported only the tests related to the upstream changes; all the other, I think, should be added in the related PRs.

miker83z (approved):
LGTM!

crates/iota-protocol-config/src/lib.rs
piotrm50: yes, we are coordinating those together

piotrm50 (approved):

tomxey (approved):

semenov-vladyslav (approved):

alexsporn (approved):

PR #12165, opened 2026-07-06
feat(move-natives, iota-framework, iota-system): per-test protocol config flag overrides for Move unit tests

### LINKED ISSUE #12149: [move] Per-test protocol config flag overrides for Move unit tests
## Summary

Move unit tests can check protocol config flags via `protocol_config::is_feature_enabled_system()` (or similar accessors), but there's currently no way to override flag values on a per-test basis within the test code itself. Environment variables can override flags globally for an entire `iota move test` run, but tests that expect a flag to be enabled will fail in CI where flags default to off.

## Current behavior

**What works:**
- Global override via environment variables when invoking `iota move test`:
  ```bash
  IOTA_PROTOCOL_CONFIG_OVERRIDE_ENABLE=1 \
  IOTA_PROTOCOL_CONFIG_FEATURE_FLAGS_OVERRIDE_ENABLE_WHITE_FLAG_FLOW=true \
  IOTA_PROTOCOL_CONFIG_FEATURE_FLAGS_OVERRIDE_ENABLE_VALIDATOR_ATTESTATION=true \
  iota move test test_name
  ```
- This successfully changes flag values for the entire test run

**What doesn't work:**
- Per-test flag control within Move test code
- Committing tests that expect flags to be enabled (they fail in CI where overrides aren't set)

## Problem

It is needed a way to override per test the flags, in a similar fashion to the e2e tests.

Example scenario:
- `test_register_attestor_requires_feature_flag` expects `ENABLE_VALIDATOR_ATTESTATION` to be **false** and tests the error path
- `test_register_attestor` in the same module expects `ENABLE_VALIDATOR_ATTESTATION` to be **true** and tests the happy path

## What's needed

A mechanism similar to e2e test flag overrides, but for Move unit tests. Developers should be able to:
1. Write Move tests that explicitly declare which protocol config state they need
2. Run tests locally with different flag configurations
3. Commit tests that work in CI without environment variable setup
4. Test both enabled and disabled paths for feature-flagged code in Move

## Starting points for investigation

- How do e2e tests currently override protocol config flags on a per-test basis?
- Is there a test-only Move API or annotation that could declare required flag state?
- Should overrides be explicit in test code, or via test-framework metadata/attributes?
- What's the interaction between Move's `#[test]` framework and the protocol config runtime?
- How does the test framework currently handle `IotaSystemAdminCap` in tests (via `new_system_admin_cap_for_testing()`) — could similar test-only plumbing work for flag overrides?

Dkwcs:
# Description of change

Previously Move unit tests could read flags/params but not set them per test - env-var overrides apply to the entire `iota move test` run, so tests expecting a flag enabled failed in CI where flags default to their on-chain state.

Adds two test-only Move natives to override protocol config per unit test:

`protocol_config::set_feature_enabled_for_testing(name, value) // feature flag`
`protocol_config::set_attr_for_testing<T>(name, value) // config parameter`

Overrides live for the whole test and are read back by `is_feature_enabled` / `get_attr`, so a single module can test both the enabled and disabled (or differently-valued) paths of feature-flagged code.

Fixes #12149 .

For `iota-framework` - `iota move test protocol_config`
For `iota-system` -  `iota move test protocol_config`


iota-execution/latest/iota-move-natives/src/protocol_config.rs
```diff
@@ -139,3 +163,53 @@ pub fn get_attr(
...
+        Err(_) => {
+            return Ok(NativeResult::err(
+                context.gas_used(),
+                E_INVALID_UTF8_PARAM_NAME,
+            ));
+        }
+    };
+
+    let protocol_config = get_extension!(context, ObjectRuntime)?.protocol_config;
+    if protocol_config.lookup_feature(name.clone()).is_none() {
```
miker83z, line 219: make it so that you can also set attributes with the same logic (i.e., you first check if the attribute is present and the user passed the correct type, then you set the attribute in ProtocolConfigTestOverrides)

crates/iota-framework/packages/iota-system/sources/protocol_config.move
```diff
@@ -27,3 +27,16 @@ public(package) native fun is_feature_enabled(feature_flag_name: vector<u8>): bo
...
+/// path. The override is read back by `is_feature_enabled` and reset between
+/// tests.
+///
+/// Aborts if `feature_flag_name` is not valid UTF-8 or does not name a flag in
+/// the current protocol config.
+#[test_only]
+public(package) native fun set_feature_enabled_for_testing(
+    feature_flag_name: vector<u8>,
+    value: bool,
+);
```
miker83z, line 41:
```suggestion
public(package) native fun set_protocol_config_value_for_testing<T: copy + drop + store>(
    name: vector<u8>,
    value: T,
);
```
see https://github.com/iotaledger/iota/pull/12165/changes#r3545068941

iota-execution/latest/iota-move-natives/src/protocol_config.rs
Dkwcs: c7c7f2360d4025484e1b0268570cffdfc629971a

crates/iota-framework/packages/iota-system/sources/protocol_config.move
Dkwcs: c7c7f2360d4025484e1b0268570cffdfc629971a

iota-execution/latest/iota-move-natives/src/protocol_config.rs
```diff
@@ -139,3 +177,78 @@ pub fn get_attr(
...
+/// reset between tests, as the [`ProtocolConfigTestOverrides`] extension is
+/// created fresh per test.
+///
+/// Aborts with `E_INVALID_UTF8_PARAM_NAME` if `name` is not valid UTF-8, with
+/// `E_PARAM_NOT_FOUND` if it names neither a flag nor a parameter present in
+/// the current protocol config, and with `E_TYPE_MISMATCH` if `T` does not
+/// match the target's type — so a mistyped name or wrong type fails loudly.
+/// ****************************************************************************
+/// *******************
+pub fn set_protocol_config_value_for_testing(
```
valeriyr: I would create two separate functions:

set_attr_for_testing
set_feature_enabled_for_testing

It will make code cleaner and more consistent.
Also, now it is impossible to set an attribute if it has the same name as a feature flag.

Dkwcs: fc4bb00fe9505500403eb95234c991046295977f

miker83z (approved):

valeriyr (approved):

PR #12183, opened 2026-07-08
feat(grpc): add `balance_changes` and `object_changes` to `ExecutedTransaction`

### LINKED ISSUE #11243: [Feature Request] Add balance_changes and object_changes fields to gRPC ExecutedTransaction
## Bug description

Briefly describe the bug.

## Context

**Components** (select all that apply):

- [ ] `iota-indexer`
- [ ] `iota-json-rpc`
- [ ] `iota-graphql-rpc`
- [ ] Other: __________

**Issue type** (select one):

- [ ] JSON-RPC API bug (failing request, unexpected response, etc.)
- [ ] Crate bug (logic bug, runtime error, performance issue, etc.)

## Version & Environment

Specify exactly which version, commit, or branch you're using, and where you observed the issue:

- **Version (tag), commit hash, or branch**:
- **Environment** (Mainnet, Testnet, Devnet, Local, other, etc.):

## Steps to reproduce the bug

Provide clear, copy-&-paste instructions to reproduce the issue.

- For JSON-RPC or GraphQL API bugs, include your request and full response.
- If relevant, add commands or scripts to set up test data locally (e.g., PTB commands).

Example JSON-RPC request:

```sh
curl 'http://localhost:9124/' \
-H 'Content-Type: application/json' \
--data-raw '{"jsonrpc":"2.0","id":3,"method":"iota_getTransactionBlock","params":["HZ6wGXArNs6gagzaVZruprPX7QaaxfPo5ZFS76wUpK4f",{"showInput":true,"showEffects":true,"showEvents":true,"showBalanceChanges":true,"showObjectChanges":true}]}' | json_pp
```

## Expected behaviour

Describe in detail what you expect to happen.

## Actual behaviour

Describe clearly what actually happens, including error messages or unexpected behavior.

## Errors & logs

Paste any errors, logs, stack traces, or screenshots here.

## System details (optional, recommended)

Provide additional system details if relevant:

- **OS & Kernel**:
- **RAM**:
- **CPU cores**:
- **Container/VM**: yes ☐ / no ☐, details:
- **Network**: VPN? Proxy? Firewall?

muXxer:
# Description of change

Adds server-side support for `balance_changes` and `object_changes` on the gRPC API, the equivalent of the JSON-RPC `showBalanceChanges` / `showObjectChanges` response options. Both fields are strictly opt-in — populated only when the request's read mask names them — and are not part of any default read mask.

Companion PR adding the proto types: iotaledger/iota-rust-sdk#1255. This PR bumps the `iota-rust-sdk` rev to pick up those types; the rev should be re-pointed to the merge commit once https://github.com/iotaledger/iota-rust-sdk/pull/1255 lands.

## What & how

- New synchronous derivation module `crates/iota-grpc-server/src/changes.rs`. It computes both change sets from data the server already holds in memory — the transaction effects plus its input/output objects — so **no new lookup tables or indexes** are introduced. Semantics mirror `crates/iota-json-rpc/src/balance_changes.rs` and `object_changes.rs`, including the gas-only balance change for failed transactions and the mocked-gas-coin exclusion for simulations.
- Wired through all four `ExecutedTransaction` producers: `get_transactions`, `execute_transactions`, `simulate_transactions`, and the checkpoint endpoints (`get_checkpoint` / `stream_checkpoints`, as `transactions.balance_changes` / `transactions.object_changes`).
- The derived fields force the fetch of effects and input/output objects (and the transaction, for the sender) even when those aren't themselves in the mask; over-fetched data never leaks into the response, as the `Merge` impls only populate mask-requested fields.

## Completeness contract on pruned objects

Requesting `input_objects` / `output_objects` / `balance_changes` / `object_changes` for a transaction whose objects the node no longer has (e.g. pruned) now returns a per-item `FAILED_PRECONDITION` naming the unavailable object, instead of a silently shortened object list or a numerically wrong balance delta. Clients should narrow the read mask, or fetch objects individually via `get_objects` for best-effort retrieval. Previously the raw object fields silently dropped missing objects; that path is closed too.

- Depends on iotaledger/iota-rust-sdk#1255 (proto types).

Unit tests in `changes.rs` cover the derivation (coin transfers, grouping, zero-delta drop, failed-tx gas-only, mocked-coin exclusion, the six object-change variants, and the missing-object error). E2e tests in `crates/iota-e2e-tests/tests/grpc/` exercise `get_transactions`, `execute_transactions`, `simulate_transactions`, the checkpoint endpoints, and a failed transaction, asserting the exact expected gRPC results (balance deltas from the transfer amount plus the gas charge read from the returned effects). `cargo ci-clippy`, `cargo +nightly fmt`, `IOTA_SKIP_SIMTESTS=1 cargo nextest run -p iota-grpc-server` (67 tests) and the full gRPC e2e suite (103 tests) pass; the new e2e tests also pass under `cargo simtest`.

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Full node gRPC API can now derive and return per-transaction balance and object changes.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] gRPC: `ExecutedTransaction` responses can include opt-in `balance_changes` and `object_changes` (via the read mask) on `get_transactions`, `execute_transactions`, `simulate_transactions` and the checkpoint endpoints. `get_transactions` now returns `FAILED_PRECONDITION` when a requested `input_objects` / `output_objects` / `balance_changes` / `object_changes` field needs an object the node has pruned, instead of silently omitting it from the response.

#### Breaking Changes Rollout

No breaking changes — the new response fields are purely additive and opt-in, and default read masks are unchanged.

### DIFF
--- a/crates/iota-e2e-tests/tests/grpc/utils.rs
+++ b/crates/iota-e2e-tests/tests/grpc/utils.rs
@@ -6,10 +6,11 @@ use std::collections::{HashMap, HashSet};
 use iota_grpc_client::{ReadMask, read_mask_fields::CheckpointResponseField};
 use iota_grpc_types::v1::types::{Address as ProtoAddress, ObjectId as ProtoObjectId};
 use iota_sdk_types::{
-    Address, ExecutionStatus, ObjectId, SignedTransaction, Transaction, TransactionDigest,
+    Address, ExecutionStatus, ObjectDigest, ObjectId, Owner, SignedTransaction, Transaction,
+    TransactionDigest, TypeTag,
 };
 use iota_test_transaction_builder::{TestTransactionBuilder, make_transfer_iota_transaction};
-use iota_types::effects::TransactionEffectsAPI;
+use iota_types::{effects::TransactionEffectsAPI, gas_coin::GAS};
 use test_cluster::{TestCluster, TestClusterBuilder};
 
 // --- Shared example package names used by filter tests ---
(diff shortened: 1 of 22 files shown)


crates/iota-grpc-server/src/types.rs
```diff
@@ -1218,37 +1237,47 @@ impl GrpcReader {
...
+                                 requested fields is unavailable (possibly pruned); narrow the \
+                                 read_mask or fetch objects individually via `get_objects` for best-effort retrieval"
+                            ),
+                        )
+                    })
+            };
+
             // Get input objects only if requested
-            let input_objects = if fields.include_input_objects {
+            let input_objects = if fields.include_input_objects || include_derived_changes {
```
bingyanglin, line 1261: This now errors on pruned objects even when someone asked for just `input_objects`/`output_objects`, not the new balance/object changes. Before it skipped the missing object and returned a partial list, so this quietly turns previously-OK requests into `FAILED_PRECONDITION`. Do we intend to do that?

alexsporn (approved):
LGTM

crates/iota-e2e-tests/tests/grpc/utils.rs
```diff
@@ -192,6 +193,231 @@ pub async fn wait_for_executed_transactions_checkpointed(
...
+pub fn assert_transfer_derived_changes(
+    executed_transaction: &iota_grpc_types::v1::transaction::ExecutedTransaction,
+    sender: Address,
+    recipient: Address,
+    amount: i128,
+    scenario: &str,
+) {
+    let gas = grpc_net_gas_usage(executed_transaction) as i128;
+    let mut expected = vec![
+        (Owner::Address(sender), GAS::type_tag(), -(amount + gas)),
```
semenov-vladyslav, line 366: Using i128 makes it seem like there is a possibility of overflow, since the absolute value must fit u64. Of course, correct values can't produce incorrect results. This should be Ok.

semenov-vladyslav (approved):

crates/iota-grpc-server/src/types.rs
muXxer: yes, that was exactly what the second commit in this PR was about, and also explained in the description and in the release notes. 😄

Thoralf-M (approved):

PR #12184, opened 2026-07-08
feat(graphql): Support fallback for tx by checkpoint queries

### LINKED ISSUE #11936: [graphql]: add archival fallback for transactions filtered by checkpoint
Make GraphQL fall back to the archival store when listing transactions in a single checkpoint.

### GraphQL paths covered
- `Query.transactionBlocks(filter: { atCheckpoint: ... })`
- `Checkpoint.transactionBlocks`

### Implementation
Route the checkpoint-filtered transaction query through `IndexerReader::query_transactions_by_checkpoint_seq_with_fallback`. JSON-RPC `iotax_queryTransactionBlocks` already uses this method when the filter is `Checkpoint`.

Depends on: #11932
Parent: #11855

tomxey:
# Description of change

Adds historical fallback for GraphQL queries that list the transactions of a single checkpoint. Covered paths:

- `Query.transactionBlocks(filter: { atCheckpoint: ... })`
- `Checkpoint.transactionBlocks`

Queries with only the `atCheckpoint` filter are served by the new `IndexerReader::query_stored_transactions_by_checkpoint_seq_with_fallback` method. A pruned checkpoint is served from the historical storage when configured, and returns a `DATA_PRUNED` error otherwise (before this change, the page came back empty). Queries with other filters keep using the SQL path.

The fallback reader's `checkpoint_transactions` now takes the cursor either by transaction digest (used by JSON-RPC) or by `tx_sequence_number` (used by GraphQL, see `CheckpointTxCursor`). The page is fetched with one extra row on each side, so `hasPreviousPage`/`hasNextPage` can be filled.

fixes https://github.com/iotaledger/iota/issues/11936

### Release Notes

- [x] GraphQL: `Query.transactionBlocks(filter: { atCheckpoint })` and `Checkpoint.transactionBlocks` can now read pruned transactions from the fallback KV store when it is configured.

### DIFF
--- a/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
+++ b/crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
@@ -2,7 +2,10 @@
 // Modifications Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, HashMap};
+use std::{
+    collections::{BTreeMap, HashMap},
+    ops::Bound,
+};
 
 use async_graphql::{connection::CursorType, dataloader::Loader, *};
 use connection::Edge;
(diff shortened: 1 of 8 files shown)


sergiupopescu199 (approved):

crates/iota-indexer/src/historical_fallback/reader.rs
```diff
@@ -451,55 +464,58 @@ impl HistoricalFallbackReader {
...
-            .client
-            .multi_get_checkpoints_contents(&[checkpoint_sequence_number])
-            .await?
-            .into_iter()
-            .next()
-            .flatten()
+        let seqs = [checkpoint_sequence_number];
+        let (summaries, contents) = tokio::try_join!(
+            self.client
+                .multi_get_checkpoints_summaries_by_sequence_numbers(&seqs),
```
kodemartin, line 525: This is only necessary when `cursor.is_some` and a `Seq`, which only serves GraphQL. And while the refactoring can be read easier than the previous version, it is moreover less efficient as it allocates eagerly the `seq_digests`.

We would be probably better off with separate methods for each cursor variant (maybe even one for without a cursor as well?), making 2 calls in the fallback only when `Seq` is passed. Maintaining laziness could be also pursued, but this is not as important.

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
```diff
@@ -498,6 +510,85 @@ impl TransactionBlock {
...
+        // The inclusive window of rows the client asked for using cursors (cursors are
+        // exclusive).
+        let window_lo = after_seq.map_or(0, |after| after.saturating_add(1));
+        let window_hi = before_seq.map_or(u64::MAX, |before| before.saturating_sub(1));
+
+        // Fill the page with up to `limit` rows. Any row that does
+        // not make it into the page sets `has_prev`/`has_next` accordingly.
+        let mut page_rows: Vec<StoredTransaction> = Vec::with_capacity(limit);
+        let mut has_prev = false;
+        let mut has_next = false;
```
kodemartin: Why don't we use `Page::paginate_results` to resolve these? It seems that this patch reimplements logic that is maintained in `paginate_results` and is used throughout the API via the `paginate_{query,raw_query}` methods.

crates/iota-indexer/src/historical_fallback/reader.rs
tomxey: Split into 2 variants in 980f3dc5eb

crates/iota-graphql-rpc/src/types/transaction_block/mod.rs
tomxey: Done in 980f3dc5eb

kodemartin (approved):
lgtm 🚵

PR #12186, opened 2026-07-08
feat(core): execution-driven pruning

muXxer:
# Description of change

Follow-up to the chain-time pruning cutoff (#12132, already on `develop`). That
change stopped the database from growing unbounded during catch-up sync, but
pruning still ran on a background timer that accumulated ~10s of backlog and then
flushed it in one burst. On a live node each burst saturates the shared RocksDB
instance (cold reads of old checkpoint data, tombstone writes, and the compaction
they trigger on the same column families execution reads), so checkpoint
execution drops to **0 tx/s** for the duration of the burst and resumes once it
ends — the node repeatedly falls behind the network.

This PR makes pruning **execution-driven** so the work is spread evenly instead of
bursting, and adds a backpressure **leash** so on-disk state stays bounded without
a per-run rate cap (which could silently let the database grow under sustained
load).

Key idea: with a chain-time retention window, ~one checkpoint's worth of data ages
out of the window per checkpoint executed — at tip *and* during catch-up. So
pruning a little after every executed checkpoint is naturally rate-matched and
fits within the per-checkpoint slack.

Changes:
- **`PruningCoordinator`** (`authority_store_pruner.rs`): a shared handle with two
  `watch` channels — the executor nudges it after each checkpoint, and the pruner
  publishes a *frontier* = the highest-executed checkpoint timestamp it has caught
  up to on its last completed drain.
- **Pruner**: replaced the timer/`select!` loop with a nudge-driven loop. On each
  nudge it drains objects/checkpoints/indexes **fully** to the chain-time cutoff
  (no per-run cap — full-drain is self-correcting and cannot fall behind), then
  publishes the frontier. The `watch` nudge coalesces many executed checkpoints
  into a single drain. Periodic background compaction is unchanged.
- **Executor** (`checkpoint_executor/mod.rs`): after a checkpoint is executed and
  broadcast, it nudges the pruner (off the propagation path); before scheduling a
  checkpoint it awaits the leash — throttling while execution has run more than
  **1h of chain-time** ahead of the pruner's last completed drain. Uniform for
  validators and full nodes.
- **Config**: removed the now-unused `pruning-run-delay-seconds`,
  `max-checkpoints-in-batch`, and `max-transactions-in-batch` (batch sizes are now
  internal constants; they only bounded `WriteBatch` memory, never total work).
  Deserialization ignores unknown fields, so existing `fullnode.yaml` files with
  these options still load.

### Why the frontier is the executed position, not `pruned + window`

The leash needs a signal for "how far behind is the pruner." An earlier revision
derived it from the pruned watermark (`pruned_ts + retention_window`), but that
**deadlocked** on far-behind nodes: the drain is also bounded by the epoch guard
(the last `num_epochs_to_retain` epochs are never pruned) and `window` uses the
*current* `epoch_duration_ms`, which doesn't match the actual durations of the
*historical* epochs a catching-up node replays. When historical epochs are longer
than the current epoch duration, `pruned_ts + window` sits permanently below
`executed − slack`; the leash never opens, execution stops nudging, the
nudge-driven pruner sleeps, and the whole pipeline freezes (only state-sync keeps
moving).

Publishing the executed timestamp the pruner has drained up to instead makes the
leash measure "how far execution has run ahead of the pruner's last completed
drain" — a real, monotonically advancing value, independent of epoch-duration
variance and the epoch guard. It cannot deadlock (each completed drain lifts the
frontier; nudges buffered during a drain trigger the next drain immediately) and
still bounds disk: the pruner drains to `executed − window` each pass and the
leash keeps `executed − frontier ≤ slack`, so the retained span stays
`≤ window + slack`.

### Behavioral note (throttles on slowness, fails open on errors)

If pruning is merely slow (IO-bound), the drain still completes fully each pass —
it just takes longer — and the leash holds execution within the slack of the
drain's start, so disk stays bounded and the node self-throttles. If pruning
*persistently errors* (a bug, a corrupt store), the frontier still advances (it
tracks the executed position, published regardless of drain success), so execution
keeps going and the DB grows — i.e. it **fails open**, with errors logged loudly.
This is deliberate: it avoids the deadlock above, and fail-open is the safer choice
for liveness. If bounded disk on prune failure is ever required, gate it explicitly
(e.g. halt only after N consecutive errors) rather than by coupling the frontier to
pruned-watermark progress.

_No tracked issue. Follow-up to #12132._

Unit tests cover the coordinator (leash passes within slack, leash blocks until
the frontier advances, nudge wakes the pruner subscription) and the chain-time
cutoff drain (stops at the timestamp cutoff, bounded by the hard ceiling, cutoff
below all, duplicate boundary timestamps, epoch-count guard takes precedence).
`cargo clippy` and `cargo +nightly fmt` are clean; `iota-core`, `iota-config`,
`iota-node`, and `iota-tool` compile.

Note: end-to-end behavior under real epoch progression is best validated by the
sim/e2e pruning suite in CI and a live-node run (verified on a testnet full node
that a far-behind node keeps executing instead of deadlocking).

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Database pruning is now driven by
  checkpoint execution rather than a background timer: after each executed
  checkpoint the node prunes data that has aged past the retention window, pacing
  the work against on-chain checkpoint timestamps so it no longer stalls execution
  in bursts. Execution is throttled if pruning falls behind, keeping on-disk size
  bounded. The `authority-store-pruning-config` options `pruning-run-delay-seconds`,
  `max-checkpoints-in-batch`, and `max-transactions-in-batch` have been removed and
  are safely ignored if still present.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-config/src/node.rs
+++ b/crates/iota-config/src/node.rs
@@ -1016,16 +1016,6 @@ pub struct AuthorityStorePruningConfig {
     /// Use `u64::MAX` to disable the pruner for the objects.
     #[serde(default)]
     pub num_epochs_to_retain: u64,
-    /// pruner's runtime interval used for aggressive mode
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub pruning_run_delay_seconds: Option<u64>,
-    /// maximum number of checkpoints in the pruning batch. Can be adjusted to
-    /// increase performance
-    #[serde(default = "default_max_checkpoints_in_batch")]
-    pub max_checkpoints_in_batch: usize,
-    /// maximum number of transaction in the pruning batch
-    #[serde(default = "default_max_transactions_in_batch")]
-    pub max_transactions_in_batch: usize,
     /// enables periodic background compaction for old SST files whose last
     /// modified time is older than `periodic_compaction_threshold_days`
     /// days. That ensures that all sst files eventually go through the
@@ -1054,14 +1044,6 @@ fn default_num_latest_epoch_dbs_to_retain() -> usize {
     3
 }
 
-fn default_max_transactions_in_batch() -> usize {
-    1000
-}
-
-fn default_max_checkpoints_in_batch() -> usize {
-    10
-}
-
 fn default_periodic_compaction_threshold_days() -> Option<usize> {
     Some(1)
 }
@@ -1071,9 +1053,6 @@ impl Default for AuthorityStorePruningConfig {
         Self {
             num_latest_epoch_dbs_to_retain: default_num_latest_epoch_dbs_to_retain(),
             num_epochs_to_retain: 0,
-            pruning_run_delay_seconds: if cfg!(msim) { Some(2) } else { None },
-            max_checkpoints_in_batch: default_max_checkpoints_in_batch(),
-            max_transactions_in_batch: default_max_transactions_in_batch(),
             periodic_compaction_threshold_days: None,
             num_epochs_to_retain_for_checkpoints: if cfg!(msim) { Some(2) } else { None },
             enable_compaction_filter: cfg!(test) || cfg!(msim),
--- a/crates/iota-core/src/authority/authority_store_pruner.rs
+++ b/crates/iota-core/src/authority/authority_store_pruner.rs
@@ -30,7 +30,10 @@ use prometheus_filtered::{
     register_int_gauge_with_registry,
 };
 use tokio::{
-    sync::oneshot::{self, Sender},
+    sync::{
+        oneshot::{self, Sender},
+        watch,
+    },
     time::Instant,
 };
 use tracing::{debug, error, info, warn};
@@ -64,6 +67,35 @@ static PERIODIC_PRUNING_TABLES: Lazy<BTreeSet<String>> = Lazy::new(|| {
 pub const EPOCH_DURATION_MS_FOR_TESTING: u64 = 24 * 60 * 60 * 1000;
 pub const MIN_EPOCHS_TO_RETAIN_FOR_INDEXES: u64 = 7;
 
+/// Maximum number of checkpoints whose data is written in a single pruning
+/// `WriteBatch`. Bounds batch memory only; it does not cap total work per run,
+/// so it cannot cause the pruner to fall behind.
+const MAX_CHECKPOINTS_IN_BATCH: usize = 10;
+/// Maximum number of transactions whose effects are written in a single pruning
+/// `WriteBatch`. Bounds batch memory only (see [`MAX_CHECKPOINTS_IN_BATCH`]).
+const MAX_TRANSACTIONS_IN_BATCH: usize = 1000;
+
+/// Chain-time slack, in milliseconds, allowed on top of the retention window
+/// before the checkpoint executor is throttled by [`PruningCoordinator`]. It
+/// absorbs transient bursts of high-contention checkpoints so execution runs at
+/// the average prune rate rather than the peak; under sustained overload the
+/// retained span stabilizes at `window + PRUNING_LEASH_SLACK_MS`, which is
+/// negligible next to a multi-epoch window.
+const PRUNING_LEASH_SLACK_MS: u64 = 60 * 60 * 1000;
+
+/// While catching up (see [`PRUNING_DEBOUNCE_MIN_LAG`]), after a nudge wakes
+/// the pruner it waits this long before draining so that more executed
+/// checkpoints accumulate and their object deletions coalesce into larger,
+/// fewer batches — which measurably improves catch-up throughput. Negligible
+/// against the leash slack, so it never risks throttling execution.
+const PRUNING_NUDGE_DEBOUNCE: Duration = Duration::from_millis(1000);
+
+/// The debounce above is only applied while the node is catching up, i.e. when
+/// execution lags the highest synced checkpoint by more than this many
+/// checkpoints. Near the tip the lag is tiny, so pruning stays prompt
+/// (per-checkpoint) and does not incur the debounce delay.
+const PRUNING_DEBOUNCE_MIN_LAG: u64 = 100;
+
 /// The `AuthorityStorePruner` manages the pruning process for object stores
 /// within the `AuthorityStore`. It includes a cancellation handle that can be
 /// used to stop the pruning task for objects.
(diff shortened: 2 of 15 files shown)


crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -63,14 +66,114 @@ static PERIODIC_PRUNING_TABLES: Lazy<BTreeSet<String>> = Lazy::new(|| {
...
+/// drains fully to its chain-time retention cutoff on every nudge. To keep
+/// on-disk state bounded without a per-run rate cap (which could silently let
+/// the database grow under sustained load), the executor is *leashed*: it stops
+/// scheduling checkpoints while the pruner has fallen more than
+/// `PRUNING_LEASH_SLACK_MS` behind its retention target.
+pub struct PruningCoordinator {
+    /// Executor -> pruner: latest executed checkpoint sequence number. Updating
+    /// it both records progress and wakes the pruner to drain.
+    executed: watch::Sender<CheckpointSequenceNumber>,
+    executed_rx: watch::Receiver<CheckpointSequenceNumber>,
```
semenov-vladyslav: I think it's not necessary to keep the receiver copy, just `self.executed.subscribe()`.

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -63,14 +66,114 @@ static PERIODIC_PRUNING_TABLES: Lazy<BTreeSet<String>> = Lazy::new(|| {
...
+    /// Pruner -> executor: the executed-checkpoint timestamp the pruner has
+    /// caught up to (the `highest_executed` it observed on its last completed
+    /// drain). The leash throttles execution while it runs more than
+    /// `PRUNING_LEASH_SLACK_MS` of chain-time ahead of this, i.e. ahead of
+    /// the pruner's last completed drain. Initialized to `u64::MAX` so the
+    /// executor is never leashed before the pruner has published a real
+    /// value.
+    frontier_ms: watch::Sender<CheckpointTimestamp>,
+    frontier_ms_rx: watch::Receiver<CheckpointTimestamp>,
```
semenov-vladyslav: Same here.

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -63,14 +66,114 @@ static PERIODIC_PRUNING_TABLES: Lazy<BTreeSet<String>> = Lazy::new(|| {
...
+    executed: watch::Sender<CheckpointSequenceNumber>,
+    executed_rx: watch::Receiver<CheckpointSequenceNumber>,
+    /// Pruner -> executor: the executed-checkpoint timestamp the pruner has
+    /// caught up to (the `highest_executed` it observed on its last completed
+    /// drain). The leash throttles execution while it runs more than
+    /// `PRUNING_LEASH_SLACK_MS` of chain-time ahead of this, i.e. ahead of
+    /// the pruner's last completed drain. Initialized to `u64::MAX` so the
+    /// executor is never leashed before the pruner has published a real
+    /// value.
+    frontier_ms: watch::Sender<CheckpointTimestamp>,
```
semenov-vladyslav, line 124: Since checkpoints are produced pretty regularly why not use `CheckpointSequenceNumber` directly?

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -63,14 +66,114 @@ static PERIODIC_PRUNING_TABLES: Lazy<BTreeSet<String>> = Lazy::new(|| {
...
+        let mut rx = self.frontier_ms_rx.clone();
+        loop {
+            let frontier = *rx.borrow_and_update();
+            if executed_timestamp_ms.saturating_sub(frontier) <= PRUNING_LEASH_SLACK_MS {
+                return;
+            }
+            // If the pruner is gone, do not block execution.
+            if rx.changed().await.is_err() {
+                return;
+            }
```
semenov-vladyslav: Not sure how can this be triggered. `changed()` only returns error in case sender is dropped which is never while there is `&self` no allowing to drop sender.

crates/iota-core/src/authority/authority_store_pruner.rs
```diff
@@ -748,25 +830,120 @@ impl AuthorityStorePruner {
...
-                        if let Err(err) = Self::prune_checkpoints_for_eligible_epochs(&perpetual_db, &checkpoint_store, grpc_indexes_store.as_deref(), pruner_db.as_ref(), config.clone(), metrics.clone(), archive_readers.clone(), epoch_duration_ms, progress_tracker.as_ref()).await {
-                            error!("Failed to prune checkpoints: {:?}", err);
-                        }
-                    },
-                    _ = indexes_prune_interval.tick(), if config.num_epochs_to_retain_for_indexes.is_some() => {
-                        if let Err(err) = Self::prune_indexes(jsonrpc_index.as_deref(), &config, epoch_duration_ms, &metrics) {
-                            error!("Failed to prune indexes: {:?}", err);
+                    _ = &mut recv => break,
+                    res = executed_rx.changed() => {
+                        if res.is_err() {
```
semenov-vladyslav: I don't think `res` can ever be Err. "executed_tx" is never dropped since it is owned by `coordinator` which is moved into this future.

crates/iota-config/src/node.rs
```diff
@@ -1016,16 +1016,6 @@ pub struct AuthorityStorePruningConfig {
...
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub pruning_run_delay_seconds: Option<u64>,
-    /// maximum number of checkpoints in the pruning batch. Can be adjusted to
-    /// increase performance
-    #[serde(default = "default_max_checkpoints_in_batch")]
-    pub max_checkpoints_in_batch: usize,
-    /// maximum number of transaction in the pruning batch
-    #[serde(default = "default_max_transactions_in_batch")]
-    pub max_transactions_in_batch: usize,
     /// enables periodic background compaction for old SST files whose last
```
semenov-vladyslav, line 1019: `fullnode.yaml` templates should also be adjusted:
- `crates/iota-config/data/fullnode-template.yaml`
- `crates/iota-config/data/fullnode-template-with-path.yaml`
- `dev-tools/iota-network/genesis/static/fullnode.yaml`
- `dev-tools/iota-private-network/configs/fullnodes/`

crates/iota-core/src/authority/authority_store_pruner.rs
muXxer: Because the whole smoothing of pruning objects and checkpoints is now based on onchain time since https://github.com/iotaledger/iota/pull/12132, and IMO it makes sense to just add some extra slack time in the same unit.

semenov-vladyslav (approved):
The logic seems sound. I would put PrunerCoordinator implementation details (the two watch channels) into AuthorityStorePruner, no need to add another glue component. Other than that looks good.

jkrvivian (approved):

sergiupopescu199 (approved):

PR #12191, opened 2026-07-09
refactor(iota-types): replace node CheckpointContents with SDK type

DaughterOfMars:
## Why

Final PR of the checkpoint-family track (cf. `Object` #11718, `CheckpointCommitment` #11976, `EndOfEpochData` #12027, `CheckpointSummary` #12091): retire the duplicated node `CheckpointContents` in favor of `iota_sdk_types::checkpoint::CheckpointContents`.

## What

- Drop the node `CheckpointContents` enum / `CheckpointContentsV1`, re-export the SDK type. Its custom serializer reproduces the exact V1 parallel-vector BCS, so the wire format is unchanged.
- Move node-only helpers (`new_with_*`, `iter`, `into_iter_with_signatures`, `enumerate_transactions`, `inner`, `into_inner`, `size`) to a `CheckpointContentsExt` trait; `digest()` comes from the SDK type's inherent method (byte-identical to the old node digest).
- Bridge signatures at the contents boundary only: the SDK type stores `Vec<UserSignature>` while the node uses `GenericSignature` elsewhere. The two are BCS-identical, so the boundary conversion is total for signatures from valid transactions. No broader `GenericSignature → UserSignature` migration.
- Accessors return owned `ExecutionDigests` (the SDK stores a different layout); the `OnceCell` digest cache is dropped.
- Delete the node↔SDK `CheckpointContents` conversions; the `CheckpointData` conversions now move the (now-identical) `checkpoint_contents` field directly.

## Format

BCS wire format is unchanged. The serde-reflection snapshot re-shapes the `CheckpointContents` entry (tuple variant, `BYTES` signatures) purely because the SDK type uses a custom serializer and `UserSignature` serializes as opaque bytes — representation-only. `generate-format` registers the type from a value sample since that opaque-bytes form can't be synthesized by `trace_type`.

## Test plan

- `cargo check --workspace --all-targets`, `cargo clippy -p iota-types -p iota-core` — clean.
- `iota-core` format snapshot test passes (inspected; representation-only).
- 880 `iota-types` / `iota-core` lib tests pass.


crates/iota-config/src/migration_tx_data.rs
```diff
@@ -95,17 +95,19 @@ impl MigrationTxData {
...
     ) -> anyhow::Result<()> {
         anyhow::ensure!(
-            checkpoint.content_digest == *contents.digest(),
+            checkpoint.content_digest == contents.digest(),
             "checkpoint's content digest is corrupted"
         );
         let mut validation_digests_queue: HashSet<TransactionDigest> =
             self.inner.keys().copied().collect();
         // We skip the genesis transaction to process only migration transactions from
         // the migration.blob.
```
Thoralf-M: Would move the comment directly above the genesis_tx_digest comparison

crates/iota-types/src/iota_sdk_types_conversions.rs
```diff
@@ -68,64 +65,6 @@ impl TryFrom<crate::object::Object> for Object {
...
-        );
-        crate::messages_checkpoint::CheckpointContents::new_with_digests_and_signatures(
-            transactions,
-            user_signatures.into_iter().collect::<Result<Vec<_>, _>>()?,
-        )
-        .pipe(Ok)
-    }
-}
-
 impl TryFrom<crate::full_checkpoint_content::CheckpointData> for CheckpointData {
```
Thoralf-M, line 68: Why keep the `full_checkpoint_content::CheckpointData` and not replace it as well?

DaughterOfMars: From claude's notes:

### Two checkpoint types that look in-scope but are blocked

`full_checkpoint_content::{CheckpointData, CheckpointTransaction}` cannot be
replaced now: `CheckpointData.checkpoint_summary` is a node
`CertifiedCheckpointSummary` (an `Envelope` over consensus signatures) and
`CheckpointTransaction.transaction` is a node `Transaction`
(`Envelope<SenderSignedData, …>`) — both in the "not feasibly replaceable"
bucket. Their conversions (conversions.rs:130-218) **stay**. They only become
removable after the transaction-envelope and consensus-signature story is
solved, which is a separate, much larger effort.

Thoralf-M (approved):

crates/iota-types/src/messages_checkpoint.rs
```diff
@@ -343,132 +343,165 @@ impl CheckpointSignatureMessage {
...
+        transaction: info.transaction,
+        effects: info.effects,
+    }
+}
+
+/// Node-only helpers for [`CheckpointContents`], which is defined in
+/// `iota_sdk_types`. They bridge the node's `ExecutionDigests` /
+/// [`GenericSignature`] representation to the SDK type's parallel
+/// [`CheckpointTransactionInfo`] / [`UserSignature`] form.
+pub trait CheckpointContentsExt: Sized {
```
thibault-martinez: seal it

thibault-martinez (changes requested):
`CheckpointContents` is not versioned in the SDK (even though its serialisation is), it probably should.

crates/iota-types/src/messages_checkpoint.rs
```diff
@@ -343,132 +343,165 @@ impl CheckpointSignatureMessage {
...
+    fn enumerate_transactions(
+        &self,
+        ckpt: &CheckpointSummary,
+    ) -> impl Iterator<Item = (u64, ExecutionDigests)> + '_;
+
+    fn into_inner(self) -> Vec<ExecutionDigests>;
+
+    fn inner(&self) -> Vec<ExecutionDigests>;
+
+    fn size(&self) -> usize;
```
thibault-martinez: I think these 3 are good SDK candidates

semenov-vladyslav (approved):

muXxer:
IMO we should keep the versioned types. Can you add them to the SDK?

crates/iota-types/src/messages_checkpoint.rs
DaughterOfMars: done

crates/iota-types/src/messages_checkpoint.rs
DaughterOfMars: done

thibault-martinez:
You're probably on it already but you need to update the rev again

crates/iota-core/tests/staged/iota.yaml
thibault-martinez, line 1: May need to regen this

crates/iota-core/tests/staged/iota.yaml
```diff
@@ -130,19 +130,14 @@ CheckpointContents:
...
-          TYPENAME: CheckpointContentsV1
+          TUPLE:
+            - SEQ:
+                TYPENAME: Digests
+            - SEQ:
+                SEQ: BYTES
 CheckpointContentsDigest:
   NEWTYPESTRUCT:
     TYPENAME: Digest
-CheckpointContentsV1:
```
thibault-martinez, line 137: Just curious as to why this is removed if we have it versioned in the SDK

DaughterOfMars: Well we removed those types. Not sure tho, maybe we need to trace it?

DaughterOfMars: I tried manually tracing the inner type but it didn't change anything

thibault-martinez: What do you mean removed?

DaughterOfMars: Well some of the types don't exist anymore in the SDK.

thibault-martinez: Ok but CheckpointContentsV1 definitely does exist

DaughterOfMars: Oh, I know why it's gone. It's because it's based on the serialization and that doesn't name a struct inside the variant. It's just the V1 variant in the outer enum.

crates/iota-core/src/generate_format.rs
```diff
@@ -531,7 +549,9 @@ fn get_registry() -> Result<Registry> {
     tracer.trace_type::<UnchangedSharedKind>(&samples).unwrap();
     tracer.trace_type::<TransactionEffects>(&samples).unwrap();
 
-    tracer.trace_type::<CheckpointContents>(&samples).unwrap();
+    // `CheckpointContents` (SDK type) is registered via its value sample above:
+    // its `UserSignature`s serialize as opaque bytes, which the tracer cannot
+    // synthesize a scheme flag for, so `trace_type` cannot be used here.
```
thibault-martinez: I'd remove this, nice to know in the context of the PR but shouldn't stay in the code

crates/iota-core/tests/staged/iota.yaml
thibault-martinez: Alrighty

thibault-martinez (approved):

sergiupopescu199 (approved):

crates/iota-core/src/generate_format.rs
DaughterOfMars: I dunno, I think these comments here are really good for future maintainers

muXxer (approved):

crates/iota-core/src/generate_format.rs
thibault-martinez: It's not, it's just AI explaining why it moved it above, there is no reason to have this comment since it's gone. Plus it's literally a copy of the comment L236.

valeriyr (approved):

crates/iota-core/src/generate_format.rs
DaughterOfMars: Oh, I see 😆 yeah bad AI

thibault-martinez: 🤖

PR #12193, opened 2026-07-09
fix(iota-node): abort validator startup when the randomness manager cannot be built

piotrm50:
# Description of change

If `RandomnessManager::try_new` returned `None` on a validator, startup silently continued without a randomness manager and the first consensus commit — which unconditionally expects one — then panicked later, far from the real cause. This hardens startup to fail immediately and loudly at the point of failure instead.

- Abort loudly (`fatal!`) at the point the randomness manager is built, so a validator that cannot construct one fails immediately at the real cause instead of panicking later on the first commit. This point is shared by initial startup and reconfiguration, so the abort is loud on both.
- Convert `try_new`'s internal `expect`s on committee-/protocol-config-derived values into the existing error-and-`None` failure path, and surface the offending value in the log.

fixes iotaledger/iota-private#456

Added `try_new_returns_none_when_reduction_lower_bound_exceeds_u16`, which exercises the `None` branch that the startup abort depends on.

### DIFF
--- a/crates/iota-node/src/lib.rs
+++ b/crates/iota-node/src/lib.rs
@@ -23,7 +23,7 @@ use arc_swap::ArcSwap;
 use futures::future::BoxFuture;
 pub use handle::IotaNodeHandle;
 use iota_archival::{reader::ArchiveReaderBalancer, writer::ArchiveWriter};
-use iota_common::debug_fatal;
+use iota_common::{debug_fatal, fatal};
 use iota_config::{
     ConsensusConfig, NodeConfig,
     node::{DBCheckpointConfig, RunWithRange},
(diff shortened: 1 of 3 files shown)


crates/iota-node/src/lib.rs
```diff
@@ -1533,18 +1533,29 @@ impl IotaNode {
...
-        if let Some(randomness_manager) = randomness_manager {
-            epoch_store
-                .set_randomness_manager(randomness_manager)
-                .await?;
-        }
+        .await
+        {
+            Ok(randomness_manager) => randomness_manager,
+            Err(err) => {
+                fatal!(
```
bingyanglin, line 1550: `set_randomness_manager(...).await?`, also leaves you with no randomness manager if it fails, and that one just propagates. Should that be fatal too, or is a storage error there meant to be recoverable?

roman1e2f5p8s (approved):

crates/iota-node/src/lib.rs
piotrm50: The error here comes from `start_dkg()` inside set_randomness_manager - the manager itself is always installed before returning. So this failure can't leave us without a randomness manager, and a failed DKG start is recoverable (`advance_dkg` tolerates it). Keeping `?` like the other storage errors in this function.

bingyanglin (approved):

PR #12194, opened 2026-07-09
fix(iota-framework): prune stale cross-epoch report records before advance_epoch

piotrm50:
# Description of change

Validator report records persist across epochs but were previously cleaned only when a validator fully departs the active set, not on committee rotation. `compute_slashed_validators` assumed every recorded reporter/reportee is a current committee member. Under `ValidatorSetV2` the committee can be a subset of the active set, so that assumption no longer holds and a stale record could cause `compute_slashed_validators` to reject valid state during `advance_epoch`.

- Prune report records referencing validators no longer in the committee (both reportee entries and reporter memberships) in `advance_epoch`, before `compute_slashed_validators`, instead of asserting. Reports between validators that remain in the committee are preserved.
- The behavior ships with the iota-system framework upgrade gated on protocol version 31 (already introduced on develop); `bytecode_snapshot/31` and the genesis snapshot are regenerated.

fixes iotaledger/iota-private#457

Added `test_advance_epoch_prunes_report_records_for_rotated_out_committee_validator`: committee size < validator count, files reports (reporter-rotates-out, reportee-rotates-out, and a control between two validators that stay), shifts stake so a reporter rotates out of the committee but stays active, crosses two epoch boundaries, and asserts `advance_epoch` completes while the control report is kept.


muXxer (approved):
Only approved the new snapshot file.

miker83z (approved):

PR #12199, opened 2026-07-09
feat(starfish): enable consensus gRPC protective limits by default

### LINKED ISSUE #12198: starfish(grpc): enable per-peer admission control on the consensus gRPC server by default
The consensus gRPC server gained per-(peer, RPC group) admission control in #11970, but it is inert by default and only takes effect when a node sets `CONSENSUS_GRPC_PROTECTIVE_LIMITS=1`. In practice the network runs without these bounds unless each operator opts in individually.

Promote the protective preset to on-by-default, retaining `CONSENSUS_GRPC_PROTECTIVE_LIMITS` as an opt-out (`0`/`false`) so a node can disable it in the field without a redeploy.

**Key points:**
- Without the bounds, a validator applies no per-peer limits on the consensus gRPC server. In the worst case an mTLS-authenticated committee member could open a disproportionate number of concurrent requests/streams and pressure an honest node's serving resources.
- These are local operational parameters, not protocol-gated: heterogeneous values across validators are safe and there is no consensus-safety impact, so the default can flip without a protocol version bump.
- Behaviour is observable via `consensus_inbound_admission_in_use{group}` and `consensus_inbound_admission_rejected{group}`.

polinikita:
# Description of change

After testing on the Testnet, we promote the consensus gRPC per-peer admission control and inbound resource bounds (added in #11970) from opt-in to on-by-default.

- The protective preset is now applied by default in the consensus manager.
- `CONSENSUS_GRPC_PROTECTIVE_LIMITS` becomes an opt-out: setting it to `0`/`false` disables the bounds so a node can roll back in the field without a redeploy.
- Local operational parameters only — no protocol version bump, and heterogeneous values across validators remain safe. The type-level default and the parameters snapshot are unchanged.

Fixes #12198

### Release Notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Validators now apply the consensus gRPC per-peer admission control and inbound resource bounds by default (full nodes are unaffected — they run no consensus server). Set `CONSENSUS_GRPC_PROTECTIVE_LIMITS=0` (or `false`) to opt out.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] gRPC:

### DIFF
--- a/crates/iota-core/src/consensus_manager/starfish_manager.rs
+++ b/crates/iota-core/src/consensus_manager/starfish_manager.rs
@@ -130,16 +130,18 @@ impl ConsensusManagerTrait for StarfishManager {
             .find(|(_, a)| a.protocol_key == own_protocol_key)
             .expect("Own authority should be among the consensus authorities!");
 
-        // Opt into the protective consensus gRPC resource limits via an
-        // environment variable, for gradual rollout on a subset of validators
-        // without changing the default (inert) behaviour.
+        // Apply the protective consensus gRPC resource limits by default. A node
+        // can opt out via the environment variable without a redeploy; these are
+        // local operational parameters, so heterogeneous values across validators
+        // are safe.
         let parameters = {
             let mut p = parameters;
             if matches!(
                 std::env::var("CONSENSUS_GRPC_PROTECTIVE_LIMITS").as_deref(),
-                Ok("1") | Ok("true")
+                Ok("0") | Ok("false")
             ) {
-                info!("Applying protective consensus gRPC limits for validator {own_index}");
+                info!("Consensus gRPC protective limits disabled for validator {own_index}");
+            } else {
                 p.tonic.apply_protective();
             }
             p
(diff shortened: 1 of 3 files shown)


muXxer (approved):

VorobyevIlya (commented):
1. The docs in `parameters.rs` still describe the preset as opt-in: `AdmissionParameters` says "leaving the mechanism inert until an operator opts in" / "the opt-in protective preset", and `apply_protective` says "the bounds never change behaviour unless explicitly enabled". With the preset now applied by default, these sentences seem misleading — same for the "must survive opting in" comment in `parameters_test.rs`.

2. `apply_protective` overwrites `max_concurrent_streams`, `request_timeout`, `max_inbound_message_size`, and `admission`, so with it running by default, values an operator sets for these fields in the node YAML are silently replaced unless `CONSENSUS_GRPC_PROTECTIVE_LIMITS=0` is set. The fields are new and default-inert, so probably nobody is affected today — but is "the env var wins over the YAML" the intended behavior here?

VorobyevIlya (approved):

PR #12202, opened 2026-07-09
feat(node): Add admin endpoint for Prometheus metrics filter

### LINKED ISSUE #12068: node: add admin endpoint for prometheus filters

jkrvivian:
# Description of change

Adds runtime control over which collected metrics the node exposes, via the admin server:

- `GET /metrics/filters` returns the filter currently in effect.
- `POST /metrics/filters` sets the runtime override from the `filter` query parameter; patterns may be group names from the `metrics.groups` config section (`default` and `hardware` included) or raw `METRICS_FILTER`-style patterns. An invalid directive rejects the whole update with a `400` naming every offending token. Each POST starts from the startup directives again rather than stacking on the previous override.
- `POST /metrics/filters/reset` drops the runtime override, restoring the startup behavior.

The directives from the three sources (config, `METRICS_FILTER` env var, runtime override) merge into one set: a higher-precedence directive replaces the lower source's directive with the same pattern, and otherwise the most specific matching pattern decides each metric (a metric-name match over a module match, then the longest pattern; among equal patterns the last one) — replacing the previous order-sensitive last-match-wins rule. A bare level (e.g. `trace`) makes its source replace the lower-precedence directives entirely — `METRICS_FILTER=trace` exposes everything regardless of the config, keeping the existing benchmarking/debugging escape hatch — while `default=LEVEL` is the surgical spelling that changes only the level for otherwise-unmatched metrics.

The `METRICS_FILTER` env var gets the same treatment: it now accepts group names and replaces same-pattern config directives instead of being concatenated onto the config string.

Config changes:

- `metrics.groups` gains a free-form `overrides` map for module paths, metric names, or group names, including ones already covered by a named group (e.g. `iota_core::checkpoints: trace`); the most specific pattern wins over the group directives, so a single module or metric can diverge from its group's level, and a group-name key expands like the same directive in `METRICS_FILTER` or the admin endpoint, replacing the group field's level. Group-name typo protection stays intact.
- The `hardware` group is now a regular group: its collector goes through the same gather-time filter as any other metric, so it can be toggled at runtime (previously decided once at startup). A hardware collector construction failure no longer prevents node startup.

Filtering is enforced at gather time: metrics registered through the wrapper macros are wrapped in a collector that consults the shared filter on every scrape, so exposure changes need no registration bookkeeping. Hidden metrics keep collecting and reappear with their up-to-date values.

```
$ curl 'http://127.0.0.1:1337/metrics/filters'
$ curl -X POST 'http://127.0.0.1:1337/metrics/filters?filter=consensus=debug,storage=trace'
$ curl -X POST 'http://127.0.0.1:1337/metrics/filters/reset'
```

## How our metrics level works
config, env, and runtime each contribute directives, and a higher source's directive replaces a lower one with the same pattern. Each metric is decided by its most specific matching directive — metric name beats module, longer beats shorter, later beats earlier. Exception: a bare level in a higher source discards the lower sources entirely. 

Closes #12068

### Release Notes

- [x] Nodes (Validators and Full nodes): the Prometheus metrics exposure filter can be inspected and overridden at runtime through the admin server; `metrics.groups` gains a free-form `overrides` map for per-module, per-metric, or per-group levels, and the `hardware` group is now toggleable at runtime (`hardware: off` hides the collector instead of skipping its registration at startup).

### DIFF
--- a/crates/iota-config/data/fullnode-template.yaml
+++ b/crates/iota-config/data/fullnode-template.yaml
@@ -40,6 +40,6 @@ authority-store-pruning-config:
 #     epoch: warn
 #     runtime: warn
 #     hardware: warn
-#     # free-form overrides for modules/metrics not covered by a group:
+#     # free-form overrides for modules/metrics:
 #     overrides:
 #       "iota_core::checkpoints": trace
--- a/crates/iota-config/data/fullnode-template.yaml
+++ b/crates/iota-config/data/fullnode-template.yaml
@@ -39,4 +39,7 @@ authority-store-pruning-config:
 #     rpc: warn
 #     epoch: warn
 #     runtime: warn
-#     hardware: warn # set to `off` to disable hardware metrics; takes effect at startup only
+#     hardware: warn
+#     # free-form overrides for modules/metrics not covered by a group:
+#     overrides:
+#       "iota_core::checkpoints": trace
--- a/crates/iota-metrics/src/lib.rs
+++ b/crates/iota-metrics/src/lib.rs
@@ -582,6 +582,11 @@ impl RegistryService {
         self.default_registry.clone()
     }
 
+    /// Returns the metrics filter shared by the service's registries.
+    pub fn filter(&self) -> Arc<Filter> {
+        self.filter.clone()
+    }
+
     // Creates a new registry that shares the service's metric filter. Prefer
     // this over `Registry::new()`/`Registry::new_custom()` for registries added
     // via `add`, so that the configured filter applies to their metrics too.
--- a/crates/iota-metrics/src/lib.rs
+++ b/crates/iota-metrics/src/lib.rs
@@ -560,8 +560,6 @@ pub struct RegistryService {
     default_registry: Registry,
     registries_by_id: Arc<DashMap<Uuid, Registry>>,
     filter: Arc<Filter>,
-    /// Serializes filter updates and the reconciles they trigger.
-    reconcile_lock: Arc<Mutex<()>>,
 }
 
 impl RegistryService {
@@ -572,7 +570,6 @@ impl RegistryService {
             filter: default_registry.filter(),
             default_registry,
             registries_by_id: Arc::new(DashMap::new()),
-            reconcile_lock: Arc::new(Mutex::new(())),
         }
     }
 
@@ -605,18 +602,13 @@ impl RegistryService {
     // registry - we expected a removal to happen explicitly.
     pub fn add(&self, registry: Registry) -> RegistryID {
         let registry_id = Uuid::new_v4();
-        let _guard = self.reconcile_lock.lock();
         if self
             .registries_by_id
-            .insert(registry_id, registry.clone())
+            .insert(registry_id, registry)
             .is_some()
         {
             panic!("Other Registry already detected for the same id {registry_id}");
         }
-        // The registry may have been populated before a filter change whose
-        // reconcile pass could not see it yet; re-evaluate it now so its
-        // membership matches the filter currently in effect.
-        registry.reconcile();
 
         registry_id
     }
@@ -644,36 +636,26 @@ impl RegistryService {
         self.get_all().iter().flat_map(|r| r.gather()).collect()
     }
 
-    /// Sets the runtime override on the shared filter and reconciles every
-    /// registry. Rejects the whole update if any directive is invalid.
-    /// `directives` drive matching; `display` (e.g. the group-form input before
-    /// expansion) is what the admin endpoint echoes back.
+    /// Sets the runtime override on the shared filter; every registry's next
+    /// gather exposes metrics per the new directives. Rejects the whole
+    /// update if any directive is invalid. `directives` drive matching;
+    /// `display` (e.g. the group-form input before expansion) is what the
+    /// admin endpoint echoes back.
     pub fn set_runtime_filter(
         &self,
         directives: &str,
         display: &str,
     ) -> std::result::Result<(), String> {
-        let _guard = self.reconcile_lock.lock();
         self.filter
             .set_runtime_filter(prometheus_filtered::FilterSource::with_display(
                 directives, display,
-            ))?;
-        self.reconcile_all();
-        Ok(())
+            ))
     }
 
-    /// Drops the runtime override on the shared filter and reconciles every
-    /// registry back to its startup exposure.
+    /// Drops the runtime override on the shared filter, restoring every
+    /// registry to its startup exposure.
     pub fn reset_runtime_filter(&self) {
-        let _guard = self.reconcile_lock.lock();
         self.filter.reset_runtime_filter();
-        self.reconcile_all();
-    }
-
-    fn reconcile_all(&self) {
-        for registry in self.get_all() {
-            registry.reconcile();
-        }
     }
 }
 
@@ -874,7 +856,7 @@ mod tests {
     }
 
     #[test]
-    fn set_runtime_filter_reconciles_all_registries() {
+    fn set_runtime_filter_applies_to_all_registries() {
         use std::sync::Arc;
 
         use prometheus_filtered::{Filter, MetricLevel};
@@ -913,7 +895,8 @@ mod tests {
             ["default_g_warn", "second_g_warn"]
         );
 
-        // One runtime update reconciles every registry in the service.
+        // One runtime update changes the exposure of every registry in the
+        // service.
         registry_service
             .set_runtime_filter("iota_metrics=debug", "iota_metrics=debug")
             .unwrap();
--- a/crates/iota-metrics/src/lib.rs
+++ b/crates/iota-metrics/src/lib.rs
@@ -637,18 +637,15 @@ impl RegistryService {
     }
 
     /// Sets the runtime override on the shared filter; every registry's next
-    /// gather exposes metrics per the new directives. Rejects the whole
-    /// update if any directive is invalid. `directives` drive matching;
-    /// `display` (e.g. the group-form input before expansion) is what the
-    /// admin endpoint echoes back.
-    pub fn set_runtime_filter(
-        &self,
-        directives: &str,
-        display: &str,
-    ) -> std::result::Result<(), String> {
+    /// gather exposes metrics per the new directives. `filter` may use
+    /// [`MetricGroups`] names, which are expanded for matching while the
+    /// string is echoed back as given. Rejects the whole update if any
+    /// directive is invalid.
+    pub fn set_runtime_filter(&self, filter: &str) -> std::result::Result<(), String> {
+        let expanded = MetricGroups::expand_directives(filter)?;
         self.filter
             .set_runtime_filter(prometheus_filtered::FilterSource::with_display(
-                directives, display,
+                &expanded, filter,
             ))
     }
 
@@ -898,7 +895,7 @@ mod tests {
         // One runtime update changes the exposure of every registry in the
         // service.
         registry_service
-            .set_runtime_filter("iota_metrics=debug", "iota_metrics=debug")
+            .set_runtime_filter("runtime=debug")
             .unwrap();
         assert_eq!(
             gathered_names(&registry_service),
@@ -909,13 +906,6 @@ mod tests {
                 "second_g_warn"
             ]
         );
-
-        // An invalid update is rejected without touching the current override.
-        let err = registry_service
-            .set_runtime_filter("bogus=nope", "bogus=nope")
-            .unwrap_err();
-        assert!(err.contains("bogus=nope"), "unexpected error: {err}");
-        assert_eq!(gathered_names(&registry_service).len(), 4);
 
         // Reset restores the startup exposure across all registries.
         registry_service.reset_runtime_filter();
--- a/crates/iota-metrics/src/lib.rs
+++ b/crates/iota-metrics/src/lib.rs
@@ -857,4 +857,73 @@ mod tests {
         assert_eq!(metric_1.name(), "iota_counter_2");
         assert_eq!(metric_1.help(), "counter_2_desc");
     }
+
+    #[test]
+    fn set_runtime_filter_reconciles_all_registries() {
+        use std::sync::Arc;
+
+        use prometheus_filtered::{Filter, MetricLevel};
+
+        fn gathered_names(service: &RegistryService) -> Vec<String> {
+            let mut names: Vec<_> = service
+                .gather_all()
+                .iter()
+                .map(|f| f.name().to_owned())
+                .collect();
+            names.sort();
+            names
+        }
+
+        // Both registries share the service's filter, which starts at `warn`
+        // for this module, hiding the default-level (`debug`) gauges.
+        let filter = Arc::new(Filter::parse("iota_metrics=warn"));
+        let default_registry =
+            Registry::new_custom(Some("default".to_string()), None, Some(filter)).unwrap();
+        let registry_service = RegistryService::new(default_registry.clone());
+        let second_registry = registry_service
+            .new_registry_custom(Some("second".to_string()), None)
+            .unwrap();
+        registry_service.add(second_registry.clone());
+
+        for registry in [&default_registry, &second_registry] {
+            prometheus_filtered::register_int_gauge_with_registry!(
+                "g_warn", "h", registry; MetricLevel::Warn
+            )
+            .unwrap();
+            prometheus_filtered::register_int_gauge_with_registry!("g_debug", "h", registry)
+                .unwrap();
+        }
+        assert_eq!(
+            gathered_names(&registry_service),
+            ["default_g_warn", "second_g_warn"]
+        );
+
+        // One runtime update reconciles every registry in the service.
+        registry_service
+            .set_runtime_filter("iota_metrics=debug", "iota_metrics=debug")
+            .unwrap();
+        assert_eq!(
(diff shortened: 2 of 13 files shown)


crates/iota-node/src/main.rs
```diff
@@ -74,15 +74,19 @@ fn main() {
...
     let metric_groups = config
         .metrics
         .as_ref()
         .and_then(|m| m.groups.clone())
         .unwrap_or_default();
-    let metrics_filter =
-        prometheus_filtered::Filter::resolve(Some(&metric_groups.to_filter_string()));
+    let env_filter = std::env::var("METRICS_FILTER")
+        .ok()
+        .map(|env| iota_metrics::MetricGroups::expand_startup_directives(&env).unwrap_or(env));
```
bingyanglin: One bad directive in METRICS_FILTER makes unwrap_or(env) silently skip group expansion for all the others, with no error visible. Maybe report the error to stderr or drop only the bad directive?

crates/iota-node/src/admin.rs
```diff
@@ -582,3 +606,47 @@ async fn traffic_control(
         Err(err) => (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()),
     }
 }
+
+async fn get_metrics_filter(State(state): State<Arc<AppState>>) -> (StatusCode, String) {
+    let filter = state.node.metrics_filter();
+    let runtime = filter
```
bingyanglin: Unset runtime prints `(none)` but unset env/config print empty. Maybe `(none)` for all three?

crates/telemetry-subscribers/observability_guides.md
```diff
@@ -135,7 +135,9 @@ is set to `info`, so only spans with level `error` and `info` will be sent as me
        runtime: debug
    ```
 
-   or via the `METRICS_FILTER` environment variable, which overrides the config:
+   or via the `METRICS_FILTER` environment variable, whose directives override
```
bingyanglin, line 138: Group expansion of `METRICS_FILTER` only happens in `iota-node`, not other binaries. Maybe qualify this line with "for iota-node"?

crates/iota-metrics/src/metric_groups.rs
semenov-vladyslav:
```suggestion
    fn modules_for_group(group: &str) -> Option<&'static [&'static str]> {
```

bingyanglin (approved):

crates/telemetry-subscribers/observability_guides.md
semenov-vladyslav: METRICS_FILTER is independent in a way from metric groups. Other binary still handle `METRICS_FILTER` env var (but not metrics groups).

crates/prometheus-filtered/src/lib.rs
semenov-vladyslav: Remove `core` module, no need for wrappers now.

semenov-vladyslav (changes requested):
- Having three filter layers - config, env, and runtime - seems to be overcomplicated and confusing, especially in the admin get endpoint: if all three layers had same group or module level declared, then it will be shown three times. Why not just keep a single layer that applies to a registry or registry service? IMO, it would be simpler in the implementation and for users.
- Filter layer `raw` value is not "raw", but expanded list of directives. Eg. if a filter is `"consensus=info"` then `raw` will contain `"starfish_core=info,iota_core::consensus_adapter=info,iota_core::consensus_manager=info,iota_core::consensus_validator=info,iota_core::epoch::consensus_store_pruner=info"`. Again, it's not very convenient in the admin endpoint. Why not keep group directives internally?
- Node config for metrics allows to configure only metric groups, but not individual module or metric overrides. It is only possible via env var or runtime overrides. Why not allow free keys in the config? That would allow for module and metric directives.
- Why are hardware metrics handled separately (`reject_hardware`)? After all, hardware metrics are registered as any other prometheus metric - via `register` method.
- Currently, `prometheus_filtered::Registry` API has diverged from `prometheus::Registry`, in particular registering metrics in macros is done via `record_collector` method instead of `register`. `record_collector` - a public function - in turn allows passing arbitrary value for `result` argument. This breaks encapsulation, allowing users a possibility to misuse.

crates/iota-node/src/admin.rs
```diff
@@ -582,3 +610,49 @@ async fn traffic_control(
...
+    fn or_none(s: String) -> String {
+        if s.is_empty() { "(none)".into() } else { s }
+    }
+    let filter = state.node.metrics_filter();
+    (
+        StatusCode::OK,
+        format!(
+            "metrics exposure filter:\n\
+             {}\n",
+            or_none(filter.filter_string()),
```
bingyanglin: The route comment promises current + startup directives, but only the current filter is printed. Maybe add a line with `startup_filter_string()`?

crates/iota-metrics/src/lib.rs
```diff
@@ -626,6 +625,36 @@ impl RegistryService {
...
+    }
+
+    /// Drops the runtime override on the shared filter and reconciles every
+    /// registry back to its startup exposure.
+    pub fn reset_runtime_filter(&self) {
+        self.filter.reset_runtime_filter();
+        self.reconcile_all();
+    }
+
+    fn reconcile_all(&self) {
```
bingyanglin: A registry populated before `add()` misses a filter change in that window, and stays wrong until the next one. Maybe `registry.reconcile()` inside `add()`?

crates/iota-metrics/src/lib.rs
```diff
@@ -626,6 +625,36 @@ impl RegistryService {
     pub fn gather_all(&self) -> Vec<prometheus_filtered::proto::MetricFamily> {
         self.get_all().iter().flat_map(|r| r.gather()).collect()
     }
+
+    /// Sets the runtime override on the shared filter and reconciles every
+    /// registry. Rejects the whole update if any directive is invalid.
+    /// `directives` drive matching; `display` (e.g. the group-form input before
+    /// expansion) is what the admin endpoint echoes back.
+    pub fn set_runtime_filter(
```
bingyanglin: Concurrent POSTs can leave membership matching the older filter, since each reconcile snapshots separately. Maybe a mutex around set + `reconcile_all`?

crates/prometheus-filtered/src/lib.rs
```diff
@@ -652,26 +451,65 @@ impl Registry {
...
+    pub fn reconcile(&self) {
+        // Snapshot the directives once so the whole pass sees a consistent
+        // filter and avoids re-locking them per metric.
+        let runtime = self.filter.runtime();
+        let registered = self.registered.read().unwrap();
+        for (name, recorded) in registered.iter() {
+            let want = threshold(&runtime.directives, name, &recorded.module)
+                >= recorded.level.verbosity();
+            if want {
+                let _ = self.register(recorded.collector.clone_box());
```
bingyanglin: A real register failure here is silent, and the metric never reappears. Maybe `warn!` on any error except `AlreadyReg`?

crates/iota-metrics/src/metric_groups.rs
```diff
@@ -262,32 +284,87 @@ impl MetricGroups {
...
+                Err(err) => errors.push(err),
+            }
+        }
+        (directives.join(","), errors)
+    }
+
+    fn expand_directive(part: &str) -> Result<Vec<String>, String> {
+        prometheus_filtered::validate_directive(part)?;
+        let (pattern, level) = prometheus_filtered::split_directive(part);
+        if pattern == "default" {
```
bingyanglin: `default=xxx` in a POST wipes all startup directives, not just the catch-all. Is it intended?

bingyanglin (approved):

crates/iota-config/data/fullnode-template.yaml
```diff
@@ -39,4 +39,7 @@ authority-store-pruning-config:
 #     rpc: warn
 #     epoch: warn
 #     runtime: warn
-#     hardware: warn # set to `off` to disable hardware metrics; takes effect at startup only
+#     hardware: warn
+#     # free-form overrides for modules/metrics not covered by a group:
```
semenov-vladyslav:
```suggestion
#     # free-form overrides for modules/metrics:
```
Overrides can be covered by a group: `iota_core::checkpoints` is covered by `checkpoints` group.

setups/fullnode/fullnode-devnet.yaml
```diff
@@ -57,4 +57,7 @@ state-archive-read-config:
 #     rpc: warn
 #     epoch: warn
 #     runtime: warn
-#     hardware: warn # set to `off` to disable hardware metrics; takes effect at startup only
+#     hardware: warn
+#     # free-form overrides for modules/metrics not covered by a group:
```
semenov-vladyslav:
```suggestion
#     # free-form overrides for modules/metrics:
```

setups/fullnode/fullnode-mainnet.yaml
```diff
@@ -61,4 +61,7 @@ state-archive-read-config:
 #     rpc: warn
 #     epoch: warn
 #     runtime: warn
-#     hardware: warn # set to `off` to disable hardware metrics; takes effect at startup only
+#     hardware: warn
+#     # free-form overrides for modules/metrics not covered by a group:
```
semenov-vladyslav:
```suggestion
#     # free-form overrides for modules/metrics:
```

setups/fullnode/fullnode-testnet.yaml
```diff
@@ -56,4 +56,7 @@ state-archive-read-config:
 #     rpc: warn
 #     epoch: warn
 #     runtime: warn
-#     hardware: warn # set to `off` to disable hardware metrics; takes effect at startup only
+#     hardware: warn
+#     # free-form overrides for modules/metrics not covered by a group:
```
semenov-vladyslav:
```suggestion
#     # free-form overrides for modules/metrics:
```

crates/iota-metrics/src/metric_groups.rs
```diff
@@ -130,13 +142,17 @@ pub struct MetricGroups {
     /// Modules: `iota_metrics` (except the `hardware` and `p2p` group
     /// submodules), `telemetry_subscribers`.
     pub runtime: MetricLevel,
-    /// Host hardware metrics (CPU / memory / disk). Registered as a collector,
-    /// so individual metrics cannot be level-filtered: `off` skips the whole
-    /// group and every other level registers it, at startup only.
+    /// Host hardware metrics (CPU / memory / disk). One collector, so the level
+    /// applies to the whole group rather than individual metrics: `off` hides
+    /// it and every other level exposes it.
```
semenov-vladyslav: This comment is confusing. No need to mention "one collector". A group metric level applies to the corresponding group, so "the level applies to the whole group" is true for all group metric levels, not just `hardware`. What is meant here for the user is that the level at which the hardware metrics are "registered" is warn (so "`off` hides it and every other level exposes it") and there is no way to override individual metrics. Although I don't see why, it's possible to split and register hardware metrics individually.

crates/iota-metrics/src/metric_groups.rs
```diff
@@ -130,13 +142,17 @@ pub struct MetricGroups {
...
     ///
     /// Rendered as an `iota_metrics::hardware_metrics` directive, the module
     /// where the collector is registered.
     pub hardware: MetricLevel,
+    /// Free-form overrides for module paths or metric names not covered by the
+    /// named groups.
```
semenov-vladyslav: Why "not covered by the named groups"? This is precisely useful to override group level to just enable/disable individual metrics. Is it really implemented so that only metrics _not covered_ by a named group is allowed here?

crates/iota-metrics/src/metric_groups.rs
```diff
@@ -253,41 +266,100 @@ impl MetricGroups {
...
         ]
-        .map(|(group, level)| {
+    }
+
+    /// Maps each group's configured level to the filter patterns it covers.
+    ///
+    /// Directive order does not matter: where one group's module prefix also
+    /// matches another group's submodules (e.g. `runtime`'s `iota_metrics`
+    /// covers the `p2p` and `hardware` groups'), the more specific pattern
+    /// wins.
```
semenov-vladyslav: Confusing description: it doesn't map group level to anything, there's neither filter patterns, nor directives - just module paths corresponding to each group, there's no matching. This comment probably belongs to its call site.
```suggestion
    /// List of metric level together with registered module paths corresponding to all groups.
```

crates/iota-node/src/lib.rs
```diff
@@ -2260,6 +2249,27 @@ impl IotaNode {
         self.randomness_handle.clone()
     }
 
+    /// Returns the metrics filter shared by the node's Prometheus registries.
+    pub fn metrics_filter(&self) -> Arc<prometheus_filtered::Filter> {
+        self.registry_service.default_registry().filter()
```
semenov-vladyslav:
```suggestion
        self.registry_service.filter()
```
Just expose `filter()` in `RegistryService`.

crates/iota-node/src/admin.rs
```diff
@@ -76,6 +77,28 @@ use crate::IotaNode;
...
+// Patterns may be group names from the `metrics.groups` config section
+// (`default` and `hardware` included) or raw METRICS_FILTER-style patterns.
+// The override is merged over the startup directives: an override directive
+// replaces the startup directive with the same pattern, and otherwise the
+// most specific matching pattern decides each metric, whichever source it
+// came from. Each POST
+// starts from the startup directives again rather than stacking on the
+// previous override.
+//
+//   $ curl -X POST 'http://127.0.0.1:1337/metrics/filters' -d 'consensus=off,typed_store=warn'
```
semenov-vladyslav: I would use query params as most other endpoints do.

crates/iota-node/src/main.rs
```diff
@@ -74,28 +74,45 @@ fn main() {
...
+            &metric_groups.to_filter_string(),
+            &metric_groups.to_display_string(),
+        ),
+        env_directives
+            .as_deref()
+            .zip(env_raw.as_deref())
+            .map(|(directives, display)| {
+                prometheus_filtered::FilterSource::with_display(directives, display)
+            }),
+    );
```
semenov-vladyslav: A lot of internal logic is exposed. Is there a way to make FilterSource private?

semenov-vladyslav (approved):
The implementation looks sound, although a bit overcomplicated. I would also clarify/simplify some comments a bit.

crates/iota-metrics/src/metric_groups.rs
jkrvivian: It works as the runtime or env filter, you can assign individual metrics.

crates/iota-node/src/main.rs
jkrvivian: Now it's only public to the crate.

bingyanglin (approved):

crates/iota-metrics/src/metric_groups.rs
semenov-vladyslav:
```suggestion
    fn modules_for_group(group: &str) -> Option<&'static [&'static str]> {
```

crates/iota-metrics/src/metric_groups.rs
```diff
@@ -35,11 +35,12 @@
...
 //!
-//! The `hw` hardware metrics are registered as a prometheus collector and
-//! so bypass the filtering macros entirely. They cannot be level-filtered
-//! individually: the `hardware` group's level is read once at registration
-//! time — [`MetricLevel::Off`] skips the whole group, any other level
-//! registers it.
+//! The `hw` hardware metrics are one collector rather than macro-registered
+//! metrics, so the whole group shares a single level: the `hardware` group's
+//! level exposes the collector (`off` hides it, any other level exposes it),
+//! and — like the other groups — it can be changed at runtime.
```
semenov-vladyslav:
```suggestion
//! The `hardware` metrics are grouped together as one collector and
//! registered with `warn` level, so the whole group shares a single level.
```
I think the sentence for `hardware` metrics could be simplified a bit like this.

crates/iota-node/src/admin.rs
```diff
@@ -76,6 +76,30 @@ use crate::IotaNode;
...
+// The override is merged over the startup directives: an override directive
+// replaces the startup directive with the same pattern, and otherwise the
+// most specific matching pattern decides each metric, whichever source it
+// came from. A bare level (e.g. `trace`) replaces the startup directives
+// entirely, exposing everything up to that level; `default=LEVEL` changes
+// only the level for the metrics no other directive matches. Each POST
+// starts from the startup directives again rather than stacking on the
+// previous override.
+//
+//   $ curl -X POST 'http://127.0.0.1:1337/metrics/filters?filter=consensus=off,typed_store=warn'
```
semenov-vladyslav, line 98: A bit unusual form for query params. Why not `curl -X POST 'http://127.0.0.1:1337/metrics/filters?consensus=off&typed_store=warn'`? With this syntax runtime metrics could be reset with `curl -X POST 'http://127.0.0.1:1337/metrics/filters`.

semenov-vladyslav (approved):

crates/iota-node/src/admin.rs
jkrvivian: I'll skip this one. The current design matches other endpoint: `/enable-tracing?filter=info&duration=10s `

VorobyevIlya (approved):

